PR #10210, opened 2022-02-08
[CPU] Execute shape_of path in FP32 for BF16 case

mandrono:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *78144*

PR #14911, opened 2023-01-03
[CPU] Execute shape_of path in FP32 for BF16 case

maxnick:
### Details:
Original PR https://github.com/openvinotoolkit/openvino/pull/10210
This PR attempts to address a problem with insufficient numerical resolution of bf16 numbers to process the results of the ShapeOf operation, especially when a floating point type cast was inserted in the original model.

### Tickets:
 - CVS-78144


maxnick:
@usstq , FYI

akladiev:

akladiev:

PR #14947, opened 2023-01-05
[GPU] Add binary_sub post-op for oneDNN

jade-cho:
### Tickets:
 - *94581*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/program_node.cpp
+++ b/src/plugins/intel_gpu/src/graph/program_node.cpp
@@ -520,6 +520,7 @@ dnnl::post_ops program_node::try_optimize_post_ops(dnnl::post_ops& p_ops, const 
             }
 
             case onednn_post_op_type::binary_add:
+            case onednn_post_op_type::binary_sub:
             case onednn_post_op_type::binary_mul:
             case onednn_post_op_type::binary_max:
             case onednn_post_op_type::binary_min:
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/src/graph/program_node.cpp
```diff
@@ -1076,6 +1078,10 @@ void program_node::init_onednn_primitive_attributes() {
                     post_ops.append_binary(dnnl::algorithm::binary_add, in_desc);
                     update_onednn_post_op_list(onednn_post_op_type::binary_add, dep_idx);
                 }
+            } else if (desc.typed_desc<eltwise>()->mode == eltwise_mode::sub) {
+                dnnl::memory::desc in_desc = onednn::layout_to_memory_desc(in);
+                post_ops.append_binary(dnnl::algorithm::binary_sub, in_desc);
+                update_onednn_post_op_list(onednn_post_op_type::binary_sub, dep_idx);
             } else {
```
isanghao, line 1085: I think we are missing some assertion logic here. What if mode is not `sum`, nor `sub`?

src/plugins/intel_gpu/src/graph/program_node.cpp
```diff
@@ -1073,6 +1079,11 @@ void program_node::init_onednn_primitive_attributes() {
                     attrs->set_output_scales(mask, {DNNL_RUNTIME_F32_VAL});
                     update_onednn_post_op_list(onednn_post_op_type::scale, dep_idx);
                 }
+            } else {
+                std::stringstream error_msg;
+                error_msg << "Unsupported eltwise mode: " << static_cast<int>(desc.typed_desc<eltwise>()->mode) << ". ";
+                error_msg << desc.desc->id << " is fused node of " + this->id() + ".";
+                OPENVINO_ASSERT(false, error_msg.str());
```
isanghao, line 1086: could you check that there is no diff from dryrun or partial run with this change?

jade-cho: No difference at dryrun test.

src/plugins/intel_gpu/src/graph/program_node.cpp
jade-cho: Added

byungilm (approved):

PR #14956, opened 2023-01-05
Fix ResultRename transformation to preserve output order

mvafin:
### Details:
 - *MO IR Reader doesn't preserve the output order because in case of `ResultRename` transformation is called names preserved in `outputs_order` list are changed*

### Tickets:
 - *10010*

### DIFF
--- a/tools/mo/openvino/tools/mo/back/ResultRename.py
+++ b/tools/mo/openvino/tools/mo/back/ResultRename.py
@@ -15,6 +15,7 @@ class ResultRename(BackReplacementPattern):
 
     def find_and_replace_pattern(self, graph: Graph):
         op_names = set()
+        result_names_map = dict()
         for node in graph.get_op_nodes():
             if node.has_valid('name'):
                 op_names.add(node['name'])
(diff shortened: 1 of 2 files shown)


tools/mo/openvino/tools/mo/back/ResultRename.py
```diff
@@ -46,4 +46,7 @@ def find_and_replace_pattern(self, graph: Graph):
                     log.warning("Tensor name for Result node with name {} wasn't found. "
                                 "Default renaming was used: {}".format(node.soft_get('name', node.id),
                                                                        result_name))
+                for idx in range(len(graph.outputs_order)):
+                    if graph.outputs_order[idx] == node['name']:
+                        graph.outputs_order[idx] = result_name
```
rkazants, line 51: I think it is better to create a dictionary prev_result_name -> new_result_name. After that you can walk though all graph.outputs_order only ones and rename them using the dictionary.

tools/mo/unit_tests/mo/back/ResultRename_test.py
```diff
@@ -45,12 +45,16 @@ def test_case4(self):
...
+        graph.outputs_order = ['result1', 'result2']
+
         ResultRename().find_and_replace_pattern(graph)
         res1_node = Node(graph, 'result1')
         res2_node = Node(graph, 'result2')
         self.assertTrue(res1_node['name'] == 'Op1_tensor')
         self.assertTrue(res2_node['name'] == 'Op2_tensor')
 
+        self.assertTrue(graph.outputs_order == ['Op1_tensor', 'Op2_tensor'])
+
```
rkazants, line 57: Can we have more interesting case with the order not necessarily topologically ordered. For example, ['result_a', 'result_d', 'result_b', 'result_c']

tools/mo/openvino/tools/mo/back/ResultRename.py
```diff
@@ -46,4 +46,7 @@ def find_and_replace_pattern(self, graph: Graph):
                     log.warning("Tensor name for Result node with name {} wasn't found. "
                                 "Default renaming was used: {}".format(node.soft_get('name', node.id),
                                                                        result_name))
+                for idx in range(len(graph.outputs_order)):
```
rkazants, line 49: Am I correct that due to this transformation `graph.outputs_order` gets useless since we lost original names of result nodes?

mvafin: Correct

rkazants (approved):

PR #13249, opened 2022-09-27
[nG] [IE] add support of mixed FP16/FP32 precision ov::Model for mixed inference

pavel-esir:
### Details:
This PR does the following 2 things:

- During offline phase when a constant is casted from fp32 to fp16 if proportion of values that do not fit into fp16 exceeds 
a certain threshold (less than 75%) then the whole constant is left in original fp32 precission. During subsequent run this prevents loosing accuracy (even on CPU). In most cases these constants eps with single value.

- It marks FP16 precission sensitive subgraphs to keep inference for them in FP16. Precission sensitive subgraphs include:

1. x/(y + eps) including it's twins like x*((y+eps)^-1) and so on

2. Exp -> [Squeeze|Unsqueeze|ArithmeticElementwise|etc.] -> ReduceSum -> [Squeeze|Unsqueeze|ArithmeticElementwise|etc.]

3. Normalization nodes: MVN, NormalizeL2. Since they contain some of the operations above: div with eps, ReduceSum and pow^2. The last can cause overflow in FP16 for some models, e.g. if input values are > 256 since 256^2 is > 65504 (FP16 max)!!!

Fortunately this does not cause any significant change in performance. Since most of the flops are spend for Convolutions/MatMuls. Validation on models affected by these changes showed degradation only on a single model: and root cause is in plugin side, our mixed precision marking just uncovered it.

`MarkSugraphsToKeepInMixedPrecision` - includes several passes which mark div with eps, exp->[...]->ReduceOp, and normalization nodes: such `MarkDivWithEpsToKeepInMixedPrecision`, `MarkExpReduceOpToKeepInMixedPrecision`
`AlignMixedFP32FP16Types` - inserts converts to avoid type inconsistency during validation.

Both these transformations are placed inside `ConvertCompressedToMixedPrecision`. Plugins with mixed FP16/FP32 precision can enable only need to switch on this pass.

### Validation:

- [x] performance validation: validation on a scope of affected models showed degradation only on a single model, created task for that. Also decided to keep track of regular validation report after merging to master. 
- [ ] accuracy validation: done manually for some models, decided to merge without accuracy and wait for regular validation report.

### Tickets:
 - xxx-91865, xxx-91866, xxx-92711

PR #14965, opened 2023-01-06
[FP16] call marking for mixed precision inside ConvertPrecision

pavel-esir:
### Details:
 - moved `MarkSugraphsToKeepInMixedPrecision` and `AlignMixedFP32FP16Types` inside `ConvertPrecision` to ease user experience
 - should be merged after https://github.com/openvinotoolkit/openvino/pull/13249
 
### Tickets:
 - xxx-99601


src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
```diff
@@ -221,27 +176,77 @@ void TransformationsPipeline::apply(std::shared_ptr<ov::Model> func) {
...
+                {ngraph::element::u64, ngraph::element::i32},
+                {ngraph::element::u16, ngraph::element::i32},
+                {ngraph::element::u32, ngraph::element::i32},
+                {ngraph::element::boolean, ngraph::element::u8},
+                {ngraph::element::i4, ngraph::element::i8},
+                {ngraph::element::u4, ngraph::element::u8},
+        };
+
+        manager.register_pass<ngraph::pass::Validate>();
+        manager.register_pass<ngraph::pass::ConvertPrecision>(int_and_f64_convert_precision_list);
```
vladimir-paramuzov, line 207: I think we wanted to keep these converts after common optimizations (as it is now) to minimize risk of breaking something else.

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -302,6 +307,14 @@ bool ov::pass::ConvertPrecision::run_on_model(const std::shared_ptr<ngraph::Func
         {opset10::Unique::get_type_info_static(), fuse_type_to_unique_v10},
         {opset8::RandomUniform::get_type_info_static(), fuse_type_to_random_uniform_v8}};
 
+    if (m_keep_precision_sensitive_in_fp32) {
+        pass::Manager manager(get_pass_config());
+        // Mark subgraphs with disable_fp16_compression to keep them in FP32
+        manager.register_pass<pass::MarkSugraphsToKeepInMixedPrecision>();
+        manager.register_pass<pass::AlignMixedFP32FP16Types>();
```
vladimir-paramuzov, line 314: As the passes above are required for fp32->fp16 conversion only, I think we may add a check that such pair exists in precisions list and don't run redundant transforms if not really needed.

pavel-esir: agreed. will do that

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
pavel-esir: Yes, you are right. At the moment when I was preparing PR I moved marking inside `ConvertPrecision` and float conversion from `CommonOptimizations` into `transformations_pipeline` it seemed not a big deal to move integer conversion as well. But you are right, let's be consistent with what we discussed. I'll move integer and int64 conversion back after `CommonOptimizations`

src/common/transformations/src/transformations/convert_precision.cpp
pavel-esir: done

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
pavel-esir: Returned back integer conversion where it was at current master: to the end of CommonOptimizations. Please take a look

vladimir-paramuzov (approved):
GPU plugin part LGTM

src/common/transformations/tests/common_optimizations/align_mixed_fp32_fp16_types_test.cpp
```diff
@@ -36,24 +35,21 @@ TEST_F(TransformationTestsF, align_mixed_fp16_fp32_1) {
         pass::Manager manager;
         manager.register_pass<pass::MarkSugraphsToKeepInMixedPrecision>();
         manager.register_pass<pass::AlignMixedFP32FP16Types>();
-        const precisions_array convert_precision_list{{ov::element::f32, ov::element::f16}};
-        type_to_fuse_map empty_type_to_fuse_map = {};
-        manager.register_pass<pass::ConvertPrecision>(convert_precision_list, empty_type_to_fuse_map, true);
```
pavel-esir, line 41: No need to call `ConvertPrecision` manually. Now `MarkSugraphsToKeepInMixedPrecision` and `AlignMixedFP32FP16Types` are called inside `ConvertPrecision`

src/common/transformations/tests/common_optimizations/align_mixed_fp32_fp16_types_test.cpp
```diff
@@ -36,24 +35,21 @@ TEST_F(TransformationTestsF, align_mixed_fp16_fp32_1) {
...
         manager.run_passes(model);
     }
 
     {
-        auto input_1 = make_shared<Parameter>(element::f16, Shape{1, 3, 224, 224});
+        auto input_1 = make_shared<Parameter>(element::f32, Shape{1, 3, 224, 224});
         auto convert_to_f32_1 = make_shared<Convert>(input_1, element::f32);
         auto exp_1 = make_shared<Exp>(convert_to_f32_1);
-        auto input_2 = make_shared<Parameter>(element::f16, Shape{1, 3, 224, 224});
```
pavel-esir, line 49: Type rewriting from `f32` to `f16` will be done inside `ConvertPrecision`. 

In `AlignMixedFP32FP16Types` we only insert `Convert`s for the future, so that when element types are changed (precision sensitive nodes remain in `f32` while others are casted to `f16`) inside `ConvertPrecision` there is will be no type inconsistency

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -17,7 +17,12 @@
 #include <vector>
 
 #include "itt.hpp"
+#include "openvino/pass/constant_folding.hpp"
+#include "openvino/pass/manager.hpp"
 #include "ov_ops/type_relaxed.hpp"
+#include "transformations/common_optimizations/align_mixed_fp32_fp16_types.hpp"
+#include "transformations/common_optimizations/mark_subgraphs_to_keep_in_mixed_precision.hpp"
+#include "transformations/enable_decompression_convert_constant_folding.hpp"
```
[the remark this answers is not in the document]
pavel-esir: I understand your reason but i think benefits of moving into a single file do not outweigh disadvantages: we will have a really huge `convert_precision.hpp` file where a lot of very different and tricky logics are included.

I've already seen transformations defined and called only inside other transformations. E.g. there is https://github.com/openvinotoolkit/openvino/blob/78995e9ac2582e80b849820d8e762076cd19ed89/src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp which is called only inside of `transpose_sinking_general.cpp` it could've been defined only in the one cpp but they opt to keep them distributed. So i think it's not a big problem, at the end for the user it would be enough just to call ConvertPrecision with an additional flag to make get mixed precision `ov::Model`

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
```diff
@@ -140,6 +139,63 @@ void TransformationsPipeline::apply(std::shared_ptr<ov::Model> func) {
...
+                }) != fp_convert_precision_list.end();
+
+                if (!has_valid_conversion) {
+                    fp_convert_precision_list.push_back(et_pair);
+                }
+            }
+        }
+
+        type_to_fuse_map empty_fuse_map = {};
+        manager.register_pass<ov::pass::Validate>();
```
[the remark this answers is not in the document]
pavel-esir: I suspect it was called just to be sure that graph came in validated state, this validate was previously in GPU pipeline, i didn't introduce it, i just kept it without removing

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
```diff
@@ -140,6 +139,63 @@ void TransformationsPipeline::apply(std::shared_ptr<ov::Model> func) {
...
+                case ov::element::f64: return device_info.supports_fp64;
+                case ov::element::bf16: return false;
+                default: return false;
+            }
+            return false;
+        };
+
+        const auto fallback_precision = ov::element::f32;
+        std::vector<ov::element::Type> fp_element_types = {
+                ov::element::f32,
```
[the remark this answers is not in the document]
pavel-esir: If we have `f64` unsupported on plugin this list specified to which type we should fallback. As far as i understood fallback to bigger precision `f64` will have no sence.
https://github.com/openvinotoolkit/openvino/blob/988847f559501c63c486f0e081fdafa1cba8c88b/src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp#L190-L206

I just moved these lines from below, i didn't create these lines

src/common/transformations/tests/common_optimizations/align_mixed_fp32_fp16_types_test.cpp
pavel-esir: Done. Added tests for `ConvertPrecision` with flag `skip_precision_sensitive=True`

src/common/transformations/tests/common_optimizations/convert_compressed_to_mixed_precision_test.cpp
```diff
@@ -1,140 +0,0 @@
...
-#include "transformations/init_node_info.hpp"
-#include "transformations/rt_info/decompression.hpp"
-
-using namespace testing;
-using namespace ov;
-
-TEST_F(TransformationTestsF, ConvertCompressedToMixedPrecision) {
-    {
-        auto input = std::make_shared<opset10::Parameter>(element::f32, Shape{1, 3, 12, 12});
-        auto const_weights =
```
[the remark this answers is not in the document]
pavel-esir: Done. Added tests for ConvertPrecision with flag skip_precision_sensitive=True

src/common/transformations/tests/common_optimizations/align_mixed_fp32_fp16_types_test.cpp
pavel-esir: I think we can do both. I added tests for `ConvertPrecision` with flag `skip_precision_sensitive=True`, but i think keeping tests for marking and alignment does also makes sense.

If we accidentally break `AlignMixedFP32FP16Types` or `ConvertPrecision` we will see only red tests for `ConvertPrecision` but for marking they will be green: such info can help us to localize place where we did something wrong. Like the essence of unit-tests - we have several steps and we test each step, not only final one.

rkazants (approved):

PR #13730, opened 2022-10-30
[GPU] Make generic logic to find formats from meme::desc

byungilm:
Signed-off-by: Min, Byungil <byungil.min@intel.com>

### Details:
 - *item1*
 - *...*

### Tickets:
 - 86189

PR #14973, opened 2023-01-06
[GPU] Use correct onednn target in unit tests

vladimir-paramuzov:
### Details:
 - Fix build issue caused by #13730 which blocks OV build w/o CPU plugin


byungilm:
I am sorry for the build failure of gpu_test. I uploaded a PR to fix it. 
https://github.com/openvinotoolkit/openvino/pull/14998

PR #14975, opened 2023-01-06
[do not review]Enable cc for legacy api

xufang-lisa:
### Details:
 - *Enable cc for legacy api*

### Tickets:
 - *98322*


akladiev:

PR #14979, opened 2023-01-06
[TF FE] Support TF2 Object detection models

rkazants:
**Details:** For support of OOB conversion of OD models (Faster RCNN, SSD models) several fixes were done for Select, BroadcastArgs, Slice, and Concat operations. Implement tests for each case

**Tickets:** 

### DIFF
--- a/src/frontends/tensorflow/src/helper_transforms/embedding_segments_feature_fusing.cpp
+++ b/src/frontends/tensorflow/src/helper_transforms/embedding_segments_feature_fusing.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2018-2022 Intel Corporation
+// Copyright (C) 2018-2023 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 //
 
@@ -103,10 +103,15 @@ ov::frontend::tensorflow::pass::EmbeddingSegmentSingleFeatureFusion::EmbeddingSe
     auto const_one = make_shared<Constant>(element::i32, Shape{1}, 1);
     auto new_subshape = make_shared<Broadcast>(const_one, num_new_axes);
     auto cond_shape = make_shared<ShapeOf>(tile, element::i32);
-    auto new_cond_shape = make_shared<Concat>(OutputVector{cond_shape, new_subshape}, 0);
+    // use extra dimensions in the begin to avoid concatenation of empty tensors that is not supported by Concat
+    auto const_1 = make_shared<Constant>(element::i32, Shape{1}, 1);
+    auto new_cond_shape = make_shared<Concat>(OutputVector{const_1, cond_shape, new_subshape}, 0);
 
     // prepare the condition to have the same rank as operands `x` and `y`
-    auto prep_cond = make_shared<Reshape>(tile, new_cond_shape, false);
+    auto prep_cond = make_shared<Reshape>(tile, new_cond_shape, false)->output(0);
+    // squeeze prep_cond by one extra dimension specially added
+    auto const_0 = make_shared<Constant>(element::i32, Shape{1}, 0);
+    prep_cond = make_shared<Squeeze>(prep_cond, const_0);
 
     auto select_pattern = make_shared<Select>(prep_cond, zeros_like, sparse_segment_op);
 
--- a/src/frontends/tensorflow/src/op/concat.cpp
+++ b/src/frontends/tensorflow/src/op/concat.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2018-2022 Intel Corporation
+// Copyright (C) 2018-2023 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 //
 
@@ -20,20 +20,21 @@ OutputVector translate_concat_op(const NodeContext& node) {
     // axis is the first input for Concat
     // and is the last input to ConcatV2
     default_op_checks(node, 2, {"Concat", "ConcatV2"});
-    auto input_size = node.get_input_size();
+    auto input_size = static_cast<int>(node.get_input_size());
 
     int64_t axis;
     OutputVector inputs;
 
     if (node.get_op_type() == "Concat") {
         std::vector<int64_t> axis_vector;
+        // const auto start_const = get_constant_from_source(start);
         get_const_input(node, 0, &axis_vector);
         TENSORFLOW_OP_VALIDATION(
             node,
             axis_vector.size() == 1,
             "Input model is incorrect: axis input for Concat operation must have exactly one element.");
         axis = axis_vector[0];
-        for (size_t input_idx = 1; input_idx < input_size; ++input_idx) {
+        for (int input_idx = 1; input_idx < input_size; ++input_idx) {
             inputs.push_back(node.get_input(input_idx));
         }
     } else if (node.get_op_type() == "ConcatV2") {
@@ -44,7 +45,7 @@ OutputVector translate_concat_op(const NodeContext& node) {
             axis_vector.size() == 1,
             "Input model is incorrect: axis input for Concat operation must have exactly one element.");
         axis = axis_vector[0];
-        for (size_t input_idx = 0; input_idx < input_size - 1; ++input_idx) {
+        for (int input_idx = 0; input_idx < input_size - 1; ++input_idx) {
             inputs.push_back(node.get_input(input_idx));
         }
     } else {
--- a/src/frontends/tensorflow/src/utils.hpp
+++ b/src/frontends/tensorflow/src/utils.hpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2018-2022 Intel Corporation
+// Copyright (C) 2018-2023 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 //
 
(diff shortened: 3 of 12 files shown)


src/frontends/tensorflow/src/op/concat.cpp
```diff
@@ -20,20 +20,21 @@ OutputVector translate_concat_op(const NodeContext& node) {
     // axis is the first input for Concat
     // and is the last input to ConcatV2
     default_op_checks(node, 2, {"Concat", "ConcatV2"});
-    auto input_size = node.get_input_size();
+    auto input_size = static_cast<int>(node.get_input_size());
```
itikhono, line 23: why was it changed?

src/frontends/tensorflow/src/helper_transforms/embedding_segments_feature_fusing.cpp
```diff
@@ -103,10 +103,15 @@ ov::frontend::tensorflow::pass::EmbeddingSegmentSingleFeatureFusion::EmbeddingSe
     auto const_one = make_shared<Constant>(element::i32, Shape{1}, 1);
     auto new_subshape = make_shared<Broadcast>(const_one, num_new_axes);
     auto cond_shape = make_shared<ShapeOf>(tile, element::i32);
-    auto new_cond_shape = make_shared<Concat>(OutputVector{cond_shape, new_subshape}, 0);
+    // use extra dimensions in the begin to avoid concatenation of empty tensors that is not supported by Concat
```
itikhono, line 106: can we just check that tensors are empty here?
anyway, it looks like a workaround, we need to solve this issue with empty tensors more generally.
does our specification describe the behavior of Concat op in this case?

src/frontends/tensorflow/src/utils.hpp
```diff
@@ -26,13 +26,21 @@ void set_node_name(const std::string& node_name, const std::shared_ptr<Node>& no
...
 
 template <typename T>
-void get_const_input(const NodeContext& node, int64_t input_index, std::vector<T>* vector) {
-    auto ng_input = node.get_input(static_cast<int>(input_index));
-    if (auto constant = std::dynamic_pointer_cast<opset8::Constant>(ng_input.get_node_shared_ptr())) {
+void get_const_input(const NodeContext& node, int input_index, std::vector<T>* vector) {
+    auto input_size = static_cast<int>(node.get_input_size());
+    auto node_name = node.get_name();
+    auto node_type = node.get_op_type();
+    FRONT_END_GENERAL_CHECK(0 <= input_index && input_index < input_size,
```
itikhono, line 33: do we have tensorflow frontend macro?
if not, I believe we need to create it to avoid using "[TensorFlow Frontend] Internal error: " everywhere

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -252,7 +252,7 @@ void FrontEnd::normalize(const std::shared_ptr<ov::Model>& function) const {
     manager.register_pass<pass::GRUBlockCellReplacer>();
 
     // TODO: reimplement TransposeSinking that does not corrupt filters for Convolution
-    manager.register_pass<ov::frontend::tensorflow::pass::TransposeSinking>();
+    // manager.register_pass<ov::frontend::tensorflow::pass::TransposeSinking>();
```
gkrivor, line 255: Did you miss a changing copyrights of this file in case it is changed?

src/frontends/tensorflow/src/op/concat.cpp
```diff
@@ -20,20 +20,21 @@ OutputVector translate_concat_op(const NodeContext& node) {
...
 
     int64_t axis;
     OutputVector inputs;
 
     if (node.get_op_type() == "Concat") {
         std::vector<int64_t> axis_vector;
+        // const auto start_const = get_constant_from_source(start);
```
gkrivor, line 30: Is it meaningful line?

src/frontends/tensorflow/src/op/concat.cpp
rkazants: It is changed because `input_size` is used below in `for` loop to align with type of `input_idx`.

tests/layer_tests/tensorflow_tests/test_tf_BroadcastArgs.py
```diff
@@ -0,0 +1,63 @@
+# Copyright (C) 2018-2023 Intel Corporation
```
gkrivor, line 1: Not sure it is correct copyright, I may expect (c) 2023 because it is a new file

tests/layer_tests/tensorflow_tests/test_tf_Slice.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (C) 2018-2023 Intel Corporation
```
gkrivor, line 1: Same concern about copyright year

src/frontends/tensorflow/src/utils.hpp
rkazants: I agree with you. Adjustment of the logging will be separate effort because all type of logs need to be adjusted in this case.

tests/layer_tests/tensorflow_tests/test_tf_BroadcastArgs.py
```diff
@@ -0,0 +1,63 @@
...
+            tf.raw_ops.BroadcastArgs(s0=s0, s1=s1)
+
+            tf.compat.v1.global_variables_initializer()
+            tf_net = sess.graph_def
+
+        ref_net = None
+        return tf_net, ref_net
+
+    test_data_basic = [
+        dict(s0_shape=[6], s1_shape=[6], input_type=tf.int32),
```
gkrivor, line 51: I may suggest to use  @pytest.mark.parametrize("input_type", [tf.int32, tf.int64])

tests/layer_tests/tensorflow_tests/test_tf_Slice.py
```diff
@@ -0,0 +1,37 @@
...
+
+            tf.compat.v1.global_variables_initializer()
+            tf_net = sess.graph_def
+
+        ref_net = None
+        return tf_net, ref_net
+
+    test_data_basic = [
+        dict(input_shape=[6], input_type=tf.float32, begin_value=[2], size_value=[2]),
+        dict(input_shape=[2, 5, 3], input_type=tf.int32, begin_value=[0, 1, 0], size_value=[-1, 1, -1]),
```
gkrivor, line 26: not sure I understand why you introduce verification of int32, but don't introduce other possible types

src/frontends/tensorflow/src/op/concat.cpp
itikhono: I see `node.get_input` method accepts `int` type.
ok

src/frontends/tensorflow/src/utils.hpp
itikhono: Do we have a ticket or could you create a new one?

tests/layer_tests/tensorflow_tests/test_tf_Slice.py
itikhono: agree, maybe we can place additional types to nightly tests, not to pre-commits

tests/layer_tests/tensorflow_tests/test_tf_BroadcastArgs.py
rkazants: Absolutely agree. Better to concentrate on technical things:)

tests/layer_tests/tensorflow_tests/test_tf_Slice.py
rkazants: this is pre-commit. For the full coverage I think we have separate activity. Now the main idea is to cover different shapes and types by mixing not overloading the pre-commit.
Nightly tests will be in separate tickets. The current tickets are to cover the common logic for the pre-commit.

src/frontends/tensorflow/src/helper_transforms/embedding_segments_feature_fusing.cpp
rkazants: Some transformation in the common part removes input with empty tensors to `Concat` so it is remained with no inputs.
The problem is in the transformation.

tests/layer_tests/tensorflow_tests/test_tf_BroadcastArgs.py
rkazants: This is the pre-commit scope, for nightly tests it make sense. The main idea to test a mix: different shapes and types to prevent different common bugs in plugin and FE.

src/frontends/tensorflow/src/op/concat.cpp
rkazants: Fixed.

src/frontends/tensorflow/src/frontend.cpp
rkazants: Done

src/frontends/tensorflow/src/helper_transforms/embedding_segments_feature_fusing.cpp
rkazants: Agreed to create a ticket for the common part and fix it there. Once the fix is done, this workaround will be removed.

rkazants: Done

gkrivor (approved):

itikhono (approved):

PR #15000, opened 2023-01-09
[GPU] GPU plugin supports ImportNetwork()

e-ddykim:
### Details:
 - `ImportNetwork()` is supported by GPU plugin from 2022.3 release.
 - So, the test case `IEClassBasicTestP.ImportNetworkThrows` is not necessary anymore.

### DIFF
--- a/src/tests/functional/plugin/shared/include/behavior/plugin/core_integration.hpp
+++ b/src/tests/functional/plugin/shared/include/behavior/plugin/core_integration.hpp
@@ -356,23 +356,6 @@ TEST_P(IEClassSpecificDeviceTestSetConfig, SetConfigSpecificDeviceNoThrow) {
 // ImportNetwork
 //
 
-TEST_P(IEClassBasicTestP, ImportNetworkThrows) {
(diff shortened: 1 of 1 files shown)


src/tests/functional/plugin/shared/include/behavior/plugin/core_integration.hpp
```diff
@@ -368,7 +368,7 @@ TEST_P(IEClassBasicTestP, ImportNetworkThrows) {
             file << "content";
         }
 
-        EXPECT_THROW(ie.ImportNetwork(modelName, target_device), InferenceEngine::NotImplemented);
+        EXPECT_THROW(ie.ImportNetwork(modelName, target_device), InferenceEngine::GeneralError);
```
vladimir-paramuzov, line 371: I think this tests is intended to check correct behavior for plugins which don't support Import/Export API. So if we support it now, then maybe whole test can be removed.

e-ddykim: I removed it as you reviewed. Thank you.

yeonbok (approved):

PR #15002, opened 2023-01-09
Add bf16 blob check in dump_checker.py

luweizhou2016:
### Details:
 - *Enhance bf16 blob check in dump_checker.py*

### Tickets:
 - *97918*


usstq (approved):
LGTM!

PR #14890, opened 2023-01-02
Review topk for shape inference aspects

praasz:
### Details:
- label and dimension propagation
- partial value and label propagation
- preserve partial value and labels
- add evaluate upper, lower and label
- introduce shape_infer which return inferred shapes as return.
- add new `get_input_const_data_as_shape` which allow validate values by element type and pre-process it.

### Tickets:
 - 97259
 - 97279
 - 97328
 - 97377
 - 97426
 - 97480
 - 97529
 - 97578

PR #15004, opened 2023-01-09
Refactor shape inference factory

praasz:
### Details:
- Add new shape inference factory which base on map.
- Add helpers to create inference factory map entries.
- Create map for `IShapeInferCommon` instead of if else switch
- Create new map for `IStaticShapeInfer`
- Update Tile shape inference to use new interface.
- Update validation utility `default_label_evaluator` to work and call op's evaluate base on `ov::Tensor`
- Improve `ov::cmp::lt` for mixed type comparison floating point and unsigned integers.

### Tickets:
 - 100374


src/plugins/intel_gpu/src/graph/tile.cpp
```diff
@@ -45,28 +45,26 @@ std::vector<layout> tile_inst::calc_output_layouts(tile_node const& /*node*/, co
...
     if (constant_mem.count(1)) {
         auto repeats_mem = constant_mem.at(1);
         cldnn::mem_lock<uint8_t, mem_lock_type::read> repeats_lock(repeats_mem, impl_param.prog->get_stream());
-        std::map<size_t, ngraph::HostTensorPtr> const_data = {
-            {1, make_host_tensor(repeats_mem->get_layout(), repeats_lock.data())}
-        };
-        ov::op::v0::shape_infer(&op, input_shapes, output_shapes, const_data);
+        const auto& layout = repeats_mem->get_layout();
+        repeats_tensor =
+            ov::Tensor(data_type_to_element_type(layout.data_type), layout.get_shape(), repeats_lock.data());
```
vladimir-paramuzov, line 60: It's illegal change as when `repeats_lock` is deleted, `repeats_lock.data()` ptr used for tensor creation may become invalid.

praasz:
The modified/added files `compare.hpp`, `shape_infer_transformations', 'utils.hpp' are taken from PR #14890 and got following changes:
* compare.hpp - improve ov::cmp::lt functions to compare float and unsigned.
* utils.hpp - modified functions `get_input_const_data_as`, `get_input_const_data_as_shape` to handle host tensor pointer and 
  ov::tensor reference.

src/plugins/intel_gpu/src/graph/tile.cpp
praasz: Corrected.

src/core/dev_api/dimension_tracker.hpp
```diff
@@ -9,12 +9,28 @@
...
 
 namespace ov {
 
 /// \brief Special label value indicate no label set.
 constexpr size_t no_label = 0;
 
+namespace element {
+/// \brief Get element::Type form label type.
+/// \return element::Type of same size in bytes as label_type (size_t)
+inline Type from_label_type() {
```
tomdol, line 22: Could you try making this function constexpr? Its logic is based entirely on the information available during the compile time so perhaps you can mark it explicitly constexpr and the compiler might inline it even more.

src/core/shape_inference/include/utils.hpp
```diff
@@ -44,37 +48,202 @@ void eltwise_shape_infer(const OpType* op, const std::vector<T>& input_shapes, s
...
+ * \param size  Data size as number of elements.
+ * \param func  Unary operation function object.
+ *
+ * \throws ov::AssertionFailure for not supported element type.
+ * \return Object of TResult with data from input pointer and transformed by unary operation.
+ */
+template <class T, class TResult = std::vector<T>, class UnaryOperation>
+TResult get_raw_data_as(const element::Type_t et, const void* const ptr, const size_t size, UnaryOperation&& func) {
+    TResult out;
+    auto out_it = std::inserter(out, out.end());
```
tomdol, line 69:
```suggestion
    auto out_it = std::back_inserter(out);
```

tomdol (approved):
Approved with optional suggestions

src/core/dev_api/dimension_tracker.hpp
praasz: To make it `constexpr`, I need to make it as template set for each `sizeof(size_t)` as `constexpr` cannot have multiple returns.
So probably then more proper solution would be introduce template specializations for element::from<size_t> for mac OS, as for others used on CI the `size_t` is alias to some unsigned integer which is supported by element::from<>.

I assume this as temporary solution and in PR #15010 there alias for `label_t` as uint32_t (or any other supported by tensor type) then this function is not required at all and can be removed.

Depends which PR will be merged first this or #15010, I will update 2nd one accordingly.

mitruska (commented):
@praasz Great effort, could you elaborate more abut the advantages of those changes and further plans?

src/core/shape_inference/include/utils.hpp
praasz: std::inserter selected as more containers support it.

src/core/shape_inference/include/shape_infer_transformations.hpp
```diff
@@ -0,0 +1,58 @@
...
+//
+
+#pragma once
+
+#include "compare.hpp"
+#include "openvino/core/except.hpp"
+
+namespace ov {
+namespace sh_infer {
+namespace tr {
```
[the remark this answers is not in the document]
praasz: For transformations, but as file name changed also change to be `ov::util`

src/core/shape_inference/include/utils.hpp
```diff
@@ -84,32 +253,82 @@ std::unique_ptr<TRes> get_input_const_data_as(const ov::Node* op,
...
+          class UnaryOperation = sh_infer::tr::InTypeRange<TDimValue>>
+std::unique_ptr<TShape> get_input_const_data_as_shape(const ov::Node* op,
+                                                      size_t idx,
+                                                      const std::map<size_t, TTensorPtr>& constant_data = {},
+                                                      UnaryOperation&& func = sh_infer::tr::InTypeRange<TDimValue>()) {
+    std::unique_ptr<TShape> shape_ptr;
+
+    if (auto d =
+            get_input_const_data_as<TShape, TDimValue>(op, idx, constant_data, std::forward<UnaryOperation>(func))) {
+        shape_ptr.reset(new TShape(std::move(*d)));
```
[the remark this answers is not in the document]
praasz: Changed.

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
```diff
@@ -386,199 +375,310 @@ class entryConvBackprop : public entryBase {
...
+#undef _OV_OP_NON_TEMPLATE_SHAPE_INFER_REG
+#undef _OV_OP_SHAPE_INFER_REG
+#undef _OV_OP_SHAPE_INFER_VA_REG
+
+template <>
+std::shared_ptr<IShapeInferCommon> make_shape_inference<IShapeInferCommon>(std::shared_ptr<Node> op) {
+    if (auto shape_infer = ShapeInferFactory::make<std::shared_ptr<IShapeInferCommon>>(op->get_type_info(), op)) {
+        return shape_infer;
+    } else if (auto shape_infer = make_shape_inference<IStaticShapeInfer>(op)) {
+        return shape_infer;
```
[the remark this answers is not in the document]
praasz: Yes,
But there are some blocking points to that in more expanded way just make an example on Tile operator.
This new interface operate `ov::Tensor` map instead host tensor so operator's `shape_infer` has to be written for `ov::Tensor` 

If `shape_infer` is done on `ov::Tensor` then evaluate in op can be reworked to use ov::Tensor version and host tensor version could be removed. But is not possible to remove evaluate version for host tensor as template plugin use it (work on this is ongoing) and some `validation_utils` can call it directly instead ov Tensor version. On tile example the evaluates are duplicated till this template plugin an `utils` will not be changed.

So from shape inference perspective we should rework the `utils` for bounds evaluation to work on new tensors and then is possible to start implement evaluates for this version tensors.

The IStaticShapeInfer can be used to create the previous version of shape inference object but not vice versa.

src/core/shape_inference/include/utils.hpp
```diff
@@ -44,37 +48,202 @@ void eltwise_shape_infer(const OpType* op, const std::vector<T>& input_shapes, s
...
+    } break;
+    case element::Type_t::f32: {
+        using dtype = fundamental_type_for<element::Type_t::f32>;
+        std::transform(static_cast<const dtype*>(ptr),
+                       static_cast<const dtype*>(ptr) + size,
+                       out_it,
+                       std::forward<UnaryOperation>(func));
+    } break;
+    default:
+        OPENVINO_ASSERT(false, "Not supported element type ", et);
```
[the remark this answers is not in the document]
praasz: The not supported type is provided in exception, added more details in exception description for easier identifications.

The test are done indirect way as this function is used by all `get_data_as` so it is use in all test cases for shape inference and there is no regression.

I think this function will be re-worked as this concept with explicit switch case for types not exactly is what I want.

src/core/shape_inference/include/compare.hpp
```diff
@@ -73,5 +77,104 @@ class Equal {
         return _exp_value == value;
     }
 };
+
+/**
+ * \brief Compare two integers (a < b) in safe way against lossy integer conversion.
```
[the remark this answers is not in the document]
praasz: Added tests

mitruska (approved):

PR #15007, opened 2023-01-09
[GPU] Serialization of read_value and assign

e-ddykim:
### Details:
 - Serialization of `read_value` and `assign` primitives

### Tickets:
 - 98690

### DIFF
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -1086,6 +1086,9 @@ void primitive_inst::save(cldnn::BinaryOutputBuffer& ob) const {
     ob << can_be_optimized();
     ob << can_share_buffer();
     ob << is_constant();
+    auto users = get_node().get_users();
+    bool is_output_event = is_any_user_cpu(users) || get_node().is_output();
+    ob << is_output_event;
 
     if (type() == cldnn::data::type_id()) {
         return;
@@ -1201,6 +1204,7 @@ void primitive_inst::load(cldnn::BinaryInputBuffer& ib) {
     ib >> _can_be_optimized;
     ib >> _can_share_buffer;
     ib >> _is_constant;
+    ib >> _is_output_event;
 
     if (type() == cldnn::data::type_id()) {
         return;
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -1248,7 +1248,8 @@ void primitive_inst::load(cldnn::BinaryInputBuffer& ib) {
             if ((!can_share_buffer()) || can_be_optimized() || is_output()) {
                 _outputs[0] = get_network().get_engine().allocate_memory(output_layout, _allocation_type);
             } else {
-                _outputs[0] = get_network().get_memory_pool().get_memory(output_layout, id(), get_network_id(), _node_mem_deps, _allocation_type, can_reuse_memory);
+                _outputs[0] = get_network().get_memory_pool().get_memory(output_layout, id(), get_network_id(),
+                                                                         _node_mem_deps, _allocation_type, can_reuse_memory);
             }
         }
     }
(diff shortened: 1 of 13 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1069,14 +1069,22 @@ void primitive_inst::save(cldnn::BinaryOutputBuffer& ob) const {
...
 
     if (type() == cldnn::data::type_id()) {
         return;
     }
 
-    ob << _outputs[0]->get_layout();
-    const auto _allocation_type = _outputs[0]->get_allocation_type();
-    ob << make_data(&_allocation_type, sizeof(_allocation_type));
+    if (_outputs[0] == nullptr) {
+        ob << true;
```
yeonbok, line 1081: How is this different from mem_allocated ?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1206,19 +1218,22 @@ void primitive_inst::load(cldnn::BinaryInputBuffer& ib) {
...
-        if (dep_id.compare("NOT_FOUND") != 0) {
-            _outputs[0] = get_network().get_engine().reinterpret_buffer(get_network().get_primitive(dep_id)->output_memory(), output_layout);
-        } else if (type() == cldnn::mutable_data::type_id()) {
-            _outputs[0] = get_network().get_engine().allocate_memory(output_layout, _allocation_type);
-        }
-    } else {
-        if ((!can_share_buffer()) || can_be_optimized() || is_output()) {
-            _outputs[0] = get_network().get_engine().allocate_memory(output_layout, _allocation_type);
+    if (!output_is_null) {
+        if (!_mem_allocated) {
```
yeonbok, line 1222: Here too. Can't mem_allocated used as !output_is_null ?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1206,19 +1218,22 @@ void primitive_inst::load(cldnn::BinaryInputBuffer& ib) {
...
+            std::string dep_id;
+            ib >> dep_id;
+            if (dep_id.compare("NOT_FOUND") != 0) {
+                _outputs[0] = get_network().get_engine().reinterpret_buffer(get_network().get_primitive(dep_id)->output_memory(), output_layout);
+            } else if (type() == cldnn::mutable_data::type_id()) {
+                _outputs[0] = get_network().get_engine().allocate_memory(output_layout, _allocation_type);
+            }
```
yeonbok, line 1229: Now I'd change the question. When is the mem_allocated is true but output is nullptr? Should't we set mem_allocated to be false if output is nullptr?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
e-ddykim: `mem_allocated` does not tell if an operation has an output memory.
For example, in the case of implicit concat, `mem_allocated` is false and output is not null.
So, we can query output layout for the concat operation.
But in the case of the read_value operation, `mem_allocated` is false and output is also null.
Output for the read_value operation is allocated by any external logic.

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
e-ddykim: As I replied above, in the case of implicit concat, `mem_allocated` is false and output is not null.

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
yeonbok: Thanks for the explanation. Then for that purpose, can we just infer as 
( !mem_allocated && can_be_optimized) => has_output mem
( !mem_allocated && !can_be_optimized) => no output mem ?

I'd like to have # of similar variables as less as possible, becasue it would be very confusing in the future :D

e-ddykim: In the case of `mutable_data`, mem_allocated is false, can_be_optimized is false, and output is not null.
I agree with your opinion, but I'm not sure if it's possible to replace it with a combination of multiple variables.

e-ddykim: I added comments for the difference between mem_allocated and is_output_null to avoid future confusion. Thank you.

yeonbok (approved):

PR #15014, opened 2023-01-09
Use unique file names in tests

olpipi:
### Details:
 - Add api to common test utils to create unique names
 - Use this api in tests

### Tickets:
 - CVS-96479

### DIFF
--- a/src/tests/functional/plugin/conformance/subgraphs_dumper/src/ops_cache.cpp
+++ b/src/tests/functional/plugin/conformance/subgraphs_dumper/src/ops_cache.cpp
@@ -196,10 +196,10 @@ OPCache::serialize_function(const std::pair<std::shared_ptr<ov::Node>, LayerTest
         }
         std::replace(op_name.begin(), op_name.end(), '/', '_');
         std::replace(op_name.begin(), op_name.end(), '\\', '_');
-        // TODO: Possible names collision
-        auto xml_path = current_op_folder + CommonTestUtils::FileSeparator + op_name + ".xml";
-        auto bin_path = current_op_folder + CommonTestUtils::FileSeparator + op_name + ".bin";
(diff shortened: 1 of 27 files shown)


ilyachur (commented):
Can we grep and fix all test cases which works with non-unique file?

p-durandin (approved):
Please fix code style

ilyachur (commented):
Please double check.
For example core tests...
I found 
 - https://github.com/openvinotoolkit/openvino/blob/master/src/core/tests/pass/serialization/const_compression.cpp#L16
 - https://github.com/openvinotoolkit/openvino/blob/master/src/core/tests/pass/serialization/tensor_iterator.cpp#L17
 
for example.

ilyachur (approved):

src/tests/functional/plugin/conformance/subgraphs_dumper/src/ops_cache.cpp
```diff
@@ -196,10 +196,10 @@ OPCache::serialize_function(const std::pair<std::shared_ptr<ov::Node>, LayerTest
         }
         std::replace(op_name.begin(), op_name.end(), '/', '_');
         std::replace(op_name.begin(), op_name.end(), '\\', '_');
-        // TODO: Possible names collision
-        auto xml_path = current_op_folder + CommonTestUtils::FileSeparator + op_name + ".xml";
-        auto bin_path = current_op_folder + CommonTestUtils::FileSeparator + op_name + ".bin";
-        auto meta_info = current_op_folder + CommonTestUtils::FileSeparator + op_name + ".meta";
+        auto filePrefix = CommonTestUtils::generateTestFilePrefix();
```
iefode, line 199: @olpipi Why do you made such changes??? It causes of the crashes in `subgraphdumper`.
Next time ask me than make such changes
FYI @ilyachur

PR #15016, opened 2023-01-09
[GPU] Added shape agnostic ref kernels for Select and Activation

Lyamin-Roman:


src/plugins/intel_gpu/src/kernel_selector/cl_kernels/select_gpu_ref.cl
```diff
@@ -24,14 +29,19 @@ const uint d1  = (uint) get_global_id(0);
...
-const uint d4  = d34 / OUTPUT_SIZES[2];
-
-uint output_offset = OUTPUT_OFFSET +
-                     d1*OUTPUT_PITCHES[0] +
-                     d2*OUTPUT_PITCHES[1] +
-                     d3*OUTPUT_PITCHES[2] +
-                     d4*OUTPUT_PITCHES[3];
+#ifdef IS_DYNAMIC
+    const uint d3 = d34 % OUTPUT_FEATURE_NUM;
+    const uint d4 = d34 / OUTPUT_FEATURE_NUM;
```
yeonbok, line 34: Is it feature_num? Curious because the shape aware kernel uses output_sizes[2] :)

Lyamin-Roman: Dims in *_sizes are stored in reverse order. So output_sizes[2] in xyfb order is feature_num

src/plugins/intel_gpu/src/plugin/ops/convert.cpp
```diff
@@ -23,9 +23,7 @@ static void CreateConvertLikeOp(Program& p, const std::shared_ptr<ngraph::op::v1
     auto reorderPrim = cldnn::reorder(layerName,
                                       inputs[0],
                                       cldnn::format::any,
-                                      outDataType,
-                                      std::vector<float>(),
-                                      cldnn::reorder_mean_mode::subtract);
+                                      outDataType);
```
yeonbok, line 26: Could you let me know which test required this change?

Lyamin-Roman: None, it's just the default parameters remaining after removing one of the arguments. This change does not affect the behavior

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/activation_ref.cl
```diff
@@ -32,7 +36,7 @@ KERNEL(activation)(
     const unsigned x = get_global_id(0);
     const uint y = (uint)get_global_id(1) % OUTPUT_SIZE_Y;
     const uint z = (uint)get_global_id(1) / OUTPUT_SIZE_Y;
-    #if OUTPUT_BATCH_NUM == 1
+    #if OUTPUT_BATCH_NUM_CONST == 1
```
sshlyapn, line 39: You can't use runtime value in preprocessing stage
Also, it will make preprocessor always choose only `else` branch (even in static scenario), since `OUTPUT_BATCH_NUM_CONST` doesn't exist from preprocessor's point of you
So, the proper solution will be keep it as it was. And in that case static path may be optimized, but in dynamic case we need to calculate batch index anyway (`else` branch)

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -274,7 +275,7 @@ void primitive_inst::update_impl() {
             return ps.to_shape();
         }
         if (ps.size() < 4) {
-            if (_node->is_type<gemm>())
+            if (_node->is_type<gemm>() || _node->is_type<select>())
```
yeonbok, line 278: Please check the latest master. extend_to_6d is moved to primivie_base so that it can be overriden.

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/activation_ref.cl
```diff
@@ -72,11 +73,11 @@ KERNEL(activation)(
 
 // GWS.feature and GWS.batch is aligned to 16. Otherwise, there are some idling WIs.
 #if (defined(OUTPUT_LAYOUT_B_FS_YX_FSV16) || defined(OUTPUT_LAYOUT_B_FS_YX_FSV32)) \
-    && OUTPUT_FEATURE_NUM % 16 != 0
+    && OUTPUT_FEATURE_NUM_CONST % 16 != 0
     if (feature >= OUTPUT_FEATURE_NUM)
         return;
 #elif (defined(OUTPUT_LAYOUT_BS_FS_YX_BSV32_FSV16) || defined(OUTPUT_LAYOUT_BS_FS_YX_BSV32_FSV32)) \
-    && (OUTPUT_FEATURE_NUM % 16 != 0 || OUTPUT_BATCH_NUM % 16 != 0)
+    && (OUTPUT_FEATURE_NUM_CONST % 16 != 0 || OUTPUT_BATCH_NUM_CONST % 16 != 0 || IS_DYNAMIC)
```
sshlyapn, line 80: Just out of curiosity: why did you add IS_DYNAMIC for double-blocked formats, but ignore B_FS_YX_FSV16 and B_FS_YX_FSV32 formats?

src/plugins/intel_gpu/src/graph/select.cpp
```diff
@@ -55,6 +55,28 @@ std::vector<layout> select_inst::calc_output_layouts(const select_node& /*node*/
...
+
+    if (ps.size() < 4) {
+        ps.insert(ps.begin(), 4 - ps.size(), ov::Dimension(1));
+    }
+
+    layout l(ps, data_types::i32, format::get_default_format(ps.size()));
+    return l.transform(format::bfwzyx).to_shape();
+}
+
+std::vector<size_t> select_inst::extend_output_shape_to_6d(kernel_impl_params const& orig_impl_param, int32_t output_idx) {
```
sshlyapn, line 69: `extend_output_shape_to_6d()` and `extend_input_shape_to_6d()` functions look mostly identical, so maybe it makes sense to extract common part to `extend_shape_to_6d()` function and reuse the code

sshlyapn (approved):

PR #15019, opened 2023-01-10
fix ip GemmJit implement issue on avx512 platform

liubo-intel:
### Details:
 - *keep to use brgemm matmul instead of gemm:jit inner_product for 4D inputs cases(sp dimensions are not all 1) on avx512_core platform*

### Tickets:
 - *95796*

### DIFF
--- a/src/plugins/intel_cpu/src/ngraph_transformations/convert_matmul_to_fc.cpp
+++ b/src/plugins/intel_cpu/src/ngraph_transformations/convert_matmul_to_fc.cpp
@@ -10,7 +10,6 @@
 #include <transformations/utils/utils.hpp>
 
 #include "itt.hpp"
-#include <cpu/x64/cpu_isa_traits.hpp>
 
 ov::intel_cpu::ConvertMatMulToFC::ConvertMatMulToFC() {
     MATCHER_SCOPE(ConvertMatMulToFC);
@@ -42,19 +41,6 @@ ov::intel_cpu::ConvertMatMulToFC::ConvertMatMulToFC() {
         if (rank_a == 1 || rank_b == 1 ||
             rank_a > 3 || rank_b > 3) {
             return false;
-        }
-
-        // keep to use brgemm matmul instead of gemm:jit inner_product for 4D inputs cases(sp dimensions are not all 1) on avx512_core platform
-        auto reshape = std::dynamic_pointer_cast<ngraph::opset1::Reshape>(matmul->get_input_node_shared_ptr(0));
-        if (reshape && reshape->input_value(0).get_partial_shape().is_static()) {
-            auto shape_in = reshape->input_value(0).get_shape();
-            auto shape_out = reshape->get_shape();
-            if (shape_in.size() == 4 && reshape->get_shape().size() == 2) {
-                if (shape_in[0] == shape_out[0] && ((shape_in[2] != 1) || (shape_in[3] != 1)) &&
-                    dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx512_core)) {
-                    return false;
-                }
-            }
         }
 
         // Check that if second inputs is Constant path and it's shape without ones dimensions has length <= 2
--- a/src/plugins/intel_cpu/src/ngraph_transformations/convert_matmul_to_fc.cpp
+++ b/src/plugins/intel_cpu/src/ngraph_transformations/convert_matmul_to_fc.cpp
@@ -10,6 +10,7 @@
 #include <transformations/utils/utils.hpp>
 
 #include "itt.hpp"
+#include <cpu/x64/cpu_isa_traits.hpp>
 
 ov::intel_cpu::ConvertMatMulToFC::ConvertMatMulToFC() {
     MATCHER_SCOPE(ConvertMatMulToFC);
@@ -41,6 +42,20 @@ ov::intel_cpu::ConvertMatMulToFC::ConvertMatMulToFC() {
         if (rank_a == 1 || rank_b == 1 ||
             rank_a > 3 || rank_b > 3) {
             return false;
+        }
+
+        // keep to use brgemm matmul instead of gemm:jit inner_product for 4D inputs cases(sp dimensions are not all 1) on avx512_core platform
+        auto reshape = std::dynamic_pointer_cast<ngraph::opset1::Reshape>(matmul->get_input_node_shared_ptr(0));
+        if (reshape && reshape->input_value(0).get_partial_shape().is_static()) {
+            auto shape_in = reshape->input_value(0).get_shape();
+            auto shape_out = reshape->get_shape();
+            if (shape_in.size() == 4 && reshape->get_shape().size() == 2) {
+                if (shape_in[0] == shape_out[0] &&
+                    std::accumulate(shape_in.begin() + 1, shape_in.end(), size_t {1}, std::multiplies<size_t>()) != shape_in[1] &&
+                    dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx512_core)) {
+                    return false;
+                }
+            }
         }
 
         // Check that if second inputs is Constant path and it's shape without ones dimensions has length <= 2
--- a/src/plugins/intel_cpu/src/ngraph_transformations/convert_matmul_to_fc.cpp
+++ b/src/plugins/intel_cpu/src/ngraph_transformations/convert_matmul_to_fc.cpp
@@ -50,8 +50,7 @@ ov::intel_cpu::ConvertMatMulToFC::ConvertMatMulToFC() {
             auto shape_in = reshape->input_value(0).get_shape();
             auto shape_out = reshape->get_shape();
             if (shape_in.size() == 4 && reshape->get_shape().size() == 2) {
-                if (shape_in[0] == shape_out[0] &&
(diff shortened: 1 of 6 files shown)


liubo-intel:
Hi, @usstq , @dmitry-gorokhov : could you please help have a look at this PR when you are free? thanks.

src/plugins/intel_cpu/src/ngraph_transformations/convert_matmul_to_fc.cpp
```diff
@@ -43,6 +44,19 @@ ov::intel_cpu::ConvertMatMulToFC::ConvertMatMulToFC() {
             return false;
         }
 
+        // keep to use brgemm matmul instead of gemm:jit inner_product for 4D inputs cases(sp dimensions are not all 1) on avx512_core platform
```
usstq, line 47: This comment is good, but not enough, imagine someone has no background knowledge of this PR saw this comment, can he know the exact reason why we are doing this? 

We can add more explanation on the fact that FullyConnected is based on oneDNN `Inner Product` primitive which shows performance worse than `Matrix Multiplication` primitive in 4D cases, and the Matmul CPU node is implemented exactly based on that primitive. so this piece of code is actually based on many other facts not easily discovered by reader, and it will become invalid when any of them becomes invalid, for example, when Matmul CPU node changed its implementation

So why not implement this WA inside FullyConnected CPU node ? just like what we do to for using brgconv1x1 to speed up some special shape?:
 - it's easier to understand because in FullyConnected we are closer to oneDNN primitive to explain the reason
 - it's not dependent on how Matmul is implemented which is more reasonable
 - it can support the case when target shape of reshape is not constant (comes from a shape-Of subgraph for example), because in `prepareParam()` stage we can get exact shape

dmitry-gorokhov: I am not fully agree with that view. 

We have MatMul operation in OpenVINO opset and corresponding node in CPU plugin opset. Besides that we have internal FullyConnected node was introfuced to easily map on oneDNN InnerProduct primitive due to performance reeasons. In fact this node is very backend specific and required only for CPU plugin + OneDNN configuration. So in other words I expect MatMul to be converted to FullyConnected only in case it it required by performance reasons and only for use-cases where FullyConnected is really supported by the backend. 

Performing MatMul to FC conversion and then implementening FC executor with Matmul primitive inside doesn't make any sence for me. Two executors (Conv1x1 and IP based) underneath FC node is reasonable solution due to the fact we might use one or another each iteration based on input shape. Here is the different story: this is enough to make sure that resulting FC has different rank on input and output - which is static information (not really neccessary to check that dimensions equal 1)

I fully agree that having target specific conditions inside transformation itself is not desired solution. Usually we have to use callback to specify such customization. I can propose the following: we can move transformation_callback call to the end of the function and make it based on resulting op (FC) instead of startinf op (MatMul) See [example](https://github.com/openvinotoolkit/openvino/blob/b8a7b3bb4301be848b1e5b3ab31a068cfb4c0bc6/src/common/transformations/src/transformations/common_optimizations/add_fake_quantize_fusion.cpp#L117).  Then we might implement target specific conditions inside dedicated FullyConnected::isSupported() method and provide as callback inside the transformation. See [example](https://github.com/openvinotoolkit/openvino/blob/b8a7b3bb4301be848b1e5b3ab31a068cfb4c0bc6/src/plugins/intel_cpu/src/transformation_pipeline.cpp#L416).  

With that approach we encapsulate target specific conditions inside target specific node, keeping transformation itself generic. Minimal changes required to fix original issue.
@usstq  What do you think?

usstq: @dmitry-gorokhov , after discussion with @liubo-intel in depth, we found the root cause is `brgemm_inner_product_fwd_t` implementation isn't following the [design of inner_product](https://oneapi-src.github.io/oneDNN/dev_guide_inner_product.html) which claimed to support inputs with spatial dimensions, we can see `gemm_inner_product_fwd_t` followed the design well, thus I think a better fix should be done inside oneDNN (a PR to fork is proposed) and maybe we can also consider upstream this fix:

oneDNN fix: https://github.com/openvinotoolkit/oneDNN/pull/175

What do you think?

dmitry-gorokhov: @usstq Sounds good. Lets merge after oneDNN 3.0 migrations. 
Please also submit ticket to oneDNN as we want this functionality to finally appear in oneDNN mainstream repo.

liubo-intel: Hi, @dmitry-gorokhov and @usstq : thanks for your comments and suggestions. if we all agree to use oneDNN fixed way for this issue, I will close this ov PR later.

usstq: @dmitry-gorokhov Sorry, @liubo-intel found https://github.com/openvinotoolkit/oneDNN/pull/175 actually doesn't work for 4D input in practice, the reason is, for example:

name | external shape | ext layout | flattened shape (by brgemm) | internal layout (by brgemm) |
:------ | :------|:------|:------|:------|
src| 1x512x7x7 | abcd | 1x25088 | ab
weight|4096x512x7x7 | **???** | 4096x25088 | AB16b64a 

we can see that there is no way to represent the `AB16b64a` layout of flattened 2D weight required by brgemm in orginal/external 4D shape (shown as ***???*** in the table, it not simply ABcd16b64a).  so I'm not sure the internal flatten design is good idea for inner product, considering the blocking requirement of weight.

For WA, to successfully use `brgemm_inner_product_fwd_t` when input has spatial dimension, we must flatten(reshape) it into 2D inside OV and only passing 2D descriptor to oneDNN.

the pattern we are trying to fix is like this
```bash
###### original onnx/IR : a typical way of connecting output of convolution to fullyconnect layer
src4d =pooling(input) <====================  with shape 1x512x7x7
src2d=reshape(src4d, 1x25088)
weight2d=const_of_shape(4096x25088)
dst2d = matmul(src2d, weight2d) <============= 2D matmul

###### after MatcherPass  ConvertMatMulToFC
src2d=reshape(src4d, 1x25088)
weight2d=const_of_shape(4096x25088)
dst2d = fullyconnect(src2d, weight2d) <========== FC with 2D inputs

###### after MatcherPass ReshapeFullyConnectedFusion
weight2d=const_of_shape(4096x25088)
weight4d = reshape(weight2d, 4096x512x7x7) <===== will be const-folded
dst2d = fullyconnect(src4d, weight4d) <========== 4D inputs to FC !!!!
```
[ConvertMatMulToFC](https://github.com/openvinotoolkit/openvino/blob/fb3d785c0662c4c0c5864162d96c0ee09b220f45/src/plugins/intel_cpu/src/ngraph_transformations/convert_matmul_to_fc.cpp#L42) avoids convert matmul with inputs rank more than 3 into FC, but [ReshapeFullyConnectedFusion](https://github.com/openvinotoolkit/openvino/blob/cabe75da1c60397e6fb101929ab47983ed66244d/src/plugins/intel_cpu/src/ngraph_transformations/reshape_fc_fusion.cpp#L34) is creating 4D FC by purposely remove the preceding reshape node, if inner product primitive can indeed support internal spatial dimension flattening, everything works fine, but as we saw,  the major implementation `brgemm_inner_product_fwd_t` has difficulty supporting that, which causes the FC to use fallback implementation `gemm`.

What @liubo-intel does in this PR is trying to fix this in `ConvertMatMulToFC` by detecting the existence of 4D=>2D reshape before Matmul and leaves it as Matmul, which would actually make the some logic in second transformation `ReshapeFullyConnectedFusion` useless since there would be no more (4D=>2D reshape + FC) pattern. which seems to be improper to me.

checking MatMul's definition we can find that it actually doesn't support spatial dimension:

> Two right-most axes in each tensor are interpreted as matrix rows and columns dimensions while all left-most axes (if present) are interpreted as multi-dimensional batch: [BATCH_DIM_1, BATCH_DIM_2,..., BATCH_DIM_K, ROW_INDEX_DIM, COL_INDEX_DIM]

so if input tensor has spatial dimension, there must be some subgraph/reshape before MatMul node to flatten it into the last right-most axis, and that should be kept rather than removed - to flatten spatial dimension outside of  the oneDNN inner product.

thus we can just remove the 4D=>2D case in  `ReshapeFullyConnectedFusion`, so we can enjoy the performance benefit of `brgemm_inner_product_fwd_t` in such case, we are not sure if that's feasible because the 4D=>2D logic must be there for a reason, we need your advice, Thanks!

liubo-intel: Hi, @dmitry-gorokhov there are some updates about performance data, FYI:
For the performance issue case mentioned in ticket 95796, if we use 'GemmJit -> Matmal' way, the model performance improved from ~470fps to ~480fps, if use 'GemmJit -> ip' way(remove the 4D=>2D case in ReshapeFullyConnectedFusion) that @usstq mentioned above, the performance improved from ~470fps to ~500fps.

akladiev:

wenjiew:
@liubo-intel Can we re-look at this PR to see the need to merge in 2023.1 release? Thanks!

liubo-intel:
AFAIK, this kind of 'fullyconnected to matmul' optimization way will be considered  together in the following CPU fullyconnected/matmul  node optimization. 
will close this PR currently.

PR #15035, opened 2023-01-10
[GPU] Fix cl source dump file name not to be overwritten

yeonbok:
### Details:
 - In dynamic shape, kernels are newly created and compiled for new shapes. Previously, dumped cl source codes for the same program id was overwritten by such new kernels, because the file name is only identified by prog_id, batch_id. 
 - Fixed dump file name to include hash value so that it will not be overwritten.

### DIFF
--- a/src/plugins/intel_gpu/src/runtime/kernels_cache.cpp
+++ b/src/plugins/intel_gpu/src/runtime/kernels_cache.cpp
@@ -205,7 +205,7 @@ void kernels_cache::build_batch(const engine& build_engine, const batch_program&
             current_dump_file_name += '/';
 
         current_dump_file_name += "clDNN_program_" + std::to_string(_prog_id) + "_bucket_" + std::to_string(batch.bucket_id)
-                               + "_part_" + std::to_string(batch.batch_id) + ".cl";
+                               + "_part_" + std::to_string(batch.batch_id) + "_" + std::to_string(batch.hash_value) + ".cl";
     }
 
     std::ofstream dump_file;


src/plugins/intel_gpu/src/runtime/kernels_cache.cpp
```diff
@@ -205,7 +205,7 @@ void kernels_cache::build_batch(const engine& build_engine, const batch_program&
             current_dump_file_name += '/';
 
         current_dump_file_name += "clDNN_program_" + std::to_string(_prog_id) + "_bucket_" + std::to_string(batch.bucket_id)
-                               + "_part_" + std::to_string(batch.batch_id) + ".cl";
+                               + "_part_" + std::to_string(batch.batch_id) + "_" + std::to_string(batch.hash_value) + ".cl";
```
isanghao, line 208: What about having something like a sequence number so that we can know when it is created? I think hash will be quite difficult to track.

yeonbok: At first I thought about it, but there are multiple (many) kernels built in the same sequence, so they will be still overwritten. 
Actually in my mind, I am thinking of adding the new impl hash (of the first kenel in the batch, if it is compiled in batch), once the https://github.com/openvinotoolkit/openvino/pull/14966 is merged. Doing so, we can retrieve the kernel with key used in the impl cache. Of course still we'll be able to add sequence number additionally. Also, the sequence num is only the sequence num when it was "created", not the sequence num being used.. So I doubt it will be very helpful... 

All in all, we can consider further improvment with those various combination. However this is a minimal change for the current overwriting problem.

kelvinchoi-intel (approved):

PR #15036, opened 2023-01-10
[GPU] Add black list to avoid fusing crop before experimental_detectron_roi_feature_extractor

sungeunk:
### Tickets:
 - 96566

### DIFF
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_buffer_fusing.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_buffer_fusing.cpp
@@ -14,6 +14,7 @@
 #include "loop_inst.h"
 #include "non_max_suppression_inst.h"
 #include "experimental_detectron_roi_feature_extractor_inst.hpp"
+#include "border_inst.h"
 
 #include "pass_manager.h"
 #include "program_helpers.h"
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_buffer_fusing.cpp
```diff
@@ -365,6 +366,8 @@ void prepare_buffer_fusing::run(program& p) {
                     if (can_reshape_be_optimized(reshape_node))
                         return;
                 }
+                if (user->is_type<experimental_detectron_roi_feature_extractor>())
+                    return;
```
isanghao, line 370: I think this is too restrictive. can we limit the scope of this black list by the specific input port of roi feature extractor?

src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_buffer_fusing.cpp
```diff
@@ -419,6 +423,18 @@ void prepare_buffer_fusing::run(program& p) {
...
+                    //  feature num of pad should be accumulated if dep has been optimized out.
+                    auto& dep = node.get_dependency(0);
+                    if (dep.is_type<crop>() && dep.can_be_optimized()) {
+                        auto dep_pad = dep.get_output_layout().data_padding;
+                        if (dep_pad.lower_size().batch[0] == 0 && dep_pad.upper_size().batch[0] == 0 &&
+                            dep_pad.lower_size().spatial[0] == 0 && dep_pad.upper_size().spatial[0] == 0 &&
+                            dep_pad.lower_size().spatial[1] == 0 && dep_pad.upper_size().spatial[1] == 0) {
+                            opt_lower_pad += dep_pad.lower_size().feature[0];
+                            opt_upper_pad += dep_pad.upper_size().feature[0];
+                        }
```
isanghao, line 435: what if it does not go into this `if` condition? Should we throw exception?
Could you also add a unit test for this case?

isanghao (changes requested):
Please see the comment

isanghao (approved):

PR #15037, opened 2023-01-11
[GPU] Let reorder_inputs pass to use not only output_layout but also input layout for dynamic shape

yeonbok:
### Details:
 - Previously, reorder_input pass checked only output layouts of each node, with an assumption that the input/output ranks are same at that pass. 
 - However in dynamic shape using ngraph shape infer, there are cases input / output ranks differ. In such cases, the reorder_input pass inserts reorder to the format of current node's output_layout in the input of the current node, which causes error.
 - Fixed the above behavior by applying set_preferred_input/output_layout.

### Tickets:
 - 99432


yeonbok:
Some functional tests are failing, which seems expecting throw exception but not any more, from the log
smoke_Multi_BehaviorTests/InferRequestIOBBlobSetLayoutTest.CanSetOutBlobWithDifferentLayouts/layout=SCALAR_target_device=MULTI

vladimir-paramuzov (approved):

PR #15039, opened 2023-01-11
Review eye class for shape inference aspects

praasz:
### Details:
 - Check label and interval dimension propagation
 - Add labels and partial values on rows and columns inputs.
 - Review and refactor template shape inference implementation and its usage in operator.

### Tickets:
 - 97229
 - 97309
 - 97407
 - 97456
 - 97510
 - 97559
 - 97608


src/plugins/intel_cpu/tests/unit/shape_inference_test/eye_shape_inference_test.cpp
```diff
@@ -0,0 +1,171 @@
...
+        {3, std::make_shared<HostTensor>(element::i32, Shape{batch.size()}, batch.data())}};
+
+    input_shapes = ShapeVector{{1}, {}, {1}, {3}};
+
+    OV_EXPECT_THROW(shape_inference(op.get(), input_shapes, output_shapes, const_data),
+                    NodeValidationFailure,
+                    HasSubstr("'num_columns' value must be a scalar or 1D tensor. Got:"));
+}
+
+TEST_F(EyeV9StaticShapeInferenceTest, assert_on_batch_shape_not_match_shape_in_const_data) {
```
mitruska, line 151: Additional test with default constructor and shape_infer could be added to ensure the inputs are not required to be set.

mitruska (approved):

src/plugins/intel_cpu/tests/unit/shape_inference_test/eye_shape_inference_test.cpp
praasz: Do you  mean test like:
https://github.com/openvinotoolkit/openvino/blob/c1200e2a6347fc10af37afcc48dbd660253f8d8a/src/core/tests/type_prop/eye.cpp#L360

I've put this kind test in type prop for default ctor also for some other ops.

mitruska: Exactly, okay.

jane-intel (approved):

PR #15043, opened 2023-01-11
[TF FE] Support While and StatelessWhile operations

rkazants:
**Details:** Implement support of While operation. Add layer tests for this operation

**Tickets:** 91480, 92853


gkrivor (approved):
Could you check copyrights in updated files?

rkazants:
> Could you check copyrights in updated files?

Done

src/frontends/tensorflow/src/op/while.cpp
```diff
@@ -0,0 +1,115 @@
...
+using namespace std;
+using namespace ov;
+using namespace ov::opset10;
+
+namespace ov {
+namespace frontend {
+namespace tensorflow {
+namespace op {
+OutputVector translate_while_op(const NodeContext& node) {
+    default_op_checks(node, 0, {"While", "StatelessWhile"});
```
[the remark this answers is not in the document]
rkazants: Agree, changed to 1 because I was unable to create `While` with empty inputs list.

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -11,6 +11,35 @@
 
 using namespace ov::frontend::tensorflow;
 
+namespace {
+template <typename T>
```
mvafin, line 15: I think this is too broad template, can it be at least `vector<T>`?

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -11,6 +11,35 @@
...
+T reorder_ops_by_names(const std::vector<std::string>& names, const T& ops) {
```
mvafin, line 16: I think this is too broad template, can it be `vector<T>` at least?

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -11,6 +11,35 @@
...
+                }
+                resulted_ops[ind] = op;
+            } else {
+                // not found operation with the requested name
+                // do not re-order then
+                return ops;
+            }
+        }
+    } else {
+        return ops;
```
mvafin, line 37: Shouldn't this case be error and not silent skip? Same question for 3 other cases of silent return

src/frontends/tensorflow/src/op/while.cpp
```diff
@@ -0,0 +1,115 @@
...
+        body_model,
+        "[TensorFlow Frontend] Internal error or incorrect input model. Cannot find body graph with name " + body_type);
+
+    // inject condition body graph prior to Loop node
+    // to check condition before to start iterations
+    ov::OutputVector ov_inputs;
+    for (int input_ind = 0; input_ind < input_size; ++input_ind) {
+        ov_inputs.push_back(node.get_input(input_ind));
+    }
+    auto cond_prior = cond_model->clone();
```
mvafin, line 47: It looks like every time you call `inject_body_model` you will do copy of the body, can it be done inside the `inject_body_model` function?

src/frontends/tensorflow/src/translate_session.cpp
rkazants: No, it should not because now Parameter and Result nodes re-ordering happens only for body-graphs. It was required to do for them first. For all other cases, we have to leave the legacy behavior, if names are not provided, for example, or something wrong is going.
This check will happen when we get more mature for these re-ordering to be used on the regular base.

src/frontends/tensorflow/src/op/while.cpp
rkazants: Let us think this sort of improvement in different PR because it will require changes in other files.
This change only for convenience to have code by one line shorter. Not critical

src/frontends/tensorflow/src/translate_session.cpp
rkazants: Done

src/frontends/tensorflow/src/translate_session.cpp
mvafin: For main body you will have empty `names` vector and can skip reordering in that case, but if you have two vectors of different sizes this is an error and should be thrown I think. Same if you didn't find the name. Otherwise you will not know what happened if reordering didn't happen for one of those reasons.

rkazants: Added checks and raise exceptions in suspicious cases

mvafin (approved):

PR #15051, opened 2023-01-11
[AUTO_BATCH]Create separated auto_batch plugin testcase

riverlijunjie:
### Details:
 - Create separated auto_batch plugin testcase sample

### Tickets:
 - 100607


src/tests/functional/plugin/meta/auto_batch/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set(TARGET_NAME ov_auto_batch_func_tests)
```
ilya-lavrenov, line 5: all tests should be located in plugin's folder.
for autobatch it's src/plugins/auto_batch/tests

riverlijunjie: ok

src/plugins/auto_batch/tests/functional/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+        LINK_LIBRARIES
+            funcSharedTests
+        DEPENDENCIES
+            openvino_auto_batch_plugin
+        ADD_CPPLINT
+        LABELS
+            Auto_Batch
+)
+if (ENABLE_INTEL_GPU)
+    target_compile_definitions(${TARGET_NAME} PRIVATE HAVE_GPU_PLUGIN)
```
riverlijunjie, line 25: @ilya-lavrenov do you know is there any unified method to decide which hw plugin should be tested or not? For example, the openvino package is built with gpu plugin enabled, but CI docker doesn't want to run any gpu plugin test case, is there any global control mechanism?

ilya-lavrenov: I don't know about such unified methods. You need in runtime understand whether specific device is here or not.. 
or maybe configure test binary with -d. So, you can run `ov_auto_func_tests -D GPU`

src/plugins/auto_batch/tests/functional/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+            openvino_auto_batch_plugin
+        ADD_CPPLINT
+        LABELS
+            Auto_Batch
+)
+
+if (ENABLE_INTEL_GPU)
+    find_package(OpenCL QUIET)
+    if(OpenCL_FOUND)
+        target_compile_definitions(${TARGET_NAME} PRIVATE HAVE_INTEL_GPU_PLUGIN)
```
ilya-lavrenov, line 48: why do we need to check OpenCL?

ilya-lavrenov: why do we need to split auto_batch tests from GPU or CPU?
I would treat having BATCH tests in cpuFUncTests are integration tests that CPU can work under auto-batch..

riverlijunjie: 1. Check OCL is to decide whether GPU plugin can be used, you can see that not all CI docker configure OCL which will cause GPU plugin cannot be loaded, and GPU plugin related auto_batch test cases failed.
2.  I'm not sure whether it is good idea to split auto_batch tests from GPU and CPU functional testcases, but if we will put all auto_batch related test cases into a separated test sample, it will be easier to maintain. Maybe we can still keep original auto_batch test cases in CPU/GPU functional test and add more auto_batch test cases into the separated test sample.

src/plugins/auto_batch/tests/functional/CMakeLists.txt
riverlijunjie: I prefer runtime checking, how about get_available_device() ?

riverlijunjie: Update to runtime decide whether run specific plugin tests or not.

ilya-lavrenov: BTW, it's even better to avoid usage of any HW plugin and test everything on MOCK objects, like it's done in caching tests

What do you think?

riverlijunjie: I think we can have 2 auto_batch test samples like cpu/gpu/gna plugin:
1. functional test: ov_auto_batch_func_test, which will use actual plugin to run test cases; 
2. unit test: ov_auto_batch_unit_test, which will use MOCK object to test auto_batch plugin.

Considered we still don't have unit test for auto_batch plugin, so firstly we create this functional test sample, and then create another separated PR for unit test with MOCK objects.

riverlijunjie: image

ilya-lavrenov: I agree that we need two types of tests, but:
- We need AUTO_BATCH (but it's applicable for other umbrella plugins) tests, which checks auto-batch only on all levels - on unit test level to check plugin internals, also it should cover AUTO_BATCH integration and invocation as a part of Core (we have some logic which checks whether auto-batch is applicable).
- We also need integration tests (behavior tests in out current notation), which are instantiated in plugin functional tests to check whether plugin is compatible with AUTO_BACTH (or any other umbrella plugin like HETERO). These tests don't test AUTO_BACTH, they test plugin. Having these tests is also important for Conformance testing to showcase that ANY plugin compliant or not with specific OpenVINO feature (auto batching, hetero execution, etc). For AUTO_BATCH it implies having optimal_batch_size property check for plugin, maybe support of model with batch > 1. But it should assume that such tests should test all operations with all possible batch sizes, because we should it as a prerequisite for AUTO_BATCH, it's not an integration point. 

Currently (in current master), we have only second type of test for AUTO_BATCH (for multi / auto we have unit tests and it's great). And as a part of this task we need to split these cases between 2 buckets: 
- does a test check whether ANY plugin can be run under AUTO_BATCH? 
- does a test is needed to test AUTO_BATCH itself?
Likely, we are OK to have some duplication here

src/plugins/auto_batch/tests/functional/CMakeLists.txt
ilya-lavrenov: AUTO_BATCH as a software feature as well its tests, should not not depend on GPU or OpenCL availability.
To check whether GPU plugin provides proper interface for communication with BATCH plugin, we need write tests specifically for this interface (e.g. whether optimal_batch_size is supported, maybe some other metrics)

src/plugins/auto_batch/tests/functional/CMakeLists.txt
riverlijunjie: Type second test above should contain integration tests:
    1). Test other plugin whether is compatible with AUTO_BATCH: such as "GPU", "HETRO:BATCH"
    2). Test whether plugin can run under AUTO_BATCH,  such as "BATCH:GPU(16)", "BATCH:CPU(4)"

For Type first test, there will be another WIP PR(https://github.com/openvinotoolkit/openvino/pull/15211) to cover:
    1) AUTO_BATCH itself internal tests
    2) Core level test for AUTO_BATCH

src/plugins/auto_batch/tests/functional/CMakeLists.txt
riverlijunjie: GPU or OCL dependents have been removed.
And we will add some tests for communication between AUTO_BATCH plugin and other plugins

src/plugins/auto_batch/tests/functional/behavior/executable_network/exec_net_base.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+using namespace BehaviorTestsDefinitions;
+namespace {
+
+const std::vector<InferenceEngine::Precision> netPrecisions = {InferenceEngine::Precision::FP32,
+                                                               InferenceEngine::Precision::FP16};
+auto autoBatchConfig = []() {
+    return std::vector<std::map<std::string, std::string>>{
+        // explicit batch size 4 to avoid fallback to no auto-batching (i.e. plain GPU)
+        {{CONFIG_KEY(AUTO_BATCH_DEVICE_CONFIG), std::string(CommonTestUtils::DEVICE_GPU) + "(4)"},
```
ilya-lavrenov, line 15: I suppose AUTO_BATCH should be tested without help of any other HW plugin, let's just use mock plugins (as in caching tests).
I understand that in some cases we still need to perform inference to check runtime part of AUTO_BATCH. For this purpose we can use TEMPLATE plugin.

src/plugins/auto_batch/tests/functional/behavior/ov_executable_network/properties.cpp
```diff
@@ -0,0 +1,52 @@
...
+const std::vector<ov::AnyMap> auto_batch_inproperties = {
+    {ov::num_streams(-100)},
+    {{CONFIG_KEY(AUTO_BATCH_DEVICE_CONFIG), std::string(CommonTestUtils::DEVICE_CPU) + "(4)"},
+     {ov::auto_batch_timeout(-1)}},
+#ifdef HAVE_TEMPLATE_PLUGIN
+    {ov::device::id("UNSUPPORTED_DEVICE_ID_STRING")},
+    {{CONFIG_KEY(AUTO_BATCH_DEVICE_CONFIG), std::string(CommonTestUtils::DEVICE_TEMPLATE) + "(4)"},
+     {ov::auto_batch_timeout(-1)}},
+#endif
+};
```
ilya-lavrenov, line 23: why do we have template plugin under conditional compilation, while CPU used without any guards?
As I wrote, better not rely on any plugin at all and use mocks.

src/plugins/auto_batch/tests/functional/behavior/plugin/auto_batching_tests.hpp
```diff
@@ -1,47 +1,46 @@
-// Copyright (C) 2018-2022 Intel Corporation
+// Copyright (C) 2018-2023 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
```
ilya-lavrenov, line 2: is this file about auto-batch testing or about integration between HW plugin and AUTO_BATCH?
If it's the second case, we need to keep this file in behavior tests

src/plugins/auto_batch/tests/functional/CMakeLists.txt
ilya-lavrenov: > Type second test above should contain integration tests:

Integration tests should be a part of plugin tests, IMO.
As I wrote, such tests should also be integrated into conformance to understand whether plugin is compatible with AUTO_BATCH?

src/plugins/auto_batch/tests/functional/behavior/ov_executable_network/properties.cpp
riverlijunjie: For AUTO_BATCH plugin tests, any hw plugin shouldn't be involved in, but we can leverage mock plugin to help create test case, in some special test cases we can leverage TEMPLATE plugin if possible.

src/plugins/auto_batch/tests/functional/CMakeLists.txt
riverlijunjie: I'm not sure the accurate scope of integration test and conformance test, could you give some clarification?

src/plugins/auto_batch/tests/functional/behavior/ov_executable_network/exec_net_base.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "behavior/ov_executable_network/exec_network_base.hpp"
+
+using namespace ov::test::behavior;
+namespace {
+auto autoBatchConfigs = []() {
+    return std::vector<ov::AnyMap>{
+        // explicit batch size 4 to avoid fallback to no auto-batching
+        {{CONFIG_KEY(AUTO_BATCH_DEVICE_CONFIG), std::string(CommonTestUtils::DEVICE_TEMPLATE) + "(4)"},
+         // no timeout to avoid increasing the test time
+         {CONFIG_KEY(AUTO_BATCH_TIMEOUT), "0 "}}};
```
ilya-lavrenov, line 14: can we use only new API in newly written tests?

src/plugins/auto_batch/tests/functional/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    ie_add_compiler_flags(/wd4273)
+    ie_add_compiler_flags(/wd4309)
+    ie_add_compiler_flags(/wd4804)
+endif()
+if(NOT MSVC)
+    ie_add_compiler_flags(-Wno-missing-declarations)
+    ie_add_compiler_flags(-Wno-sign-compare)
+    ie_add_compiler_flags(-Wno-odr)
+    ie_add_compiler_flags(-Wno-all)
+endif()
```
ilya-lavrenov, line 25: can we remove all warnings suppressions above?

src/plugins/auto_batch/tests/functional/CMakeLists.txt
wangleis: @ilya-lavrenov Could we handle AUTO_BATCH internal test and conformance test in separated PR?

wangleis (approved):
Please add TODO for conformance test

riverlijunjie:
> Please add TODO for conformance test

OK, let's create a separated PR for AUTO_BATCH conformance test

peterchen-intel (approved):

peterchen-intel (approved):

PR #15056, opened 2023-01-12
[GPU] Use bfyx format as defalut for scalar layout

wilson-seok:
### Details:
 - Use bfyx format as defalut for scalar layout

### Tickets:
 - 99948


wilson-seok:
No topology from dryrun for fp32/fp16/int8.

yeonbok (approved):

PR #15061, opened 2023-01-12
[GPU] Use in-order queue for devices with dpas support

vladimir-paramuzov:
### Details:
 - Fixes perf drop on dgpu caused by #14635  as OOO queue was created for all cases and onednn couldn't be used.


isanghao (approved):

isanghao:
It works fine now.

isanghao:
Perf issues are now all clear.

PR #14327, opened 2022-11-30
[Snippets][CPU] Added FP32 MHA tokenization support

a-sidorova:
### Details:
 - *Added FP32 MHA-pattern support to Snippets*

### Tickets:
 - *92360*

### Todo:
- [x]  *register new ops as CPU extension like in the [serialization PR#14064](https://github.com/openvinotoolkit/openvino/pull/14064)* (this branch rebase needed)
- [x] *Move changes in the common test infrastructure ([1](https://github.com/IvanNovoselov/openvino/blob/89490eae2b997c9dc13f865f8654663d6706a962/src/tests/functional/shared_test_classes/src/base/utils/generate_inputs.cpp#L46-L60) and [2](https://github.com/IvanNovoselov/openvino/blob/89490eae2b997c9dc13f865f8654663d6706a962/src/tests/functional/shared_test_classes/include/shared_test_classes/base/ov_subgraph.hpp#L64-L70)) to a separate PR, or drop 1 and move 2 to the snippets utils*
- [x] *Reset src/tests/functional/shared_test_classes/include/shared_test_classes/base/ov_subgraph.hpp and
src/tests/functional/shared_test_classes/src/base/utils/generate_inputs.cpp to master after [14343](https://github.com/openvinotoolkit/openvino/pull/14343) is merged.*
- [ ] *Follow-up: remove extra parameters from `emit_brgemm_kernel_call` in the separate PR*
### Blocked by:
- small fix on test infrastructure [14343](https://github.com/openvinotoolkit/openvino/pull/14343). UPD: Unblocked

PR #15063, opened 2023-01-12
[Snippets] Added support of BF16/I8/U8 for MatMul

a-sidorova:
### Details:
 - *Added support of INT8, BF16 for MatMul on platforms with VNNI, bf16 support and AMX*

### Tickets:
 - *102166*

### Blockers:
- [x] [PR#14996](https://github.com/openvinotoolkit/openvino/pull/14996)

### TODO:
- [x]  After merging of [PR#14996](https://github.com/openvinotoolkit/openvino/pull/14996) need to add support of `get_supported_precisions` for `Brgemm`


rkazants:
Wow, tons of the code:)

a-sidorova:
> Wow, tons of the code:)

The main part of the code is here https://github.com/openvinotoolkit/openvino/pull/14327 :yum:

src/common/snippets/src/op/memory_access.cpp
```diff
@@ -12,32 +12,66 @@ namespace ngraph {
...
+        visitor.on_attribute("offset_out_" + std::to_string(i), port.m_offset);
+    }
     return true;
 }
 
-size_t MemoryAccess::get_count() const {
-    return m_count;
+void MemoryAccess::set_input_port_descriptor(const PortDescriptor& desc, const size_t i) {
+    // Logic is as same as ov::Node::get_input_descriptor
+    while (m_input_ports.size() <= i) {
```
IvanNovoselov, line 33: Why don't we check that `i < get_input_size()`? It's unclear why we can set 100 input descriptors for an operation that has only 2 inputs. Same for outputs.

src/common/snippets/src/op/store.cpp
```diff
@@ -12,13 +12,19 @@ namespace ngraph {
 namespace snippets {
 namespace op {
 
-snippets::op::Store::Store(const Output<Node>& x, const size_t count, const size_t offset) : MemoryAccess({x}, count, offset) {
+snippets::op::Store::Store(const Output<Node>& x, const size_t count, const size_t offset) : MemoryAccess({x}) {
     constructor_validate_and_infer_types();
+    set_output_port_descriptor({count, offset}, 0);
```
IvanNovoselov, line 17: `set_output_port_descriptor` is (almost?) always called with an rvalue, so why don't we add a corresponding parameter to a `MemoryAccess` constructor? The parameter could even be mandatory to protect us from the cases when one simply forgets to manually call `set_output_port_descriptor` (and this will happen sooner or later).

src/common/snippets/src/op/subgraph.cpp
```diff
@@ -432,21 +432,20 @@ void snippets::op::Subgraph::initialize_buffer_scratchpad_size() {
...
-            if (auto store = ov::as_type_ptr<snippets::op::Store>(parent)) {
-                store->set_offset(offset);
-            } else if (const auto brgemm = ov::as_type_ptr<snippets::op::Brgemm>(parent)) {
-                // Brgemm encapsulates work with loading and storing of data
-                brgemm->set_offset_c(offset);
-            } else {
-                throw ngraph_error("Buffer::set_offset() was called when Buffer didn't have the corresponding Store op for offset propagation");
+            if (!ov::is_type<ngraph::op::v0::Constant>(parent)) {
+                auto idx = buffer->input(0).get_source_output().get_index();
+                while (ov::is_type<snippets::op::LoopBase>(parent)) {
```
IvanNovoselov, line 437: This approach works nicely in the cases when there are several consecutive LoopEnd's, and it's perfect for now. However, it would throw in the following pattern:
```
EltwiseSomething
Store (connected to Buffer)
LoopEnd
HorizonSomething
LoopEnd
Buffer
```
Let's leave it as is, but keep such cases in mind for the future.

IvanNovoselov (commented):
Part #1

src/plugins/intel_cpu/src/emitters/jit_snippets_emitters.cpp
```diff
@@ -904,29 +942,228 @@ void BrgemmEmitter::emit_brgemm_kernel_call(const brgemm_kernel_t *brgKernel, in
...
+}
+
+void BrgemmEmitter::kernel_call(const brgemm_kernel_t *brg_kernel,
+                                const std::vector<Reg64>& regs,
+                                const std::vector<size_t>& offsets) const {
+    if (regs.size() != offsets.size() || regs.size() != 3) {
+        IE_THROW() << "Brgemm emitter got unexpected register count and offset count: " << regs.size() << " and " << offsets.size();
+    }
+
+    size_t gpr_size = 8;
```
IvanNovoselov, line 1008: looks like a perfect candidate to become a static constexpr in the base class)

src/plugins/intel_cpu/src/emitters/jit_snippets_emitters.cpp
```diff
@@ -904,29 +942,228 @@ void BrgemmEmitter::emit_brgemm_kernel_call(const brgemm_kernel_t *brgKernel, in
...
+std::vector<size_t> BrgemmEmitter::init_kernel_offsets(size_t mb, size_t M_blk, size_t LDA, size_t LDC,
+                                                       size_t k, size_t K0_step0, size_t K0_step1,
+                                                       size_t n, size_t N0_step0, size_t N0_step1) const {
+    const size_t in0_offset = load_offset_a + (k * K0_step0 + mb * M_blk * LDA) * io_data_size[0];
+    const size_t in1_offset = load_offset_b + (k * K0_step1 + n * N0_step0) * io_data_size[1];
+    const size_t out0_offset = store_offset_c + (n * N0_step1 + mb * M_blk * LDC) * io_data_size[2];
+    return { in0_offset, in1_offset, out0_offset };
+}
+
+void BrgemmEmitter::kernel_call(const brgemm_kernel_t *brg_kernel,
```
IvanNovoselov, line 1001: Code of this emitter is 99.32% identical to the `BrgemmWithScratchEmitter::kernel_call`. Can we generalize them to avoid repetition? It seems at he first glance that we just need to insert a couple of `if(regs.size() == 3)`

src/plugins/intel_cpu/src/emitters/jit_snippets_emitters.cpp
```diff
@@ -904,29 +942,228 @@ void BrgemmEmitter::emit_brgemm_kernel_call(const brgemm_kernel_t *brgKernel, in
...
+                                                              void*,
+                                                              void*,
+                                                              int)>(kernel_execute);
+    h->mov(h->rbp, reinterpret_cast<uintptr_t>(brgemm_kernel_overload));
+    // todo: several of addr_{A, B, C} could be also abi_paramX, so one of them could be corrupted
+    //  if moving directly h->uni_vmovq(abi_paramX, adr_X). Save them to vector regs to avoid corruption.
+    //  It's likely that a more efficient solution exists.
+    h->uni_vmovq(Xmm(0), regs[0]); // A
+    h->uni_vmovq(Xmm(1), regs[1]); // B
+    h->uni_vmovq(Xmm(2), regs[2]); // Scratch
```
IvanNovoselov, line 1138: We can reorder registers in `regs` to match better with `BrgemmEmitter::kernel_call` (e.g. Scratch, if present, goes last)

src/plugins/intel_cpu/src/emitters/jit_snippets_emitters.cpp
```diff
@@ -937,87 +1174,348 @@ void BrgemmEmitter::emit_brgemm_kernel_call(const brgemm_kernel_t *brgKernel, in
...
     for (int i = n_gprs_to_save - 1; i >= 0; --i)
         h->mov(gprs_to_save[i], h->ptr[h->rsp + i * gpr_size]);
     h->add(h->rsp, n_gprs_to_save * gpr_size);
 }
 
-void BrgemmEmitter::kernel_execute(const brgemm_kernel_t *brg_kernel, const void *A, const void *B, void *C) {
-    // TODO: There are 4 available abi_params on Windows so we have the copy of brgemm_kernel_execute() function
-    //       with 4 runtime parameters (kernel and I/O) and 4 default parameter values (batch, bs and scratch)
-    brgemm_kernel_params_t brgemm_p;
+void BrgemmCopyBBaseEmitter::execute(matmul::jit_brgemm_matmul_copy_b_t *kernel, const void *src,
```
IvanNovoselov, line 1354: Similarly to `BrgemmEmitter` and `BrgemmWithScratchEmitter`, `BrgemmCopyBEmitter::kernel_call` and `BrgemmCopyBWithCompensationsEmitter::kernel_call` could be merged into one method using just a couple tweaks, and it looks that we should do it.

A more ambitious goal to think about, is whether we can create a wrapper that will be able to emit code to call an arbitrary precompiled binary. It seems that it's completely doable if we add a vector uint64_t values to the kernell_call args. The values will be treated as immediates and will be passed as arguments after the `regs`. We might even need two maps instead of vectors: map{position:pair{reg, offset}} and map{position:imm} to allow for an arbitrary order of reg and imm arguments. What do you think about the idea?

src/plugins/intel_cpu/src/emitters/jit_snippets_emitters.cpp
```diff
@@ -937,87 +1174,348 @@ void BrgemmEmitter::emit_brgemm_kernel_call(const brgemm_kernel_t *brgKernel, in
...
+    brgCopyKernelConf.wei_n_blk = N_blk;
+    brgCopyKernelConf.wei_tag = dnnl_abcd;  // What's about other ranks?
+    brgCopyKernelConf.copy_B_wei_stride = 0;
+    brgCopyKernelConf.LDB = LDB;
+    brgCopyKernelConf.N = N;
+    brgCopyKernelConf.N_tail = N_tail;
+    brgCopyKernelConf.N_blk = N_blk;
+    brgCopyKernelConf.K = K;
+    brgCopyKernelConf.K_blk = K;
+    brgCopyKernelConf.N_chunk_elems = brgCopyKernelConf.N_blk;
```
IvanNovoselov, line 1250: Minor: 
1) To avoid dealing with code-analyzer reports later, it would be nice to eliminate implicit signed-unsigned conversions here (L1241, 12,44-12499)
2) The function parameter have the same names as class fields.

IvanNovoselov (commented):
Part No 2

src/common/snippets/include/snippets/op/load.hpp
```diff
@@ -20,11 +20,15 @@ namespace op {
...
 class Load : public MemoryAccess {
 public:
-    OPENVINO_OP("Load", "SnippetsOpset");
+    OPENVINO_OP("Load", "SnippetsOpset", MemoryAccess);
 
     Load(const Output<Node>& x, const size_t count = 1lu, const size_t offset = 0lu);
     Load() = default;
 
+    size_t get_offset() const { return get_input_port_descriptor(0).m_offset; }
+    size_t get_count() const { return get_input_port_descriptor(0).m_count; }
```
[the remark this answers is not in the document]
IvanNovoselov: Minor (here and in LIR loop fusion also) when the class members are public (as for PortDescriptor), there is no need for the `m_` prefix, do it would be simply `get_input_port_descriptor(0).offset`. What do you think?

src/common/snippets/include/snippets/op/buffer.hpp
```diff
@@ -13,6 +13,8 @@ namespace op {
 /**
  * @interface Buffer
  * @brief This is a base class for memory storage.
+ *        If Buffer has a parent, the operation is for intermediate data storage - Intermediate type.
```
IvanNovoselov, line 16: Minor: Let's be precise: `Intermediate` --> `IntermediateMemory`, `Empty` --> `NewMemory`

src/common/snippets/include/snippets/op/buffer.hpp
```diff
@@ -26,20 +23,66 @@ class Buffer : public ngraph::op::Op {
...
+ *        saves the input shape for shape inference and input element type.
+ *        For example,
+ *              Parameter [5, 3, 128]    Constant [2] (with values {3, 128})
+ *                     \                 /
+ *                  Buffer with allocated memory 3x128 size
+ *                              |
+ *                       Result [5, 3, 128]
+ * @ingroup snippets
+ */
+class IntermediateBuffer : public Buffer {
```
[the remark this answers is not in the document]
IvanNovoselov: Obsolete after refactoring: there is one class for all buffers now

src/common/snippets/include/snippets/op/memory_access.hpp
```diff
@@ -48,14 +23,46 @@ class MemoryAccess : public ngraph::op::Op {
...
     void set_output_port_descriptor(const PortDescriptor& desc, const size_t i);
-    PortDescriptor get_input_port_descriptor(const size_t i) const;
-    PortDescriptor get_output_port_descriptor(const size_t i) const;
-    PortDescriptor& get_input_port_descriptor(const size_t i);
-    PortDescriptor& get_output_port_descriptor(const size_t i);
+    const PortDescriptor& get_input_port_descriptor(const size_t i) const;
+    const PortDescriptor& get_output_port_descriptor(const size_t i) const;
+
+    void set_input_count(size_t count, size_t idx);
+    void set_output_count(size_t count, size_t idx);
```
IvanNovoselov, line 54: 1) It makes sense to have a default `idx = 0` value
2) We used to have only one count for both inputs and outputs, so I'm a little confused here now. If I want to set count for `Load` should I use `set_input_count` or `set_output_count`? And if only `set_input_count` is legal for such operations, what happens if I `set_output_count`? I've seen that you created `set_count` for `Load` and `Store`, and I think this is a right direction, but still `set_input_count` [ could be called](https://github.com/openvinotoolkit/openvino/blob/5deb3c326e8845fd43cce555d70aa066c2f8db9d/src/common/snippets/src/pass/vector_to_scalar.cpp#L27)  on load, which is a bit confusing. I don't insist, but should we make a separate class for single-port operations maybe?

src/common/snippets/src/pass/collapse_subgraph.cpp
```diff
@@ -49,9 +49,16 @@ auto outputs_are_not_broadcastable(const std::shared_ptr<const Node>& node) -> b
...
         const auto& out_shape = n->get_output_partial_shape(0);
-        return matmul && out_shape.is_static() && out_shape.size() == 4;
+        if (!matmul || out_shape.is_dynamic() || out_shape.size() != 4)
+            return false;
+        const auto intype_0 = matmul->get_input_element_type(0);
+        const auto intype_1 = matmul->get_input_element_type(1);
+        const bool is_f32 = intype_0 == element::f32 && intype_1 == element::f32;
+        const bool is_int8 = (intype_0 == element::i8 || intype_0 == element::u8) && (intype_1 == element::i8);
+        const bool is_bf16 = intype_0 == element::bf16 && intype_1 == element::bf16;
+        return is_f32 || is_bf16 || is_int8;
```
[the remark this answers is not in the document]
IvanNovoselov: Just to clarify, it's not in plugin callback now, but in [collapse subgraph](https://github.com/openvinotoolkit/openvino/blob/10d5eb896b5ad94ef46db5a817196645a3ae5639/src/common/snippets/src/pass/collapse_subgraph.cpp#L58-L61)

src/plugins/intel_cpu/src/emitters/jit_snippets_emitters.hpp
```diff
@@ -336,33 +336,34 @@ class BrgemmEmitter : public jit_emitter {
         bool is_with_comp;
         float beta;
     };
-    void initBrgemm(brgemmCtx& ctx, std::unique_ptr<brgemm_kernel_t>& brgKernel, bool use_amx) const;
-    void callBrgemm(brgemmCtx& ctx, std::unique_ptr<brgemm_kernel_t>& brgKernel, const void* pin0, const void* pin1, void* pout, void* wsp) const;
+    void initBrgemm(brgemmCtx& ctx, std::unique_ptr<dnnl::impl::cpu::x64::brgemm_kernel_t>& brgKernel, bool use_amx) const;
+    void callBrgemm(brgemmCtx& ctx, std::unique_ptr<dnnl::impl::cpu::x64::brgemm_kernel_t>& brgKernel,
```
IvanNovoselov, line 340: `callBrgemm` is not implemented, please remove

src/plugins/intel_cpu/src/emitters/jit_snippets_emitters.cpp
```diff
@@ -734,54 +738,53 @@ BrgemmEmitter::BrgemmEmitter(dnnl::impl::cpu::x64::jit_generator* h, dnnl::impl:
...
-    brg0VnniFactor = 4 / brg0Prc.size();
-    bool brgWithAMX = isAMXSupported && brg0Prc != Precision::FP32 && (K % brg0VnniFactor == 0) && (N % brg0VnniFactor == 0);
+    m_brg0VnniFactor = 4 / brg0Prc.size();
+    bool brgWithAMX = brgemm_node->is_amx();
 
-    with_scratch = brgemm_node->get_input_size() == 3;
-    with_comp = !brgWithAMX && brg0Prc == Precision::I8;
+    m_with_comp = brgemm_node->is_with_compensations();
+    m_with_scratch = brgemm_node->is_with_scratchpad();
+    OPENVINO_ASSERT((m_with_scratch && brgemm_node->get_input_size() == 3) || !m_with_scratch, "Brgemm with scratchpad expect 3 inputs");
```
IvanNovoselov, line 762: It's a bit to late to check graph validity, isn't it?)
It looks like this check should be in brgemm_cpu validate_and_infer_types.
What do you think?

src/plugins/intel_cpu/src/emitters/jit_snippets_emitters.cpp
```diff
@@ -734,54 +738,53 @@ BrgemmEmitter::BrgemmEmitter(dnnl::impl::cpu::x64::jit_generator* h, dnnl::impl:
...
+                auto N_ = n ? m_N_tail : m_N - m_N_tail;
+                auto K_ = k ? m_K_tail : m_K - m_K_tail;
+                auto beta = k && m_brgCtxs0[getBrgIdx(m, 0, n)].K != 0 ? 1.0f : 0.0f;
 
                 brgemmCtx.M = M_;
                 brgemmCtx.N = N_;
                 brgemmCtx.K = K_;
                 brgemmCtx.LDA = leading_dimensions[0];
-                brgemmCtx.LDB = brg1Prc == Precision::FP32 ? leading_dimensions[1] : rnd_up(N, N_blk);
+                brgemmCtx.LDB = brg1Prc == Precision::FP32 ? leading_dimensions[1] : rnd_up(m_N, m_N_blk);
```
IvanNovoselov, line 787: Just curious, why do LBD values depend on precision? Is it a feature of non fp-32 primitives?

src/common/snippets/include/snippets/op/memory_access.hpp
a-sidorova: 1. Nice idea. let's do it, thanks
2. `Load` works with data on input and with vector register on output. So you should work with `set_input_count`. And opposite for `Store`. I thought that it seems like logical things. We can discuss it

src/plugins/intel_cpu/src/emitters/jit_snippets_emitters.cpp
a-sidorova: I transferred it from MHA custom implementation. I think it depends on precision because if there is FP32, we don't need `BrgemmCopyB` otherwise we need it. I think that we should here another condition. For example, `is_with_data_repacking()`. Thanks a lot for the catch!

src/common/snippets/src/op/load.cpp
```diff
@@ -13,12 +13,15 @@ namespace snippets {
 namespace op {
 
 Load::Load(const Output<Node>& x, const size_t count, const size_t offset) : MemoryAccess({x}) {
-    constructor_validate_and_infer_types();
+    m_input_ports.resize(get_output_size());
     set_input_port_descriptor({count, offset}, 0);
```
IvanNovoselov, line 17: `set_input_port_descriptor` is a virtual member now, so we should explicitly specify the class 
`MemoryAccess::set_input_port_descriptor(...)`

src/common/snippets/src/op/store.cpp
```diff
@@ -13,12 +13,15 @@ namespace snippets {
 namespace op {
 
 snippets::op::Store::Store(const Output<Node>& x, const size_t count, const size_t offset) : MemoryAccess({x}) {
-    constructor_validate_and_infer_types();
+    m_output_ports.resize(get_output_size());
     set_output_port_descriptor({count, offset}, 0);
```
IvanNovoselov, line 17: `MemoryAcess::set....`

src/common/snippets/src/op/load.cpp
```diff
@@ -29,13 +32,11 @@ std::shared_ptr<Node> Load::clone_with_new_inputs(const OutputVector& new_args)
...
 void Load::set_output_port_descriptor(const MemoryAccess::PortDescriptor& desc, const size_t i) {
-    // Load is one-port MemoryAccess operation. To simulate this behavior input_desc = output_desc
-    set_input_port_descriptor(desc, i);
+    throw ov::Exception("Load node doesn't have memory access output port");
 }
 
 const MemoryAccess::PortDescriptor& Load::get_output_port_descriptor(const size_t i) const {
-    // Load is one-port MemoryAccess operation. To simulate this behavior input_desc = output_desc
-    return get_input_port_descriptor(i);
+    throw ov::Exception("Load node doesn't have memory access output port");
```
IvanNovoselov, line 39: 1) You don't have to override `get_output_port_descriptor` now, since the MemoryAcess implementation will throw [anyway](https://github.com/openvinotoolkit/openvino/blob/6b80c5de66f042f7d414ed03748c75d7dfdbd97a/src/common/snippets/src/op/memory_access.cpp#L53), as far as we don't have and output port desc.
2) The same is true for `Store::get_input_port_desc`
3) We still have one vulnerability in this approach: one can dynamic_cast Load to MemoryAccess and create (invalid) out port descriptor by calling `MemoryAcess::validate_and_infer_types()`. Therefore I suggest to remove `m_input_ports` [resize](https://github.com/openvinotoolkit/openvino/blob/6b80c5de66f042f7d414ed03748c75d7dfdbd97a/src/common/snippets/src/op/memory_access.cpp#L18-L23) from there, it should happen in a child class constructor, not here.

IvanNovoselov (approved):
Approve with minor leftovers

src/common/snippets/src/op/load.cpp
a-sidorova: 1. Oh, really, thanks!
2. Done
3. Right. I have added parameters `input_count` and `output_coun` for `MemoryAcccess` ctor to avoid code duplication in general. But anyway we can manually set them in derived class ctors. Thanks!

src/common/snippets/src/op/load.cpp
a-sidorova: When we needed override methods, it's valid. But now as you said, we don't have to override them and make virtual them. So I removed `virtual` and left just s`et_input_port_descriptor()`. Thanks!

dmitry-gorokhov (approved):

PR #15073, opened 2023-01-12
[Core] Safe loading of default plugins

vurusovs:
### Details:
 - In case of default `plugins.xml` all plugins are registered by absolute paths
 - In case of user API user is able to specify abs or rel path or plugin name to be found in CWD or ENV. But searching library in `openvino-<version>` folder was intentionally removed, to do so user should use default `plugins.xml` for packing custom libraries together with default libraries.
**NOTE:** it breaks behaviour like
```
core.unregister_plugin("CPU")
core.register_plugin("CPU", "libopenvino_intel_cpu.so")
```
because Core doesn't store information about default places of plugins. This is also a reason why some C and Py "register_plugin*" tests **FAIL** - `register_plugin("CPU", "libopenvino_intel_cpu.so")` works only if correct path to libs is specified in LD_LIBRARY_PATH. So the tests will be moved on "libmock_engine.so" because tests know where that lib is
 - Update `ov::util::get_absolute_file_path()` in order to prevent checking file exists or can be accessed
 - For testing purposes `getPluginPath()` were moved from anonymous namespace to ov

### Tickets:
 - 96201, 97906, 97899


src/inference/src/core.cpp
```diff
@@ -48,11 +48,17 @@ class Core::Impl : public CoreImpl {
 Core::Core(const std::string& xmlConfigFile) {
     _impl = std::make_shared<Impl>();
 
+    auto _registerPlugins = [&]() {
```
ilyachur, line 51: Why do you need this change?

src/inference/tests/functional/shared_object_loader_test.cpp
```diff
@@ -1,68 +0,0 @@
...
-#include <cpp/ie_plugin.hpp>
-
-#include "common_test_utils/file_utils.hpp"
-#include "openvino/util/shared_object.hpp"
-
-using namespace std;
-using namespace InferenceEngine;
-using namespace InferenceEngine::details;
-
-class SharedObjectLoaderTests : public ::testing::Test {
```
ilyachur, line 17: Why did you remove this file?

src/common/util/src/file_util.cpp
```diff
@@ -351,14 +352,26 @@ std::wstring ov::util::string_to_wstring(const std::string& string) {
 std::string ov::util::get_absolute_file_path(const std::string& path) {
     std::string absolutePath;
     absolutePath.resize(MAX_ABS_PATH);
-    auto absPath = get_absolute_path(&absolutePath[0], path);
-    if (!absPath) {
-        std::stringstream ss;
-        ss << "Can't get absolute file path for [" << path << "], err = " << strerror(errno);
-        throw std::runtime_error(ss.str());
+    (void*)get_absolute_path(&absolutePath[0], path);
```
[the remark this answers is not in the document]
vurusovs: The original problem is that Linux `realpath` checks file existence and read access, but `_fullpath` on Windows - no. And to guarantee full path generation without exceptions I need to skip/workaround that check. It's required to prevent such exceptions on plugins registration step in order to postpone failure on plugin loading step (e.g. if there is no GNA plugin file, I expect failure only on loading of GNA, but not on registration of paths to CPU/GPU/GNA etc. common step)

According to https://man7.org/linux/man-pages/man3/realpath.3.html, `realpath() returns a pointer to the resolved_path` (i.e. `&absolutePath[0]` and return pointer will be same). But in notes:
>    GNU extensions
       If the call fails with either EACCES or ENOENT and resolved_path
       is not NULL, then the prefix of path that is not readable or does
       not exist is returned in resolved_path.

So, if file doesn't exist or no access, `realpath` will return NULL, `&absolutePath[0]` is undefined, but according to notes it should contain the path that we expect. So with the change I just make a workaround on `EACCES or ENOENT` errors

To be sure it works, I checked it with GCC and CLANG, and both works. Otherwise I expect `&absolutePath[0]` should contain NULL as `realpath` return value (but it wasn't guaranteed). 

P.S. Windows `_fullpath` is very close to Linux `realpath`, but doesn' check existence and access (it checks correctness of disk letter or size of char array, but I can't achieve NULL from `_fullpath`). Details: https://learn.microsoft.com/ru-RU/cpp/c-runtime-library/reference/fullpath-wfullpath?view=msvc-170#return-value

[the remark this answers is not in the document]
vurusovs: I may add more strict checks like:
`
retval = get_absolute_path(&absolutePath[0], path);
if (retval == NULL)
    if (errno == ENOENT || errno == EACCES)  // error codes are equal both on Win or Lin
        if (&absolutePath[0])
            return &absolutePath[0];
    throw std::runtime_error;
return retval;

`

src/inference/src/core.cpp
vurusovs: For the moment `Core` ctor may throw next exceptions - ov::Exception, InferenceEngine::Exception (via IE_THROW), std::runtime_error. And previosly, when `register_plugins` was called, IE::Exception was converted to ov::Exception by `OV_CORE_CALL_STATEMENT`. Because I extended number of arguments in 
`RegisterPluginsInRegistry()`, but not for `register_plugins`, I call `RegisterPluginsInRegistry()` directly. So `Core` started throwing IE::Exception, and it led to tests failures

After introducing of `OV_CORE_CALL_STATEMENT` it was like that:
```
OV_CORE_CALL_STATEMENT({
#ifdef OPENVINO_STATIC_LIBRARY
        _impl->RegisterPluginsInRegistry(::getStaticPluginsRegistry());
#else
        // If XML is default, load default plugins by absolute paths
        auto loadByAbsPath = xmlConfigFile.empty();
        _impl->RegisterPluginsInRegistry(findPluginXML(xmlConfigFile), loadByAbsPath);
#endif
})
```
but on Windows it led to build failure:
```
'#': invalid character: possibly the result of a macro expansion
'ifdef': undeclared identifier
syntax error: missing ';' before identifier 'OPENVINO_STATIC_LIBRARY'
'OPENVINO_STATIC_LIBRARY': undeclared identifier
syntax error: missing ';' before identifier '_impl'
'getStaticPluginsRegistry': is not a member of '`global namespace''
'getStaticPluginsRegistry': identifier not found
illegal else without matching if
'endif': undeclared identifier
syntax error: missing ';' before '}'
```

So the final solution is like that

src/inference/tests/functional/shared_object_loader_test.cpp
vurusovs: Because same tests are presented in [src\inference\tests\functional\ov_shared_object_test.cpp](https://github.com/openvinotoolkit/openvino/blob/master/src/inference/tests/functional/ov_shared_object_test.cpp)

src/inference/src/core.cpp
vurusovs: I decided to remove lambda and replace it like that:
```
#ifdef OPENVINO_STATIC_LIBRARY
        OV_CORE_CALL_STATEMENT(_impl->RegisterPluginsInRegistry(::getStaticPluginsRegistry());)
#else
        OV_CORE_CALL_STATEMENT({
             // If XML is default, load default plugins by absolute paths
            auto loadByAbsPath = xmlConfigFile.empty();
            _impl->RegisterPluginsInRegistry(findPluginXML(xmlConfigFile), loadByAbsPath);
        })
#endif
```

src/inference/include/ie/ie_core.hpp
```diff
@@ -293,13 +293,12 @@ class INFERENCE_ENGINE_API_CLASS(Core) {
...
-     * prefix to identify library full name
+     * @param plugin Path (absolute or relative) or name of a plugin. Depending on platform `plugin` is wrapped with
+     * shared library suffix and prefix to identify library full name
      *
-     * @param deviceName A device name to register plugin for. If device name is not specified, then it's taken from
-     * plugin itself.
+     * @param deviceName A device name to register plugin for
      */
-    void RegisterPlugin(const std::string& pluginName, const std::string& deviceName);
+    void RegisterPlugin(const std::string& plugin, const std::string& deviceName);
```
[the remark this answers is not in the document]
vurusovs: Done

src/bindings/python/src/compatibility/openvino/inference_engine/ie_api.pyx
```diff
@@ -541,10 +541,10 @@ cdef class IECore:
 
 
     def register_plugin(self, plugin_name: str, device_name: str = ""):
```
vurusovs, line 543: @ilyachur @ilya-lavrenov I decided to leave argument name as is because it may be used as `register_plugin(plugin_name=...)`. We may also change it to achieve full compatibility with other APIs, but it will break Py compatibility

src/inference/include/openvino/runtime/core.hpp
```diff
@@ -638,17 +638,20 @@ class OPENVINO_RUNTIME_API Core {
...
+     * shared library suffix and prefix to identify library full name.
      * For example, on Linux platform, plugin name specified as `plugin_name` will be wrapped as `libplugin_name.so`.
      * Plugin search algorithm:
-     * - If plugin is located in the same directory as OpenVINO runtime library, it will be used.
-     * - If no, plugin is tried to be loaded from paths pointed by PATH/LD_LIBRARY_PATH/DYLD_LIBRARY_PATH
+     * - If `plugin` points to an exact library path (absolute or relative), it will be used.
+     * - If `plugin` specifies file name (`libplugin_name.so`) or plugin name (`plugin_name`), it will be searched by
+     *   file name (`libplugin_name.so`) in CWD or in paths pointed by PATH/LD_LIBRARY_PATH/DYLD_LIBRARY_PATH
      *   environment variables depending on the platform.
+     * Notes:
```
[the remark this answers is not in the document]
vurusovs: Done, thanks!

src/common/util/src/file_util.cpp
```diff
@@ -351,14 +352,26 @@ std::wstring ov::util::string_to_wstring(const std::string& string) {
...
+    if (!absolutePath.empty()) {
+        // on Linux if file not exist or no access, function will return NULL, but
```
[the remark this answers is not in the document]
vurusovs: Done, thanks!

src/inference/include/openvino/runtime/core.hpp
```diff
@@ -681,10 +683,12 @@ class OPENVINO_RUNTIME_API Core {
      *
      * - `name` identifies name of a device enabled by a plugin.
      * - `location` specifies absolute path to dynamic library with a plugin.
-     *    The path can also be relative to inference engine shared library. It allows having common config
+     *    The path can also be relative to XML file directory. It allows having common config
      *    for different systems with different configurations.
      * - `properties` are set to a plugin via the ov::Core::set_property method.
      * - `extensions` are set to a plugin via the ov::Core::add_extension method.
+     * Notes:
```
vurusovs, line 690: @note

src/inference/src/core.cpp
```diff
@@ -49,9 +49,13 @@ Core::Core(const std::string& xmlConfigFile) {
...
 
 #ifdef OPENVINO_STATIC_LIBRARY
-    _impl->register_plugins_in_registry(::getStaticPluginsRegistry());
+    OV_CORE_CALL_STATEMENT(_impl->register_plugins_in_registry(::getStaticPluginsRegistry());)
 #else
-    register_plugins(findPluginXML(xmlConfigFile));
+    OV_CORE_CALL_STATEMENT({
+        // If XML is default, load default plugins by absolute paths
+        auto loadByAbsPath = xmlConfigFile.empty();
+        _impl->register_plugins_in_registry(findPluginXML(xmlConfigFile), loadByAbsPath);
```
ilyachur, line 57: Nitpick: just to avoid introduction of additional local variables
```suggestion
    OV_CORE_CALL_STATEMENT({
        // If XML is default, load default plugins by absolute paths
        _impl->register_plugins_in_registry(findPluginXML(xmlConfigFile), xmlConfigFile.empty());
```

src/inference/src/dev/core_impl.cpp
```diff
@@ -102,7 +59,65 @@ ov::CoreImpl::CoreImpl(bool _newAPI) : m_new_api(_newAPI) {
     }
 }
 
-void ov::CoreImpl::register_plugins_in_registry(const std::string& xmlConfigFile) {
+ov::util::FilePath ov::getPluginPath(const std::string& plugin) {
```
ilyachur, line 62: Could you use snake_case for new methods?

src/inference/src/dev/core_impl.hpp
```diff
@@ -250,8 +253,9 @@ class CoreImpl : public InferenceEngine::ICore, public std::enable_shared_from_t
      * @brief Register plugins for devices which are located in .xml configuration file.
      * @note The function supports UNICODE path
      * @param xmlConfigFile An .xml configuraion with device / plugin information
+     * @param ByAbsPath A boolean value - register plugins by absolute file path or not
      */
-    void register_plugins_in_registry(const std::string& xmlConfigFile);
+    void register_plugins_in_registry(const std::string& xmlConfigFile, const bool& ByAbsPath = false);
```
ilyachur, line 258: Please use snake_case
```suggestion
    void register_plugins_in_registry(const std::string& xmlConfigFile, bool by_absolute_path = false);
```

src/inference/src/core.cpp
vurusovs: Is it readable/clear enough that `xmlConfigFile.empty()` means plugins will be loaded by absolute paths only?

src/inference/src/dev/core_impl.hpp
vurusovs: Are mixed types OK here? I mean `xmlConfigFile` and `by_absolute_path`

ilyachur: You can change all names :)

src/inference/src/dev/core_impl.cpp
vurusovs: I will do it in next PR

src/inference/src/dev/core_impl.hpp
vurusovs: I will do it in next PR

src/inference/src/core.cpp
vurusovs: I will do it in next PR

src/bindings/python/src/compatibility/openvino/inference_engine/ie_api.pyx
```diff
@@ -541,10 +541,10 @@ cdef class IECore:
...
-        """Registers plugins specified in an `.xml` configuration file
+        """Register a new device and plugin that enables this device inside OpenVINO Runtime.
         
-        :param plugin_name: A name of a plugin. Depending on a platform, plugin_name is wrapped with a shared
-                            library suffix and a prefix to identify a full name of the library
+        :param plugin_name: A path (absolute or relative) or name of a plugin. Depending on platform, 
+                            `plugin` is wrapped with shared library suffix and prefix to identify library full name
```
vurusovs, line 547: rename `plugin` to `plugin_name`

.ci/azure/linux_debian.yml
```diff
@@ -225,7 +225,10 @@ jobs:
 
     # Skip test_onnx/test_zoo_models and test_onnx/test_backend due to long execution time
   - script: |
+      # TODO (vurusovs): revert skip of test_core.py::test_register_plugin*,
```
ilyachur, line 228: Please create a ticket on this

Can we be sure that we won't have another issues in other places?

.ci/azure/linux_debian.yml
```diff
@@ -348,7 +354,9 @@ jobs:
     displayName: 'CPU FuncTests'
 
   - script: |
-      $(INSTALL_TEST_DIR)/InferenceEngineCAPITests --gtest_output=xml:$(INSTALL_TEST_DIR)/TEST-InferenceEngineCAPITests.xml
+      # TODO (vurusovs): revert skip of ie_core_*register_plugin*,
+      # tests should be fixed
+      $(INSTALL_TEST_DIR)/InferenceEngineCAPITests --gtest_filter=-*register_plugin* --gtest_output=xml:$(INSTALL_TEST_DIR)/TEST-InferenceEngineCAPITests.xml
```
ilyachur, line 359: Doesn't it work only in tests? Or doesn't it work at all on debian?

vurusovs:
Current changes doesn't break behavior of install package or local (developer) build since it's compatible with previous behavior. Only those who used `register_plugin(s)` in order to search plugin in libopenvino.so directory will be impacted due that way of library loading is prohibited now

.ci/azure/linux_debian.yml
vurusovs: We have other tests on `register_plugin()` to be sure it works on debian

.ci/azure/linux_debian.yml
vurusovs: Done - #101968

We prohibit only searching in libopenvino.so folder, other behavior is aligned with previous implementation

ilyachur (approved):

PR #15075, opened 2023-01-12
[GPU] Remove legacy fused ops

vladimir-paramuzov:
### Details:
 - Removed pass with legacy activation primitive fusion and added conversion during kernel selector params creation for single activation in fused ops
 -  Fix winograd kernel compilation

### DIFF
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_buffer_fusing.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_buffer_fusing.cpp
@@ -67,9 +67,7 @@ bool concat_noop_optimization::match(concatenation_node& node) {
         return false;
     if (node.is_dynamic())
         return false;
-    return node.get_dependencies().size() == 1 &&
-        !node.has_fused_primitives() &&
-        node.get_fused_activations_funcs().empty();
+    return node.get_dependencies().size() == 1 && !node.has_fused_primitives();
 }
 
 bool concat_noop_optimization::optimize(concatenation_node& node) {
@@ -84,7 +82,7 @@ bool concat_noop_optimization::optimize(concatenation_node& node) {
 bool concat_in_place_optimization::match(concatenation_node& node) {
     if (node.is_output())
         return false;
-    if (node.has_fused_primitives() || !node.get_fused_activations_funcs().empty())
+    if (node.has_fused_primitives())
         return false;
     if (node.is_dynamic())
         return false;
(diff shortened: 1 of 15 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_buffer_fusing.cpp
```diff
@@ -322,11 +320,11 @@ void prepare_buffer_fusing::run(program& p) {
         // The condition below check only output layout as cases like
         // (dyn_shape) -> reshape -> (static_shape) -> some_static_primitive
         // may have invalid set_arguments call as output memory of reshape won't be available until reshape primitive is executed
-        if (node->is_type<reshape>() && is_dynamic && is_planar && no_pad && !node->is_output() && node->get_fused_activations_funcs().empty()) {
+        if (node->is_type<reshape>() && is_dynamic && is_planar && no_pad && !node->is_output() && !node->has_fused_primitives()) {
```
yeonbok, line 323: This assumes that rehsape can have "fused operation" but there is no "GetSupportedFusedOps" in reshape. Does it make sense?

src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_buffer_fusing.cpp
```diff
@@ -322,11 +320,11 @@ void prepare_buffer_fusing::run(program& p) {
...
             return true;
         }
 
-        if (node->is_dynamic() || node->is_output() || (!node->get_fused_activations_funcs().empty())) {
+        if (node->is_dynamic() || node->is_output() || node->has_fused_primitives()) {
```
yeonbok, line 327: Does this behave same? Previously it was excluding only "fused activation" but now it excludes all fused cases.

vladimir-paramuzov: I think missing fused_primitives check before is a bug. This pass handles concat,crop and reshape primitives and if those have any type of fused operation - they can't be optimized out.

src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_buffer_fusing.cpp
vladimir-paramuzov: For reshape we allow fusion for single activation via fused_ops mechanism and it's supposed to be converted later to legacy desc for jitter. So missing `GetSupportedFusedOps` impl is not a big issue. Also, we can try to prohibit all fusions into reshape later, as those seems to be inefficient (optimized out reshape + optimized activation can be faster than single reshape kernel)

yeonbok: Is the idea to handle such cases later when detected by "GPU_DEBUG_LOG << activation_node.id() << " is fused by legacy conditions! Consider adding selected kernel with fused ops support" ?

yeonbok: After I added the second comment I saw your response. Resolving.

yeonbok (approved):

PR #15092, opened 2023-01-13
[CPU]FC shape infer with an CPU shapeInfer object

chenhu-wang:
### Details:
 - *cpu specific shape infer for FC*

### Tickets:
 - *100505*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
+++ b/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
@@ -93,16 +93,19 @@ public:
         const VectorDims& weightShape = input_shapes[1].get();
         size_t activationRank = activationShape.size();
         size_t channelRank = weightShape.size() - 1;
-        VectorDims outputShape;
-        // set batch dims
-        for (size_t i = 0; i < activationRank - channelRank; ++i) {
-            outputShape.push_back(activationShape[i]);
-        }
-        // append Co
-        outputShape.push_back(weightShape[0]);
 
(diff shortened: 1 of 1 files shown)


src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -83,6 +83,50 @@ bool FCKey::operator==(const FCKey &rhs) const {
...
+        // set batch dims
+        for (size_t i = 0; i < activationRank - channelRank; ++i) {
+            outputShape.push_back(activationShape[i]);
+        }
+        // append Co
+        outputShape.push_back(weightShape[0]);
+
+        while (outputShape.size() < out_rank) {
+            outputShape.insert(outputShape.begin(), 1);
+        }
```
maxnick, line 106:
```suggestion
        VectorDims outputShape(out_rank, 1);
         // set Co
        outputShape.back() = weightShape[0];
        // set batch dims
        size_t batchDims = activationRank - channelRank
        size_t stratIdx = out_rank - (activationDims + 1);
        for (size_t i = 0; i < activationDims ; ++i) {
            outputShape[stratIdx ] = activationShape[i];
        }
```
Or something like that. The idea is to allocate enough memory at once to avoid possible internal buffer reallocations. Also, operation insert at the beginning over a std::vector means copy of n elements.

chenhu-wang: done, thank you Maksim!

maxnick (approved):

PR #15100, opened 2023-01-13
[MO][TF FE] Switch MO to TF FE in default mode

rkazants:
**Details:** Switch Model Optimizer to use TF FE in the default mode. There are still some cases when MO will use the legacy FE when input model contains operations for TF1 `If` and `While` operations, for example.

**Ticket:** 99257


rkazants:
@gkrivor, fyi

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -301,16 +305,53 @@ def check_fallback(argv: argparse.Namespace):
...
+        "LoopCond", "Enter", "NextIteration", "Exit",  # corresponds to TF1 While operation
+        "Switch", "Merge",  # corresponds to TF1 If and TF1 While operations
+        "TensorListLength", "TensorListReserve", "TensorListFromTensor",
+        "TensorListSetItem", "TensorListStack",  # corresponds to TF2 While operations
+        "FFT", "FFT2D", "FFT3D", "IFFT", "IFFT2D", "IFFT3D",
+        "RFFT", "RFFT2D", "RFFT3D", "IRFFT", "IRFFT2D", "IRFFT3D",
+        "Complex", "ComplexAbs", "Real", "Imag",  # corresponds to operations with complex tensors
+        "FIFOQueueV2", "QueueDequeueUpToV2", "QueueDequeueManyV2",
+        "QueueDequeue", "QueueDequeueV2", "IteratorGetNext",
+        "LookupTableInsert", "LookupTableInsertV2"  # corresponds to automatic pruning
```
mvafin, line 340: nit: Comments for each groups of operations are hard to match with actual groups, it would be better to rearrange it so that comment describing the group comes first and list of operations comes after it.

mvafin (approved):

tools/mo/openvino/tools/mo/convert_impl.py
rkazants: Done

PR #15102, opened 2023-01-13
[ShapeInference] GridSample shape infer review

mitruska:
### Details:
 - Improve output batch dim by merging data and grid dimensions
 - Add setter for GridSample attributes

### Tickets:
 - 97234


src/plugins/intel_cpu/tests/unit/shape_inference_test/grid_sample_shape_inference_test.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <gtest/gtest.h>
+
+#include "utils.hpp"
+#include "openvino/opsets/opset9.hpp"
+#include "grid_sample_shape_inference.hpp"
+
+using namespace ov;
+using namespace ov::intel_cpu;
+
+TEST(StaticShapeInferenceTest, GridSample) {
```
praasz, line 14: Optional changes:

As these are new test consider use the helper test fixture `OpStaticShapeInferenceTest` and convert these test to `TEST_F` macro

This helper got input and output shapes member (output shapes can be set to one in setup test) and there is `make_op` method. 
It can improve maintenance of tests as operator under test is specified as template parameter, and it can easily replace by different version of operator.

praasz (approved):

jane-intel (approved):

PR #15104, opened 2023-01-13
[CPU] Use static post ops in the FC node

maxnick:
### Details:
Since the FC node always has a constant tenzor on its second input the OC size is always static, and we can exploit this fact and initialize the `dnnl::primitive_attr` object only once on the `createPrimitive` stage. This modification helps to save a lot of `prepareParams` time in the case of dynamic shapes.

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
+++ b/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
@@ -265,6 +265,13 @@ void FullyConnected::getSupportedDescriptors() {
     }
 }
 
+void FullyConnected::createPrimitive() {
+    Node::createPrimitive();
+    setPostOps(attr, outDims);
+    attr.set_scratchpad_mode(dnnl::scratchpad_mode::user);
+    appendPostOpArgs(attr, primArgs, postOpsArgs);
+}
+
 void FullyConnected::prepareParams() {
     auto srcMemPtr = getParentEdgesAtPort(0)[0]->getMemoryPtr();
     auto dstMemPtr = getChildEdgesAtPort(0)[0]->getMemoryPtr();
@@ -283,10 +290,6 @@ void FullyConnected::prepareParams() {
     if (selected_pd == nullptr)
         IE_THROW() << "Preferable primitive descriptor is not set for node " << getName() << ".";
 
-    AttrPtr attr = std::make_shared<dnnl::primitive_attr>();
-    setPostOps(*attr, dstMemPtr->getStaticDims());
-    (*attr).set_scratchpad_mode(dnnl::scratchpad_mode::user);
-
     DnnlMemoryDescPtr weightDesc = MemoryDescUtils::convertToDnnlMemoryDesc(weightDescIP);
     DnnlMemoryDescCPtr biasDesc = nullptr;
     if (biasMemPtr) {
@@ -301,7 +304,7 @@ void FullyConnected::prepareParams() {
                  weightDesc,
                  biasDesc,
                  outDesc,
-                 *attr,
+                 attr,
                  implementationTypeIP,
                  useConv1x1};
 
@@ -424,8 +427,6 @@ void FullyConnected::prepareParams() {
         if (withBiases) {
             primArgs[DNNL_ARG_BIAS] = biasMemPtr->GetPrimitive();
         }
-
-        appendPostOpArgs(*attr, primArgs, postOpsArgs);
 
         auto pd = execPtr->getPrimitiveDesc();
         auto scratchpadMem = getScratchPadMem(pd);
--- a/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
+++ b/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
@@ -266,9 +266,9 @@ void FullyConnected::getSupportedDescriptors() {
 }
 
 void FullyConnected::createPrimitive() {
-    Node::createPrimitive();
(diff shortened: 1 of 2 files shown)


maxnick:
@usstq, could you please review?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -265,6 +265,13 @@ void FullyConnected::getSupportedDescriptors() {
     }
 }
 
+void FullyConnected::createPrimitive() {
+    setPostOps(attr, outDims);
+    attr.set_scratchpad_mode(dnnl::scratchpad_mode::user);
+    Node::createPrimitive();
+    appendPostOpArgs(attr, primArgs, postOpsArgs);
```
usstq, line 272: Is there any reason why `Node::createPrimitive();` is not the last statement, or does `appendPostOpArgs()` depending on something happened inside `Node::createPrimitive();`?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -265,6 +265,13 @@ void FullyConnected::getSupportedDescriptors() {
     }
 }
 
+void FullyConnected::createPrimitive() {
+    setPostOps(attr, outDims);
```
usstq, line 269: `outDims` may be a dummy shape in dynamic model, is that a problem when a binary post Ops is being fused into FullyConnected ?

maybe it's not a problem for FullyConnected,  since OC is always static and so far we only use `outDims` in setPostOps when fusing per-OC post-Ops.

If so, why not take one more step, when the weights for convolution is also constant and the OC is static, we can also do this optimization which benefits more, what do you think?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
maxnick: Usually filling up the `primArgs` map is the last step of the parameter preparation, and, though is this particular case is does not have any matter, I wanted to preserve thins sequence of actions.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
maxnick: On the high level, yes, using `outDims` being a "dummy" shape is rather dangerous, since the `DnnlPostOpsComposer` requires the output shape to be equal to the output shape. But, on the FC operation level we know that the post operations may only depend on the channel size. So yes, this is a coupling, but staying on such a low level we can sacrifice the coupling for better performance. At least such coupling is bounded only on the node implementation level and does not impact anything outside it.
From the implementation behavior, following https://github.com/openvinotoolkit/openvino/blob/0c9abf43a91fbe2dc95974b6cb2651bd883d1b12/src/plugins/intel_cpu/src/dnnl_postops_composer.cpp#L28-L29 we see that only `OC` dimension is used for the binary post ops, all others are always 1. So that should work. And in opposite, if it didn't work, that would mean that something was wrong with the `DnnlPostOpsComposer`, since it requires dependency on the other dims when it is meaningless.
As for `Convolution`. We already have the very same optimization, but implemented in slightly different fashion, please take a look at https://github.com/openvinotoolkit/openvino/blob/0c9abf43a91fbe2dc95974b6cb2651bd883d1b12/src/plugins/intel_cpu/src/nodes/conv.cpp#L1309-L1318 the only difference is that, due to some comprehensive fusing handling, we still have to call the attributes initialization under some conditions.

usstq: Yes, indeed, from oneDNN API binary postOps can support any shape with broadcast semantic that may depending on exact dimension value other than OC, but we didn't observe such use case for now, so in theory we can re-build attr only when OC is changed, this may also apply to matmul.  for example, in matmul, the batch dimension may be dynamic but the channel dimension is usually static. this method can be generalized, maybe next time.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
usstq: personally I prefer to use same mechanism for similar task, for example we can follow what convolution node does inside `prepareParam` instead of introducing `createPrimitive()` overload, this overload becomes tricky due to the timing, and you must be cautious on what shapes are fake/dummy and what information is truly determined.

usstq (approved):
LGTM, maybe we can generalize this idea to other nodes in the future

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
maxnick: The thing is, in Convolution node we still have a situation when we rebuild `attr` every `preparePrams()` call due to `sum` fusing. But here we always can do it only once. That means this is implementation dependent feature, and given the fact that the node implementation is the most low level, probably we can not do it in a generic fashion. I mean if in the Convolution node we could initialize `attr` only once to in all cases, we probably would do it in the same way as in FC.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
maxnick: I'm just wondering, maybe it does makes sense to reflect such a property in the `DnnlPostOpsComposer` constructor? I mean add a constructor which takes OC dimension only, so that to ensure that all other dims are not used?

maxnick:
@dmitry-gorokhov , the PR is ready to be merged.

PR #15112, opened 2023-01-13
[GPU] Move tuning cache loading to kernel selector

vladimir-paramuzov:
### Details:
 - Move tuning cache file loading from program to kernel_selector::AutoTuner to have less kernel selector deps in common graph module
 - Removed TuningCache internal property and related processing (including kernels_runner) as we haven't exposed tuning capabilities to users and don't plan to do that
 - Simplified `GetAutoTuneBestKernel` impl and avoided double call for `GetAllImplementations` by adding overloaded `GetNaiveBestKernel` version
 - Removed virtual keyword from `GetAutoTuneBestKernel` and `GetNaiveBestKernel` as those are never overridden.


src/plugins/intel_gpu/src/kernel_selector/auto_tuner.h
```diff
@@ -47,6 +47,8 @@ class TuningCache {
 
     bool NeedsSave() const { return needsSave; }
 
+    static TuningCache* get();
```
sshlyapn, line 50: Why do you left TuningCache::{StoreKernel|RemoveKernel|Save|RemoveKernel_v1|RemoveKernel_v2} functions? I think they will never be used and may be safely removed

sshlyapn (approved):

src/plugins/intel_gpu/src/kernel_selector/auto_tuner.h
vladimir-paramuzov: Good catch. Will remove them in a separate PR.

PR #15113, opened 2023-01-13
update Linux CPU map parser and add unit test

wangleis:
### Details:
 - *Update Linux CPU map parser for validation*
 - *Add unit test to validate Linux CPU map parser based on simulation dataset*

### Tickets:
 - *100600*


src/inference/tests/unit/cpu_map_parser.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+#include <gtest/gtest.h>
+
+#include "ie_system_conf.h"
+
+using namespace InferenceEngine;
+
+#ifdef __linux__
+TEST(CPUMapParserTest, LinuxMapping) {
+    struct linux_system_data_str {
```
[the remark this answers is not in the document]
wangleis: Update to `TEST_P`. 

Let's discuss how to expose device name in meeting.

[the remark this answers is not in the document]
ilya-lavrenov: Maybe we suppose we can export not device names, like Ice Lake, etc, but description of device, e.g. `_2sockets_4cores_hyperthreading`, etc. So, we don't expose internal or external name, but the description is still useful.
```c++
linux_system_data_str  _2sockets_4cores_hyperthreading = {...};
linux_system_data_str  _1socket_4cores_hyperthreading = {...};
```
And then pass such values to `TEST_P`

ilya-lavrenov (approved):

src/inference/src/os/lin/lin_system_conf.cpp
```diff
@@ -188,9 +91,155 @@ struct CPU {
                 _cores = _processors;
             }
         }
+        std::vector<std::vector<std::string>>().swap(system_info_table);
     }
 };
 static CPU cpu;
+
+void parse_processor_info_linux(const int _processors,
```
dmitry-gorokhov, line 99: Not necessary for this PR, but nice to have in future. It is better to explain what structure `system_info_table` table has to simplify future maintenance of `parse_processor_info` both for linux and windows.

src/inference/tests/unit/cpu_map_parser.cpp
```diff
@@ -0,0 +1,571 @@
...
+struct LinuxCpuMapTestCase {
+    int _processors;
+    int _sockets;
+    int _cores;
+    std::vector<std::vector<int>> _proc_type_table;
+    std::vector<std::vector<int>> _cpu_mapping_table;
+    std::vector<std::vector<std::string>> system_info_table;
+};
+
+class LinuxCpuMapParserTests : public CommonTestUtils::TestsCommon,
```
dmitry-gorokhov, line 27: Will you create same tests for Windows?

dmitry-gorokhov (approved):

src/inference/tests/unit/cpu_map_parser.cpp
wangleis: Yes, will create tests for windows as well.

src/inference/src/os/lin/lin_system_conf.cpp
wangleis: So far looks like we cannot use unified `system_info_table` for both Linux and Windows. 

We would implement separated parser and test cases for Linux and Windows. Then let's review and check if we can merge them.

PR #15117, opened 2023-01-13
[CPU] Fix StridedSlice fix for inputs on a constant path

maxnick:
### Details:
The problem is that the StridedSlice implementation considered constant inputs only if the parent node is Const ngraph op. However, there is so called constant path, when the constant value is propagated through several operations of different types. So the fix is aimed to correctly process such cases.


maxnick:
@dmitry-gorokhov, could you please review? This fix is related to your latest changes.

maxnick:
Test TBD

maxnick:
Close as the bug fix should be on the ngraph side.

PR #15123, opened 2023-01-16
optimize inference perf of model with asymmetric quantization

ermubuzhiming:
Signed-off-by: Lin Ying <yinglin@intel.com>

### Details:
 - *item1* found the perf of model with asymmetric quantization similar with that of FP32 model
 - *item2* the perf of model with asymmetric quantization is between perf of model with symmetric quantization and perf of FP32 model.


akladiev:

akladiev:

PR #15128, opened 2023-01-16
Review ReverseSequence class for shape inference aspects

praasz:
### Details:
- partial shapes and labels propagation
- template implementation of shape infer
- refactor `shape_infer` to use it when op created with default `ctor`

### Tickets:
 - 97250
 - 97288
 - 97337
 - 97386
 - 97435
 - 97489
 - 97538
 - 97587


src/core/shape_inference/include/reverse_sequence_shape_inference.hpp
```diff
@@ -26,25 +28,33 @@ void shape_infer(const ReverseSequence* op, const std::vector<T>& input_shapes,
                           seq_lengths_rank.compatible(1),
                           "Sequence lengths rank must be equal to 1. Got: ",
                           seq_lengths_pshape);
-    auto& output_pshape = output_shapes[0];
-
-    output_pshape = data_pshape;
+    auto output_pshape = data_pshape;
     if (data_rank.is_static() && seq_lengths_rank.is_static()) {
+        const auto normalized_batch_axis = ov::normalize_axis(op, op->m_batch_axis, data_rank);
```
jane-intel, line 33: let's call `get_batch_axis` here and get rid of this friend function in the ReverseSequence class. It just makes it more complex with no obvious reason why.

praasz: Good catch, `get_origin_batch_axis()` method as it give `m_batch_axis`. The `get_batch_axis ` can not work as it use input 0 for rank value which can be not set when operator created with default ctor.

src/core/src/op/reverse_sequence.cpp
```diff
@@ -59,25 +55,14 @@ shared_ptr<Node> op::ReverseSequence::clone_with_new_inputs(const OutputVector&
...
-    const auto& data_rank = data_pshape.rank();
-    NODE_VALIDATION_CHECK(op,
-                          data_rank.is_dynamic() || data_rank.get_length() >= 2,
-                          "Data input rank should be equal or greater than 2. Got: ",
-                          data_pshape);
-    op->m_normalized_batch_axis = ov::normalize_axis(op, op->m_batch_axis, data_rank);
-    op->m_normalized_seq_axis = ov::normalize_axis(op, op->m_seq_axis, data_rank);
+void op::ReverseSequence::set_sequence_axis(int64_t sequence_axis) {
+    m_seq_axis = sequence_axis;
+    m_normalized_seq_axis = ov::normalize_axis(this, m_seq_axis, get_input_partial_shape(0).rank());
```
mitruska, line 67: `get_input_partial_shape(0)` makes impossible to call `set_sequence_axis` on default constructed op without inputs, 
can it be changed to perform normalization `if inputs.size() > 0`?
Also such normalization will throw for dynamic rank, but it wasn't supported anyway.

src/core/src/op/reverse_sequence.cpp
```diff
@@ -59,25 +55,14 @@ shared_ptr<Node> op::ReverseSequence::clone_with_new_inputs(const OutputVector&
...
 }
 
-void op::ReverseSequence::set_sequence_axis(int64_t sequence_axis) {
-    m_seq_axis = sequence_axis;
-    const auto& data_pshape = get_input_partial_shape(0);
-    const auto& data_rank = data_pshape.rank();
-    m_normalized_seq_axis = ov::normalize_axis(this, m_seq_axis, data_rank);
+size_t op::ReverseSequence::get_batch_axis() const {
+    const auto& data_rank = get_input_partial_shape(0).rank();
+    return static_cast<size_t>(ov::normalize_axis(this, m_batch_axis, data_rank));
```
mitruska, line 62: Is there a specific reason for moving axis normalization from setter to getter?

And similar comment to `set_sequence_axis`:
`get_input_partial_shape(0)` makes impossible to call `get_batch_axis` on default constructed op without inputs, 
can it be changed to perform normalization `if inputs.size() > 0`?

src/core/tests/type_prop/reverse_sequence.cpp
```diff
@@ -65,255 +67,274 @@ TEST(type_prop, reverse_sequence_invalid_seq_lengths_et) {
...
+    EXPECT_EQ(op->get_sequence_axis(), 0);
+    EXPECT_EQ(op->get_input_size(), 2);
+    EXPECT_EQ(op->get_output_size(), 1);
+    EXPECT_EQ(op->get_output_element_type(0), element::f32);
+    EXPECT_EQ(op->get_output_partial_shape(0), (PartialShape{4, 3, 2}));
+}
+
+TEST_F(TypePropReverseSequenceV0Test, default_ctor_no_arguments) {
+    auto op = make_op();
+    op->set_batch_axis(1);
```
mitruska, line 301: `set_sequence_axis` should be tested here as well

src/core/src/op/reverse_sequence.cpp
praasz: Yes as `m_batch_axis` is required for shape inference and if it is possible to set without input this is why is moved to getter so you can only get when you got input set. 
Get this value without normalization when there is no inputs will give not correct value so exception is OK for me.
For not normalized value there is `get_origin_batch_axis`.

src/core/src/op/reverse_sequence.cpp
praasz: using if there is no inputs is pointless and this value has no meaning for shape inference.

src/core/tests/type_prop/reverse_sequence.cpp
praasz: Is skipped as it has no meaning for shape inference.

jane-intel (approved):

mitruska (approved):

PR #15130, opened 2023-01-16
[TF FE] Support Group Convolutions

rkazants:
**Details:** It turned out that `Conv2D` and `Conv3D` (despite this case is missed in the TensorFlow specification) represent grouped convolutional operations. So we need to support this case raised in some e2e model.

**Ticket:** 98461


src/frontends/tensorflow/src/utils.cpp
```diff
@@ -182,8 +184,27 @@ ov::OutputVector ov::frontend::tensorflow::translate_convolution_op(const ov::fr
...
+    auto input_cin = make_shared<Slice>(input_shape, one_const, two_const, one_const);
+    auto filter_cin = make_shared<Slice>(filter_shape, one_const, two_const, one_const);
+    auto num_groups = make_shared<Divide>(input_cin, filter_cin);
+
+    // reshape the filter based on the number of groups information
+    auto int_max_const = make_shared<Constant>(element::i32, Shape{1}, std::numeric_limits<int>::max());
+    auto filter_cout = make_shared<Slice>(filter_shape, zero_const, one_const, one_const);
+    auto filter_new_cout = make_shared<Divide>(filter_cout, num_groups);
+    auto shape_cin_xy = make_shared<Slice>(filter_shape, one_const, int_max_const, one_const);
+    auto filter_new_shape = make_shared<Concat>(OutputVector{num_groups, filter_new_cout, shape_cin_xy}, 0);
```
itikhono, line 203: can we use -1 instead of shape_cin_xy?

PR #15143, opened 2023-01-16
[GPU] Use onednn fc/gemm in dGPU.

hyunback:
Huge performance improvement FC and gemm in oneDNN3.0
Update to use onednn FC and gemm in OV and remove W.A code force using cldnn in FC, gemm
Fix the issues used from onednn fc, gemm(eltwise post op)
 - FC b16 ACC issue: cldnn updates yxfb format in b16 for opt kernel, but FC onednn should bfyx
 - FC ACC issue: 3 input dims with eltwise post-po.
 - gemm ACC issue: multiple batch with eltwise post-po.

Signed-off-by: hyunback <hyunback.kim@intel.com>

### Tickets:
 - *101284*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/program_node.cpp
+++ b/src/plugins/intel_gpu/src/graph/program_node.cpp
@@ -9,6 +9,7 @@
 #include "intel_gpu/runtime/debug_configuration.hpp"
 #ifdef ENABLE_ONEDNN_FOR_GPU
 #include "convolution_inst.h"
+#include "fully_connected_inst.h"
 #include "deconvolution_inst.h"
 #include "quantize_inst.h"
 #include "reorder_inst.h"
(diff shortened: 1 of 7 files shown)


src/plugins/intel_gpu/src/graph/program_node.cpp
```diff
@@ -948,6 +950,25 @@ void program_node::init_onednn_primitive_attributes() {
             auto dep_idx = desc.dep_start_idx;
             auto in = get_dependency(dep_idx).get_output_layout();
 
+            auto get_input_desc = [&](cldnn::layout in) {
+                if (is_type<fully_connected>()) {
+                    const kernel_impl_params& impl_params = *get_kernel_impl_params();
+                    auto prim = impl_params.typed_desc<fully_connected>();
+                    if (prim->input_size == 3) {
```
isanghao, line 957: why do we have special treatment for 3? As this is called for post operations, I think we can just remove this condition.

isanghao (changes requested):

src/plugins/intel_gpu/src/graph/program_node.cpp
hyunback: oneDNN FC only support 2 dims output, so current OV implementation(fully_connected_onednn.cpp:128) changes shape bfyx to bf using combine_bf_with_first_spatial_dim() in case 3 input dims. Because full tensor eltwise post operations calculate on the changed FC result, we should check FC's input_size condition.
Previous master code also check the condition (add_onednn_optimization_attributes.cpp) and now it move to here, because we should not change OV shape itself.
And below is accuracy_checker result from bert-small-uncased-whole-word-masking-squad-int8-0002(PR vs PR w/o check-condition here) and shows the huge different result.
* PR
F1: 53.08% [FAILED:  abs error = 38.32 | relative error = 0.4192]
EM: 33.54% [FAILED:  abs error = 50.86 | relative error = 0.6026]
* PR w/o check-condition here
F1: 4.44% [FAILED:  abs error = 86.96 | relative error = 0.9514]
EM: 0.63% [FAILED:  abs error = 83.77 | relative error = 0.9925]

isanghao:
Performance was good ../log/2023-02-01--11-46.report

isanghao (approved):

PR #15145, opened 2023-01-16
[TF FE] Fix LRN operation and test it

rkazants:
**Details:** It allows to fix inference results mismatch for some models

**Tickets:** 95068, 92866

### DIFF
--- a/src/frontends/tensorflow/src/op/lrn.cpp
+++ b/src/frontends/tensorflow/src/op/lrn.cpp
@@ -4,6 +4,7 @@
 
 #include "op_table.hpp"
 #include "openvino/opsets/opset8.hpp"
+#include "utils.hpp"
 
 using namespace std;
 using namespace ov::opset8;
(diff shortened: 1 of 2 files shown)


mvafin (approved):

src/frontends/tensorflow/src/op/lrn.cpp
```diff
@@ -14,23 +15,30 @@ namespace tensorflow {
...
+    size_t attr_size = static_cast<size_t>(depth_radius) * 2 + 1;
+    alpha *= static_cast<double>(attr_size);
+
+    // currently, plugins fallback to the template to execute this operation
+    // the current implementation supports only the normalization across channel (axes={1})
+    // and spatial dimensions (axes={2,3})
+    // so we have to transpose inputs and outputs due to this limitation
+    auto axis = make_shared<Constant>(element::i32, Shape{1}, 1);
+    ov::frontend::tensorflow::convert_nhwc_to_nchw(true, input);
+    auto lrn = make_shared<LRN>(input, axis, alpha, beta, bias, attr_size)->output(0);
```
itikhono, line 38: nitpick: you can remove ->output(0) call and also remove .get_node_shared_ptr() from line40

itikhono (approved):

src/frontends/tensorflow/src/op/lrn.cpp
rkazants: unfortunately, this is not possible because `convert_nchw_to_nhwc` accepts `ov::Output` by reference so I have to work with `ov::Output` instead of `ov::Node`.

itikhono: You actually can. This ctor will be used in this case: https://github.com/openvinotoolkit/openvino/blob/89e2c0e2fa5c13f693a8240310395c557fef1748/src/core/include/openvino/core/node_output.hpp#L46
It works only for Nodes with a single output

rkazants: image

rkazants: `convert_nchw_to_nhwc` is complaining.

PR #15161, opened 2023-01-17
[GPU] Fix segfault with exec_graph_path option

hyunback:
The issue is reproduced in models included implicit concat
Opt-out primtive_onednn_imple doesn't share ExecutionConfig issue

Signed-off-by: hyunback <hyunback.kim@intel.com>

### Tickets:
 - *101368*


hyunback:
@vladimir-paramuzov Could you review it? This issue is reproduced since https://github.com/openvinotoolkit/openvino/pull/15061

vladimir-paramuzov (approved):

PR #15163, opened 2023-01-17
optimize TensorIterator DynamicBuffer by preallocating a large chunk …

ceciliapeng2011:
…of intermediate buffer.

### Details:
 - *Two points from from memory allocation standpoint:

In case of TI number of iterations is determined by size of axis we are slicing. So if TI body on each iteration will produce the same shape on output we can easily preallocate enough memory before the TI computation, The same for Loop with trip count input - we can just read value from this input, make shape_inference for the body and this definitely determines required amount of memory.
More complicated story is when we don't now exact number of iterations before Loop inference (e.g. #iterations is determined by ExecutionCondition input). In that case do the following: lets have an output buffer where we put the Loop output. Once the buffer doesn't have enough space we reallocate it on new size.

Such approach will allow us significantly decrease number of reallocation and not to allocate to much memory at the same time.*
 - *...*

### Tickets:
 - *87788*
 - *90171*


src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
```diff
@@ -303,24 +338,49 @@ void DynamicBuffer::move_data() {
...
-                DnnlExtensionUtils::convertToVectorDims(mem_holder_buffer->get_desc().dims()));
-        redefineToMemories(to, desc);
+                DnnlExtensionUtils::convertToVectorDims(dims));
 
-        copy(get_ptr(*mem_holder_buffer.get()), reinterpret_cast<uint8_t*>(to.front()->GetPtr()), 0, 0, 1, to.front()->GetSize());
+        if (dims[axis] == mem_holder_dims[axis]) {
+            // straightly share mem_holder_buffer to outside graph
+            redefineToMemories(to, desc, mem_holder_buffer->get_data_handle());
+        } else {
+            redefineToMemories(to, desc);  // TODO: redefine only if changed since last inference
```
maxnick, line 369: As far as I understand it, `redefineToMemories` itself internally checks whether the shape has changed

src/plugins/intel_cpu/src/nodes/tensoriterator.h
```diff
@@ -74,6 +74,14 @@ class DynamicBuffer {
     void init(const dnnl::engine& eng);
 
     /* methods for resize and refill buffer */
+    inline bool check_buffer() {
```
maxnick, line 77: Actually the `inline` declaration is not necessary at all.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
```diff
@@ -505,14 +562,18 @@ void TensorIterator::prepareParams() {
     before_mappers.clear();
     back_mappers.clear();
 
+    buffers.clear();
```
maxnick, line 565: I'm wondering, do we really need to reinstantiate the buffers? M.b. it does make sense to just reset the `max_iterations` value? These memory reallocations for all the auxiliary data are not free of charge.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
```diff
@@ -270,19 +299,9 @@ std::shared_ptr<dnnl::memory> DynamicBuffer::create_buffer(const dnnl::engine& e
     const auto old_desc = mem_holder_buffer->get_desc();
     auto dims = old_desc.dims();
 
-    if (from->getStaticDims()[axis] != abs_stride)
-        IE_THROW() << "TensorIterator (Loop) has incorrect output shape[axis] after iteration for concatenation. " << abs_stride <<
-        " is expected, but actual: " << from->getStaticDims()[axis];
-
-    dims[axis] += abs_stride;
+    dims[axis] += abs_stride * 1;  // TODO: the increasing amount
```
maxnick, line 302: Do you plan to implement this within this PR?

src/plugins/intel_cpu/src/nodes/tensoriterator.h
```diff
@@ -85,7 +93,9 @@ class DynamicBuffer {
     size_t count = 1lu;
     size_t elem_size = 0lu;
     ptrdiff_t chunk_offset_in_byte = 0;
-    ptrdiff_t buffer_offset_in_byte = 0;
+    size_t mem_holder_unit = 0;   // the amount of bytes copied per each count per each execution (iteration)
+    size_t num_iters = 0lu;      // number of executions happened
```
maxnick, line 97: `execute` method take `iter` as a parameter and it means the number of the tensor iterator iteration. If this number of interactions is different in the sense, so m.b. it does make sense to rename it somehow, so just to avoid confusion.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
```diff
@@ -303,24 +338,49 @@ void DynamicBuffer::move_data() {
...
+        dims[axis] = abs_stride * num_iters;
         const auto desc = node->getBaseMemDescAtOutputPort(map_rule.from)->cloneWithNewDims(
-                DnnlExtensionUtils::convertToVectorDims(mem_holder_buffer->get_desc().dims()));
-        redefineToMemories(to, desc);
+                DnnlExtensionUtils::convertToVectorDims(dims));
 
-        copy(get_ptr(*mem_holder_buffer.get()), reinterpret_cast<uint8_t*>(to.front()->GetPtr()), 0, 0, 1, to.front()->GetSize());
+        if (dims[axis] == mem_holder_dims[axis]) {
+            // straightly share mem_holder_buffer to outside graph
+            redefineToMemories(to, desc, mem_holder_buffer->get_data_handle());
```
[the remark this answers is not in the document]
ceciliapeng2011: > The lifetime of the `to` memory objects is longer than the lifetime of the `mem_holder_buffer`, which may lead to the situation when the `data_handle` pointer is expired but the memory objects steel reference it. 
My idea is "to" memory objects will take over the ownership of  data_handle of "mem_holder_buffer" since then. Will it expire if it's taken?

> So Does it really worth the cost of one extra copy? What do you think?
Do you mean keep the copy and won't share the data_handle outside?

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
```diff
@@ -54,12 +54,15 @@ static NodeConfig make_plain_config(const std::shared_ptr<ov::Node>& op) {
...
 
-static void redefineToMemories(const std::vector<MemoryPtr>& to_mems, MemoryDescPtr new_desc) {
+static void redefineToMemories(const std::vector<MemoryPtr>& to_mems, MemoryDescPtr new_desc, void* data = nullptr) {
     const auto &currDesc = to_mems.front()->getDesc();
     if (currDesc.getShape().isDynamic() || currDesc.getShape().getStaticDims() != new_desc->getShape().getStaticDims()) {
         // TODO : check the entire dstMemPtrs usage considering the proper memory sharing
         for (size_t j = 0; j < to_mems.size(); j++) {
-            to_mems[j]->redefineDesc(new_desc);
+            if (nullptr != data)
+                to_mems[j]->Create(new_desc, data, false);
```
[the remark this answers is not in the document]
ceciliapeng2011: The motivation of this code change is to reset internal memory buffer with a new data buffer. M.b. a better way is to overload redefineDesc(desc). How do you think?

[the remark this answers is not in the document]
maxnick: No the problem is that `Creat` call completely resets the memory object, and, for example if it share the memory manager with some other objects, this connection will be lost. Also, which is more important, from the design perspective, the memory object, being a black box, may have any other important internal state, which may be rest by the `Creat` call. I think this is not what we want here.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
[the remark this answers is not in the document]
maxnick: AFAIK, `mem_holder_buffer->get_data_handle()` call does not transfer ownership, it only returns a pointer to the internal data blob. Following line 277 `mem_holder_buffer.reset(new memory(new_buffer_desc, eng));`, mem_holder_buffer maintains the lifespan of the internal memory block. Which means when mem_holder_buffer object is destroyed, the memory pointer returned by the `mem_holder_buffer->get_data_handle()` expires.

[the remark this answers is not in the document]
ceciliapeng2011: This memory sharing contributes 2~3% performance improvement with Dien model. 
Basically it is a data buffer. We just need to pass it outside and let "to" memory object share it directly.
A decoupling abstraction layer probably is asked here.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
ceciliapeng2011: yes, the comment is a bit nonsense. I will remove it and other commented code.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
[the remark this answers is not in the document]
ceciliapeng2011: > No, the problem is that `Creat` call completely resets the memory object, and, for example if it share the memory manager with some other objects, this connection will be lost. Also, which is more important, from the design perspective, the memory object, being a black box, may have any other important internal state, which may be rest by the `Creat` call. I think this is not what we want here. BTW, if we want to reset data pointer, we use `setDataHandle` call. But this means that we switch the memory object to another externally controlled memory blob. Which in case of memory sharing breaks the connection with common mem manager.


Hence got a question here - "redefineDesc" basically is "Create". How could the original code work as it uses this method to reset "to" memory object?

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
```diff
@@ -225,41 +228,67 @@ class staticValueCheck : public PortChecker {
...
 
 DynamicBuffer::DynamicBuffer(const MemoryPtr &from_, const std::vector<MemoryPtr> &to_,
-                             const PortMap &map_rule_) : from(from_), to(to_), map_rule(map_rule_) {
+                             const PortMap &map_rule_, const int max_iter_count_) : from(from_), to(to_), map_rule(map_rule_), max_iter_count(max_iter_count_) {
     elem_size = DnnlExtensionUtils::sizeOfDataType(from->GetDataType());
 }
 
+DynamicBuffer::~DynamicBuffer() {
+    mem_holder_buffer.reset();
+}
```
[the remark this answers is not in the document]
ceciliapeng2011: Hmm, it is a legacy of debugging. Will remove it.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
```diff
@@ -225,41 +228,67 @@ class staticValueCheck : public PortChecker {
...
+        auto src_mem = from->GetPrimitive();
+        auto src_desc = src_mem.get_desc();
+        auto dims = src_desc.dims();
+
+        count = std::accumulate(dims.begin(), dims.begin() + map_rule.axis, size_t(1), std::multiplies<size_t>());
+        len = std::accumulate(dims.begin() + map_rule.axis + 1, dims.end(), elem_size, std::multiplies<size_t>());
+        mem_holder_unit = abs_stride * len;
+
+        // preallocate a large chunk of memory to hold intermediate concated outputs of all iterations.
+        auto estimated_iters = max_iter_count != -1 ? max_iter_count : 1;  // TODO: the initial amount
```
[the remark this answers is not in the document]
ceciliapeng2011: When max_iter_count == -1 in which case we have no idea of the upper bound of memory requirement, how to preallocate the extra memory? This is what the comment is.

src/plugins/intel_cpu/src/nodes/tensoriterator.h
ceciliapeng2011: changed.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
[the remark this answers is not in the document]
ceciliapeng2011: defer it as it contributes a little bit but asks much effort.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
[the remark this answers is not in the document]
ceciliapeng2011: defer it as it contributes a little bit but asks much effort.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
ceciliapeng2011: updated.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
ceciliapeng2011: updated. with growth_factor=2.0

src/plugins/intel_cpu/src/nodes/tensoriterator.h
```diff
@@ -65,33 +65,41 @@ class PortChecker {
...
-    dnnl::memory create_buffer(const dnnl::engine& eng);
-    void move_buffer(dnnl::memory new_buffer);
+    bool check_buffer();
+    MemoryPtr create_buffer(const dnnl::engine& eng);
+    void move_buffer(MemoryPtr& new_buffer);
     void move_data();
 
     static void copy(const uint8_t* src, uint8_t* dst, const size_t src_stride, const size_t dst_stride, const size_t count, const size_t len);
-    static uint8_t* get_ptr(dnnl::memory& prim);
+    static uint8_t* get_ptr(MemoryPtr& prim);
```
maxnick, line 84: Either `const MemoryPtr& prim` or `MemoryPtr prim` depending on the ownership requirements.

src/plugins/intel_cpu/src/nodes/tensoriterator.h
```diff
@@ -65,33 +65,41 @@ class PortChecker {
...
+
 private:
     void init(const dnnl::engine& eng);
 
     /* methods for resize and refill buffer */
-    dnnl::memory create_buffer(const dnnl::engine& eng);
-    void move_buffer(dnnl::memory new_buffer);
+    bool check_buffer();
+    MemoryPtr create_buffer(const dnnl::engine& eng);
+    void move_buffer(MemoryPtr& new_buffer);
```
maxnick, line 80: Either `const MemoryPtr& prim` or `MemoryPtr prim` depending on the ownership requirements.

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
```diff
@@ -231,97 +231,148 @@ DynamicBuffer::DynamicBuffer(const MemoryPtr &from_, const std::vector<MemoryPtr
...
+
+        // in case of no idea of memory upper boundary
+        return (num_execs == 0) ? 1 : 2 * num_execs; // growth factor 2
+    };
+    const auto estimated_iters = estimate_iters();
+    const Shape _shape = Shape({count, static_cast<size_t>(abs_stride * estimated_iters), len/elem_size});
+    auto _descCreator = BlockedDescCreator::getCommonCreators().at(LayoutType::ncsp);
+    auto new_buffer_desc = _descCreator->createSharedDesc(from->getDesc().getPrecision(), _shape);
+
+    auto _ptr = MemoryPtr(new Memory(eng));
```
maxnick, line 301:
```suggestion
    auto _ptr = std::make_shared<Memory>(eng);
```

src/plugins/intel_cpu/src/nodes/tensoriterator.cpp
```diff
@@ -231,97 +231,148 @@ DynamicBuffer::DynamicBuffer(const MemoryPtr &from_, const std::vector<MemoryPtr
...
-void DynamicBuffer::move_buffer(dnnl::memory new_buffer) {
-    const auto axis = map_rule.axis;
-    const auto src_stride = mem_holder_buffer.get_desc().dims()[axis] * len;
-    const auto dst_stride = new_buffer.get_desc().dims()[axis] * len;
+void DynamicBuffer::move_buffer(MemoryPtr& new_buffer) {
+    const auto stride = map_rule.stride;
+
+    // copy data from old buffer to new buffer
+    const auto src_stride = chunk_stride_in_byte;
+    const auto dst_stride = new_buffer->getDesc().as<BlockedMemoryDesc>()->getBlockDims()[1] * len;
```
maxnick, line 311:
```suggestion
    const auto dst_stride = new_buffer->getStaticDims()[1] * len;
```
Since we use plain format.

src/plugins/intel_cpu/src/nodes/tensoriterator.h
ceciliapeng2011: updated.

src/plugins/intel_cpu/src/nodes/tensoriterator.h
ceciliapeng2011: updated.

maxnick (approved):

PR #15169, opened 2023-01-17
[GPU] Optimize realloc for dynamic shape 

yeonbok:
### Details:
 - Enable pre-alloc for bounded dynamic shape 
 - Reuse internal buffer for smaller memory

### Tickets:
 - 93614

### DIFF
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -285,6 +285,10 @@ void primitive_inst::realloc_if_needed() {
                     max_intermediates_memory_sizes[i] = std::max(max_intermediates_memory_sizes[i], _intermediates_memory[i]->size());
                }
             }
+        } else {
+            for (size_t i = 0; i < _intermediates_memory.size(); ++i) {
+                _intermediates_memory[i] = _network.get_engine().reinterpret_buffer(*_intermediates_memory[i], ibuf_layouts[i]);
+            }
         }
     }
 }
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -613,7 +613,7 @@ primitive_inst::primitive_inst(network& network, program_node const& node, bool 
     }
 
     if (_outputs[0])
-        max_output_layout_size = _outputs[0]->get_layout().get_max_tensor().count();
+        max_output_layout_size = _outputs[0]->get_layout().get_tensor().count();
 }
 
 void primitive_inst::allocate_internal_buffers(void) {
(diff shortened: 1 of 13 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -261,7 +261,36 @@ void primitive_inst::realloc_if_needed() {
...
     // intermediate memory allocation is required for primitives consisting of multiple kernels in dynamic case
-    allocate_internal_buffers();
+    {
+        const auto& ibuf_layouts = _impl->get_internal_buffer_layouts();
+        if (ibuf_layouts.empty())
+            return;
+        bool can_reuse_internal_buffer = (max_intermediates_memory_sizes.size() == ibuf_layouts.size());
+        if (can_reuse_internal_buffer) {
+            for (size_t i = 0; i < max_intermediates_memory_sizes.size(); ++i) {
+                if (ibuf_layouts[i].bytes_count() > max_intermediates_memory_sizes[i]) {
```
sshlyapn, line 271: I'm not sure, but maybe there are some cases where intermediates_memory can be reused partially? And then, it's better to have decision about memory reuse independently for each buffer

src/plugins/intel_gpu/src/runtime/layout.cpp
```diff
@@ -337,6 +337,32 @@ tensor layout::get_tensor() const {
...
+    }
+    std::vector<tensor::value_type> dims(shape.begin(), shape.end());
+
+    auto rank = std::max(format.dimension(), dims.size());
+    auto default_fmt = format::get_default_format(rank, format::is_weights_format(format), format::is_grouped(format));
+    if (default_fmt.dimension() > dims.size()) {
+        dims.insert(dims.end(), default_fmt.dimension() - dims.size(), 1);
+    }
+
+    while (dims.size() > default_fmt.dimension()) {
```
sshlyapn, line 358: Is it possible? As I understand, it occurs only if rank will be bigger than 6, and then maybe it's better to throw an exception instead of implicit reduce dimensions to bfyx format (which will be returned for rank size => 7)

sshlyapn (approved):
In general looks good to me, just left some minor comments

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
yeonbok: What do you mean? It is deciding reuse for each ibuf [i]. Do you mean something else?

src/plugins/intel_gpu/src/runtime/layout.cpp
yeonbok: This part was just copied from get_tensor(). However now I think we dont need get_max_tensor separately for this PR's requirement . So I merged it to original get_max_tensor. Could you review again?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
sshlyapn: I mean situation when only single buffer requires memory reallocation, but this will cause memory reallocation for all internal inst's buffers, not only for one

yeonbok: I got your point. Will update it.

yeonbok: Updated, please review again: https://github.com/openvinotoolkit/openvino/pull/15169/commits/5206fe37e88dd3dfb7f4fcf1eafd7891f8326736

sshlyapn: Looks good, thank you!

PR #15176, opened 2023-01-18
[GPU] Update ScatterNDUpdate Op to use ngraph shape infer

kelvinchoi-intel:
### Details:
 - *Update ScatterNDUpdate Op to use ngraph shape infer*

### Tickets:
 - *95092*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
+++ b/src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
@@ -28,6 +28,7 @@
 #include "depth_to_space_inst.h"
 #include "region_yolo_inst.h"
 #include "prior_box_inst.h"
+#include "scatter_nd_update_inst.h"
 #include "to_string_utils.h"
 #include <vector>
 #include <memory>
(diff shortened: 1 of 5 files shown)


ahnyoung-paul (approved):
LGTM

src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
```diff
@@ -1834,6 +1835,10 @@ format layout_optimizer::get_preferred_format(program_node& node) {
         // Set default format for issue 92967/98750
         // TODO: will remove when arg_max_min_ref supports blocked format
         expected = format::get_default_format(node.get_input_layouts()[0].get_rank(), false, false);
+    } else if (node.is_type<scatter_nd_update>()) {
+        // 2nd, 3rd input should not be reordered
+        node.set_preferred_input_fmt(1, get_preferred_format(node.get_dependency(1)));
+        node.set_preferred_input_fmt(2, get_preferred_format(node.get_dependency(2)));
```
yeonbok, line 1841: I think we should do it not only for scatter_nd_update but also for all primitive in new_shape_infer scenario.

kelvinchoi-intel: done

yeonbok (approved):

PR #15182, opened 2023-01-18
[GPU] Serialization logic updates for OneDNN 3.0

e-ddykim:
### Details:
 - OneDNN 3.0 has some changes over OneDNN 2.7.
 - This PR updated GPU serialization logic according the changes

### Tickets:
 - 97996

### DIFF
--- a/src/plugins/intel_gpu/src/plugin/graph.cpp
+++ b/src/plugins/intel_gpu/src/plugin/graph.cpp
@@ -67,7 +67,11 @@ Graph::Graph(cldnn::BinaryInputBuffer &ib, RemoteContextImpl::Ptr context, const
         m_config.set_property(ov::intel_gpu::max_dynamic_batch(m_program->m_max_batch));
 
 #ifdef ENABLE_ONEDNN_FOR_GPU
-    get_engine().create_onednn_engine(config);
+    bool need_onednn_engine = false;
+    ib >> need_onednn_engine;
+    if (need_onednn_engine) {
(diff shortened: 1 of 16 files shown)


src/plugins/intel_gpu/src/plugin/graph.cpp
```diff
@@ -470,6 +478,15 @@ std::shared_ptr<ngraph::Function> Graph::GetExecGraphInfoByPrimitivesInfo(std::v
...
 void Graph::Export(cldnn::BinaryOutputBuffer &ob) {
+#ifdef ENABLE_ONEDNN_FOR_GPU
+    try {
+        get_engine().get_onednn_engine();
+        ob << true;
+    } catch (ov::AssertFailure &) {
+        ob << false;
+    }
+#endif  // ENABLE_ONEDNN_FOR_GPU
+
```
yeonbok, line 489: Dont we need to handle  such a case that:
1. Model is dumped with openvino built with "ENABLE_ONEDNN_FOR_GPU" => need_onednn_engine flag is stored in the cache 
2. Next, one try to load cache with openvino built without "ENABLE_ONEDNN_FOR_GPU" by mistake => However, since the commit and driver are same, it is trying to use the cache and this data is misread?

e-ddykim: I updated the logic that checks if OneDNN is required and is enabled in the current gpu plugin, as you commented. Thank you.

yeonbok (approved):

PR #15185, opened 2023-01-18
[PT FE]: support aten::glu and aten::sigmoid_

eaidova:
### Details:
 - *support glu and inplace sigmoid activations in pytorch frontent*
 - *added tests for sigmoid*

### Tickets:
 - *101112*
 - *101126*


src/frontends/pytorch/src/op/glu.cpp
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
```
mvafin, line 6: Same comment about opset vs operation include

eaidova: done

mvafin: I still see you are using opset

tests/layer_tests/pytorch_tests/test_glu.py
```diff
@@ -0,0 +1,34 @@
...
+            def forward(self, x):
+                return F.glu(x, self.dim)
+
+        ref_net = None
+
+        return aten_glu(dim), ref_net, "aten::glu"
+
+    @pytest.mark.nightly
+    @pytest.mark.precommit
+    @pytest.mark.parametrize("dim", [0, 1, 2, 3, -1, -2])
```
mvafin, line 32: Could you also add `None` case?

src/frontends/pytorch/src/op/glu.cpp
eaidova: oops, looks like I forgot to push :)

tests/layer_tests/pytorch_tests/test_glu.py
eaidova: Torchscript failoed, 
E       glu(Tensor input, int dim=-1) -> Tensor:
E       Expected a value of type 'int' for argument 'dim' but instead found type 'NoneType'

src/frontends/pytorch/src/op/glu.cpp
eaidova: updated

src/frontends/pytorch/src/op/glu.cpp
```diff
@@ -0,0 +1,31 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/op/constant.hpp"
+#include "openvino/op/multiply.hpp"
+#include "openvino/op/sigmoid.hpp"
+#include "openvino/op/split.hpp"
+#include "utils.hpp"
```
mvafin, line 10:
```suggestion
```
Don't think you need this header here

src/frontends/pytorch/src/op_table.cpp
```diff
@@ -4,7 +4,46 @@
 
 #include "op_table.hpp"
 
-#include "openvino/opsets/opset10.hpp"
```
mvafin, line 7: Well, here using opset header actually makes sense...

mvafin: Please remove changes

eaidova: return back

mvafin (approved):

PR #15193, opened 2023-01-19
[PT FE]: support aten::var_mean and aten::remainder

eaidova:
### Details:
 - *aten::var_mean*
 - *aten::remainder*

### Tickets:
 - *101059*
 - *101056*


src/frontends/pytorch/src/op/remainder.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_remainder(NodeContext& context) {
+    auto x = context.get_input(0);
```
rkazants, line 14: can you please add a check that the node has at least two inputs? You can take a loot at TF FE where we have such routine:)

src/frontends/pytorch/src/op/remainder.cpp
```diff
@@ -0,0 +1,25 @@
...
+    auto y = context.get_input(1);
+    auto div = context.mark_node(std::make_shared<opset10::Divide>(x, y, true));
+    auto floor = context.mark_node(std::make_shared<opset10::Floor>(div));
+    auto quo = context.mark_node(std::make_shared<opset10::Multiply>(floor, y));
```
rkazants, line 18: Please check to use `Mod` operation. It looks performing the same.

eaidova: Mod has different rounding policy in OV, I get accuracy error during mod usage for this case

src/frontends/pytorch/src/op/remainder.cpp
eaidova: We discussed it with @mvafin, probably it is better to have common helper, so I make it as separated request for covering all ops

PR #15195, opened 2023-01-19
[PT FE]: Add aten::meshgrid + prim::ListUnpack transformation

mmikolajcz:
### Details:
 - *Add aten::meshgrid + prim::ListUnpack transformation*
 - *Add tests*

### Tickets:
 - *94264*
 - *101079*


mvafin (approved):

tests/layer_tests/pytorch_tests/test_listunpack.py
```diff
@@ -124,3 +124,80 @@ def test_listconstruct_getitem_listunpack(
...
+                a, b, c, d = torch.meshgrid(in1, in2, in3, in4, indexing=self.idx)
+                return a, b, c, d
+
+        ref_net = None
+
+        return prim_listunpack(idx), ref_net, "prim::ListUnpack"
+
+    @pytest.mark.parametrize("idx", ["ij", "xy"])
+    @pytest.mark.parametrize("inp", [1, 2, 3, 4])
+    @pytest.mark.nightly
```
mvafin, line 200:
```suggestion
    @pytest.mark.nightly
    @pytest.mark.precommit
```
Please add tests to precommit

mmikolajcz: Done

src/frontends/pytorch/src/transforms/prim_list_unpack_replacer.cpp
```diff
@@ -137,6 +138,81 @@ PrimListUnpackReplacer::PrimListUnpackReplacer() {
...
+            }
+
+            std::string indexing = "";
+            // Indexing - Constant with str "ij" or "xy"
+            if (meshgrid->inputs().size() == 1) {
+                // No input for indexing argument, use default.
+                indexing = "ij";
+            } else if (meshgrid->inputs().size() == 2) {
+                // Get indexing value from second input to node.
+                // TODO - if meshgrid would be placed inside loop body, cast into prim::Constant would fail due to being
```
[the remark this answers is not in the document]
mmikolajcz: Thanks for you help with explaining how to solve it, added translation for meshgrid, manual tests show that this should fix issue with indexing input

src/frontends/pytorch/src/op/meshgrid.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include "pt_framework_node.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_meshgrid(NodeContext& context) {
+    auto node = context.mark_node(std::make_shared<PtFrameworkNode>(context.get_decoder(), context.inputs()));
+    node->set_argument(0, context.get_input_from_visible_context(0));
```
mvafin, line 15: `context.get_input_from_visible_context` can return not only a constant. If it is not constant we cannot move it inside of the body graph. In this case you need to call `const_input<std::string>` for string constant (which will propagate constant from outside internally) or you can check if the node returned by `context.get_input_from_visible_context` is a `Constant` and in that case use it as input for framework node. Also you add inputs here to `PtFrameworkNode` while you need to create this node with such inputs just like any other operation.

mmikolajcz: Thanks, modified translation to use const_input<std::string> for indexing input and passed it as PtFrameworkNode attribute

src/frontends/pytorch/src/op/meshgrid.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "pt_framework_node.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_meshgrid(NodeContext& context) {
+    OutputVector inputs{context.get_input_from_visible_context(0)};
```
mvafin, line 14: Again, you should check that `context.get_input_from_visible_context(0)` is `Constant` only after that use it as input otherwise you may break graph logic if you add some non-constant node inside the body, that will add not only that node, but all subgraph it belongs to. From the looks of it you do not need to do `get_input_from_visible_context` for input 0, just use regular `context.get_input(0)`

mvafin (changes requested):

src/frontends/pytorch/src/op/meshgrid.cpp
mmikolajcz: Sorry, misunderstood your previous comment, thought that issue was only for input(1). Replaced `context.get_input_from_visible_context(0)` with `context.get_input(0)` as you suggested

src/frontends/pytorch/src/transforms/prim_list_unpack_replacer.cpp
```diff
@@ -137,6 +137,62 @@ PrimListUnpackReplacer::PrimListUnpackReplacer() {
...
+                return false;
+            }
+            NodeVector rt_copy_from{list_unpack, input_node, meshgrid_input_node};
+            OutputVector meshgrid_inputs;
+            for (auto& input : meshgrid_input_node->inputs()) {
+                meshgrid_inputs.push_back(input.get_source_output());
+            }
+
+            auto meshgrid_attrs = meshgrid->get_attrs();
+            std::string indexing = meshgrid_attrs["indexing"];
```
mvafin, line 155:
```suggestion
            std::string indexing = meshgrid_attrs.at("indexing");
```
Should be safer to do it that way

mvafin: Also please check that `indexing` exist in `attrs`

mmikolajcz: Added check if "indexing" exist in attributes, added check for indexing value correctness and implemented suggested change

mvafin (approved):

eaidova (approved):

PR #14380, opened 2022-12-02
[DOCS] Document Model Optimizer Python API

popovaan:
### Details:
 - documentation for convert_model().

PR #15197, opened 2023-01-19
[MO][PT FE] POC Integrate PT FE into MO convert_model

eaidova:
### Details:
- integrate pytorch frontend into mo.convert_model API

### Tickets:
 - *101752*


tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -672,6 +676,51 @@ def convert_pytorch_to_onnx(model, input_shape, opset_version, example_inputs, o
...
+        from openvino.frontend import FrontEndManager
+        from openvino.frontend.pytorch.decoder import TorchScriptPythonDecoder, pt_to_ov_type_map
+        fe_manager = FrontEndManager()
+        fe = fe_manager.load_by_framework('pytorch')
+    except Exception as e:
+        print("PyTorch frontend loading failed")
+        raise e
+    inputs = prepare_torch_inputs(example_inputs, input_shape)
+    if isinstance(inputs, dict):
+        example_inputs_list = list(inputs.values())
```
slyalin, line 691: Not sure that this issue wasn't there before (in ONNX export part), but what about order of this inputs, why should it be the same as PT model expects?

eaidova: onnx support dict as example_inputs, I'm not sure that torchscript support them.
After checking I found answer that it is supported, so I'll update

eaidova: updated

slyalin: Need more information about how it woks. Are keys used as parameter names in forward method?

slyalin: Yeah, already see this in the code.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -731,6 +735,62 @@ def convert_pytorch_to_onnx(model, input_shape, opset_version, example_inputs, o
...
+        pt_dtype = input_data.dtype if isinstance(input_data, torch.Tensor) else type(input_data)
+        dtype = pt_to_ov_type_map.get(str(pt_dtype))
+        if dtype is None:
+            raise f"Unknown input dtype {pt_dtype}"
+
+        input_tensor.get_node().set_element_type(dtype)
+        shape = [-1] * len(input_data.shape) if not provide_shapes else input_shape[idx]
+        input_tensor.get_node().set_partial_shape(PartialShape(shape))
+    om.validate_nodes_and_infer_types()
+    return om
```
rkazants, line 791: I think we need to re-use FE configuration (loading, setting inputs and outputs) and rely on the existing implementation from https://github.com/openvinotoolkit/openvino/blob/master/tools/mo/openvino/tools/mo/moc_frontend/pipeline.py

rkazants (changes requested):
please rely on the existing functionality from https://github.com/openvinotoolkit/openvino/blob/master/tools/mo/openvino/tools/mo/moc_frontend/pipeline.py.

tools/mo/openvino/tools/mo/convert_impl.py
eaidova: reused

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -255,7 +256,8 @@ def raise_ie_not_found():
     if is_tf:
         if argv.saved_model_tags is not None:
             if ' ' in argv.saved_model_tags:
-                raise Error('Incorrect saved model tag was provided. Specify --saved_model_tags with no spaces in it')
+                raise Error(
+                    'Incorrect saved model tag was provided. Specify --saved_model_tags with no spaces in it')
```
rkazants, line 260: please do not change it - not relevant change

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -412,8 +414,10 @@ def prepare_ir(argv: argparse.Namespace):
         if len(fallback_reasons) == 0:
             try:
                 t.send_event("mo", "conversion_method", moc_front_end.get_name() + "_frontend")
-                moc_front_end.add_extension(TelemetryExtension("mo", t.send_event, t.send_error, t.send_stack_trace))
-                moc_front_end.add_extension(ProgressReporterExtension(progress_printer(argv)))
+                # pt is not support extensions yet
```
rkazants, line 417: Please avoid grammar mistakes. It looks very bad.

rkazants, line 417:
```suggestion
                # PyTorch does not support extensions yet
```

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -394,7 +396,7 @@ def get_moc_frontends(argv: argparse.Namespace):
 def prepare_ir(argv: argparse.Namespace):
     # TODO: remove this workaround once new TensorFlow frontend supports non-frozen formats: checkpoint, MetaGraph, and SavedModel
     # Now it converts all TensorFlow formats to the frozen .pb format in case new TensorFlow frontend
-    is_tf, _, _, _, _ = deduce_legacy_frontend_by_namespace(argv)
+    is_tf, _, _, _, _, is_pt = deduce_legacy_frontend_by_namespace(argv)
```
rkazants, line 399: you do not need to change it. Please revert it:)

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -412,8 +414,10 @@ def prepare_ir(argv: argparse.Namespace):
         if len(fallback_reasons) == 0:
             try:
                 t.send_event("mo", "conversion_method", moc_front_end.get_name() + "_frontend")
-                moc_front_end.add_extension(TelemetryExtension("mo", t.send_event, t.send_error, t.send_stack_trace))
-                moc_front_end.add_extension(ProgressReporterExtension(progress_printer(argv)))
+                # pt is not support extensions yet
+                if not is_pt:
```
rkazants, line 418: you can deduce PyTorch case by just calling `moc_front_end.get_name()` :)

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -760,7 +799,8 @@ def parse_input_shapes(argv):
                 if isinstance(shapes, torch.Size):
                     return [shapes]
             except ImportError:
-                raise Error("Unknown type of input shape {}.".format(type(shapes)))
+                raise Error(
+                    "Unknown type of input shape {}.".format(type(shapes)))
```
rkazants, line 803: Please revert it

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -894,9 +934,11 @@ def input_model_is_object(argv):
 
 
 def pack_params_to_args_namespace(args: dict, cli_parser: argparse.ArgumentParser):
+    ignored_params = ["example_input", "input_signature"]
     if len(args) > 0:
         args_string = params_to_string(**args)
-        argv, _ = cli_parser.parse_known_args(args_dict_to_list(cli_parser, **args_string))
+        argv, _ = cli_parser.parse_known_args(
+            args_dict_to_list(cli_parser, **args_string))
```
rkazants, line 941: please revert it. With new line it gets worse:)

tools/mo/openvino/tools/mo/moc_frontend/serialize.py
```diff
@@ -4,14 +4,53 @@
...
+                if input_signature is not None:
+                    tensor = input_tensor.get_tensor()
+                    input_names = tensor.names
+                    input_names.update(input_signature[idx])
+                    tensor.set_names(input_names)
+                if not provide_shapes:
+                    # prevent dynamic rank issue
+                    shape = [-1] * len(input_data.shape)
+                input_tensor.get_node().set_partial_shape(PartialShape(shape))
+            ngraph_function.validate_nodes_and_infer_types()
```
rkazants, line 53: Please do not add this code here. we have already common part for FE configuration. Already left comment about it.
Also this routine is responsible for executing MOC and pre-processing transformation and IR serialization

eaidova: I'm not familar with mo code, could you please provide link, where the right place?
For Inow,  can only guess which part of the code you mean... Also it can not be done as part of preprocessing on input model, only after convert

rkazants: I have already put the link above. Duplicate one more time: https://github.com/openvinotoolkit/openvino/blob/master/tools/mo/openvino/tools/mo/moc_frontend/pipeline.py#L24

eaidova: ok, moved, thanks for clarification

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
```diff
@@ -213,4 +213,48 @@ def shape_to_array(shape: PartialShape):
...
+            if input_signature is not None:
+                tensor = input_tensor.get_tensor()
+                input_names = tensor.names
+                input_names.update(input_signature[idx])
+                tensor.set_names(input_names)
+            if not provide_shapes:
+                # prevent dynamic rank issue
+                shape = [-1] * len(input_data.shape)
+                input_tensor.get_node().set_partial_shape(PartialShape(shape))
+        ngraph_function.validate_nodes_and_infer_types()
```
rkazants, line 259: again it is a duplicate of https://github.com/openvinotoolkit/openvino/blob/master/tools/mo/openvino/tools/mo/moc_frontend/pipeline.py.
We need to adopt PT case to use `moc_pipeline`

rkazants (changes requested):

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
eaidova: it is done before running convert, in my case, it is required to do after, 
input_model.inputs() returns empty list and as I know right now, it is impossible to implement.
When, it will be ready, I remove this part of code

slyalin: Let me understand this first. Is it a correct assumption that we need this part to set dynamic dimensions in shapes and types when they do not match ones that provided with `example_input`? If that is correct, I don't understand why we need to set types in this case, because when using `example_input` we should use `torch.jit.trace` and all the types are fused into the model. No?

@eaidova, I don't understand why we cannot implement setting types and shapes in PT FE -- looks like it is just moving this code to PT FE in the right place? I mean why this is impossible to implement fe.inputs()? Do you mean that now it is not implemented and this is why we are implementing it here?

slyalin (changes requested):

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
eaidova: yes, I mean that it is not implemented now. I have nothing against to migrate into common flow, when it will be implemented (I believe, it is separated task as well as support extensions, right?)

rkazants: Now we should understand that we miss to utilize some functionality (like freezing inputs) that MO can demonstrate if FE supports it or gracefully complains on it. In the common part different logic already processed gracefully by MO and we need to utilize this logic for PT FE.
I think this is the main task of this story to have appropriate integration of PT FE into MO.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -412,8 +412,10 @@ def prepare_ir(argv: argparse.Namespace):
         if len(fallback_reasons) == 0:
             try:
                 t.send_event("mo", "conversion_method", moc_front_end.get_name() + "_frontend")
-                moc_front_end.add_extension(TelemetryExtension("mo", t.send_event, t.send_error, t.send_stack_trace))
-                moc_front_end.add_extension(ProgressReporterExtension(progress_printer(argv)))
+                # PyTorch does not support extensions yet
+                if not moc_front_end.get_name() == "pytorch":
+                    moc_front_end.add_extension(TelemetryExtension("mo", t.send_event, t.send_error, t.send_stack_trace))
+                    moc_front_end.add_extension(ProgressReporterExtension(progress_printer(argv)))
```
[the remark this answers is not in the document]
slyalin: @eaidova, can we just remove this part from PR?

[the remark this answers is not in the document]
eaidova: if I just remove it failed with error that extension loading is not implemented.
Is there simple way to ignore specific extensions and throw error for other?

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
slyalin: Yes we are missing all this things for now. Will be added later. This is a temporary workaround because a lack of shapes is a killer for the whole feature. There is the ticket particularly for inputs: CVS-102339. So this is not a long-term solution.

tools/mo/openvino/tools/mo/convert_impl.py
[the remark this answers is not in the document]
slyalin: The whole idea of the extensions is they are not mandatory to pass or to be mandatory to implement. So probably somebody changed the initial design. @mvafin, could you help with that?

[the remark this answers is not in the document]
eaidova: done

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -119,7 +119,8 @@ void FrontEnd::normalize(const std::shared_ptr<ov::Model>& model) const {
 }
 
 void FrontEnd::add_extension(const std::shared_ptr<ov::Extension>& extension) {
-    FRONT_END_NOT_IMPLEMENTED(add_extension);
+    // Extension loading mechanism is not implemented, any extensions will be ignored
```
rkazants, line 122: Can you please create a ticket and JIRA about adding extensions for PT FE and list it here

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -938,38 +1007,21 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
...
             model_framework = check_model_object(args)
-            if model_framework == "pytorch" and not os.environ.get('USE_PYTORCH_FRONTEND'):
-
-                opset_version = None
-                if 'onnx_opset_version' in args and args['onnx_opset_version'] is not None:
-                    opset_version = args['onnx_opset_version']
-
+            print(args)
+            if model_framework == "pytorch":
+                use_pt_fe = os.environ.get('USE_PYTORCH_FRONTEND') or not args.get("use_legacy_frontend", False)
```
rkazants, line 1012: Please do not use `use_legacy_frontend` option. It is only used for switching from ONNX FE to the legacy (Python) FE.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -938,38 +1007,21 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
...
                 example_inputs = None
                 if 'example_input' in args and args['example_input'] is not None:
                     example_inputs = args['example_input']
+                if not use_pt_fe:
+                    if "use_legacy_frontend" in args:
```
rkazants, line 1017: some kind of mess of MO args. We need to clarify these cases what we are going to implement

rkazants (commented):
Can you please describe the concept about switching to ONNX FE in the ticket description?
Should it be automatic switch or by option?
cc @slyalin

tools/mo/openvino/tools/mo/convert_impl.py
popovaan: @slyalin suggested to use `use_legacy_frontend`/`use_new_frontend `option to switch between PyTorch FE and converting via tmp ONNX model. So it looks correct to use `use_legacy_frontend` here. 

Check @slyalin's comments to this PR https://github.com/openvinotoolkit/openvino/pull/14380

src/frontends/pytorch/src/frontend.cpp
eaidova: added ticket number

tools/mo/openvino/tools/mo/convert_impl.py
eaidova: as agreed, we should use `use_legacy_frontend` for the switch in development purposes without mention in docs. This option will be removed later according to CVS-103726 (if the user wants to use onnx export, it should be done explicitly with all necessary parameters, which can not be guaranteed with pt->onnx->it way in conver_model due to limited parameters support).

Added TO DO comment in the code

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -1019,4 +1072,4 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
         telemetry.send_event('mo', 'conversion_result', 'fail')
         telemetry.end_session('mo')
         telemetry.force_shutdown(1.0)
-        raise e.with_traceback(None)
+        raise e#.with_traceback(None)
```
popovaan, line 1075: Don't forget to uncomment it when PR is ready.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -938,38 +1007,22 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
         inp_model_is_object = input_model_is_object(args)
         if inp_model_is_object:
             model_framework = check_model_object(args)
-            if model_framework == "pytorch" and not os.environ.get('USE_PYTORCH_FRONTEND'):
-
-                opset_version = None
-                if 'onnx_opset_version' in args and args['onnx_opset_version'] is not None:
-                    opset_version = args['onnx_opset_version']
-
+            print(args)
```
popovaan, line 1010: Please remove this print.

eaidova: thanks, removed

tools/mo/openvino/tools/mo/convert_impl.py
eaidova: thanks, return back.
BTW, is there a better solution for getting an error stack trace?

popovaan: We actually remove stack trace here to make error message smaller. Otherwise user sees a big stack stare with inner MO files which is not very nice.

eaidova: I mean in development purposes, not as part of the product (it is not a change request or something else, just some hints for me in the future :)), so it is the only known solution, it is ok.

popovaan: Yes, for development purposes commenting this line is the only way to obtain stack trace now. 
Maybe we could make a better approach. Using some env variable or if loglevel=DBUG is set. @rkazants what do you think?

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -694,12 +692,22 @@ def convert_pytorch_to_onnx(model, input_shape, opset_version, example_inputs, o
...
-        raise Error("Please provide input_shape or example_input for converting PyTorch model.")
+        if not allow_none:
+            raise Error("Please provide input_shape or example_input for converting PyTorch model.")
+    return inputs
+
+
+def convert_pytorch_to_onnx(model, input_shape, opset_version, example_inputs, output_dir):
+    import io
+    import torch
+
```
rkazants, line 708: please put all PyTorch specific routines like `convert_pytorch_to_onnx` and  `get_pytorch_decoder` to the separate python file like `moc_frontend/pytorch_frontend_utils.py`. Now `convert_impl.py` becomes (or already became) too overloaded:)

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -925,6 +965,35 @@ def remove_tmp_onnx_model(out_dir):
...
+    out_dir = args['output_dir'] if 'output_dir' in args else None
+
+    model_onnx = convert_pytorch_to_onnx(args['input_model'],
+                                         parse_input_shapes(args),
+                                         opset_version,
+                                         example_inputs,
+                                         out_dir)
+
+    args['input_model'] = model_onnx
+    if os.environ.get('SAVE_TO_BYTES_IO_ONNX_MODEL'):
```
rkazants, line 982: please make a comment what this env variable means and where this is used.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -925,6 +965,35 @@ def remove_tmp_onnx_model(out_dir):
...
+        args['use_legacy_frontend'] = True
+    args['example_input'] = None
+    args['onnx_opset_version'] = None
```
rkazants, line 985: please make a comment why we re-set these arguments to `None`

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -925,6 +965,35 @@ def remove_tmp_onnx_model(out_dir):
...
+
+    try:
```
rkazants, line 987: I think we need to move a call of `convert_pytorch_to_onnx` also under `try`. Just to avoid remaining temporal onnx file in case of failures of `convert_pytorch_to_onnx`, for example.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -938,38 +1007,21 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
...
         if inp_model_is_object:
             model_framework = check_model_object(args)
-            if model_framework == "pytorch" and not os.environ.get('USE_PYTORCH_FRONTEND'):
-
-                opset_version = None
-                if 'onnx_opset_version' in args and args['onnx_opset_version'] is not None:
-                    opset_version = args['onnx_opset_version']
-
+            if model_framework == "pytorch":
+                use_pt_fe = not args.get("use_legacy_frontend", False)
```
rkazants, line 1011: I think we should have a deal with the default value of the option here. It is already defined in the common place. Can we do as follows:

```
use_pt_fe = not args['use_legacy_frontend']
```

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -938,38 +1007,21 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
...
                 example_inputs = None
                 if 'example_input' in args and args['example_input'] is not None:
                     example_inputs = args['example_input']
+                if not use_pt_fe:
+                    # TO DO: remove this path, when pytorch frontend productization is finished, CVS-103726
+                    if "use_legacy_frontend" in args:
+                        args.pop("use_legacy_frontend")
```
rkazants, line 1018: Am I correct that we need to do this in order to avoid use of Python ONNX FE? Can you leave a comment here about it?

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -938,38 +1007,21 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
...
-                    opset_version = args['onnx_opset_version']
-
+            if model_framework == "pytorch":
+                use_pt_fe = not args.get("use_legacy_frontend", False)
                 example_inputs = None
                 if 'example_input' in args and args['example_input'] is not None:
                     example_inputs = args['example_input']
+                if not use_pt_fe:
+                    # TO DO: remove this path, when pytorch frontend productization is finished, CVS-103726
+                    if "use_legacy_frontend" in args:
```
rkazants, line 1017:
```suggestion
                    if 'use_legacy_frontend' in args and args['use_legacy_frontend']:
```

rkazants: I think this `if` will be sufficient and `use_pt_fe` is not needed at all

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
```diff
@@ -213,4 +213,56 @@ def shape_to_array(shape: PartialShape):
             input_model.set_partial_shape(place, new_partial_shape)
 
     ngraph_function = moc_front_end.convert(input_model)
+    if argv.framework == "pytorch":
+        pytorch_process_after_convert(argv, ngraph_function)
+    return ngraph_function
+
+
+def pytorch_process_after_convert(argv, ngraph_function):
```
rkazants, line 221:
```suggestion
def pytorch_process_after_convert(argv, ov_model):
```

rkazants: `ngraph_function` is outdated:)

tools/mo/openvino/tools/mo/utils/cli_parser.py
```diff
@@ -704,7 +704,8 @@ def transform_param_to_str(value):
     'example_input': ParamDescription('Sample of model input in original framework. '
                                        'For PyTorch it can be torch.Tensor.', '', '', None),
     'onnx_opset_version': ParamDescription('Version of ONNX opset that is used for converting from PyTorch to ONNX.',
-                                           '', '', None)
+                                           '', '', None),
+    'input_signature': ParamDescription('PyTorch model forward method input signature', '', '', None)
```
rkazants, line 708: can we have a shor example in the description please?

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
```diff
@@ -213,4 +213,56 @@ def shape_to_array(shape: PartialShape):
             input_model.set_partial_shape(place, new_partial_shape)
 
     ngraph_function = moc_front_end.convert(input_model)
+    if argv.framework == "pytorch":
```
rkazants, line 216: please add `TODO: finish integration of PyTorch FE into MO` and list tikcet here. I remember @andrei-kochin  already created it.
As I understand the code above should also work ideally and special `pytorch_process_after_convert` not needed

rkazants (commented):
Left last comments.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -728,6 +736,38 @@ def convert_pytorch_to_onnx(model, input_shape, opset_version, example_inputs, o
...
+    if isinstance(model, torch.nn.Module) and not isinstance(model, torch.jit._trace.TopLevelTracedModule):
+        input_signature = list(inspect.signature(model.forward).parameters.keys())
+        try:
+            scripted = torch.jit.script(model)
+        except Exception as scripting_err:
+            if example_inputs is not None:
+                try:
+                    scripted = torch.jit.trace(model, inputs)
+                except Exception as tracing_e:
+                    print('Both traicing and scripting failed')
```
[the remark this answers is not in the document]
eaidova: loger is not intialized here, as I understand

tools/mo/openvino/tools/mo/convert_impl.py
eaidova: I belive we can have cyclic dependency if I move conversion via onnx, as is it uses _convert inside...

tools/mo/openvino/tools/mo/convert_impl.py
eaidova: I do not know :) it was here before :)
probably @popovaan should better know

tools/mo/openvino/tools/mo/convert_impl.py
eaidova: added

tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
```diff
@@ -0,0 +1,214 @@
...
+            dynamic_dims = get_dynamic_dims(shape)
+            if len(dynamic_dims) > 0:
+                dynamic_dims_dict[input_names[shape_idx]] = dynamic_dims
+    additional_params = {}
+    if len(dynamic_dims_dict) > 0:
+        additional_params.update({'dynamic_axes': dynamic_dims_dict})
+    if input_names is not None and len(input_names) > 0:
+        additional_params.update({'input_names': input_names})
+
+    if os.environ.get('SAVE_TO_BYTES_IO_ONNX_MODEL'):
```
rkazants, line 124: @popovaan, please explain why we need `SAVE_TO_BYTES_IO_ONNX_MODEL`.

rkazants (approved):

rkazants:
@eaidova, please update `tools\mo\automation\package_BOM.txt` with new files. Now MO UT test fails due to this.

rkazants:
@andrei-kochin, @eaidova, @slyalin, can we use environment variable like `ONNX_FE_PASS_ENABLE` instead of using `use_legacy_frontend`? I think this solution will be more clear. How do you think?
Last comment:)

tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
popovaan: This variable switches two possible ways to pass onnx model converted from pytorch to inner convert_model(). 
1. Save to temporary onnx file - default way
2. Pass as BytesIO as object in memory - if  `SAVE_TO_BYTES_IO_ONNX_MODEL` is set. Currently only legacy MO supports input_model as BytesIO, so for inner convert_model() `use_legacy_frontend` is set.

Measurements showed that there were no significant advantage in any approach, so both were left as possible conversion way.

tools/mo/openvino/tools/mo/convert_impl.py
popovaan: Answered in comment below.

slyalin (approved):

PR #15210, opened 2023-01-19
DOCS: Update doxygen version

sgolebiewski-intel:
Updating doxygen to the latest version - 1.9.6. due to the sourceforge issues.


ilya-lavrenov:
> @ilya-lavrenov, @ilyachur it appears there's an issue with: [src/core/include/openvino/pass/pattern/op/branch.hpp](https://github.com/openvinotoolkit/openvino/blob/master/src/core/include/openvino/pass/pattern/op/branch.hpp) file:
> 
> image
> 
> Can you help? Thanks

We need to remove these 2 lines:
https://github.com/openvinotoolkit/openvino/blob/d7e77e30a8a76a3caadfd144388e2ea64b40be59/src/core/include/openvino/pass/pattern/op/branch.hpp#L26-L27
Because there is no actually parameters to that constructor. Could you please do it in current PR?

sgolebiewski-intel:
> > @ilya-lavrenov, @ilyachur it appears there's an issue with: [src/core/include/openvino/pass/pattern/op/branch.hpp](https://github.com/openvinotoolkit/openvino/blob/master/src/core/include/openvino/pass/pattern/op/branch.hpp) file:
> > image
> > Can you help? Thanks
> 
> We need to remove these 2 lines:
> 
> https://github.com/openvinotoolkit/openvino/blob/d7e77e30a8a76a3caadfd144388e2ea64b40be59/src/core/include/openvino/pass/pattern/op/branch.hpp#L26-L27
> 
> 
> Because there is no actually parameters to that constructor. Could you please do it in current PR?

Will do. Thank you for your help :)

ilya-lavrenov (approved):
Let's check by eyes that documentation is properly generated

ilyachur (approved):

ilya-lavrenov:
Let's also port changes to 2022.3?
The fixes seem useful

PR #15212, opened 2023-01-19
Add test to verify add_extension with relative path

artkuli:
Adding test to the changes from #15047 
### Tickets:
 - *98973*


src/tests/ie_test_utils/common_test_utils/file_utils.cpp
```diff
@@ -43,8 +49,74 @@ std::string getExecutableDirectory() {
...
+    auto to_vec = split_path(to);
+
+    auto mismatch_it = std::mismatch(from_vec.begin(), from_vec.end(), to_vec.begin());
+    if (mismatch_it.first == from_vec.end() && mismatch_it.second == to_vec.end()) {
+        return {};
+    }
+
+    std::string separator(1, ov::util::FileTraits<char>::file_separator);
+    std::string output;
+    if (mismatch_it.first != from_vec.end()) {
```
ilyachur, line 105: Could we add comments to this methods?
Does it work in case: `/aaa/bbb/cc/test_app` and cwd `/aaa/bbb/ddd/eee/`?

ilyachur (commented):
Some nitpicks but in general LGTM

ilyachur:
@artkuli Do you have a plan to fix comments?

artkuli:
@ilyachur I will make the suggested changes today

src/tests/ie_test_utils/common_test_utils/file_utils.cpp
```diff
@@ -2,8 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-#include <openvino/util/file_util.hpp>
 #include <cstring>
+#include <iostream>
```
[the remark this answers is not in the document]
artkuli: Removed

src/tests/ie_test_utils/common_test_utils/file_utils.cpp
artkuli: Yes, it works the generated path is: `../../../cc/test_app`

vurusovs: @artkuli could you also check case with `/` and `/some/path`?

artkuli: @vurusovs When start path is: `/` and destination path is: `/test/test_app` it generates: `../test/test_app` 
In the case where start path is: `/test/test_app` and destination is: `/` generates: `../../`

vurusovs: Case with `../test/test_app` isn't correct, but it works, so I have no issues here

vurusovs: But about the first case:
> /aaa/bbb/cc/test_app and cwd /aaa/bbb/ddd/eee/
It generates wrong output `../../../cc/test_app`, it should be `../../cc/test_app` (2 times of `../`)

Looks like there is a problem with `/path/to` and `/path/to/` difference (with last symbol which is /)

artkuli: @vurusovs  Good catch! Thanks! I figured it out, now the path is correct: `../../cc/test_app` also in the case where start path is: `/` and destination path is: `/test/test_app`, now  it generates: `test/test_app`

ilyachur (approved):

PR #15221, opened 2023-01-20
add aten::topk

bszmelcz:
### Details:
 - add translator for `aten::topk`
 - add tests

### Tickets:
 - 101120


src/frontends/pytorch/src/op/topk.cpp
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2018-2022 Intel Corporation
```
mvafin, line 1:
```suggestion
// Copyright (C) 2018-2023 Intel Corporation
```

src/frontends/pytorch/src/op/topk.cpp
```diff
@@ -0,0 +1,44 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <climits>
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
```
mvafin, line 8: Could you please use individual operation includes instead of big opset include to speed up build.

src/frontends/pytorch/src/op/topk.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "utils.hpp"
```
mvafin, line 9:
```suggestion
```
Looks like you don't need this include in this file

src/frontends/pytorch/src/op/topk.cpp
```diff
@@ -0,0 +1,44 @@
...
+    const auto sorted = context.const_input<bool>(4);
+    auto k = context.get_input(1);
+    int64_t axis{-1};
+    auto mode = opset10::TopK::Mode::MAX;
+    auto sort = opset10::TopK::SortType::SORT_VALUES;
+
+    if (!context.input_is_none(2)) {
+        axis = context.const_input<int64_t>(2);
+    }
+    if (!largest) {
```
mvafin, line 28: nit: The logic for this seems backward. It would be better to have `mode` and `sort` be defined as default first and if `largest` or `sorted` is provided modify it.

mvafin (changes requested):

src/frontends/pytorch/src/op/topk.cpp
bszmelcz: okok, I thought I will follow the default values from Pytorch.

src/frontends/pytorch/src/op/topk.cpp
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "openvino/op/topk.hpp"
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/op/convert.hpp"
```
rkazants, line 8: @mvafin, @bszmelcz, I think also we need to include `opset` header as well here.

tests/layer_tests/pytorch_tests/test_topk.py
```diff
@@ -0,0 +1,71 @@
...
+                self.dim = dim
+                self.largest = largest
+                self.sort = sort
+
+            def forward(self, input_tensor):
+                if self.dim is None:
+                    a, b = torch.topk(input_tensor, k=self.k, largest=self.largest, sorted=self.sort)
+                else:
+                    a, b = torch.topk(input_tensor, k=self.k, dim=self.dim, largest=self.largest, sorted=self.sort)
+                return b
```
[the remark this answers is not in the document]
bszmelcz: In general this way of writing tests is to force the underlying torchscript graph to not include certain operations that were npt supported before but I guess ListConstruct should work for this case so I will change that to return the topk operation

tests/layer_tests/pytorch_tests/test_topk.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @pytest.mark.parametrize(("largest"), [
+        True,
+        False,
+    ])
+    # For False it is hard to test because in Pytorch implementation
+    # there is not promise on the order of output values
+    @pytest.mark.parametrize(("sort"), [
+        True,
+    ])
```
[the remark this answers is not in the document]
bszmelcz: I will trim test cases a bit.

tests/layer_tests/pytorch_tests/test_topk.py
```diff
@@ -0,0 +1,71 @@
...
+        -1,
+        None,
+    ])
+
+    @pytest.mark.parametrize(("largest"), [
+        True,
+        False,
+    ])
+    # For False it is hard to test because in Pytorch implementation
+    # there is not promise on the order of output values
```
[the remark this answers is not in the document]
bszmelcz: @rkazants I do not know if that is possible to support this test case since the unpredictability of the order of the output values is the expected thing. The values itself match but the order does not.

tests/layer_tests/pytorch_tests/test_topk.py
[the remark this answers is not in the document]
bszmelcz: Done

src/frontends/pytorch/src/op/topk.cpp
mvafin: Do not agree, operation headers are good here to speed up building

mvafin (approved):

src/frontends/pytorch/src/op/topk.cpp
```diff
@@ -0,0 +1,43 @@
...
+    }
+    if (largest) {
+        mode = ov::op::TopKMode::MAX;
+    }
+    if (sorted) {
+        sort = ov::op::TopKSortType::SORT_VALUES;
+    }
+
+    auto topk = context.mark_node(std::make_shared<ov::op::v3::TopK>(input_tensor, k, axis, mode, sort));
+    auto indices = context.mark_node(std::make_shared<ov::op::v0::Convert>(topk->output(1), element::i64));
```
rkazants, line 35: nit comments: please use the latest opset and `using namespace ov::opset10;` to avoid extra code `v::op::v3::` etc.
You can fix it in the next PR.

rkazants (approved):

src/frontends/pytorch/src/op/topk.cpp
bszmelcz: okok thanks!

PR #15235, opened 2023-01-20
[GPU] Move local_block_io_support check to kernel selector helper

vladimir-paramuzov:
### Details:
 - `is_local_block_io_supported` moved to kernel selector helper

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.cpp
@@ -43,6 +43,9 @@
 #include "kernel_selector/kernels/quantize/quantize_kernel_params.h"
 #include "kernel_selector/kernels/reorder/reorder_kernel_base.h"
 
+#include "runtime/kernels_cache.hpp"
+#include "kernel_base.h"
+
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.h
```diff
@@ -226,12 +140,10 @@ inline params_t get_default_params(const kernel_impl_params& param_info) {
         size_t op_id = 0;
         for (auto& fused_prim : param_info.fused_desc) {
             kernel_selector::fused_operation_desc desc;
-            desc.op_params = std::move(fused_prim.f_param);
+            desc.op_params = convert_fuse_params(fused_prim.f_param);
+
+            OPENVINO_ASSERT(desc.op_params != nullptr, "[GPU] Invalid fused operation (", param_info.desc->id , ") of type ", param_info.desc->type_string());
```
sshlyapn, line 145: nit: looks like `convert_fuse_params` will never return nullptr

sshlyapn (approved):
Looks good to me, just left some minor comments

src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.cpp
```diff
@@ -1160,7 +1217,7 @@ void set_params(const kernel_impl_params& param_info, kernel_selector::params& p
     params.engineInfo.enable_sub_groups_emulation = true;
     params.engineInfo.bOptHintsSupport = false;
 
-    params.engineInfo.bLocalBlockIOSupport = device_info.supports_local_block_io && program->is_local_block_io_supported();
+    params.engineInfo.bLocalBlockIOSupport = query_local_block_io_supported(engine, config);
```
yeonbok, line 1220: If we query this here, we'll query every time of update_impl in dynamic shape which will affect the performance.. wont't it?

vladimir-paramuzov: I don't expect visible perf impact as starting from second set_params call result will be returned from cache, so it's just a matter of getting element from map

PR #15239, opened 2023-01-20
[GNA] Excluded EXACT mode from the test

mryzhov:
### Details:
 - Internal quantization algorithm is deprecating soon, the test model should be quantized using POT, that is why the EXACT mode has been excluded from the test.

### Tickets:
 - *71070*


src/plugins/intel_gna/tests/functional/shared_tests_instances/behavior/ov_infer_request/inference_chaining.cpp
```diff
@@ -8,22 +8,14 @@
 using namespace ov::test::behavior;
 namespace {
 
+// GNA_SW_EXACT mode exluded from the tests because without quantization we got the bad accuracy for this test
```
mryzhov, line 11:
```suggestion
// GNA_SW_EXACT mode excluded from the tests because without quantization we got the bad accuracy for this test
```

evkotov (approved):

sirzabek:
By removing SW_EXACT mode from the test we avoid quantization at all, be it internal or POT. How will this model's accuracy be checked in integer math then?

mryzhov:
> By removing SW_EXACT mode from the test we avoid quantization at all, be it internal or POT. How will this model's accuracy be checked in integer math then?

This test verifies inference chaining and not the particular model itself. We have such a models in subgraph tests.

PR #15242, opened 2023-01-20
TF FE import model from memory

popovaan:
Description: 
Implemented passing the model to TF FE from memory using saving model to temporary .pb file.

Ticket: 99426

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests

Documentation:


tools/mo/openvino/tools/mo/front/tf/loader.py
```diff
@@ -329,6 +329,17 @@ def convert_to_pb(argv: argparse.Namespace):
...
+        if not isinstance(argv.input_model, (str, pathlib.Path)):
+            # Convert model to Protobuf
+            model_pb = prepare_graph_def(argv.input_model)[0]
+
+            # Save model to temporary .pb file
+            tmp_dir = argv.output_dir if argv.output_dir != '.' else os.getcwd()
+            tf.io.write_graph(model_pb, tmp_dir, 'model.pb', False)
+
+            argv.input_model = os.path.join(os.path.realpath(tmp_dir), 'model.pb')
+            setattr(argv, 'need_remove_tmp_model', True)
```
rkazants, line 342: May be we should rather update `load_tf_graph_def` that will return `GraphDef` object below and the all logic for creation and maintance of the temporal file exists?

popovaan: So you suggest to move saving to file logic to `load_tf_graph_def`? So for legacy pipeline `load_tf_graph_def `should return `GraphDef `and for frontend path to .pb file?

rkazants: saving is already performed below. You need to care of providing `GraphDef` object to the logic below where we re-saved one format to another TF format

popovaan: Yes, it looks like all the logic is already implemented. `load_tf_graph_def` converts all supported formats to Protobuf and saving to file exists below in this method. So I only corrected condition to return `None` only for `string` input_model and it looks enough.

rkazants (approved):

PR #15246, opened 2023-01-21
[TF FE] Support ClipByValue operation

rkazants:
**Details:** Support ClipByValue operation and test it

**Ticket:** TBD

### DIFF
--- a/src/frontends/tensorflow/src/op/clip_by_value.cpp
+++ b/src/frontends/tensorflow/src/op/clip_by_value.cpp
@@ -1,0 +1,34 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "op_table.hpp"
+#include "openvino/opsets/opset10.hpp"
+
+using namespace std;
+using namespace ov::opset10;
+
+namespace ov {
+namespace frontend {
+namespace tensorflow {
+namespace op {
+OutputVector translate_clip_by_value_op(const NodeContext& node) {
+    default_op_checks(node, 3, {"ClipByValue"});
+    auto t = node.get_input(0);
+    auto clip_value_min = node.get_input(1);
+    auto clip_value_max = node.get_input(2);
+
+    // it can be case that clip_value_min > clip_value_max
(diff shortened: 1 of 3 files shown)


src/frontends/tensorflow/src/op/clip_by_value.cpp
```diff
@@ -0,0 +1,34 @@
...
+    default_op_checks(node, 3, {"ClipByValue"});
+    auto t = node.get_input(0);
+    auto clip_value_min = node.get_input(1);
+    auto clip_value_max = node.get_input(2);
+
+    // it can be case that clip_value_min > clip_value_max
+    // in this case both values are equal to clip_value_min
+    clip_value_max = make_shared<Maximum>(clip_value_min, clip_value_max);
+
+    auto clip_by_min = make_shared<Maximum>(t, clip_value_min);
```
itikhono, line 25: is it possible to use ov Clamp operation in case of input1,2 are const?

src/frontends/tensorflow/src/op/clip_by_value.cpp
```diff
@@ -0,0 +1,34 @@
...
+using namespace ov::opset10;
+
+namespace ov {
+namespace frontend {
+namespace tensorflow {
+namespace op {
+OutputVector translate_clip_by_value_op(const NodeContext& node) {
+    default_op_checks(node, 3, {"ClipByValue"});
+    auto t = node.get_input(0);
+    auto clip_value_min = node.get_input(1);
```
itikhono, line 18: please double check the case when `clip_value_min == clip_value_max`

itikhono (approved):

src/frontends/tensorflow/src/op/clip_by_value.cpp
rkazants: It is checked, all values for the input tensor is set to `clip_value_min`. It works.

src/frontends/tensorflow/src/op/clip_by_value.cpp
rkazants: Sure, it can be any operation with eligible output type.

PR #13944, opened 2022-11-10
[DON'T REVIEW] Use new evaluate method in template plugin

artkuli:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*

PR #15252, opened 2023-01-23
Use ov tensor in validation utils for bounds evaluation

praasz:
### Details:
 - *item1*
 - *...*

### Tickets:
 - 101296

### DIFF
--- a/src/core/include/openvino/core/validation_util.hpp
+++ b/src/core/include/openvino/core/validation_util.hpp
@@ -196,6 +196,7 @@ OPENVINO_API bool are_unique(const std::vector<int64_t>& data);
 /// \param max    Maximum value boiund
 ///
 /// \return Value if between min, max otherwise min or max.
-OPENVINO_API
-int64_t clip(const int64_t& value, const int64_t& min, const int64_t& max);
+OPENVINO_API int64_t clip(const int64_t& value, const int64_t& min, const int64_t& max);
+
+OPENVINO_API bool could_propagate(const Output<Node>& output, std::vector<Node*>& order);
 }  // namespace ov
--- a/src/core/include/openvino/core/validation_util.hpp
+++ b/src/core/include/openvino/core/validation_util.hpp
@@ -199,4 +199,53 @@ OPENVINO_API bool are_unique(const std::vector<int64_t>& data);
 OPENVINO_API int64_t clip(const int64_t& value, const int64_t& min, const int64_t& max);
 
 OPENVINO_API bool could_propagate(const Output<Node>& output, std::vector<Node*>& order);
+
+/// \brief Checks if all the elements of the bound Tensor are positive
+OPENVINO_API bool tensor_is_positive(const Tensor& bound);
+
+/// \brief Estimates upper bound for node output tensors using only upper bounds of the nodes
+/// inputs.
+/// \param node Operation to be performed
+/// \param output_values Vector of Tensors representing resulting upper value estimations
+/// \return boolean status if value evaluation was successful.
+OPENVINO_API bool default_upper_bound_evaluator(const Node* node, TensorVector& output_values);
+
+/// \brief Estimates lower bound for node output tensors using only lower bounds of the nodes
+/// inputs.
+/// \param node Operation to be performed
+/// \param output_values Vector of Tensors representing resulting lower value estimations
+/// \return boolean status if value evaluation was successful.
+OPENVINO_API bool default_lower_bound_evaluator(const Node* node, TensorVector& output_values);
+
+/// \brief Evaluates lower value estimation of the output tensor. Traverses graph up to deduce
+/// estimation through it.
+/// \param Node output pointing to the tensor for estimation.
+/// \return Tensor to estimated value.
+OPENVINO_API Tensor evaluate_lower_bound(const Output<Node>& output);
+
+/// \brief Evaluates lower value estimation of the output tensor. Traverses graph up to deduce
+/// estimation through it.
+/// \param output Tensor to be estimated.
+/// \return Tensor to estimated value.
+OPENVINO_API Tensor evaluate_upper_bound(const Output<Node>& output);
+
+/// \brief Evaluates lower and upper value estimations of the output tensor. Traverses graph up
+/// to deduce estimation through it.
+/// \param output Node output pointing to the tensor for estimation.
+/// \return pair with Tensors for lower and upper value estimation.
+OPENVINO_API std::pair<Tensor, Tensor> evaluate_both_bounds(const Output<Node>& output);
+
+/// \brief Estimates both bounds for node output tensors using both bounds of inputs. Works for
+/// operations with two inputs (in_1 and in_2). Brute forces all the pairs of bounds for inputs
+/// and evaluates all of them: {in_1_lower, in_2 lower}, {in_1_lower, in_2 upper}, {in_1_upper,
+/// in_2_lower}, {in_1_upper, in_2_upper}. Lower and upper values are selected from all the
+/// outputs calculated using input pairs.
+///
+/// \param node Operation to be performed
+/// \param lower_output_values Vector of Tensors representing resulting lower value estimations
+/// \param upper_output_values Vector of Tensors representing resulting upper value estimations
+/// \return boolean status if value evaluation was successful.
+OPENVINO_API bool interval_bound_evaluator(const Node* node,
+                                           TensorVector& lower_output_values,
+                                           TensorVector& upper_output_values);
 }  // namespace ov
--- a/src/core/include/openvino/op/add.hpp
+++ b/src/core/include/openvino/op/add.hpp
@@ -43,6 +43,7 @@ public:
     OPENVINO_SUPPRESS_DEPRECATED_START
     bool evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const override;
     OPENVINO_SUPPRESS_DEPRECATED_END
+    bool evaluate(TensorVector& outputs, const TensorVector& inputs) const override;
     bool has_evaluate() const override;
 };
 }  // namespace v1
--- a/src/core/include/openvino/op/clamp.hpp
+++ b/src/core/include/openvino/op/clamp.hpp
@@ -47,9 +47,9 @@ public:
     }
     OPENVINO_SUPPRESS_DEPRECATED_START
     bool evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const override;
-    bool evaluate_lower(const HostTensorVector& outputs) const override;
-    bool evaluate_upper(const HostTensorVector& outputs) const override;
     OPENVINO_SUPPRESS_DEPRECATED_END
+    bool evaluate_lower(TensorVector& outputs) const override;
+    bool evaluate_upper(TensorVector& outputs) const override;
     bool has_evaluate() const override;
 
 private:
--- a/src/core/include/openvino/op/concat.hpp
+++ b/src/core/include/openvino/op/concat.hpp
@@ -54,10 +54,9 @@ public:
     bool evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const override;
     OPENVINO_SUPPRESS_DEPRECATED_END
     bool has_evaluate() const override;
-    OPENVINO_SUPPRESS_DEPRECATED_START
-    bool evaluate_lower(const HostTensorVector& output_values) const override;
-    bool evaluate_upper(const HostTensorVector& output_values) const override;
-    OPENVINO_SUPPRESS_DEPRECATED_END
+    bool evaluate(TensorVector& outputs, const TensorVector& inputs) const override;
+    bool evaluate_lower(TensorVector& output_values) const override;
+    bool evaluate_upper(TensorVector& output_values) const override;
     bool evaluate_label(TensorLabelVector& output_labels) const override;
 
 protected:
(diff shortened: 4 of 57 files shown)


praasz:
@ilyachur have a look, please.

src/core/include/openvino/core/validation_util.hpp
```diff
@@ -196,6 +196,56 @@ OPENVINO_API bool are_unique(const std::vector<int64_t>& data);
...
+/// \brief Evaluates lower value estimation of the output tensor. Traverses graph up to deduce
+/// estimation through it.
+/// \param Node output pointing to the tensor for estimation.
+/// \return Tensor to estimated value.
+OPENVINO_API Tensor evaluate_lower_bound(const Output<Node>& output);
+
+/// \brief Evaluates lower value estimation of the output tensor. Traverses graph up to deduce
+/// estimation through it.
+/// \param output Tensor to be estimated.
+/// \return Tensor to estimated value.
```
ilyachur, line 229: Why do we need to have this API in public? 
Can it be a part of developer API?

src/core/include/openvino/op/add.hpp
```diff
@@ -43,6 +43,7 @@ class OPENVINO_API Add : public util::BinaryElementwiseArithmetic {
     OPENVINO_SUPPRESS_DEPRECATED_START
     bool evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const override;
```
ilyachur, line 44: Cab we remove old implementation?

src/core/include/openvino/op/clamp.hpp
```diff
@@ -47,9 +47,9 @@ class OPENVINO_API Clamp : public util::UnaryElementwiseArithmetic {
     }
     OPENVINO_SUPPRESS_DEPRECATED_START
     bool evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const override;
```
ilyachur, line 49: Can we migrate to ov::Tensor?

src/core/include/openvino/op/concat.hpp
```diff
@@ -54,10 +54,9 @@ class OPENVINO_API Concat : public Op {
     bool evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const override;
```
ilyachur, line 54: Please remove this old implementation

ilyachur (commented):
3 common comments:

- Can we hide API which is not targeted for external users and move it to developer API?
- You can remove legacy evaluate methods if you migrate to ov::Tensor.
- Can you also migrate evaluate to ov::Tensor for operations where you have changed evaluate_lower/upper?

src/core/include/openvino/op/concat.hpp
praasz: I've want to remove to not have duplicates but it looks like is not possible now. When there is no method for host tensor the `ov_template_func_tests` will fail, there is exception that operator not support evaluate. It looks like these test direct call evaluate for host tensor version.

src/core/include/openvino/op/clamp.hpp
praasz: I've add evaluate for ov::Tensor only there where looks is needed mostly in cases when there is a case for inputs which can have Shape{0} or Shape{max} which can trigger exception when converting ov::Tensor to HostTensor in base evaluate method in Node class.
And also it looks like the old version of evaluate cannot be dropped because of template plugin tests.

src/core/include/openvino/op/add.hpp
praasz: Same comment as below it looks like old implementation is required by tests.

src/core/include/openvino/core/validation_util.hpp
praasz: I'm not sure about but it looks like could be dev API part. When I modified these functions to use ov::Tensor I decide to move them from ngraph to ov namespace and move between `validation_util.hpp` files.
I've done some check if possible but when there function (or whole file) is in dev API it cannot be included in the version for ngraph namespace and all usage in all cpp file has to be updated. It looks like a lot of changes which is not main goal of this PR so would like to postpone these changes.
I will add task to shape inference improvements to move these functions to dev API and do it in separate PR.

src/core/src/op/add.cpp
```diff
@@ -78,6 +78,17 @@ bool op::v1::Add::evaluate(const HostTensorVector& outputs, const HostTensorVect
     return add::evaluate_add(inputs[0], inputs[1], outputs[0], get_autob());
 }
 
+bool op::v1::Add::evaluate(ov::TensorVector& outputs, const ov::TensorVector& inputs) const {
+    OV_OP_SCOPE(v1_Add_evaluate);
+    if (std::none_of(inputs.cbegin(), inputs.cend(), [](const ov::Tensor& t) {
+            return is_vector(t.get_shape()) && t.get_shape().front() == 0;
+        })) {
+        return BinaryElementwiseArithmetic::evaluate(outputs, inputs);
```
mitruska, line 86: `BinaryElementwiseArithmetic` doesn't override `evaluate`, so it will call `Node::evaluate` that converts Tensors to HostTensors and finally will run old `v1::Add::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)`. 
Is that a temporary workaround?

src/core/src/op/add.cpp
```diff
@@ -78,6 +78,17 @@ bool op::v1::Add::evaluate(const HostTensorVector& outputs, const HostTensorVect
...
+    } else {
+        return true;
```
mitruska, line 88: Is `set_shape` not needed for output tensor with shape that contains `0` dimension?

src/core/src/op/constant.cpp
```diff
@@ -552,6 +552,7 @@ bool ov::op::v0::Constant::visit_attributes(AttributeVisitor& visitor) {
 bool ov::op::v0::Constant::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const {
     OV_OP_SCOPE(v0_Constant_evaluate);
     auto output = outputs[0];
+    output->set_shape(get_shape());
```
mitruska, line 555: For operations other than `Constant` such `get_shape()` on node is not safe to call in evaluate, 
but it seems to be safe for `Constant` because there is guarantee it has static shape.

src/core/src/op/add.cpp
```diff
@@ -78,6 +78,17 @@ bool op::v1::Add::evaluate(const HostTensorVector& outputs, const HostTensorVect
     return add::evaluate_add(inputs[0], inputs[1], outputs[0], get_autob());
 }
 
+bool op::v1::Add::evaluate(ov::TensorVector& outputs, const ov::TensorVector& inputs) const {
+    OV_OP_SCOPE(v1_Add_evaluate);
+    if (std::none_of(inputs.cbegin(), inputs.cend(), [](const ov::Tensor& t) {
+            return is_vector(t.get_shape()) && t.get_shape().front() == 0;
```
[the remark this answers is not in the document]
praasz: No this check is for one dimension of size 0 (Shape{0}) only.

src/core/src/op/add.cpp
praasz: Is temporary when will be possible to migrate from host tensor to ov tensor and not keep both version will be not required. Just to avoid duplication for now.

src/core/src/op/constant.cpp
praasz: Generally in evalute methods the inputs should have only static shapes only (most evaluates assume static shape and gets shape from inputs), Now is possible to have dynamic because of conversion from ov::Tensor (Shape{0}) will produce dynamic Host tensor. 
But true, consent has always static shape.

src/core/src/op/add.cpp
praasz: If tensor has no dimension is not possible to change shape.

src/core/src/op/concat.cpp
```diff
@@ -107,18 +107,47 @@ bool op::Concat::evaluate(const HostTensorVector& outputs, const HostTensorVecto
     NGRAPH_CHECK(validate_host_tensor_vector(outputs, 1));
     auto concat_axis = get_axis() < 0 ? get_axis() + inputs[0]->get_shape().size() : get_axis();
     return evaluate_concat(inputs, outputs[0], concat_axis);
+    // return false;
```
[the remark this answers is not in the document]
praasz: Will remove.

src/core/src/validation_util.cpp
```diff
@@ -1348,63 +1370,75 @@ bool ov::evaluate_as_partial_shape(const Output<Node>& output, PartialShape& psh
...
     return false;
 }
 
-inline bool default_bound_evaluator(const Node* node, const HostTensorVector& output_values, bool is_upper) {
-    HostTensorVector input_tensors;
-    for (const auto& input : node->input_values()) {
-        if (auto bound = is_upper ? input.get_tensor().get_upper_value() : input.get_tensor().get_lower_value())
-            input_tensors.push_back(bound);
-        else
+bool ov::default_lower_bound_evaluator(const Node* node, TensorVector& output_values) {
```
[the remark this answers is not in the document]
praasz: Because this flag is not required and as runtime value is checked every iteration. The decision of upper/lower bound evaluation is done during programing (compile time) but this parameter is not compile time value.
Changed to have common evaluator but without check of `is_upper`.

mitruska (approved):

src/core/include/openvino/core/validation_util.hpp
ilyachur: So, in general i don't want to introduce these new functions to the public API. Because in this case we cannot remove it without deprecation (I mean released API) and we can forget to move all this API to dev before the release

src/core/include/openvino/op/clamp.hpp
ilyachur: Why do you have a problem with template functional tests?

src/core/include/openvino/op/concat.hpp
ilyachur: Can we modify template plugin?
Or it requires a lot of changes?

src/core/src/op/add.cpp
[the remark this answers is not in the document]
praasz: The shape{0} is special case in evaluate when there is transformation for host tensor and only this shape is converted to dynamic tensor (if on input can result with exception ) other shapes like Shape{0, 0} will not be converted to dynamic tensor and can be processed as normal.
If there will be only implementation for ov::Tensor this check will be not required at all.

src/core/include/openvino/op/concat.hpp
praasz: I think  modification of template plugin is part of #13944. 
@artkuli can share opinion how much work it can be to update template plugin regarding testing and dev?

Regarding migration of evaluate method to ov::tensor version it could be some work as there is a lot of operators so I would like to postpone this migration after merge #13944 and this PR. Then migration can start.

src/core/include/openvino/op/clamp.hpp
praasz: I think is related to line [evaluate in template](https://github.com/openvinotoolkit/openvino/blob/3ad92084d69d0a9a143b244e139890f8df78e552/src/plugins/template/backend/int_executable.cpp#L163) and work is ongoing in #13944

src/core/include/openvino/core/validation_util.hpp
praasz: Ok, I moved these new added functions to dev API.

artkuli (approved):

src/core/dev_api/bound_evaluation_util.hpp
```diff
@@ -0,0 +1,58 @@
...
+/// \param output_values Vector of Tensors representing resulting upper value estimations
+/// \return boolean status if value evaluation was successful.
+bool default_upper_bound_evaluator(const Node* node, TensorVector& output_values);
+
+/// \brief Estimates lower bound for node output tensors using only lower bounds of the nodes
+/// inputs.
+/// \param node Operation to be performed
+/// \param output_values Vector of Tensors representing resulting lower value estimations
+/// \return boolean status if value evaluation was successful.
+bool default_lower_bound_evaluator(const Node* node, TensorVector& output_values);
```
ilyachur, line 27: If it is a part of API and we use it in other components it should be exported.
If we use this API only inside core, I propose to remove these methods from the developer API and move to `src` folder.

praasz: Moved not required functions to `src` folder and add export macro for those which are in dev API.
Extract sources of bound evaluation functions to separate file.

src/core/dev_api/bound_evaluation_util.hpp
```diff
@@ -0,0 +1,30 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include "openvino/core/node.hpp"
+
+namespace ov {
+
+OPENVINO_API bool could_propagate(const Output<Node>& output, std::vector<Node*>& order);
```
ilyachur, line 11: Please add doxygen description

ilyachur (commented):
@jane-intel Please take a look

src/core/dev_api/bound_evaluation_util.hpp
praasz: Added

ilyachur (approved):

Issue #15138, opened 2023-01-16
Error in converting TF2 model with multiple input signature


mnl12:
Operating system linux ubuntu
Openvino, Openvino-dev, Openvino-tensorflow (all 2022.3.0)
Tensorflow 2.9.1
Model: Delg (https://github.com/tensorflow/models/tree/master/research/delf)

Hi,

I am trying to convert a saved TF2 model with multiple Input/output containing signatures. I use the following code

`mo --saved_model_dir tf_scaled_model --input input_image,input_scales,input_max_feature_num,input_abs_thres,input_global_scales_ind --input_shape [1024,1024,3],[1],[0],[0],[1]`

 

where the signature in the original code is defined as 

```
@tf.function(input_signature=[
    tf.TensorSpec(shape=[None, None, 3], dtype=tf.uint8, name='input_image'),
    tf.TensorSpec(shape=[None], dtype=tf.float32, name='input_scales'),
    tf.TensorSpec(shape=(None), dtype=tf.int32, name='input_max_feature_num'),
    tf.TensorSpec(shape=(None), dtype=tf.float32, name='input_abs_thres'),
    tf.TensorSpec(
        shape=[None], dtype=tf.int32, name='input_global_scales_ind')
])
```
However, I receive the following error,

```
[ ERROR ]  Cannot infer shapes or values for node "StatefulPartitionedCall/StatefulPartitionedCall/Any".

[ ERROR ]  Input 0 of node StatefulPartitionedCall/StatefulPartitionedCall/Any was passed int32 from StatefulPartitionedCall/StatefulPartitionedCall/Equal_port_0_ie_placeholder:0 incompatible with expected bool.
```

 

which is weird because I do not have bool input. I was wondering if you could help me in this issue.

 

Thanks.

rkazants:
Hi @mnl12,

Let me clarify your request a bit. What pre-trained model did you try to convert? Is it this one: http://storage.googleapis.com/delf/delf_gld_20190411.tar.gz?

Best regards,
Roman

mnl12:
Thanks for your reply.
Very similar to it. It is actually https://storage.googleapis.com/delf/r101delg_gldv2clean_20200914.tar.gz

rkazants:
Hi @mnl12,

Investigated the problem. It turned out that `reverse_shape_infer` for Unary operations is incorrect and uses the logic for eltwise operations. Let me fix it today.

@pavel-esir, fyi.

Best regards,
Roman

[...]

PR #15255, opened 2023-01-23
[MO] add extractor for TF Any operation

pavel-esir:
Description: There was not extractor for TF Any operation, and during shape inference MO were offloading inference into TF and this process is not always reliable and caused type inconsistency errors. Added extractor for Any according to example of other TF Reduce operations: https://github.com/openvinotoolkit/openvino/blob/5c10e3741ed14cec3a6ba8a2984dc4172a541661/src/frontends/tensorflow/src/op_table.cpp#L201-L208

Other reduce ops: All, EuclideanNorm, Max, Mean, Min, Prod, Sum - already had extractors in MO.

Ticket: xxx-101680 and issue https://github.com/openvinotoolkit/openvino/issues/15138, and partly xxx-95499

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests: for extractors no need for tests, operation itself `ReduceOps.py` already had unit-tests
* [x]  Framework operation tests: already had https://github.com/openvinotoolkit/openvino/blob/951be58dc344d6a8a7de8503d1d6e7ee636ce88e/tests/layer_tests/tensorflow_tests/test_tf_LogicalOps.py
* [x]  Transformation tests: n/a

Documentation:
* [x]  Supported frameworks operations list: done
* [x]  Guide on how to convert the **public** model: n/a
* [x]  User guide update: n/a


mvafin (commented):
Looks good, can we have a test for it?

tools/mo/openvino/tools/mo/front/tf/reduce_ext.py
```diff
@@ -17,6 +18,16 @@ def extract(cls, node: Node):
         return cls.enabled
 
 
+class AnyExtractor(FrontExtractorOp):
+    op = 'Any'
+    enabled = True
+
+    @classmethod
+    def extract(cls, node):
+        ReduceLogicalOr.update_node_stat(node, {'keep_dims':  node.pb.attr['keep_dims'].b})
```
rkazants, line 27: Shouldn't we check if `keep_dims` in attributes?
```
        keep_dims = False
        if 'keep_dims' in node.pb.attr:
            keep_dims = node.pb.attr['keep_dims'].b
```
This is according to the specification: https://www.tensorflow.org/api_docs/python/tf/raw_ops/Any

rkazants:
> Looks good, can we have a test for it?

Please add layer test for this operation. So we can re-use it for TF FE as well.

pavel-esir:
> > Looks good, can we have a test for it?
> 
> @pavel-esir, Please add layer test for this operation. So we can re-use it for TF FE as well)

There are already layer tests for Any, i will check if they work correctly 
https://github.com/openvinotoolkit/openvino/blob/951be58dc344d6a8a7de8503d1d6e7ee636ce88e/tests/layer_tests/tensorflow_tests/test_tf_LogicalOps.py

rkazants:
> > > Looks good, can we have a test for it?
> > 
> > 
> > @pavel-esir, Please add layer test for this operation. So we can re-use it for TF FE as well)
> 
> There are already layer tests for Any, i will check if they work correctly https://github.com/openvinotoolkit/openvino/blob/951be58dc344d6a8a7de8503d1d6e7ee636ce88e/tests/layer_tests/tensorflow_tests/test_tf_LogicalOps.py

Very interesting. How do the tests pass in the pre-commit then?

pavel-esir:
@rkazants i checked existing layer test, it was disabled for legacy frontend, looks like for very this reason there was not `Any` extractor and it was failing on master. But with the current extractor we can safely enable test for both legacy and cpp TF frontend.  I enabled Any layer tests for all cases

tools/mo/openvino/tools/mo/front/tf/reduce_ext.py
pavel-esir: It's not necessary, because `keep_dims` is always present in the pb, even if user didn't specify it TF puts default `False` into protobuf. We have exactly same extractor for all tensorflow logical reduce ops
https://github.com/openvinotoolkit/openvino/blob/0c9abf43a91fbe2dc95974b6cb2651bd883d1b12/tools/mo/openvino/tools/mo/front/tf/reduce_ext.py#L19-L47

Also i double checked locally. Here it is, i didn't specify `keeps_dims` when created the op and it's still in the pb
image

image

mvafin (approved):

PR #15263, opened 2023-01-23
[GPU] Perform memory transfer from usm_host to usm_device for dGPU only 

sshlyapn:
### Details:
 - This PR disables usm_host -> usm_device memory transfers in case of iGPUs since it doesn't provide any gain

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing.cpp
@@ -329,31 +329,39 @@ void prepare_primitive_fusing::fuse_bias(program &p) {
             new_node.recalc_output_layout();
         };
 
-        auto recalculate_biases = [&](data_node& original_node, data_node& new_node) -> bool {
+        auto recalculate_biases = [&](data_node& original_node, data_node& second_node) -> bool {
             auto original_mem = original_node.get_attached_memory_ptr();
-            auto new_mem = new_node.get_attached_memory_ptr();
-            if (original_mem->count() != new_mem->count() || original_mem->get_layout().data_type != new_mem->get_layout().data_type)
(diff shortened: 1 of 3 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing.cpp
```diff
@@ -329,31 +329,39 @@ void prepare_primitive_fusing::fuse_bias(program &p) {
...
                     uint16_t* original_data = original_bias_mem.data();
-                    uint16_t* new_data = new_bias_mem.data();
+                    uint16_t* new_data = second_bias_mem.data();
                     for (size_t i = 0; i < original_bias_mem.size(); i++) {
                         float new_val = half_to_float(original_data[i]) + half_to_float(new_data[i]);
-                        original_data[i] = float_to_half(new_val);
+                        new_bias_mem[i] = float_to_half(new_val);
                     }
+
+                    original_node.attach_memory(new_mem);
```
isanghao, line 364: why do you replace original_data? is it also related to igpu optimization?

sshlyapn: I faced a problem in fusion tests for this pass, because both (fused_ and ref_) networks use the same data memory, and during program build of fused_network this memory is modified by this optimization pass which affects ref_network's memory as well and leads to incorrect results
Also, I think it would be more correctly, since in real scenario during Constant Op creation different constants may be reused by several primitives (if constant values are the same), and then modification of data w/o taking 
into account it's users may lead to data invalidation for other users after this optimization. This change allows to keep original memory unchanged and attaches new one

isanghao: oh I see.. thanks for the explanation.

isanghao (approved):

PR #15265, opened 2023-01-23
[PyOV] Allow `numpy` patch releases

p-wysocki:
### Details:
 - Allow for numpy patch releases
 - Python bindings numpy version stays on 1.23 beacuse ONNX==1.13 does not support numpy==1.24. It will be updated once ONNX=1.13.1 is released and updated in OV repo.

### Tickets:
 - N/A, proposed in https://github.com/openvinotoolkit/openvino/pull/15247


src/frontends/tensorflow/tests/requirements.txt
```diff
@@ -1,2 +1,2 @@
-numpy>=1.16.6,<=1.23.4
```
p-wysocki, line 1: @rkazants Can we bump numpy here?

rkazants: yes, let us bump version here as well.

ilya-lavrenov (approved):

tools/mo/requirements_onnx.txt
```diff
@@ -1,7 +1,7 @@
 onnx>=1.8.1,<=1.12
 networkx~=2.5; python_version <= "3.6"
 networkx<=2.8.8; python_version > "3.6"
-numpy>=1.16.6,<=1.24.0
+numpy>=1.16.6,<1.25.0
```
jiwaszki, line 4:
```suggestion
numpy>=1.16.6,<1.24
```
Am I wrong?
> Python bindings numpy version stays on 1.23 beacuse ONNX==1.13 does not support numpy==1.24.

p-wysocki: I'm not certain - ONNX doesn't support numpy==1.24 only when it comes to running ONNX backend tests through OV, since MO is a separate component I think it technically *could* be bumped. We could downgrade the numpy version just to be safe though.

rkazants: @p-wysocki, can you check please if MO works for ONNX models converted by the legacy MO `--use_legacy_frontend` using your requirements?

rkazants: No, I didn't find them. I looked for models for which we fallback to the legacy ONNX.

p-wysocki: @rkazants I converted `yolov3-tiny.onnx` using master version of MO and `numpy==1.24`, the IR has been generated successfully. Is this test sufficient?

rkazants: yes, it is. Thanks!

PR #15268, opened 2023-01-23
Support rt info attributes which start from numbers

ilyachur:
### Details:
 - XML cannot have attributes which starts from numbers.
 - Pugixml cannot read such IRs
 - The service prefix added for such attributes

### Tickets:
 - 101287


src/core/src/pass/serialize.cpp
```diff
@@ -795,7 +795,8 @@ void auto_pad_resolving(ov::Node* node) {
 }
 
 void serialize_rt_info(pugi::xml_node& root, const std::string& name, const ov::Any& data) {
-    auto child = root.append_child(name.c_str());
```
rkazants, line 798: Before serialization should we check that there is no such attributes starting with the prefix `_ov_prefix_`?

ilyachur: We introduce this prefix ONLY in order to serialize and deserialize tags which start from numbers. It is why we introduce it. So serialized IR will have such prefix.

akladiev:

akladiev:

PR #15273, opened 2023-01-23
[PyOV] Add deepcopy for Model

akuporos:
### Details:
 - add the `deepcopy` magic method for Model class

### Tickets:
 - 100919


p-wysocki (approved):
LGTM

src/bindings/python/src/pyopenvino/graph/model.cpp
```diff
@@ -706,6 +706,19 @@ void regclass_graph_Model(py::module m) {
         return "<" + class_name + ": '" + self.get_friendly_name() + "'\ninputs[\n" + inputs_str + "\n]\noutputs[\n" +
                outputs_str + "\n]>";
     });
+
+    model.def(
+        "__deepcopy__",
+        [](ov::Model& self, py::dict& memo) {
```
jiwaszki, line 712:
```suggestion
        [](ov::Model& self, py::dict) {
```
Small nit, there is no need for `memo` in this scenario. It could work without argument.

src/bindings/python/src/pyopenvino/graph/model.cpp
```diff
@@ -706,6 +706,19 @@ void regclass_graph_Model(py::module m) {
         return "<" + class_name + ": '" + self.get_friendly_name() + "'\ninputs[\n" + inputs_str + "\n]\noutputs[\n" +
                outputs_str + "\n]>";
     });
+
+    model.def(
+        "__deepcopy__",
```
jiwaszki, line 711: Can this class have a whole copy-interface? `__copy__` can be raised with error - does it make sense?

src/bindings/python/src/pyopenvino/graph/model.cpp
akuporos: done

ilyachur (approved):

PR #15274, opened 2023-01-24
[PT FE] Add aten::repeat_interleave operator

sikorsl1:
### Details:
 - `aten::repeat_interleave` operator has been added
 
### Tickets:
 - 99228


src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
+#include "utils.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector generate_indices_from_repeats_tensor(std::vector<int64_t> repeats, NodeContext& context) {
```
mvafin, line 14: Please add helper functions to nameless namespace

tests/layer_tests/pytorch_tests/test_repeat_interleave.py
```diff
@@ -0,0 +1,42 @@
...
+                self.dim = dim
+
+            def forward(self, input_tensor):
+                return input_tensor.repeat_interleave(self.repeats, self.dim)
+
+        ref_net = None
+
+        return aten_repeat_interleave(), ref_net, "aten::repeat_interleave"
+
+    @pytest.mark.nightly
```
mvafin, line 37:
```suggestion
    @pytest.mark.nightly
    @pytest.mark.precommit
```
You need to also add tests to precommit

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,85 @@
+// Copyright (C) 2018-2022 Intel Corporation
```
mvafin, line 1:
```suggestion
// Copyright (C) 2018-2023 Intel Corporation
```

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,85 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
+#include "utils.hpp"
```
mvafin, line 7: Do you really need this header include?

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector generate_indices_from_repeats_tensor(std::vector<int64_t> repeats, NodeContext& context) {
+    OutputVector all_indices;
+    for (int i = 0; i < repeats.size(); i++) {
+        Shape indices_shape{(unsigned int)repeats.at(i)};
```
mvafin, line 17:
```suggestion
        Shape indices_shape{static_cast<size_t>(repeats.at(i))};
```
`Shape` has `size_t` type underneath it

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,85 @@
...
+    auto const_0 = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {0}));
+    auto const_1 = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {1}));
+    auto const_neg_1 = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {-1}));
+
+    // inputs
+    auto input = context.get_input(0);
+    auto repeats = context.const_input<std::vector<int64_t>>(1);
+
+    auto const_repeats =
+        context.mark_node(opset10::Constant::create(element::i64, Shape{2}, {repeats.at(0), (int64_t)1}));
```
mvafin, line 36: Could you please add check that  `repeats` has at least 1 element?

mvafin (changes requested):
What if `repeats` is not constant? Is it possible to support such case? Is it constant in all models we have?

src/frontends/pytorch/src/op/repeat_interleave.cpp
sikorsl1: done

tests/layer_tests/pytorch_tests/test_repeat_interleave.py
sikorsl1: done

src/frontends/pytorch/src/op/repeat_interleave.cpp
sikorsl1: done

src/frontends/pytorch/src/op/repeat_interleave.cpp
sikorsl1: Oh thanks, it is not needed.

sikorsl1: done

src/frontends/pytorch/src/op/repeat_interleave.cpp
sikorsl1: Thanks, done

src/frontends/pytorch/src/op/repeat_interleave.cpp
sikorsl1: done

sikorsl1:
> What if `repeats` is not constant? Is it possible to support such case? Is it constant in all models we have?

Generally, `repeats` does not have to be constant, but without any assumption about `repeats` shape or input shape usage of `Loop` or `TensorIterator` will be needed. So we have a dilemma here - we can cover the most general case using `TensorIterator` or we can make some assumptions about input shape and do not cover the most general case or we can treat `repeats` as constant, but it also won't cover the most general case.  
@mvafin What do you think about it?

mvafin:
> > What if `repeats` is not constant? Is it possible to support such case? Is it constant in all models we have?
> 
> Generally, `repeats` does not have to be constant, but without any assumption about `repeats` shape or input shape usage of `Loop` or `TensorIterator` will be needed. So we have a dilemma here - we can cover the most general case using `TensorIterator` or we can make some assumptions about input shape and do not cover the most general case or we can treat `repeats` as constant, but it also won't cover the most general case. @mvafin What do you think about it?

The question is: do we have models with `repeats` not constant. Could you check that?

sikorsl1:
> > > What if `repeats` is not constant? Is it possible to support such case? Is it constant in all models we have?
> > 
> > 
> > Generally, `repeats` does not have to be constant, but without any assumption about `repeats` shape or input shape usage of `Loop` or `TensorIterator` will be needed. So we have a dilemma here - we can cover the most general case using `TensorIterator` or we can make some assumptions about input shape and do not cover the most general case or we can treat `repeats` as constant, but it also won't cover the most general case. @mvafin What do you think about it?
> 
> The question is: do we have models with `repeats` not constant. Could you check that?

I found out that models use non constant `repeats`. But what I also found out is that in models that we are currently working on, `repeats` contains only one element, which makes the implementation of "non constant repeats case" quite easy.

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,119 @@
...
+    auto const_0 = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {0}));
+    auto const_1 = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {1}));
+    auto const_1_list = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {1}));
+    auto const_neg_1 = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {-1}));
+
+    // inputs
+    auto input = context.get_input(0);
+    std::shared_ptr<ov::Node> result;
+
+    try {
```
mvafin, line 37: Try catch is not required in this case, please avoid it. You need to get `reapeats` as non constant and after that try to do `dynamic_pointer_cast` to `Constant` if it casted fine you can get const value from it.

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+                // case (repeats=tensor, dim=number)
+                OutputVector all_indices = generate_indices_from_repeats_tensor(repeats, context);
+                auto concat = context.mark_node(std::make_shared<opset10::Concat>(all_indices, 0));
+                result = std::make_shared<opset10::Gather>(input, concat, const_dim);
+            }
+        }
+    } catch (...) {
+        // repeats is not Constant
+        auto repeats_input = std::make_shared<opset10::Reshape>(context.get_input(1), const_1_list, false);
```
mvafin, line 86: Would be good to have a comment here that we only support 1 element `repeats` and in other cases this `Reshape` will fail

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,119 @@
...
+            // case (repeats=number, dim=None)
+            auto flat_shape = context.mark_node(opset10::Constant::create(element::i64, Shape{2}, {1, -1}));
+            auto reshape = context.mark_node(std::make_shared<opset10::Reshape>(input, flat_shape, false));
+            auto tile = context.mark_node(std::make_shared<opset10::Tile>(reshape, repeats));
+            auto shape_perm = context.mark_node(opset10::Constant::create(element::i64, Shape{2}, {1, 0}));
+            auto transpose = context.mark_node(std::make_shared<opset10::Transpose>(tile, shape_perm));
+            result = std::make_shared<opset10::Reshape>(transpose, const_neg_1, false);
+        } else {
+            // case (repeats=number, dim=number)
+            auto const_dim = context.get_input(2);
```
mvafin, line 98: Name seems misleading dim is not const here

mvafin (changes requested):

src/frontends/pytorch/src/op/repeat_interleave.cpp
sikorsl1: done

src/frontends/pytorch/src/op/repeat_interleave.cpp
sikorsl1: done

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,93 @@
...
+            result = std::make_shared<opset10::Gather>(reshape, concat, const_0);
+        } else {
+            // case (repeats=tensor, dim=number)
+            auto dimension = context.get_input(2);
+            OutputVector all_indices = generate_indices_from_repeats_tensor(repeats, context);
+            auto concat = context.mark_node(std::make_shared<opset10::Concat>(all_indices, 0));
+            result = std::make_shared<opset10::Gather>(input, concat, dimension);
+        }
+    } else {
+        // repeats is not Constant
```
mvafin, line 58:
```suggestion
        // repeats is not Constant or single element constant
```

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,93 @@
...
+    auto const_neg_1 = context.mark_node(opset10::Constant::create(element::i32, Shape{1}, {-1}));
+
+    // inputs
+    auto input = context.get_input(0);
+    std::shared_ptr<ov::Node> result;
+
+    auto repeats_ext_node = context.get_input_from_visible_context(1).get_node_shared_ptr();
+    auto repeats_fw_node = std::dynamic_pointer_cast<opset10::Constant>(repeats_ext_node);
+    if (repeats_fw_node && repeats_fw_node->cast_vector<int32_t>().size() > 1) {
+        // repeats is Constant
```
mvafin, line 41:
```suggestion
        // repeats is Constant with more then 1 element
```

mvafin (approved):

src/frontends/pytorch/src/op/repeat_interleave.cpp
```diff
@@ -0,0 +1,93 @@
...
+        } else {
+            // case (repeats=tensor, dim=number)
+            auto dimension = context.get_input(2);
+            OutputVector all_indices = generate_indices_from_repeats_tensor(repeats, context);
+            auto concat = context.mark_node(std::make_shared<opset10::Concat>(all_indices, 0));
+            result = std::make_shared<opset10::Gather>(input, concat, dimension);
+        }
+    } else {
+        // repeats is not Constant or single element constant
+        // Curently we support only case when repeats contains only one element. Otherwise next Reshape will fail.
```
mmikolajcz, line 59: I've checked how it fails, unfortunately it seem to just raise runtime error with exception from Reshape instead of just returning message of unconverted operators. Since this is known limitation, I think better exception handling would be needed

mmikolajcz (changes requested):

src/frontends/pytorch/src/op/repeat_interleave.cpp
sikorsl1: Actually no exception here is thrown, because on the conversion stage shape of `repeats` is not known. Exception is thrown during "infer shapes and types" stage.

mmikolajcz: My bad, I though that failure was during conversion, not infer_shapes. In that case, I don't think we can do much more

mmikolajcz (approved):

PR #15277, opened 2023-01-24
Reduce the number of validate and infer types in ConvertPrecision

mateusztabaka:
Currently, ConvertPrecision pass frequently runs validate and infer types. This is due to the fact that it iterates over every precision pair, then over the whole model followed by validate and infer types. The proposed solution is to iterate over the model: for each node iterate over precisions array, update the node if required followed by validate and infer types.

Ticket: 81311


itikhono:
@rkazants could you try this PR for with obj detection models?

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -527,85 +604,121 @@ bool fuse_type_to_multiclass_nms(const std::shared_ptr<ngraph::Node>& node, ngra
...
     auto generate_proposals = ov::as_type_ptr<opset9::GenerateProposals>(node);
     if (!generate_proposals) {
         return false;
     }
 
-    if ((idx == 2) && (to == ov::element::i32 || to == ov::element::i64)) {
-        generate_proposals->set_roi_num_type(to);
-        return true;
+    auto it = precisions.find(node->get_output_element_type(2));
+    if (it != precisions.end()) {
```
itikhono, line 626: I believe we can implement a separate helper function for all these cases,e.g.
```
bool update_type(size_t idx, function update_method) {
auto it = precisions.find(node->get_output_element_type(idx));
    if (it != precisions.end()) {
        const auto& to = it->second;
        if (to == ov::element::i32 || to == ov::element::i64) {
            update_method();
            return true;
        }
    }
    }
```
    
and after that we will be able to simplify all these calls, e.g.
`update_type(2, [](){generate_proposals->set_roi_num_type(to);}) 
`
what do you think?

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -136,63 +141,64 @@ bool convert_precision(ov::pass::PassBase& pass,
...
                 const auto t2f_it = type_to_fuse.find(node->get_type_info());
                 if (t2f_it != type_to_fuse.end()) {
-                    res |= t2f_it->second(node, to, output.get_index());
+                    res = t2f_it->second(node, precisions);
                 }
-            }
-        }
-        return res;
-    };
+                if ((is_changed || is_output_precision_changed || res) && !node_is_replaced(node)) {
```
itikhono, line 172: Are you sure that we need to invoke `node_is_replaced`?
As I can see, the `ops` is updated on line 204 
image
I believe the topological sort (get_ordered_ops) won't return nodes without consumers

src/plugins/intel_cpu/src/transformation_pipeline.cpp
```diff
@@ -112,7 +112,15 @@ namespace intel_cpu {
 
 using const_node_ptr = const std::shared_ptr<const ov::Node>;
 
-bool Transformations::fuse_type_to_convert(const std::shared_ptr<ngraph::Node>& node, ov::element::Type to, size_t idx) {
+bool Transformations::fuse_type_to_convert(const std::shared_ptr<ngraph::Node>& node, const precisions_map& precisions) {
+    const auto& from = node->get_output_element_type(0);
+    auto type_pair = std::find_if(precisions.begin(), precisions.end(),
```
praasz, line 117: As `precisions` is map maybe find function member will work (maybe cast from to Type_t to be same as key is required)?
```suggestion
    auto type_pair = precisions.find(from);
```

mateusztabaka: done

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -136,63 +141,64 @@ bool convert_precision(ov::pass::PassBase& pass,
...
 
-    auto convert_node_output_precision = [&](const std::shared_ptr<ngraph::Node>& node) {
-        bool res = false;
-        // Handle case with Constants as they can have consumers from other nGraph Function object
-        const auto constant = ov::as_type_ptr<opset10::Constant>(node);
-        const auto it = const_to_internal_output.find(node.get());
-        if (constant && constant->get_output_element_type(0) == from && it != const_to_internal_output.end()) {
-            return fuse_type_to_constant(node, to, it->second);
-        }
+    std::function<bool(const std::shared_ptr<Model>&, bool)> convert_function_precision =
```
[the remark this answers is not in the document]
mateusztabaka: done

src/common/transformations/src/transformations/convert_precision.cpp
mateusztabaka: done

src/common/transformations/src/transformations/convert_precision.cpp
mateusztabaka: the node handler (called at line 170) can also replace the node. If that's the case we don't have to revalidate the old one and sometimes we even can't (e.g. if the new node is type relaxed the old one can throw during validate_and_infer_types

praasz (approved):

itikhono (approved):

itikhono:
please restart pre-commit

PR #15280, opened 2023-01-24
[ONNX FE] Added support of DFT operator

mbencer:
Ticket: 79173
Spec: https://github.com/onnx/onnx/blob/main/docs/Operators.md#DFT


src/frontends/onnx/frontend/src/op/dft.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace ngraph {
+namespace onnx_import {
+namespace op {
+namespace set_1 {
+
+namespace {
+// For DFT, IDFT, IRDFT cases, if real data are provided (with shape [D_0, D_1, ..., D_{N-1}, 1])
+// it's needed to fill tensors with zero imaginary part to be aligned with Core ops requirements.
+ov::Output<ov::Node> convert_real_to_complex_if_needed(const ov::Output<ov::Node>& data,
```
mbencer, line 18: @mateusztabaka Do you think that it makes sense to fill with zero imaginary part also IDFT, IRDFT case? ONNX spec is not very clear for such cases (in ONNX tests there is only case of using DFT(inversed=0,oneside=0) with real numbers as input)

mateusztabaka: > Do you think that it makes sense to fill with zero imaginary part also IDFT, IRDFT case

Yes, I think so.

Also, should we pad output of IRDFT with zeros?

mbencer: @mateusztabaka I've checked Pytorch and I don't see unsqueeze for IRDFT output case ([..,..,1]), for ONNXRT IRDFT (inverse + onesided) is not supported

src/frontends/onnx/frontend/src/op/dft.cpp
```diff
@@ -0,0 +1,85 @@
...
+namespace {
+// For DFT, IDFT, IRDFT cases, if real data are provided (with shape [D_0, D_1, ..., D_{N-1}, 1])
+// it's needed to fill tensors with zero imaginary part to be aligned with Core ops requirements.
+bool try_convert_real_to_complex(ov::Output<ov::Node>& data) {
+    if (data.get_partial_shape().rank().is_static()) {
+        const auto length = data.get_partial_shape().rank().get_length();
+        const auto last_axis_pos = length - 1;
+        const auto last_dim = data.get_partial_shape()[last_axis_pos];
+        if (last_dim.is_static()) {
+            if (last_dim.get_length() == 1) {
```
tomdol, line 24: I think you can merge this if with the one above.

src/frontends/onnx/frontend/src/ops_bridge.cpp
```diff
@@ -332,7 +333,7 @@ OperatorsBridge::OperatorsBridge() {
     REGISTER_OPERATOR("DequantizeLinear", 1, dequantize_linear);
     REGISTER_OPERATOR("DequantizeLinear", 13, dequantize_linear);
     REGISTER_OPERATOR("Div", 1, div);
-    REGISTER_OPERATOR("Div", 7, div);
```
tomdol, line 335: Accidental removal of Div-7?

tomdol (changes requested):

src/frontends/onnx/frontend/src/op/dft.cpp
```diff
@@ -0,0 +1,85 @@
...
+    // importing.
+    return false;
+}
+}  // namespace
+
+OutputVector dft(const Node& node) {
+    const OutputVector ng_inputs{node.get_ng_inputs()};
+    ov::Output<ov::Node> data = ng_inputs.at(0);
+
+    const auto dft_length_provided = ng_inputs.size() > 1;
```
mitruska, line 43: There can be a null node provided as the optional input, it's worth to add a check:
```suggestion
    const auto dft_length_provided = ng_inputs.size() > 1 && !is_null(ng_inputs.at(1));

```

mbencer: a good catch  ;)

src/frontends/onnx/frontend/src/ops_bridge.cpp
mbencer: not really :D

mitruska (approved):

tomdol (approved):

PR #15290, opened 2023-01-24
[PyOV] Adding optimizations and shared memory mode to inference calls

jiwaszki:
### Details:
 - WIP

### Tickets:
 - *101379*


yury-gorbachev (commented):
Generally looks good to me, would love to see if there is performance difference with new callable implementation...

src/bindings/python/src/openvino/runtime/ie_api.py
```diff
@@ -265,12 +198,71 @@ def infer_new_request(self, inputs: Union[dict, list, tuple, Tensor, np.ndarray]
...
+
+        The allowed types of keys in the `inputs` dictionary are:
+
+        (1) `int`
+        (2) `str`
+        (3) `openvino.runtime.ConstOutput`
+
+        The allowed types of values in the `inputs` are:
+
+        (1) `numpy.array`
```
AlexKoff88, line 225: `numpy.array` should be replaced by `numpy.ndarray`. It was an opening to me that some other types can also work, e.g. `torch.Tensor` which is cast to numpy.ndarray implicitly. Based on that I would rephrase the item `(1)` to:
```
(1) `numpy.ndarray` and all the types that are castable to it, e.g. `torch.Tensor`.
```

src/bindings/python/src/openvino/runtime/ie_api.py
```diff
@@ -265,12 +198,71 @@ def infer_new_request(self, inputs: Union[dict, list, tuple, Tensor, np.ndarray]
...
+        (2) `openvino.runtime.Tensor`
+
+        Can be called with only one `openvino.runtime.Tensor` or `numpy.array`,
+        it will work only with one-input models. When model has more inputs,
+        function throws error.
+
+        :param inputs: Data to be set on input tensors.
+        :type inputs: Union[Dict[keys, values], List[values], Tuple[values], Tensor, numpy.array], optional
```
AlexKoff88, line 233: Please replace `numpy.array` with `numpy.ndarray` everywhere.

jiwaszki: Combining with comment above - good point. Currently the same comment is done by adding `* array-like object with __array__ attribute`. However, it makes more sense to just squeeze it to one-liner comment.

jiwaszki: Done!

src/bindings/python/src/openvino/runtime/ie_api.py
jiwaszki: Done!

src/bindings/python/tests/test_runtime/test_infer_request.py
```diff
@@ -475,12 +483,17 @@ def callback(request, job_id):
...
+
+    if not shared_flag:
+        img = generate_image()
     infer_queue.set_callback(callback)
     assert infer_queue.is_ready()
 
     for i in range(jobs):
-        infer_queue.start_async({"data": img}, i)
+        if shared_flag:
+            img = generate_image()
```
p-wysocki, line 495: Could you please explain why a new `img` needs to be created for every async if `shared_memory=True`? Would it matter if the random image seed is constant? Or is it about asserting that the original `img` object is still alive after overwriting it in the next loop iterations? Not a change request, just want to understand what's going on :)

p-wysocki (approved):
LGTM, I really like the idea of setting `shared_memory=True` by default for `__call__`.

src/bindings/python/tests/test_runtime/test_infer_request.py
jiwaszki: Discussed in details offline. Short note: due to potential memory conflicts/data-races, there must should be separated images.

AlexKoff88 (approved):

jane-intel (approved):
🚀

akuporos (approved):

PR #15291, opened 2023-01-24
Fix "(un)register_plugin(s)" C and Py tests

vurusovs:
### Details:
 - After https://github.com/openvinotoolkit/openvino/pull/15073 C and Py "register_plugin(s)" tests start failing because it depends on libopenvino_intel_cpu_plugin.so which place is unknown for reregistering

### Tickets:
 - 101968


src/bindings/python/tests/test_runtime/test_core.py
```diff
@@ -273,27 +271,15 @@ def test_register_plugins(device):
     full_device_name = core.get_property(device, "FULL_DEVICE_NAME")
     plugins_xml = plugins_path(device, full_device_name)
     core.register_plugins(plugins_xml)
-    model = core.read_model(model=test_net_xml, weights=test_net_bin)
-    compiled_model = core.compile_model(model, "CUSTOM")
-    os.remove(plugins_xml)
```
akuporos, line 278: keep this line as it cleaned produced file `plugins_xml`

akuporos, line 278:
```suggestion
os.remove(plugins_xml)
```
keep this line as it cleaned produced file plugins_xml

vurusovs: Thanks! Reverted

ilyachur (approved):

PR #15300, opened 2023-01-25
Print unknown argument itself instead of pointer address

Jackojc:
### Details:
Fixes issue in compile_tool whereby unknown argument error prints address instead of string.

### Tickets:
 - 102048


ilya-lavrenov (approved):

Jackojc:
@ilya-lavrenov Do you think this is okay to merge now?

PR #15301, opened 2023-01-25
[CPU] [DEBUG CAPS] Extend performance info for dynamic shapes

EgorDuplensky:
### Details:
 - Measurements for node shapeInfer and prepareParams stages were added
including support of aggregate tables in csv format.
 - Verbose output was appended by corresponding durations
and prepareParams cache hit info.

### Tickets:
 - 73282

### DIFF
--- a/src/plugins/intel_cpu/src/graph.cpp
+++ b/src/plugins/intel_cpu/src/graph.cpp
@@ -998,11 +998,13 @@ void Graph::PullOutputData(BlobMap &out) {
 }
 
 void Graph::InferStatic(InferRequestBase* request) {
+    CPU_DEBUG_CAP_ENABLE(const PerfKey perfKey = perfGetKey(*this));
+
     dnnl::stream stream(getEngine());
 
     for (const auto& node : executableGraphNodes) {
-        VERBOSE(node, getConfig().debugCaps.verbose);
-        PERF(node, getConfig().collectPerfCounters);
+        VERBOSE(node, getConfig().debugCaps.verbose, infer_count);
+        PERF(node, getConfig().collectPerfCounters, perfKey);
 
         if (request)
             request->ThrowIfCanceled();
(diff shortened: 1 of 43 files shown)


EgorDuplensky:
The performance impact of the new counters will be double checked before merge

akladiev:

src/plugins/intel_cpu/src/node.cpp
```diff
@@ -554,6 +554,7 @@ std::vector<memory::format_tag> Node::getAvailableFormatsForDims(const Shape &di
 void Node::updateShapes() {
     IE_ASSERT(isDynamicNode()) << "Node::updateShapes() is called to a static shape node of type: " << getTypeStr() << " with name: " << getName();
     if (needShapeInfer()) {
+        PERF_SHAPE_INFER(this);
         auto result = shapeInfer();
         if (ShapeInferStatus::success == result.status) {
             redefineOutputMemory(result.dims);
```
maxnick, line 560: It does make sense to measure `redefineOutputMemory` call separatelly from `shapeInfer`. Also, it may be usefull to include `needShapeInfer()` into shape inference time measurements.

src/plugins/intel_cpu/src/node.cpp
```diff
@@ -565,6 +566,7 @@ void Node::updateDynamicParams() {
     IE_ASSERT(isDynamicNode()) << "Node::updateDynamicParams() is called to a static shape node of type: " << getTypeStr() << " with name: " << getName();
     if (isExecutable()) {
         if (needPrepareParams()) {
+            PERF_PREPARE_PARAMS(this);
```
maxnick, line 569: It may be useful to include `needPrepareParams` call into prepareParams time measurements.

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1224,14 +1226,15 @@ void Graph::InferDynamic(InferRequestBase* request) {
     } else {
         updateNodes.reset(new UpdateNodesSeq(executableGraphNodes));
     }
-    size_t inferCounter = 0;
 
+    CPU_DEBUG_CAP_ENABLE(const PerfKey perfKey = perfGetKey(*this));
+    size_t inferCounter = 0;
     for (auto stopIndx : syncIndsWorkSet) {
         updateNodes->run(stopIndx);
```
maxnick, line 1233: It will be good to have high level time counters, which measure time on this method level, so that we see the real time spent on nodes execution and nodes update. The former of course may be deduced from the node counters, but the latter is not that simple, since updating shapes and prepare params stages are run in parallel.

src/plugins/intel_cpu/src/perf_count.h
```diff
@@ -53,5 +52,10 @@ class PerfHelper {
...
 
-#define GET_PERF(_node) std::unique_ptr<PerfHelper>(new PerfHelper(_node->PerfCounter()))
-#define PERF(_node, _need) auto pc = _need ? GET_PERF(_node) : nullptr;
+#  define PERF(_node, _need, _itrKey) PERF_COUNTER(_need, PerfHelper, _node->PerfCounter())
+#  define PERF_SHAPE_INFER(_node)
+#  define PERF_PREPARE_PARAMS(_node)
+#endif // ifndef CPU_DEBUG_CAPS
+
+#define GET_PERF(_helper, ...) std::unique_ptr<_helper>(new _helper(__VA_ARGS__))
+#define PERF_COUNTER(_need, _helper, ...) auto pc = _need ? GET_PERF(_helper, __VA_ARGS__) : nullptr;
```
maxnick, line 61: Du we still need to allocate the counter on the heap. May be the `nullptr` may be replaced with an "Empty" object, that will reduce the amount of dynamic memory allocations/deallocations, which are not that cheap.

PR #15316, opened 2023-01-26
[PT FE] Add aten::unfold operator with layer test

sikorsl1:
### Details:
 - `aten::unfold` operator has been added

### Tickets:
 - 94310

### Comments:

 - Due to some overlaps in PyTorch operator names, it has to be clarified that this PR corresponds to the following PyTorch operator: https://pytorch.org/docs/stable/generated/torch.Tensor.unfold.html#torch.Tensor.unfold and NOT this one: https://pytorch.org/docs/stable/generated/torch.nn.functional.unfold.html#torch.nn.functional.unfold


tests/layer_tests/pytorch_tests/test_unfold.py
```diff
@@ -0,0 +1,40 @@
...
+
+import pytest
+from pytorch_layer_test_class import PytorchLayerTest
+import numpy as np
+import torch
+
+
+@pytest.mark.parametrize('dimension', (0, 1, 2))
+@pytest.mark.parametrize('size', (1, 2))
+@pytest.mark.parametrize('step', (1, 2))
```
rkazants, line 12: please test a case with `step=3,4`  as well. these values are also valid

tests/layer_tests/pytorch_tests/test_unfold.py
```diff
@@ -0,0 +1,40 @@
...
+                self.step = step
+
+            def forward(self, input_tensor):
+                return input_tensor.unfold(dimension=self.dimension, size=self.size, step=self.step)
+
+        ref_net = None
+
+        return aten_unfold(dimension, size, step), ref_net, "aten::unfold"
+
+    @pytest.mark.nightly
```
rkazants, line 36: please add one complex case to the precommit

tests/layer_tests/pytorch_tests/test_unfold.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+from pytorch_layer_test_class import PytorchLayerTest
+import numpy as np
+import torch
```
rkazants, line 7: please apply formatting to Python file

src/frontends/pytorch/src/op/unfold.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    auto dimension =
+        context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {context.const_input<int64_t>(1)}));
+    auto dimension_plus_1 = context.mark_node(std::make_shared<opset10::Add>(dimension, const_1_list));
+
+    int size_int = context.const_input<int64_t>(2);
+    auto size_scalar = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {size_int}));
+    auto size_list = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {size_int}));
+
+    auto step = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {context.const_input<int64_t>(3)}));
```
rkazants, line 34: please support dynamic cases of `size` and `step`. They can vary for models with dynamic shapes

src/frontends/pytorch/src/op/unfold.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    // calculate number of "slices"
+    auto sizedim = context.mark_node(std::make_shared<opset10::Gather>(sizes, dimension, const_0));
+    auto sizedim_minus_size = std::make_shared<opset10::Subtract>(sizedim, size_list);
+    auto fraction = std::make_shared<opset10::Divide>(sizedim_minus_size, step);
+    auto slices_count = std::make_shared<opset10::Add>(fraction, const_1);
+    auto slices_count_scalar = context.mark_node(std::make_shared<opset10::Reshape>(slices_count, const_1, false));
+
+    // generate indices for Gather
```
rkazants, line 43: it would be great to have more comments with example about what indices are generated.

tests/layer_tests/pytorch_tests/test_unfold.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@pytest.mark.parametrize('dimension', (0, 1, 2))
+@pytest.mark.parametrize('size', (1, 2))
+@pytest.mark.parametrize('step', (1, 2))
+@pytest.mark.parametrize('input_tensor', (np.random.randn(2, 2, 5).astype(np.float32),
+                                          np.random.randn(3, 3, 3, 3).astype(np.float32)))
```
rkazants, line 14: please try to use shapes with different dimensions like `(2,3,4,5)`

src/frontends/pytorch/src/op/unfold.cpp
```diff
@@ -0,0 +1,88 @@
...
+    auto const_0 = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {0}));
+    auto const_1 = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {1}));
+    auto const_0_list = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {0}));
+    auto const_1_list = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {1}));
+    auto const_neg_1_list = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {-1}));
+
+    // get inputs and prepare auxiliary nodes
+    auto input = context.get_input(0);
+    auto sizes = context.mark_node(std::make_shared<opset10::ShapeOf>(input));
+    auto input_rank = context.mark_node(std::make_shared<opset10::ShapeOf>(sizes));
```
rkazants, line 24: it is great to create PT common helper to compute tensor rank. Please consider it and re-use in other translators:)

src/frontends/pytorch/src/op/unfold.cpp
```diff
@@ -0,0 +1,88 @@
...
+    // constants
+    auto const_0 = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {0}));
+    auto const_1 = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {1}));
+    auto const_0_list = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {0}));
+    auto const_1_list = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {1}));
+    auto const_neg_1_list = context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {-1}));
+
+    // get inputs and prepare auxiliary nodes
+    auto input = context.get_input(0);
+    auto sizes = context.mark_node(std::make_shared<opset10::ShapeOf>(input));
```
rkazants, line 23: Now we have a confusing names between PyTorch specification for unfold and your names in the code. May be it is better to call not `sizes` and rather `input_shape`?

src/frontends/pytorch/src/op/unfold.cpp
```diff
@@ -0,0 +1,88 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
+
```
rkazants, line 7:
```suggestion

using namespace ov::opset10;

```

rkazants: Please consider to use

src/frontends/pytorch/src/op/unfold.cpp
```diff
@@ -0,0 +1,88 @@
...
+    auto input = context.get_input(0);
+    auto sizes = context.mark_node(std::make_shared<opset10::ShapeOf>(input));
+    auto input_rank = context.mark_node(std::make_shared<opset10::ShapeOf>(sizes));
+
+    auto dimension =
+        context.mark_node(opset10::Constant::create(element::i64, Shape{1}, {context.const_input<int64_t>(1)}));
+    auto dimension_plus_1 = context.mark_node(std::make_shared<opset10::Add>(dimension, const_1_list));
+
+    int size_int = context.const_input<int64_t>(2);
+    auto size_scalar = context.mark_node(opset10::Constant::create(element::i64, Shape{}, {size_int}));
```
rkazants, line 31:
```suggestion
    auto size_scalar = context.mark_node(Constant::create(element::i64, Shape{}, {size_int}));
```
can simply like this due to using namespace

tests/layer_tests/pytorch_tests/test_unfold.py
sikorsl1: done

tests/layer_tests/pytorch_tests/test_unfold.py
sikorsl1: done

tests/layer_tests/pytorch_tests/test_unfold.py
sikorsl1: done

src/frontends/pytorch/src/op/unfold.cpp
sikorsl1: done

src/frontends/pytorch/src/op/unfold.cpp
sikorsl1: done

tests/layer_tests/pytorch_tests/test_unfold.py
sikorsl1: done

src/frontends/pytorch/src/op/unfold.cpp
sikorsl1: done

src/frontends/pytorch/src/op/unfold.cpp
sikorsl1: done

src/frontends/pytorch/src/op/unfold.cpp
sikorsl1: I think that in general it is not a good idea - we usually use both shape and rank. Calculating rank in an additional sub-procedure will result in redundant shape calculation.

rkazants (approved):

src/frontends/pytorch/src/op/unfold.cpp
```diff
@@ -0,0 +1,85 @@
...
+    auto const_0_list = context.mark_node(Constant::create(element::i64, Shape{1}, {0}));
+    auto const_1_list = context.mark_node(Constant::create(element::i64, Shape{1}, {1}));
+    auto const_neg_1_list = context.mark_node(Constant::create(element::i64, Shape{1}, {-1}));
+
+    // get inputs and prepare auxiliary nodes
+    auto input = context.get_input(0);
+    auto input_shape = context.mark_node(std::make_shared<ShapeOf>(input));
+    auto input_rank = context.mark_node(std::make_shared<ShapeOf>(input_shape));
+
+    auto dimension = context.mark_node(Constant::create(element::i64, Shape{1}, {context.const_input<int64_t>(1)}));
```
mvafin, line 28:
```suggestion
    auto dimension = context.get_input(1);
```
Since you anyway create `Constant` node there is no reason to convert it to value first. The code I suggested will also work if 1st input is not a constant for example, also it reduces complexity. The only problem may be the shape of existing constant, it can be scalar, if so I still think the best idea would be to add `Unsqueeze` after.

src/frontends/pytorch/src/op/unfold.cpp
```diff
@@ -0,0 +1,85 @@
...
+    auto dimension_plus_1 = context.mark_node(std::make_shared<Add>(dimension, const_1_list));
+
+    auto size_scalar = context.mark_node(std::make_shared<Convert>(context.get_input(2), element::i64));
```
mvafin, line 31: It looks like there is no reason to cast to `i64`, correct me if I am wrong. All `i64` values here are created by you, `ShapeOf` can return `i32` values. Please try to avoid cast to `i64`.

src/frontends/pytorch/src/op/unfold.cpp
```diff
@@ -0,0 +1,85 @@
...
+    auto size_list = context.mark_node(std::make_shared<Reshape>(size_scalar, const_1_list, false));
+
+    auto step = context.mark_node(std::make_shared<Convert>(context.get_input(3), element::i64));
```
mvafin, line 34: Same comment

mvafin (changes requested):
It would be interesting to compare performance of this implementation vs the previous one with `Loop`, but this is not a blocker for this PR.

src/frontends/pytorch/src/op/unfold.cpp
sikorsl1: When get those inputs as "non constant inputs" I have to convert them to integer type, because otherwise it doesn't work. I think that this is approach is not a fault, because these parameters can only be integer type.

src/frontends/pytorch/src/op/unfold.cpp
sikorsl1: Thanks, done.

src/frontends/pytorch/src/op/unfold.cpp
sikorsl1: I introduced changes that you proposed.

mvafin (approved):

PR #15317, opened 2023-01-26
[GPU] Added shape agnostic optimized GEMM kernel

andrew-k-park:
### Details:
 - Shape agnostic optimized GEMM kernel implementation

### Tickets:
 - 99596

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/gemm.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/gemm.cpp
@@ -40,9 +40,7 @@ public:
             params.inputs.push_back(convert_data_tensor(input_layouts[i]));
         }
         params.outputs[0] = convert_data_tensor(output_layout);
-        if (impl_param.get_program().get_node(primitive->id).is_dynamic()) {
-            params.is_shape_agnostic = true;
-        }
+
         params.alpha = primitive->alpha;
         params.beta = primitive->beta;
         params.transpose_input0 = primitive->transpose_input0;
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/gemm.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/gemm.cpp
@@ -40,7 +40,9 @@ public:
             params.inputs.push_back(convert_data_tensor(input_layouts[i]));
         }
         params.outputs[0] = convert_data_tensor(output_layout);
-
+        if (impl_param.get_program().get_node(primitive->id).is_dynamic()) {
+            params.is_shape_agnostic = true;
+        }
         params.alpha = primitive->alpha;
         params.beta = primitive->beta;
         params.transpose_input0 = primitive->transpose_input0;
--- a/src/plugins/intel_gpu/src/graph/include/kernel_selector_helper.h
+++ b/src/plugins/intel_gpu/src/graph/include/kernel_selector_helper.h
@@ -201,6 +201,7 @@ kernel_selector::dim_tensor<T> convert_dim_vector(const tensor& t) {
 
 void convert_fused_ops_to_legacy_activations(const kernel_impl_params& param_info, std::vector<kernel_selector::base_activation_params>& activations);
 bool use_legacy_fused_ops(const kernel_impl_params& param_info);
+bool is_shape_agnostic(const kernel_impl_params& param_info);
 
 void set_params(const kernel_impl_params& param_info, kernel_selector::params& params);
 
@@ -215,6 +216,9 @@ inline params_t get_default_params(const kernel_impl_params& param_info) {
 
     params.inputs[0] = convert_data_tensor(input_layout);
     params.outputs[0] = convert_data_tensor(output_layout);
+    if (is_shape_agnostic(param_info)) {
+        params.is_shape_agnostic = true;
+    }
     params.layerID = param_info.desc->id;
 
     if (use_legacy_fused_ops(param_info)) {
--- a/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/gemm_tiled_opt.cl
+++ b/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/gemm_tiled_opt.cl
@@ -27,7 +27,7 @@
     #define BLOCK_WRITE_C(ptr, offset, data) BLOCK_WRITEN(OUTPUT_TYPE, 1, ptr, offset, data)
 #endif // TILE_N > SIMD_WIDTH
 
-inline uint FUNC(get_input0_batch_offset)(uint b, uint f, uint w, uint z) {
+inline uint FUNC(get_input0_batch_offset)(OPTIONAL_SHAPE_INFO_ARG uint b, uint f, uint w, uint z) {
 #if INPUT0_SIMPLE
     return GET_DATA_INDEX_6D_SAFE(INPUT0, b, f, w, z, 0, 0);
 #else // INPUT0_SIMPLE
@@ -35,7 +35,7 @@ inline uint FUNC(get_input0_batch_offset)(uint b, uint f, uint w, uint z) {
 #endif // INPUT0_SIMPLE
 }
 
-inline uint FUNC(get_input1_batch_offset)(uint b, uint f, uint w, uint z) {
+inline uint FUNC(get_input1_batch_offset)(OPTIONAL_SHAPE_INFO_ARG uint b, uint f, uint w, uint z) {
 #if INPUT1_SIMPLE
     return GET_DATA_INDEX_6D_SAFE(INPUT1, b, f, w, z, 0, 0);
 #else // INPUT1_SIMPLE
@@ -44,7 +44,7 @@ inline uint FUNC(get_input1_batch_offset)(uint b, uint f, uint w, uint z) {
 }
 
 #ifdef INPUT2_TYPE
-inline uint FUNC(get_input2_batch_offset)(uint b, uint f, uint w, uint z) {
+inline uint FUNC(get_input2_batch_offset)(OPTIONAL_SHAPE_INFO_ARG uint b, uint f, uint w, uint z) {
 #if INPUT2_SIMPLE
     return GET_DATA_INDEX_6D_SAFE(INPUT2, b, f, w, z, 0, 0);
 #else // INPUT2_SIMPLE
@@ -53,7 +53,7 @@ inline uint FUNC(get_input2_batch_offset)(uint b, uint f, uint w, uint z) {
 }
 #endif // INPUT2_TYPE
 
-inline uint FUNC(get_output_batch_offset)(uint b, uint f, uint w, uint z) {
+inline uint FUNC(get_output_batch_offset)(OPTIONAL_SHAPE_INFO_ARG uint b, uint f, uint w, uint z) {
 #if OUTPUT_SIMPLE
     return GET_DATA_INDEX_6D(OUTPUT, b, f, w, z, 0, 0);
 #else // OUTPUT_SIMPLE
@@ -65,6 +65,7 @@ inline uint FUNC(get_output_batch_offset)(uint b, uint f, uint w, uint z) {
 REQD_SUB_GROUP_SIZE(SIMD_WIDTH)
 __attribute__((reqd_work_group_size(SIMD_WIDTH, 1, 1)))
 KERNEL(gemm_tiled_opt)(
+    OPTIONAL_SHAPE_INFO_ARG
     const __global INPUT0_TYPE* input0,
     const __global INPUT1_TYPE* input1,
 #ifdef INPUT2_TYPE
@@ -86,18 +87,26 @@ KERNEL(gemm_tiled_opt)(
 
     // Setting x and y for fusings indexing
     // TODO: investigate how we can use only TILE_N_NOT_DIVISIBLE here for getting stable results in fusings
+#if IS_DYNAMIC
+    const uint x = (uint)get_global_id(0);
+#else // IS_DYNAMIC
 #if TILE_N_NOT_DIVISIBLE || B_VEC_SIZE == 1
     const uint x = (uint)get_global_id(0);
 #else // TILE_N_NOT_DIVISIBLE || B_VEC_SIZE == 1
     const uint x = tile_n_num * SIMD_WIDTH * B_VEC_SIZE;
 #endif // TILE_N_NOT_DIVISIBLE || B_VEC_SIZE == 1
+#endif // IS_DYNAMIC
     uint y = tile_m_offset;
 
+#if IS_DYNAMIC
+    const uint tile_m_iterations = TILE_M_NOT_DIVISIBLE ? (tile_m_num == (tile_m_size - 1) ? TILE_M_LEFTOVER : TILE_M) : TILE_M;
+#else // IS_DYNAMIC
 #if TILE_M_NOT_DIVISIBLE
     const uint tile_m_iterations = tile_m_num == (tile_m_size - 1) ? TILE_M_LEFTOVER : TILE_M;
 #else // TILE_M_NOT_DIVISIBLE
     const uint tile_m_iterations = TILE_M;
 #endif // TILE_M_NOT_DIVISIBLE
+#endif // IS_DYNAMIC
 
     const uint z = batch_number % OUTPUT_SIZE_Z;
     batch_number /= OUTPUT_SIZE_Z;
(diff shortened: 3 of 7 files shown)


src/plugins/intel_gpu/src/kernel_selector/kernels/gemm/gemm_kernel_tiled_opt.cpp
```diff
@@ -53,25 +55,34 @@ GemmKernelTiledOpt::GemmTuningData GemmKernelTiledOpt::SetTuningParams(const gem
 
     GemmKernelTiledOpt::GemmTuningData tuning_data;
 
-    auto m_size = output.Y().v;
-    auto n_size = output.X().v;
-    auto k_size = params.transpose_input0 ? params.inputs[0].Y().v : params.inputs[0].X().v;
+     if (!params.has_dynamic_tensors()) {
+        auto m_size = output.Y().v;
```
sshlyapn, line 59: During network execution, after update_shape() function call, params will not contain dynamic tensors, so even for shape agnostic dynamic kernel tuning will be performed in this if..else branch. Tuning function may change tiles sizes which will lead to incorrect workgroup sizes configuration in SetDefault() function after update_dispatch_data_func() call. It looks like you need to have some information if SetTuningParams call performed for dynamic kernel or fully static, and if it is dynamic case then always return tiles sizes aligned with simd_size=16, isn't it?

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/gemm_tiled_opt.cl
```diff
@@ -86,18 +87,26 @@ KERNEL(gemm_tiled_opt)(
...
 #if TILE_N_NOT_DIVISIBLE || B_VEC_SIZE == 1
     const uint x = (uint)get_global_id(0);
 #else // TILE_N_NOT_DIVISIBLE || B_VEC_SIZE == 1
     const uint x = tile_n_num * SIMD_WIDTH * B_VEC_SIZE;
 #endif // TILE_N_NOT_DIVISIBLE || B_VEC_SIZE == 1
+#endif // IS_DYNAMIC
     uint y = tile_m_offset;
 
+#if IS_DYNAMIC
+    const uint tile_m_iterations = TILE_M_NOT_DIVISIBLE ? (tile_m_num == (tile_m_size - 1) ? TILE_M_LEFTOVER : TILE_M) : TILE_M;
```
vladimir-paramuzov, line 102: Can we have this version for all cases (including static ones)? I'd expect that the compiler will emit the same code for
```
const uint tile_m_iterations = TILE_M_NOT_DIVISIBLE ? (tile_m_num == (tile_m_size - 1) ? TILE_M_LEFTOVER : TILE_M) : TILE_M;
```

and 

```
#if TILE_M_NOT_DIVISIBLE
    const uint tile_m_iterations = tile_m_num == (tile_m_size - 1) ? TILE_M_LEFTOVER : TILE_M;
#else // TILE_M_NOT_DIVISIBLE
    const uint tile_m_iterations = TILE_M;
#endif // TILE_M_NOT_DIVISIBLE
```

when `TILE_M_NOT_DIVISIBLE` is compile-time constant

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/gemm_tiled_opt.cl
```diff
@@ -281,17 +316,43 @@ KERNEL(gemm_tiled_opt)(
         }
     } // Loading leftovers of the matrix A and tile C calculation end
 #endif // TILE_K_NOT_DIVISIBLE
+#endif // IS_DYNAMIC
 
 #if HAS_FUSED_OPS && FUSED_OPS_CAN_USE_PRELOAD
+#if IS_DYNAMIC
+    FUSED_OPS_PRELOAD_SCALAR;
+#else // IS_DYNAMIC
 #if TILE_N_NOT_DIVISIBLE || B_VEC_SIZE == 1
```
vladimir-paramuzov, line 325: || IS_DYNAMIC

src/plugins/intel_gpu/src/kernel_selector/kernels/gemm/gemm_kernel_tiled_opt.cpp
```diff
@@ -86,33 +97,62 @@ JitConstants GemmKernelTiledOpt::GetJitConstants(const gemm_params& params) cons
...
+    if (params.has_dynamic_tensors()) {
+        auto m_size = toCodeString(output.Y(), 16);
+        auto n_size = toCodeString(output.X(), 17);
+        auto k_size = params.transpose_input0 ? toCodeString(params.inputs[0].Y(), 4) : toCodeString(params.inputs[0].X(), 5);
+        const std::string leftover_m = "(" + m_size + "%" + std::to_string(tuning_data.tile_m_size) + ")";
+        const std::string leftover_n = "(" + k_size + "%" + std::to_string(tuning_data.tile_n_size) + ")";
+        const std::string leftover_k = "(" + k_size + "%" + std::to_string(tuning_data.tile_k_size) + ")";
+        const std::string not_devisible_m = "(" + leftover_m + "!=0)";
+        const std::string not_devisible_n = "(" + leftover_n + "!=0)";
+        const std::string not_devisible_k = "(" + leftover_k + "!=0)";
```
vladimir-paramuzov, line 112: nit: not_d**i**visible

andrew-k-park: done

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/gemm_tiled_opt.cl
```diff
@@ -86,18 +87,26 @@ KERNEL(gemm_tiled_opt)(
 
     // Setting x and y for fusings indexing
     // TODO: investigate how we can use only TILE_N_NOT_DIVISIBLE here for getting stable results in fusings
+#if IS_DYNAMIC
+    const uint x = (uint)get_global_id(0);
+#else // IS_DYNAMIC
 #if TILE_N_NOT_DIVISIBLE || B_VEC_SIZE == 1
```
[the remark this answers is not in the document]
andrew-k-park: In shape agnostic kernel case, `TILE_N_NOT_DIVISIBLE` is defined as constant as follows:
```
#define TILE_N_NOT_DIVISIBLE ((shape_info[17]%16)!=0)
```
So, this causes build error like:
```
2:5377:5: error: token is not a valid binary operator in a preprocessor subexpression
#if TILE_N_NOT_DIVISIBLE || B_VEC_SIZE == 1 || IS_DYNAMIC
    ^~~~~~~~~~~~~~~~~~~~
```
Therefore, this code is separated into dynamic and static cases.

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/gemm_tiled_opt.cl
andrew-k-park: For the same reason as above, this code was separated.

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/gemm_tiled_opt.cl
andrew-k-park: For the same reason as above, this code was separated.

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/gemm_tiled_opt.cl
andrew-k-park: I understand your suggestion. removed unnecessarily redundant code as suggested.

vladimir-paramuzov: Thanks. But please check the performance in static case to ensure that compiler emits the same code.

src/plugins/intel_gpu/src/kernel_selector/kernels/gemm/gemm_kernel_tiled_opt.cpp
```diff
@@ -184,7 +224,7 @@ bool GemmKernelTiledOpt::Validate(const Params& params, const optional_params& o
     const auto& gmm_params = static_cast<const gemm_params&>(params);
     bool gemm_leftovers = gmm_params.inputs[0].X().v % 16 || gmm_params.inputs[0].Y().v % 16 ||
                           gmm_params.inputs[1].X().v % 16 || gmm_params.inputs[1].Y().v % 16;
-    if ((gmm_params.transpose_input0 || gmm_params.transpose_input1) && gemm_leftovers)
+    if ((gmm_params.transpose_input0 || gmm_params.transpose_input1) && (gemm_leftovers || gmm_params.has_dynamic_inputs()))
```
yeonbok, line 227: Is there any fundamental restriction for the dynamic impl for transposed cases or just to leave it to be done in the future?

src/plugins/intel_gpu/tests/test_cases/gemm_gpu_test.cpp
```diff
@@ -342,6 +342,168 @@ TEST(gemm_gpu, dynamic) {
...
+    config.set_property(ov::intel_gpu::allow_new_shape_infer(true));
+    network network(engine, topology, config);
+
+    {
+        network.set_input_data("input1", input1_1);
+        network.set_input_data("input2", input2);
+
+        auto outputs = network.execute();
+        ASSERT_EQ(outputs.size(), size_t(1));
+        ASSERT_EQ(outputs.begin()->first, "gemm");
```
yeonbok, line 478: Could you add a check whether the gemm_tiled_opt kernel is selected ?

src/plugins/intel_gpu/src/kernel_selector/kernels/gemm/gemm_kernel_tiled_opt.cpp
yeonbok: Yeah, we can add a common kernel param of is_shape_agnostic ?

ahnyoung-paul:
If we create shape agnostic optimized GEMM kernel, is it necessary to compile GEMM optimized kernel on async compilation in update_impl?

ahnyoung-paul (commented):
If shape agnostic optimized GEMM kernel is provided and it has good performance , why don't you disable async compilation for GEMM optimized kernel in update_impl?

ahnyoung-paul:
> If shape agnostic optimized GEMM kernel is provided and it has good performance , why don't you disable async compilation for GEMM optimized kernel in update_impl?

Through an offline discussion with @yeonbok , I understood it will be applied next time to disable gemm in async compilation. thanks.

src/plugins/intel_gpu/tests/test_cases/gemm_gpu_test.cpp
andrew-k-park: done

src/plugins/intel_gpu/src/kernel_selector/kernels/gemm/gemm_kernel_tiled_opt.cpp
andrew-k-park: Yes. If gmm_params has dynamic inputs, leftovers cannot be calculated. Added corresponding comment.

src/plugins/intel_gpu/src/kernel_selector/kernels/gemm/gemm_kernel_tiled_opt.cpp
andrew-k-park: Updated to add `is_shape_agnostic` to `base_params`

ahnyoung-paul (approved):
LGTM

src/plugins/intel_gpu/src/graph/impls/ocl/gemm.cpp
```diff
@@ -40,7 +40,9 @@ struct gemm_impl : typed_primitive_impl_ocl<gemm> {
             params.inputs.push_back(convert_data_tensor(input_layouts[i]));
         }
         params.outputs[0] = convert_data_tensor(output_layout);
-
+        if (impl_param.get_program().get_node(primitive->id).is_dynamic()) {
+            params.is_shape_agnostic = true;
```
vladimir-paramuzov, line 44: Why this flag is needed? You can always get the value inside kernel selector by checking if any of inputs/outputs tensor is dynamic

vladimir-paramuzov (changes requested):

src/plugins/intel_gpu/src/graph/impls/ocl/gemm.cpp
andrew-k-park: In case of gemm_tiled_opt shape agnotic kernel, when `update_dispatch_data_func` is called, it is necessary to check whether the current selected kernel is static kernel or shape agnostic kernel in `SetTuningParams` of kernel selector. However, since the inputs/outputs tensor is already static, it cannot be checked. Therefore, additional flag is needed to determine this.

vladimir-paramuzov: What I don't like in current impl is that you've added is_shape_agnostic flag into common params, but fill it properly only for gemm primitive. So I see 2 possible options to solve it:
1. Determine if kernel is shape agnostic in gemm kernel by checking if tensors are dynamic, and then just capture the flag value in `update_dispatch_data_func`
2. Move initialization of this flag to `get_default_params`

The second option can be used if multiple primitives may require similar flags, otherwise the first option is preferrable.

src/plugins/intel_gpu/tests/test_cases/gemm_gpu_test.cpp
```diff
@@ -342,6 +342,178 @@ TEST(gemm_gpu, dynamic) {
     }
 }
 
+TEST(gemm_gpu, dynamic_multi_inference_same_shape) {
+    auto& engine = get_test_engine();
+
+    auto in1_dyn_layout = layout{ ov::PartialShape{ 1, 1, ov::Dimension(1, 10), 4 }, data_types::f32, format::bfyx };
+    auto in1_actual_layout = layout{ ov::PartialShape{ 1, 1, 3, 4 }, data_types::f32, format::bfyx };
+    auto in2_layout = layout{ ov::PartialShape{ 4, 1 }, data_types::f32, format::bfyx };
```
vladimir-paramuzov, line 350: it's better to use dynamic shape for second input as well to ensure that it's handled properly in the kernel

src/plugins/intel_gpu/tests/test_cases/gemm_gpu_test.cpp
```diff
@@ -342,6 +342,178 @@ TEST(gemm_gpu, dynamic) {
...
+        network.set_input_data("input2", input2);
+
+        auto outputs = network.execute();
+        ASSERT_EQ(outputs.size(), size_t(1));
+        ASSERT_EQ(outputs.begin()->first, "gemm");
+
+        auto prog = network.get_program();
+        auto& node = prog->get_node("gemm");
+        auto kernel_name = node.get_selected_impl()->get_kernel_name();
+        ASSERT_EQ(kernel_name, "gemm_tiled_opt");
```
vladimir-paramuzov, line 401: I'd suggest changing this check to something like:
```
    auto impl = node.get_selected_impl();
    ASSERT_TRUE(impl != nullptr);
    ASSERT_TRUE(impl->is_dynamic());
```

vladimir-paramuzov:
@andrew-k-park could you also check accuracy on some dynamic transformer? I did a quick experiment and it seems that this patch leads to accuracy drop.

andrew-k-park:
>

andrew-k-park:
> @andrew-k-park could you also check accuracy on some dynamic transformer? I did a quick experiment and it seems that this patch leads to accuracy drop.

Yes, I also confirmed that there was dynamic shape accuracy drop, and found the root cause and fixed. Currently F1: 91.81%, EM: 85.25% @bert-small-uncased-whole-word-masking-squad-0001 w/ SQuAD v1.1

src/plugins/intel_gpu/src/graph/impls/ocl/gemm.cpp
yeonbok: Yes we can move it to common param

src/plugins/intel_gpu/tests/test_cases/gemm_gpu_test.cpp
andrew-k-park: done

src/plugins/intel_gpu/tests/test_cases/gemm_gpu_test.cpp
andrew-k-park: done

src/plugins/intel_gpu/src/graph/impls/ocl/gemm.cpp
andrew-k-park: Because maybe FC is also expected to require the same flag, it's updated as second option.

src/plugins/intel_gpu/src/graph/include/kernel_selector_helper.h
```diff
@@ -215,6 +216,9 @@ inline params_t get_default_params(const kernel_impl_params& param_info) {
 
     params.inputs[0] = convert_data_tensor(input_layout);
     params.outputs[0] = convert_data_tensor(output_layout);
+    if (use_shape_agnostic(param_info)) {
```
yeonbok, line 219: It is really minor.. but the kernel "is" shape agnostic, not "use".. could you please rename?

andrew-k-park: done

yeonbok (approved):

vladimir-paramuzov (approved):

PR #15325, opened 2023-01-26
Don't use could_propagate in MarkDequantizationSubgraph

mateusztabaka:


src/common/transformations/src/transformations/low_precision/mark_dequantization_subgraph.cpp
```diff
@@ -10,9 +10,24 @@
...
+    if (!node->has_evaluate() || ov::constant_folding_is_disabled(node))
+        return false;
+
+    for (size_t i = 0; i < num_inputs; i++) {
+        const auto& input = node->get_input_node_shared_ptr(i);
+        if (!is_constfoldable(input))
+            return false;
+    }
+    return true;
+}
```
eshoguli, line 30: Can you, please, copy exists and well tested not recursive `ngraph::could_propagate` function and just remove `ShapeOf` logic, please?

eshoguli (commented):
There is comment. I got agreement with nGraph team, that we will have similar function like `ngraph::could_propagate` which will be moved to `utils` later by nGraph team later.

src/common/transformations/src/transformations/low_precision/mark_dequantization_subgraph.cpp
mateusztabaka: done

eshoguli (approved):
LGTM

PR #15339, opened 2023-01-26
[PT FE]: Add aten::view transformations

mmikolajcz:
### Details:
 - *Refactor view translation into transformation*
 - *Collect runtime info*
 - *Fix issue with translation of aten::Int returning i64 instead of i32, causing issues with cat on view input*
 - *Add more test cases*

### Tickets:
 - *99230*


src/frontends/pytorch/src/transforms/aten_view_replacer.cpp
```diff
@@ -0,0 +1,78 @@
...
+                auto unsqueeze = std::make_shared<opset10::Unsqueeze>(input.get_source_output(), axis_0);
+                inputs.push_back(unsqueeze);
+            }
+            auto concat = std::make_shared<opset10::Concat>(inputs, 0);
+            auto reshape = std::make_shared<opset10::Reshape>(view_op->get_input_source_output(0), concat, false);
+            copy_runtime_info({view_op, shape_node}, reshape);
+            replace_node(view_op, reshape);
+            return true;
+        };
+        if (auto constant_node = std::dynamic_pointer_cast<opset10::Constant>(shape_node)) {
```
mmikolajcz, line 42: Instead checking every translated node known to possibly return Tensor being PyTorch list, maybe handle it in more generic way?

src/frontends/pytorch/src/op/int.cpp
```diff
@@ -12,10 +12,10 @@ namespace pytorch {
 namespace op {
 
 OutputVector translate_int(NodeContext& context) {
-    return {context.mark_node(std::make_shared<opset10::Convert>(context.get_input(0), element::i64))};
+    return {context.mark_node(std::make_shared<opset10::Convert>(context.get_input(0), element::i32))};
```
mmikolajcz, line 15: Alternative to it for view would be to `Convert` all inputs to `prim::ListConstruct` into common precision.

mvafin:
I don't think I like this approach. And my initial comment in `translate_view` didn't mean to remove translation of `view` and do it on transformation stage.
In this PR we basically add copy-paste code for `Constant`, `ShapeOf` etc cases, while we could leave translation for `view` where we do `view`->`Reshape` conversion, but only the case when second input is `ListConstruct` do a transformation, where we replace pattern: `ListConstruct`->`Reshape` with `Concat`->`Reshape`.

mmikolajcz:
Thanks @mvafin, modified transformation as you suggested to only look for prim::ListConstruct + ov::Reshape combination and restored previous translation with ListConstruct part removed. Also applied changes to translation of aten::reshape due to similarity

src/frontends/pytorch/src/op/reshape.cpp
```diff
@@ -13,25 +13,8 @@ namespace pytorch {
 namespace op {
 
 OutputVector translate_reshape(NodeContext& context) {
```
mvafin, line 15: Please add comment here that view has the same semantics. Also would be very nice to have schema in comments for both reshape and view

src/frontends/pytorch/src/transforms/listconstruct_reshape_replacer.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "openvino/pass/pattern/matcher.hpp"
+#include "openvino/pass/pattern/op/wrap_type.hpp"
+#include "utils.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace pass {
+
+ListConstructReshapeReplacer::ListConstructReshapeReplacer() {
```
mvafin, line 18: Could you add comments for this transformation. What it does what ops can become Reshape, etc

mvafin (commented):
Looks good, please add comments

src/frontends/pytorch/src/transforms/listconstruct_reshape_replacer.cpp
mmikolajcz: Done

src/frontends/pytorch/src/op/reshape.cpp
mmikolajcz: Done

mvafin (approved):

PR #15343, opened 2023-01-26
[PyOV] Simplify requirement files

p-wysocki:
### Details:
 - Add `constraints.txt` files which allow to keep dependency versions in a single place: one in `openvino/tools`, one in `openvino/src/bindings/python` and one in `openvino/tests`
 - Overriding values in `constraints.txt` causes pip conflicts, so some packages cannot be there, for example `numpy` or `onnx` - they have different versions in different components (for example MO allows for a range of `onnx` versions while ONNX FE does not). This is a direct cause of some dependency versions not being present in `constraints.txt` files, but still in `requirements.txt`.
 - `setup.py` files do not work with `-c` flag for `pip`, so a custom requirement parser has been written
 - This change should result in dependencies being easier to maintain and upgrade
 - Modify dependabot configuration

### Tickets:
 - 98876


tools/mo/requirements.txt
```diff
@@ -1,6 +1,8 @@
-numpy>=1.16.6
-networkx~=2.5; python_version <= "3.6"
-networkx<=2.8.8; python_version > "3.6"
-defusedxml>=0.7.1
-importlib-metadata; python_version < "3.8" and sys_platform == 'win32'
-openvino-telemetry>=2022.1.0
\ No newline at end of file
+-c ../../src/bindings/python/constraints.txt
+numpy>=1.16.6,<1.25.0
```
rkazants, line 2: disagree with this change:
1. please remove upper-bound
2. why do we need the constrain files here? This requirements file is going to the release package and now we have relative path that can be invalidated in the release package

rkazants (changes requested):

tools/mo/requirements_mxnet.txt
```diff
@@ -1,9 +1,4 @@
...
-networkx<=2.8.8; python_version > "3.6"
-numpy>=1.16.6,<1.25.0
-defusedxml>=0.7.1
-urllib3>=1.26.4
-requests>=2.25.1
-fastjsonschema~=2.15.1
+-c ../../src/bindings/python/constraints.txt
+-r ./requirements.txt
+mxnet
+urllib3
```
rkazants, line 4: mxnet, urllib3 should have versions here

tools/mo/requirements.txt
```diff
@@ -1,6 +1,8 @@
...
+networkx
+requests
+fastjsonschema
```
rkazants, line 5: please remove `fastjsonschema` because this is optional module and install only in case the validated version

tools/mo/requirements.txt
```diff
@@ -1,6 +1,8 @@
-numpy>=1.16.6
-networkx~=2.5; python_version <= "3.6"
-networkx<=2.8.8; python_version > "3.6"
-defusedxml>=0.7.1
-importlib-metadata; python_version < "3.8" and sys_platform == 'win32'
-openvino-telemetry>=2022.1.0
\ No newline at end of file
+-c ../../src/bindings/python/constraints.txt
+numpy>=1.16.6,<1.25.0
+networkx
```
rkazants, line 3: risky change to have `networkx` without upper-bound. Please recover it.

tools/mo/requirements_onnx.txt
```diff
@@ -1,8 +1,4 @@
+-c ../../src/bindings/python/constraints.txt
+-r ./requirements.txt
```
rkazants, line 2: I think you need to create separate `requirements_extra.txt` file to move all overlappings among extra files.

tools/mo/requirements.txt
jiwaszki: `Add constraints.txt which allows to keep dependency versions in a single place` - FYI it is in the description. It preserves upper-bound this way.

tools/mo/requirements_mxnet.txt
jiwaszki: `Add constraints.txt which allows to keep dependency versions in a single place` - FYI

tools/mo/requirements_onnx.txt
p-wysocki: Currently all the overlapping requirements are contained in `requirements.txt`, which is linked here as `-r ./requirements.txt`. Running `pip install -r requirements_onnx.txt` will also install everything from the linked `requirements.txt`.

tools/mo/requirements.txt
p-wysocki: I moved the version ('fastjsonschema~=2.15.1') into `constraints.txt`. If the version is specified (it was not changed in this PR) is this okay? Or should I remove it even if version is specified?

tools/mo/requirements.txt
p-wysocki: > 1. please remove upper-bound for NumPy

All of the requirement files in `tools/mo` have upper-bound requirement, except for `requirements.txt`. Approach proposed in this PR allows us to specify one numpy version requirement for MO, but if you wish I can move it back to the previous state (where no upper-bound is present only in `requirements.txt`.

> 2. why do we need the constrain files here? This requirements file is going to the release package and now we have relative path that can be invalidated in the release package

That's a good point, I'm currently testing the `openvino_dev` wheel.

rkazants: Please remove upper-bound in `requirements.txt` because it corresponds to the default installation. We agreed to have the default installation preserving user environment without overriding NumPy version

tools/mo/requirements_onnx.txt
rkazants: They are anyway different because extra installation requires to have fixed upper-bound version, in the mean time. the default installation does not have it. So we cannot use `requirement.txt` file for collecting such overlapping items.

.github/github_org_control/requirements.txt
```diff
@@ -1,2 +1,3 @@
-PyGithub==1.55
-ldap3==2.7
+-c ../../src/bindings/python/constraints.txt
```
ilya-lavrenov, line 1: do you really think we need to have repository-level constraints? 
Some files like `.github/github_org_control/requirements.txt` are used in completely different environment rather than product and I don't think we need to put everything under a single umbrella

tools/mo/requirements.txt
p-wysocki: I understand now. Upper bound has been removed.

tools/mo/requirements_onnx.txt
p-wysocki: This makes sense now. I reverted the changes.

.github/github_org_control/requirements.txt
p-wysocki: That's a good point, I guess I got lost in the dream of having everything in one place. I removed the requirement files which do not belong to this PR.

tools/openvino_dev/setup.py
```diff
@@ -214,7 +272,7 @@ def concat_files(output_file, input_files):
                 outfile.write(content)
     return output_file
 
-
+install_reqs = read_requirements(SCRIPT_DIR / 'requirements.txt')
```
jiwaszki, line 275: Nit: move it to `setup`.

jiwaszki (approved):

tools/openvino_dev/setup.py
p-wysocki: Done.

tools/mo/requirements_onnx.txt
```diff
@@ -1,8 +1,8 @@
...
-defusedxml>=0.7.1
-requests>=2.25.1
-fastjsonschema>=2.15.1,<2.17
-protobuf>=3.18.1,<4.0.0
+onnx
+networkx
+defusedxml
+requests
+fastjsonschema
+protobuf
```
rkazants, line 8: Some tests directly install these files like `pip install -r requirements_onnx.txt` that means they will ignore versions?
Please check this one, for example: https://github.com/openvinotoolkit/openvino/blob/master/.ci/azure/linux.yml#L165

p-wysocki: Pip understands `-c` flag, so it will use versions from `openvino/tools/constraints.txt`. It can be seen in latest `openvino-lin` job at https://dev.azure.com/openvinoci/dldt/_build/results?buildId=481344&view=logs&j=982848f2-7f8d-5e57-730a-8df1e5f991ba&t=83ea7ca3-25f9-5f6d-cb63-d032d85612bf, in the "Install dependencies" step that the versions installed are correct.

tools/mo/requirements_tf.txt
```diff
@@ -1,7 +1,7 @@
-numpy>=1.16.6,<1.25.0
+-c ../constraints.txt
 tensorflow>=1.15.5,<=2.11.0
```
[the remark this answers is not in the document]
rkazants: they vary in tf and tf2 files

rkazants (changes requested):
Please ping me to review once CI gets green

src/bindings/python/constraints.txt
```diff
@@ -0,0 +1,39 @@
+# used in multiple components
```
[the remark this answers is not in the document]
p-wysocki: If you're asking about CI pipelines then no, as far as I know they all use `pip install -r requirements.txt`, and `-c` is a flag natively supported by `pip`.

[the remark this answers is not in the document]
ilya-lavrenov: asking about https://github.com/openvinotoolkit/openvino/blob/master/.github/dependabot.yml

ceciliapeng2011 (approved):
LGTM from PaddleFE's perspective

cmake/developer_package/python_requirements.cmake
```diff
@@ -97,7 +97,10 @@ function(ov_check_pip_packages)
 
     if(PYTHONINTERP_FOUND)
         execute_process(
-            COMMAND ${PYTHON_EXECUTABLE} -c "import pkg_resources ; pkg_resources.require(open('${ARG_REQUIREMENTS_FILE}', mode='r'))"
+            COMMAND ${PYTHON_EXECUTABLE} -c "import pkg_resources ;
+            requirements_raw = open('${ARG_REQUIREMENTS_FILE}', mode='r').readlines() ;
+            requirements = [package for package in requirements_raw if not package.startswith('-c')] ;
+            pkg_resources.require(requirements)"
```
ilya-lavrenov, line 103: I suppose we need concrete versions here

tools/mo/requirements_onnx.txt
mlukasze: can we mark this conversation as solved and move forward? :)

rkazants: there are a lot of failires in CI for this PR. Propose to anticipate fixing it and we will resolve it:)

cmake/developer_package/python_requirements.cmake
p-wysocki: I just pushed a commit with concrete requirement for just `onnx=1.12`, it's far from perfect but I want to see if it fixes the issues.

cmake/developer_package/python_requirements.cmake
```diff
@@ -97,8 +97,46 @@ function(ov_check_pip_packages)
...
+        else:
+            constraint = constraints.get(line) ;
+            if constraint:
+                for marker in constraint: 
+                    requirements.append(line+marker) ;
+            else:
+                requirements.append(line) ;
+else:
+    requirements = raw_requirements ;
+pkg_resources.require(requirements)"
```
ilya-lavrenov, line 139: we can put it into python script inside this folder and call inside the `execute_process`

praasz (approved):

src/bindings/python/constraints.txt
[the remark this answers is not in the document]
p-wysocki: Good point, I updated dependabot config.

cmake/developer_package/python_requirements.cmake
p-wysocki: I tried it and there were issues due to cmake being called from different directories, which results in the Python script not being found due to relative paths. I'll get back to it once the CI gets green.

ilya-lavrenov: we need to use path relative to folder with IEDevScripts.cmake
Example https://github.com/openvinotoolkit/openvino/blob/73bedced873d75d509f06d237e70ba7cfbb52208/cmake/developer_package/api_validator/api_validator.cmake#L169

cmake/developer_package/python_requirements.cmake
p-wysocki: The current parser now works correctly - just like before, but while taking `constraints.txt` into account.

cmake/developer_package/python_requirements.cmake
p-wysocki: Thank you, that helped. I moved the script into a separate `.py` file.

ilya-lavrenov (approved):

.github/dependabot.yml
```diff
@@ -6,7 +6,7 @@ updates:
   # Python product dependencies
   #
```
ilya-lavrenov, line 8: do we need to add a constraints file for tests as well?

tests/samples_tests/CMakeLists.txt
```diff
@@ -6,4 +6,11 @@ cmake_minimum_required(VERSION 3.13)
 
 project(samples_tests)
 
-install(DIRECTORY smoke_tests/ DESTINATION tests/smoke_tests COMPONENT tests EXCLUDE_FROM_ALL)
+install(DIRECTORY smoke_tests/ DESTINATION tests/smoke_tests COMPONENT tests EXCLUDE_FROM_ALL PATTERN "requirements.txt" EXCLUDE)
+install(FILES ../constraints.txt DESTINATION tests COMPONENT tests EXCLUDE_FROM_ALL)
+
+file(READ smoke_tests/requirements.txt REQUIREMENTS_REPO)
+string(REPLACE "-c ../../constraints.txt" "-c ../constraints.txt" REQUIREMENTS_TMP ${REQUIREMENTS_REPO})
+file(WRITE ./requirements.txt ${REQUIREMENTS_TMP})
```
ilya-lavrenov, line 14: it's bad practice to write inside source folder.
Use `CMAKE_CURRENT_BINARY_DIR` to replicate current sub-folder to build tree (binary tree)

tools/mo/requirements.txt
```diff
@@ -1,6 +1,7 @@
+-c ../constraints.txt
 numpy>=1.16.6
-networkx~=2.5; python_version <= "3.6"
-networkx<=2.8.8; python_version > "3.6"
-defusedxml>=0.7.1
-importlib-metadata; python_version < "3.8" and sys_platform == 'win32'
-openvino-telemetry>=2022.1.0
\ No newline at end of file
+singledispatchmethod
```
rkazants, line 3: please remove - not needed for production

rkazants (changes requested):

src/frontends/tensorflow_lite/tests/requirements.txt
```diff
@@ -1,2 +1,3 @@
+-c ../../../bindings/python/constraints.txt
 numpy>=1.16.6,<1.25.0
```
rkazants, line 2: why it still remained with versions and does not use constaints

.github/dependabot.yml
p-wysocki: Good point, added.

tests/samples_tests/CMakeLists.txt
p-wysocki: Done, please take a look if it's alright now.

tools/mo/requirements.txt
p-wysocki: I reverted it because it was a debug commit - however, @jiwaszki do we need `singledispatchmethod` in `tools`?

src/frontends/tensorflow_lite/tests/requirements.txt
p-wysocki: Because `numpy` versions differ between different `requirements.txt` files under the scope of the `constraints.txt`. This is sadly a limitation of pip `-c` flag, there are more details in the PR description.

p-wysocki: I believe we can align it everywhere, now that `onnx==1.13.1` has been merged into master. However, this PR already does like five different things and we need to merge it soon. I would prefer doing the `numpy` alignment in a separate PR in order to save time.

tools/mo/requirements.txt
p-wysocki: Agreed with @jiwaszki offline to consider doing it in a separate PR. The package is currently removed, as you @rkazants requested.

src/frontends/tensorflow_lite/tests/requirements.txt
rkazants: @p-wysocki, please create a ticket to fix it this release.
@mlukasze, fyi.

mlukasze: I'm aware, but thanks for mentioning anyway :)

ilya-lavrenov:
testd via product-config. The CI was green

src/frontends/tensorflow_lite/tests/requirements.txt
p-wysocki: @rkazants Done in https://github.com/openvinotoolkit/openvino/pull/16652, let's wait for green CI with review.

PR #15346, opened 2023-01-26
Snippets: linker issue fix in tests

eshoguli:
### Details:
 - *Linker issue fix in snippets tests. There are two TransposeMatMul tests in the same namespace ov::test::snippets*

### DIFF
--- a/src/tests/functional/plugin/shared/include/snippets/matmul.hpp
+++ b/src/tests/functional/plugin/shared/include/snippets/matmul.hpp
@@ -56,15 +56,6 @@ protected:
     void SetUp() override;
 };
 
-class TransposeMatMul : public testing::WithParamInterface<ov::test::snippets::TransposeMatMulParams>,
-                        virtual public ov::test::SnippetsTestsCommon {
-public:
-    static std::string getTestCaseName(testing::TestParamInfo<ov::test::snippets::TransposeMatMulParams> obj);
-
-protected:
-    void SetUp() override;
-};
-
 } // namespace snippets
 } // namespace test
 } // namespace ov
--- a/src/tests/functional/plugin/shared/src/snippets/matmul.cpp
+++ b/src/tests/functional/plugin/shared/src/snippets/matmul.cpp
@@ -98,41 +98,6 @@ void ExplicitTransposeMulMatMulBias::SetUp() {
     }
 }
 
-std::string TransposeMatMul::getTestCaseName(testing::TestParamInfo<ov::test::snippets::TransposeMatMulParams> obj) {
-    std::vector<ov::PartialShape> input_shapes;
-    size_t transpose_position;
-    ov::element::Type elem_type;
-    std::string targetDevice;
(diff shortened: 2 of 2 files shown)


a-sidorova:
Seems like we have two identical test classes: in [transpose_matmul.*](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/tests/functional/plugin/shared/src/snippets/transpose_matmul.cpp#L35) and [matmul.*](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/tests/functional/plugin/shared/src/snippets/matmul.cpp#L121)
Moreover, as I can see class only from [transpose_matmul.*](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/tests/functional/plugin/shared/src/snippets/transpose_matmul.cpp#L35) is [used](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/transpose_matmul.cpp#L19).
I suggest to remove test class from [matmul.*](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/tests/functional/plugin/shared/src/snippets/matmul.cpp#L121) and leave it in [matmul_transpose.*](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/transpose_matmul.cpp#L19) . @eshoguli what do you think?
Also I remember that I have already removed this class before... It's a black magic :smile:

eshoguli:
> Seems like we have two identical test classes: in [transpose_matmul.*](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/tests/functional/plugin/shared/src/snippets/transpose_matmul.cpp#L35) and [matmul.*](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/tests/functional/plugin/shared/src/snippets/matmul.cpp#L121)
> Moreover, as I can see class only from [transpose_matmul.*](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/tests/functional/plugin/shared/src/snippets/transpose_matmul.cpp#L35) is [used](https://github.com/openvinotoolkit/openvino/blob/ea48786bc67a3284ff9f514945edff7ffba4d72a/src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/transpose_matmul.cpp#L19).
> I suggest to r

done, thanks!

dmitry-gorokhov (approved):

PR #15353, opened 2023-01-27
[GPU][DG2] Fix fusings_gpu/gemm_2in_act_scale_eltwise.basic/4

tuxedcat:
### Details:
 - Onednn only supports 2D/3D gemm but openvino GPU plugin policy enforces 4D~6D. This API mismatch causes problems in the post-op axis and requires massive code changes. Therefore we decided to insert throw code for now and fix this issue later if some models require non-(per tensor/full tensor) post-ops.

### Tickets:
 - 67491

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/onednn/gemm_onednn.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/onednn/gemm_onednn.cpp
@@ -167,6 +167,14 @@ public:
     }
 
     static std::unique_ptr<primitive_impl> create(const gemm_node& arg, const kernel_impl_params& impl_params) {
+        bool full_tensor_or_per_tensor = true;
+        for (auto prim : arg.get_fused_primitives()) {
+            full_tensor_or_per_tensor &=
+                prim.input_layout.count() == prim.output_layout.count() || prim.input_layout.count() == 1;
+        }
+        if (!full_tensor_or_per_tensor) {
+            IE_THROW() << "FIXME: Onednn gemm has an issue for post-op axis.";
+        }
         auto& engine = impl_params.prog->get_engine();
         auto& config = impl_params.prog->get_config();
         auto attr = arg.get_onednn_primitive_attributes();
--- a/src/plugins/intel_gpu/tests/fusions/gemm_fusion_test.cpp
+++ b/src/plugins/intel_gpu/tests/fusions/gemm_fusion_test.cpp
@@ -74,6 +74,11 @@ public:
     }
 
     layout get_per_channel_layout(gemm_test_params& p) {
+        // WA: refer PR message
+        if (engine.get_device_info().supports_immad){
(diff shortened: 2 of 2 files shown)


tuxedcat:
@hyunback Could you check whether your [PR](https://github.com/openvinotoolkit/openvino/pull/15143) fix this issue?

tuxedcat:
> @hyunback Could you check whether your [PR](https://github.com/openvinotoolkit/openvino/pull/15143) fix this issue?

Nevermind. It doesn't fix this issue, but I think I can handle it based on onednn::combine_bf_with_first_spatial_dim(). WIP

hyunback:
@tuxedcat Sorry, late, I see it now. Yes you can support other dims using onednn::combine_bf_with_first_spatial_dim() likes 3 input dims.  https://github.com/openvinotoolkit/openvino/pull/15143 support 3 input dims with full tensor binary post-op.

isanghao (approved):

tuxedcat:
I'll look at it later because it's hard to solve this issue right away.

src/plugins/intel_gpu/src/graph/impls/onednn/gemm_onednn.cpp
```diff
@@ -301,6 +301,14 @@ struct gemm_onednn : typed_primitive_onednn_impl<gemm> {
     }
 
     static std::unique_ptr<primitive_impl> create(const gemm_node& arg, const kernel_impl_params& impl_params) {
+        bool full_tensor_or_per_tensor = true;
+        for (auto prim : arg.get_fused_primitives()) {
+            full_tensor_or_per_tensor &=
+                prim.input_layout.count() == prim.output_layout.count() || prim.input_layout.count() == 1;
+        }
+        if (!full_tensor_or_per_tensor) {
+            IE_THROW() << "FIXME: Onednn gemm has some issues in the post-op axis. Refer PR(#15353) message.";
```
isanghao, line 310: Please state clearly what is not supported. Better message would be "Unimplemented: Onednn gemm does not support per channel binary post-operation"

src/plugins/intel_gpu/tests/fusions/gemm_fusion_test.cpp
```diff
@@ -74,6 +74,11 @@ class GemmFusingTest : public ::BaseFusingTest<gemm_test_params> {
     }
 
     layout get_per_channel_layout(gemm_test_params& p) {
+        // WA: refer PR(#15353) message
```
isanghao, line 77: Please state it directly here.
"WA: per channel binary post-operation is not supported for Onednn gemm. Use single value for such case.

isanghao:
Let me check performance for this PR.

src/plugins/intel_gpu/src/graph/impls/onednn/gemm_onednn.cpp
tuxedcat: Done.

PR #15359, opened 2023-01-27
Split mo between openvino and openvino-dev

jane-intel:
mo_lite/ is for openvino module
mo/ is for openvino-dev module

both legacy and new fe works if run mo.py directly

- @popovaan @slyalin @andrei-kochin


rkazants:
@jane-intel, I think `mo_lite` is not lucky name:)
Can we consider a different name like `mo_core`?

rkazants (commented):
After this PR, will we still have MO a part of `openvino-dev` because I can't see changes in packaging or is it the first stage of splitting?

rkazants:
Also we need to check that both installation version will work appropriately:
1. python setup.py install develop
2. How will we setup developer version for `openvino-dev` version. Will we have setup.py for openvino-dev?
3. check that MO unit-tests coverage is preserved

jane-intel:
@rkazants this PR was for @popovaan use only. It doesn't require review.

akladiev:

PR #15364, opened 2023-01-27
Remove `OV_FRONTEND_PATH`

vurusovs:
### Details:
 - Remove `OV_FRONTEND_PATH` env var from FEM loading mechanism to fix security issue with unexpected injections
 - Add `register_front_end(name, library_path)` user API to C++ and Py  in order to provide possibility to use custom FEs. It's also required for testing FEs (`mock_py`, `mock1`)
 - Move `get_plugin_path()` to utils in order to reuse such login in `register_front_end` (user is able to load by abs, relative to CWD or from ENV)

### Tickets:
 - 101687


rkazants (commented):
Please double-check that it will not break Windows system that is absent in the pre-commit testing

src/frontends/common/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ else()
     set(FRONTEND_LIB_SUFFIX "${FRONTEND_LIB_SUFFIX}${CMAKE_SHARED_LIBRARY_SUFFIX}${OpenVINO_VERSION_SUFFIX}")
 endif()
 
-target_compile_definitions(${TARGET_NAME}_obj PRIVATE
+target_compile_definitions(${TARGET_NAME}_obj PUBLIC
```
vurusovs, line 53: @ilya-lavrenov @ilyachur please, take a look. I need `PREFIX` and `SUFFIX` to determine FE file name inside tests, maybe you have any ideas how to do it better?

src/core/tests/frontend/frontend_manager.cpp
```diff
@@ -70,35 +40,27 @@ TEST(FrontEndManagerTest, testAvailableFrontEnds) {
...
-        FrontEndManager fem;
-        auto frontends = fem.get_available_front_ends();
-        EXPECT_EQ(std::find(frontends.begin(), frontends.end(), "mock1"), frontends.end());
-    }
-
-    {
-        // update env
-        SetTestEnvrionment obj;
+TEST(FrontEndManagerTest, testFailRegisterFEByWrongPath) {
+    ASSERT_THROW(FrontEndManager fem; fem.register_front_end("mock1", mock_fe_path() + "_wrong");
```
ilyachur, line 44:
```suggestion
    FrontEndManager fem;
    ASSERT_THROW(fem.register_front_end("mock1", mock_fe_path() + "_wrong");
```

src/frontends/common/CMakeLists.txt
ilyachur: Why cannot you add these prefix and suffix to test binary?

src/core/tests/frontend/frontend_manager.cpp
vurusovs: Done

src/frontends/common/CMakeLists.txt
vurusovs: Done. But I deliberately ignore `OpenVINO_VERSION_SUFFIX` since there is also a symlink pointed to versioned library

ilya-lavrenov (approved):

ilyachur (approved):

PR #15365, opened 2023-01-27
[PT FE] Add aten::__and__ operator with layer test

sikorsl1:
### Details:
 - `aten::unfold` operator has been added

### Tickets:
 - 101069

### Comment:
- This is simplified version of `aten::__and___` operator. It does not cover case with numerical data. We need it to enable specific model but in the future this implementation has to be developed.


src/frontends/pytorch/src/op_table.cpp
```diff
@@ -111,6 +111,7 @@ OP_CONVERTER(translate_zeros_like);
 
 const std::map<std::string, CreatorFunction> get_supported_ops() {
     return {
+        {"aten::__and__", op::translate_1to1_match_2_inputs<opset10::LogicalAnd>},  // TODO: cover numerical cases
```
mvafin, line 114: Do we have numerical cases for models we target for?

mvafin (approved):

eaidova (approved):

PR #15366, opened 2023-01-27
[MO] fix UnaryElementwise reverse_infer

pavel-esir:
**Description**: for UnaryElementwise operations by default is used reverse_infer for binary elementwise operation. The lates one tries to access input ports with index 1 which do not exist in unary elementwise ops
**Solution**: call `reverse_bypass_infer` in `UnaryElementwise`, which is designed exactly for this reason to pass exact shape value form single out port into input port.

Ticket: xxx-101680 and issue https://github.com/openvinotoolkit/openvino/issues/15138

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests: done
* [x]  Framework operation tests n/a
* [x]  Transformation tests: no new transformations

Documentation:
* [x]  Supported frameworks operations list: n/a
* [x]  Guide on how to convert the **public** model n/a
* [x]  User guide update: n/a


mvafin (approved):
I wonder how `UnaryElementwise` operation class is different from `Activation`.

tools/mo/unit_tests/mo/front/common/partial_infer/eltwise_test.py
```diff
@@ -53,10 +53,12 @@ def test_eltwise_infer(self, value1, shape1, value2, shape2, shape_infer, exp_va
...
-                             'node_1': {'shape': shape_array(value1).shape if value1 is not None else shape_array(shape1),
-                                        'value': value1},
-                             'node_2': {'shape': shape_array(value2).shape if value2 is not None else shape_array(shape2),
-                                        'value': value2}
+                             'node_1': {
+                                 'shape': shape_array(value1).shape if value1 is not None else shape_array(shape1),
+                                 'value': value1},
+                             'node_2': {
+                                 'shape': shape_array(value2).shape if value2 is not None else shape_array(shape2),
+                                 'value': value2}
```
rkazants, line 61: not needed to change

tools/mo/unit_tests/mo/front/common/partial_infer/eltwise_test.py
```diff
@@ -8,7 +8,7 @@
 
 from openvino.tools.mo.front.common.partial_infer.eltwise import eltwise_infer, eltwise_reverse_infer
 from openvino.tools.mo.front.common.partial_infer.utils import shape_array, strict_compare_tensors, \
-    dynamic_dimension_value
+    dynamic_dimension_value, reverse_bypass_infer
```
rkazants, line 11: please check Python formatting:  not sorted

tools/mo/unit_tests/mo/front/common/partial_infer/eltwise_test.py
pavel-esir: autoformat changed it, returned back

pavel-esir:
> I wonder how `UnaryElementwise` operation class is different from `Activation`.

indeed, for shape and type inference no difference at all

rkazants (approved):

tools/mo/unit_tests/mo/front/common/partial_infer/eltwise_test.py
rkazants: It should be like this:

```
from openvino.tools.mo.front.common.partial_infer.utils import dynamic_dimension_value, reverse_bypass_infer, shape_array, strict_compare_tensors,
```

PR #15367, opened 2023-01-27
Graph comparator: added consumers number check

v-Golubev:
### Details:
`Comparator::compare_outputs`: added consumers number check


v-Golubev:
@itikhono the test was added, thanks.

itikhono (approved):

PR #15368, opened 2023-01-27
[PT FE]: support transformation for case aten::size + aten::__getitem__

eaidova:
### Details:
 - *support aten::__getitem__*

### Tickets:
 - *94276*


src/frontends/pytorch/src/transforms/aten_getitem_replacer.cpp
```diff
@@ -95,6 +95,17 @@ AtenGetItemReplacer::AtenGetItemReplacer() {
             }
             return true;
         }
+        if (auto shape_of = std::dynamic_pointer_cast<opset10::ShapeOf>(input_node)) {
+            // case aten::size as input
+            auto getitem_idx = getitem->input(1).get_source_output();
```
rkazants, line 100:
```suggestion
            auto getitem_idx = getitem->input_value(1);
```

rkazants (approved):
@mvafin, please review it.

src/frontends/pytorch/src/transforms/aten_getitem_replacer.cpp
```diff
@@ -95,6 +95,17 @@ AtenGetItemReplacer::AtenGetItemReplacer() {
             }
             return true;
         }
+        if (auto shape_of = std::dynamic_pointer_cast<opset10::ShapeOf>(input_node)) {
```
mvafin, line 98: I think we can support the case when `__getitem__` is located in body and input is located outside. But there is a problem when `ListConstruct` is located outside. Basically we can translate `__getitem__` to `Gather` if input is `Parameter` and `ListConstruct` (with non zero number of inputs) to `Concat` with axis 0. That should cover majority of cases, but we can still have some issues with that approach.

mvafin (commented):
Please add PR description

src/frontends/pytorch/src/transforms/aten_getitem_replacer.cpp
eaidova: moved to translators

src/core/src/op/util/multi_subgraph_base.cpp
```diff
@@ -156,7 +156,9 @@ void ov::op::util::MultiSubGraphOp::validate_and_infer_type_body(
 
         auto body_parameter = params.at(input_description->m_body_parameter_index);
         auto input_partial_shape = input_value(index).get_partial_shape();
+        auto dtype = input_value(index).get_element_type();
```
mvafin, line 159: Could you add test in https://github.com/openvinotoolkit/openvino/blob/master/src/core/tests/type_prop/if.cpp that checks this logic?

mvafin (changes requested):

src/core/src/op/util/multi_subgraph_base.cpp
eaidova: added

src/core/tests/type_prop/if.cpp
```diff
@@ -335,3 +335,46 @@ TEST(type_prop, if_scalar_and_1d_static_union) {
...
+    Shape out0_shape{32, 40, 10};
+    auto sh = result0->get_output_shape(0);
+    EXPECT_EQ(sh, out0_shape);
+    // Check that If validation validates both bodies
+    Xt->set_element_type(ov::element::f16);
+    Yt->set_element_type(ov::element::f16);
+    Xe->set_element_type(ov::element::f16);
+    Ye->set_element_type(ov::element::f16);
+    X->set_element_type(ov::element::f16);
+    Y->set_element_type(ov::element::f16);
```
mvafin, line 376:
```suggestion
```
This is not needed here. Only out body types should be provided

src/core/tests/type_prop/if.cpp
```diff
@@ -335,3 +335,46 @@ TEST(type_prop, if_scalar_and_1d_static_union) {
     auto sh = result0->get_output_partial_shape(0);
     EXPECT_EQ(sh, out_shape);
 }
+
+TEST(type_prop, if_element_type_dynamic) {
+    // That which we iterate over
+    auto X = make_shared<op::Parameter>(element::dynamic, Shape{32, 40, 10});
+    auto Y = make_shared<op::Parameter>(element::dynamic, Shape{32, 40, 10});
```
mvafin, line 342:
```suggestion
    auto X = make_shared<op::Parameter>(ov::element::f16, Shape{32, 40, 10});
    auto Y = make_shared<op::Parameter>(ov::element::f16, Shape{32, 40, 10});
```

mvafin (changes requested):

mvafin (approved):

PR #15370, opened 2023-01-27
update NV12

kblaszczak-intel:
add single-plane input information
create single-plane cpp snippet


docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1,55 @@
...
+    //! [batched_case]
+    cl::Image2D img_y_plane_0, img_y_plane_l;
+    auto input_y = model_with_preproc->get_parameters().at(0); 
+    auto remote_y_tensor_0 = remote_context.create_tensor(input_y->get_element_type(), in_shape, img_y_plane);
+    auto remote_y_tensor_1 = remote_context.create_tensor(input_y->get_element_type(), in_shape, img_y_plane);
+    std::vector<ov::Tensor> y_tensors = {remote_y_tensor_0, remote_y_tensor_1}; 
+    infer_request.set_tensors(*input_y->output(0).get_tensor().get_names().begin(), y_tensors);
+    infer_request.infer();
+    //! [batched_case]
+}
```
ilya-lavrenov, line 53: why do we need new code snippets? is it extension of what we have?

docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1 @@
+#include <openvino/runtime/core.hpp>
```
ilya-lavrenov, line 1: remove this file

docs/snippets/gpu/preprocessing_single-plane.cpp
kblaszczak-intel: It is supposed to be an exact copy of the preprocessing.cpp snippet, with changes for single-plane input. If you see the option of putting all the code in one file, I'm all for it.
Even more so that there are some build issues... apparently the first line in this file breaks things... But in the original it does not.

ilya-lavrenov: it's OK to have several files

tsavina (approved):

docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1,55 @@
...
```
ilya-lavrenov, line 1: please, add file docs/snippets/gpu/preprocessing_single-plane.cpp here https://github.com/openvinotoolkit/openvino/blob/master/docs/snippets/CMakeLists.txt#L52-L58

kblaszczak-intel: Done

docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1,55 @@
...
+    inf_req_remote.set_tensor(*input_y->output(0).get_tensor().get_names().begin(), remote_y_tensor);
+    inf_req_remote.infer() ;
+    //! [single_batch]
+}
+
+{
+    auto y_tensor_0 = get_y_tensor();
+    auto y_tensor_1 = get_y_tensor();
+    auto uv_tensor_0 = get_uv_tensor();
+    auto uv_tensor_1 = get_uv_tensor();
```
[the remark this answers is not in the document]
kblaszczak-intel: removed

docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1,55 @@
...
+    auto context = compiled_model.get_context().as<ov::intel_gpu::ocl::ClContext>();
+    auto input = model->get_parameters().at(0);
+    auto infer_request = compiled_model.create_infer_request();
+
+{
+    //! [single_batch]
+    cl::Image2D img_y_plane;
+    auto input_y = model_with_preproc->get_parameters().at(0); 
+    auto remote_y_tensor = remote_context.create_tensor(input_y->get_element_type(), in_shape, img_y_plane);
+    inf_req_remote.set_tensor(*input_y->output(0).get_tensor().get_names().begin(), remote_y_tensor);
```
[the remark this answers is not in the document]
kblaszczak-intel: @vladimir-paramuzov 
That's a very good point. I have no expertise to improve the snippet, but if it is how you describe it, I assume we may want to reword the description also - lines 188/189. Do you think it would be proper to write:
calling the `create_tensor()` function creates one `ov::RemoteTensor` object, representing both the Y and UV planes, stored in that order;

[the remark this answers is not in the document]
sshlyapn: @vladimir-paramuzov , actually, it was an intention to describe exactly that case (where only one Y plane from two is used) for NV12 to Gray conversion. Should we call it something like "single_Y_plane" processing or something else to avoid misleading?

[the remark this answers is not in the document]
vladimir-paramuzov: @sshlyapn If the intention is to describe nv12->gray conversion, then it should be called nv12->gray, not single-plane preprocessing, as those are different use cases. Ideally, we need to describe both scenarios and cover them with snippets.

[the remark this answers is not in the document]
kblaszczak-intel: That's all very interesting :) So am I correct in thinking that we can store the planes like so:
- both planes in two files
- both planes in one file
- one plane in one file

And the snippet presents the case with only Y stored in one object? If so, how about we change the description to something like:

* two-plane NV12 video surface input - calling the `create_tensor_nv12()` function creates
  a pair of `ov::RemoteTensor` objects, one per each plane;
* single-plane NV12 video surface input - calling the `create_tensor()` function creates one 
  `ov::RemoteTensor` object, representing either both planes together (Y elements 
  before UV elements) or just a single plane.

Would that justify the "single-plane" snippet?

docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    auto compiled_model = core.compile_model(model_with_preproc, "GPU");
+    auto remote_context = compiled_model.get_context().as<ov::intel_gpu::ocl::ClContext>();
+    auto input = model->get_parameters().at(0);
+    auto infer_request = compiled_model.create_infer_request();
+
+{
+    //! [single_batch]
+    cl::Image2D img_y_plane;
+    auto input_y = model_with_preproc->get_parameters().at(0);
```
ilya-lavrenov, line 31:
```suggestion
    auto input_y = model_with_preproc->input(0);
```

docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1,50 @@
...
+    auto remote_y_tensor = remote_context.create_tensor(input_y->get_element_type(), input->get_shape(), img_y_plane);
+    infer_request.set_tensor(*input_y->output(0).get_tensor().get_names().begin(), remote_y_tensor);
```
ilya-lavrenov, line 33:
```suggestion
    infer_request.set_tensor(input_y->get_any_name(), remote_y_tensor);
```

docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1,50 @@
...
+    infer_request.infer() ;
+    //! [single_batch]
+}
+
+{
+    //! [batched_case]
+    cl::Image2D img_y_plane_0, img_y_plane_l;
+    auto input_y = model_with_preproc->get_parameters().at(0);
```
ilya-lavrenov, line 41: the same here

docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1,50 @@
...
+    p.input().tensor().set_element_type(ov::element::u8)
+                      .set_layout("NHWC")
+                      .set_memory_type(ov::intel_gpu::memory_type::surface);
+    p.input().model().set_layout("NCHW");
+    auto model_with_preproc = p.build();
+    //! [init_preproc]
+
+    auto compiled_model = core.compile_model(model_with_preproc, "GPU");
+    auto remote_context = compiled_model.get_context().as<ov::intel_gpu::ocl::ClContext>();
+    auto input = model->get_parameters().at(0);
```
ilyachur, line 25:
```suggestion
    auto input = model->input(0);
```

docs/snippets/gpu/preprocessing_single-plane.cpp
```diff
@@ -0,0 +1,50 @@
...
+    auto infer_request = compiled_model.create_infer_request();
+
+{
+    //! [single_batch]
+    cl::Image2D img_y_plane;
+    auto input_y = model_with_preproc->get_parameters().at(0);
+    auto remote_y_tensor = remote_context.create_tensor(input_y->get_element_type(), input->get_shape(), img_y_plane);
+    infer_request.set_tensor(*input_y->output(0).get_tensor().get_names().begin(), remote_y_tensor);
+    infer_request.infer() ;
```
ilyachur, line 34:
```suggestion
    infer_request.infer();
```

kblaszczak-intel:
Many changes have been made, the article now presents three options:
two-plane,
single-plane,
to grey

The third snippet has been included and filenames have been fixed to be more straightforward.
all checks are green :) Can we merge?

vladimir-paramuzov (approved):
lgtm

ababushk:
Folks, master branch is failing in Jenkins after this commit, I've created a PR to revert it.

PR #15380, opened 2023-01-28
[CPU] MEMC loading failed RuntimeError: There should be only one inst…

nshchego:
…ance of RegistersPool per thread.

### Details:
 - *Fix for CPU RegistersPool. It could not be released after the code generation and created for the next kernel.*

### Tickets:
 - *99549*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/kernels/registers_pool.hpp
+++ b/src/plugins/intel_cpu/src/nodes/kernels/registers_pool.hpp
@@ -30,6 +30,7 @@ using namespace dnnl::impl::cpu;
 class RegistersPool {
 public:
     using Ptr = std::shared_ptr<RegistersPool>;
+    using WeakPtr = std::weak_ptr<RegistersPool>;
     static constexpr int anyIdx = -1;
 
     /**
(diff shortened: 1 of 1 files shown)


dmitry-gorokhov:
@maxnick could you please take a look?

src/plugins/intel_cpu/src/nodes/kernels/registers_pool.hpp
```diff
@@ -65,12 +66,12 @@ class RegistersPool {
...
         void release() {
-            if (regPool) {
-                regPool->returnToPool(reg);
+            if (auto pool = regPool.lock()) {
+                pool->returnToPool(reg);
                 regPool.reset();
             }
         }
-        bool isInitialized() const { return static_cast<bool>(regPool); }
+        bool isInitialized() const { return static_cast<bool>(!regPool.expired()); }
```
maxnick, line 74:
```suggestion
        bool isInitialized() const { return !regPool.expired(); }
```

nshchego: Done

dmitry-gorokhov (approved):

PR #15386, opened 2023-01-30
[GPU] Allocate output to host if it is to be used as shape infer dependency 

yeonbok:
### Details:
 - Shape inference with memory inputs are time consuming if those values are in the device memory, because it needs copy to host
 - Allocate such memories to host improved shape inference performance in dGPU

### Tickets:
 - 93592


ahnyoung-paul (approved):
LGTM

PR #15390, opened 2023-01-30
[GNA] Fix supported layers validator

tadamowicz:
### Details:
 - Skip unconnected inputs when validating supported layers and only report exceptions if all inputs are unconnected during the search for the first input layer in the graph.

### Tickets:
 - 102360


tadamowicz:
accidently created

PR #15391, opened 2023-01-30
[GNA] Skip unconnected inputs when searching for the first input laye…

tadamowicz:
…r in the graph.

### Details:
 - Skip unconnected inputs when validating supported layers and only report exceptions if all inputs are unconnected during the search for the first input layer in the graph.

### Tickets:
 - 102360


akladiev:

PR #15392, opened 2023-01-30
Remove `setupTestEnv()` from  `frontend_shared_test_classes`

vurusovs:
### Details:
 - Remove extra dependency - OV_FRONTEND_PATH points to the same place from what FrontEndManager loads frontends by default

### Tickets:
 - 101687

### DIFF
--- a/src/frontends/tests/frontend/shared/src/basic_api.cpp
+++ b/src/frontends/tests/frontend/shared/src/basic_api.cpp
@@ -16,7 +16,6 @@ std::string FrontEndBasicTest::getTestCaseName(const testing::TestParamInfo<Basi
 }
 
 void FrontEndBasicTest::SetUp() {
-    FrontEndTestUtils::setupTestEnv();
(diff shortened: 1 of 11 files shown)


src/frontends/tests/frontend/shared/src/basic_api.cpp
```diff
@@ -16,7 +16,6 @@ std::string FrontEndBasicTest::getTestCaseName(const testing::TestParamInfo<Basi
 }
 
 void FrontEndBasicTest::SetUp() {
-    FrontEndTestUtils::setupTestEnv();
```
ilyachur, line 19: Shouldn't you remove `setupTestEnv` function at all?

vurusovs: Yes, I'll do it later when will remove all mentions of OV_FRONTEND_PATH

ilya-lavrenov (approved):

rkazants (approved):

PR #15396, opened 2023-01-30
[GPU] Shape agnostic FC opt tiled kernel

sshlyapn:
### Details:
 - Shape agnostic FC bf_tiled kernel

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/fully_connected.cpp
+++ b/src/plugins/intel_gpu/src/graph/fully_connected.cpp
@@ -166,7 +166,7 @@ std::vector<layout> fully_connected_inst::calc_output_layouts(fully_connected_no
 
 
 kernel_impl_params fully_connected_inst::get_fake_aligned_params(kernel_impl_params const& orig_impl_param) {
-    // fc_tiled_opt kernel is optimized for row shape aligned by 16.
+    // fc_tiled_opt kernel is optimized for row shape aligned by 8.
     // Thus, use fake aligned shape at kernel execution for better performance.
     auto orig_input_layout = orig_impl_param.get_input_layout();
     auto orig_output_layout = orig_impl_param.get_output_layout();
@@ -176,10 +176,10 @@ kernel_impl_params fully_connected_inst::get_fake_aligned_params(kernel_impl_par
         auto updated_param = orig_impl_param;
         auto input_shape = orig_input_layout.get_partial_shape().to_shape();
         auto input_row_idx = input_shape.size() - 2;
-        input_shape[input_row_idx] = align_to(input_shape[input_row_idx], 16);
+        input_shape[input_row_idx] = align_to(input_shape[input_row_idx], 8);
         auto output_shape = orig_output_layout.get_partial_shape().to_shape();
         auto output_row_idx = output_shape.size() - 2;
-        output_shape[output_row_idx] = align_to(output_shape[output_row_idx], 16);
+        output_shape[output_row_idx] = align_to(output_shape[output_row_idx], 8);
 
         updated_param.input_layouts[0] = layout(ov::PartialShape(input_shape),
                                                 orig_input_layout.data_type,
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/fully_connected.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/fully_connected.cpp
@@ -118,6 +118,9 @@ public:
             params.quantization = kernel_selector::QuantizationType::NONE;
         }
 
+        if (impl_param.get_program().get_node(primitive->id).is_dynamic())
+            params.is_dynamic = true;
+
         return {params, optional_params};
     }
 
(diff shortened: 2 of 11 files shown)


src/plugins/intel_gpu/src/graph/include/kernel_selector_helper.h
```diff
@@ -129,6 +129,8 @@ struct kernel_impl_params {
 
     memory::ptr reordered_weights = nullptr;
 
+    bool is_dynamic = false;
```
yeonbok, line 132: Do we need this new flag? Just have a new flag in kernel_selector's base param to have such a flag so that it can be set when the kernel is created?

src/plugins/intel_gpu/src/kernel_selector/kernel_selector_params.h
```diff
@@ -627,6 +627,7 @@ struct fused_operation_desc {
 struct base_params : public Params {
     virtual ~base_params() {}
 
+    bool is_dynamic = false;
```
yeonbok, line 630: Yes this can be added, but I think this can be set at the creation time of the dynamic impl ?

src/plugins/intel_gpu/tests/test_cases/fully_connected_gpu_test.cpp
```diff
@@ -2059,3 +2059,175 @@ TEST(fully_connected_gpu, dynamic_multi_inference_multiple_shapes) {
...
+            for (int ofm = 0; ofm < output_f; ofm++) {
+                AccT acc = static_cast<AccT>(bias[ofm]);
+                for (int ifm = 0; ifm < input_f; ifm++) {
+                    acc += weights[ofm * input_f + ifm] * input[b * input_f + ifm];
+                }
+                result[b * output_f + ofm] = acc;
+            }
+        }
+
+        return result;
```
yeonbok, line 2086: Not that important, but why not just call the ngraph reference code?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -359,7 +359,10 @@ void primitive_inst::update_impl() {
                             return;
                     }
 
-                    auto impl = _node->type()->choose_impl(*_node, updated_params);
+                    auto static_params = updated_params;
+                    static_params.is_dynamic = false;
```
yeonbok, line 363: If we have this flag in kernel param only, not in the impl param, this is unnecessary.

src/plugins/intel_gpu/src/graph/include/kernel_selector_helper.h
sshlyapn: Yes, agree, removed

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
sshlyapn: Yeah, thank you

src/plugins/intel_gpu/tests/test_cases/fully_connected_gpu_test.cpp
yeonbok: Oh I  see. Thanks!

src/plugins/intel_gpu/src/graph/fully_connected.cpp
```diff
@@ -173,10 +173,10 @@ kernel_impl_params fully_connected_inst::get_fake_aligned_params(kernel_impl_par
         auto updated_param = orig_impl_param;
         auto input_shape = orig_input_layout.get_partial_shape().to_shape();
         auto input_row_idx = input_shape.size() - 2;
-        input_shape[input_row_idx] = align_to(input_shape[input_row_idx], 16);
+        input_shape[input_row_idx] = align_to(input_shape[input_row_idx], 8);
         auto output_shape = orig_output_layout.get_partial_shape().to_shape();
         auto output_row_idx = output_shape.size() - 2;
-        output_shape[output_row_idx] = align_to(output_shape[output_row_idx], 16);
+        output_shape[output_row_idx] = align_to(output_shape[output_row_idx], 8);
```
yeonbok, line 179: How was this number decided?

src/plugins/intel_gpu/src/graph/impls/ocl/fully_connected.cpp
```diff
@@ -118,6 +118,9 @@ struct fully_connected_impl : typed_primitive_impl_ocl<fully_connected> {
             params.quantization = kernel_selector::QuantizationType::NONE;
         }
 
+        if (impl_param.get_program().get_node(primitive->id).is_dynamic())
+            params.is_dynamic = true;
```
andrew-k-park, line 122: similar flag(`is_shape_agnostic`) has already been added on this PR(https://github.com/openvinotoolkit/openvino/pull/15317) for the same requirement. Please rebase and update to use it from base kernel params

src/plugins/intel_gpu/src/graph/fully_connected.cpp
sshlyapn: As I can see, blocking by batch dimension is more effective in case of alignment to 8, especially in case of small batch sizes

src/plugins/intel_gpu/src/graph/impls/ocl/fully_connected.cpp
sshlyapn: Updated, thank you

vladimir-paramuzov (approved):

PR #15400, opened 2023-01-31
[GPU] Add layout supports for shuffle_channels

isanghao:
### Details:
- Add onednn layout support for shuffle_channels

### Tickets:
 - 102500

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/shuffle_channels.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/shuffle_channels.cpp
@@ -48,9 +48,9 @@ struct shuffle_channels_impl : typed_primitive_impl_ocl<shuffle_channels> {
 namespace detail {
 
 attach_shuffle_channels_impl::attach_shuffle_channels_impl() {
-    static const auto types =
+    auto types =
         {data_types::f16, data_types::f32, data_types::i8, data_types::u8};
-    static const auto formats = {
+    auto formats = {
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gpu/src/graph/impls/ocl/shuffle_channels.cpp
```diff
@@ -48,26 +48,24 @@ struct shuffle_channels_impl : typed_primitive_impl_ocl<shuffle_channels> {
...
+        format::bs_fs_yx_bsv32_fsv32,
+        format::bs_fs_yx_bsv32_fsv16,
+    };
+    std::set<std::tuple<data_types, format::type>> keys;
+    for (const auto t : types) {
+        for (const auto f : formats) {
+            keys.emplace(t, f);
+        }
+    }
+    implementation_map<shuffle_channels>::add(impl_types::ocl, typed_primitive_impl_ocl<shuffle_channels>::create<shuffle_channels_impl>, keys);
```
vladimir-paramuzov, line 68: There is an `implementation_map::add` overload which does combine internally, so you can just call:
```
implementation_map<shuffle_channels>::add(impl_types::ocl, typed_primitive_impl_ocl<shuffle_channels>::create<shuffle_channels_impl>, types, formats);
```

isanghao: oh thanks for the comment. Fixed.

PR #15404, opened 2023-01-31
[CPU]Fix bf16 marking logic & remove useless convert

zhangYiIntel:
### Details:
 - *fix bf16 marking logic*
 - *remove useless convert*

### Tickets:
 - *95678*


zhangYiIntel:
@dmitry-gorokhov  This is the PR to fix BF16 marking logic when the output precision of model is BF16

dmitry-gorokhov:
@EgorDuplensky  would appreciate if you can take a look. 

The major idea here is to relax current BF16 mark-up logic for tails in case some outputs are explictily forced to have BF16 precision by user. BF16 outputs are forced using post-processing API which assumes following subgraphs Node [FP32] -> Convert [BF16] -> Outputs [BF16]. 
The PRs forces BF16 precision even for tails in case corresponding output is BF16 and eliminates duplicated Convert operations.

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2221,5 +2225,22 @@ void GraphOptimizer::reshapeRnnSeq(Graph &graph) {
     }
 }
```
EgorDuplensky, line 2227: I propose to add a corresponding description to the transformation regarding the motivation.
i.e.
"Removes redundant Convert node.
Example: BF16 model output is forced by post-processing API which results in the following subgraph:
Node [FP32] -> Convert [BF16] -> Outputs [BF16]
After EnforceBF16 routine the subgraph becomes:
Node [BF16] -> Convert [BF16] -> Outputs [BF16]
So Convert is redundant."

EgorDuplensky (approved):

dmitry-gorokhov:
@zhangYiIntel As we discussed offline lets also extend Snippets mark-up skipped logic to avoid Convert operations to be tokenized before newly introduced operation.

zhangYiIntel:
> @zhangYiIntel As we discussed offline lets also extend Snippets mark-up skipped logic to avoid Convert operations to be tokenized before newly introduced optimization pass.

The BF16 marking routine happens inside CPU graph, the Ngraph model doesn't have correct precision info, therefore the Snippets mark-up stage of ngraph doesn't have enough info to skip that `Convert`. Maybe I need to find a way to skip inside `graph_optimizer.cpp`

dmitry-gorokhov:
> > @zhangYiIntel As we discussed offline lets also extend Snippets mark-up skipped logic to avoid Convert operations to be tokenized before newly introduced optimization pass.
> 
> The BF16 marking routine happens inside CPU graph, the Ngraph model doesn't have correct precision info, therefore the Snippets mark-up stage of ngraph doesn't have enough info to skip that `Convert`. Maybe I need to find a way to skip inside `graph_optimizer.cpp`

I believe the condition should be pretty simple: this is Convert right before Result op and it does conversion from FP32 to BF16. Isn't it enough to mark-up it?

zhangYiIntel:
> > > @zhangYiIntel As we discussed offline lets also extend Snippets mark-up skipped logic to avoid Convert operations to be tokenized before newly introduced optimization pass.
> > 
> > 
> > The BF16 marking routine happens inside CPU graph, the Ngraph model doesn't have correct precision info, therefore the Snippets mark-up stage of ngraph doesn't have enough info to skip that `Convert`. Maybe I need to find a way to skip inside `graph_optimizer.cpp`
> 
> I believe the condition should be pretty simple: this is Convert right before Result op and it does conversion from FP32 to BF16. Isn't it enough to mark-up it?

Yes, I think you are correct, the related change in Snippets skipped  is submit

src/plugins/intel_cpu/src/ngraph_transformations/snippets_mark_skipped.cpp
```diff
@@ -458,6 +484,9 @@ bool SnippetsMarkSkipped::run_on_model(const std::shared_ptr<ov::Model> &m) {
         } else if (isSuitableSubtractAsZeroPointsParent(node)) {
             SetSnippetsNodeType(node, snippets::pass::SnippetsNodeType::SkippedByPlugin);
             channelAxis = DEFAULT_AXIS;
+        } else if (enableBF16 && isSuitableConvert(node)) {
```
zhangYiIntel, line 487: @dmitry-gorokhov The skipped logic requires that CPU plugin decides to use BF16 otherwise this convert may not be correct since snippets may construct it own convert subgraph which has similar pattern but has different meaning e.g.  example code [here](https://github.com/openvinotoolkit/openvino/blob/master/src/tests/functional/plugin/shared/include/snippets/convert.hpp)

zhangYiIntel:
@dmitry-gorokhov Hi Dmitry, any further suggestions ?

src/plugins/intel_cpu/src/ngraph_transformations/snippets_mark_skipped.cpp
dmitry-gorokhov: As we agreed lets provide comment why the skip is needed and create feature request on Snippets

dmitry-gorokhov (approved):

PR #15407, opened 2023-01-31
[GNA] Refactor target configuration

sirzabek:
### Details:
 - refactor target configuration and use

### Tickets:
 -102677

### DIFF
--- a/src/plugins/intel_gna/src/backend/gna_limitations.cpp
+++ b/src/plugins/intel_gna/src/backend/gna_limitations.cpp
@@ -19,6 +19,7 @@
 
 namespace ov {
 namespace intel_gna {
+using namespace common;
 namespace limitations {
 namespace cnn2d {
 
@@ -423,13 +424,17 @@ bool Validator_35::ShouldUseOnlyConv2DGnaIface() const {
     return true;
 }
 
-std::unique_ptr<AbstractValidator> AbstractValidator::Create(const std::string& target) {
-    if (target == common::kGnaTarget3_0) {
+std::unique_ptr<AbstractValidator> AbstractValidator::Create(const common::DeviceVersion& target) {
+    switch (target) {
+    case DeviceVersion::GNA3_0:
+    case DeviceVersion::GNAEmbedded3_1:
         return tools::make_unique<Validator_30>();
-    } else if (target == common::kGnaTarget3_5) {
+    case DeviceVersion::GNA3_5:
+    case DeviceVersion::GNAEmbedded3_5:
         return tools::make_unique<Validator_35>();
+    default:
+        return nullptr;
     }
-    return nullptr;
 }
 
 void AbstractValidator::ThrowIfNotEmpty(const std::string& prefix, const std::string& error) {
@@ -448,6 +453,11 @@ bool AbstractValidator::ValidationSuccesful(const bool throwOnError,
     }
 
     return error.empty();
+}
+
+bool UseOnly16BitConvolutionWeights(const DeviceVersion& compile_target) {
+    return (compile_target == common::DeviceVersion::GNA2_0 || compile_target == common::DeviceVersion::GNA3_0) ||
+           compile_target == common::DeviceVersion::GNAEmbedded3_1;
 }
 
 }  // namespace cnn2d
--- a/src/plugins/intel_gna/src/backend/gna_limitations.hpp
+++ b/src/plugins/intel_gna/src/backend/gna_limitations.hpp
@@ -10,6 +10,7 @@
 #include <cstdint>
 #include <ie_algorithm.hpp>
 
+#include "common/gna_target.hpp"
 #include "dnn_types.hpp"
 #include "gna_lib_ver_selector.hpp"
 
@@ -199,7 +200,7 @@ public:
                                OvGnaType inPrecision,
                                bool exception = true) const = 0;
 
-    static std::unique_ptr<AbstractValidator> Create(const std::string&);
+    static std::unique_ptr<AbstractValidator> Create(const common::DeviceVersion& target);
 };
 
 class Validator_30 : public AbstractValidator {
@@ -350,6 +351,8 @@ public:
                        OvGnaType inPrecision,
                        bool exception = true) const override;
 };
+
+bool UseOnly16BitConvolutionWeights(const common::DeviceVersion& compile_target);
 
 }  // namespace cnn2d
 
--- a/src/plugins/intel_gna/src/common/gna_target.hpp
+++ b/src/plugins/intel_gna/src/common/gna_target.hpp
@@ -4,16 +4,51 @@
 
 #pragma once
 
+#include <string>
+
+#include "gna2-common-api.h"
+#include "openvino/runtime/intel_gna/properties.hpp"
+
 namespace ov {
 namespace intel_gna {
 namespace common {
 
-static constexpr const char* kGnaTargetUnspecified = "";
-static constexpr const char* kGnaTarget2_0 = "GNA_TARGET_2_0";
-static constexpr const char* kGnaTarget3_0 = "GNA_TARGET_3_0";
-static constexpr const char* kGnaTarget3_1 = "GNA_TARGET_3_1";
-static constexpr const char* kGnaTarget3_5 = "GNA_TARGET_3_5";
-static constexpr const char* kGnaDefaultTarget = kGnaTarget3_0;
+enum class DeviceVersion {
+    NotSet = -1,
+    SoftwareEmulation = 0,
+    GNAEmbedded1_0 = 0x10e,
+    GNA2_0 = 0x20,
+    GNA3_0 = 0x30,
+    GNAEmbedded3_1 = 0x31e,
+    GNA3_5 = 0x35,
+    GNAEmbedded3_5 = 0x35e,
+    Default = GNA3_0
+};
+
+class Target {
+public:
+    DeviceVersion get_detected_device_version() const;
+    DeviceVersion get_user_set_execution_target() const;
+    DeviceVersion get_user_set_compile_target() const;
+    void set_detected_device_version(DeviceVersion detected_device);
+    void set_user_set_execution_target(DeviceVersion execution_target);
+    void set_user_set_compile_target(DeviceVersion compile_target);
+    DeviceVersion get_effective_execution_target() const;
+    DeviceVersion get_effective_compile_target() const;
+
+private:
+    DeviceVersion detected_device_version = DeviceVersion::SoftwareEmulation;
+    DeviceVersion user_set_execution_target = DeviceVersion::NotSet;
+    DeviceVersion user_set_compile_target = DeviceVersion::NotSet;
+};
+
+DeviceVersion HwGenerationToDevice(const HWGeneration& target);
+HWGeneration DeviceToHwGeneration(const DeviceVersion& target);
+DeviceVersion GnaToDevice(const Gna2DeviceVersion& target);
+Gna2DeviceVersion DeviceToGna(const DeviceVersion& target);
+DeviceVersion StringToDevice(const std::string& target);
+std::string DeviceToString(const DeviceVersion& target);
+bool IsEmbeddedDevice(const DeviceVersion& target);
 
 }  // namespace common
 }  // namespace intel_gna
--- a/src/plugins/intel_gna/src/gna2_model_export_helper.cpp
+++ b/src/plugins/intel_gna/src/gna2_model_export_helper.cpp
@@ -8,7 +8,6 @@
 #include <fstream>
 #include <numeric>
 
-#include "common/gna_target.hpp"
 #include "common/versioning.hpp"
 #include "gna/gna_config.hpp"
 #include "gna2-device-api.h"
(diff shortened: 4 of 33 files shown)


src/plugins/intel_gna/src/backend/gna_limitations.cpp
```diff
@@ -15,11 +15,14 @@
 #include "gna/gna_config.hpp"
 #include "gna_graph_tools.hpp"
 #include "gna_lib_ver_selector.hpp"
+#include "gna_limitations.hpp"
```
marcinkusm, line 18: redundant include

src/plugins/intel_gna/src/backend/gna_limitations.cpp
```diff
@@ -15,11 +15,14 @@
...
+#include "gna_plugin_config.hpp"
 #include "log/log.hpp"
 
 namespace ov {
 namespace intel_gna {
 namespace limitations {
+using namespace common;
```
marcinkusm, line 25: are you sure that it should be ov::intel_gna::limitations::common namespace?

src/plugins/intel_gna/src/backend/gna_limitations.cpp
```diff
@@ -450,6 +457,12 @@ bool AbstractValidator::ValidationSuccesful(const bool throwOnError,
     return error.empty();
 }
 
+bool UseOnly16BitConvolutionWeights(const Config& gna_config) {
+    const auto compile_target = gna_config.target->get_effective_compile_target();
```
marcinkusm, line 461: wouldn't passing `const common::DeviceVersion& target` be enough?

src/plugins/intel_gna/src/backend/gna_limitations.hpp
```diff
@@ -12,6 +12,7 @@
 
 #include "dnn_types.hpp"
 #include "gna_lib_ver_selector.hpp"
+#include "gna_plugin_config.hpp"
```
marcinkusm, line 15: Why does this file require gna_plugin_config.hpp? I think gna_target.hpp would be enough

src/plugins/intel_gna/src/backend/gna_limitations.cpp
```diff
@@ -15,11 +15,14 @@
 #include "gna/gna_config.hpp"
 #include "gna_graph_tools.hpp"
 #include "gna_lib_ver_selector.hpp"
+#include "gna_limitations.hpp"
+#include "gna_plugin_config.hpp"
```
marcinkusm, line 19: why do we need two  includes of "gna_plugin_config.hpp" one in header file and second in source file?

src/plugins/intel_gna/src/common/gna_target.hpp
```diff
@@ -4,16 +4,48 @@
...
+    GNA2_0 = 0x20,
+    GNA3_0 = 0x30,
+    GNAEmbedded3_1 = 0x31e,
+    GNA3_5 = 0x35,
+    GNAEmbedded3_5 = 0x35e,
+    Default = GNA3_0
+};
+
+class Target {
+    DeviceVersion detected_device_version = DeviceVersion::SoftwareEmulation;
```
marcinkusm, line 29: suggest to align with following order of declaration.
```
class MyClass{
public:
// internal types
// constructor, destructors, assignment Operators
// methods
// fields
 
protected:
// internal types
// constructor, destructors, assignment Operators
// methods
// fields
 
private:
// internal types
// constructor, destructors, assignment Operators
// methods
// fields
}
```

src/plugins/intel_gna/src/common/gna_target.hpp
```diff
@@ -4,16 +4,48 @@
...
 namespace intel_gna {
 namespace common {
 
-static constexpr const char* kGnaTargetUnspecified = "";
-static constexpr const char* kGnaTarget2_0 = "GNA_TARGET_2_0";
-static constexpr const char* kGnaTarget3_0 = "GNA_TARGET_3_0";
-static constexpr const char* kGnaTarget3_1 = "GNA_TARGET_3_1";
-static constexpr const char* kGnaTarget3_5 = "GNA_TARGET_3_5";
-static constexpr const char* kGnaDefaultTarget = kGnaTarget3_0;
+enum class DeviceVersion {
```
marcinkusm, line 16: I would initialize this values with values from Gna2DeviceVersion or event better create conversion function/operator
from Gna2DeviceVersion to DeviceVersion  and vice versa.

further you are sing casting from DeviceVersion  to Gna2DeviceVersion .
and e.g.

DeviceVersion::NotSet is -1 which does not map to any value Gna2DeviceVersion - it may lead to undefined behaviour

src/plugins/intel_gna/src/gna2_model_export_helper.cpp
```diff
@@ -147,34 +87,33 @@ std::string WriteAllEndpoints(std::ostream& outStream,
...
+    }
 
     uint32_t exportConfig;
     auto status = Gna2ModelExportConfigCreate(gnaUserAllocatorAlignedPage, &exportConfig);
     GNADeviceHelper::checkGna2Status(status, "Gna2ModelExportConfigCreate");
 
     status = Gna2ModelExportConfigSetSource(exportConfig, deviceIndex, modelId);
     GNADeviceHelper::checkGna2Status(status, "Gna2ModelExportConfigSetSource");
-    status = Gna2ModelExportConfigSetTarget(exportConfig, deviceVersionToExport);
+    status = Gna2ModelExportConfigSetTarget(exportConfig, static_cast<Gna2DeviceVersion>(compile_target));
```
marcinkusm, line 112: I would use coversion function from DeviceVersion to Gna2DeviceVersion - this are enums and values does not map 1:1

src/plugins/intel_gna/src/gna2_model_export_helper.cpp
```diff
@@ -278,3 +215,40 @@ void ExportTlvModel(uint32_t modelId,
...
+    void* bufferSueCreekHeader = nullptr;
+    uint32_t bufferSueCreekHeaderSize;
+
+    status = Gna2ModelExport(exportConfig,
+                             Gna2ModelExportComponentLegacySueCreekHeader,
+                             &bufferSueCreekHeader,
+                             &bufferSueCreekHeaderSize);
+    GNADeviceHelper::checkGna2Status(status, "Gna2ModelExport(LegacySueCreekHeader)");
+
+    (*modelHeader) = *(reinterpret_cast<Gna2ModelSueCreekHeader*>(bufferSueCreekHeader));
```
marcinkusm, line 239: shouldn't we check the pointer? It may lead to the segfault.

src/plugins/intel_gna/src/gna2_model_export_helper.cpp
```diff
@@ -278,3 +215,40 @@ void ExportTlvModel(uint32_t modelId,
         THROW_GNA_EXCEPTION << "Not succesfull status returned: " << tlv_status << ", from Gna2ExportTlv() function\n";
     }
 }
+
+void* ExportSueLegacyUsingGnaApi2(uint32_t modelId, uint32_t deviceIndex, Gna2ModelSueCreekHeader* modelHeader) {
```
marcinkusm, line 219: Correct mi if I am wrong, purpose of this method is to copy header of model to modelHeader variable and at the same time return dump.

src/plugins/intel_gna/src/gna_device.cpp
```diff
@@ -502,42 +474,42 @@ void GNADeviceHelper::dumpTLVForDeviceVersion(const uint32_t modelId,
...
+                   target->get_effective_compile_target(),
                    inputsContainer,
                    outputsContainer,
                    allAllocations);
 }
 
-void GNADeviceHelper::createVirtualDevice(Gna2DeviceVersion devVersion) {
-    const auto status = Gna2DeviceCreateForExport(devVersion, &nGnaDeviceIndex);
-    GNADeviceHelper::checkGna2Status(status, "Gna2DeviceCreateForExport(" + std::to_string(devVersion) + ")");
+void GNADeviceHelper::createVirtualDevice(DeviceVersion devVersion) {
```
marcinkusm, line 486: it is not bug but inconsistence in some method above you passes DeviceVersion but const reference but here by value.

src/plugins/intel_gna/src/gna_device.cpp
```diff
@@ -502,42 +474,42 @@ void GNADeviceHelper::dumpTLVForDeviceVersion(const uint32_t modelId,
...
         updateGnaDeviceVersion();
-    } else if (!executionTarget.empty() && gnaExecTarget != detectedGnaDevVersion) {
-        createVirtualDevice(gnaExecTarget);
+    } else if (execution_target != DeviceVersion::NotSet && execution_target != target->get_detected_device_version()) {
+        createVirtualDevice(execution_target);
         updateGnaDeviceVersion();
-        if (detectedGnaDevVersion != gnaExecTarget) {
-            THROW_GNA_EXCEPTION << "Wrong virtual GNA device version reported: " << detectedGnaDevVersion
-                                << " instead of: " << gnaExecTarget;
+        if (target->get_detected_device_version() != execution_target) {
```
marcinkusm, line 509: Correct me I am wrong but there is no chance to reach this statement having condition true due the fact updateGnaDeviceVersion() will set detected_device_version of target  on success, but in case of issue it will throw.

Maybe this THROW_GNA_EXCEPTION should be moved to the updateGnaDeviceVersion method?

src/plugins/intel_gna/src/gna_device.hpp
```diff
@@ -94,14 +94,14 @@ class GNADeviceHelper : public ov::intel_gna::GNADevice {
...
     static uint32_t selectGnaDevice();
-    static bool isGnaHw(const Gna2DeviceVersion dev) {
-        return Gna2DeviceVersionSoftwareEmulation != dev;
+    static bool is_hw_target(const common::DeviceVersion device_version) {
+        return common::DeviceVersion::SoftwareEmulation != device_version;
     }
-    bool hasGnaHw() const {
-        return isGnaHw(detectedGnaDevVersion);
+    bool is_hw_detected() const {
+        return is_hw_target(target->get_detected_device_version());
```
marcinkusm, line 101: hmm interesting that we require both methods to have public is_hw_target and is_hw_detected (this is static)
do we need this static method at all?

src/plugins/intel_gna/src/gna_device.hpp
```diff
@@ -176,11 +175,8 @@ class GNADeviceHelper : public ov::intel_gna::GNADevice {
 
     static void enforceLegacyCnns(Gna2Model& gnaModel);
     static void enforceLegacyCnnsWhenNeeded(Gna2Model& gnaModel);
-    static Gna2DeviceVersion parseTarget(const std::string& target);
-    Gna2DeviceVersion getDefaultTarget() const;
-    Gna2DeviceVersion getTargetDevice(bool execTarget) const;
 
-    void createVirtualDevice(Gna2DeviceVersion devVersion);
+    void createVirtualDevice(common::DeviceVersion devVersion);
```
marcinkusm, line 179: in method below you pass DeviceVersion const reference

marcinkusm (changes requested):
Nice work small changes needed.
Please update enum to enum conversion and fix includes.

I could not find tests for gna_target it self - its implementation is not huge but also not trivial.
Maybe it would be good to add them, but it is not mandatory, due the fact they are mostly checked by config.

src/plugins/intel_gna/src/backend/gna_limitations.cpp
sirzabek: Good catch.

src/plugins/intel_gna/src/backend/gna_limitations.hpp
sirzabek: Probably a leftover from some earlier version.

src/plugins/intel_gna/src/gna2_model_export_helper.cpp
sirzabek: This is out of scope of this PR.

src/plugins/intel_gna/src/gna2_model_export_helper.cpp
sirzabek: This is out of scope of this PR.

src/plugins/intel_gna/src/gna_device.hpp
sirzabek: is_up_to_20_hw will be made private.

src/plugins/intel_gna/src/gna2_model_export_helper.cpp
```diff
@@ -278,3 +215,40 @@ void ExportTlvModel(uint32_t modelId,
...
+    uint32_t exportConfig;
+    auto status = Gna2ModelExportConfigCreate(gnaUserAllocatorAlignedPage, &exportConfig);
+    GNADeviceHelper::checkGna2Status(status, "Gna2ModelExportConfigCreate");
+
+    status = Gna2ModelExportConfigSetSource(exportConfig, deviceIndex, modelId);
+    GNADeviceHelper::checkGna2Status(status, "Gna2ModelExportConfigSetSource");
+    status =
+        Gna2ModelExportConfigSetTarget(exportConfig, static_cast<Gna2DeviceVersion>(DeviceVersion::GNAEmbedded1_0));
```
marcinkusm, line 227: I would use specific value of Gna2DeviceVersion here due the fact it is hardocded eventually I would use conversion method.

sirzabek: Conversion method will be used, omitted by mistake..

src/plugins/intel_gna/src/gna2_model_export_helper.cpp
sirzabek: Omitted by mistake.

marcinkusm (approved):
Looks ok, 
nice work

mryzhov (changes requested):
Looks good generally, but I'm concerning about "common" namespace
I suggest using ov::intel_gna::target namespace for this functionality.

mryzhov (approved):

PR #15410, opened 2023-01-31
Enable TransposeSinking transformation in Tensorflow FrontEnd

itikhono:
Bug fix + enabling TransposeSinking in TF FE

### Tickets:
102477, 99678, 102475, 102952, 103318


src/common/transformations/src/transformations/common_optimizations/transpose_sinking_utils.cpp
```diff
@@ -129,17 +129,33 @@ ov::Output<ov::Node> FixInputNodeRank(ov::Output<ov::Node> input_node, ov::Rank:
...
+
+    AxisVector new_transpose_order;
+    if (rank > num_of_val) {
+        const auto diff = rank - num_of_val;
+        new_transpose_order.resize(rank);
+        std::iota(new_transpose_order.begin(), new_transpose_order.end(), 0);
+        auto transpose_axis_order = transpose_input_info.transpose_const->get_axis_vector_val();
+        for (int64_t i = diff; i < rank; ++i) {
+            new_transpose_order[i] = transpose_axis_order[i - diff] + diff;
+        }
```
jane-intel, line 144: I am worried. Could you explain what are the use cases for such transpose?

itikhono: disscused offline, it will be fixed for Concat op

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp
```diff
@@ -171,7 +180,18 @@ ov::pass::TransposeSinkingSplitForward::TransposeSinkingSplitForward() {
 
         auto& main_node_output = pattern_to_output.at(main_node_label);
         auto main_node = main_node_output.get_node_shared_ptr();
-
+        auto split_node = as_type_ptr<Split>(main_node);
+        auto split_axis_constant = as_type_ptr<Constant>(split_node->input_value(1).get_node_shared_ptr());
```
rkazants, line 184: I think we should check that this is not `nullptr`?

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp
```diff
@@ -171,7 +180,18 @@ ov::pass::TransposeSinkingSplitForward::TransposeSinkingSplitForward() {
...
+        auto split_axis = split_axis_constant->cast_vector<int64_t>()[0];
```
rkazants, line 185: We should check that the vector is not empty

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp
```diff
@@ -171,7 +180,18 @@ ov::pass::TransposeSinkingSplitForward::TransposeSinkingSplitForward() {
...
+        auto rank = split_node->input_value(0).get_partial_shape().rank();
+        if (split_axis < 0 && rank.is_static()) {
+            if (rank.is_static()) {
```
rkazants, line 188: duplicated check

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_utils.cpp
```diff
@@ -129,17 +129,33 @@ ov::Output<ov::Node> FixInputNodeRank(ov::Output<ov::Node> input_node, ov::Rank:
 }  // namespace
 
 namespace sink_forward {
+AxisVector AlignTransposeOrder(const Output<Node>& output, const TransposeInputsInfo& transpose_input_info) {
+    auto num_of_val = static_cast<int64_t>(shape_size(transpose_input_info.transpose_const->get_shape()));
+    const auto rank = output.get_partial_shape().rank().get_length();
```
rkazants, line 134: please check `output.get_partial_shape().rank()` to be static

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_utils.cpp
```diff
@@ -129,17 +129,33 @@ ov::Output<ov::Node> FixInputNodeRank(ov::Output<ov::Node> input_node, ov::Rank:
 }  // namespace
 
 namespace sink_forward {
+AxisVector AlignTransposeOrder(const Output<Node>& output, const TransposeInputsInfo& transpose_input_info) {
+    auto num_of_val = static_cast<int64_t>(shape_size(transpose_input_info.transpose_const->get_shape()));
```
rkazants, line 133: should we check that `transpose_input_info.transpose_const` is not `nullptr`?

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp
```diff
@@ -116,14 +116,23 @@ ov::pass::TransposeSinkingSplitBackward::TransposeSinkingSplitBackward() {
 
         NodePtr split = FindInputNode<Split>(transpose_label_node);
         auto split_axis_constant = as_type_ptr<Constant>(split->input_value(1).get_node_shared_ptr());
+        auto split_axis = split_axis_constant->cast_vector<int64_t>()[0];
+        auto rank = split->input_value(0).get_partial_shape().rank();
+        if (split_axis < 0 && rank.is_static()) {
+            if (rank.is_static()) {
```
rkazants, line 122: redundant check. Please correct this code

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_utils.cpp
itikhono: the current implementation doesn't support dynamic ranks for Concat inputs, in this way this method should never add this alignment. 
dynamic inputs to Concat will be processed separately in the next PR

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp
itikhono: done

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp
itikhono: fixed

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_utils.cpp
itikhono: fixed

itikhono: but actually it was invoked only from `UpdateInputTransposes` function where we have a check:
image

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_utils.cpp
itikhono: fixed
but actually it was invoked only from `UpdateInputTransposes` function where we have a check:
image

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp
itikhono: fixed

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_concat.cpp
```diff
@@ -32,18 +32,26 @@ ov::pass::TransposeSinkingConcatForward::TransposeSinkingConcatForward() {
         auto main_node = main_node_output.get_node_shared_ptr();
 
         TransposeInputsInfo transpose_input_info = GetFirstTransposeInput(main_node);
+        auto concat_node = as_type_ptr<Concat>(main_node);
+        auto concat_axis = concat_node->get_concatenation_axis();
+        if (concat_axis < 0) {
+            return false;
+        }
```
rkazants, line 39: why do we decline to handle this case? Can we normalize axis for the concatenation to have non-negative?

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp
```diff
@@ -76,6 +76,22 @@ bool IsSplitSinked(const Output<Node>& output) {
     return HasInputSplitAndTransposeSiblings(output) && is_sinking_node(output);
 }
 
+bool GetSplitAxis(const std::shared_ptr<Constant>& split_axis, const ov::Rank& rank, int64_t& axis) {
```
rkazants, line 79: I think it is better to generalize this function to execute normalization of `axis` for Concat and Split, etc.

rkazants (commented):
Last comments, other looks good.

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_concat.cpp
itikhono: if `concat_node->get_concatenation_axis();` returns negative value, it means that normalization has failed already (see validate_and_infer_type method of Concat op). Most probably rank is dynamic

src/common/transformations/src/transformations/common_optimizations/transpose_sinking_split.cpp
itikhono: normalization of axis for Concat op is already implemented as part of validate_and_infer_type method of Concat
if we call concat_node->get_concatenation_axis(); method, it should return normalized axis already

rkazants (approved):

itikhono:
the tests will be updated in the separate PR

evkotov (approved):

PR #15415, opened 2023-01-31
Add eltwise types resolving. Support big int constants.

mvafin:
### Details:
 - *Support big Python int constants*
 - *Implement PyTorch type promotion for eltwise ops (based on https://pytorch.org/docs/master/tensor_attributes.html#type-promotion-doc)*

### Tickets:
 - *101869, 102334*


src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -25,6 +25,8 @@ def get_type_from_py_type(value):
 def ivalue_to_constant(ivalue):
     ov_type = get_type_from_py_type(ivalue)
     if ov_type.is_static():
+        if ov_type == OVType.i32 and torch.iinfo(torch.int).max < ivalue or torch.iinfo(torch.int).min > ivalue:
+            ov_type = OVType.i64
         return op.Constant(ov_type, Shape([]), [ivalue]).outputs()
 
     if isinstance(ivalue, list):
```
mvafin, line 32:
```suggestion
    if isinstance(ivalue, (list, tuple)):
```

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -16,7 +16,10 @@ def get_type_from_py_type(value):
     if isinstance(value, float):
         return OVType.f32
     if isinstance(value, int):
-        return OVType.i32
+        # Python int is 64 bit, but we will convert it to int32 except cases when it can't fit in 32 bits
+        if torch.iinfo(torch.int).min <= value <= torch.iinfo(torch.int).max:
+            return OVType.i32
+        return OVType.i64
```
rkazants, line 22: why can't we convert it as? Can it be a problem that some operand will be decoded as `int32` and the other operand as `int64` and we will have type inconsistency?

src/frontends/pytorch/src/op/add.cpp
```diff
@@ -12,12 +15,14 @@ namespace pytorch {
 namespace op {
 
 OutputVector translate_add(NodeContext& context) {
+    auto lhs = context.get_input(0);
```
rkazants, line 18: please check a number of inputs to the operation

src/frontends/pytorch/src/op/div.cpp
```diff
@@ -14,21 +19,27 @@ namespace op {
...
-                                          " for aten::div");
+        rounding_mode = context.const_input<std::string>(2);
+    }
+    if (rounding_mode.empty()) {
+        // if no rounding mode and both inputs are ints cast BOTH to fp32
+        const auto x_dtype = x.get_element_type();
+        const auto y_dtype = y.get_element_type();
+        if (x_dtype.is_static() && x_dtype.is_integral() && y_dtype.is_static() && y_dtype.is_integral()) {
+            x = context.mark_node(std::make_shared<v0::Convert>(x, element::f32));
+            y = context.mark_node(std::make_shared<v0::Convert>(y, element::f32));
```
rkazants, line 32: may be it is better always cast them into `f32` without those checks? What if the type is dynamic and during inference they will be of `int32` type?

src/frontends/pytorch/src/op/mul.cpp
```diff
@@ -0,0 +1,29 @@
...
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_mul(NodeContext& context) {
+    auto x = context.get_input(0);
+    auto y = context.get_input(1);
+    auto input_size = context.get_input_size();
+    FRONT_END_OP_CONVERSION_CHECK(input_size >= 2, "Operation has less then 2 inputs.");
```
rkazants, line 18:
```suggestion
    auto input_size = context.get_input_size();
    FRONT_END_OP_CONVERSION_CHECK(input_size >= 2, "Operation has less then 2 inputs.");
    auto x = context.get_input(0);
    auto y = context.get_input(1);
```

src/frontends/pytorch/src/utils.cpp
```diff
@@ -459,6 +459,106 @@ Any simplified_type_interpret(Any type) {
...
+    {8, element::i8},
+    {16, element::i16},
+    {32, element::i32},
+    {64, element::i64},
+};
+}  // namespace
+
+void align_eltwise_input_types(const NodeContext& context,
+                               ov::Output<ov::Node>* lhs,
+                               ov::Output<ov::Node>* rhs,
```
rkazants, line 479: Let us pass them by reference instead. Simple pointer is not good here.

tests/layer_tests/py_frontend_tests/test_torch_decoder.py
```diff
@@ -21,6 +21,7 @@ def get_scripted_model(model):
         model = torch.jit.script(model)
         model.eval()
         model = torch.jit.freeze(model)
+        print(model.inlined_graph)  # will help debugging
```
rkazants, line 24: Will we clutter logs this way? May be it is better to comment it out if someone needs

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
mvafin: We had discussion with @slyalin about this issue, he was in favor of converting to int32 when we can do it, to reduce conversions later

src/frontends/pytorch/src/utils.cpp
mvafin: There is pros and cons in refs and ptrs, is there a strong reason why ptr is bad? I think it makes sense from the readability side, you always know when value will be changed inside the function when you call it

rkazants: Someone can pass nullptr and there are no checks on it.
Using pointers will lead to extra checks to have clean code.

src/frontends/pytorch/src/op/div.cpp
mvafin: No, fp64 should be left as fp64, also mixed case with int x float is better handled by `align_eltwise_input_types`

src/frontends/pytorch/src/op/add.cpp
mvafin: @eaidova adds a function for that in parallel PR, I will later update all operations to check number of inputs, so lets leave this out of scope for now

src/frontends/pytorch/src/utils.cpp
mvafin: Hmm, you may be right, okay I will change to refs

src/frontends/pytorch/src/utils.cpp
```diff
@@ -459,6 +459,106 @@ Any simplified_type_interpret(Any type) {
...
+        if (is_lhs_scalar && is_rhs_scalar) {
+            // if both scalar, align to lhs
+            rhs = context.mark_node(std::make_shared<opset10::ConvertLike>(rhs, lhs));
+            return;
+        }
+        auto lhs_dst_type = lhs_type;
+        auto rhs_dst_type = rhs_type;
+        if (is_lhs_scalar) {
+            if (lhs_type.is_real() && !rhs_type.is_real()) {
+                // if div we need to also align float types to highest biyness regardless of scalar
```
rkazants, line 509:
```suggestion
                // if div we need to also align float types to highest bitness regardless of scalar
```

src/frontends/pytorch/src/utils.cpp
```diff
@@ -459,6 +459,106 @@ Any simplified_type_interpret(Any type) {
...
+                    lhs_dst_type = element::f32;
+                rhs_dst_type = element::f32;
+            } else {
+                lhs = context.mark_node(std::make_shared<opset10::ConvertLike>(lhs, rhs));
+                return;
+            }
+        } else if (is_rhs_scalar) {
+            if (!lhs_type.is_real() && rhs_type.is_real()) {
+                lhs_dst_type = element::f32;
+                // if div we need to also align float types to highest biyness regardless of scalar
```
rkazants, line 520:
```suggestion
                // if div we need to also align float types to highest bitness regardless of scalar
```

src/frontends/pytorch/src/utils.cpp
```diff
@@ -459,6 +459,106 @@ Any simplified_type_interpret(Any type) {
...
+        if (lhs_dst_type.bitwidth() != rhs_dst_type.bitwidth()) {
+            const auto dst_bitness = std::max(lhs_dst_type.bitwidth(), rhs_dst_type.bitwidth());
+            element::Type* type_to_align = &lhs_dst_type;
+            if (rhs_dst_type.bitwidth() < dst_bitness)
+                type_to_align = &rhs_dst_type;
+            if (type_to_align->is_real()) {
+                *type_to_align = bit_to_float.at(dst_bitness);
+            } else {
+                *type_to_align = bit_to_int.at(dst_bitness);
+            }
```
rkazants, line 549: Please check - it looks `type_to_align` is not needed

mvafin: It is needed because it points to one of two types which has lowest bitness

rkazants: I didn't find that `type_to_align` is used below.

rkazants (approved):

eaidova (approved):

PR #15378, opened 2023-01-27
[ShapeInference] Improve GatherND shape inference

mitruska:
### Details:
GatherND V5 and V8 improvements:
 - Add preserving interval dimensions
 - Add preserving labels
 - Add shape_infer function for v5::GatherND and v8::GatherND
 - Register GatherND shape infer for CPU
 - Add tests
 
### Tickets:
 - 97231

PR #15416, opened 2023-01-31
[ShapeInference] GatherND improvements review

mitruska:
### Details:
 - No logic changes, small code refactoring, applied review suggestions from:
 https://github.com/openvinotoolkit/openvino/pull/15378
 
### Tickets:
 - Related to: 97231


src/core/shape_inference/include/gather_nd_shape_inference.hpp
```diff
@@ -73,12 +75,14 @@ void shape_infer(const GatherND* op, const std::vector<TShape>& input_shapes, st
     auto batch_dims = op->get_batch_dims();
     if (batch_dims > 1 && output_shapes[0].rank().is_static()) {
         const auto& output_base_shape = output_shapes[0];
```
praasz, line 77:
```suggestion
        auto& output_base_shape = output_shapes[0];
        auto output_dims = std::vector<DimType>{output_base_shape[0]};
        std::for_each(output_base_shape.begin() + 1,
                      output_base_shape.begin() + batch_dims,
                      [&output_dims](DimType& dim) {
                          output_dims[0] *= dim;
                      });
        output_dims.insert(output_dims.begin() + 1, output_base_shape.begin() + batch_dims, output_base_shape.end());
        output_shapes[0] = TShape(std::move(output_dims));
```
I'm sorry for inconvenience, but it looks like using accumulate is not good choice as at every iteration it will create new dimension (is not using move until c++20).
But I suggest to start this accumulate from next element to use first element as start value to reduce number of iterations and create dummy dimension with value {1}.
So propose alternative.
And `erase` is not possible for partial shape 😞

praasz (approved):

src/core/shape_inference/include/gather_nd_shape_inference.hpp
mitruska: No worries, thank you for the clarification, good catch.
Update applied.

PR #15422, opened 2023-02-01
[GPU] add different type support in range kernel

wilson-seok:
### Details:
 -  Add different type support in range kernel for such case of start:4.9(float) stop:19(int) step:1(int) and output is float

### Tickets:
 - 102037


yeonbok:
Could you add a functional test for this?

wilson-seok:
> Could you add a functional test for this?

@yeonbok I added mixed data type inputs in dynamic Range test case.

yeonbok (approved):

PR #15423, opened 2023-02-01
[CPU] Register printing utilities

xuchen-intel:
### Details:
 - *Register printing utilities for JIT kennels of CPU plugin. Invoke RegPrinter::print method inside JIT kernels.*
 -  *Param:*
 ```
The following combinations of Register types and precisions are supported.

           fp32         int32         int8         u8
Vmm        Yes          Yes           No          No
Reg64      Yes          Yes           No          No
Reg32      Yes          Yes           No          No
Reg16      No           No            Yes         Yes
Reg8       No           No            Yes         Yes

```
 - *Usage:*
```
Here are some examples. Note that user friendly register name will be printed, if it has been set.
While original Xbyak register name will always be printed.

Example 1:
Invocation: RegPrinter::print<float>(*this, vmm_val, "vmm_val");
Console:    vmm_val | ymm0: {30, 20, 25, 29, 24, 31, 27, 23}

Example 2:
Invocation: RegPrinter::print<float>(*this, vmm_val);
Console:    ymm0: {30, 20, 25, 29, 24, 31, 27, 23}

Example 3:
Invocation: RegPrinter::print<int>(*this, vmm_idx, "vmm_idx");
Console:    vmm_idx | ymm1: {5, 6, 0, 2, 0, 6, 6, 6}

Example 4:
Invocation: RegPrinter::print<int>(*this, reg_work_amount, "reg_work_amount");
Console:    reg_work_amount | r13: 8

Example 5:
Invocation: RegPrinter::print<int>(*this, reg_work_amount);
Console:    r13: 8

Example 6:
Invocation: RegPrinter::print<float>(*this, reg_tmp_64, "reg_tmp_64");
Console:    reg_tmp_64 | r15: 1
```

### Tickets:
 - *101222*

### DIFF
--- a/src/plugins/intel_cpu/src/emitters/utils.cpp
+++ b/src/plugins/intel_cpu/src/emitters/utils.cpp
@@ -12,75 +12,20 @@ namespace intel_cpu {
 using namespace Xbyak;
 using namespace dnnl::impl::cpu::x64;
 
-template <>
-void RegPrinter::print<float, Xmm>(jit_generator &h, Xmm reg, const char *name) {
-    print_vmm<float, Xmm>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<int, Xmm>(jit_generator &h, Xmm reg, const char *name) {
-    print_vmm<int, Xmm>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<float, Ymm>(jit_generator &h, Ymm reg, const char *name) {
-    print_vmm<float, Ymm>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<int, Ymm>(jit_generator &h, Ymm reg, const char *name) {
-    print_vmm<int, Ymm>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<float, Zmm>(jit_generator &h, Zmm reg, const char *name) {
-    print_vmm<float, Zmm>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<int, Zmm>(jit_generator &h, Zmm reg, const char *name) {
-    print_vmm<int, Zmm>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<float, Reg64>(jit_generator &h, Reg64 reg, const char *name) {
-    print_reg<float, Reg64>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<int, Reg64>(jit_generator &h, Reg64 reg, const char *name) {
-    print_reg<int, Reg64>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<float, Reg32>(jit_generator &h, Reg32 reg, const char *name) {
-    print_reg<float, Reg32>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<int, Reg32>(jit_generator &h, Reg32 reg, const char *name) {
-    print_reg<int, Reg32>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<char, Reg16>(jit_generator &h, Reg16 reg, const char *name) {
-    print_reg<char, Reg16>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<unsigned char, Reg16>(jit_generator &h, Reg16 reg, const char *name) {
-    print_reg<unsigned char, Reg16>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<char, Reg8>(jit_generator &h, Reg8 reg, const char *name) {
-    print_reg<char, Reg8>(h, reg, name);
-}
-
-template <>
-void RegPrinter::print<unsigned char, Reg8>(jit_generator &h, Reg8 reg, const char *name) {
-    print_reg<unsigned char, Reg8>(h, reg, name);
-}
+template void RegPrinter::print<float, Xmm>(jit_generator &h, Xmm reg, const char *name);
+template void RegPrinter::print<int, Xmm>(jit_generator &h, Xmm reg, const char *name);
+template void RegPrinter::print<float, Ymm>(jit_generator &h, Ymm reg, const char *name);
+template void RegPrinter::print<int, Ymm>(jit_generator &h, Ymm reg, const char *name);
+template void RegPrinter::print<float, Zmm>(jit_generator &h, Zmm reg, const char *name);
+template void RegPrinter::print<int, Zmm>(jit_generator &h, Zmm reg, const char *name);
+template void RegPrinter::print<float, Reg64>(jit_generator &h, Reg64 reg, const char *name);
+template void RegPrinter::print<int, Reg64>(jit_generator &h, Reg64 reg, const char *name);
+template void RegPrinter::print<float, Reg32>(jit_generator &h, Reg32 reg, const char *name);
+template void RegPrinter::print<int, Reg32>(jit_generator &h, Reg32 reg, const char *name);
+template void RegPrinter::print<char, Reg16>(jit_generator &h, Reg16 reg, const char *name);
+template void RegPrinter::print<unsigned char, Reg16>(jit_generator &h, Reg16 reg, const char *name);
+template void RegPrinter::print<char, Reg8>(jit_generator &h, Reg8 reg, const char *name);
+template void RegPrinter::print<unsigned char, Reg8>(jit_generator &h, Reg8 reg, const char *name);
 
 void RegPrinter::print_reg_fp32(const char *name, int val) {
     std::stringstream ss;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/plugins/intel_cpu/src/utils/debug_capabilities.h [hunk 1]
```diff
@@ -15,12 +15,71 @@
...
+    RegPrintManager manager(*this, REG_VAL, #REG_VAL, REG_PRECISION); \
+    manager.PrintRegister();                                          \
+}
+
+#define PRINT_REGISTER(...)                                      \
+    PRINT_REGISTER_SELECTOR(__VA_ARGS__,                         \
+                            PRINT_REGISTER_WITH_VALUE_PRECISION, \
+                            PRINT_REGISTER_WITH_VALUE)(__VA_ARGS__)
+
+class RegPrintManager {
```
dmitry-gorokhov, line 46: Some general thoughts. 
1. From architecture perspective I would prefer to have this functionality incapsulated into base class for all JIT functionality. We have a plan to create base class for all JIT kernels and JIT emitters inside the plugin and put this auxiliary staff inside. Since now we don't have such hierarhy lets leave as a separate class.
2. Given 1st statement it is better to avoid creation of objects of this class. My proposal it to keep all print methods static and hide costructors, while fields should be const expression. This is very simple printing functionality, so it shouldn't have any state.
3. Maybe we can call it RegPrinter and have print public methods like:
```
template< typename PRC_T, typename REG_T>
static void print(jit_generator &h, REG_T reg)
```
Then usage might look like:
```
RegPrinter::print<float>(*this, vmm_src);
```
Precision type is deduced from template param (PRC_T). Such approach will allow you to specialize template for only required precisions and avoid functions like CheckPrecision().
Register type is deduced from function argument (REG_T). It wil allow you to avoid checks like 
```
reg_val.isREG(64) || reg_val.isREG(32) || reg_val.isREG(16) || reg_val.isREG(8)
```
and just use template specialization mechanism.

4. Given 1st statement I would prefer to avoid MACRO usage. It was introduced by two reasons as I can see: object creation and register name printing. First is solved by static print method, second can be solved by overloading print method with additional register name attribute:
```
template< typename PRC_T, typename REG_T>
static void print(jit_generator &h, REG_T reg, std::string name)
```
By default if name is not passed you can just print number of register and its type getting it from Xbyak. So if vmm_src is actually YMM0 you can print exactly YMM0 before register content.

What do you think? If you have any concerns feel free to discuss.

src/plugins/intel_cpu/src/utils/debug_capabilities.h
```diff
@@ -15,12 +15,71 @@
...
+public:
+    RegPrintManager(dnnl::impl::cpu::x64::jit_generator &h, const Xbyak::Reg &reg_val, const char *reg_name): h(h),
+                    reg_val(reg_val), reg_name(reg_name), reg_dt(dnnl::memory::data_type::undef) {
```
dmitry-gorokhov, line 49: Please don't use dnnl types. This functionality should be fully independent from oneDNN. The only exceptions are `dnnl::impl::cpu::x64::jit_generator` and Xbyak itself.

src/plugins/intel_cpu/src/utils/debug_capabilities.h
```diff
@@ -15,12 +15,71 @@
...
 #include <onednn/dnnl.h>
 #include <dnnl_debug.h>
+#include <cpu/x64/jit_generator.hpp>
 
 #include "openvino/core/model.hpp"
 
 namespace ov {
 namespace intel_cpu {
 
+class RegPrintManager;
```
dmitry-gorokhov, line 25: There is no much sense to put this functionality inside major debug capabilities file, since it shouldn't depend on DEBUG_CAPS option. Lets move inside smt like  `utils.hpp` inside emitters folder.

src/plugins/intel_cpu/src/utils/debug_capabilities.h
```diff
@@ -15,12 +15,71 @@
...
+public:
+    RegPrintManager(dnnl::impl::cpu::x64::jit_generator &h, const Xbyak::Reg &reg_val, const char *reg_name): h(h),
+                    reg_val(reg_val), reg_name(reg_name), reg_dt(dnnl::memory::data_type::undef) {
+        InitDefaultPrecision();
+    }
+    RegPrintManager(dnnl::impl::cpu::x64::jit_generator &h, const Xbyak::Reg &reg_val, const char *reg_name,
+                    dnnl::memory::data_type reg_dt) : h(h), reg_val(reg_val), reg_name(reg_name), reg_dt(reg_dt) {
+        InitDefaultPrecision();
+    }
+    RegPrintManager(dnnl::impl::cpu::x64::jit_generator &h, const Xbyak::Xmm &xmm_val, const char *reg_name): h(h),
```
dmitry-gorokhov, line 56: I would reccomed to remove ability to print regiters with default precision, since it creates more confusion. Lets force the user to explicilty specify desired precision by template argument as I mentioned above.

src/plugins/intel_cpu/src/utils/debug_capabilities.h
xuchen-intel: @dmitry-gorokhov Hi Dmitry, I like your proposal! And I want to make sure that I think it the same way as you do for your first point. By saying "From architecture perspective I would prefer to have this functionality incapsulated into base class for all JIT functionality", do you mean, for example, to revise
`struct jit_uni_topk_kernel_f32 : public jit_uni_topk_kernel, public jit_generator` to
`struct jit_uni_topk_kernel_f32 : public jit_uni_topk_kernel, public jit_generator,  public RegPrinter`
and apply such revision to all the other jit kernals of CPU plugin nodes?

dmitry-gorokhov: @xuchen-intel My first statement was about view on future achitecture. This is not needed for current PR.

In short I expect to have jit_base class that is derived from jit_generator. All Jit kernels and emitters will be derived from jit_base. This jit_base will have print capability implemented as its protected method.

xuchen-intel: Good suggestions! Applied. Thanks Dmitry!

src/plugins/intel_cpu/src/utils/debug_capabilities.h
xuchen-intel: Fixed. Thanks Dmitry!

src/plugins/intel_cpu/src/emitters/utils.hpp
```diff
@@ -0,0 +1,87 @@
...
+namespace ov {
+namespace intel_cpu {
+
+// Usage
+// 1. Include this headfile where JIT kennels of CPU plugin are implemented for Register printing
+// 2. Invoke RegPrinter::print method. Here are some examples. Note that if register name is not set,
+//    original Xbyak register name will be printed by default.
+//    Example 1:
+//    Invocation: RegPrinter::print<float>(*this, vmm_val, "vmm_val");
+//    Console:    vmm_val: {30, 20, 25, 29, 24, 31, 27, 23}
```
dmitry-gorokhov, line 18: Thinking about console output: it sounds like it would always good to print also register type and number (like ymm0) even if name provided. Like:
ex if name provided: `"vmm_val" | <ymm0>: {...}`
ex if name not provided: `<ymm0>: {...}`

src/plugins/intel_cpu/src/emitters/utils.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+void RegPrinter::save_reg(jit_generator &h) {
+    h.sub(h.rsp, reg_len * reg_cnt);
+    for (size_t i = 0; i < reg_cnt; i++) {
+        h.mov(h.ptr[h.rsp + i * reg_len], Reg64(i));
+    }
+}
+
+template <>
+void RegPrinter::save_vmm<Xmm>(jit_generator &h) {
```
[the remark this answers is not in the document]
xuchen-intel: Good suggestion! Applied. Thanks Dmitry!

src/plugins/intel_cpu/src/emitters/utils.cpp
```diff
@@ -0,0 +1,267 @@
...
+#include <sstream>
+
+namespace ov {
+namespace intel_cpu {
+
+using namespace Xbyak;
+using namespace dnnl::impl::cpu::x64;
+
+template <>
+void RegPrinter::print<float, Xmm>(jit_generator &h, Xmm reg, const char *name) {
```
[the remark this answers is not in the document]
xuchen-intel: Good point! Applied. Thanks Dmitry!

src/plugins/intel_cpu/src/emitters/utils.hpp
xuchen-intel: Applied. Thanks Dmitry!

dmitry-gorokhov (approved):
LGTM now.
I think this is a good initial state of the debugging feature.

PR #15426, opened 2023-02-01
Review reverse class for shape inference aspects

praasz:
### Details:
 - Add test for partial shape and label propagation
 - Add template implementation for `shape_infer`
 - Add test for static shape inference

### Tickets:
 - 97249
 - 97289
 - 97338
 - 97387


src/core/tests/type_prop/reverse.cpp
```diff
@@ -155,39 +160,175 @@ TEST(type_prop, reverse_partial_rank_dynamic) {
...
+        NodeValidationFailure,
+        HasSubstr("Some of the provided axes (AxisSet{0, 2, 4}) are out of bounds (input rank: 4)."));
+}
+
+TEST_F(TypePropReverseV1Test, param_static_rank_partial_shape_axes_negatives) {
+    PartialShape param_shape{-1, {2, -1}, {-1, 3}, 5};
+    set_shape_labels(param_shape, 10);
+    auto param = make_shared<op::Parameter>(element::f32, param_shape);
+
+    auto op = make_op(param, Constant::create(element::i64, {3}, {0, -1, 2}), op::v1::Reverse::Mode::INDEX);
```
mitruska, line 280: Should negative axis values be allowed and pass the validation?

Unfortunately, there is no direct info about negative axes support in the [Reverse_1.md](https://github.com/openvinotoolkit/openvino/blob/da9470864c1e099f46611e1db9534ad0c89d4b24/docs/ops/movement/Reverse_1.md) operation specification.

praasz: Yes, I've got same doubts but I left this behavior unchanged.
It was hidden in `get_axis_set_val` when get AxisSet from constant value.

mitruska (approved):

src/core/shape_inference/include/reverse_shape_inference.hpp
```diff
@@ -0,0 +1,106 @@
...
+    } else if (data_rank.is_static()) {
+        // mode index & data rank is static
+        using TAxis = typename AxisSet::value_type;
+        static_assert(std::is_same<TAxis, util::ClipNegative::value_type>(),
+                      "AxisSet::value_type != ClipNegative::value_type");
+
+        if (const auto axes =
+                get_input_const_data_as<TShape, TAxis, AxisSet>(op, 1, constant_data, util::ClipNegative())) {
+            NODE_VALIDATION_CHECK(op,
+                                  all_of(axes->begin(), axes->end(), cmp::Less<TAxis>(data_rank.get_length())),
```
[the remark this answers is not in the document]
praasz: That was original implementation all negatives are treated as 0 and there is no difference if 0 was in tensor data.
This validation (L:74) checks if axis is greater than data rank (index points not existing dimension in data input) then throw.

jane-intel (approved):

PR #15430, opened 2023-02-01
[CPU][Snippets] Dynamism via recompilation and cache

chenhu-wang:
### Details:
 - *dynamism via recompilation*

### Tickets:
 - *99116*


dmitry-gorokhov:
@IvanNovoselov Could you please start the review?

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -394,79 +376,50 @@ std::vector<VectorDims> Snippet::shapeInfer() const {
         IE_THROW() << errorMessage.str();
     }
 
+    std::vector<ov::Shape> new_shapes;
+    for (const auto& s : normInputShapes)
+        new_shapes.emplace_back(s);
+    copy_snippet();
+    const auto& outputShapes = snippetAttrs.snippet->reshape_body(new_shapes);
```
IvanNovoselov, line 383: If I'm not missing anything, `ShapeInfer()` would be called every time `inputShapesModified()`. So currently it's every time input shapes are updated, even when we have them in the cache and don't actually need to perform expensive operations such as `copy_snippet()` and `reshape_body()`. Please correct me if I'm mistaken.

My second concern is, even when we really need to make shape infer (new set of shapes not in the cache), why do we need to `copy_snippet` every time? Why can't we reshape the original body?

src/plugins/intel_cpu/src/nodes/subgraph.h
```diff
@@ -47,6 +46,17 @@ class Snippet : public Node {
 
     // if generator is set, it would execute generated code otherwise it would fallback to nGraph reference
     void execute(dnnl::stream strm) override;
+    void executeDynamicImpl(dnnl::stream strm) override;
+
+    struct SnippetAttrs {
+        // Local copy of subgraph node for canonization & code generation
+        std::shared_ptr<ngraph::snippets::op::Subgraph> snippet;
```
IvanNovoselov, line 53: Why do we need a deep `snippet` copy in every `SnippetsAttrs` instance? In order to perform lowering and code generation in `SnippetJitExecutor(...)` it's enough to have a local copy of the reshaped `snippet` (with a life scope inside the constructor).

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -35,7 +38,95 @@ using namespace Xbyak;
 namespace ov {
 namespace intel_cpu {
 namespace node {
+namespace {
+
+struct SnippetKey {
+    Snippet::SnippetAttrs attrs;
```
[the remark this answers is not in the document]
chenhu-wang: SnippetKey is in an anonymous namespace, which indicate only valid in current file. This also prevent conflict or mis-used, and is consistent with Key declare in other nodes. such as: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/nodes/conv.cpp#L38

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -35,7 +38,95 @@ using namespace Xbyak;
...
+
+bool SnippetKey::operator==(const SnippetKey& rhs) const {
+    bool retVal = true;
+    if (attrs.inputMemDesc.size() != rhs.attrs.inputMemDesc.size())
+        return false;
+    if (attrs.outputMemDesc.size() != rhs.attrs.outputMemDesc.size())
+        return false;
 
+    for (size_t i = 0; i < attrs.inputMemDesc.size(); i++) {
+        if (attrs.inputMemDesc[i] != rhs.attrs.inputMemDesc[i]) {
```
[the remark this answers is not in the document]
chenhu-wang: A more light-weight key is used now. i.e. blockedDim, order and precision is used in generation stage besides the body. 
Only when all three is the same, the cached maybe hit in recompilation solution. Snippets body of dynamic shape together with exact blockedDim, order and precision determine if cache reused. Thank you!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -45,9 +136,10 @@ Snippet::Snippet(const std::shared_ptr<ngraph::Node>& op, const GraphContext::CP
     if (!original_snippet) {
         IE_THROW(NotImplemented) << "Node is not an instance of snippets::op::Subgraph";
     }
+    init_snippet_hash();
```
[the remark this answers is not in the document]
chenhu-wang: ok, init_body_hash() is better, changed, thank you!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -61,11 +153,18 @@ void Snippet::copy_snippet() {
...
+    snippetAttrs.snippet = std::make_shared<ngraph::snippets::op::Subgraph>(subgraph_node_inputs, new_body);
+    ngraph::copy_runtime_info(original_snippet, snippetAttrs.snippet);
+    snippetAttrs.snippet->set_friendly_name(original_snippet->get_friendly_name());
+    snippetAttrs.snippet->set_generator(std::make_shared<CPUGenerator>(host_isa));
+}
+
+void Snippet::init_snippet_hash() {
+    uint64_t seed = 0;
+    ov::pass::Manager m;
+    m.register_pass<ov::pass::Hash>(seed);
```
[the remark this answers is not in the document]
chenhu-wang: will do.

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -211,169 +321,41 @@ InferenceEngine::Precision Snippet::getRuntimePrecision() const {
...
         }
         return success;
     };
+
+    std::vector<bool> isBlockedInput(inputShapes.size(), false);
+    bool isBlockedMaster = false;
+    for (size_t i = 0; i < inputShapes.size(); i++) {
+        auto blockedShape = getBlockedShape(getParentEdgeAt(i)->getMemory().GetDescWithType<BlockedMemoryDesc>());
+        isBlockedInput[i] = std::get<0>(blockedShape).size() != std::get<1>(blockedShape).size();
+        isBlockedMaster = isBlockedMaster || isBlockedInput[i];
```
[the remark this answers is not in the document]
chenhu-wang: isBlockedInput and isBlockedMaster is done offline after get selected pd. Thank you Ivan!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: As the local body is modified through out shapeInfer(), canonicalization() etc based on previous input shapes. This lead for example some broadcast or other snippets dialect is insert. At current shapeInfer(), canonicalization() can not based on the modified body. seems like can not reshape back to original topology. I originally always use the single local body, there are lot of failures in test, so restart from original for new shape. Thank you!

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/add.cpp
```diff
@@ -14,8 +14,15 @@ namespace {
...
-std::vector<ov::Shape> inShapesStatic2{{1, 16, 29,  1}, {1, 16, 1, 1}, {1, 1, 1, 1}};
+std::vector<ov::test::InputShape> inShapesStatic1{{{}, {{1, 16, 29, 1}}},
+                                                  {{}, {{1, 16, 29, 7}}},
+                                                  {{}, {{1, 16, 29, 8}}},
+                                                  {{}, {{1, 16, 29, 15}}},
+                                                  {{}, {{1, 16, 29, 16}}},
+                                                  {{}, {{1, 16, 29, 31}}}};
+std::vector<ov::test::InputShape> inShapesStatic2{{{}, {{1, 16, 29, 1}}},
+                                                  {{}, {{1, 16, 1, 1}}},
+                                                  {{}, {{1, 1, 1, 1}}}};
```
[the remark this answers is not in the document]
IvanNovoselov: UPD: just discussed the issue with @dmitry-gorokhov. He suggests to go with the implemented approach and to rewrite all static tests to use InputShape's. The motivation behind that is to avoid tests class duplication for dynamic cases.

akladiev:

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/add.cpp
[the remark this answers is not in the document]
chenhu-wang: snippets test is updated with InputShape. Dynamic test case is added.
Note: Not cover MHA, MatMul related as recompilation to matmul takes long time, and critical for LLM customers. Team is seeking shape agnostic MHA solution.

src/plugins/intel_cpu/src/nodes/subgraph.h
chenhu-wang: done, thank you Ivan!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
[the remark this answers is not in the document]
chenhu-wang: @IvanNovoselov The static model load time test is linked on ticket. Thank you!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -43,63 +46,207 @@ namespace intel_cpu {
...
+        attrs.inMemPrecs.size() != rhs.attrs.inMemPrecs.size())
+        return false;
+    if (attrs.outMemBlockedDims.size() != rhs.attrs.outMemBlockedDims.size() ||
+        attrs.outMemOrders.size() != rhs.attrs.outMemOrders.size() ||
+        attrs.outMemPrecs.size() != rhs.attrs.outMemPrecs.size())
+        return false;
+
+    bool retVal = true;
+    for (size_t i = 0; i < attrs.inMemBlockedDims.size(); i++)
+        retVal = retVal && attrs.inMemBlockedDims[i] == rhs.attrs.inMemBlockedDims[i];
```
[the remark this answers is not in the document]
a-sidorova: Also probably we can improve the method if we order checks. I mean that `inMemPrecs` and `outMemPrecs` should be the end of method since precisions aren't changed after each infer request (or very rarely anyway?)? for example. @chenhu-wang what do you think?

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -43,63 +46,207 @@ namespace intel_cpu {
...
         const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
-        return {m_node->shapeInfer(), ShapeInferStatus::success};
+        auto broadcast_merge = [](VectorDims& dst, const VectorDims& src) {
+            // Ranks are both static.
+            auto dst_rank = dst.size();
+            auto src_rank = src.size();
+            const auto new_rank = std::max(dst_rank, src_rank);
+            dst.insert(dst.begin(), new_rank - dst_rank, 1);
+            bool success = true;
+            for (size_t i = 0; i < new_rank; i++) {
```
a-sidorova, line 144:
```suggestion
            for (size_t i = 0; i < new_rank && success; i++) {
```

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -43,63 +46,207 @@ namespace intel_cpu {
...
-#if defined(OPENVINO_ARCH_X86_64)
-    snippet->set_generator(std::make_shared<CPUGenerator>(host_isa));
-    isa_num_lanes =  snippet->get_generator()->get_target_machine()->get_lanes();
-#else
-    IE_THROW(NotImplemented) << "CPU plugin: code-generation is not supported on non-x64 platforms";
-#endif // OPENVINO_ARCH_X86_64
+    snippetAttrs.snippet = std::make_shared<snippets::op::Subgraph>(subgraph_node_inputs, new_body);
+    ov::copy_runtime_info(original_snippet, snippetAttrs.snippet);
+    snippetAttrs.snippet->set_friendly_name(original_snippet->get_friendly_name());
+    snippetAttrs.snippet->set_generator(std::make_shared<CPUGenerator>(host_isa));
```
a-sidorova, line 241: Seems like macros have been missed
```suggestion
#if defined(OPENVINO_ARCH_X86_64)
    snippetAttrs.snippet->set_generator(std::make_shared<CPUGenerator>(host_isa));
#else
    IE_THROW(NotImplemented) << "CPU plugin: code-generation is not supported on non-x64 platforms";
#endif // OPENVINO_ARCH_X86_64
```

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -43,63 +46,207 @@ namespace intel_cpu {
...
+}
+
+void Snippet::init_body_hash() {
+    uint64_t seed = 0;
+    ov::pass::Manager m;
+    m.register_pass<ov::pass::Hash>(seed);
+    m.run_passes(std::const_pointer_cast<ov::Model>(original_snippet->body_ptr()));
+    snippetAttrs.bodyHash = seed;
```
a-sidorova, line 249: Maybe it would be better
```suggestion
    uint64_t seed = 0;
    ov::pass::Hash hash_function(seed); 
    hash_function.run_on_model(original_snippet->body_ptr());
    snippetAttrs.bodyHash = seed;
```

Moreover as I know, to use `std::const_pointer_cast` is not good practice. So let's try to avoid this using if possible 🙂

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -506,50 +674,114 @@ void Snippet::prepareParams() {
...
-    }
-    if (getChildEdges().size() != 1) {
+ov::PartialShape Snippet::SnippetJitExecutor::reshapeCanonicalizedBody() {
+    ov::snippets::op::Subgraph::BlockedShapeVector input_blocked_shapes = getBlockedShapes(
+        snippetAttrs.inMemBlockedDims, snippetAttrs.inMemOrders, snippetAttrs.inMemPrecs);
+
+    ov::snippets::op::Subgraph::BlockedShapeVector output_blocked_shapes = getBlockedShapes(
+        snippetAttrs.outMemBlockedDims, snippetAttrs.outMemOrders, snippetAttrs.outMemPrecs);
+
+    const auto& canonicalShape = snippetAttrs.snippet->reshape_canonicalized_body(output_blocked_shapes, input_blocked_shapes);
```
a-sidorova, line 709: Seems like we can unite 2 methods `reshapeCanonicalizedBody` and `canonicalizeBody` into one `canonicalizeBody` but just add the parameter `reshape`. If `true`, use `reshape_canonicalized_body`. Otherwise use `canonicalize`.

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -567,71 +799,16 @@ void Snippet::generate(const jit_snippets_compile_args* jcp) {
...
-            for (ptrdiff_t j = work_size.size() - 2; j >= 0; j--) {
-                indexes[j] = tmp % work_size[j];
-                tmp /= work_size[j];
-            }
-
-            schedule.get_callable<kernel>()(indexes.data(), &call_args);
-        }
-    });
+bool Snippet::SnippetJitExecutor::schedule_created() {
+    return schedule.ptr == nullptr ? false : true;
```
a-sidorova, line 811:
```suggestion
    return schedule.ptr != nullptr;
```

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -119,12 +266,12 @@ void Snippet::initSupportedPrimitiveDescriptors() {
...
     const size_t ndims = outputShapes[0].getRank();
     // Domain sensitive operations support only Planar layout
-    const bool isOnlyPlanarApplicable = snippet->has_domain_sensitive_ops();
+    const bool isOnlyPlanarApplicable = snippetAttrs.snippet->has_domain_sensitive_ops();
     const bool isChannelsFirstApplicable = dnnl::impl::utils::one_of(ndims, 1u, 2u, 3u, 4u, 5u) && dimRanksAreEqual && !isOnlyPlanarApplicable;
     // Todo: Snippets currently don't support per-channel broadcasting of Blocked descriptors because
     //  canonicalization can't distinguish between <N, C, H, W, c> and <N, C, D, H, W> cases.
     //  See snippets::op::Subgraph::canonicalize for details.
-    bool isBlockedApplicable = dnnl::impl::utils::one_of(ndims,  4u, 5u) && dimRanksAreEqual && !isOnlyPlanarApplicable;
+    bool isBlockedApplicable = dnnl::impl::utils::one_of(ndims,  3u, 4u, 5u) && dimRanksAreEqual && !isOnlyPlanarApplicable;
```
a-sidorova, line 274: Do we have the tests for blocked 3D layout? If not, could you please add? Thanks!

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/add.cpp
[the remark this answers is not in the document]
a-sidorova: Maybe there is sense to use `static_shapes_to_test_representation` for static shape tests to avoid a lot of `{{}}`. What do you think?

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/ngram.cpp
```diff
@@ -207,7 +207,7 @@ class NgramCPUTest : public testing::WithParamInterface<NgramTestParams>, virtua
 
 TEST_P(NgramCPUTest, CompareWithRefs) {
     run();
-    CheckNumberOfNodesWithType(compiledModel, "Ngram", 1);
```
a-sidorova, line 210: Could you please explain why the number has been changed?

src/plugins/intel_cpu/src/nodes/subgraph.h
```diff
@@ -56,60 +67,94 @@ class Snippet : public Node {
     // Create a deep local copy of the input snippet to perform canonicalization & code generation
     // TODO: Probably better to implement a proper copy constructor
     // NOTE: Before call mutex should be initialized
```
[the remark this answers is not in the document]
chenhu-wang: ok, I see. Let me remove

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: Yes, we have some 3D block dynamic eltwise tests cases. After enable dynamic node, it's tokenized and run as subgraph, that's why I extend here. I can add some specific snippets 3d block test case. Thanks

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -236,7 +383,25 @@ void Snippet::initSupportedPrimitiveDescriptors() {
 
 void Snippet::selectOptimalPrimitiveDescriptor() {
     selectPreferPrimitiveDescriptor(getPrimitivesPriority(), true);
+    // memory order and precision is determined now, there is no need to prepare for each dynamic shapes.
+    const auto config = getSelectedPrimitiveDescriptor()->getConfig();
```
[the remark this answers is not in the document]
chenhu-wang: the fixed part of snippetAttrs(precision, layout) can be set only once as long as the best memory descriptor is selected. On other reasons, let me try move it. Thanks!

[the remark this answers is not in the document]
chenhu-wang: Moved to initOptimalPrimitiveDescriptor. Thanks Ivan.

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -249,204 +414,229 @@ InferenceEngine::Precision Snippet::getRuntimePrecision() const {
...
-                                 VectorDims &domain, size_t& TileRank) const {
-    const size_t minimalConcurrency = parallel_get_max_threads();
-    const size_t minimalJitWorkAmount = 256;
-    const size_t ds = domain.size();
-    if ( ds <= 2 || // not enough dimensions to collapse
-         domain[ds-1] >= minimalJitWorkAmount || // There is enough work for 1D Tiles, no need to collapse
-         domain[ds-1] * domain[ds-2] >= fullWorkAmount / minimalConcurrency) // There won't be enough work for every thread (even one iter) if we collapse
+void Snippet::prepareParams() {
+    auto inputSize = inputShapes.size();
+    snippetAttrs.inMemBlockedDims.resize(inputSize);
```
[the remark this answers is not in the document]
chenhu-wang: removed repeated resize. Thanks Ivan!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -249,204 +414,229 @@ InferenceEngine::Precision Snippet::getRuntimePrecision() const {
...
-                    if (TileRank < maxTileRank) {
-                        TileRank++;
-                        continue;
-                    }
+void Snippet::execute(dnnl::stream strm) {
+    if (!execPtr) {
+        IE_THROW() << "Can't execute Subgraph node. Primitive didn't created";
+    }
+    auto inputSize = inputShapes.size();
+    srcMemPtrs.resize(inputSize);
```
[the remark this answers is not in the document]
chenhu-wang: Applied. Thanks Ivan!

src/plugins/intel_cpu/tests/functional/single_layer_tests/conversion.cpp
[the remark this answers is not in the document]
chenhu-wang: do you think current eltwise dynamic test should be disabled for snippets tokenization?

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: Here success means the dst dimenion is updated, not a fail of broadcast. Need continue broadcast for next dimension. seems success here is not needed and removed. welcome comments if concerns. thanks Alexandra!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: yes, applied, thanks Alexandra!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: done.

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: There are already 3d block dynamic test, such as in https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/tests/functional/subgraph_tests/src/eltwise_caching.cpp#L531. If snippets test disable eltwise dynamic test, I will add separate cases. What do you think? Thanks Alexandra!

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/ngram.cpp
chenhu-wang: Ngram is internal node and it's fused from graph based on some pattern. After tokenization, the pattern can not match, so there is no ngram node. Maybe I need disable tokenization for this case, it's intended for ngram test. Thank you Alexandra.

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -43,65 +46,218 @@ namespace intel_cpu {
...
-    Snippet* m_node;
+    std::shared_ptr<ov::Model> m_body;
 };
 
 class SnippetShapeInferFactory : public ShapeInferFactory {
 public:
-    SnippetShapeInferFactory(Snippet* node) : m_node(node) {}
+    SnippetShapeInferFactory(const std::shared_ptr<ov::Node>& op) {
+        auto subgraph = ov::as_type_ptr<snippets::op::Subgraph>(op);
+        snippet_body = subgraph->body_ptr()->clone();
```
IvanNovoselov, line 214: Why do we need to clone body function here? Can't we just copy the `body_ptr()` return value?

IvanNovoselov: We may have to do it, if some operations from snippets opset break shape propagation. If this is the case, could you identify such ops, so we can fix their `validate_and_infer_types`?

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/ngram.cpp
a-sidorova: I'd suggest to disable Snippets here to allow test to validate `Ngram` implementation. To validate subgraph we have eltwise tests and snippets tests. You can disable them using config (like [here](https://github.com/openvinotoolkit/openvino/blob/52b9df4a6d7cf80833c0e441c11fb80f12d6bfce/src/tests/functional/plugin/shared/src/snippets/mha.cpp#L58) but with `DISABLE` key). Thank you!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: The reason of a clone() here is that body_ptr() is shared across streams. Different streams should have its own shape infer with possibly different input shapes, should not impact each other. Am I right?

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/ngram.cpp
chenhu-wang: done, thank you!

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/add.cpp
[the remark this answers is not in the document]
chenhu-wang: The unified paramSet structure with
[inputShape(using InputShape = std::pair<ov::PartialShape, std::vector<ov::Shape>>;)](https://github.com/openvinotoolkit/openvino/blob/master/src/tests/functional/shared_test_classes/include/shared_test_classes/base/ov_subgraph.hpp#L16)
is shared for static and dynamic test cases. If define two paramSet structure, seems like we need two setup() and getName(). I am not sure it's deserve to do it. Maybe we can use static_shapes_to_test_representation for test only include static test cases. Thank you Alexandra!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
IvanNovoselov: Yes, that's correct. Previously we solved this problem by making a deep copy in the Snippet node constructor.
Now I see that you moved the deep copy from constructor to SnippetShapeInferFactory, so we still have one copy per node.
That's all right as well, so let's leave as is.

src/plugins/intel_cpu/tests/functional/single_layer_tests/conversion.cpp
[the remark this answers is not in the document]
IvanNovoselov: @dmitry-gorokhov suggested that we disable snippets in original dynamic eltwise single-layer tests, since dynamic eltwise ops will be evaluaeted by plugin for now (because dynamic snippets are slower than plugin for now).
However, we still need to test snippets dynamic pipeline, so we can create separate test instances with `SNIPPETS_MODE = IGNORE_CALLBACK`.

src/plugins/intel_cpu/src/nodes/subgraph.cpp
a-sidorova: Should we throw exception anyway if dimensions aren't broadcastable? What do you think?

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -250,204 +433,219 @@ InferenceEngine::Precision Snippet::getRuntimePrecision() const {
...
+
+    ov::PartialShape canonicalShape;
+    if (is_canonicalized) {
+        // just reshape with new input shapes, and get updated master shape
+        canonicalShape = canonicalizeBody(true);
+    } else {
+        // determine canonicalize, determine master_shape
+        // canonicalShape and compute_master_shape is always on snippetAttrs.snippet, only generation on a copy.
+        canonicalShape = canonicalizeBody(false);
     }
```
a-sidorova, line 616: Minor
```suggestion
    ov::PartialShape canonicalShape = canonicalizeBody(is_canonicalized);
```

src/common/snippets/src/op/subgraph.cpp
```diff
@@ -388,27 +412,30 @@ ov::PartialShape snippets::op::Subgraph::canonicalize(const BlockedShapeVector&
...
                                                                                  ::ov::op::AutoBroadcastType::NUMPY);
             NODE_VALIDATION_CHECK(this, compatibleWithOtherOutputs,
                                   "Snippets output shapes must be numpy broadcastable");
         }
     }
-
-    // We should insert Converts after Parameters and Constant and before Results
-    // to align precision inside Subgraph body that is supported by Plugin
-    align_element_types(outputShapes, inputShapes);
+    if (!reshape) {
```
a-sidorova, line 434: Honestly, it's not clear why if not `reshape`, we shouldn't insert Converts (I understand but another developer without background will be confused).
On the [previous](https://github.com/openvinotoolkit/openvino/pull/15430#discussion_r1230494695) comment I meant to unite only plugin methods but to leave the Snippets so because now the method `snippets::op::Subgraph::canonicalize` looks complicated now. What do you think?

src/frontends/tensorflow/tests/compilation.cpp
```diff
@@ -20,7 +20,6 @@ TEST_F(CompileModelsTests, NgramCompilation) {
     ov::CompiledModel compiled_model = core.compile_model(model, "CPU");
     const auto runtime_model = compiled_model.get_runtime_model();
 
-    EXPECT_EQ(runtime_model->get_ordered_ops().size(), 4);
     EXPECT_EQ(runtime_model->get_parameters().size(), 2);
```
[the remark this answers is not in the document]
a-sidorova: Maybe we can provide the plugin config with disabled Snippets to allow tests to validate Ngram? @chenhu-wang @IvanNovoselov what do you think?

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/codegen_gelu.cpp
```diff
@@ -13,15 +13,42 @@ namespace ov {
...
+        ::testing::Values(ov::element::f32),
+        ::testing::ValuesIn(inShapes0),
+        ::testing::ValuesIn(inShapes1),
+        ::testing::Values(true, false),
+        ::testing::Values(CommonTestUtils::DEVICE_CPU)),
+        CodegenGelu::getTestCaseName);
+
+// DS
+std::vector<InputShape> inShapesDynamic0{
+    {{-1, -1, -1}, {{1, 384, 4096}, {1, 384, 1}, {1, 384, 4096}}},
```
a-sidorova, line 36: I'd suggest to reduce Shapes to avoid big execution for simple test (the main idea - to have innermost Loop with several iterations, so you can set shape smth like `1x12x128`). What do you think?

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/add.cpp
```diff
@@ -26,41 +32,77 @@ INSTANTIATE_TEST_SUITE_P(smoke_Snippets_Eltwise, Add,
...
+        {{{}, {{1, 128, 1, 9}}}, {{}, {{1, 128, 1, 9}}}},
+        {{{}, {{1, 128, 1, 16}}}, {{}, {{1, 128, 1, 16}}}},
+        {{{}, {{1, 128, 1, 17}}}, {{}, {{1, 128, 1, 17}}}},
+        {{{}, {{1, 128, 1, 29}}}, {{}, {{1, 128, 1, 29}}}},
+        {{{}, {{1, 128, 1, 33}}}, {{}, {{1, 128, 1, 33}}}},
+        {{{}, {{1, 128, 9, 30}}}, {{}, {{1, 128, 1, 30}}}},
+        {{{}, {{1, 128, 9, 1}}}, {{}, {{1, 128, 1, 30}}}},
+        {{{}, {{1, 128, 9, 16}}}, {{}, {{1, 128, 9, 1}}}},
+        // DS
+        {{{1, -1, {1, 10}, {1, 33}}, {{1, 128, 1, 1}, {1, 128, 1, 9}, {1, 128, 1, 17}, {1, 128, 1, 29}, {1, 128, 9, 1}, {1, 128, 1, 1}}},
```
a-sidorova, line 72: Could you please add tests with sequence of the same shapes? To validate how Snippets work if we provide the same input shapes several times in a row

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: I agree, added, thanks Alexandra!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: updated, thanks Alexandra!

src/common/snippets/src/op/subgraph.cpp
chenhu-wang: I agree, created separate snippets function canonicalized_body_shape_infer() for reshape case. thanks Alexandra!

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/codegen_gelu.cpp
chenhu-wang: updated, also for static shape case as seems like the big size is not intentional and not a must have.

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/add.cpp
chenhu-wang: added, thanks Alexandra!

src/frontends/tensorflow/tests/compilation.cpp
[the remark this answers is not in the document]
chenhu-wang: after [disable ](https://github.com/openvinotoolkit/openvino/pull/15430#issuecomment-1618252848)tokenization for dynamic in fallback, seem like  we do not need set config.

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
```diff
@@ -728,13 +734,29 @@ void Transformations::MainSnippets(void) {
...
                                                         [&](const ov::Output<const ov::Node>& out) {
                                                             return rank_is_too_large(out.get_tensor());
                                                         });
-                return has_only_const_inputs || bad_input_rank || bad_output_rank || is_unsupported_swish ||
-                    is_disabled_tokenization;
+                if (bad_output_rank)
+                    return true;
+                const bool has_dynamic_input = std::any_of(inputs.begin(), inputs.end(),
+                                                        [&](const ov::Input<const ov::Node>& in) {
+                                                            return !in.get_tensor().get_partial_shape().is_static();
```
a-sidorova, line 748: We already have the check for dynamic shapes 🙂 
```suggestion
                                                            return in.get_partial_shape().is_dynamic();
```

BTW The `ov::Node` has interface `is_dynamic()` which check for dynamic shapes of inputs. Since Snippets don't support nodes with internal dynamism we don't expect to get node with input static shapes but dynamic output shapes (I understand that probably it will be supported in the future but you have already insert early `returns` to make callback quicker). What do you think can we reuse here just `node->is_dynamic()`?

Also there may be sense to reorder checks here. For example, insert `checks for dynamism` earlier than checks for `HSwish`. Maybe not in this PR...

a-sidorova (approved):

src/frontends/tensorflow/tests/compilation.cpp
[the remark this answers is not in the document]
chenhu-wang: cpu plugin fallback in [87c4a73](https://github.com/openvinotoolkit/openvino/pull/15430/commits/87c4a73bd1b99500158cfbc84bce2c992afbb738), so leave ngram untouched.

src/plugins/intel_cpu/tests/functional/single_layer_tests/conversion.cpp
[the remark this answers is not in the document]
chenhu-wang: Eltwise single layer test have both plugin node and snippets backend now. snippets test is enforce with snippets backend. in [87c4a73](https://github.com/openvinotoolkit/openvino/pull/15430/commits/87c4a73bd1b99500158cfbc84bce2c992afbb738), Thanks!

src/plugins/intel_cpu/src/nodes/subgraph.cpp
chenhu-wang: Now 3d eltwise test have snippets backend. Thanks.

chenhu-wang:
@IvanNovoselov , @ilyachur , could you please take a look snippets custom hash pass changes in [a86f9d1](https://github.com/openvinotoolkit/openvino/pull/15430/commits/a86f9d16264bf6bf98f63f89626ea6159ce3235c), thanks!

src/core/src/pass/serialize.cpp
```diff
@@ -127,7 +128,8 @@ void ngfunction_2_ir(pugi::xml_node& node,
                      const std::map<std::string, ngraph::OpSet>& custom_opsets,
                      ConstantWriter& constant_write_handler,
                      int64_t version,
-                     bool deterministic);
+                     bool deterministic,
+                     bool ignore_names);
```
ilyachur, line 132: I expect that you don't need to modify this file in case of custom pass.

Please revert all changes!

chenhu-wang: I think so, will remove after finish a custom model hash.

chenhu-wang: removed, thanks Ilya!

src/core/CMakeLists.txt
```diff
@@ -48,7 +48,8 @@ target_include_directories(ov_core_dev INTERFACE
     $<BUILD_INTERFACE:${OV_CORE_INCLUDE_PATH}>
     $<BUILD_INTERFACE:${OpenVINO_SOURCE_DIR}/src/core/dev_api>
     $<BUILD_INTERFACE:${OpenVINO_SOURCE_DIR}/src/frontends/common/include>
-    $<BUILD_INTERFACE:${OpenVINO_SOURCE_DIR}/src/common/transformations/include>)
+    $<BUILD_INTERFACE:${OpenVINO_SOURCE_DIR}/src/common/transformations/include>
+    $<BUILD_INTERFACE:${OpenVINO_SOURCE_DIR}/src/common/snippets/include>)
```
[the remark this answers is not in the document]
chenhu-wang: removed!

src/core/src/pass/serialize.cpp
```diff
@@ -1292,4 +1304,27 @@ bool pass::Hash::run_on_model(const std::shared_ptr<ov::Model>& model) {
 
 pass::Hash::Hash(uint64_t& output_hash_value) : m_hash(output_hash_value) {}
 
+bool snippets::pass::Hash::run_on_model(const std::shared_ptr<ov::Model>& model) {
```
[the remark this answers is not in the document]
chenhu-wang: done in [c224dc1](https://github.com/openvinotoolkit/openvino/pull/15430/commits/c224dc1809ea96801623617bc60d2f247b8c3d54), could you please take a look?

src/core/src/pass/serialize.cpp
```diff
@@ -1292,4 +1304,27 @@ bool pass::Hash::run_on_model(const std::shared_ptr<ov::Model>& model) {
...
+    using namespace ov::pass;
+
+    RUN_ON_MODEL_SCOPE(Hash);
+    OstreamHashWrapper xmlHash;
+    OstreamHashWrapper binHash;
+    std::ostream xml(&xmlHash);
+    std::ostream bin(&binHash);
+
+    // Determinism is important for hash calculation
+    serializeFunc(xml, bin, model, Serialize::Version::UNSPECIFIED, {}, true, true);
```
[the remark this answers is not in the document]
chenhu-wang: done in [c224dc1](https://github.com/openvinotoolkit/openvino/pull/15430/commits/c224dc1809ea96801623617bc60d2f247b8c3d54), could you please take a look?

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+    void on_adapter(const std::string& name, ov::ValueAccessor<std::shared_ptr<ov::Model>>& adapter) override {
+        if (name == "net") {
+            ovfunction_2_hash(m_hash, *adapter.get());
+        } else {
+            OPENVINO_THROW("Unsupported Model name.");
+        }
+    }
+};
+
+const std::unordered_map<ov::Node*, int> create_layer_ids(const ov::Model& model) {
```
IvanNovoselov, line 239: Minor: top-level const qualifier makes a little sense, since it doesn't prevent returned object from modification [details](https://stackoverflow.com/questions/1579435/should-useless-type-qualifiers-on-return-types-be-used-for-clarity). Also on L248

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+        int port_id = 0;
+        // <layers/input>
+        if (node->get_input_size() > 0) {
+            hash = hash_combine(hash, std::string("input"));
+            for (auto& i : node->inputs()) {
+                hash = hash_combine(hash, std::string("port"));
+                hash = hash_combine(hash_combine(hash, std::string("id")), port_id++);
+                hash = hash_combine(hash_combine(hash, std::string("precision")), get_precision_name(i.get_element_type()));
+                for (auto d : i.get_partial_shape()) {
```
IvanNovoselov, line 409: `PartialShape` has [to_string()](https://github.com/openvinotoolkit/openvino/blob/f4fe8400a71077491fef94beafc5e659cae1a3fb/src/core/include/openvino/core/partial_shape.hpp#L355) method, so we don't have to convert it to string manually here.

Here, and also on L429-435

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+    // Serialize rt info
+    hash = hash_combine(hash, std::string("rt_info"));
+    for (const auto& it : model.get_rt_info()) {
+        hash_rt_info(hash, it.first, it.second);
+    }
+}
+
+void snippets_to_hash(uint64_t& hash,
```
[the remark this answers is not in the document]
chenhu-wang: removed this function, thanks.

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+};
+
+} // namespace rt_info
+
+class SnippetsHasher : public ov::AttributeVisitor {
+    uint64_t& m_hash;
+    const std::string& m_node_type_name;
+
+    template <typename T>
+    std::string create_atribute_list(ov::ValueAccessor<std::vector<T>>& adapter) {
```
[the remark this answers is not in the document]
chenhu-wang: fixed, thanks!

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+                m_hash = hash_combine(hash_combine(m_hash, attr.first), attr.second);
+            }
+        } else if (const auto& a = ov::as_type<ov::AttributeAdapter<ov::element::TypeVector>>(&adapter)) {
+            const auto& attrs = a->get();
+            m_hash = hash_combine(hash_combine(m_hash, name), join(attrs));
+        } else if (const auto& a = ov::as_type<ov::AttributeAdapter<ov::PartialShape>>(&adapter)) {
+            const auto& attrs = a->get();
+            auto shape_str = attrs.to_string();
+            if (shape_str[0] == '[' && shape_str[shape_str.size() - 1] == ']')
+                shape_str = shape_str.substr(1, shape_str.size() - 2);
```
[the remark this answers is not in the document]
chenhu-wang: My original intention is based on assumption that [1,3,2,2] should be the same as 1,3,2,2. The two shape should generate the same hash, to increase hit probability. But I am not sure in some special case or in the future, such as 1 and {1} is the same. vector with 1 element and scalar maybe different. Removed the strip now, thanks Ivan!

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+        : m_hash(hash),
+          m_node_type_name(node_type_name) {}
+
+    void on_adapter(const std::string& name, ov::ValueAccessor<void>& adapter) override {
+        if (const auto& a = ov::as_type<ov::AttributeAdapter<std::shared_ptr<ngraph::Variable>>>(&adapter)) {
+            m_hash = hash_combine(hash_combine(m_hash, name), a->get()->get_info().variable_id);
+        } else if (const auto& a =
+                       ov::as_type<ov::AttributeAdapter<std::shared_ptr<ngraph::runtime::AlignedBuffer>>>(&adapter)) {
+            if (name == "value" && m_node_type_name == "Constant") {
+                m_hash = hash_combine(m_hash, std::string("Constant"));
```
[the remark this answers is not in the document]
chenhu-wang: done, thanks Ivan!

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+struct Edge {
+    int from_layer = 0;
+    int from_port = 0;
+    int to_layer = 0;
+    int to_port = 0;
+};
+
+template <typename T>
+static uint64_t hash_combine(uint64_t seed, const T& a) {
```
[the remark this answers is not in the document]
chenhu-wang: removed, thanks Ivan!

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+        m_hash = hash_combine(hash_combine(m_hash, name), create_atribute_list(adapter));
+    }
+    void on_adapter(const std::string& name, ov::ValueAccessor<std::vector<float>>& adapter) override {
+        m_hash = hash_combine(hash_combine(m_hash, name), create_atribute_list(adapter));
+    }
+    void on_adapter(const std::string& name, ov::ValueAccessor<std::vector<std::string>>& adapter) override {
+        m_hash = hash_combine(hash_combine(m_hash, name), create_atribute_list(adapter));
+    }
+    void on_adapter(const std::string& name, ov::ValueAccessor<std::shared_ptr<ov::Model>>& adapter) override {
+        if (name == "net") {
```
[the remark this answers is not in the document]
chenhu-wang: removed, thanks Ivan!

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+                result.emplace_back(node);
+        }
+        for (const auto& sink : model.get_sinks()) {
+            result.emplace_back(sink);
+        }
+        for (const auto& res : model.get_results()) {
+            result.emplace_back(res);
+        }
+        sorted_ops = result;
+    }
```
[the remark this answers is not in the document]
chenhu-wang: removed, thanks Ivan!

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+    for (const auto& n : sorted_ops) {
+        ov::Node* node = n.get();
+        const std::string& node_type_name{node->get_type_name()};
+
+        OPENVINO_ASSERT(layer_ids.find(node) != layer_ids.end(), "Internal error");
```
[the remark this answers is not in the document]
chenhu-wang: done, thanks Ivan.

src/common/snippets/src/pass/hash.cpp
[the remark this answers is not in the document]
chenhu-wang: done, thanks Ivan!

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,487 @@
...
+        // <layers/output>
+        if ((node->get_output_size() > 0) && !ov::op::util::is_output(node)) {
+            hash = hash_combine(hash, std::string("output"));
+            // pugi::xml_node output = layer.append_child("output");
+            for (auto& o : node->outputs()) {
+                hash = hash_combine(hash, std::string("port"));
+                hash = hash_combine(hash_combine(hash, std::string("id")), port_id++);
+                hash = hash_combine(hash_combine(hash, std::string("precision")), get_precision_name(o.get_element_type()));
+
+                for (auto d : o.get_partial_shape()) {
```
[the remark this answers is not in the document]
chenhu-wang: done, thanks Ivan!

src/common/snippets/src/pass/hash.cpp
chenhu-wang: done, thanks Ivan!

src/common/snippets/src/pass/hash.cpp
chenhu-wang: nice,  `PartialShape` `to_string()` is used, thanks Ivan!

IvanNovoselov (approved):

src/common/snippets/src/pass/hash.cpp
```diff
@@ -0,0 +1,453 @@
...
+            edges.push_back(e);
+        }
+    }
+    std::sort(begin(edges), end(edges), [](const Edge& a, const Edge& b) -> bool {
+        return a.from_layer < b.from_layer;
+    });
+    return edges;
+}
+
+std::string get_precision_name(const ov::element::Type& elem_type) {
```
dmitry-gorokhov, line 282: `ov::element::Type` already has overrided `<<` operator  and even `hash()` method. Can we reuse smt instead?

chenhu-wang: That's great! element::Type hash() is used. Thanks Dmitry.

src/plugins/intel_cpu/src/shape_inference/custom/subgraph.hpp
```diff
@@ -10,34 +10,91 @@ namespace ov {
 namespace intel_cpu {
 namespace node {
 using Result = IShapeInfer::Result;
-/* This class implementation is a temporal WA
-   TODO: revise the implementation to remove the node reference*/
 class SnippetShapeInfer : public ShapeInferEmptyPads {
 public:
-    SnippetShapeInfer(Snippet* node) : m_node(node) {}
+    SnippetShapeInfer(std::shared_ptr<ov::Model> body) : m_body(body) {}
```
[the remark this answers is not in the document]
chenhu-wang: In general case, subgraph's `validate_nodes_and_infer_types` is actually used: https://github.com/openvinotoolkit/openvino/pull/15430/files#diff-0cedaaf60bc2518ef9628e775185635df0ddc5f4b1fad1a7de1ca6aa7e875320R65,
Just in case of specific(one output), A quicker one is implemented. Unified validate_nodes_and_infer_types should be used if it's optimized in the future.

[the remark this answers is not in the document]
dmitry-gorokhov: I want to avoid Snippets users to implement custom code for shape inference themselves. 
We have PR that adopts new memory accessors for shape inference: https://github.com/openvinotoolkit/openvino/pull/17719. That should significantly improve shape inference via IStaticShapeInfer interface. 
If custom impl still will be required it cab be implemented reusing existing approach https://github.com/openvinotoolkit/openvino/tree/master/src/core/shape_inference/include, but placed into snippets component.

dmitry-gorokhov (approved):

src/plugins/intel_cpu/src/shape_inference/custom/subgraph.hpp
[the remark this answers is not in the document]
IvanNovoselov: @dmitry-gorokhov, I agree that shape inference implementation should be hidden inside op::Subgraph. However we still need a wrapper of some kind to bridge cpu and snippets shape_inference, but it should be as simple as forwarding arguments: subgraph->shape_infer(....), or smth like that.
Anyway, snippets shape-inference will be refactored and moved to op::Subgraph in [PR 18563](https://github.com/openvinotoolkit/openvino/pull/18563), so I would be happy to address your concern there.

PR #15436, opened 2023-02-01
Added compilation flag to build with spectre mitigated libraries

gkrivor:
### Details:
 - Added flag to build with spectre mitigated libraries for prevent usage of Spectre and Meltdown vulnerabilities.

### Tickets:
 - 102790

### DIFF
--- a/cmake/developer_package/compile_flags/os_flags.cmake
+++ b/cmake/developer_package/compile_flags/os_flags.cmake
@@ -389,6 +389,8 @@ if(WIN32)
         ie_add_compiler_flags(/Qdiag-disable:161,177,556,1744,1879,2586,2651,3180,11075,15335)
     endif()
 
+	ie_add_compiler_flags(/Qspectre)
+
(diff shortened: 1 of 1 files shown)


cmake/developer_package/compile_flags/os_flags.cmake
```diff
@@ -333,6 +333,8 @@ if(WIN32)
         ie_add_compiler_flags(/Qdiag-disable:161,177,556,1744,1879,2586,2651,3180,11075,15335)
     endif()
 
+	ie_add_compiler_flags(/Qspectre)
```
ilyachur, line 336: Please add explanation why do we need this flag

gkrivor: Updated title and details. Is it enough?

ilyachur (approved):

ilyachur:
> does it affect performance?

Performance numbers are in the ticket 106464

PR #15447, opened 2023-02-02
[GPU] Fix unit test generator

tuxedcat:
### Details:
 - Fix it to generates FLOAT16 correctly
   - `VF<uint16_t> rnd_vec = generate_random_1d<uint16_t>(s.count(), -1, 1);`
   - float16 is converted from uint16 bits. This might contain NaN value. We should use proper random generator.
 - Fix it to generates values for blocked format correctly
   - When data is not aligned with block size(e.g. fsv16 [b:1, f:15, x:2, y:1]), tensor should be filled with `linear_offset` function. But we are filling with raw data buffer index with data size.
   - FillType::All - Generally recommended
   - FillType::OnlyInsideShape - Required for reduce_sum because reduce kernel assumes the margin to be filled with zero.

### Tickets:
 - 67491


tuxedcat:
Consideration: Should it generate 0 value for the case that index is outside of shape but valid memory? Or is a random value sufficient?

tuxedcat:
We cannot fix fp16 generator now because it makes many tests to fail(not only onednn, cldnn also fails). Therefore I close this PR.

akladiev:

akladiev:

PR #15459, opened 2023-02-02
[MO]: fix mo.convert_model for pytorch

eaidova:
### Details:
Fix for issue:
```python
model = torchvision.models.resnet50(pretrained=True)
mo.convert_model(model, example_input=torch.zeros(1, 3, 224, 224))
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "openvino/tools/mo/openvino/tools/mo/convert.py", line 55, in convert_model
    ov_model, _ = _convert(cli_parser, framework, args)
TypeError: cannot unpack non-iterable openvino._pyopenvino.Model object*

### DIFF
--- a/tools/mo/openvino/tools/mo/convert_impl.py
+++ b/tools/mo/openvino/tools/mo/convert_impl.py
@@ -965,13 +965,14 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
                 args['onnx_opset_version'] = None
 
                 try:
-                    ov_model = _convert(cli_parser, framework, args)
+                    ov_model, argv = _convert(cli_parser, framework, args)
                 except Exception as e:
                     remove_tmp_onnx_model(out_dir)
                     raise e
 
                 remove_tmp_onnx_model(out_dir)
-                return ov_model, None
+
+                return ov_model, argv
 
         argv = pack_params_to_args_namespace(args, cli_parser)
 
--- a/tools/mo/openvino/tools/mo/convert_impl.py
+++ b/tools/mo/openvino/tools/mo/convert_impl.py
@@ -930,7 +930,7 @@ def remove_tmp_onnx_model(out_dir):
 def _convert(cli_parser: argparse.ArgumentParser, framework, args):
     if 'help' in args and args['help']:
         show_mo_convert_help()
-        return None
+        return None, None
 
     telemetry = tm.Telemetry(tid=get_tid(), app_name='Model Optimizer', app_version=get_simplified_mo_version())
     telemetry.start_session('mo')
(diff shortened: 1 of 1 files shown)


tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -971,7 +971,7 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
                     raise e
 
                 remove_tmp_onnx_model(out_dir)
-                return ov_model
+                return ov_model, None
```
rkazants, line 974: the second returned value is `argv`. It shoud not be `None` in this case

rkazants (commented):
one comment about the returned value
and please add test to test this case

tools/mo/openvino/tools/mo/convert_impl.py
eaidova: argv parsing becomes later, after this return, if it should not be, then I can move it

eaidova: argv parsing becomes later, after this return, if it should not be, then I can move it

rkazants: yes, correct. But this is the second issue and parsing or at least valid arguments should be returned.
`None` is not correct value in this case.

eaidova: looks like it contains some parameters which are not accepted by cli parser, attempt to parse it before model creation:
```
AttributeError: 'Namespace' object has no attribute 'example_input'
```
@popovaan probably I need your help to understand this functionality design and exdpected behaviour

eaidova:
> one comment about the returned value and please add test to test this case

there is set of tests which should find this issue (looks like they are not in precommit)

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -965,13 +965,14 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args):
...
-                    ov_model = _convert(cli_parser, framework, args)
+                    ov_model, argv = _convert(cli_parser, framework, args)
                 except Exception as e:
                     remove_tmp_onnx_model(out_dir)
                     raise e
 
                 remove_tmp_onnx_model(out_dir)
-                return ov_model
+
+                return ov_model, argv
```
popovaan, line 975: This does not change the logic of method. 
In previous version ov_model contained tuple with (ov_model, argv) and this tuple was returned as result.
Renaming looks good, but it is not the root cause of the issue.

popovaan:
> > one comment about the returned value and please add test to test this case
> 
> there is set of tests which should find this issue (looks like they are not in precommit)

Tests are in precommit. You can find them in openvino-lin job in group "MO Python API Tests - Import PyTorch model from memory".

PR #15015, opened 2023-01-09
[GNA] Changed max layer limit tests to avoid SEH exceptions

mryzhov:
### Details:
 - Test model should be spitted on several branches to avoid stack overflow error on windows in legacy DFS function 

### Tickets:
 - *98431*

PR #15460, opened 2023-02-02
[GNA] Changed max layer limit tests to avoid SEH exceptions (#15015)

mryzhov:
### Details:
 -  Merge of #15015 

### Tickets:
 - *98431*


ilya-lavrenov (approved):

src/plugins/intel_gna/tests/functional/limitations/layers_limit.cpp
```diff
@@ -40,34 +40,34 @@ class GNALayersLimitTest : public testing::WithParamInterface<GNALayersLimitTest
...
+            std::vector<std::shared_ptr<Add>> add_nodes;
+            add_nodes.push_back(std::make_shared<Add>(add_const, params[i]));
 
-        for (size_t i = 0; i < (layers_number - 2) / 2; ++i) {
-            auto add_next_x = std::make_shared<ngraph::opset8::Add>(add_nodes_x.back(), params[0]);
-            auto add_next_y = std::make_shared<ngraph::opset8::Add>(add_nodes_y.back(), params[1]);
-            add_nodes_x.push_back(add_next_x);
-            add_nodes_y.push_back(add_next_y);
+            for (size_t j = 0; j < (layers_number - branch_count) / branch_count; ++j) {
+                add_nodes.push_back(std::make_shared<Add>(add_nodes.back(), params[i]));
```
marcinkusm, line 65: are you sure that number of final layers in Gna2Model will be the same as in input params?

src/plugins/intel_gna/tests/functional/limitations/layers_limit.cpp
```diff
@@ -40,34 +40,34 @@ class GNALayersLimitTest : public testing::WithParamInterface<GNALayersLimitTest
 protected:
     void SetUp() override {
         ov::element::Type net_type = ov::element::f32;
-        auto params = ngraph::builder::makeParams(net_type, {{1, 2}, {1, 2}});
```
marcinkusm, line 43: Change looks fine for me.

I just wonder how we should proceeds with core issue itself.
AFAIK the problem is too many recurrent invocation of DFS when applying scale factor.
It happens on windows due the fact default stack size for app is 1MB when on Linux is 8MB.

marcinkusm (approved):
Looks good to me.

src/plugins/intel_gna/tests/functional/limitations/layers_limit.cpp
```diff
@@ -9,7 +9,7 @@
 #include "shared_test_classes/base/layer_test_utils.hpp"
 
 using namespace InferenceEngine;
-// using namespace ov::test;
+using namespace ov::opset8;
```
marcinkusm, line 12: opset9?

src/plugins/intel_gna/tests/functional/limitations/layers_limit.cpp
mryzhov: This is legacy code that we will be deprecated, the issue will not be applicable after migration to ngraph

src/plugins/intel_gna/tests/functional/limitations/layers_limit.cpp
mryzhov: This is merged fix, the same test in master

PR #15476, opened 2023-02-02
[IE][nG][LoadTime] speed up constant folding

pavel-esir:
### Details:
 - got rid of redundant copying during constant fold: created `HostTensorVector` directly from constant data pointer instead of packing into `ov::Tensor`s and `TensorVector`s with copies
 - distilbert-base-cased on 12th Gen Intel(R) Core(TM) i9-12900K load time reduced by ~**20%** from ~**270 ms** to ~**200 ms**

### Tickets:
 - xxx-102751


ilya-lavrenov:
> distilbert-base-cased on 12th Gen Intel(R) Core(TM) i9-12900K load time reduced by ~20% from ~270 ms to ~200 ms

How much time from these 200ms CF takes now? I mean if we optimize some reference implementations, can we achieve even better results?

src/core/src/node.cpp
```diff
@@ -790,26 +790,28 @@ bool ov::Node::constant_fold(OutputVector& output_values, const OutputVector& in
         return false;
 
     NodeVector nodes;
-    TensorVector input_tensors;
+    HostTensorVector input_tensors;
```
ilyachur, line 793: I disagree with this change because you remove code which works with `ov::Tensor` and replace it by code which uses `ngraph::HostTensor` (but `HostTensor` will be removed soon).

ilyachur: The better option is to migrate evaluates from `HostTensor` to `ov::Tensor`

pavel-esir: ok, i'll do that

pavel-esir:
> How much time from these 200ms CF takes now? I mean if we optimize some reference implementations, can we achieve even better results?

After this changes, from **~200** ms of load/compile model CF takes **~50 ms**, most of which is `Convert::evaluate`. Attached screen of a trace
redundant_copying_traces

Looks like optimizing other reference implementations will not give significant improvements

src/core/include/openvino/op/convert.hpp
```diff
@@ -41,7 +41,7 @@ class OPENVINO_API Convert : public Op {
     }
 
     OPENVINO_SUPPRESS_DEPRECATED_START
-    bool evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const override;
+    bool evaluate(ov::TensorVector& output_values, const ov::TensorVector& input_values) const override;
```
ilya-lavrenov, line 44: I suppose it should not be wrapped with `OPENVINO_SUPPRESS_DEPRECATED_START` now

pavel-esir: removed

pavel-esir:
- Corrected Convert evaluate to use `ov::Tensor`s instead of `HostTensorPtr`s. 
- Added Const constructor to create Consts from data pointers shared with `ov::Tensor`. 
- Corrected Node::constant_fold so that for getting and creating Consts data blobs are passed by reference instead of value (correction from the initial solution is that now we use the new `ov::Tensor` for that, instead of a deprecated `HostTensor`)
@ilyachur @ilya-lavrenov please take a look

src/core/src/node.cpp
pavel-esir: done. Please take a look

pavel-esir:
returned back old Convert's old evaluate with `HostTensor`s because they are used in tests accuracy comparator. Because of that some pruning unit-tests were failing, e.g. `PropagateMasksFakeQuantizePerChannel` and so on.

ilyachur (approved):

ilya-lavrenov (approved):

PR #15487, opened 2023-02-03
[TF FE] Refactor LeakyRelu and add layer tests

rkazants:
**Details:** Refactor LeakyRelu and add layer tests
 
**Ticket:** 92787

### DIFF
--- a/src/frontends/tensorflow_common/src/op/leaky_relu.cpp
+++ b/src/frontends/tensorflow_common/src/op/leaky_relu.cpp
@@ -13,12 +13,13 @@ namespace frontend {
 namespace tensorflow {
 namespace op {
 ov::OutputVector translate_leaky_relu_op(const NodeContext& node) {
-    auto in = node.get_input(0);
-    auto alpha_attr = node.get_attribute<float>("alpha", 0.f);
-    auto alpha_const = make_shared<Constant>(element::f32, Shape{1}, alpha_attr);
-
-    auto leaky_relu = make_shared<PRelu>(in, alpha_const);
-    return leaky_relu->outputs();
+    default_op_checks(node, 1, {"LeakyRelu"});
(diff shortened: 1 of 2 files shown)


itikhono (approved):

gkrivor (approved):

src/frontends/tensorflow_common/src/op/leaky_relu.cpp
```diff
@@ -13,12 +13,13 @@ namespace frontend {
 namespace tensorflow {
 namespace op {
 ov::OutputVector translate_leaky_relu_op(const NodeContext& node) {
-    auto in = node.get_input(0);
-    auto alpha_attr = node.get_attribute<float>("alpha", 0.f);
-    auto alpha_const = make_shared<Constant>(element::f32, Shape{1}, alpha_attr);
-
-    auto leaky_relu = make_shared<PRelu>(in, alpha_const);
-    return leaky_relu->outputs();
+    default_op_checks(node, 1, {"LeakyRelu", "LEAKY_RELU"});
```
popovaan, line 16: Why we need check both "LeakyRelu" and "LEAKY_RELU"? Should we add similar checks for all TF ops? It looks like other ops have only one name in check.

popovaan (approved):

src/frontends/tensorflow_common/src/op/leaky_relu.cpp
rkazants: This translator is used also by TensorFlow Lite frontend. In their models operation type name is `LEAKY_RELU`. we are checking for which operation names this translator is called.

rkazants: translators for some other operations are not used by TFL FE so there are just one name.

PR #15490, opened 2023-02-03
[PT FE] Add aten::empty operator with layer test

sikorsl1:
### Details:
 - `aten::empty` operator has been added

### Tickets:
 - 101075

### Comments:

- Generally in PyTorch `aten::empty` operator returns a tensor with uninitialized data, but we don't support such case in OV, so I decided to return a tensor filled with zeros. My solution is based on: https://github.com/pytorch/pytorch/blob/b41e2779f2e24364515d3980e660a6bb23ef5926/torch/onnx/symbolic_opset9.py#L3554

### DIFF
--- a/tests/layer_tests/pytorch_tests/test_empty.py
+++ b/tests/layer_tests/pytorch_tests/test_empty.py
@@ -1,0 +1,77 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import numpy as np
+import pytest
+import torch
+
+from pytorch_layer_test_class import PytorchLayerTest
+
+
+class TestEmptyNumeric(PytorchLayerTest):
+
+    def _prepare_input(self):
+        return (np.random.randn(2, 3),)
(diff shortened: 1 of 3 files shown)


mmikolajcz (approved):

tests/layer_tests/pytorch_tests/test_empty.py
```diff
@@ -0,0 +1,77 @@
...
+                # so we do "and" operation with False.
+                return empty & self.false
+
+        ref_net = None
+
+        return aten_empty(), ref_net, "aten::empty"
+
+    @pytest.mark.nightly
+    @pytest.mark.precommit
+    def test_empty(self, ie_device, precision, ir_version, ):
```
mvafin, line 76:
```suggestion
    def test_empty_bool(self, ie_device, precision, ir_version, ):
```

mvafin (changes requested):

mvafin (approved):

tomdol (approved):

PR #15493, opened 2023-02-03
Add aten narrow

bszmelcz:
### Details:
- [x] add support for `aten::narrow`
- [x] add test coverage
- [x] update `op_table.cpp`
### Tickets:
 - 101085


tests/layer_tests/pytorch_tests/test_narrow.py
```diff
@@ -0,0 +1,50 @@
...
+    @pytest.mark.parametrize("input_tensor", [
+        np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]]),
+        np.array([[[1, 2, 3], [4, 5, 6], [7, 8, 9]], [[5, 2, 4], [4, 9, 0], [7, 7, 9]], [[5, 2, 4], [4, 9, 0], [7, 7, 9]]])
+    ])
+    @pytest.mark.parametrize("dim", [
+        0, 1, -1
+    ])
+    @pytest.mark.parametrize("start", [
+        0, 1
+    ])
```
sikorsl1, line 40: PyTorch documentation allows `start` to be `int` or `Tensor`: https://pytorch.org/docs/stable/generated/torch.narrow.html#torch.narrow:~:text=start%20(Tensor%20or%20int)%20%E2%80%93%20the%20starting%20dimension. Why don't you test it for `Tensor`?

tests/layer_tests/pytorch_tests/test_narrow.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def create_model(self, dim, start, length):
+        import torch
+
+        class aten_narrow(torch.nn.Module):
+            def __init__(self, dim, start, length):
+                super(aten_narrow, self).__init__()
+                self.dim = dim
+                self.start = start
+                self.length = length
```
sikorsl1, line 22: I think that you should consider test case, where these values are not constant.

src/frontends/pytorch/src/op/narrow.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <climits>
```
mvafin, line 5: Do you need this include?

src/frontends/pytorch/src/op/narrow.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
+#include "utils.hpp"
```
mvafin, line 9: Don't think you need this include

src/frontends/pytorch/src/op/narrow.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <climits>
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
```
mvafin, line 8: It would be better to use individual op includes for such small files, but opset is also fine

mvafin (changes requested):

src/frontends/pytorch/src/op/narrow.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "utils.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_narrow(NodeContext& context) {
+    auto input_tensor = context.get_input(0);
```
mvafin, line 17: Please check number of inputs using `num_inputs_check` helper function

sikorsl1:
This PR should be closed without merging because I can't push to @bszmelcz fork. Continuation of this task can be found in the following PR: https://github.com/openvinotoolkit/openvino/pull/15788

PR #15496, opened 2023-02-03
[LPT] AddTransformation: checking constants for infinity

v-Golubev:
### Details:
 - *NetworkHelper: added a function which checks constants for infinity*
 - *AddTransformation: avoid transformation in case if result constants contain infinite values.*

### Tickets:
 - *102722*


eshoguli (approved):
LGTM

PR #15503, opened 2023-02-03
[TF FE] Support TensorList operations and RNN layers

rkazants:
**Details:** Support TensorList operations and RNN layers. The idea is to have tensor imitating the tensor list.
For cases when the rank of tensor elements in the list is unknown, we flatten all tensor elements to have the list of shape [num_elements, -1].
This PR unblock support any RNN layers of different structure. Also makes some models enabled via TF FE.

**Tickets:** 92643, 92642, 92641, 92640, 98867, 92843, 92844, 92845, 92846


gkrivor (approved):

mvafin (commented):
Looks good, but what about tests, it seems we do not have tests for all these ops

rkazants:
> Looks good, but what about tests? it seems we do not have tests for all these ops

In all of the layer tests for LSTM, GRU, Simple RNN we have these operations TensorList*. They are met in the combination in models so no way to test them separately.

mvafin:
> > Looks good, but what about tests? it seems we do not have tests for all these ops
> 
> In all of the layer tests for LSTM, GRU, Simple RNN we have these operations TensorList*. They are met in the combination in models so no way to test them separately. Now I added these tests in the pre-commit.

Ah, I see. Okay

mvafin (approved):

PR #15512, opened 2023-02-05
[TF FE] Add layer tests for If operation

rkazants:
**Details:** Add layer tests for If operation

**Tickets:** 92783, 92832

### DIFF
--- a/tests/layer_tests/tensorflow_tests/test_tf_If.py
+++ b/tests/layer_tests/tensorflow_tests/test_tf_If.py
@@ -1,0 +1,128 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import numpy as np
+import pytest
+import tensorflow as tf
+from common.tf_layer_test_class import CommonTFLayerTest
+
+
+class TestIfFloat(CommonTFLayerTest):
+    def _prepare_input(self, inputs_info):
+        assert 'cond' in inputs_info, "Test error: inputs_info must contain `cond`"
+        assert 'x' in inputs_info, "Test error: inputs_info must contain `x`"
+        assert 'y' in inputs_info, "Test error: inputs_info must contain `y`"
+        cond_shape = inputs_info['cond']
+        x_shape = inputs_info['x']
+        y_shape = inputs_info['y']
+        inputs_data = {}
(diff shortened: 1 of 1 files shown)


tests/layer_tests/tensorflow_tests/test_tf_If.py
```diff
@@ -0,0 +1,128 @@
...
+    def _prepare_input(self, inputs_info):
+        assert 'cond' in inputs_info, "Test error: inputs_info must contain `cond`"
+        assert 'ind' in inputs_info, "Test error: inputs_info must contain `ind`"
+        assert 'y' in inputs_info, "Test error: inputs_info must contain `y`"
+        cond_shape = inputs_info['cond']
+        ind_shape = inputs_info['ind']
+        y_shape = inputs_info['y']
+        inputs_data = {}
+        inputs_data['cond'] = np.random.randint(0, 2, cond_shape).astype(bool)
+        inputs_data['ind'] = np.random.randint(1, 10, ind_shape).astype(np.int32)
```
itikhono, line 80: is the difference between these 2 tests only in the input types?
why can't we parameterize them?

rkazants: not only types, but also we have difference in bodies.

rkazants: also, in the second case there is a mix of types in inputs and outputs

itikhono: the test names are confusing in this case "TestIfInt" and "TestIfFloat". Please describe the difference in the comment or select other names.

gkrivor: Overall - agree. Also I don't understand where is difference between "x" and "ind". I also see some difference between expressions, but not sure I understand are they meaningful or not.
I suggest to carefully review this part, but don't think it is merge blocker

gkrivor (approved):
Non-blocking comment from my side

tests/layer_tests/tensorflow_tests/test_tf_If.py
rkazants: @itikhono, well, let me fix this in the separate PR. I think this is too minor.
Will they be fine to call as `TestIfWithFloatInputs` and `TestIfWithMixedTypes`?

itikhono (approved):

PR #15514, opened 2023-02-05
[POC] Add signature verification for plugins

vurusovs:
### Details:
- Add signature verification (for Win) before plugin loading
- Self-signed library can be loaded successfully with current implementation. To reproduce lib signing, use https://stackoverflow.com/questions/84847/how-do-i-create-a-self-signed-certificate-for-code-signing-on-windows

### Tickets:
 - 96205

### DIFF
--- a/src/common/util/include/openvino/util/shared_object.hpp
+++ b/src/common/util/include/openvino/util/shared_object.hpp
@@ -21,6 +21,7 @@ namespace util {
  * @param path Full or relative path to the plugin library
  * @return Reference to shared object
  */
+std::shared_ptr<void> load_shared_object(const char* path, const bool& verify_signature);
 std::shared_ptr<void> load_shared_object(const char* path);
 
 #ifdef OPENVINO_ENABLE_UNICODE_PATH_SUPPORT
@@ -29,6 +30,7 @@ std::shared_ptr<void> load_shared_object(const char* path);
  * @param path Full or relative path to the plugin library
  * @return Reference to shared object
  */
+std::shared_ptr<void> load_shared_object(const wchar_t* path, const bool& verify_signature);
 std::shared_ptr<void> load_shared_object(const wchar_t* path);
 #endif  // OPENVINO_ENABLE_UNICODE_PATH_SUPPORT
 /**
--- a/src/common/util/src/os/win/win_shared_object_loader.cpp
+++ b/src/common/util/src/os/win/win_shared_object_loader.cpp
@@ -70,9 +70,74 @@
 #endif
 
 #include <windows.h>
+#include <wintrust.h>
+#include <Softpub.h>
+
+// #include <wincrypt.h>
+// #include <wintrust.h>
+
+// #pragma comment (lib, "wintrust")
+
+namespace {
+bool verify_embedded_signature(LPCWSTR pwszSourceFile)
+{
+	GUID WintrustVerifyGuid = WINTRUST_ACTION_GENERIC_VERIFY_V2;
+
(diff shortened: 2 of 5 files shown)


src/common/util/src/os/win/win_shared_object_loader.cpp
```diff
@@ -70,9 +70,74 @@
...
+
+// #include <wincrypt.h>
+// #include <wintrust.h>
+
+// #pragma comment (lib, "wintrust")
+
+namespace {
+bool verify_embedded_signature(LPCWSTR pwszSourceFile)
+{
+	GUID WintrustVerifyGuid = WINTRUST_ACTION_GENERIC_VERIFY_V2;
```
vurusovs, line 84: Add links:
- https://github.com/konstantin89/windows-pe-signature-verifying/blob/master/pe-signature-utils/src/native_api_wrappers/TrustVerifyWrapper.cpp#L21
- https://learn.microsoft.com/en-us/windows/win32/seccrypto/example-c-program--verifying-the-signature-of-a-pe-file

src/common/util/include/openvino/util/shared_object.hpp
```diff
@@ -21,6 +21,7 @@ namespace util {
  * @param path Full or relative path to the plugin library
  * @return Reference to shared object
  */
+std::shared_ptr<void> load_shared_object(const char* path, const bool& verify_signature);
```
rkazants, line 24: I think we need to update description with new param:)

vurusovs: @rkazants Thanks! For the moment it's just a POC to verify CI lib signs are compatible with Runtime signs verification

src/common/util/src/os/win/win_shared_object_loader.cpp
```diff
@@ -70,9 +70,74 @@
...
+bool verify_embedded_signature(const char* path) {
+    return verify_embedded_signature(ov::util::string_to_wstring(path).c_str());
+}
+}  // namespace
 
 namespace ov {
 namespace util {
+std::shared_ptr<void> load_shared_object(const char* path, const bool& verify_signature) {
+    if (verify_signature) {
+        if (!is_absolute_file_path(path)) {
```
vurusovs, line 126: Add new field `verify_signature` to `PluginDescriptor` in order to check signs only for required plugins

vurusovs:
Self-signed library can be loaded successfully with current implementation. To reproduce lib signing, use https://stackoverflow.com/questions/84847/how-do-i-create-a-self-signed-certificate-for-code-signing-on-windows

src/common/util/src/os/win/win_shared_object_loader.cpp
```diff
@@ -124,6 +189,23 @@ std::shared_ptr<void> load_shared_object(const char* path) {
 }
 
 #ifdef OPENVINO_ENABLE_UNICODE_PATH_SUPPORT
+std::shared_ptr<void> load_shared_object(const wchar_t* path, const bool& verify_signature) {
+    if (verify_signature) {
+        if (!is_absolute_file_path(ov::util::wstring_to_string(path))) {
+            // TODO: check how it works with file names
+            std::stringstream ss;
+            ss << "Cannot verify signature of library '" << path << "': path isn't absolute.";
```
vurusovs, line 197: add `wstring_to_string` for correct message

vurusovs:
For the moment we decided to decline signature verification because initial problem which we are trying to solve is covered by OS through the differentiation of user access rights (more details in a ticket)

PR #15525, opened 2023-02-06
[GPU] Show num_ccs for RANGE_FOR_STREAMS

isanghao:
### Details:
 - Expose num_ccs through RANGE_FOR_STREAMS


vladimir-paramuzov (approved):

ngaloppo:
Wait, if `ov::range_for_streams` returns `2`, then we don't actually know if the hardware has one or two CCS. How is that useful?

isanghao:
> Wait, if `ov::range_for_streams` returns `2`, then we don't actually know if the hardware has one or two CCS. How is that useful?

right. But it is still useful when hardware has 4 ccs. A770 has 4 ccs.

ngaloppo:
Got it. Why would someone want to use 2 streams if only 1 CCS is available?

isanghao:
> Got it. Why would someone want to use 2 streams if only 1 CCS is available?

Nice question :) Actual performance is better to use 2 streams for 1-ccs hardware. I guess this is because of kernel synchronization / network enqueue overhead. That is why we are using 2streams for 1-ccs hardware.

Then you may ask "8 streams would be better for 4-ccs hardware"? It did not bring meaning perf gain for such case. I guess this is because there is no idling time, already with 4 streams.

PR #15542, opened 2023-02-07
[GPU] Optimized permute kernel cannot be fused with both eltwise + reorder.

yeonbok:
### Details:
 - Optimized permute kernel cannot be fused with both eltwise + reorder. So prevented selection of optimized kernel for such case. However it may be better not to fuse reorder instead in such case. Need further investigation with performance impact. 

### Tickets:
 - 93264


andrew-k-park (approved):

e-ddykim (approved):
Looks good to me

PR #15548, opened 2023-02-07
[TF FE] Refactor Range and add layer test

rkazants:
**Details:** Refactor Range and add layer test

**Ticket:** 92816

### DIFF
--- a/tests/layer_tests/tensorflow_tests/test_tf_Range.py
+++ b/tests/layer_tests/tensorflow_tests/test_tf_Range.py
@@ -1,0 +1,58 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import numpy as np
+import pytest
+import tensorflow as tf
+from common.tf_layer_test_class import CommonTFLayerTest
+
+
+class TestRange(CommonTFLayerTest):
+    def _prepare_input(self, inputs_info):
+        inputs_data = {}
+        if self.negative_delta:
+            inputs_data['start'] = np.random.randint(1, 10, []).astype(self.input_type)
+            inputs_data['limit'] = np.random.randint(-10, 0, []).astype(self.input_type)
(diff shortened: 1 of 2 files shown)


tests/layer_tests/tensorflow_tests/test_tf_Range.py
```diff
@@ -0,0 +1,58 @@
...
+        }
+        assert input_type in types_map, "Incorrect test case"
+        tf_type = types_map[input_type]
+        tf.compat.v1.reset_default_graph()
+        # Create the graph and model
+        with tf.compat.v1.Session() as sess:
+            start = tf.compat.v1.placeholder(tf_type, [], 'start')
+            limit = tf.compat.v1.placeholder(tf_type, [], 'limit')
+            delta = tf.compat.v1.placeholder(tf_type, [], 'delta')
+            tf.raw_ops.Range(start=start, limit=limit, delta=delta)
```
itikhono, line 39: please try to set `dtype` attribute with different values to Range op here.
It looks like, it can be related with this deleted line
`auto out_type = node.get_attribute<ov::element::Type>("Tidx");`

rkazants: `tf.raw_ops.Range` does not have `dtype` attribute: https://www.tensorflow.org/api_docs/python/tf/raw_ops/Range
And according to the specification `Tidx` can be absent at all so we can't rely on it. It explicitly says that output type is the same as `start`

itikhono: but the python version has https://www.tensorflow.org/api_docs/python/tf/range
image
please check how it works

rkazants: I just checked. Everytime it finally converted to `raw_ops.Reshape` version that follows to the specification.
It is the same for all operations. So we need to look at `raw_ops`:)

itikhono (approved):

popovaan (approved):

PR #15559, opened 2023-02-07
[CONFORMANCE] Add support of models and lists of models to the conformance rrunner

iefode:
### Details:
 - *Add support of download models by list file*
 - *Add support of working with original models via frontends*

### Tickets:
 - *103277*

### DIFF
--- a/src/tests/functional/plugin/conformance/subgraphs_dumper/src/main.cpp
+++ b/src/tests/functional/plugin/conformance/subgraphs_dumper/src/main.cpp
@@ -33,6 +33,12 @@ static std::vector<std::regex> getRegexByFrontend() {
 #ifdef ENABLE_OV_IR_FRONTEND
     result.push_back(std::regex(R"(.*\.xml)"));
 #endif
+#ifdef ENABLE_OV_TF_LITE_FRONTEND
+    result.push_back(std::regex(R"(.*\.tflite)"));
+#endif
+#ifdef ENABLE_OV_PYTORCH_FRONTEND
+    result.push_back(std::regex(R"(.*\.pt)"));
+#endif
     return result;
 }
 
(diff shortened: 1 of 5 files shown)


src/tests/functional/plugin/conformance/subgraphs_dumper/src/main.cpp
```diff
@@ -32,6 +32,12 @@ static std::vector<std::regex> getRegexByFrontend() {
 #endif
```
rkazants, line 32: btw, we look to have missed other TF formats. Please double-check:)

rkazants, line 32: btw, we look to have missed other TF formats. Please double-check:)

iefode: @rkazants Please provide all formats. I take extensions from the tests

rkazants: 1. **Checkpoint**. In this case, a model consists of two files: `inference_graph.pb` (or `inference_graph.pbtxt`) and `checkpoint_file.ckpt`.
```sh
 mo --input_model <INFERENCE_GRAPH>.pb --input_checkpoint <INPUT_CHECKPOINT>
```

2. **MetaGraph**. In this case, a model consists of three or four files stored in the same directory: `model_name.meta`, `model_name.index`,

```sh
 mo --input_meta_graph <INPUT_META_GRAPH>.meta
```

3. **SavedModel format**. In this case, a model consists of a special directory with a `.pb` file
and several subfolders: `variables`, `assets`, and `assets.extra`. For more information about the SavedModel directory, refer to the [README](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/python/saved_model#components) file in the TensorFlow repository.
```sh
 mo --saved_model_dir <SAVED_MODEL_DIRECTORY>
```

iefode: Discussed offline with @rkazants. These extensions are not valid for c++ tool.
Should be implemented on python side of conformance tools in the future.

sbalandi (approved):

PR #15561, opened 2023-02-07
[Remove API] remove api add_extension & related func

zhaixuejun1993:
Signed-off-by: xuejun <xuejun.zhai@intel.com>

### Details:
 - ``` void add_extension(const std::shared_ptr<InferenceEngine::IExtension>& extension); ```
 - ``` virtual void add_extension(const std::shared_ptr<InferenceEngine::IExtension>& extension); ```

### Tickets:
 - 100091

### DIFF
--- a/src/plugins/intel_cpu/src/plugin.cpp
+++ b/src/plugins/intel_cpu/src/plugin.cpp
@@ -637,10 +637,6 @@ Parameter Engine::GetMetric(const std::string& name, const std::map<std::string,
     return GetMetricLegacy(name, options);
 }
 
-void Engine::AddExtension(const InferenceEngine::IExtensionPtr& extension) {
-    extensionManager->AddExtension(extension);
-}
-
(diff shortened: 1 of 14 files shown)


src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -637,10 +637,6 @@ Parameter Engine::GetMetric(const std::string& name, const std::map<std::string,
     return GetMetricLegacy(name, options);
 }
 
-void Engine::AddExtension(const InferenceEngine::IExtensionPtr& extension) {
-    extensionManager->AddExtension(extension);
-}
```
riverlijunjie, line 642: Supposed add_extension API is deprecated and removed, do we also need remove ExtensionManager class? @ilyachur

ilyachur:
Hi, please don't remove this API.
I propose to remove this API together with legacy IE API in 2024.0 release

riverlijunjie:
@zhaixuejun1993 per Ilya's comments, deprecated APIs related legacy API and with 2024.0 deprecated flag APIs will not removed, so we can close this PR.

PR #15563, opened 2023-02-07
[PT FE] Add skip for prim::device

mmikolajcz:
### Details:
 - *Skip prim::device operation*
Operator `prim::device` in PyTorch is used to indicate on with device Tensors will be allocated. Skipping it, since in OV we don't have operations that support such actions

### Tickets:
 - *99232*

### DIFF
--- a/src/frontends/pytorch/src/op_table.cpp
+++ b/src/frontends/pytorch/src/op_table.cpp
@@ -301,6 +301,8 @@ const std::map<std::string, CreatorFunction> get_supported_ops() {
         {"aten::zeros", op::translate_zeros},
         {"aten::zeros_like", op::translate_zeros_like},
         {"prim::Constant", op::translate_constant},
+        {"prim::device", op::skip_node},  // In openvino how tensors are stored in memory is internal plugin detail,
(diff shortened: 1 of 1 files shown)


sikorsl1 (approved):
Impressive piece of code!

mvafin:
I don't think we can just skip `prim::device`

eaidova:
@mmikolajcz  are you looked how it is used in models?

I see some conditional usage for if like that https://github.com/pytorch/vision/blob/main/torchvision/ops/boxes.py#L72
where device detection used for selecting implementation details, so just skipping is not variant here

mmikolajcz:
Thanks @mvafin @eaidova. I mostly though for cases like "aten::to" where device is just ignored, didn't though about it used in control flow like that with prim::type.
How output from deice is used is quite depended at consumer op, maybe best way would be to not skip device or do anything on this op level like @mvafin suggested and rely on consumers of this op to remove it?
I guess hardcoding it to always return "cpu" probably wouldn't be best solution, and I don't think we can retrieve in OV on what device tensors were located.

PR #15572, opened 2023-02-08
[TF FE] Support SaveV2 operation

rkazants:
**Details:** Support SaveV2 operation. Also, implement transformation to remove UnsupportedConstant to Result isolated sub-graphs. It allows to convert one more model OOB.

**Ticket:** 102648

### DIFF
--- a/src/frontends/tensorflow/tests/convert_tricky_models.cpp
+++ b/src/frontends/tensorflow/tests/convert_tricky_models.cpp
@@ -246,3 +246,19 @@ TEST_F(TransformationTestsF, DISABLED_ModelWithDilatedGroupConvolution) {
         model_ref = make_shared<Model>(OutputVector{transpose_after}, ParameterVector{x});
     }
 }
+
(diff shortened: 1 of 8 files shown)


gkrivor (approved):

src/frontends/tensorflow/tests/convert_tricky_models.cpp
```diff
@@ -246,3 +246,19 @@ TEST_F(TransformationTestsF, DISABLED_ModelWithDilatedGroupConvolution) {
         model_ref = make_shared<Model>(OutputVector{transpose_after}, ParameterVector{x});
     }
 }
+
+TEST_F(TransformationTestsF, ModelWithSaveV2) {
```
popovaan, line 250: Does this test cover UnsupportedConstToResultRemover transform?

rkazants: Yes, it includes both SaveV2 and StringConst->Identity case.

popovaan (approved):

PR #14156, opened 2022-11-22
Add ITensor

ilyachur:
### Details:
 - Port https://github.com/openvinotoolkit/openvino/pull/11057

### Tickets:
 - *ticket-id*

PR #15574, opened 2023-02-08
[WIP] Data accessor for shape infer method

praasz:
### Details:
 - *item1*
 - *...*

### Tickets:
 - 101672

### DIFF
--- a/src/core/shape_inference/include/tensor_accessor.hpp
+++ b/src/core/shape_inference/include/tensor_accessor.hpp
@@ -4,13 +4,85 @@
 
 #pragma once
 
-// #include "openvino/core/node.hpp"
+#include "ngraph/runtime/host_tensor.hpp"
 #include "openvino/runtime/tensor.hpp"
 
 namespace ov {
 
+struct ITensorDataAdapter;
+
+using ITensorDataAdapterPtr = std::unique_ptr<ITensorDataAdapter>;
 /** \brief Function type to get tensors from memory container. */
-using get_tensor_func_t = std::function<Tensor(size_t)>;
+using get_tensor_func_t = std::function<ITensorDataAdapterPtr(size_t)>;
+
+struct ITensorDataAdapter {
+    virtual ~ITensorDataAdapter() = default;
+
+    virtual element::Type_t get_element_type() const = 0;
+    virtual size_t get_size() const = 0;
+    virtual const void* data() const = 0;
+};
+
+class TensorDataAdapter : public ITensorDataAdapter {
+public:
+    TensorDataAdapter(const Tensor& tensor) : m_tensor{tensor} {}
+
+    element::Type_t get_element_type() const override {
+        return m_tensor.get_element_type();
+    }
+
+    size_t get_size() const override {
+        return m_tensor.get_size();
+    };
+
+    const void* data() const override {
+        return m_tensor.data();
+    }
+
+private:
+    const Tensor& m_tensor;
+};
+
+template <class TContainer>
+class ContainerDataAdapter : public ITensorDataAdapter {
+public:
+    ContainerDataAdapter(const TContainer& c) : m_c{c} {}
+
+    element::Type_t get_element_type() const override {
+        return element::from<typename TContainer::value_type>();
+    }
+
+    size_t get_size() const override {
+        return m_c.size();
+    };
+
+    const void* data() const override {
+        return m_c.data();
+    }
+
+private:
+    const TContainer& m_c;
+};
+
+class HostTensorDataAdapter : public ITensorDataAdapter {
+public:
+    HostTensorDataAdapter(const HostTensor& tensor) : m_tensor{tensor} {}
+
+    element::Type_t get_element_type() const override {
+        return m_tensor.get_element_type();
+    }
+
+    size_t get_size() const override {
+        return m_tensor.get_element_count();
+    }
+
+    const void* data() const override {
+        return m_tensor.get_data_ptr();
+    }
+
+private:
+    const HostTensor& m_tensor;
+};
 
 /**
  * \brief Functor implements get_tensor_func_t.
@@ -27,14 +99,18 @@ struct TensorAccessor {
     TensorAccessor(const TContainer& tensors) : m_tensors{tensors} {}
 
     template <class C = TContainer, typename std::enable_if<std::is_same<C, TensorVector>::value>::type* = nullptr>
-    Tensor operator()(size_t idx) {
-        return idx < m_tensors.size() ? m_tensors[idx] : Tensor();
+    auto operator()(size_t idx) -> ITensorDataAdapterPtr {
+        if (idx < m_tensors.size()) {
+            return std::unique_ptr<TensorDataAdapter>(new TensorDataAdapter{m_tensors[idx]});
+        } else {
+            return {};
+        }
     }
 
     template <class C = TContainer, typename std::enable_if<std::is_same<C, HostTensorVector>::value>::type* = nullptr>
-    Tensor operator()(size_t idx) {
+    auto operator()(size_t idx) -> ITensorDataAdapterPtr {
         if (idx < m_tensors.size()) {
-            return {m_tensors[idx]->get_element_type(), m_tensors[idx]->get_shape(), m_tensors[idx]->get_data_ptr()};
+            return std::unique_ptr<HostTensorDataAdapter>(new HostTensorDataAdapter{*m_tensors[idx]});
         } else {
             return {};
         }
@@ -42,10 +118,10 @@ struct TensorAccessor {
 
     template <class C = TContainer,
               typename std::enable_if<std::is_same<C, std::map<size_t, HostTensorPtr>>::value>::type* = nullptr>
-    Tensor operator()(size_t idx) {
+    auto operator()(size_t idx) -> ITensorDataAdapterPtr {
         const auto t_iter = m_tensors.find(idx);
         if (t_iter != m_tensors.cend()) {
-            return {t_iter->second->get_element_type(), t_iter->second->get_shape(), t_iter->second->get_data_ptr()};
+            return std::unique_ptr<HostTensorDataAdapter>(new HostTensorDataAdapter{*t_iter->second});
         } else {
             return {};
         }
@@ -56,12 +132,12 @@ private:
 };
 
 /**
- * \brief Get default tensor for any index number.
+ * \brief Get null data adapter pointer for any index number.
  *
  * \param idx  Index to get tensor.
- * \return Default (null) ov::Tensor.
+ * \return nullptr.
  */
-inline auto null_tensor_accessor(size_t idx) -> ov::Tensor {
+inline auto null_tensor_accessor(size_t idx) -> ITensorDataAdapterPtr {
     return {};
 };
 
--- a/src/core/shape_inference/include/tensor_accessor.hpp
+++ b/src/core/shape_inference/include/tensor_accessor.hpp
@@ -1,0 +1,80 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+// #include "openvino/core/node.hpp"
+#include "openvino/runtime/tensor.hpp"
+
+namespace ov {
+
+/** \brief Function type to get tensors from memory container. */
+using get_tensor_func_t = std::function<Tensor(size_t)>;
+
+/**
+ * \brief Functor implements get_tensor_func_t.
+ *
+ *  Is not owning TContainer and supports following tensor containers:
+ *  - ov::TensorVector
+ *  - ov::HostTensorVector
+ *  - std::map<size_t, ov::HostTensorPtr>
+ *
+ * \tparam TContainer
+ */
+template <class TContainer>
+struct TensorAccessor {
+    TensorAccessor(const TContainer& tensors) : m_tensors{tensors} {}
+
+    template <class C = TContainer, typename std::enable_if<std::is_same<C, TensorVector>::value>::type* = nullptr>
+    Tensor operator()(size_t idx) {
+        return idx < m_tensors.size() ? m_tensors[idx] : Tensor();
+    }
+
+    template <class C = TContainer, typename std::enable_if<std::is_same<C, HostTensorVector>::value>::type* = nullptr>
+    Tensor operator()(size_t idx) {
+        if (idx < m_tensors.size()) {
+            return {m_tensors[idx]->get_element_type(), m_tensors[idx]->get_shape(), m_tensors[idx]->get_data_ptr()};
+        } else {
+            return {};
+        }
+    }
+
+    template <class C = TContainer,
+              typename std::enable_if<std::is_same<C, std::map<size_t, HostTensorPtr>>::value>::type* = nullptr>
+    Tensor operator()(size_t idx) {
+        const auto t_iter = m_tensors.find(idx);
+        if (t_iter != m_tensors.cend()) {
+            return {t_iter->second->get_element_type(), t_iter->second->get_shape(), t_iter->second->get_data_ptr()};
+        } else {
+            return {};
+        }
+    }
+
+private:
+    const TContainer& m_tensors;
+};
+
+/**
+ * \brief Get default tensor for any index number.
+ *
+ * \param idx  Index to get tensor.
+ * \return Default (null) ov::Tensor.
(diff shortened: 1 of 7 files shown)


jane-intel:
@maxnick and @Lyamin-Roman please have a look into new way of data passing between plugin and shape inference

Lyamin-Roman (approved):
Looks good to me as a temporary solution for GPU side

praasz:
> Looks good to me as a temporary solution for GPU side

Yes it is example.
Can you suggest where class `cldnn::MemoryAdapter` class could stored as separate file in GPU plugin to use in other ops in future? 

It looks like in example lambda in tile operator it can be generalized for accessing data from `kernel_impl_params` as it stores as a map, but there are also some operator specific data like  [repeats](https://github.com/openvinotoolkit/openvino/blob/b98e9504e4a919977871eceb962ad4037b32f05a/src/plugins/intel_gpu/src/graph/tile.cpp#L83). 

Do you think is possible to make this kind of lambda more generic (as class) which could be re-used in other operators to access `kernel_impl_params` data and op specific data?

yeonbok:
> Yes it is example. Can you suggest where class `cldnn::MemoryAdapter` class could stored as separate file in GPU plugin to use in other ops in future?

I'd like to suggest under intel_gpu/runtime/include/memory.hpp.

src/core/dev_api/tensor_data_accessor.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * This accessor not take ownership of tensors container.
+ * Supports following containers:
+ * - ov::TensorVector
+ * - ov::HostTensorVector
+ * - std::map<size_t, ov::HostTensorPtr>
+ *
+ * @tparam TContainer Type of tensor container.
+ */
+template <class TContainer>
+struct OPENVINO_API TensorAccessor {
```
maxnick, line 28: Just wondering, why we cannot simply introduce such accessor interface and every plugin just implement it on its side? So we just pass the accessor to the infer call.

maxnick:
@praasz , the main intention of this PR is great. Plugins will remove responsibility on maintaining knowledge on which inputs must be passed to the shape inference. However, I think one important feature is missing. From the plugin standpoint (at least CPU), it is important to know which data inputs are required for shape inference in advance (before the shape infer). In other words, we need to know if the operation shape inference procedure has any data dependency. This is important, because not all the operations require additional data to calculate the output shapes. In many cases only input shapes are enough (plus m.b. some internal parameters) that means such chains of operations can run shape inference before the actual data propagation, which may be exploited for optimizations. But with your proposed solution we still have no way to request such info in abstract fashion.

src/core/dev_api/tensor_data_accessor.hpp
praasz: I assumed this interface is specified by `tensor_data_accessor_func_t` which define function object which can be passed.
So in plugin side it can be implemented by anything (not bind to specific class) like lambda, functor, function which satisfy this function object.
The class `ov::TensorAccessr` is to provide this function object mainly for tensors (ov::Tensor, HostTensor) where I specified specialization for `operator()` for different kind of tensors and containers which hold them. It looks like these specializations cover all cases from core perspective. They are in dev API because they are shared between libraries.

As I don't know what exactly is required by plugin I left it open for implementation. Plugin can have own class or maybe will be enough to add in plugin own `tensor adapter class` and then specialization for `operator()` and re-use `make_tensor_accessor` helper to create it as in tile operator [here](https://github.com/openvinotoolkit/openvino/blob/b98e9504e4a919977871eceb962ad4037b32f05a/src/core/src/op/tile.cpp#L59) and [here](https://github.com/openvinotoolkit/openvino/blob/b98e9504e4a919977871eceb962ad4037b32f05a/src/core/src/op/tile.cpp#L84)

maxnick: I got the idea, thank you. But, this functional approach being more flexible may have some performance drawbacks. The thing is, the shape infer operations should be as quick as possible. Also some of them are really simple, like merging two vectors of 4 elements each. Though, any ongoing overheads might be really big in comparison with the shape infer algorithm itself. For example, passing the functor as a value may involve additional copy of internal structures, even though this is not a big problem from the interface stand point, since it is always possible to say that the user should implement more lightweight functors, it is still better to avoid such copy. Another point is the return type, which is a unique_ptr. That means that most likely we should allocate dynamic memory to store the adapter every access request call. Which requires `malloc`, which is not cheap.
But, in opposite, we can take a reference to some accessor abstract interface, which always return a tensor of specific type, so all the internal optimizations are on the plugin side and we do not impose any additional overheads on the interface level. That means on the plugin side I can instantiate such adapter implementation object only once, as well as all the auxiliary structures and pass it as a reference every infer call. Also the accessor implementation may return a reference to some thin wrapper over the internal tensor implementation that provides direct access to the internal data with minimal overhead (at lest in the case of CPU plugin since the data are stored in the same DRAM).

praasz:
> @praasz , the main intention of this PR is great. Plugins will remove responsibility on maintaining knowledge on which inputs ...

@maxnick thanks for comment. I suspect there will be something missing from plugin perspective.
So this missing information about data dependency in operator do mean that e.g. the `IStaticShapeInfer` should provide something like `port map` like [port_map](https://github.com/openvinotoolkit/openvino/blob/8839c66ea6a4cc2d52c01298cd23e2db3cd05d41/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_cpu.hpp#L50)?

If yes I think I can easily add to `IStaticShapeInfer` class and it looks like this mask by default can be easily generated in generic way and also I can `ctor` which accept custom value for specific operators if required. 
Please share if this is missing data or something else or idea how you see what should be provided in CPU plugin.

@openvinotoolkit/openvino-ie-gpu-developers 
Can you share opinion if this port map will useful from GPU perspective or there is required something else?

src/core/dev_api/tensor_data_accessor.hpp
praasz: Regarding passing accessor by reference and avoid using unique_ptr I can re-factor it to not use.

The current tensor accessor returns `ITensorDataAdapter` which can store tensor pointer or memory pointer and the `data()` member function return pointer to data. There are additional member functions to get element type and size to jump over data correctly. Do you think about something thinner?

In case of GPU plugin, this example adapter holds also lock to data to be sure that during shape inference data are locked and released after. If accessor return reference not sure how to hold correctly this lock and provide correct lifetime.

For `ov:tensor` and `host tensor` it warp tensors to unify the interface between tensors.

Lyamin-Roman:
> It looks like in example lambda in tile operator it can be generalized for accessing data from `kernel_impl_params` as it stores as a map, but there are also some operator specific data like [repeats](https://github.com/openvinotoolkit/openvino/blob/b98e9504e4a919977871eceb962ad4037b32f05a/src/plugins/intel_gpu/src/graph/tile.cpp#L83).
> 
> Do you think is possible to make this kind of lambda more generic (as class) which could be re-used in other operators to access `kernel_impl_params` data and op specific data?

Unfortunately, I have no idea how at the moment it is possible to generalize this lambda completely as a class, only one part of this code can be generalized into a separate method common to all operations:
```
    const auto mem_it = impl_param.memory_deps.find(i);
    if (mem_it != impl_param.memory_deps.cend()) {
        return std::unique_ptr<MemoryAdapter>(new MemoryAdapter{mem_it->second, impl_param.prog->get_stream()});
    } else {
        return nullptr;
    }
```
The second part of the code is specific to a particular operation.

> Can you suggest where class cldnn::MemoryAdapter class could stored as separate file in GPU plugin to use in other ops in future?
>> I'd like to suggest under intel_gpu/runtime/include/memory.hpp.

On this point, I agree with @yeonbok

src/core/dev_api/tensor_data_accessor.hpp
maxnick: > The current tensor accessor returns ITensorDataAdapter which can store tensor pointer or memory pointer and the data() member function return pointer to data. There are additional member functions to get element type and size to jump over data correctly. Do you think about something thinner?

Honestly it completely depends on how it is used, if this interface is sufficient, then OK.

> In case of GPU plugin, this example adapter holds also lock to data to be sure that during shape inference data are locked and released after. If accessor return reference not sure how to hold correctly this lock and provide correct lifetime.

That is why I'm asking about passing the accessor instead of the factory method. In case we pass the accessor by reference, its lifetime is controlled by the plugin, so the shape infer implementation should not bother about the accessor lifespan, and, in its turn, it can return a reference to the tensor adapter, and again the lifespan is controlled by the plugin (since all the locks and auxiliary data structures are managed by the accessor). Or m.b. I'm missing something and we want to store the adapter reference beyond the `shape_infer` call and want to mange its lifetime from the shape inference subsystem?

maxnick:
> `@maxnick` thanks for comment. I suspect there will be something missing from plugin perspective. So this missing information about data dependency in operator do mean that e.g. the `IStaticShapeInfer` should provide something like `port map` like [port_map](https://github.com/openvinotoolkit/openvino/blob/8839c66ea6a4cc2d52c01298cd23e2db3cd05d41/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_cpu.hpp#L50)?

Yes, the port map was selected as the fast possible method to retrieve the data dependency info (just read one 32 bit value). I do not expect that we have operations with more than 32 input ports.

src/core/dev_api/tensor_data_accessor.hpp
praasz: There is version to pass accessor via reference

praasz:
> > It looks like in example lambda in tile operator it can be generalized for accessing data from `kernel_impl_params` as it stores as a map, but there are also some operator specific data like [repeats](https://github.com/openvinotoolkit/openvino/blob/b98e9504e4a919977871eceb962ad4037b32f05a/src/plugins/intel_gpu/src/graph/tile.cpp#L83).
> > Do you think is possible to make this kind of lambda more generic (as class) which could be re-used in other operators to access `kernel_impl_params` data and op specific data?
> 
> Unfortunately, I have no idea how at the moment it is possible to generalize this lambda completely as a class, only one part of this code can be generalized into a separate method common to all operations:
> 
> ```
>     const auto mem_it = impl_param.memory_deps.find(i);
>     if (mem_it != impl_param.memory_deps.cend()) {
>         return std::unique_ptr<MemoryAdapter>(new MemoryAdapter{mem_it->second, impl_param.prog->get_stream()});
>     } else {
>         return nullptr;
>     }
> ```
> 
> The second part of the code is specific to a particular operation.
> 
> > Can you suggest where class cldnn::MemoryAdapter class could stored as separate file in GPU plugin to use in other ops in future?
> > > I'd like to suggest under intel_gpu/runtime/include/memory.hpp.
> 
> On this point, I agree with @yeonbok

I've put in separate close to operators in GPU plugin. The reason for that is core dev API headers are not visible in suggested location. There probably should public core API to be visible there. There is also work on ITensor interface and some other classes which could probably replace this ITensorDataAdapter and then move of classes should be easy.

src/core/dev_api/tensor_data_accessor.hpp
```diff
@@ -0,0 +1,106 @@
...
+ * This accessor not take ownership of tensors container.
+ * Supports following containers:
+ * - ov::TensorVector
+ * - ov::HostTensorVector
+ * - std::map<size_t, ov::HostTensorPtr>
+ *
+ * @tparam TContainer Type of tensor container.
+ */
+template <class TContainer>
+class OPENVINO_API TensorAccessor : public ITensorAccessor {
```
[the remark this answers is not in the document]
praasz: The `ITensorAccessor` is to provide common way to access tensor-like data from containers (looks like map, and vector is commonly used). The shape_infer function then can access tensor data by requesting it by port index from plugin or core.
Main purpose for this accessor is bound check for data container and return `ITensorDataAdapter` to provide common way to access data as the `ITensor` is not done yet.

I was trying using the `ov::Tensor` instead but got some issues and also try follow @maxnick suggestion to do this access as simple as possible:

1. To avoid additional allocation. As plugin already has data so creating the ov::Tensor which holds at least one smart pointer and `ctor` are not very simple will cause additional allocations. The `DataAccessor` holds only the pointer to data adapter which should be as simple as possible (except GPU version it got only raw pointer). During accessing data the pointer value is replaced and returned. It should be cheap.
   - I think CPU plugin can do it similar as done for Tensors and creating new object will be not required.
   - I know that current version of data accessor for GPU use unique pointer (dynamic allocation) but I think can be optimized to use same pattern as for tensor to have only raw pointer copy but probably will required different approach to create memory lock. I think GPU team can optimize it for their needs.

2. In case of GPU plugin there is also required to create lock for accessed data. I don't know how e.g. create `ov::Tensor` and store the lock inside so this tensor adapter allow store lock and release when object will be deleted.

3. In case the ov::Tensor, I agree that adapter is overhead but the `ITensor` and `TensorView` classes from #14156 are not available yet. But I need common interface to access data. When `ITensor` will be don I think I can refactor and replace `ITensorDataAdapter` by `ITensor`. Then Accessor for `ov::Tensor` will be very simple as just bound check and return pointer or ref to data from tensor container.

4. In case of HostTensor it can be beneficial as access to data and conversion to adapter is raw pointer replace. Currently in CPU plugin the memories are converted to HostTensorPtr so it allow avoid again conversion to ov::Tensor just use pointers. If `HostTensor` implement `ITensor` then simplification will be same as for `ov::Tensor`.
If will not implement it then the `TensorView` could be used as adapter.

5. If plugin memories will implement `ITensor` then accessing data will be very simple if not maybe `TensorView` can be used to provide common interface.

I've put this files in dev API to share it between core, shape_inference, plugins and if possible share code from on lib. I can move it to shape_infer but then without exportin each library will have its own code.

I'm open to suggestions if there is possibility to have Tensor API implementation and avoid allocation etc. please share the idea.

maxnick (approved):
The proposed design looks good.

src/core/dev_api/tensor_data_accessor.hpp
```diff
@@ -0,0 +1,106 @@
...
+//
+
+#pragma once
+
+#include "openvino/core/node.hpp"
+#include "openvino/runtime/tensor.hpp"
+
+namespace ov {
+/** @brief Interface for data accessor. */
+class OPENVINO_API ITensorAccessor {
```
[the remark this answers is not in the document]
praasz: It not replace the ov::tensor it provide interface how get e.g like plugin own implementation.
But if the `MemoryAccesser` can be removed because of the use the `RemoteTensor` then can remove it.

src/core/shape_inference/include/reverse_shape_inference.hpp
```diff
@@ -39,7 +39,7 @@ namespace v1 {
 template <class TShape>
 std::vector<TShape> shape_infer(const Reverse* op,
                                 const std::vector<TShape>& input_shapes,
-                                const std::map<size_t, std::reference_wrapper<const ov::Tensor>>& constant_data = {}) {
+                                const ITensorAccessor& tensor_accessor = null_tensor_accessor()) {
```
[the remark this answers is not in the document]
praasz: The intention was not used the `std::function` to create this wrapper every call, instead the object of accessor is create on caller side and pass reference only ( the function pointer will not work to pass TensorAccessor).
The ITensorAccessor is just interface with requirement to have call operator to be able to call specific implementation (core, plugin etc).
I think the ITensorAccessor can be integrated into TnsorAccessor.

src/plugins/intel_gpu/src/graph/include/memory_accessor.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include "intel_gpu/runtime/memory.hpp"
+#include "tensor_data_accessor.hpp"
+
+namespace cldnn {
+
+/**
+ * @brief CLDNN memory accessor implementing ov::ITensorAccessor to get data as tensor from CLDNN container.
+ */
+struct MemoryAccessor : public ov::ITensorAccessor {
```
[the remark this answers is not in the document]
praasz: It handles the cldnn memory map similar as  core TensorAccessor and also has additional feature to provide data as tensor from GPU operators properties e.g see Tile shape inference which use repeats from tile  `primitive`.

ilyachur (approved):
Common part LGTM

ilyachur (approved):
Common part LGTM

Lyamin-Roman (approved):

jane-intel (approved):

PR #15575, opened 2023-02-08
[Core] fix compress_float_constants.cpp for denormal float16

pavel-esir:
### Details:
 - Because of a typo fp16 denormal values in IR bin file were not compressed and were filled by blank zeros. This led to accuracy degradation. Fixed that

### Tickets:
 - xxx-103313, xxx-103382, xxx-103379, xxx-103911


src/common/transformations/src/transformations/common_optimizations/compress_float_constants.cpp
```diff
@@ -29,6 +29,7 @@ std::shared_ptr<ov::Node> change_constant_precision_to_fp16(std::shared_ptr<ov::
     for (size_t i = 0; i < size; ++i) {
         // if it is smaller than the smallest positive normal fp16 but not zero
         if (std::abs(src_data[i]) <= ov::float16::from_bits(0x0400) && src_data[i] != 0.0f) {
+            dst_data[i] = static_cast<ov::float16>(src_data[i]);
```
mvafin, line 32: Will it be casted to zero anyway here? Shouldn't we have smallest fp16 value here?

pavel-esir: I was also surprised that it indeed will not be casted to zero. I checked that.

BTW, i forgot to add unittest, will do it :)

mvafin: It might mean that `0x0400` is not really smallest fp16 number. As far as I can see here https://en.wikipedia.org/wiki/Half-precision_floating-point_format smallest is actually `0x0001`

pavel-esir: I knew that `0x0004` was not the smallest fp16, it was the smallest normal value (looks like a reflected that in the comment line above). For some reason i forgot to change that :(

Now I indeed use the smallest fp16, which is `0x0001`, as you stated. Added unit-test for this fix as well. Please take a look

mvafin (approved):

ilya-lavrenov:
is it valid to port this fix to 2022.3?

pavel-esir:
> is it valid to port this fix to 2022.3?

On 2022.3 we don't have this degradation, bug was introduced recently, therefore it's not valid to port to 2022.3.

src/common/transformations/src/transformations/common_optimizations/compress_float_constants.cpp
```diff
@@ -27,8 +27,8 @@ std::shared_ptr<ov::Node> change_constant_precision_to_fp16(std::shared_ptr<ov::
 
     int num_out_of_range = 0;
     for (size_t i = 0; i < size; ++i) {
-        // if it is smaller than the smallest positive normal fp16 but not zero
-        if (std::abs(src_data[i]) <= ov::float16::from_bits(0x0400) && src_data[i] != 0.0f) {
+        // if abs value is smaller than the smallest positive fp16, but not zero
+        if (std::abs(src_data[i]) < ov::float16::from_bits(0x0001) && src_data[i] != 0.0f) {
```
rkazants, line 31: I see the different meaning about the minimum positive FP16 number. Please see: https://stackoverflow.com/questions/56711084/represents-fp16-minimum-number-in-hex-format
Your current version is aligned with it.

rkazants, line 31: Here is a different opinion that the version in the master is correct: https://stackoverflow.com/questions/56711084/represents-fp16-minimum-number-in-hex-format

This is saved in bits representation for which 0x0400 is correct.
Please elaborate why?

pavel-esir: Version in the master is not quite correct because it causes accuracy degradation :) `0x0004` is minimal positive **normalized** value but minimal positive **denormalized/subnormal** is `0x0001`, and by the link you provide this information is confirmed.

image

Therefore the current fix is correct

rkazants (approved):

PR #15576, opened 2023-02-08
[PT FE] Fix aten::flatten, add more tests

mvafin:
### Details:
 - *Fix `aten::flatten` and add tests*
 - *Add tests for `aten::linear` and `aten::transpose`*

### Tickets:
 - *None*


eaidova (approved):

src/frontends/pytorch/src/op/flatten.cpp
```diff
@@ -3,59 +3,65 @@
...
 
 namespace ov {
 namespace frontend {
 namespace pytorch {
 namespace op {
 
+using namespace ov::op;
+
 OutputVector translate_flatten(NodeContext& context) {
     auto start_dim = context.const_input<int64_t>(1);
```
rkazants, line 23: please check a number of inputs

src/frontends/pytorch/src/op/flatten.cpp
```diff
@@ -3,59 +3,65 @@
...
     auto end_dim = context.const_input<int64_t>(2);
 
-    auto shape = std::make_shared<opset10::ShapeOf>(context.get_input(0), element::i32);
-    auto rank_ = std::make_shared<opset10::ShapeOf>(shape, element::i32);
-    auto rank = std::make_shared<opset10::Squeeze>(rank_);
-    // Use opset::If for dim normalization
+    auto shape = std::make_shared<v3::ShapeOf>(context.get_input(0), element::i32);
+    auto rank_ = std::make_shared<v3::ShapeOf>(shape, element::i32);
+    auto rank = std::make_shared<v0::Squeeze>(rank_);
```
rkazants, line 28:
```suggestion
    Output<Node> rank = std::make_shared<v3::ShapeOf>(shape, element::i32);
    rank = std::make_shared<v0::Squeeze>(rank);
```

rkazants: Also, it is good to have such primitive in utils to compute a rank. Please consider - very useful stuff:)

src/frontends/pytorch/src/op/flatten.cpp
```diff
@@ -3,59 +3,65 @@
...
+    // Use opset::If for dim normalization. For now we only have flatten with constant start and end
```
rkazants, line 29:
```suggestion
```

rkazants: Looks like artifact

src/frontends/pytorch/src/op/flatten.cpp
```diff
@@ -3,59 +3,65 @@
...
     auto start_dim_node = context.get_input(1);
     auto end_dim_node = context.get_input(2);
     if (start_dim < 0) {
-        start_dim_node = std::make_shared<opset10::Add>(rank, start_dim_node);
+        start_dim_node = context.mark_node(std::make_shared<v1::Add>(rank, start_dim_node));
     }
```
rkazants, line 34: I think we can avoid any restriction to have constant `start_dim`, etc. Can we use Min, Max and Abs operations to avoid any `if`?

src/frontends/pytorch/src/op/flatten.cpp
```diff
@@ -3,59 +3,65 @@
...
-    auto neg_1_const = opset10::Constant::create(element::i32, Shape{1}, {-1});
-    auto axis_0 = opset10::Constant::create(element::i32, Shape{1}, {0});
-    auto start_dim_node_ = std::make_shared<opset10::Unsqueeze>(start_dim_node, axis_0);
-    auto new_shape = std::make_shared<opset10::ScatterElementsUpdate>(zeros, start_dim_node_, neg_1_const, axis_0);
+    // Slice shape from begin and end, then concat with -1, if slice return empty tensor concat shuold still be able to
+    // work with it
+    auto zero = v0::Constant::create(element::i32, Shape{1}, {0});
+    auto one = v0::Constant::create(element::i32, Shape{1}, {1});
+    auto int_max = v0::Constant::create(element::i32, Shape{1}, {std::numeric_limits<int32_t>::max()});
+    auto start_dim_u = std::make_shared<v0::Unsqueeze>(start_dim_node, zero);
```
rkazants, line 43: I think you can not squeeze `rank` and in this case you do not need this `Unsqueeze`. Please consider this optimization.

src/frontends/pytorch/src/op/flatten.cpp
mvafin: Not really, current implementation rely on constant dims

src/frontends/pytorch/src/op/flatten.cpp
mvafin: Agree, will do it separately

src/frontends/pytorch/src/op/flatten.cpp
mvafin: Have to unsqueeze since not always we add `rank` to it

src/frontends/pytorch/src/op/flatten.cpp
mvafin: Not sure how to do it and it is not required for now

rkazants (approved):

PR #15583, opened 2023-02-08
TopK v11 - specification

tomdol:
### Details:
 - Specification of the new version of the TopK operator
 - Introduction of opset11

### Tickets:
 - 101498


docs/OV_Runtime_UG/Operations_specifications.md
```diff
@@ -202,6 +202,7 @@
    Tile-1 <openvino_docs_ops_movement_Tile_1>
    TopK-1 <openvino_docs_ops_sort_TopK_1>
    TopK-3 <openvino_docs_ops_sort_TopK_3>
+   TopK-4 <openvino_docs_ops_sort_TopK_4>
```
rkazants, line 205: I think it should be `TopK-11` since this is introduced in opset11

docs/ops/sort/TopK_4.md
```diff
@@ -0,0 +1,116 @@
+`# TopK  {#openvino_docs_ops_sort_TopK_4}
+
+**Versioned name**: *TopK-4*
```
rkazants, line 3: TopK-11

docs/ops/sort/TopK_4.md
```diff
@@ -0,0 +1,116 @@
...
+* *axis*
+
+  * **Description**: Specifies the axis along which the values are retrieved.
+  * **Range of values**: An integer. Negative values means counting dimension from the back.
+  * **Type**: `int`
+  * **Required**: *yes*
+
+* *mode*
+
+  * **Description**: Specifies whether TopK selects the largest or the smallest elements from each slice.
```
rkazants, line 20:
```suggestion
  * **Description**: Specifies whether *TopK* selects the largest or the smallest elements from each slice.
```

docs/ops/sort/TopK_4.md
```diff
@@ -0,0 +1,116 @@
...
+* *sort*
+
+  * **Description**: Specifies the order of corresponding elements of the output tensor.
+  * **Range of values**: `value`, `index`, `none`
+  * **Type**: `string`
+  * **Required**: *yes*
+
+* *stable*
+
+  * **Description**: Specifies whether the equivalent elements should maintain their relative order from the input tensor. Takes effect only if the `sort` attribute is set to `value` or `index`.
```
rkazants, line 34: Not sure how it will work for `index` case. Please check. It makes sense to me only in case `value` mode

docs/ops/sort/TopK_4.md
```diff
@@ -0,0 +1,116 @@
...
+**Detailed Description**
+
+The output tensor is populated by values computed in the following way:
+
+    output[i1, ..., i(axis-1), j, i(axis+1) ..., iN] = top_k(input[i1, ...., i(axis-1), :, i(axis+1), ..., iN]), k, sort, mode)
+
+So for each slice `input[i1, ...., i(axis-1), :, i(axis+1), ..., iN]` *TopK* values are computed individually.
+
+Sorting and minimum/maximum are controlled by `sort` and `mode` attributes:
+  * *mode*=`max`, *sort*=`value` - descending by value
```
rkazants, line 76: please add `stable` attribute value in this list

docs/ops/sort/TopK_4.md
```diff
@@ -0,0 +1,116 @@
...
+            <dim>224</dim>
+        </port>
+        <port id="1">
+        </port>
+    <output>
+        <port id="2">
+            <dim>1</dim>
+            <dim>3</dim>
+            <dim>10</dim>
+            <dim>224</dim>
```
rkazants, line 106: `axis=3` shoud be this dimension. Please check

rkazants, line 106:
```suggestion
            <dim>224</dim>
            <dim>10</dim>
```

docs/ops/sort/TopK_4.md
```diff
@@ -0,0 +1,116 @@
...
+        </port>
+        <port id="3">
+            <dim>1</dim>
+            <dim>3</dim>
+            <dim>10</dim>
+            <dim>224</dim>
```
rkazants, line 112:
```suggestion
            <dim>224</dim>
            <dim>10</dim>
```

docs/OV_Runtime_UG/Operations_specifications.md
```diff
@@ -202,6 +202,7 @@
    Tile-1 <openvino_docs_ops_movement_Tile_1>
    TopK-1 <openvino_docs_ops_sort_TopK_1>
    TopK-3 <openvino_docs_ops_sort_TopK_3>
+   TopK-4 <openvino_docs_ops_sort_TopK_11>
```
mitruska, line 205:
```suggestion
   TopK-11 <openvino_docs_ops_sort_TopK_11>
```

docs/ops/sort/TopK_11.md
```diff
@@ -0,0 +1,118 @@
+`# TopK  {#openvino_docs_ops_sort_TopK_11}
```
mitruska, line 1:
```suggestion
# TopK  {#openvino_docs_ops_sort_TopK_11}
```

docs/ops/sort/TopK_11.md
```diff
@@ -0,0 +1,118 @@
...
+<layer ... type="TopK" ... >
+    <data axis="3" mode="max" sort="value" stable="true" index_element_type="i64"/>
+    <input>
+        <port id="0">
+            <dim>1</dim>
+            <dim>3</dim>
+            <dim>224</dim>
+            <dim>224</dim>
+        </port>
+        <port id="1">
```
nshchego, line 101: Please add the actual value like [here.](https://github.com/openvinotoolkit/openvino/blob/master/docs/ops/sort/TopK_1.md)

tomdol: There's a comment just above the example XML that `K` is assumed to be equal to 10. This section is supposed to describe the shapes(not tensor values) and I thought that this would be more consistent.

docs/ops/sort/TopK_11.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Inputs**:
+
+*   **1**: tensor with arbitrary rank and type *T*. **Required.**
+
+*   **2**: The value of *K* -- a scalar of any integer type that specifies how many elements from the input tensor should be selected. **Required.**
+
+**Outputs**:
+
+*   **1**: Output tensor of type *T* with at most *k* values from the input tensor along specified dimension *axis*. The shape of the tensor is `[input1.shape[0], ..., input1.shape[axis-1], 1..k, input1.shape[axis+1], ..., input1.shape[input1.rank - 1]]`.
```
rkazants, line 57: can you please elaborate how we pad the output if `k` greater `input1.shape[axis]`?

rkazants (commented):
Last comment left. The rest looks good to me.

docs/ops/sort/TopK_11.md
tomdol: TF, PT and ONNXRT throw an exception and cancel the inference in such case. I think we should do the same.

tomdol: The CPU plugin throws an exception and doesn't allow an inference with K value larger than the dimension at axis. The GPU plugin does not throw and it returns a tensor with exactly K values.

rkazants: I think we need to say about the undefined behavior in this case explicitly. This is important to cover this corner case in the spec explicitly. It will simplify understanding of this specification. Moreover, the different plugins behave differently for this corner case now so this statement will be aligned with it.

tomdol: I've added this information to the second input description. It now defines a range of allowed values of K.

rkazants (approved):

docs/ops/sort/TopK_4.md
xuchen-intel: @rkazants Hi Roman! If I understand it correctly, for sorting by index, sometimes we don't have the chance to collect all the equivalent values in the stage of sorting by data, to do the next stage of sorting by index, because of the capacity of specific k. For such case, if we just pick one of these equivalent values randomly, that would violate our description "the equivalent elements should maintain their relative order from the input tensor". So I think the attribute of stable also make sense to sorting by index.

Here is the case I met, when I was implementing TopK-3 for CPU plugin.

```
Input tensor: {5, 3, 1, 2, 5, 5}, sort_mode=min, sort_type=indices, k=4

Possible result1:
Output data tensor: {5, 3, 1, 2}
Output index tensor: {0, 1, 2, 3}

Possible result2:
Output data tensor: {3, 1, 2, 5}
Output index tensor: {1, 2, 3, 4}

Possible result3:
Output data tensor: {3, 1, 2, 5}
Output index tensor: {1, 2, 3, 5}
```

If the attribute of stable is true, "Possible result1" is the only right result.

cc @tomdol @dmitry-gorokhov

docs/ops/sort/TopK_1.md
```diff
@@ -51,15 +51,15 @@
...
+The output tensor is populated by values computed in the following way:
 
     output[i1, ..., i(axis-1), j, i(axis+1) ..., iN] = top_k(input[i1, ...., i(axis-1), :, i(axis+1), ..., iN]), k, sort, mode)
 
 So for each slice `input[i1, ...., i(axis-1), :, i(axis+1), ..., iN]` which represents 1D array, top_k value is computed individually.
 
 Sorting and minimum/maximum are controlled by `sort` and `mode` attributes:
   * *mode*=`max`, *sort*=`value` - descending by value
-  * *mode*=`max`, *sort*=`index` - ascending by index
+  * *mode*=`max`, *sort*=`index` - descending by index
```
xuchen-intel, line 62: @tomdol Hi Tomasz, could you please elaborate the reason why we want descending sort for sort=index now, even for mode=max? Because if this is the behavior we want for TopK_1, as well as TopK_3 and TopK_11, then I think all the implementation in ref topk and plugin topk should revise accordingly.
cc @dmitry-gorokhov

tomdol: I've only updated the spec because I think the previous statement was incorrect and could have been a copy-and-paste error. The `mode` attribute determines the order of elements while `sort` decides whether the sorting should be done by comparing the indices or by comparing the values. I could revert those changes in the spec if the previous version was aligned with the code.

xuchen-intel: Thanks Tomasz for the prompt reply. I think the original spec, always applying ascend sorting for sort=index, is reasonable. And yes, they are aligned with the code. Please revert those changes. Thanks a lot!

cc @dmitry-gorokhov @rkazants

docs/ops/sort/TopK_4.md
rkazants: @xuchen-intel, you seem to be correct. How does the current version work for `sort_type=indices`? It does not guarantee "Possible result1", right?

I think we need to cover this case as well for TensorFlow TopK operations.
@xuchen-intel, how hard is it to add into the CPU plugin?

cc, @tomdol

xuchen-intel: @rkazants Hi Roman! Current topk-11 in CPU plugin has already supported stable sorting for sort_type=indices. When stable is set for sort_type=indices, it guarantees "Possible result1", because sorting by indices reuses routines of sorting by values in CPU plugin.

cc @tomdol  @dmitry-gorokhov

rkazants: @xuchen-intel Hi Chen! That is just awesome that the CPU plugin already supports it.

@tomdol, can you please enable it on the core part and cover this case in the specification? I thin we should have it in this release to be aligned with TF.

Best regards,
Roman

PR #15594, opened 2023-02-09
Support dynamic output models with all possible devices instead of CPU only

songbell:
### Details:
 - multi need to use setBlob to target hardware infer requests, when the network is with dynamic output, the setBlob will fail at blob dimension checking. previously in 22.3 release, we temporarily work around this by directing CPU to run these kinds of networks
 - use upper bound of dims to replace the zero value dimension
 - restruct auto dir to src+tests/unit
 - add some infer request tests

### Tickets:
 - 99668


src/plugins/auto/CMakeLists.txt
```diff
@@ -9,14 +9,14 @@ endif()
 
 set (TARGET_NAME "openvino_auto_plugin")
 
-file(GLOB SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp ${CMAKE_CURRENT_SOURCE_DIR}/utils/*.cpp)
+file(GLOB SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp ${CMAKE_CURRENT_SOURCE_DIR}/src/utils/*.cpp)
```
ilyachur, line 12: If you move files, please also update the README for the plugin

songbell: yeah, thanks for reminding, will update then

wangleis (approved):

PR #15601, opened 2023-02-09
optimize shape_infer time of matmul

xczhai:
### Details:
 - simplify the shape infer flow for matmul

### Tickets:
 - 100506

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/matmul.cpp
+++ b/src/plugins/intel_cpu/src/nodes/matmul.cpp
@@ -159,7 +159,6 @@ public:
     }
 
 private:
-    const std::shared_ptr<const ngraph::opset1::MatMul> m_matmul;
     VectorDims m_shapeY;
     const size_t m_out_rank;
     const bool m_transpose_a;
@@ -170,11 +169,14 @@ class MMShapeInferFactory : public ShapeInferFactory {
 public:
     MMShapeInferFactory(const std::shared_ptr<ngraph::Node>& op) : m_op(op) {}
     ShapeInferPtr makeShapeInfer() const override {
-        const auto matmul = std::dynamic_pointer_cast<const ngraph::opset1::MatMul>(m_op);
-        const auto output_rank = matmul->get_output_partial_shape(0).rank().get_length();
-        const bool transpose_a = matmul->get_transpose_a();
-        const bool transpose_b = matmul->get_transpose_b();
-        return std::make_shared<MMShapeInfer>(output_rank, transpose_a, transpose_b);
+        if (const auto matmul = ov::as_type_ptr<const ngraph::opset1::MatMul>(m_op)) {
+            const auto output_rank = matmul->get_output_partial_shape(0).rank().get_length();
+            const bool transpose_a = matmul->get_transpose_a();
+            const bool transpose_b = matmul->get_transpose_b();
+            return std::make_shared<MMShapeInfer>(output_rank, transpose_a, transpose_b);
+       } else {
+             IE_THROW() << "Unexpected operation type in the MatMul shape inference factory";
+       }
     }
 private:
     std::shared_ptr<ngraph::Node> m_op;
--- a/src/plugins/intel_cpu/src/nodes/matmul.cpp
+++ b/src/plugins/intel_cpu/src/nodes/matmul.cpp
@@ -112,17 +112,16 @@ bool MatMul::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op,
 
 class MMShapeInfer : public ShapeInferEmptyPads {
 public:
-    MMShapeInfer(const std::shared_ptr<const ngraph::opset1::MatMul>& op) : m_matmul(op) {
-        m_out_rank = m_matmul->get_output_partial_shape(0).rank().get_length();
+    MMShapeInfer(const size_t& out_rank, const bool& transpose_a, const bool& transpose_b) : m_out_rank(out_rank), m_transpose_a(transpose_a), m_transpose_b(transpose_b) {
         m_YShape = VectorDims(m_out_rank, 1); // for output and cache
     }
     std::vector<VectorDims> infer(
         const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
         const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
-        VectorDims AShape = input_shapes[0].get();
-        VectorDims BShape = input_shapes[1].get();
-        const size_t& Aranks = AShape.size();
-        const size_t& Branks = BShape.size();
+        const VectorDims& shapeA = input_shapes[0].get();
+        const VectorDims& shapeB = input_shapes[1].get();
+        const size_t rankA = shapeA.size();
+        const size_t rankB = shapeB.size();
 
         // getSupportedDescriptors has done some shape check.
         // 1. Needn't assert the scalar type since the matmul_shape_inference has checked.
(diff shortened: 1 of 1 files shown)


src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -110,8 +110,75 @@ bool MatMul::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op,
...
+public:
+    MMShapeInfer(const std::shared_ptr<const ngraph::opset1::MatMul>& op) : m_matmul(op) {
+        m_out_rank = m_matmul->get_output_partial_shape(0).rank().get_length();
+        m_YShape = VectorDims(m_out_rank, 1); // for output and cache
+    }
+    std::vector<VectorDims> infer(
+        const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
+        const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
+        VectorDims AShape = input_shapes[0].get();
+        VectorDims BShape = input_shapes[1].get();
```
maxnick, line 123:
```suggestion
        const VectorDims& shapeA = input_shapes[0].get();
        const VectorDims& shapeB = input_shapes[1].get();
```
When you declare just `VectorDims` copy constructor is called. But we need just to read them, so taking the reference is better.

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -110,8 +110,75 @@ bool MatMul::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op,
...
+        const size_t& Aranks = AShape.size();
+        const size_t& Branks = BShape.size();
```
maxnick, line 125: Here in opposite. The size of the `ptr` is the same as the `size_t` so there is no sense to take by reference.
```suggestion
        const size_t rankA = AShape.size();
        const size_t rankB = BShape.size();
```
Regarding the naming. Usually we try to start variables and method names with a lowercase letter.

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -110,8 +110,75 @@ bool MatMul::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op,
...
+        // 4. transpose is necessary
+        // 5. Just support the same rank of matmul
+        // 6. simplify the broadcast check
+        if (Aranks == 1 && Branks == 1 && AShape[0] == BShape[0]) {
+            return {m_YShape};
+        }
+
+        // transpose
+        const bool& transpose_a = m_matmul->get_transpose_a();
+        const bool& transpose_b = m_matmul->get_transpose_b();
```
maxnick, line 140: Again, why not just take by value?

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -110,8 +110,75 @@ bool MatMul::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op,
...
+
+        return {m_YShape};
+    }
+
+    port_mask_t get_port_mask() const override {
+        return EMPTY_PORT_MASK;
+    }
+
+private:
+    const std::shared_ptr<const ngraph::opset1::MatMul> m_matmul;
```
maxnick, line 164: I would say it is better to store `bool transpose_a` and `bool transpose_b`. So it gives faster access to those fields than through the operation pointer.

maxnick (changes requested):

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -110,8 +110,75 @@ bool MatMul::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op,
...
+
+        for (size_t i=0; i < m_out_rank-2; ++i) {
+            size_t max = std::max(AShape.at(i), BShape.at(i));
+            size_t min = std::min(AShape.at(i), BShape.at(i));
+            if ((AShape.at(i) == BShape.at(i)) || (min == 1)) {
+                m_YShape[i] = max;
+            } else {
+                IE_THROW() << "Incompatible MatMul batch dimension. Cant merge the first input dimension=" <<
+                              AShape[i] << " with second input dimension=" << BShape[i] << " at index=" << i;
+            }
```
maxnick, line 153:
```suggestion
            size_t max = std::max(AShape.at(i), BShape.at(i));
            size_t min = std::min(AShape.at(i), BShape.at(i));
            if (AShape.at(i) != BShape.at(i)) {
                if (BShape.at(i) == 1) {
                    m_YShape[i] = AShape.at(i);
                } else if (AShape.at(i) != 1) {
                                  IE_THROW() << "Incompatible MatMul batch dimension. Cant merge the first input dimension=" <<
                              AShape[i] << " with second input dimension=" << BShape[i] << " at index=" << i;
                }
            }
             m_YShape[i] = BShape.at(i);
```
So that we alwost always just copy the data and skip the min/max comparisons.

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -110,8 +110,75 @@ bool MatMul::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op,
...
+        // 6. simplify the broadcast check
+        if (Aranks == 1 && Branks == 1 && AShape[0] == BShape[0]) {
+            return {m_YShape};
+        }
+
+        // transpose
+        const bool& transpose_a = m_matmul->get_transpose_a();
+        const bool& transpose_b = m_matmul->get_transpose_b();
+
+        m_YShape.at(m_out_rank-2) = transpose_a ? AShape[Aranks-1] : AShape[Aranks-2];
```
maxnick, line 142: Regarding `at` call. It is more expensive than `[]` since it contains range check. I suggest checking the sizes once at the beginning, that should be enough.

src/plugins/intel_cpu/src/nodes/matmul.cpp
xczhai: > When you declare just `VectorDims` copy constructor is called. But we need just to read them, so taking the reference is better.

Done

src/plugins/intel_cpu/src/nodes/matmul.cpp
xczhai: > Here in opposite. The size of the `ptr` is the same as the `size_t` so there is no sense to take by reference.
> 
> Regarding the naming. Usually we try to start variables and method names with a lowercase letter.

Done

src/plugins/intel_cpu/src/nodes/matmul.cpp
xczhai: > Again, why not just take by value?

Done

src/plugins/intel_cpu/src/nodes/matmul.cpp
xczhai: > I would say it is better to store `bool transpose_a` and `bool transpose_b`. So it gives faster access to those fields than through the operation pointer.

Yes. move it to private member.

src/plugins/intel_cpu/src/nodes/matmul.cpp
xczhai: > 

Yes. Indeed. at is safe but slow. replace at by []

src/plugins/intel_cpu/src/nodes/matmul.cpp
xczhai: Your method is better.

xczhai: > 

Your method is better.

xczhai: > So that we alwost always just copy the data and skip the min/max comparisons.

Your method is better.

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -110,8 +110,76 @@ bool MatMul::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op,
...
+            return {m_shapeY};
+        }
+
+        m_shapeY[m_out_rank-2] = m_transpose_a ? shapeA[rankA-1] : shapeA[rankA-2];
+        m_shapeY[m_out_rank-1] = m_transpose_b ? shapeB[rankB-2] : shapeB[rankB-1];
+
+        for (size_t i=0; i < m_out_rank-2; ++i) {
+            if (shapeA[i] != shapeB[i]) {
+                if (shapeB[i] == 1) {
+                    m_shapeY[i] = shapeA[i];
```
maxnick, line 143:
```suggestion
                if (shapeB[i] == 1) {
                    m_shapeY[i] = shapeA[i];
                    continue;
```

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -110,8 +110,78 @@ bool MatMul::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op,
...
+};
+
+class MMShapeInferFactory : public ShapeInferFactory {
+public:
+    MMShapeInferFactory(const std::shared_ptr<ngraph::Node>& op) : m_op(op) {}
+    ShapeInferPtr makeShapeInfer() const override {
+        const auto matmul = std::dynamic_pointer_cast<const ngraph::opset1::MatMul>(m_op);
+        const auto output_rank = matmul->get_output_partial_shape(0).rank().get_length();
+        const bool transpose_a = matmul->get_transpose_a();
+        const bool transpose_b = matmul->get_transpose_b();
```
maxnick, line 176:
```suggestion
        if (const auto matmul = ov::as_type_ptr<const ngraph::opset1::MatMul>(m_op) ) {
            const auto output_rank = matmul->get_output_partial_shape(0).rank().get_length();
            const bool transpose_a = matmul->get_transpose_a();
            const bool transpose_b = matmul->get_transpose_b();
       } else {
             IE_TROW() << "Unexpected operation type in the MatMul shape inference factory";
       }
```
OV RTTI usually is faster than compiler generated one.

src/plugins/intel_cpu/src/nodes/matmul.cpp
alanzhai219: > Why do we still have this data member then?

Sry for this. Have removed it.

xczhai: > Why do we still have this data member then?

Sry for this. Have removed it. thx

maxnick (approved):

PR #15610, opened 2023-02-09
Review scatter nd update class for shape inference aspects

praasz:
### Details:
 - Review for interval shape and label propagation.
 - Review template implementation of `shape_infer`, add default ctor tests.
 - Add bounds and label evaluators.
 - Add new default label evaluator which propagate labels from listed inputs.
 - Use new label evaluator scatter update and previous version of evaluator.

### Tickets:
 - 97255
 - 97283
 - 97332
 - 97381
 - 97533
 - 97582


mitruska (approved):

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
```diff
@@ -579,7 +579,6 @@ const IShapeInferCommonFactory::TRegistry IShapeInferCommonFactory::registry{
     _OV_OP_SHAPE_INFER_REG(ScatterElementsUpdate, entryIOC),
     _OV_OP_SHAPE_INFER_REG(ScatterNDUpdate, entryIO),
     _OV_OP_SHAPE_INFER_REG(Select, entryIO),
-    _OV_OP_SHAPE_INFER_REG(Select, entryIO),
```
jane-intel, line 582: Is it intentional?

jane-intel (approved):

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
praasz: Yes it is just duplicated line (missed or some merge result).

PR #15620, opened 2023-02-09
Fix time test example when the tensor shape is higher than INT16

pereanub:
### Details:
 - *Fix time test example when the tensor shape is INT32 or INT64*

### Tickets:
 - [*CVS-102074*](https://jira.devtools.intel.com/browse/CVS-102074)


pereanub:
The code has been changed in [PR15644](https://github.com/openvinotoolkit/openvino/pull/15644)

PR #15628, opened 2023-02-10
[TF FE] Fix sporadic failure for TF1 LRN layer test

rkazants:
**Details:** Fix sporadic failure for TF1 LRN layer test

**Ticket:** 103487

### DIFF
--- a/src/frontends/tensorflow_common/src/op/lrn.cpp
+++ b/src/frontends/tensorflow_common/src/op/lrn.cpp
@@ -26,7 +26,12 @@ OutputVector translate_lrn_op(const NodeContext& node) {
     auto beta = static_cast<double>(node.get_attribute<float>("beta", 0.5));
 
     // adjust attribute values for opset LRN operation
-    size_t attr_size = static_cast<size_t>(depth_radius) * 2 + 1;
+    size_t attr_size = 1;
+    if (depth_radius % 2 == 0) {
+        attr_size = static_cast<size_t>(depth_radius) * 2 + 1;
+    } else {
+        attr_size = static_cast<size_t>(depth_radius - 1) * 2 + 1;
+    }
     alpha *= static_cast<double>(attr_size);
 
     // currently, plugins fallback to the template to execute this operation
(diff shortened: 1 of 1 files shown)


src/frontends/tensorflow_common/src/op/lrn.cpp
```diff
@@ -26,7 +26,12 @@ OutputVector translate_lrn_op(const NodeContext& node) {
     auto beta = static_cast<double>(node.get_attribute<float>("beta", 0.5));
 
     // adjust attribute values for opset LRN operation
-    size_t attr_size = static_cast<size_t>(depth_radius) * 2 + 1;
+    size_t attr_size = 1;
+    if (depth_radius % 2 == 0) {
+        attr_size = static_cast<size_t>(depth_radius) * 2 + 1;
+    } else {
+        attr_size = static_cast<size_t>(depth_radius - 1) * 2 + 1;
```
mvafin, line 33:
```suggestion
        attr_size = static_cast<size_t>(depth_radius) * 2 - 1;
```

src/frontends/tensorflow_common/src/op/lrn.cpp
```diff
@@ -35,7 +40,7 @@ OutputVector translate_lrn_op(const NodeContext& node) {
     // so we have to transpose inputs and outputs due to this limitation
     auto axis = make_shared<Constant>(element::i32, Shape{1}, 1);
     ov::frontend::tensorflow::convert_nhwc_to_nchw(true, input);
-    auto lrn = make_shared<LRN>(input, axis, alpha, beta, bias, attr_size)->output(0);
+    Output<Node> lrn = make_shared<LRN>(input, axis, alpha, beta, bias, attr_size);
```
mvafin, line 43: should work with `auto` too, but this is good also

src/frontends/tensorflow_common/src/op/lrn.cpp
mvafin: In this case, could you add comment about what is done here?

mvafin (approved):
I approve, to not block the fix, but I would like to see a layer test for this.

PR #15636, opened 2023-02-10
[PyOV] Fix issues with RTMap

akuporos:
### Details:
 - implemented custom iterator `PyRTMapIterator` in order to return `OVAny` python object from `ov::Any`.
 - `model.get_rt_info(*args)` returns now `OVAny` object (previously we tried to convert it to some value)
 - added methods `.astype(dtype)` and `.aslist(dtype)` for `OVAny` class using which a user can cast a runtime attribute to desired value from OVAny. 

### Tickets:
 - 98169
 - 97897


src/bindings/python/src/pyopenvino/graph/any.cpp
```diff
@@ -3,7 +3,7 @@
 //
 
 #include "openvino/core/any.hpp"
-
+#include "meta_data.hpp"
```
akuporos, line 6: move all things connected with `meta_data` to `utils.cpp`

src/bindings/python/src/pyopenvino/utils/utils.cpp
```diff
@@ -97,6 +105,9 @@ py::object from_ov_any(const ov::Any& any) {
     // Check for std::map<element::Type, float>
     else if (any.is<std::map<ov::element::Type, float>>()) {
         return py::cast(any.as<std::map<ov::element::Type, float>>());
+    }  // Check for ov::AnyMap (std::map<std::string, ov::Any)
```
akuporos, line 108:
```suggestion
    }  // Check for ov::AnyMap (std::map<std::string, ov::Any>)
```

src/bindings/python/tests/test_runtime/test_model.py
```diff
@@ -560,6 +560,23 @@ def check_rt_info(model, serialized):
...
+                                                          "min": min_val, "threshold": threshold},
+                                     "type_of_model": "classification"}
+
+        for key, value in rt_info.items():
+            if key == "config":
+                for config_value in value:
+                    assert config_value in ["type_of_model", "converter_type", "model_parameters"]
+
+        for rt_info_val in model.get_rt_info(["config", "model_parameters", "labels", "label_tree"]):
+            assert rt_info_val in ["float_empty", "nodes", "type", "directed"]
```
ilyachur, line 578: I think you need to remove this if:
https://github.com/openvinotoolkit/openvino/pull/15636/files#diff-fc5bff420a93e90ef3bbea6d6b2bf55e76901d076d9584be0bcd98135479b4f3R523
In C++ we don't have difference between original and serialized models.

src/bindings/python/src/pyopenvino/utils/utils.cpp
```diff
@@ -108,6 +119,9 @@ py::object from_ov_any(const ov::Any& any) {
             PyDict_SetItemString(dict, property_name.c_str(), PyUnicode_FromString(mutability.c_str()));
         }
         return py::cast<py::object>(dict);
+    } else if (any.is<std::shared_ptr<ov::Meta>>()) {
+        const ov::AnyMap& as_map = *any.as<std::shared_ptr<ov::Meta>>();
+        return from_ov_any_map(as_map);
     } else if (any.is<ov::element::Type>()) {
```
jiwaszki, line 125:
```suggestion
    } else if (any.is<ov::Meta>()) {
        // impl...
    } else if (any.is<std::shared_ptr<ov::Meta>>()) {
        const ov::AnyMap& as_map = *any.as<std::shared_ptr<ov::Meta>>();
        return from_ov_any_map(as_map);
    } else if (any.is<ov::element::Type>()) {
```
Is `ov::Meta` without shared pointer a case as well?

jiwaszki (approved):

src/bindings/python/src/pyopenvino/utils/utils.cpp
akuporos: it doesn't make sense as `ov::Meta` is interface class with pure virtual methods.

src/bindings/python/tests/test_runtime/test_model.py
akuporos: done

src/bindings/python/src/pyopenvino/graph/rt_map.cpp
```diff
@@ -27,19 +28,56 @@ using PyRTMap = ov::RTMap;
...
+        } else {
+            std::pair<std::string, py::object> res = {result.first,
+                                                      Common::utils::from_ov_any_no_leaves(result.second)};
+            return py::cast(res);
+        }
+    }
+
+    const PyRTMap& py_rt_map;
+    bool is_value = false;
+    py::object ref;  // keep a reference
```
jiwaszki, line 57:
```suggestion
    py::object ref;  // keep a reference
```
Isn't this a copy? Can it be switched to actual reference or is there some mechanism under `py::object` to do it when copy occurs?

src/core/src/model.cpp
```diff
@@ -955,7 +955,7 @@ bool ov::Model::has_rt_info(const std::vector<std::string>& args) const {
             return false;
         if (i == args.size() - 1)
             break;
-        const ov::Any& rt_attr = get_rt_arg<std::string>(info, args[i]);
+        const ov::Any rt_attr = get_rt_arg<std::string>(info, args[i]);
```
jiwaszki, line 958: Is there a reason for a copy?

akuporos: yes,
`get_map_from_attr` just performs some checks, casting and returns *a reference*, so lets ignore it in the explanation.
`info = rt_attr` at line 959 performs two actions: destroys data inside `info` and copies data from `rt_attr` into `info`. I don't see if an order of actions is strictly defined, but it looks like on our Windows and MacOS configurations it is Destroy->Copy. If `rt_attr` is a reference to something inside `info`, then at the moment of copy data under `rt_attr` would be already destroyed.

src/bindings/python/src/pyopenvino/graph/any.cpp
```diff
@@ -62,6 +67,72 @@ void regclass_graph_Any(py::module m) {
...
+
+            :return: A runtime attribute.
+            :rtype: Any
+    )");
+    ov_any.def(
+        "aslist",
+        [](ov::Any& self, py::object dtype) {
+            // before serialization
+            if (self.is<Common::utils::EmptyList>() || dtype.is_none()) {
+                return py::cast(std::vector<int>());
```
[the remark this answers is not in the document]
akuporos: There isn't default behavior in C++ for empty vectors which are expected from `get_rt_info(args)`.  
So according to [the test in c++ api](https://github.com/openvinotoolkit/openvino/blob/master/src/core/tests/pass/serialization/serialize.cpp#L432) user still should define vector of `float/string/something_else` is expected even for empty.

I decided to use `int` because it's a very basic type and in this case there no difference what type of vector will be.

src/bindings/python/src/pyopenvino/graph/rt_map.cpp
jiwaszki: Discussed offline

akuporos: discussed offline

src/bindings/python/src/pyopenvino/graph/any.cpp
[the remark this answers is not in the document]
akuporos: discussed offline -- decided to return python list.

jiwaszki (approved):
🥇

src/core/src/model.cpp
ilyachur: Why do you need this change? Do you have a segfault or that? I don't expect such changes...
Moreover below in the line 969 we have the same code and it works well.
Please explain this change.

akuporos: Yes, I had a segfault on macos job and failed test on windows.
During the debug process I found that `info` had empty strings for keys after update.

p-wysocki (approved):
LGTM

src/core/src/model.cpp
ilyachur: Please create a ticket to understand the root cause

ilyachur (approved):

PR #15640, opened 2023-02-10
[GPU] Added shape agnostic kernel support for Range + fix dynamic tests

Lyamin-Roman:
### Tickets:
 - *95104*


sshlyapn (approved):
Looks good to me

PR #15644, opened 2023-02-10
Correct random number generation

evgenytalanin-intel:
### Details:
 - avoid division by zero error (and thus failing test suites) on Windows where RAND_MAX is only 0x7FFF

### Tickets:
 - 102074

### DIFF
--- a/tests/lib/src/common_utils.h
+++ b/tests/lib/src/common_utils.h
@@ -7,6 +7,7 @@
 #include <inference_engine.hpp>
 #include <openvino/openvino.hpp>
 #include <limits>
+#include <random>
 
 
 /**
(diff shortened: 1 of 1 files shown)


tests/lib/src/common_utils.h
```diff
@@ -88,8 +89,18 @@ ov::Tensor fillTensorRandom(T &input) {
     ov::Tensor tensor{input.get_element_type(), input.get_shape()};
     std::vector<U> values(ov::shape_size(input.get_shape()));
 
-    for (size_t i = 0; i < values.size(); ++i)
-        values[i] = 1 + static_cast<U> (rand()) / (static_cast<U> (RAND_MAX / (std::numeric_limits<U>::max() - 1)));
+    std::random_device rd;
+    std::mt19937 gen(rd());
+
+    if (std::is_floating_point<U>::value == true) {
```
vurusovs, line 95: @evgenytalanin-intel the same may be applicable to `fillBlobRandom` (but looks like it's for API1.0)

vurusovs (approved):

dkrupnov (approved):

PR #15650, opened 2023-02-10
[Python API] GraphComparator parameters propagated to python function

v-Golubev:


src/bindings/python/src/pyopenvino/test_utils/test_utils.cpp
```diff
@@ -13,20 +13,52 @@ namespace py = pybind11;
...
+           bool compare_accuracy,
+           bool compare_subgraph_descriptors,
+           bool compare_consumers_count) {
             const auto lhs_ptr = std::const_pointer_cast<ov::Model>(lhs.shared_from_this());
             const auto rhs_ptr = std::const_pointer_cast<ov::Model>(rhs.shared_from_this());
 
-            auto fc = FunctionsComparator::with_default()
-                          .enable(FunctionsComparator::ATTRIBUTES)
-                          .enable(FunctionsComparator::CONST_VALUES);
+            auto fc = FunctionsComparator::no_default().enable(FunctionsComparator::NODES);
```
akuporos, line 30: I believe it would be better to support enums in Python. For the reference, please, look [here](https://pybind11.readthedocs.io/en/stable/classes.html#enumerations-and-internal-types) and [here](https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/src/pyopenvino/graph/passes/transformations.cpp#L26).

akladiev:

akladiev:

PR #15651, opened 2023-02-10
Turn on onnx fallthrough in convert_model()

popovaan:
Description: 
ONNX_FALLTHROUGH allows to export PyTorch models to ONNX with unsupported operations as is. Without this option torch.onnx.export() fails to export such models.
Use case of turning on the option - converting PyTorch models with custom user extensions.
Ticket: 98995

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests - N/A
* [x]  Checked conversion on TorchVision and timm models, no regression occurred.
* [x]  Checked conversion on examples from https://github.com/openvinotoolkit/openvino_contrib/pull/481 This PR allows to convert all examples using following line:
`convert_model(model, example_input=inp, extensions="libuser_ov_extensions.so", use_legacy_frontend=True)`

Documentation:


tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -723,6 +723,7 @@ def convert_pytorch_to_onnx(model, input_shape, opset_version, example_inputs, o
     torch.onnx.export(model,
                       inputs,
                       model_onnx,
+                      operator_export_type=torch.onnx.OperatorExportTypes.ONNX_FALLTHROUGH,
```
rkazants, line 726: @popovaan, @eaidova, is it possible to add one small PyTorch model with unsupported operation please?

popovaan: Added test with small custom op, which checks that conversion to ONNX step passes.

tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
```diff
@@ -131,6 +131,7 @@ def convert_pytorch_to_onnx(model, input_shape, opset_version, example_inputs, o
     torch.onnx.export(model,
                       inputs,
                       model_onnx,
+                      operator_export_type=torch.onnx.OperatorExportTypes.ONNX_FALLTHROUGH,
                       **additional_params)
```
rkazants, line 135: not a blocker for this PR but we need to consider `try-except` to properly handle possible exceptions.
@popovaan, can you create separate PR please for this?

rkazants (approved):

rkazants:
@eaidova, can you please also take a look?

eaidova (approved):

tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
popovaan: Yes, I can add `try-except` is separate PR

PR #15652, opened 2023-02-10
[TF FE] Refactor Reciprocal translator and add layer test

rkazants:
**Details:** Refactor Reciprocal translator and add layer test

**Ticket:** 92874

### DIFF
--- a/src/frontends/tensorflow_common/src/op/reciprocal.cpp
+++ b/src/frontends/tensorflow_common/src/op/reciprocal.cpp
@@ -14,11 +14,13 @@ namespace tensorflow {
 namespace op {
 
 OutputVector translate_reciprocal_op(const NodeContext& node) {
+    // computes element-wise 1/x, where x - input
+    default_op_checks(node, 1, {"Reciprocal"});
     auto x = node.get_input(0);
-    auto ng_exponent = make_shared<Constant>(x.get_element_type(), Shape{}, -1);
-    auto res = make_shared<Power>(x, ng_exponent);
-    set_node_name(node.get_name(), res);
-    return res->outputs();
+    auto minus_one_const = make_shared<Constant>(x.get_element_type(), Shape{}, -1);
+    auto reciprocal = make_shared<Power>(x, minus_one_const);
+    set_node_name(node.get_name(), reciprocal);
+    return {reciprocal};
 }
 }  // namespace op
 }  // namespace tensorflow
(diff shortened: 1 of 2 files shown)


tests/layer_tests/tensorflow_tests/test_tf_Reciprocal.py
```diff
@@ -0,0 +1,43 @@
...
+        inputs_data['x'] = np.random.randint(1, 30, x_shape).astype(np.float32)
+
+        return inputs_data
+
+    def create_reciprocal_net(self, x_shape, x_type):
+        tf.compat.v1.reset_default_graph()
+        # Create the graph and model
+        with tf.compat.v1.Session() as sess:
+            x = tf.compat.v1.placeholder(tf.float32, x_shape, 'x')
+            tf.raw_ops.Reciprocal(x=x)
```
gkrivor, line 24: Could you check, this test looks like unary test like a listed in test_tf_UnaryOps.py. I understand when we test different dtypes, but in case of single float32 maybe better to extend UnaryOps.py

src/frontends/tensorflow_common/src/op/reciprocal.cpp
```diff
@@ -14,11 +14,13 @@ namespace tensorflow {
...
 
 OutputVector translate_reciprocal_op(const NodeContext& node) {
+    // computes element-wise 1/x, where x - input
+    default_op_checks(node, 1, {"Reciprocal"});
     auto x = node.get_input(0);
-    auto ng_exponent = make_shared<Constant>(x.get_element_type(), Shape{}, -1);
-    auto res = make_shared<Power>(x, ng_exponent);
-    set_node_name(node.get_name(), res);
-    return res->outputs();
+    auto minus_one_const = make_shared<Constant>(x.get_element_type(), Shape{}, -1);
```
gkrivor, line 20: Shape{1} or Shape{} is correct?

rkazants: a scalar (i.e. `Shape{}`) is the correct value because the input can be a scalar.

tests/layer_tests/tensorflow_tests/test_tf_Reciprocal.py
rkazants: Let me fix in the different PR. The idea originally was to test `int32` type as well but TF does not support it:)

gkrivor (approved):

PR #15653, opened 2023-02-10
[TF FE] Refactor Size translator and add layer test

rkazants:
**Details:** Refactor Size translator and add layer test

**Ticket:** 92876

### DIFF
--- a/src/frontends/tensorflow_common/src/op/size.cpp
+++ b/src/frontends/tensorflow_common/src/op/size.cpp
@@ -6,6 +6,7 @@
 #include "openvino/opsets/opset8.hpp"
 
 using namespace std;
+using namespace ov;
 using namespace ov::opset8;
 
 namespace ov {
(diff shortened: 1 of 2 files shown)


src/frontends/tensorflow_common/src/op/size.cpp
```diff
@@ -14,13 +15,23 @@ namespace tensorflow {
...
+    // retrive attribute of the output type
+    auto out_type = node.get_attribute<element::Type>("out_type", element::i32);
+
+    // introduce extra dimension in order to compute size in case of a scalar input
+    auto const_zero = make_shared<Constant>(element::i32, Shape{1}, 0);
+    input = make_shared<Unsqueeze>(input, const_zero);
+
+    // compute the input tensor size
+    auto shape_of = make_shared<ShapeOf>(input, out_type);
+    auto axis = make_shared<Constant>(element::i32, Shape{}, 0);
```
gkrivor, line 31: I don't get, should we use Shape{1} or we may use Shape[] instead. Could you take a look on this line before merge?

tests/layer_tests/tensorflow_tests/test_tf_Size.py
```diff
@@ -0,0 +1,45 @@
...
+
+        return inputs_data
+
+    def create_size_net(self, input_shape, input_type, out_type):
+        self.input_type = input_type
+        tf.compat.v1.reset_default_graph()
+        # Create the graph and model
+        with tf.compat.v1.Session() as sess:
+            input = tf.compat.v1.placeholder(input_type, input_shape, 'input')
+            tf.raw_ops.Size(input=input, out_type=out_type)
```
gkrivor, line 26: Could you check this place. What do we test in case of out_type==None? Because regarding https://www.tensorflow.org/api_docs/python/tf/raw_ops/Size default value is int32. I may expect we need to test some "default behavior" in case of None. In such case we need to add here a condition, like we do it previously.

rkazants: `None` means the default behavior or value is undefined for this parameter.
Yes, we did in the different way previously but I think it the same. TF accepts it.

gkrivor (approved):

PR #15656, opened 2023-02-12
[PT FE] Check number of inputs, use rank function

mvafin:
### Details:
 - *Check number of inputs*
 - *Use rank function in op translators*
 - *Remove `TupleConstruct` translator, since it was replaced by transformation*
 - *Speed up `NodeContext` by reducing number of decoder calls (~70% speed up)*

### Tickets:
 - *None*


eaidova (approved):

src/frontends/pytorch/src/op/elu.cpp
```diff
@@ -12,9 +13,16 @@ namespace pytorch {
 namespace op {
 
 OutputVector translate_elu(NodeContext& context) {
+    // aten::elu(Tensor self, Scalar alpha=1, Scalar scale=1, Scalar input_scale=1) -> Tensor
+    num_inputs_check(context, 2, 4);
     auto x = context.get_input(0);
     auto alpha = context.const_input<float>(1);
-    return {context.mark_node(std::make_shared<opset10::Elu>(x, alpha))};
+    // TODO: Figure out what scale and input_scale do
```
rkazants, line 20: please create a ticket for this one.

rkazants (approved):

src/frontends/pytorch/src/op/elu.cpp
mvafin: Created 103682

PR #15661, opened 2023-02-12
Remove redundant reorder

byungilm:
+ Reorder 1d data
+ Reorder which only changes format

Signed-off-by: Min, Byungil <byungil.min@intel.com>

### Details:
 - Update general function 'are_layouts_identical'

### Tickets:
 - 82743

### DIFF
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
@@ -23,7 +23,6 @@ using namespace cldnn;
 #define LOG_NODE_REMOVAL(id)      GPU_DEBUG_LOG_PASS << "Remove node: " << (id) << std::endl;
 #define LOG_NODE_REPLACEMENT(id)  GPU_DEBUG_LOG_PASS << "Replace node: " << (id) << std::endl;
 
-
 remove_redundant_reorders::remove_redundant_reorders(layout_optimizer& lo_ref, bool enable_reorder_fusing, bool update_implementations,
     bool remove_output_reorders)
     : base_pass("remove_redundant_reorders"), lo(lo_ref), enable_reorder_fusing(enable_reorder_fusing), update_implementations(update_implementations),
--- a/src/plugins/intel_gpu/src/runtime/layout.cpp
+++ b/src/plugins/intel_gpu/src/runtime/layout.cpp
@@ -9,6 +9,7 @@
 #include <algorithm>
 
 namespace cldnn {
+static inline bool check_redundant_1d_along_feature(layout const& l1, layout const& l2);
 namespace {
 // pair.first tells whether l1 and l2 are absolutely identical
 // pair.second tells whether l1 and l2 can be reinterpreted to each other without need of reordering
@@ -26,10 +27,10 @@ std::pair<bool, bool> are_layouts_identical(layout const& l1, layout const& l2) 
 
     auto l1_size = l1.get_tensor();
     auto l2_size = l2.get_tensor();
-    int64_t offset_last_element_l1 = l1.get_linear_offset(l1_size - tensor{1});
-    int64_t offset_last_element_l2 = l2.get_linear_offset(l2_size - tensor{1});
     if (l1 == l2)
         return {true, true};
+    if (check_redundant_1d_along_feature(l1, l2))
+        return {false, true};
     if (l1.data_type != l2.data_type)
         return {false, false};
     // Reorders between bfyx, bfzyx, bfwzyx can pe reinterpeted as reshape when
@@ -72,13 +73,6 @@ std::pair<bool, bool> are_layouts_identical(layout const& l1, layout const& l2) 
         check_format(format::bs_fs_zyx_bsv32_fsv16) ||
         check_format(format::bs_fs_zyx_bsv32_fsv32))
         return {false, false};
-
-    // If data is actually 1d along f and dense, the layouts are identical
-    if (l1.data_type == l2.data_type && l1_size == l2_size && !l1_pad && !l2_pad && l1_size.batch[0] == 1 &&
-        ((l1.format.spatial_num() == 2 && l1_size.spatial[0] == 1 && l1_size.spatial[1] == 1) ||
-        ((l1.format.spatial_num() == 3 && l1_size.spatial[0] == 1 && l1_size.spatial[1] == 1 && l1_size.spatial[2] == 1))) &&
-        (offset_last_element_l1 + 1 == l1_size.feature[0] && offset_last_element_l2 + 1 == l2_size.feature[0]))
-        return {false, true};
 
     auto l1_pitch = l1.get_pitches();
     auto l2_pitch = l2.get_pitches();
@@ -623,6 +617,37 @@ ov::PartialShape layout::transform(cldnn::format new_fmt) const {
 
     ov::Shape new_shape(new_dims.begin(), new_dims.end());
     return ov::PartialShape(new_shape);
+}
+
+// Check a reorder is 1d along feature axis. Or feature size fits to inner block size of feature axis
+static inline bool check_redundant_1d_along_feature(layout const& l1, layout const& l2) {
+    // No padding, double blocked format and different data_type
+    if (!l1.data_padding && !l2.data_padding && !format::is_multi_blocked(l1.format) && !format::is_multi_blocked(l2.format) &&
+        l2.data_type == l1.data_type && l2.count() == l1.count()) {
+        auto l1_inner_blk = format::is_single_blocked(l1.format) ? format::traits(l1.format).block_sizes.at(0).second : 1;
+        auto l2_inner_blk = format::is_single_blocked(l2.format) ? format::traits(l2.format).block_sizes.at(0).second : 1;
+        auto max_inner_blk = std::max(l1_inner_blk, l2_inner_blk);
+        if (static_cast<size_t>(l2.feature()) == l1.count() && l2.feature() == l1.feature() &&
+           (l2.feature() % max_inner_blk == 0)) {
+            return true;
+        }
+
+        // Acceptable if a feature size of l2 'byxf' fits to l1's inner block size of 'b_fs_yx_fsv'
+        if ((l2.format == format::byxf && (l1.format == format::b_fs_yx_fsv16 ||  l1.format == format::b_fs_yx_fsv32) &&
+            l2.feature() == l1_inner_blk) ||
+            (l1.format == format::byxf && (l2.format == format::b_fs_yx_fsv16 ||  l2.format == format::b_fs_yx_fsv32) &&
+            l1.feature() == l2_inner_blk)) {
+            // each spatial axis should be same
+            for (size_t i = 0 ; i < l2.get_spatial_rank() ; i++) {
+                if (l2.spatial(i) != l1.spatial(i))
+                    return false;
+            }
+
+            return true;
+        }
+    }
+
+    return false;
 }
 
 }  // namespace cldnn
--- a/src/plugins/intel_gpu/tests/test_cases/reorder_gpu_test.cpp
+++ b/src/plugins/intel_gpu/tests/test_cases/reorder_gpu_test.cpp
@@ -2552,8 +2552,6 @@ INSTANTIATE_TEST_SUITE_P(DISABLED_REORDER,
                         tests::generic_test::custom_param_name_functor());
 
 
-
-
 struct reorder_test_param {
     tensor in_shape;
     tensor out_shape;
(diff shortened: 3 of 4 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
```diff
@@ -265,6 +265,35 @@ void remove_redundant_reorders::run(program& p) {
...
         auto i_layout = r_node.get_dependency(0).get_output_layout();
 
+        // Check redundant 1d reorder
+        if (i_layout.data_type == o_layout.data_type && i_layout.count() == o_layout.count()) {
+            // Check feature size of byxf input fits to output's inner block size
+            if (i_layout.format == format::byxf &&
+                ((o_layout.format == format::b_fs_yx_fsv16 && i_layout.feature() == 16) ||
+                (o_layout.format == format::b_fs_yx_fsv32 && i_layout.feature() == 32)) &&
+                i_layout.spatial(0) == o_layout.spatial(0) && i_layout.spatial(1) == o_layout.spatial(1) &&
+                i_layout.batch() == o_layout.batch()) {
```
sshlyapn, line 275: I think we need to check paddings here too

src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
```diff
@@ -265,6 +265,35 @@ void remove_redundant_reorders::run(program& p) {
...
+                r_node.can_be_optimized(true);
```
sshlyapn, line 276: Activations can be fused to Reorder primitive, so we need to check it before optimizing out reorder

src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
```diff
@@ -265,6 +265,35 @@ void remove_redundant_reorders::run(program& p) {
...
+            size_t size = 0;
+            for (auto dim : i_layout.get_dims()) {
+                if (dim == 1)
+                    cnt++;
+                else
+                    size = dim;
+            }
+
+            if ((i_layout.get_dims().size() == cnt + 1) && size == o_layout.count()) {
+                r_node.can_be_optimized(true);
```
sshlyapn, line 291: It looks like we can have some problems here in case of blocked/aligned dimension equal to 1. For example bfyx [b=1,f=1,y=1,x=32] reordering to b_fs_yx_fsv16 format
And also we need to check paddings and fused primitives here as well

src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
```diff
@@ -265,6 +265,35 @@ void remove_redundant_reorders::run(program& p) {
...
+                r_node.requires_reinterpret(true);
+                continue;
+            }
+        }
+
```
sshlyapn, line 296: Could you please add some tests for optimization?

src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
byungilm: Thanks. It's added.

src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
byungilm: Checking padding and fusing is added : 
if (!r_node.has_fused_primitives() && !i_layout.data_padding && !o_layout.data_padding &&

And modified condition that 1d axis is only feature axis : 
static_cast<size_t>(i_layout.feature()) == o_layout.count()

src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
byungilm: Thanks. added.

src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
byungilm: Added

src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
```diff
@@ -23,6 +23,38 @@ using namespace cldnn;
...
+                if (i_layout.spatial(i) != o_layout.spatial(i))
+                    return false;
+            }
+
+            return true;
+        }
+    }
+
+    return false;
+}
```
isanghao, line 57: could you explain why you decide to introduce a new function instead of existing function(`are_layouts_identical`)?

byungilm: It moved to are_layouts_identical

src/plugins/intel_gpu/src/graph/graph_optimizer/remove_redundant_reorders.cpp
```diff
@@ -305,7 +304,7 @@ void remove_redundant_reorders::run(program& p) {
             continue;
         }
 
-        if (!o_layout.compatible(i_layout))
+        if (!o_layout.compatible(i_layout) || r_node.has_fused_primitives())
```
isanghao, line 307: We already have this condition at line 272. Is it a different one?

src/plugins/intel_gpu/src/runtime/layout.cpp
```diff
@@ -623,4 +617,32 @@ ov::PartialShape layout::transform(cldnn::format new_fmt) const {
     return ov::PartialShape(new_shape);
 }
 
+// Check a reorder is 1d along feature axis. Or feature size fits to inner block size of feature axis
+static inline bool check_redundant_1d_along_feature(layout const& l1, layout const& l2) {
+    if (!l1.data_padding && !l1.data_padding && l2.data_type == l1.data_type && l2.count() == l1.count()) {
```
isanghao, line 622: `!l1.data_padding && !l1.data_padding` -> `!l1.data_padding && !l2.data_padding`

src/plugins/intel_gpu/src/runtime/layout.cpp
```diff
@@ -623,4 +617,32 @@ ov::PartialShape layout::transform(cldnn::format new_fmt) const {
...
+        // Get inner block size
```
isanghao, line 623: I think this comment is not expressing what this code is doing..

src/plugins/intel_gpu/src/runtime/layout.cpp
```diff
@@ -623,4 +617,32 @@ ov::PartialShape layout::transform(cldnn::format new_fmt) const {
     return ov::PartialShape(new_shape);
 }
 
+// Check a reorder is 1d along feature axis. Or feature size fits to inner block size of feature axis
+static inline bool check_redundant_1d_along_feature(layout const& l1, layout const& l2) {
```
isanghao, line 621: It is nice to make a separate function

src/plugins/intel_gpu/src/runtime/layout.cpp
```diff
@@ -623,4 +617,32 @@ ov::PartialShape layout::transform(cldnn::format new_fmt) const {
...
+    if (!l1.data_padding && !l1.data_padding && l2.data_type == l1.data_type && l2.count() == l1.count()) {
+        // Get inner block size
+        if (static_cast<size_t>(l2.feature()) == l1.count() && l2.feature() == l1.feature() &&
+            static_cast<size_t>(l1.feature()) == l2.count() &&
+            (format::traits(l1.format).block_sizes.size() == 1 || format::traits(l2.format).block_sizes.size() == 1)) {
+            return true;
+        }
```
isanghao, line 628: If the tensor size is [1,4,1,1] and input format is fsv4 and output format is fsv16, it is not compatible. But this code will return true...

src/plugins/intel_gpu/src/runtime/layout.cpp
```diff
@@ -623,4 +617,32 @@ ov::PartialShape layout::transform(cldnn::format new_fmt) const {
...
+
```
isanghao, line 629: What about bfyx [1,16,1,1] -> fsv16 [1,16,1,1]? Previous code returns true for such case, but new code will return false.

src/plugins/intel_gpu/tests/test_cases/reorder_gpu_test.cpp
```diff
@@ -2761,6 +2771,129 @@ INSTANTIATE_TEST_SUITE_P(reorder_gpu_testing, testing_removal_reorder,
...
+                                                                data_types::f16, format::byxf, data_types::f16, format::goiyx, data_types::f16, format::b_fs_yx_fsv16, true},
+                                            redundant_reorder_test_param{{1, 64, 8, 8}, {1, 64, 8, 8}, {1, 1, 1, 1}, {1, 1}, {0, 0},
+                                                                data_types::f16, format::byxf, data_types::f16, format::goiyx, data_types::f16, format::b_fs_yx_fsv16, false},
+                                            redundant_reorder_test_param{{1, 1, 1, 32}, {1, 1, 1, 32}, {1, 1, 1, 1}, {1, 1}, {0, 0},
+                                                                data_types::f16, format::byxf, data_types::f16, format::goiyx, data_types::f16, format::b_fs_yx_fsv16, false},
+                                            redundant_reorder_test_param{{1, 1, 1, 32}, {1, 1, 1, 32}, {1, 1, 1, 1}, {1, 1}, {0, 0},
+                                                                data_types::f16, format::b_fs_yx_fsv16, data_types::f16, format::goiyx, data_types::f16, format::b_fs_yx_fsv16, false},
+                                            }));
+
```
isanghao, line 2896: what about making testing `are_layouts_identical` directly, without topology?

src/plugins/intel_gpu/src/runtime/layout.cpp
```diff
@@ -623,4 +617,32 @@ ov::PartialShape layout::transform(cldnn::format new_fmt) const {
...
+        // Acceptable if a feature size of l2 'byxf' fits to l1's inner block size of 'b_fs_yx_fsv'
+        auto block_size = format::traits(l1.format).block_sizes;
+        if (l2.format == format::byxf && (l1.format == format::b_fs_yx_fsv16 ||  l1.format == format::b_fs_yx_fsv32) &&
```
isanghao, line 632: what about `fsv16 -> byxf`, not `byxf -> fsv16`? That is also compatible.

src/plugins/intel_gpu/src/runtime/layout.cpp
```diff
@@ -623,4 +617,32 @@ ov::PartialShape layout::transform(cldnn::format new_fmt) const {
...
+            block_size.size() == 1 && l2.feature() == block_size.at(0).second &&
+            l2.batch() == l1.batch()) {
```
isanghao, line 634: I think it is not necessary to check same batch size...

isanghao (changes requested):

src/plugins/intel_gpu/src/runtime/layout.cpp
byungilm: Discussion : need to consider input-output in are_layout_identical

byungilm: Solution : instead of checking input-output, check a feature size is multiple of bigger inner block.
i.e. fsv4 and fsv16 => multiple of 16 / fsv16 and fsv32 => multiple of 32

src/plugins/intel_gpu/src/runtime/layout.cpp
byungilm: Solution : added general function "format::is_multi_blocked" : use it

src/plugins/intel_gpu/tests/test_cases/reorder_gpu_test.cpp
byungilm: Solution : Make a ticket to make test-case for are_layouts_identical. This TC describes which case returns 'identical' or 'compatible'.

isanghao (approved):
LGTM, let me check the performance.

isanghao:
Performance was good. --this ../log/2023-03-20--01-47.report --ref ../log/2023-03-17--01-34.report

src/plugins/intel_gpu/src/runtime/layout.cpp
aalanwyr: Hi @byungilm，  Recently I find this patch cannot remove the reoder of this pattern:
input float16 [1,144,1,1] b_fs_yx_fsv32 --> reorder -->output float16 [1,144,1,1]  b_fs_yx_fsv16,  even the data layout is 1d along feature.  144 is not the multiple of inner block fsv32.  I try to remove this logical judgment “(l2.feature() % max_inner_blk == 0)”, but this reoder still cannot be removed.  Can we further fuse this reoder into the previous node ?

PR #15663, opened 2023-02-13
Support using bf16 in INT8 models

usstq:
### Details:
 - Removed `BF16QuantizeNodeFusing` to allow FQ to be fused into BF16 conv/fullyconnect node
 - Removed restrictions in `MatMul::canFuse` to allow FQ to be fused into Matmul
 - Fix the output data type deduction logic in conv/fullyconnect/matmul node
 - Fix oscale_mask in DnnlPostOpsComposer
 - Removed special rules on enableBF16 on quantized models

### oneDNN changes PR:
 - https://github.com/openvinotoolkit/oneDNN/pull/177

### Tickets:
 - CVS-53371


src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -342,12 +314,18 @@ void MatMul::getSupportedDescriptors() {
...
     }
 
-    if (!canBeExecutedInInt8(firstInPortPrec, secondInPortPrec) && one_of(outPortPrec, Precision::U8, Precision::I8))
-        outPortPrec = Precision::FP32; // INT output is not supported for non-INT inputs
+    if (canBeExecutedInInt8(firstInPortPrec, secondInPortPrec)) {
+        // INT8 mode support wide range of output precisions
+        outPortPrec = postOpsPrec;
+    } else {
+        // non-INT8 mode only support limited precision
+        // we ignore postOpsPrec and stay with getOriginalOutputPrecisionAtPort(0)
```
maxnick, line 327: What if the `postOpsPrec` is `fp32` and `outPortPrec` is `bf16`? In that case we will have unnecessary rounding, which may affect the accuracy.

src/plugins/intel_cpu/src/utils/debug_capabilities.cpp
```diff
@@ -506,11 +506,6 @@ std::ostream & operator<<(std::ostream & os, const PrintableModel& model) {
         os << std::endl;
 
         // recursively output subgraphs
```
maxnick, line 508: Could you please elaborate a little bit on the reason of removing the case `if (auto subgraph = std::dynamic_pointer_cast<ngraph::snippets::op::Subgraph>(op))`? Is `if (auto msubgraph = std::dynamic_pointer_cast<op::util::MultiSubGraphOp>(op))` enough for both the cases?

usstq: Yes, now `ngraph::snippets::op::Subgraph` is derived from `ov::op::util::SubGraphOp` which is derived from `MultiSubGraphOp`, so print `MultiSubGraphOp` is enough.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -229,29 +229,29 @@ void FullyConnected::getSupportedDescriptors() {
...
+    if (inputDataType == memory::data_type::f32) {
+        // oneDNN only support f32 output when input is f32, even if FQ is fused
         outputDataType = memory::data_type::f32;
-    }
-
-    if (inputDataType == memory::data_type::bf16
-        && one_of(outputDataType , memory::data_type::u8, memory::data_type::s8)) {
-        outputDataType = memory::data_type::bf16;
+    } else if (inputDataType == memory::data_type::bf16) {
+        // bf16 input only supports bf16/f32 output, even if FQ is fused as post-ops
```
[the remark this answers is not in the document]
usstq: from the performance perspective,  current solution may have little advantage over standalone FQ node(by the optimizations done in `DnnlPostOpsComposer` and `FakeQuantize::updateOptimizedFormula`).

According to @dmitry-gorokhov, the future plan is to support `int8` output in `bf16` oneDNN-primitives eventually, fusing FQ into bf16 primitive is the first step along the roadmap, the good news is due to widely supported binary-postops, we don't have to porting legacy postops to be able to fuse FQ node into `bf16` primitives (CVS-55903 CVS-55904).

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -503,10 +498,34 @@ void Convolution::getSupportedDescriptors() {
...
-        outputDataType = (getOriginalOutputPrecisionAtPort(0) == Precision::BF16
-                && !(isDepthWise() && ndims == 5)) ? memory::data_type::bf16 : memory::data_type::f32;
+        // FQ is allowed to be fused into non-INT8 parent Node, in such case it doesn't change the outputDataType
+
+        // in oneDNN, depthwise 3D convolution is only customized for f32, in bf16 it will fallbacks to
+        // normal convolution with much worse performance. thus we keep using f32 precision in that case.
+        // this WA can be removed after brdgmm_dw is enabled.
+        //    GroupConv([N, GROUPS * C_IN, D, H, W], [GROUPS, C_OUT=1, C_IN=1, kd, kh, kw])
+        // TODO when kd=1, and src layout is ndhwc, N & D can be combined so it can be done as a 2D depthwise
+        bool is_dw3DConv = (isDepthWise() && ndims == 5);
```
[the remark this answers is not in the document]
usstq: OK, will rename it to `isDW3DConv`

src/plugins/intel_cpu/src/nodes/matmul.cpp
usstq: your concern is valid, I was assuming `postOpsPrec` to be `int8` when wrote this logic, will change

usstq:
@maxnick I've fixed according to your review comment, please review again, Thx!

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1506,11 +1506,6 @@ bool Graph::InsertNode(NodePtr parent, NodePtr child, NodePtr node, int parentPo
 
 // Set all non const data paths precision to BF16
 void Graph::EnforceBF16() {
-    // Floating point parts of FP32 + INT8 or FP32 + BIN mixed precision models will be executed in BF16 precision
-    // only if enforceBF16 flag was set manually because current performance is not good enough to enable it by default
-    if (!implication(context->isGraphQuantized(), getConfig().manualEnforceBF16))
```
dmitry-gorokhov, line 1511: Seems like we don't need `manualEnforceBF16`  field at all inside the Config

usstq: Yes, manualEnforceBF16 is removed

PR #15668, opened 2023-02-13
[PT FE ] Add aten::index_put_ operator with layer test

sikorsl1:
### Details:
 - `aten::index_put_` operator has been added

### Tickets:
 - 101093

### Comments:
- Small fix in `src/frontends/pytorch/src/op/list_construct.cpp` has been added


src/frontends/pytorch/src/op/list_construct.cpp
```diff
@@ -24,7 +26,8 @@ OutputVector translate_list_construct(NodeContext& context) {
         if (c_node->get_shape().size() == 0) {
             c_node = std::make_shared<v0::Constant>(c_node->get_element_type(), Shape{1}, c_node->get_data_ptr());
         }
-        consts.push_back(c_node);
+        auto unsqueezed_c_node = context.mark_node(std::make_shared<v0::Unsqueeze>(c_node, const_0));
```
mvafin, line 29: Why is this needed? The code on lines 2-28 ensures all constants are at least 1D, and if you do `Unsqueeze`, you will make them 2D, which is incorrect in case when `prim::ListConstruct` is used for scalars.

src/frontends/pytorch/src/op/index_put_.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+namespace {
+Output<Node> generate_zeros_with_convertlike(NodeContext& context, Output<Node> sizes, Output<Node> tensor_of_type) {
+    auto const_0 = context.mark_node(v0::Constant::create(element::i32, Shape{}, {0}));
+    auto zeros = context.mark_node(std::make_shared<v3::Broadcast>(const_0, sizes));
+    return context.mark_node(std::make_shared<v1::ConvertLike>(zeros, tensor_of_type));
+}
+}  // namespace
+
+OutputVector translate_index_put_(NodeContext& context) {
```
mvafin, line 32: Please check number of inputs using `num_inputs_check` helper

src/frontends/pytorch/src/op/index_put_.cpp
```diff
@@ -0,0 +1,101 @@
...
+    auto input = context.get_input(0);
+    auto input_shape = context.mark_node(std::make_shared<v3::ShapeOf>(input, element::i32));
+    auto indices = context.get_input(1);
+    auto values = context.get_input(2);
+    auto accumulate = context.const_input<bool>(3);
+
+    auto indices_partial_shape = indices.get_partial_shape();
+    auto indices_first_dim = indices_partial_shape[0];
+    FRONT_END_OP_CONVERSION_CHECK(indices_first_dim.is_static(),
+                                  "We support only lists of tensors with v0::Constant number of elements.");
```
mvafin, line 48:
```suggestion
                                  "We support only lists of tensors with static number of elements.");
```

src/frontends/pytorch/src/op/index_put_.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    auto sub_data_shape =
+        context.mark_node(std::make_shared<v8::Slice>(input_shape, const_indices_list_len, const_max_int, const_1));
+    auto values_shape =
+        context.mark_node(std::make_shared<v0::Concat>(OutputVector{broadcast_index_shape, sub_data_shape}, 0));
+    values = context.mark_node(std::make_shared<v3::Broadcast>(values, values_shape));
+    values = context.mark_node(std::make_shared<v1::ConvertLike>(values, input));
+
+    std::shared_ptr<ov::Node> result;
```
mvafin, line 86:
```suggestion
    Output<Node> result;
```

src/frontends/pytorch/src/op/index_put_.cpp
```diff
@@ -0,0 +1,101 @@
...
+    if (accumulate) {
+        auto zeros = generate_zeros_with_convertlike(context, input_shape, input);
+        result = context.mark_node(std::make_shared<v3::ScatterNDUpdate>(zeros, index, values));
```
mvafin, line 89: Can `index` be negative?

mvafin (changes requested):

src/frontends/pytorch/src/op/list_construct.cpp
sikorsl1: I added this `Unsqueeze` operator, because after that there is a `Concat` operator. Without this `Unsqueeze` for inputs like [1,2],[3,4] we will obtain [1,2,3,4] which doesn't look like a list of tensors. I think that result [[1,2],[3,4]] (which is obtained by adding this `Unsqueeze`) would be much more appropriate. On the other hand, I didn't notice that there is a special case for scalars and I agree with your comment. Maybe it would be enough to put `Unsqueeze` after the `else` statement?

src/frontends/pytorch/src/op/index_put_.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+using namespace ov::op;
+
+namespace {
+Output<Node> generate_zeros_with_convertlike(NodeContext& context, Output<Node> sizes, Output<Node> tensor_of_type) {
```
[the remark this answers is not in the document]
sikorsl1: done

src/frontends/pytorch/src/op/index_put_.cpp
sikorsl1: done

src/frontends/pytorch/src/op/index_put_.cpp
sikorsl1: done

src/frontends/pytorch/src/op/index_put_.cpp
sikorsl1: done

src/frontends/pytorch/src/op/index_put_.cpp
sikorsl1: I general yes, but out `ScatterNDUpdate` from OV doesn't support this case.

sikorsl1: I added support for negative indices.

mvafin (approved):

mmikolajcz (approved):

PR #15673, opened 2023-02-13
[Pruning] Handle Reshape with nonconstant second input with -1 or 0 dims

mateusztabaka:
Ticket: 100573


daniil-lyakhov (approved):

AlexKoff88 (approved):

vuiseng9:
Thanks for the patch @mateusztabaka @daniil-lyakhov. I've built locally and the mo --transfrom=Pruning works for wav2vec2-JPQD IR, swin-base JPQD IR,  swin-base FP32 structured pruned IR which failed previously.

src/common/offline_transformations/src/pruning/shrink_weights.cpp
```diff
@@ -60,13 +59,25 @@ static bool maybe_adopt_reshape_node(std::shared_ptr<ov::Node> reshape, ngraph::
         return false;
     }
 
+    const auto new_shape = get_constant_from_source(shape)->cast_vector<int64_t>();
     auto sub_const_vector = std::vector<int64_t>();
```
itikhono, line 63: nitpick: you can pre-allocate the vector here for `mask->size()` size

src/common/offline_transformations/src/pruning/shrink_weights.cpp
```diff
@@ -41,15 +42,13 @@ static bool is_static_reshape_op(std::shared_ptr<ov::Node> node) {
...
+    const auto& output_shape = output_shape_const_op->cast_vector<int64_t>();
     // below casts are needed due to VC warning C4244, literals are not enough in this case
     const int64_t input_elems =
         std::accumulate(input_shape.begin(), input_shape.end(), static_cast<int64_t>(1), std::multiplies<int64_t>());
     const auto output_elems =
         std::accumulate(output_shape.begin(), output_shape.end(), static_cast<int64_t>(1), std::multiplies<int64_t>());
-    if (output_elems <= 0 || input_elems == output_elems)
-        return false;
-    return true;
+    return input_elems != output_elems;
```
itikhono, line 51: I guess if output_shape contains only one `-1` is also static case, we can calcualate the actual value based on static input shape and other values in the output_shape (if they are positive)

itikhono (changes requested):

src/common/offline_transformations/src/pruning/shrink_weights.cpp
```diff
@@ -60,13 +59,25 @@ static bool maybe_adopt_reshape_node(std::shared_ptr<ov::Node> reshape, ngraph::
         return false;
     }
 
+    const auto new_shape = get_constant_from_source(shape)->cast_vector<int64_t>();
```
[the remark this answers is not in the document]
mateusztabaka: it's already checked in is_static_reshape_op, but we can have a check here as well.

src/common/offline_transformations/src/pruning/shrink_weights.cpp
mateusztabaka: done

src/common/offline_transformations/src/pruning/shrink_weights.cpp
mateusztabaka: if output_shape contains only one -1, this function will return true.

itikhono (approved):

PR #15675, opened 2023-02-13
[TF FE] Fix FusedBatchNormV3 in case of mean and variance empty tensors

rkazants:
**Details:** Fix FusedBatchNormV3 in case of mean and variance empty tensors

**Ticket:** 99675

### DIFF
--- a/src/frontends/tensorflow_common/src/op/fused_batch_norm.cpp
+++ b/src/frontends/tensorflow_common/src/op/fused_batch_norm.cpp
@@ -104,18 +104,28 @@ void compute_weighted_batch_mean_and_variance(const Output<Node>& x,
     // for weighted_variance it is similar
     // (1 - exponential_avg_factor) * variance + exponential_avg_factor * batch_variance,
     // where batch_variance is the variance of the current batch in x.
-
-    // compute weighted_batch_mean
     auto const_one = make_shared<Constant>(exp_avg_factor_const.get_element_type(), Shape{}, 1);
     auto one_minus_exp_avg_factor = make_shared<Subtract>(const_one, exp_avg_factor_const);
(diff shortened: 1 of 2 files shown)


mvafin (approved):

gkrivor (approved):

src/frontends/tensorflow_common/src/op/fused_batch_norm.cpp
```diff
@@ -104,18 +104,28 @@ void compute_weighted_batch_mean_and_variance(const Output<Node>& x,
...
-    // compute weighted_batch_mean
     auto const_one = make_shared<Constant>(exp_avg_factor_const.get_element_type(), Shape{}, 1);
     auto one_minus_exp_avg_factor = make_shared<Subtract>(const_one, exp_avg_factor_const);
-    auto bt_mean_by_exp_avg = make_shared<Multiply>(batch_mean, exp_avg_factor_const);
-    weighted_batch_mean = make_shared<Multiply>(mean, one_minus_exp_avg_factor)->output(0);
-    weighted_batch_mean = make_shared<Add>(bt_mean_by_exp_avg, weighted_batch_mean);
+
+    // compute weighted_batch_mean
+    // no need to weight in case of empty tensor mean
+    if (mean.get_partial_shape().is_static() && shape_size(mean.get_shape()) > 0) {
```
itikhono, line 112: I believe we need some shapeof expression to handle this case.
If(NotEqual(CumSum(ShapeOf(mean))))

rkazants: It will not help because both branches of `If` operation will be shape inferred and one of the branch will complain on inconsistence of `Multiply` on empty operand and non-empty operand.

PR #15676, opened 2023-02-13
[PT FE] Remove warnings suppression in frontend code

mvafin:
### Details:
 - *Remove warnings suppression in frontend code*

### Tickets:
 - *103431*

### DIFF
--- a/src/frontends/pytorch/src/transforms/append_list_unpack_replacer.cpp
+++ b/src/frontends/pytorch/src/transforms/append_list_unpack_replacer.cpp
@@ -62,11 +62,12 @@ AppendListUnpackReplacer::AppendListUnpackReplacer() {
             auto getitem_index_ptr = getitem_node->input_value(1).get_node_shared_ptr();
             auto getitem_index_const = std::dynamic_pointer_cast<ov::op::v0::Constant>(getitem_index_ptr);
             auto index_val = getitem_index_const->cast_vector<int64_t>();
-            auto index = 0;
-            if (index_val[0] >= 0) {
-                index = index_val[0];
-            } else {
-                index = inputs.size() + index_val[0];
+            if (index_val.size() != 1) {
+                return false;
+            }
+            auto index = index_val[0];
+            if (index_val[0] < 0) {
+                index = inputs.size() + index;
             }
             auto axis_0 = ov::op::v0::Constant::create(element::i64, Shape{}, {0});
             auto split = std::make_shared<ov::op::v1::Split>(inputs[index], axis_0, list_unpack->get_output_size());
--- a/src/frontends/pytorch/src/transforms/aten_getitem_replacer.cpp
+++ b/src/frontends/pytorch/src/transforms/aten_getitem_replacer.cpp
@@ -93,15 +93,16 @@ AtenGetItemReplacer::AtenGetItemReplacer() {
             } else {
                 auto getitem_index_ptr = getitem->input_value(1).get_node_shared_ptr();
                 auto getitem_index_const = std::dynamic_pointer_cast<ov::op::v0::Constant>(getitem_index_ptr);
-                auto index_val = getitem_index_const->cast_vector<int64_t>();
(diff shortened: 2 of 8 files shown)


eaidova (approved):

src/frontends/pytorch/src/transforms/append_list_unpack_replacer.cpp
```diff
@@ -62,11 +62,12 @@ AppendListUnpackReplacer::AppendListUnpackReplacer() {
...
-            if (index_val[0] >= 0) {
-                index = index_val[0];
-            } else {
-                index = inputs.size() + index_val[0];
+            if (index_val.size() != 1) {
+                return false;
+            }
+            auto index = index_val[0];
+            if (index_val[0] < 0) {
+                index = inputs.size() + index;
```
rkazants, line 70: here we try to sum-up `size_t` and `int64_t` element. Please check

src/frontends/pytorch/src/transforms/aten_getitem_replacer.cpp
```diff
@@ -93,15 +93,16 @@ AtenGetItemReplacer::AtenGetItemReplacer() {
...
-                    index = index_val[0];
-                } else {
-                    index = split->outputs().size() + index_val[0];
+                auto index_val = getitem_index_const->cast_vector<int64_t>();
+                if (index_val.size() != 1) {
+                    return false;
+                }
+                auto index = index_val[0];
+                if (index < 0) {
+                    index = split->outputs().size() + index;
```
rkazants, line 105: Please also check

src/frontends/pytorch/src/utils.cpp
```diff
@@ -232,7 +232,7 @@ OutputVector make_framework_node(NodeContext* context) {
                                           context->get_output_size() - num_body_outs + num_skip_body_outputs);
     fw_node->set_friendly_name(context->get_op_type());
     for (size_t i = 0; i < bodies.size(); ++i) {
-        fw_node->set_function(i, bodies[i]);
+        fw_node->set_function(static_cast<int>(i), bodies[i]);
```
rkazants, line 235: may be it is a problem of API `set_function`? Why do we use `int` type for body index. `size_t` is better.

src/frontends/pytorch/src/utils.hpp
```diff
@@ -22,7 +22,7 @@ void num_inputs_check(const NodeContext& context, size_t min_inputs, size_t max_
 
 Output<Node> make_optional_bias(const Output<Node>& base_op,
                                 const NodeContext& context,
-                                size_t bias_input_idx,
+                                int bias_input_idx,
```
rkazants, line 25: `size_t` is more logical.

rkazants:
@ilya-lavrenov, please take a look at this PR

src/frontends/pytorch/src/utils.hpp
mvafin: It is used for API that uses `int` as input, so same type makes sense.

src/frontends/pytorch/src/transforms/append_list_unpack_replacer.cpp
mvafin: If compiler doesn't output warnings I think it is cleaner to not use extra casts, @ilya-lavrenov what do you think?

src/frontends/pytorch/src/utils.cpp
rkazants: if it is not FE code, I am fine then. I thought this is PT FE FW Node API:)

src/frontends/pytorch/src/transforms/append_list_unpack_replacer.cpp
rkazants: What is fine for one compiler can be "death" for another one:)

ilya-lavrenov (approved):

PR #15684, opened 2023-02-13
Allow dynamic data type in NMS

tomdol:
### Details:
 - Allow dynamic data type in NMS shape inference
 - This unblocks one of the steps in the import of Detectron2 MaskRCNN model
 - The data type can later become static and be subject to the full shape/type inference (later == when more data about the whole graph is known)


src/core/src/op/non_max_suppression.cpp
```diff
@@ -591,7 +591,7 @@ constexpr size_t score_threshold_port = 4;
 constexpr size_t soft_nms_sigma_port = 5;
 
 inline bool is_float_type_admissible(const element::Type& t) {
-    return t == element::f32 || t == element::f16 || t == element::bf16;
+    return t == element::dynamic || t == element::f32 || t == element::f16 || t == element::bf16;
```
praasz, line 594: Add/update some tests in  [type_prop tests](https://github.com/openvinotoolkit/openvino/blob/c8afb902180b403fbaff404e6417f9df068c6690/src/core/tests/type_prop/non_max_suppression.cpp) to create op with dynamic type on inputs which support it.

tomdol: I've covered it with a single test. Both inputs use the same validator for the data type.

praasz (approved):

PR #15685, opened 2023-02-13
[GPU] Refactoring weights reorder

tuxedcat:
### Status:
 - Many legacy codes making accuracy problems. (e.g os_is_yx_isv16_osv16, os_iyx_osv16, SUB_GROUP_SIZE, ...)
 - Too many copy & paste

### Details:
 - Make general calculation function and replace legacy codes

### Tickets:
 - *ticket-id*


tuxedcat:
Accuracy failed. Maybe it should consider padding info?

isanghao:
This is necessary change but we cannot afford this change at this moment :(

PR #15689, opened 2023-02-13
[CONFORMANCE] Add Shape mode to the conformance runner

iefode:


src/tests/ie_test_utils/functional_test_utils/layer_tests_summary/run_conformance.py
```diff
@@ -49,6 +49,7 @@ def parse_arguments():
     gtest_filter_helper = "Specify gtest filter to apply when running test. E.g. *Add*:*BinaryConv*. The default value is None"
     ov_config_path_helper = "Specify path to file contains plugin config"
     dump_conformance_help = "Set '1' if you want to create Conformance IRs from custom/downloaded models. In other cases, set 0. The default value is '1'"
+    shape_mode_help = "Specify shape mode for conformance. Default value is `Both`. Possible values: `static`, `dynamic`, ``"
```
ArtemySkrebkov, line 52: Probably, the default value is `` (empty) which means that we run both configurations.

ArtemySkrebkov (approved):

src/tests/ie_test_utils/functional_test_utils/layer_tests_summary/run_conformance.py
iefode: yes

PR #15695, opened 2023-02-14
optimize the shape infer for transpose

xczhai:
### Details:
 - *simplify the shape infer flow for transpose*

### Tickets:
 - *100508*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/transpose.cpp
+++ b/src/plugins/intel_cpu/src/nodes/transpose.cpp
@@ -39,17 +39,21 @@ bool Transpose::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, 
 
 class TransposeShapeInfer : public ShapeInferEmptyPads {
 public:
-    TransposeShapeInfer(const size_t& out_rank, const std::vector<size_t>& axes_vec) : m_out_rank(out_rank), m_axes_vec(axes_vec) {
-        m_outputShape = VectorDims(out_rank, 1);
-        m_needReverse = m_axes_vec.empty();
-    }
+    TransposeShapeInfer(const size_t& out_rank, const std::vector<size_t>& axes_vec)
+    : m_out_rank(out_rank), m_axes_vec(axes_vec), m_outputShape(std::move(VectorDims(out_rank, 1))), m_needReverse(axes_vec.empty()) {}
 
     std::vector<VectorDims> infer(
         const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
         const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
         const VectorDims& shapeIn = input_shapes[0].get();
-        for (auto i = 0; i < m_out_rank; ++i) {
-            m_outputShape[i] = m_needReverse ? shapeIn[m_out_rank - 1 - i] : shapeIn[m_axes_vec[i]];
+        if (m_needReverse) {
+            for (auto i = 0; i < m_out_rank; ++i) {
+                m_outputShape[i] = shapeIn[m_out_rank - 1 - i];
+            }
+        } else {
+            for (auto i = 0; i < m_out_rank; ++i) {
+                m_outputShape[i] = shapeIn[m_axes_vec[i]];
+            }
         }
         return {m_outputShape};
     }
@@ -62,7 +66,7 @@ private:
     const size_t m_out_rank;
     const std::vector<size_t> m_axes_vec;
     VectorDims m_outputShape;
-    bool m_needReverse;
+    const bool m_needReverse;
 };
 
 class TransposeShapeInferFactory : public ShapeInferFactory {
@@ -78,7 +82,7 @@ public:
     }
 
 private:
-    const std::shared_ptr<ngraph::Node> m_op;
+    const std::shared_ptr<ov::Node> m_op;
 };
 
 Transpose::Transpose(const std::shared_ptr<ov::Node>& op, const GraphContext::CPtr context)
--- a/src/plugins/intel_cpu/src/nodes/transpose.cpp
+++ b/src/plugins/intel_cpu/src/nodes/transpose.cpp
@@ -37,6 +37,20 @@ bool Transpose::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, 
     return true;
 }
 
+class TransposeDynShapeInfer : public ShapeInferEmptyPads {
+public:
+    TransposeDynShapeInfer() = default;
+    std::vector<VectorDims> infer(
+        const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
+        const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
+        IE_THROW(NotImplemented) << "TODO: Support parameterized Order input for dynamic shapes.";
+    }
+    port_mask_t get_port_mask() const override {
+        return EMPTY_PORT_MASK;
+    }
+private:
+};
+
 class TransposeShapeInfer : public ShapeInferEmptyPads {
 public:
     TransposeShapeInfer(const size_t& out_rank, const std::vector<size_t>& axes_vec)
@@ -77,7 +91,7 @@ public:
             const auto axes_vec = order->cast_vector<size_t>();
             return std::make_shared<TransposeShapeInfer>(m_op->get_output_partial_shape(0).rank().get_length(), axes_vec);
         } else {
-            IE_THROW() << "The second input is NOT a constant node, which is unexpected in the Transpose shape inference factory";
+            return std::make_shared<TransposeDynShapeInfer>();
         }
     }
 
--- a/src/plugins/intel_cpu/src/nodes/transpose.cpp
+++ b/src/plugins/intel_cpu/src/nodes/transpose.cpp
@@ -37,8 +37,52 @@ bool Transpose::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, 
     return true;
 }
 
+class TransposeShapeInfer : public ShapeInferEmptyPads {
+public:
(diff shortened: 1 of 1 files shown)


src/plugins/intel_cpu/src/nodes/transpose.cpp
```diff
@@ -37,8 +37,52 @@ bool Transpose::isSupportedOperation(const std::shared_ptr<const ov::Node>& op,
...
+        if (const auto order = ov::as_type_ptr<const ov::op::v0::Constant>(m_op->get_input_node_shared_ptr(ov::op::v1::Transpose::ORDER))) {
+            const auto axes_vec = order->cast_vector<size_t>();
+            return std::make_shared<TransposeShapeInfer>(m_op->get_output_partial_shape(0).rank().get_length(), axes_vec);
+        } else {
+            IE_THROW() << "Unexpected operation type in the Transpose shape inference factory";
+        }
+    }
+
+private:
+    const std::shared_ptr<ngraph::Node> m_op;
```
praasz, line 81:
```suggestion
    const std::shared_ptr<ov::Node> m_op;
```
or without namespace as this class is inside `ov` namespace?

src/plugins/intel_cpu/src/nodes/transpose.cpp
```diff
@@ -37,8 +37,52 @@ bool Transpose::isSupportedOperation(const std::shared_ptr<const ov::Node>& op,
     return true;
 }
 
+class TransposeShapeInfer : public ShapeInferEmptyPads {
+public:
+    TransposeShapeInfer(const size_t& out_rank, const std::vector<size_t>& axes_vec) : m_out_rank(out_rank), m_axes_vec(axes_vec) {
+        m_outputShape = VectorDims(out_rank, 1);
```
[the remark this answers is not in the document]
xczhai: > Is possible to move these members initialization to `ctor` initialization list? To avoid default initialization and then copy.

Yes. It's a good idea. Have done it.

src/plugins/intel_cpu/src/nodes/transpose.cpp
```diff
@@ -37,8 +37,52 @@ bool Transpose::isSupportedOperation(const std::shared_ptr<const ov::Node>& op,
...
+        m_needReverse = m_axes_vec.empty();
+    }
+
+    std::vector<VectorDims> infer(
+        const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
+        const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
+        const VectorDims& shapeIn = input_shapes[0].get();
+        for (auto i = 0; i < m_out_rank; ++i) {
+            m_outputShape[i] = m_needReverse ? shapeIn[m_out_rank - 1 - i] : shapeIn[m_axes_vec[i]];
```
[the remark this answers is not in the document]
xczhai: > Consider re-factor to something like (to avoid check bool flag in every iteration):
> 
> ```c++
> if (m_needReverse){
>     // do reverse copy
> } else {
>     // do normal copy.
> }
> ```

Okay. Have re-factor it.

src/plugins/intel_cpu/src/nodes/transpose.cpp
xczhai: > 

Yes. Fix it.

src/plugins/intel_cpu/src/nodes/transpose.cpp
```diff
@@ -37,8 +37,56 @@ bool Transpose::isSupportedOperation(const std::shared_ptr<const ov::Node>& op,
...
+
+class TransposeShapeInferFactory : public ShapeInferFactory {
+public:
+    TransposeShapeInferFactory(const std::shared_ptr<ov::Node>& op) : m_op(op) {}
+    ShapeInferPtr makeShapeInfer() const override {
+        if (const auto order = ov::as_type_ptr<const ov::op::v0::Constant>(m_op->get_input_node_shared_ptr(ov::op::v1::Transpose::ORDER))) {
+            const auto axes_vec = order->cast_vector<size_t>();
+            return std::make_shared<TransposeShapeInfer>(m_op->get_output_partial_shape(0).rank().get_length(), axes_vec);
+        } else {
+            IE_THROW() << "Unexpected operation type in the Transpose shape inference factory";
```
maxnick, line 80: I think the error message should be about the fact that the input is not constant.

src/plugins/intel_cpu/src/nodes/transpose.cpp
```diff
@@ -37,8 +37,56 @@ bool Transpose::isSupportedOperation(const std::shared_ptr<const ov::Node>& op,
     return true;
 }
 
+class TransposeShapeInfer : public ShapeInferEmptyPads {
+public:
+    TransposeShapeInfer(const size_t& out_rank, const std::vector<size_t>& axes_vec)
+    : m_out_rank(out_rank), m_axes_vec(axes_vec), m_outputShape(std::move(VectorDims(out_rank, 1))), m_needReverse(axes_vec.empty()) {}
```
maxnick, line 43: std::move here is not necessary since `VectorDims(out_rank, 1)` is already an rvalue. Moreover you can directly call the constructor `m_outputShape(out_rank, 1)`,

src/plugins/intel_cpu/src/nodes/transpose.cpp
xczhai: > I think the error message should be about the fact that the input is not constant.

Yes, update the error info.

src/plugins/intel_cpu/src/nodes/transpose.cpp
xczhai: > 

correct this bug.

maxnick (approved):

maxnick:
@praasz , do you have any additional comments?

praasz (approved):

src/plugins/intel_cpu/src/nodes/transpose.cpp
xczhai: @maxnick @praasz 
Hi all, 
I have something to discuss here.

- The lastest pytorch test hits this bug in the PR https://github.com/openvinotoolkit/openvino/pull/15614. To be honest, former tests don't verify this unusual case because the CI  pass all tests before 15614 merged.
- In the CPU plugin, The second input is always a constant node, which is expected.
- But if the second input is not a constant node, the framework should NOT stop here, right. As we can see https://github.com/openvinotoolkit/openvino/blob/310c4deab9f3428a95347059f144563a1aea2c22/src/plugins/intel_cpu/src/nodes/transpose.cpp#L21, the operation will fall back to the ngraph implementaion to keep the model going.
- As a result, I remove such an **_exception_** here.
looking forward to your comments. thx.

praasz: If you mean `ngraph` fallback implementation is template implementation for transpose shape inference then:

1. If shape infer on `ov::PartialShape` the result will some dynamic shape.
2. If working with `StaticShapes` then if order is not constant (parameter) and data are passed in tensor map the inference produce output shape, but if 2nd input is Parameter and no data pass it will try create dynamic shape which will trigger to throw exception

So if 2nd input is not constant but there are data which can be pass in tensor map then it should be fine from shape inference perspective, otherwise it will produce dynamic output shape.

src/plugins/intel_cpu/src/nodes/transpose.cpp
```diff
@@ -76,9 +77,8 @@ class TransposeShapeInferFactory : public ShapeInferFactory {
         if (const auto order = ov::as_type_ptr<const ov::op::v0::Constant>(m_op->get_input_node_shared_ptr(ov::op::v1::Transpose::ORDER))) {
             const auto axes_vec = order->cast_vector<size_t>();
             return std::make_shared<TransposeShapeInfer>(m_op->get_output_partial_shape(0).rank().get_length(), axes_vec);
-        } else {
-            IE_THROW() << "The second input is NOT a constant node, which is unexpected in the Transpose shape inference factory";
         }
+        return std::make_shared<TransposeShapeInfer>();
```
maxnick, line 81: The main problem is that the node instantiation pipeline try to instantiate a node of a specific type even though the specific configuration is not supported by the plugin (isSupportedOperation returns false). That means that the base `Node` class constructor is always called, and the factory also called, notwithstanding the fact that the node object will not be created due to exception in the specific type constructor. Thus, we have two options here:
1. The factory may return NgraphShapeInfer object in case the specific configuration is not supported. That is we should store the ngraph op in the factory and in the else branch instead of throwing exception we should instantiate NgraphShapeInfer factory for this op with corresponding port mask and call `makeShapeInfer()` to that factory object. That will automatically make the support for a non constant `ORDER` input working, however the `ngraph` implementation is suboptimal.
2. We can make yet another class `TransposeDynInputShapeInfer` which should support non constant data on the `ORDER` input, but its implementation will throw `not implemented` exception since at the current moment of time the CPU plugin `Transpose` node does not support such configuration and when someone removes this constraint he will be forced to update the specific shape inference implementation as well. So the factory will simply return an instance of that class on the else branch.
Personally I prefer the second option, since it is not silently work with bad performance.

praasz (approved):

src/plugins/intel_cpu/src/nodes/transpose.cpp
xczhai: The second option sounds good and is friendly.
I update a TransposeDynShapeInfer class for the UnConstant Order. In such class, the constructor is empty and the infer function throws *NotImplemented* info.

src/plugins/intel_cpu/src/nodes/transpose.cpp
xczhai: Thanks for your comments, which helps understand the flow.

PR #15701, opened 2023-02-14
[GPU] Fix a bug of permute optimization

jade-cho:
For int8 models, if there is FakeQuantize between permute and convolution, an operation like data type casting could be fused to permute. In this case, do not optimize permute.

### Tickets:
 - *103408*
 - *103386*
 - *103378*

### DIFF
--- a/src/plugins/intel_gpu/tests/fusions/convolution_fusion_test.cpp
+++ b/src/plugins/intel_gpu/tests/fusions/convolution_fusion_test.cpp
@@ -4171,7 +4171,7 @@ INSTANTIATE_TEST_SUITE_P(implicit_crop_concat_conv_fusings_gpu, implicit_crop_co
 
 class PermuteOptimizingTestOnednn : public BaseFusingTest<convolution_test_params> {
 public:
-    void execute(convolution_test_params& p, const std::string& expected_kernel = "undef") {
+    void execute(convolution_test_params& p, bool is_permute_optimized = true) {
         if (!engine.get_device_info().supports_immad)
             return;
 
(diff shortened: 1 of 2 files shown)


sshlyapn (commented):
Looks good to me
@jade-cho , could you please add some test for such cases?

jade-cho:
@sshlyapn I will add a unit test after merge.

src/plugins/intel_gpu/tests/fusions/convolution_fusion_test.cpp
```diff
@@ -4172,6 +4172,45 @@ INSTANTIATE_TEST_SUITE_P(fusings_gpu, conv_after_permute_optimizing, ::testing::
...
+        permute("permute", input_info("input"), {0, 3, 1, 2}),
+        quantize("quantize1", input_info("permute"), input_info("in_lo1"), input_info("in_hi1"),
+                 input_info("out_lo1"), input_info("out_hi1"), 256, data_types::i8),
+        convolution("conv_prim", input_info("quantize1"), { "weights" }, { "bias" }, p.groups, p.stride, p.pad, p.dilation),
+        activation("activation", input_info("conv_prim"), activation_func::abs),
+        reorder("reorder_bfyx", input_info("activation"), p.default_format, data_types::f32)
+    );
+
+    tolerance = default_tolerance(p.default_type);
+    execute(p, "permute_ref__f16");
```
isanghao, line 4207: I think permute primitive name might change. What about just checking the primitive count or primitive type(if we have such information)?

jade-cho: It is not possible to get number of unexecuted primitives. I modified to use boolean value instead of string.

isanghao (approved):

PR #15705, opened 2023-02-14
[GPU] Add mem_lock_type of get_sliced_mem() in loop

wilson-seok:
### Details:
 - Add mem_lock_type of get_sliced_mem() in loop

### Tickets:
 - 103594


wilson-seok:
No exec_graph diff from dryrun on fp32/fp16/int8 models.

yeonbok (approved):

PR #15711, opened 2023-02-14
Review space to and shuffle channels operators for shape inference aspects

praasz:
### Details:
Review operators SpaceToBatch, SpaceToDepth, ShuffleChannels, DepthToSpace, BatchToSpace for:
 - Check interval shape and label propagation
 - Check/update template implementation of `shape_infer`.
 - Check shape inference when operators created by default `ctor`.
 - Update tests

### Tickets:
 - 97257
 - 97281
 - 97330
 - 97379


src/core/shape_inference/include/space_to_batch_shape_inference.hpp
```diff
@@ -15,75 +15,72 @@ namespace ov {
...
+        pads_end_shape);
 
     NODE_VALIDATION_CHECK(op,
                           inputs_same_ps.rank().compatible(1),
                           "block_shape and pads inputs must have rank 1. Got: ",
                           inputs_same_ps.rank());
 
-    if (data_rank.is_static()) {
+    if (data_shape.rank().is_static()) {
+        static constexpr size_t spatial_dim_offset = 1;
```
t-jankowski, line 47:
```suggestion
        constexpr size_t spatial_dim_offset = 1;
```

src/core/shape_inference/include/space_to_batch_shape_inference.hpp
```diff
@@ -15,75 +15,72 @@ namespace ov {
...
+        " and ",
+        pads_end_shape);
 
     NODE_VALIDATION_CHECK(op,
                           inputs_same_ps.rank().compatible(1),
                           "block_shape and pads inputs must have rank 1. Got: ",
                           inputs_same_ps.rank());
 
-    if (data_rank.is_static()) {
+    if (data_shape.rank().is_static()) {
```
t-jankowski, line 46: Does it work with `ov::Shape`?

src/core/shape_inference/include/space_to_batch_shape_inference.hpp
```diff
@@ -15,75 +15,72 @@ namespace ov {
...
-            get_data_as_int64<T>(3, op, pads_end_val, constant_data)) {
-            got_const_data = true;
-            int64_t block_prod =
-                std::accumulate(begin(block_val), end(block_val), int64_t(1), std::multiplies<int64_t>());
+        auto out_shape = data_shape;
+        std::vector<int64_t> block, pads_begin, pads_end;
+        if (get_data_as_int64<TShape>(1, op, block, constant_data) &&
+            get_data_as_int64<TShape>(2, op, pads_begin, constant_data) &&
+            get_data_as_int64<TShape>(3, op, pads_end, constant_data)) {
+            TVal block_prod = std::accumulate(begin(block), end(block), 1, std::multiplies<int64_t>());
```
t-jankowski, line 59: If `signed` is not considered identical to `int64_t` by some compilator it might break compilation. If guess VC doesn't like that.

src/core/shape_inference/include/space_to_batch_shape_inference.hpp
praasz: The shape_infer is used tested with StaticShape and PartialShape. Right no these shape_infer functions are not used with ov::Shape.
I think will not work for ov::Shape as this class interface is not compatible with StaticShape and PartialShape it will required some specialization or make this class compatible. There is more member functions which not exist in ov::Shape.

src/core/shape_inference/include/space_to_batch_shape_inference.hpp
praasz: It looks like there is no complain from compilers on CI, so leave it. Will add cast if required.

src/core/shape_inference/include/batch_to_space_shape_inference.hpp
```diff
@@ -45,10 +45,11 @@ void shape_infer(const ov::op::v1::BatchToSpace* op,
 
     const ov::Rank data_rank = data_shape.rank();
     if (data_rank.is_static()) {
+        static constexpr size_t spatial_dim_offset = 1;
```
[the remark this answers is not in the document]
praasz: Change as there is no benefit from static it will be optimized by compiler.

src/core/tests/type_prop/shuffle_channels.cpp
```diff
@@ -30,14 +32,16 @@ TEST(type_prop, shuffle_channels_basic_4D) {
...
+    auto data_input_shape = PartialShape{Dimension::dynamic(), Dimension(3, 9), 4, Dimension(4, 15)};
+    set_shape_labels(data_input_shape, 10);
     const auto data = make_shared<op::Parameter>(element::f32, data_input_shape);
     const auto axis = 1;
     const auto group = 3;
     const auto shuffle_channels = make_shared<op::v0::ShuffleChannels>(data, axis, group);
 
     EXPECT_EQ(shuffle_channels->get_element_type(), element::f32);
     EXPECT_EQ(shuffle_channels->get_output_partial_shape(0), data_input_shape);
+    EXPECT_THAT(get_shape_labels(shuffle_channels->get_output_partial_shape(0)), ElementsAre(10, 11, 12, 13));
```
jane-intel, line 44: Actually, the data is being reshuffled, so channel dimension must have no label. please clear it in the shape inference. Please check for other operations too

jane-intel (commented):
All good! I have only one concern.

src/core/tests/type_prop/shuffle_channels.cpp
praasz: Changed.

jane-intel (approved):

PR #15714, opened 2023-02-14
[PT FE] Add prim::PythonOp

mmikolajcz:
### Details:
 - *Add handler to allow loading prim::PythonOp subgraph into OV model*

### Tickets:
 - *101096*


tests/layer_tests/pytorch_tests/test_pythonop.py
```diff
@@ -0,0 +1,61 @@
...
+
+        class _ExpF(Function):
+            @staticmethod
+            def forward(ctx, input_tensor):
+                exp = torch.exp(input_tensor)
+                ctx.save_for_backward(exp)
+                return exp
+
+            @staticmethod
+            def backward(ctx, output_grad):
```
mvafin, line 26: why do we need `backward` here?

tests/layer_tests/pytorch_tests/test_pythonop.py
```diff
@@ -0,0 +1,61 @@
...
+                (result,) = ctx.saved_tensors
+                return output_grad * result
+
+        exp_f = _ExpF.apply
+
+        class prim_pythonop(torch.nn.Module):
+            def forward(self, input_tensor):
+                return torch.add(exp_f(input_tensor), 1)
```
mvafin, line 34: Could you also add test for DeformableConvolution that we have in models?

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -185,7 +186,10 @@ def get_output_transpose_order(self, index: int) -> list:
         return []
 
     def get_subgraph_size(self) -> int:
-        return len(self.get_subgraphs()) if hasattr(self.graph_element, "blocks") else 1
+        try:
+            return len(self.get_subgraphs())
```
mvafin, line 190: I don't like using try-except here, could you just change the if statement to avoid this?

tests/layer_tests/pytorch_tests/test_pythonop.py
mmikolajcz: We don't need it, removed, thanks

tests/layer_tests/pytorch_tests/test_pythonop.py
mmikolajcz: This op would add `torchvision::deform_conv2d` into unconverted OP, adding test would require to enable this op as well. I think it might be better to enable it in different PR to not mix it up, but if you want, I could enable this in current PR

src/frontends/pytorch/src/op/pythonop.cpp
```diff
@@ -0,0 +1,51 @@
...
+namespace op {
+
+OutputVector translate_pythonop(NodeContext& context) {
+    auto decoder = context.get_decoder();
+    FRONT_END_OP_CONVERSION_CHECK(decoder->get_subgraph_size() == 1,
+                                  "PythonOp must have 1 subgraph to be able to translate it to OV.");
+    auto body = context.convert_subgraph(0);
+
+    std::map<size_t, ParameterVector> inputs_map;
+    for (auto param : body->get_parameters()) {
```
tomdol, line 21:
```suggestion
    for (const auto& param : body->get_parameters()) {
```

src/frontends/pytorch/src/op/pythonop.cpp
```diff
@@ -0,0 +1,51 @@
...
+        auto name = param->get_output_tensor(0).get_any_name();
+        size_t input_idx = (size_t)std::stoll(name);
```
tomdol, line 23: This call might throw if the conversion to a number cannot be performed. I'm not sure if the tensor names are guaranteed to be numbers here so I think you should add a try-catch around this code.

mmikolajcz: I see that better handling for tensor names is being implemented in https://github.com/openvinotoolkit/openvino/pull/15718 by @mvafin, maybe this ticket should wait for those changes to be merged?

src/frontends/pytorch/src/op/pythonop.cpp
mmikolajcz: Done

src/frontends/pytorch/src/op/pythonop.cpp
mvafin: Tensors are guranteed to be numbers, that is true even before #15718

src/frontends/pytorch/src/op/pythonop.cpp
```diff
@@ -0,0 +1,51 @@
...
+OutputVector translate_pythonop(NodeContext& context) {
+    auto decoder = context.get_decoder();
+    FRONT_END_OP_CONVERSION_CHECK(decoder->get_subgraph_size() == 1,
+                                  "PythonOp must have 1 subgraph to be able to translate it to OV.");
+    auto body = context.convert_subgraph(0);
+
+    std::map<int, ParameterVector> inputs_map;
+    for (const auto& param : body->get_parameters()) {
+        auto name = param->get_output_tensor(0).get_any_name();
+        int input_idx = (int)std::stoll(name);
```
mvafin, line 23:
```suggestion
        auto tensor_idx = m_translate_session->decode_tensor_name(param->output(0));
```

src/frontends/pytorch/src/op/pythonop.cpp
```diff
@@ -0,0 +1,51 @@
...
+namespace pytorch {
+namespace op {
+
+OutputVector translate_pythonop(NodeContext& context) {
+    auto decoder = context.get_decoder();
+    FRONT_END_OP_CONVERSION_CHECK(decoder->get_subgraph_size() == 1,
+                                  "PythonOp must have 1 subgraph to be able to translate it to OV.");
+    auto body = context.convert_subgraph(0);
+
+    std::map<int, ParameterVector> inputs_map;
```
mvafin, line 20:
```suggestion
    std::map<size_t, ParameterVector> inputs_map;
```

mvafin (changes requested):

tomdol (approved):

src/frontends/pytorch/src/op/pythonop.cpp
```diff
@@ -0,0 +1,51 @@
...
+            inputs_map[tensor_idx] = {param};
+        } else {
+            inputs_map[tensor_idx].push_back(param);
+        }
+    }
+    for (const auto& input : inputs_map) {
+        auto external_output = context.get_tensor_from_model(input.first);
+        if (external_output.get_node()) {
+            for (auto input_node : input.second) {
+                replace_node(input_node, context.get_input((int)input.first).get_node_shared_ptr());
```
mvafin, line 34: better would be to use `output.replace(new_output)`

src/frontends/pytorch/src/op/pythonop.cpp
```diff
@@ -0,0 +1,51 @@
...
+    auto body = context.convert_subgraph(0);
+    auto session = context.get_session();
+
+    std::map<size_t, ParameterVector> inputs_map;
+    for (const auto& param : body->get_parameters()) {
+        auto tensor_idx = session->decode_tensor_name(param->output(0));
+        if (!inputs_map.count(tensor_idx)) {
+            inputs_map[tensor_idx] = {param};
+        } else {
+            inputs_map[tensor_idx].push_back(param);
```
mvafin, line 27: Is it not expected and should throw exception? If more then 1 parameter has the same input?

src/frontends/pytorch/src/op/pythonop.cpp
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+    for (const auto& input : inputs_map) {
+        auto external_output = context.get_tensor_from_model(input.first);
+        if (external_output.get_node()) {
+            for (auto input_node : input.second) {
```
mvafin, line 33: If you have multiple nodes with same id it will replace it in sequence and result in only the last one existing. I think we need to throw exception in such cases, because it is wrong obviously.

mvafin (changes requested):

src/frontends/pytorch/src/op/pythonop.cpp
mmikolajcz: Done

src/frontends/pytorch/src/op/pythonop.cpp
mmikolajcz: Added `FRONT_END_OP_CONVERSION_CHECK` to not allow more than 1 parameter with the same input.

src/frontends/pytorch/src/op/pythonop.cpp
mmikolajcz: Added `FRONT_END_OP_CONVERSION_CHECK` not allowing multiple parameters for same input should prevent it. Removed for loop over input.second.

src/frontends/pytorch/src/op/pythonop.cpp
```diff
@@ -0,0 +1,47 @@
...
+        if (external_output.get_node()) {
+            input.second[0]->output(0).replace(external_output);
+        }
+    }
+
+    OutputVector outputs{};
+    for (auto result : body->get_results()) {
+        auto output = result->get_input_source_output(0).get_node_shared_ptr();
+        context.mark_node(output);
+        outputs.push_back(output);
```
mvafin, line 39:
```suggestion
        auto output = result->get_input_source_output(0);
        outputs.push_back(context.mark_output(output));
```

mvafin (approved):

PR #15719, opened 2023-02-15
[GPU] Improve OV_GPU_DumpLayers debug configuration

tuxedcat:
### Details:
 - Merge master && resolve conflict
 - Add program id prefix to prevent overwriting dump file.
 - `mem->count()` to `lock.size()`
 - Support wildcard `OV_GPU_DumpLayers="*"`

### Tickets:
 -


src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -979,11 +995,13 @@ void network::execute_impl(const std::vector<event::ptr>& events) {
...
 
-            GPU_DEBUG_IF(debug_config->dump_layers_dst_only == 0 &&
-                            debug_config->is_dumped_layer(layer_name)) {
+            GPU_DEBUG_IF(debug_config->dump_layers_dst_only == 0 && debug_config->is_dumped_layer(layer_name)) {
                 for (size_t i = 0; i < get_primitive(inst->id())->dependencies().size(); i++) {
-                    log_memory_to_file(get_primitive(inst->id())->dep_memory_ptr(i), get_stream(),
-                                       layer_name + "_src_" + std::to_string(i));
+                    log_memory_to_file(get_primitive(inst->id())->dep_memory_ptr(i),
+                                       get_stream(),
+                                       "program_" + std::to_string(get_program()->get_id()) + "_" + layer_name +
```
isanghao, line 1002: I think, to avoid conflict, it is necessary to use network id, too.. It is because a program may be instantiated into several networks.

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
```diff
@@ -117,6 +117,7 @@ static void print_help_messages() {
     message_list.emplace_back("OV_GPU_DumpLayersResult", "Dump output buffers of result layers only");
     message_list.emplace_back("OV_GPU_DumpLayersDstOnly", "Dump only output of layers");
     message_list.emplace_back("OV_GPU_DumpLayersLimitBatch", "Limit the size of batch to dump");
+    message_list.emplace_back("OV_GPU_DumpLayersRaw", "Dump memory of layers. If true, dump data is not formated.");
```
isanghao, line 120: `When dumping memory, store in raw memory format`

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
```diff
@@ -242,6 +245,8 @@ bool debug_configuration::is_dumped_layer(const std::string& layerName, bool is_
         return true;
     if (dump_layers.empty() && dump_layers_result == 0)
         return true;
+    if (dump_layers.size() == 1 && dump_layers[0] == "*")
+        return true;
```
isanghao, line 249: You don't need this. If you set DumpLayersPath, without DumpLayers it will dump all layers because `dump_layers.empty() ==  True`

isanghao (changes requested):

src/plugins/intel_gpu/src/graph/network.cpp
tuxedcat: Done.

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
tuxedcat: Done.

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
tuxedcat: Done.

isanghao (approved):

PR #15721, opened 2023-02-15
[PT FE]: fix aten::embedding realization for integer-like indicies an…

eaidova:
### Details:
 - *fix aten::embedding realization for integer-like indicies and add tests*
 tested on moussaKam/barthez-orangesum-abstract from hf

### Tickets
- *101072*

### DIFF
--- a/src/frontends/pytorch/src/op/embedding.cpp
+++ b/src/frontends/pytorch/src/op/embedding.cpp
@@ -24,7 +24,7 @@ OutputVector translate_embedding(NodeContext& context) {
     // padding_idx - if specified, the entries at padding_idx do not contribute to the gradient
     // scale_grad_by_freq - if given, this will scale gradients by the inverse of frequency of
     //                      the words in the mini-batch.
-    // sparce - if True, gradient will be represented as sparce tensor
+    // sparse - if True, gradient will be represented as sparse tensor
     auto axis_0 = context.mark_node(ov::op::v0::Constant::create(element::i64, Shape{}, {0}));
     return {context.mark_node(std::make_shared<ov::op::v8::Gather>(data, indices, axis_0))};
 };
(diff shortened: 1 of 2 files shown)


src/frontends/pytorch/src/op/embedding.cpp
```diff
@@ -13,10 +14,12 @@ namespace pytorch {
 namespace op {
 
 OutputVector translate_embedding(NodeContext& context) {
+    // aten::embedding(Tensor weight, Tensor indices, SymInt padding_idx=-1, bool scale_grad_by_freq=False, bool
+    // sparse=False) parameters 2, 3 and 4 used only for training case and connected with gradients propagation
```
mvafin, line 18: I don't think `sparse` is used only for training. `sparse` means that weight is sparse tensor, not dense. Also `padding_idx` is unclear how it is only for training.

eaidova: according to docs:
sparse ([bool](https://docs.python.org/3/library/functions.html#bool), optional) – If True, gradient w.r.t. weight matrix will be a sparse tensor. See Notes for more details regarding sparse gradients.



Note

Keep in mind that only a limited number of optimizers support sparse gradients: currently it’s optim.SGD (CUDA and CPU), optim.SparseAdam (CUDA and CPU) and optim.Adagrad (CPU)

https://pytorch.org/docs/stable/generated/torch.nn.Embedding.html#torch.nn.Embedding

eaidova: padding_idx ([int](https://docs.python.org/3/library/functions.html#int), optional) – If specified, the entries at padding_idx do not contribute to the gradient; therefore, the embedding vector at padding_idx is not updated during training, i.e. it remains as a fixed “pad”. For a newly constructed Embedding, the embedding vector at padding_idx will default to all zeros, but can be updated to another value to be used as the padding vector.

so it is padding for gradients.

eaidova: other parameters like max_norm and norm_type made on weights, made before aten::embeddings, they are not a part of op

mvafin: Okay, thank you for this analysis, that would be great if you add more detailed comment about why we skip those inputs.

eaidova: updated

mvafin (approved):

PR #15731, opened 2023-02-15
[TF FE] Support auto-pruning for Iterator, IteratorGetNext, Queue and Lookup

rkazants:
**Details:** Support auto-pruning for Iterator, IteratorGetNext, Queue and Lookup

**Tickets:** 92605


gkrivor (approved):

src/frontends/tensorflow/src/op/queue_dequeue.cpp
```diff
@@ -0,0 +1,77 @@
...
+        output_num == output_types.size(),
+        "[TensorFlow Frontend] Incorrect input model: lenghts of output_types and output_shapes do not match.");
+
+    OutputVector queue_dequeue_outputs;
+    // perform auto-pruning: create Parameter nodes for each output of QueueDequeue
+    for (size_t output_ind = 0; output_ind < output_num; ++output_ind) {
+        auto output_shape = output_shapes[output_ind];
+        auto output_type = output_types[output_ind];
+        output_shape.insert(output_shape.begin(), batch_dim);
+        auto parameter = make_shared<Parameter>(output_type, output_shape);
```
mvafin, line 63: I guess it would make sense to print a warning to user about this new input so that he knew what happened

rkazants: @mvafin, this is not reasonable to do for OV library and we won't clutter cout. Yes, this is minus of it and user should understand it by tensor names, for example, that are called usually `IteratorGetNext`.

discussed this with @ilyachur, @slyalin

mvafin: Do we have documentation for this case? That user can find by searching `IteratorGetNext`?

mvafin (approved):
Please ensure documentation cover that case in a way that user can easily find.

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -347,7 +346,16 @@ void TranslateSession::translate_graph(const ov::frontend::InputModel::Ptr& inpu
...
+    // after auto-pruning (IteratorGetNext), isolated Parameter nodes created by the auto-pruning
+    // can be remained in the graph so we need to clean them up
+    ov::ParameterVector final_params;
+    for (const auto& param : ordered_params) {
+        if (param->get_output_target_inputs(0).size() > 0) {
+            final_params.push_back(param);
+        }
+    }
+
+    ov_model = std::make_shared<ov::Model>(ordered_results, final_params, m_model_name);
```
rkazants, line 358:
```suggestion
    ov_model = std::make_shared<ov::Model>(ordered_results, ordered_params, m_model_name);
```

PR #15736, opened 2023-02-15
[GPU] Added operator== for cldnn primitives

Lyamin-Roman:
### Tickets:
 - *103414*


src/plugins/intel_gpu/include/intel_gpu/primitives/proposal.hpp
```diff
@@ -186,6 +186,35 @@ struct proposal : public primitive_base<proposal> {
         return seed;
     }
 
+    bool operator==(const primitive& rhs) const override {
+        if (!compare_common_params(rhs))
+            return false;
+
+        auto rhs_casted = downcast<const proposal>(rhs);
+
+        #define cmp_fields(name) name == rhs_casted.name
```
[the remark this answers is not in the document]
Lyamin-Roman: Added

src/plugins/intel_gpu/include/intel_gpu/primitives/binary_convolution.hpp
```diff
@@ -76,6 +76,20 @@ struct binary_convolution : public primitive_base<binary_convolution> {
...
+            return false;
+
+        auto rhs_casted = downcast<const binary_convolution>(rhs);
+
+        return pad == rhs_casted.pad &&
+               stride == rhs_casted.stride &&
+               dilation == rhs_casted.dilation &&
+               groups == rhs_casted.groups &&
+               pad_value == rhs_casted.pad_value &&
+               weights.size() == weights.size();
```
yeonbok, line 90: I think we can refactor these out as what @e-ddykim  done for proposal in this PR https://github.com/openvinotoolkit/openvino/pull/15335/files#diff-af2a6f9cbad55e0232a3e2689b9eb7702c7c33f0ae147c660fa28aa5618ed445, 
like, have a common binary dumper, use that binary for both hash and this comparator.

yeonbok: But that kind of refactoring can be done in a separate work. Just let me leave a note.

src/plugins/intel_gpu/include/intel_gpu/primitives/proposal.hpp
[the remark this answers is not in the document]
yeonbok: why this is only applied to this primitive? Will this be applied to others too?

src/plugins/intel_gpu/include/intel_gpu/primitives/binary_convolution.hpp
sshlyapn: @yeonbok, it would be nice to have such unification. Will primitive serialization include fields like `second_output` of `ctc_greedy_decoder`? and other which are compared as `field_name.empty() == rhs_casted.field_name.empty()` in Roman's PR? 
And did you have a chance to check performance impact of such approach? Wouldn't it be slowly than simple filed by field comparison?

sshlyapn, line 90: @Lyamin-Roman 
```suggestion
               weights.size() == rhs_casted.weights.size();
```

sshlyapn (approved):

PR #15739, opened 2023-02-15
[TF FE] Add Const to Result removing transformation

rkazants:
**Details:** Add Const to Result removing transformation. TensorFlow can remain training artifacts in models. It can be multiple isolated Constant nodes storing training artifacts. In IR they are expressed in a form of Const to Result sub-graphs. Multiple sub-graphs in the model can significantly affect performance.

**Ticket:** 103927

### DIFF
--- a/src/frontends/tensorflow_common/src/helper_transforms/const_to_result_remover.cpp
+++ b/src/frontends/tensorflow_common/src/helper_transforms/const_to_result_remover.cpp
@@ -21,12 +21,8 @@ bool ConstToResultRemover::run_on_model(const std::shared_ptr<ov::Model>& m) {
     // also, find isolated Constant->Result sub-graphs to remove
     for (const auto& result : m->get_results()) {
         auto unsupported_const = as_type_ptr<UnsupportedConstant>(result->get_input_node_shared_ptr(0));
-        if (unsupported_const && unsupported_const->output(0).get_target_inputs().size() == 1) {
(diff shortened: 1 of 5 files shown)


src/frontends/tensorflow_common/src/helper_transforms/const_to_result_remover.cpp
```diff
@@ -2,24 +2,32 @@
...
     // look for isolated UnsupportedConst->Result sub-graphs to remove
+    // also, find isolated Constant->Result sub-graphs to remove
     for (const auto& result : m->get_results()) {
         auto unsupported_const = as_type_ptr<UnsupportedConstant>(result->get_input_node_shared_ptr(0));
         if (unsupported_const && unsupported_const->output(0).get_target_inputs().size() == 1) {
             results_to_remove.push_back(result);
+            continue;
+        }
+        auto const_node = as_type_ptr<Constant>(result->get_input_node_shared_ptr(0));
+        if (const_node && const_node->output(0).get_target_inputs().size() == 1) {
```
itikhono, line 29: `const_node->output(0).get_target_inputs().size() == 1` I guess this check is unnecessary.
Constant will remain in the graph if we delete useless result here
image

rkazants: @itikhono, optimized

itikhono (approved):

PR #15740, opened 2023-02-15
[CPU] Select via Eltwise implementation

v-Golubev:
### Details:
 - *Select reimplemented via Eltwise node*
 - *Select single layer test class refactored, dynamic shapes instances added*

### Tickets:
 - *104055*


maxnick:
@a-sidorova , could you please review?

src/plugins/intel_cpu/src/cpu_types.cpp
```diff
@@ -143,7 +143,7 @@ const InferenceEngine::details::caseless_unordered_map<std::string, Type> type_t
         { "GridSample", Type::GridSample},
         { "OneHot", Type::OneHot},
         { "RegionYolo", Type::RegionYolo},
-        { "Select", Type::Select},
+        { "Select", Type::Eltwise},
```
a-sidorova, line 146: Minor:
I'd just to suggest to move `Select` to `Eltwise` [part](https://github.com/openvinotoolkit/openvino/blob/b793e1e8ae74595584a31f1dc6ccb7e11b92e588/src/plugins/intel_cpu/src/cpu_types.cpp#L28) of list

src/plugins/intel_cpu/tests/functional/single_layer_tests/select.cpp
```diff
@@ -1,174 +1,206 @@
...
+        auto select = ngraph::builder::makeSelect(paramOuts, broadcast);
+
+        function = makeNgraphFunction(precision, parameters, select, "Eltwise");
+   }
+
+   void generate_inputs(const std::vector<ov::Shape>& targetInputStaticShapes) override {
+        inputs.clear();
+        const auto& funcInputs = function->inputs();
+
+        auto cond_tensor = utils::create_and_fill_tensor(funcInputs[0].get_element_type(), targetInputStaticShapes[0], 1, 0, 1);
```
a-sidorova, line 74: To be honest I'm not sure that the tensor contains the both `true` and `false` here. I think there are only `false` values. By the way, I have locally checked values of tensor and I saw really only `false` values.

So I'd suggest to use the same [generation](https://github.com/openvinotoolkit/openvino/blob/bc663878eb5ecf2afb7a982ec85918285bc9c5b6/src/tests/functional/plugin/shared/src/snippets/select.cpp#L16) as the `Select` in `Snippets`. It generates different values. Also I'd suggest to generate different values for `then` and `else` inputs to see that `Select` correctly chooses values.

src/plugins/intel_cpu/src/cpu_types.cpp
v-Golubev: Done, thanks

src/plugins/intel_cpu/tests/functional/single_layer_tests/select.cpp
v-Golubev: My bad. Thanks a lot for noticing this defect! Corrected

a-sidorova (approved):

src/plugins/intel_cpu/src/nodes/eltwise.cpp
```diff
@@ -1660,6 +1666,13 @@ bool Eltwise::isSupportedOperation(const std::shared_ptr<const ngraph::Node>& op
                 return false;
             }
         }
+        if (const auto select = std::dynamic_pointer_cast<const ov::op::v1::Select>(op)) {
```
maxnick, line 1669: mgraph RTTI is more preferred to use with ngraph types.
```suggestion
        if (const auto select = ov::as_type_ptr<const ov::op::v1::Select>(op)) {
```

src/plugins/intel_cpu/tests/functional/single_layer_tests/select.cpp
```diff
@@ -1,174 +1,204 @@
...
+        inputs.insert({modelInputs[2].get_node_shared_ptr(), elseTensor});
+    }
+};
+
+TEST_P(SelectLayerCPUTest, CompareWithRefs) {
+    run();
+    CheckPluginRelatedResults(compiledModel, "Eltwise");
+}
+
+const std::vector<ElementType> precisions = {
```
maxnick, line 87: Das it make sense to add `i8` precision as well?

src/plugins/intel_cpu/src/nodes/eltwise.cpp
v-Golubev: OK, I replaced dynamic cast with ngraph RTTI in eltwise.cpp for ngraph types

src/plugins/intel_cpu/tests/functional/single_layer_tests/select.cpp
v-Golubev: Done, thanks

maxnick (approved):
Can be merged after internal validation

v-Golubev:
@maxnick, internal validation passed. Details can be found in the ticket

dmitry-gorokhov (approved):

PR #14666, opened 2022-12-15
convert_model() slow conversion time fix

popovaan:
Root cause analysis: 
Non-default convert_model() parames are serialized to rt_info section. If "input_model" model is an object then it lead to serializing of the model object to IR. Example of such line:
<input_model value="&lt;tensorflow.python.saved_model.load.Loader._recreate_base_user_object.&lt;locals>._UserObject object at 0x7fcc14508f70>" />

For big models it leads to long conversion time.

Solution: Not to serialize input_model value to rt_info if input model is model object.

Ticket: 99306

Code:
* [x]  Comments
* [x]  Code style (PEP8)
* [x]  Transformation generates reshape-able IR
* [x]  Transformation preserves original framework node names
* [x]  Transformation preserves tensor names

Validation:
* [x]  Unit tests
* [x]  Framework operation tests
* [x]  Transformation tests
* [x]  Model Optimizer IR Reader check

Documentation:
* [x]  Supported frameworks operations list
* [x]  Guide on how to convert the **public** model
* [x]  User guide update

PR #15742, opened 2023-02-15
convert_model() legacy extensions fix

popovaan:
Root cause analysis: 
cli_parser unites lists of paths into single string for most of parametes. But 'extensions' are expected to be stored as list, as it may contain non-string parameters came from convert_model(). This bug became visible after convert_model() refactoring https://github.com/openvinotoolkit/openvino/pull/14666 which removed redurant postprocessing of params from cli tool.

Solution: 
Keep extensions list as list and not unite to single string in cli parser.

Ticket: 103867

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests

Documentation:


mvafin (approved):

tests/layer_tests/mo_python_api_tests/test_mo_convert_extensions.py
```diff
@@ -1,15 +1,21 @@
 # Copyright (C) 2018-2023 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
```
rkazants, line 3: please add this test running in the pre-commit. Now it isn't, as I understand, so it can be easily broken:)

rkazants: `@pytest.mark.precommit` mark should not work

tests/layer_tests/mo_python_api_tests/test_mo_convert_extensions.py
```diff
@@ -120,3 +126,99 @@ def test_mo_convert_extensions(self, params, ie_device, precision, ir_version,
...
+            ov_model = convert_model(model, extensions=ext_path1)
+            flag, msg = compare_functions(ov_model, create_ref_model_1(), False)
+            assert flag, msg
+
+            ov_model = convert_model(model, extensions=[ext_path1, ext_path2])
+            flag, msg = compare_functions(ov_model, create_ref_model_2(), False)
+            assert flag, msg
+
+            ov_model = convert_model(model, extensions=','.join([ext_path1, ext_path2]))
+            flag, msg = compare_functions(ov_model, create_ref_model_2(), False)
```
rkazants, line 197: may be it is better to move it into `test_mo_convert_legacy_extensions.py`?

rkazants (commented):
one comment is to add this into the pre-commit.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -405,6 +399,11 @@ def prepare_ir(argv: argparse.Namespace):
     if moc_front_end:
         fallback_reasons = check_fallback(argv)
         if len(fallback_reasons) == 0:
```
popovaan, line 401: @rkazants I suggest to move creating of tmp .pb file here, because otherwise if we fallback to legacy path (due to usage of legacy extensions for example), .pb file is not removed. 
tmp .pb file is not used for legacy path, so I suggest to create it after we checked that fallback not needed.

tests/layer_tests/mo_python_api_tests/test_mo_convert_extensions.py
popovaan: Done.

rkazants:
@popovaan, please double-check e2e tests are failing.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -390,12 +390,6 @@ def prepare_ir(argv: argparse.Namespace):
     # TODO: remove this workaround once new TensorFlow frontend supports non-frozen formats: checkpoint, MetaGraph, and SavedModel
     # Now it converts all TensorFlow formats to the frozen .pb format in case new TensorFlow frontend
     is_tf, _, _, _, _ = deduce_legacy_frontend_by_namespace(argv)
```
rkazants, line 392: @popovaan, please move this comment along with the code.
Please do it this in the separate PR.

rkazants (approved):

PR #15743, opened 2023-02-15
ONNX functions expanding

tomdol:
### Details:
 - Fixes the TypeProto retrieval for ONNX context dependent functions
 - When the function Node is connected to an initializer it's possible to create a TypeProto on the fly using initializer's descriptor

### Tickets:
 - 103452

### DIFF
--- a/src/frontends/onnx/tests/onnx_import.in.cpp
+++ b/src/frontends/onnx/tests/onnx_import.in.cpp
@@ -509,6 +509,17 @@ NGRAPH_TEST(${BACKEND_NAME}, onnx_expand_context_dependent_function) {
     test_case.run();
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, onnx_expand_function_with_initializers) {
+    const auto function =
+        onnx_import::import_onnx_model(file_util::path_join(CommonTestUtils::getExecutableDirectory(),
+                                                            SERIALIZED_ZOO,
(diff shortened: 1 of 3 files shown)


src/frontends/onnx/tests/onnx_import.in.cpp
```diff
@@ -509,6 +509,17 @@ NGRAPH_TEST(${BACKEND_NAME}, onnx_expand_context_dependent_function) {
...
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, onnx_expand_function_with_initializers) {
+    const auto function =
+        onnx_import::import_onnx_model(file_util::path_join(CommonTestUtils::getExecutableDirectory(),
+                                                            SERIALIZED_ZOO,
+                                                            "onnx/transformations/celu_with_initializers.onnx"));
+
+    auto test_case = test::TestCase(function, s_device);
+    test_case.add_expected_output<float>({0.5, 1.0, 1.5, 2.0});
```
t-jankowski, line 519: Consider adding expected shape value as well.

t-jankowski (approved):

src/frontends/onnx/tests/onnx_import.in.cpp
tomdol: It's in the model and the TestCase class can use this information. The shapes have always been optional in these tests.

PR #15747, opened 2023-02-16
Fix the AUTO loadnetwork failure triggered  by AUTO:-xxx

yangwang201911:
### Details:
 - Check if the device is supported when AUTO retrieves the device list based on the ov::device::priorities.

### Tickets:
 - 99669


src/tests/unit/auto/get_device_list.cpp
```diff
@@ -33,7 +33,7 @@ const char igpuFullDeviceName[] = "Intel(R) Gen9 HD Graphics (iGPU)";
 const char dgpuFullDeviceName[] = "Intel(R) Iris(R) Xe MAX Graphics (dGPU)";
 // const char myriadFullDeviceName[] = "Intel Movidius Myriad X VPU";
 // const char vpuxFullDeviceName[] = "";
-const std::vector<std::string>  availableDevs = {"CPU", "GPU.0", "GPU.1", "VPUX"};
+const std::vector<std::string> availableDevs = {"CPU", "GPU.0", "GPU.1", "VPUX", "UNSUPPORTED_DEVICE"};
```
wangleis, line 36: I guess "UNSUPPORTED_DEVICE" needs different behavior compared to other available devices

peterchen-intel:
Add NVIDIA back and remove MYRIAD

peterchen-intel (approved):

src/plugins/auto/plugin.cpp
```diff
@@ -887,12 +884,17 @@ std::string MultiDeviceInferencePlugin::GetDeviceList(const std::map<std::string
                 updateDeviceVec(iter);
             }
             for (auto&& device : deviceVec) {
-                allDevices += device;
-                allDevices += ((device == deviceVec[deviceVec.size()-1]) ? "" : ",");
+                if (!_pluginConfig.isSupportedDevice(device))
+                    continue;
```
songbell, line 888: I'm thinking, how can you judge if device is supported when in future we remove the white list(which was originally used for secondary property checking?)

peterchen-intel: Will check if device is in blacklist in future, currently is checking if device is in whitelist.

wangleis (approved):

PR #15725, opened 2023-02-15
[GPU] Fix test tensor generator for blocked format

tuxedcat:
### Related Testcases
 - fusings_gpu/reduce_eltwise_activation_quantize.basic/53

### Details:
 - This is follow-up of #15447 without fp16 Fix (to avoid additional test failure as of now)
 - Fix it to generates values for blocked format correctly
   When data is not aligned with block size(e.g. fsv16 [b:1, f:15, x:2, y:1]), tensor should be filled with linear_offset function. But we are filling with raw data buffer index with data size.
      FillType::All - Generally recommended
      FillType::OnlyInsideShape - Required for reduce_sum because reduce kernel assumes the margin to be filled with zero.

### Tickets:
 - 67491

PR #15060, opened 2023-01-12
[Don't review] Fix reduce fsv16 bug

kelvinchoi-intel:
### Details:
 - *Fix reduce fsv16 kernel bug for MIN and MAX mode*
 - *Use fsv16 kernel for some condition or ref*

### Tickets:
 - *100837*

PR #15748, opened 2023-02-16
[GPU] Fix outside value of shape issue in reduce_gpu_b_fs_yx_fsv16

tuxedcat:
### Description
 - This is follow-up PR of  #15060
 - #15725 required to test outside value of shape
 - Not only `min` and `max`, but also `prod` and `logical_and` should not use outside value of shape.

### Details:
 - Write fail testcase
 - Change default behavior to handle outside value, and manage white list of optimizable operations.

### Tickets:
 - 67491


tuxedcat:
Waiting #15725

src/plugins/intel_gpu/src/kernel_selector/kernels/reduce/reduce_kernel_b_fs_yx_fsv16.cpp
```diff
@@ -216,13 +216,13 @@ JitConstants ReduceKernel_b_fs_yx_fsv16::GetJitConstants(const reduce_params& pa
...
-            if (params.inputs[0].Feature().v % 16 != 0) {
-                jit.AddConstant(MakeJitConstant("HANDLE_FEATURE_REMAINDER", 1));
-            }
-        }
+    bool zero_invariant_mode = params.reduceMode == ReduceMode::L1 || params.reduceMode == ReduceMode::L2 ||
+                               params.reduceMode == ReduceMode::LOG_SUM ||
+                               params.reduceMode == ReduceMode::LOG_SUM_EXP || params.reduceMode == ReduceMode::MEAN ||
+                               params.reduceMode == ReduceMode::OR || params.reduceMode == ReduceMode::SUM ||
+                               params.reduceMode == ReduceMode::SUM_SQUARE;
+    if (zero_invariant_mode || params.inputs[0].Feature().v % 16 == 0) {
```
isanghao, line 224: Let's revert condition
`        if (count(params.reduceAxes.begin(), params.reduceAxes.end(), 1) > 0) {`

src/plugins/intel_gpu/src/kernel_selector/kernels/reduce/reduce_kernel_b_fs_yx_fsv16.cpp
```diff
@@ -216,13 +216,14 @@ JitConstants ReduceKernel_b_fs_yx_fsv16::GetJitConstants(const reduce_params& pa
...
+        if (count(params.reduceAxes.begin(), params.reduceAxes.end(), 1) > 1)
```
isanghao, line 225: I think this should be         `if (count(params.reduceAxes.begin(), params.reduceAxes.end(), 1) == 1)`. isn't it?

tuxedcat: Sorry for the typo. `count(axes, 1) > 0` is correct because kernel only use `ZERO_INVARIANT_MODE` only when it has feature_axis(= axis 1) reduce.

src/plugins/intel_gpu/src/kernel_selector/kernels/reduce/reduce_kernel_b_fs_yx_fsv16.cpp
tuxedcat: Done.

byungilm:
Reducing logic of feature and batch axis does NOT have exception logic for un-aligned feature size.
It currently just use whole  value of all 16 sub_group which means feature axis.
In this PR, ZERO_INVARIANT_MODE should not be selected by mode. 
If feature axis is not aligned, it needs exception logic.

byungilm:
Moved fix to : https://github.com/openvinotoolkit/openvino/pull/17477

PR #15749, opened 2023-02-16
[GNA] Refactor memory alignment macro usage

tadamowicz:
### Details:
 - The 'memory requests' interface has been changed, and the alignment parameter is now provided in the constructor
 - Removed all occurrences of the magic number 64 and the ALIGN64 macro used in relation to memory alignment
 - Bug fix for issues related to memory alignment for graph compiler primitives, mostly not enough memory reserved for outputs. Issues observed in GNA_SW_FP32 mode.
 - Minor test update

### Tickets:
 - 102860

### DIFF
--- a/src/plugins/intel_gna/src/backend/gna_limitations.hpp
+++ b/src/plugins/intel_gna/src/backend/gna_limitations.hpp
@@ -55,6 +55,8 @@ constexpr uint32_t bytesPerSplitElement = 2;
 // In fp32 mode this is not necessary but is useful for testing
 constexpr uint32_t bytesPerCropElement = 2;
 
+constexpr uint32_t kMemoryAlignmentBytes = 64;
+
 inline bool isCropAffinedOffset(size_t numberOfElements) {
     const auto cropOffset = numberOfElements * bytesPerCropElement;
     return (ALIGN64(cropOffset) != cropOffset);
--- a/src/plugins/intel_gna/src/common/gna_target.cpp
+++ b/src/plugins/intel_gna/src/common/gna_target.cpp
@@ -74,10 +74,6 @@ HWGeneration DeviceToHwGeneration(const DeviceVersion& target) {
     return GetKeyForValue<HWGeneration, DeviceVersion>(target, HWGenerationDeviceMap);
 }
 
-size_t DeviceToGnaMemoryAlignment(const DeviceVersion& target) {
-    return 64;
-}
-
 DeviceVersion GnaToDevice(const Gna2DeviceVersion& target) {
     return GetValueForKey<Gna2DeviceVersion, DeviceVersion>(target, GnaDeviceMap);
 }
--- a/src/plugins/intel_gna/src/common/gna_target.cpp
+++ b/src/plugins/intel_gna/src/common/gna_target.cpp
@@ -74,6 +74,10 @@ HWGeneration DeviceToHwGeneration(const DeviceVersion& target) {
     return GetKeyForValue<HWGeneration, DeviceVersion>(target, HWGenerationDeviceMap);
 }
 
+size_t DeviceToGnaMemoryAlignment(const DeviceVersion& target) {
+    return 64;
+}
+
 DeviceVersion GnaToDevice(const Gna2DeviceVersion& target) {
     return GetValueForKey<Gna2DeviceVersion, DeviceVersion>(target, GnaDeviceMap);
 }
--- a/src/plugins/intel_gna/src/gna_device.cpp
+++ b/src/plugins/intel_gna/src/gna_device.cpp
@@ -48,6 +48,8 @@ GNADeviceHelper::GNADeviceHelper(std::shared_ptr<Target> targetIn, bool isPerfor
     GetGnaLibraryVersion();
 
     maxLayersCount_ = retrieveMaxLayersCount();
+
+    memAlignment = DeviceToGnaMemoryAlignment(targetIn->get_effective_compile_target());
 }
 
 GNADeviceHelper::~GNADeviceHelper() {
@@ -583,6 +585,5 @@ uint32_t GNADeviceHelper::retrieveMaxLayersCount() {
         return kMaxLayersCountGNA3_X;
     }
 }
-
 }  // namespace intel_gna
 }  // namespace ov
--- a/src/plugins/intel_gna/src/gna_device.cpp
+++ b/src/plugins/intel_gna/src/gna_device.cpp
@@ -49,7 +49,7 @@ GNADeviceHelper::GNADeviceHelper(std::shared_ptr<Target> targetIn, bool isPerfor
 
     maxLayersCount_ = retrieveMaxLayersCount();
 
-    memAlignment = DeviceToGnaMemoryAlignment(targetIn->get_effective_compile_target());
+    m_mem_alignment = limitations::kMemoryAlignmentBytes;
 }
 
 GNADeviceHelper::~GNADeviceHelper() {
--- a/src/plugins/intel_gna/src/gna_device.hpp
+++ b/src/plugins/intel_gna/src/gna_device.hpp
@@ -47,6 +47,7 @@ class GNADeviceHelper : public GNADevice {
     uint32_t nGnaDeviceIndex = 0;
     bool useDeviceEmbeddedExport = false;
     uint32_t maxLayersCount_ = 0;
+    size_t memAlignment = 0;
 
     static const uint32_t TotalGna2InstrumentationPoints = 2;
     Gna2InstrumentationPoint gna2InstrumentationPoints[TotalGna2InstrumentationPoints] = {
@@ -125,6 +126,11 @@ public:
 
     const GnaAllocations& getAllAllocations() const {
         return allAllocations;
+    }
+
+    const size_t getMemAlignment() const {
+        IE_ASSERT(memAlignment);
+        return memAlignment;
     }
 
     /**
--- a/src/plugins/intel_gna/src/gna_device.hpp
+++ b/src/plugins/intel_gna/src/gna_device.hpp
@@ -47,7 +47,7 @@ class GNADeviceHelper : public GNADevice {
     uint32_t nGnaDeviceIndex = 0;
     bool useDeviceEmbeddedExport = false;
     uint32_t maxLayersCount_ = 0;
-    size_t memAlignment = 0;
+    size_t m_mem_alignment = 0;
 
     static const uint32_t TotalGna2InstrumentationPoints = 2;
     Gna2InstrumentationPoint gna2InstrumentationPoints[TotalGna2InstrumentationPoints] = {
@@ -129,8 +129,7 @@ public:
     }
 
     const size_t getMemAlignment() const {
-        IE_ASSERT(memAlignment);
-        return memAlignment;
+        return m_mem_alignment;
     }
 
     /**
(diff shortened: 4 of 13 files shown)


src/plugins/intel_gna/src/common/gna_target.cpp
```diff
@@ -60,6 +60,10 @@ HWGeneration DeviceToHwGeneration(const DeviceVersion& target) {
     return GetKeyForValue<HWGeneration, DeviceVersion>(target, HWGenerationDeviceMap);
 }
 
+size_t DeviceToGnaMemoryAlignment(const DeviceVersion& target) {
+    return 64;
```
sirzabek, line 64: I suggest to keep it in gna_limitations as a macro definition.

src/plugins/intel_gna/src/gna_device.cpp
```diff
@@ -48,6 +48,8 @@ GNADeviceHelper::GNADeviceHelper(std::shared_ptr<Target> targetIn, bool isPerfor
     GetGnaLibraryVersion();
 
     maxLayersCount_ = retrieveMaxLayersCount();
+
+    memAlignment = DeviceToGnaMemoryAlignment(targetIn->get_effective_compile_target());
```
sirzabek, line 52: This is not according to our naming style convention.

src/plugins/intel_gna/src/gna_device.hpp
```diff
@@ -47,6 +47,7 @@ class GNADeviceHelper : public GNADevice {
     uint32_t nGnaDeviceIndex = 0;
     bool useDeviceEmbeddedExport = false;
     uint32_t maxLayersCount_ = 0;
+    size_t memAlignment = 0;
```
sirzabek, line 50:
```suggestion
    size_t m_mem_alignment = 0;
```

src/plugins/intel_gna/src/gna_device.hpp
```diff
@@ -127,6 +128,11 @@ class GNADeviceHelper : public GNADevice {
         return allAllocations;
     }
 
+    const size_t getMemAlignment() const {
+        IE_ASSERT(memAlignment);
+        return memAlignment;
```
sirzabek, line 133:
```suggestion
        return m_mem_alignment;
```

src/plugins/intel_gna/src/gna_graph_compiler.hpp
```diff
@@ -42,6 +42,7 @@ class GNAGraphCompiler {
     SplitConnection split_connection;
     CropConnection crop_connection;
     const Config& gna_config;
+    size_t mem_alignment = 0;
```
sirzabek, line 45:
```suggestion
    size_t m_mem_alignment = 0;
```

src/plugins/intel_gna/src/memory/gna_memory.hpp
```diff
@@ -173,6 +180,10 @@ class GNAMemory : public GNAMemoryInterface {
 #endif
     }
 
+    size_t getDataMemAlignment() const override {
```
sirzabek, line 183: Seems to be unused.

src/plugins/intel_gna/src/gna_plugin.cpp
```diff
@@ -852,10 +853,6 @@ void GNAPlugin::LoadNetwork(const CNNNetwork& _network) {
         }
         portId++;
     }
-    // TODO: how active list will work in multioutput case
-    // make room for active list
-    gnamem->getQueue(REGION_OUTPUTS)
```
sirzabek, line 857: Was it not needed?

src/plugins/intel_gna/src/memory/gna_mem_requests_queue.hpp
```diff
@@ -38,39 +38,36 @@ inline uint16_t getCNNLayerId(InferenceEngine::CNNLayerPtr layer) {
...
+        : _region_type(region),
+          _alignment(mem_alignment) {}
     virtual ~GNAMemRequestsQueue() {}
 
     rRegion _region_type;
     size_t _size = 0;
     std::vector<MemRequest> _mem_requests;
     std::list<std::vector<char>> _local_storage;
     std::shared_ptr<uint8_t> _basePtr = nullptr;
+    size_t _alignment;
```
sirzabek, line 51:
```suggestion
    size_t m_alignment;
```

src/plugins/intel_gna/src/memory/gna_mem_requests_queue.hpp
```diff
@@ -38,39 +38,36 @@ inline uint16_t getCNNLayerId(InferenceEngine::CNNLayerPtr layer) {
...
 
     /**
      * @brief register initialiser to access memory once it is actually allocated
      * @param ptr_out
      * @param ptr_in
      * @param num_bytes
-     * @param alignment
```
sirzabek, line 55: Missing initializer @param

src/plugins/intel_gna/src/memory/gna_mem_requests_queue.hpp
```diff
@@ -114,14 +114,23 @@ class GNAMemRequestsQueue {
      * @param dest - source is binded to dest pointer after allocation
      * @param offset - offset in bytes in source that will be set in dest
      * @param num_bytes - bind can request for bigger buffer that originally allocated via reserve(),
+     * @param aling_num_bytes - when true then align the number of bytes to the alignment predefined for the queue
```
sirzabek, line 117:
```suggestion
     * @param align_num_bytes - when true then align memory size to the alignment predefined for the queue
```

src/plugins/intel_gna/src/memory/gna_mem_requests_queue.hpp
```diff
@@ -100,9 +93,16 @@ class GNAMemRequestsQueue {
      *
      * @param ptr_out
      * @param num_bytes
+     * @param aling_num_bytes
```
sirzabek, line 96:
```suggestion
     * @param align_num_bytes
```

src/plugins/intel_gna/src/memory/gna_mem_requests_queue.hpp
```diff
@@ -100,9 +93,16 @@ class GNAMemRequestsQueue {
...
      */
-    void reserve_ptr(InferenceEngine::CNNLayerPtr layer, void* ptr_out, size_t num_bytes, size_t alignment = 1) {
-        futureHeap().push_back({regionType(), REQUEST_ALLOCATE, ptr_out, nullptr, 1, num_bytes, alignment});
+    void reserve_ptr(InferenceEngine::CNNLayerPtr layer, void* ptr_out, size_t num_bytes, bool aling_num_bytes = true) {
```
sirzabek, line 98:
```suggestion
    void reserve_ptr(InferenceEngine::CNNLayerPtr layer, void* ptr_out, size_t num_bytes, bool align_num_bytes = true) {
```

src/plugins/intel_gna/src/memory/gna_mem_requests_queue.hpp
```diff
@@ -114,14 +114,23 @@ class GNAMemRequestsQueue {
...
      *      if that happens - reserved request parameters will be updated before committing memory
      */
     void bind_ptr(InferenceEngine::CNNLayerPtr layer,
                   void* source,
                   const void* dest,
                   size_t offset = 0,
-                  size_t num_bytes = 0) {
-        futureHeap().push_back({regionType(), REQUEST_BIND, source, dest, 1, num_bytes, 1, offset});
+                  size_t num_bytes = 0,
+                  bool aling_num_bytes = true) {
```
sirzabek, line 125:
```suggestion
                  bool align_num_bytes = true) {
```

sirzabek (changes requested):

src/plugins/intel_gna/src/gna_plugin.cpp
tadamowicz: When determining the required memory size or allocating memory, a check is performed which excludes this function call from being effective if the argument "_ptr_out" is a null pointer.

    /**
     * @brief allocates memory and updates pointers
     */
    void allocateRegion(GNAMemRequestsQueue* mRequests) {
        size_t r_size = ALIGN(mRequests->getSize(), _page_alignment);
        for (auto& re : mRequests->_mem_requests) {
            // skipping Bind, crossregion and empty requests
            if (re._type == REQUEST_BIND || re._ptr_out == nullptr)
                continue;

src/plugins/intel_gna/src/backend/gna_limitations.hpp
```diff
@@ -55,6 +55,8 @@ constexpr uint32_t bytesPerSplitElement = 2;
 // In fp32 mode this is not necessary but is useful for testing
 constexpr uint32_t bytesPerCropElement = 2;
 
+constexpr uint32_t kMemoryAlignment = 64;
```
sirzabek, line 58:
```suggestion
constexpr uint32_t kMemoryAlignmentBytes = 64;
```

sirzabek (approved):

src/plugins/intel_gna/src/gna_device.hpp
```diff
@@ -47,6 +47,7 @@ class GNADeviceHelper : public GNADevice {
     uint32_t nGnaDeviceIndex = 0;
     bool useDeviceEmbeddedExport = false;
     uint32_t maxLayersCount_ = 0;
+    size_t m_mem_alignment = 0;
```
marcinkusm, line 50: 1. why not to set this value in header?
```
size_t m_mem_alignment  {limitations::kMemoryAlignment};
```
2. I don't see place where this value is changes shouldn't it be const/constexpr ?

src/plugins/intel_gna/src/gna_device.hpp
```diff
@@ -127,6 +128,11 @@ class GNADeviceHelper : public GNADevice {
         return allAllocations;
     }
 
+    const size_t getMemAlignment() const {
+        IE_ASSERT(m_mem_alignment);
```
marcinkusm, line 132: Why do we need this assertion?
Is there a chance to have 0 here?
Shouldn't be it rather done on set?

marcinkusm (approved):
👍 
Looks nice,
just minor comments

src/plugins/intel_gna/src/gna_device.hpp
tadamowicz: good finding, there is no need for assertion here

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -78,6 +78,7 @@ GNAGraphCompiler::GNAGraphCompiler(const Config& gna_config) : gna_config(gna_co
 
 void GNAGraphCompiler::setGNAMemoryPtr(std::shared_ptr<gna_memory_type> gnaMemPtr) {
     this->gnamem = std::move(gnaMemPtr);
+    this->m_mem_alignment = this->gnamem->getDataMemAlignment();
```
mryzhov, line 81: we don't need this class property, because it is already set in gnamem

mryzhov (changes requested):

mryzhov (commented):
I would suggest another implementation.  Alignment is a device specific property and we initialize device when setting up the plugin. It would be better to make alignment hidden in the plugin itself, for example in the code will use device helper method.
E.g.:
instead of ALIGN(size, _data_alignment) -> device::align(size)

There are a lot of places where we have to keep in mind the alignment size

sirzabek:
> I would suggest another implementation. Alignment is a device specific property and we initialize device when setting up the plugin. It would be better to make alignment hidden in the plugin itself, for example in the code will use device helper method. E.g.: instead of ALIGN(size, _data_alignment) -> device::align(size)
> 
> There are a lot of places where we have to keep in mind the alignment size

It's not always true that we initialize device when setting up the plugin - in some scenarios it can be initialized on a later stage or even reinitialized during plugin existence.

mryzhov (approved):

PR #15754, opened 2023-02-16
[WIP] increase threshold to keep Consts in FP32 during compression

pavel-esir:
- tbd: provide more solid justification about choosing threshold value
- tbd: find out why Const values cause performance degradation on CPU

### Tickets:
 - xxx-103558


pavel-esir:
Marked as ready for review to get jenkins build and generate reference IRs. Do not review yet

pavel-esir:
Closing. There is no need to increase threshold, instead CPU should take a look why resnext101-32x16d-swsl model if its constants do not go though fp16 compression is run much slower.

PR #15759, opened 2023-02-16
[PyOV][CI] Simplify requirement files - 103925

rafalstaroszczyk1:


rkazants (commented):
why do we need this PR if we have already opened: https://github.com/openvinotoolkit/openvino/pull/15343?

akladiev:

PR #15769, opened 2023-02-16
[PT FE] Add aten::ArgSort & aten::Sort

PiotrKrzem:
### Details:
- Add sort implementation for PyTorch frontend
 - Add argsort implementation for PyTorch frontend
 - Note: argument order based on PyTorch implementation [here](https://github.com/pytorch/pytorch/pull/87572/)

### Tickets:
 - 101131
 - 101081


mvafin (approved):

src/frontends/pytorch/src/op/argsort.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+#define mark(...) context.mark_node(__VA_ARGS__)
+#define shared    std::make_shared
```
mvafin, line 13: Please remove these defines, I don't think this is a good practice to have them.

mvafin (changes requested):

src/frontends/pytorch/src/op/argsort.cpp
```diff
@@ -0,0 +1,49 @@
...
+        descending = context.const_input<bool>(2);
+    }
+    auto mode = descending ? TopKMode::MAX : TopKMode::MIN;
+
+    auto zero_axis = mark(Constant::create(element::i32, Shape{1}, {0}));
+    auto dim_axis = mark(Constant::create(element::i64, Shape{1}, {dim}));
+    auto shape = mark(shared<ShapeOf>(input_tensor));
+    auto k_values_node = mark(shared<Gather>(shape, dim_axis, zero_axis));
+    auto k_values = mark(shared<Squeeze>(k_values_node));
+    auto topk = mark(shared<TopK>(input_tensor, k_values, dim, mode, TopKSortType::NONE));
```
tomdol, line 40: I think you should sort by values here to match what aten::argsort does https://pytorch.org/docs/stable/generated/torch.argsort.html

tomdol (changes requested):

tests/layer_tests/pytorch_tests/test_argsort.py
```diff
@@ -0,0 +1,76 @@
...
+    @pytest.mark.parametrize("dim", [
+        0
+    ])
+    @pytest.mark.parametrize("descending", [
+        False
+    ])
+    @pytest.mark.parametrize("stable", [
+        True
+    ])
+    @pytest.mark.xfail
```
mmikolajcz, line 73: Wouldn't this xfail allow to fail for all tests? Maybe it would be better to do something like that?
`@pytest.mark.parametrize(
        "stable",
        [
            False, None,
            pytest.param(
                True,
                marks=pytest.mark.xfail(
                    reason="Failed due to TopK not supporting... ticket: ..."
                ),
            ),
        ],
    )`

mmikolajcz (changes requested):

tests/layer_tests/pytorch_tests/test_argsort.py
mmikolajcz: and add:
```
@pytest.mark.nightly
@pytest.mark.precommit
```

src/frontends/pytorch/src/op/argsort.cpp
PiotrKrzem: Right, updated!

tests/layer_tests/pytorch_tests/test_argsort.py
PiotrKrzem: Good idea, added!

src/frontends/pytorch/src/op/argsort.cpp
PiotrKrzem: Removed

PiotrKrzem:
The sort implementation is present here as well because I merged the two implementations together with a common utils function

tomdol (approved):
Since this PR contains 2 ops, shouldn't you close https://github.com/openvinotoolkit/openvino/pull/15957?

src/frontends/pytorch/src/op/argsort.cpp
```diff
@@ -0,0 +1,34 @@
...
+        dim = context.const_input<int64_t>(2);
+        descending = context.const_input<bool>(3);
+        FRONT_END_OP_CONVERSION_CHECK(stable == false, "Stable sorting in aten::argsort is not yet supported.");
+    } else {
+        dim = context.const_input<int64_t>(1);
+        descending = context.const_input<bool>(2);
+    }
+
+    auto sort = sort_elements(context, input_tensor, stable, dim, descending);
+    return {sort->output(1)};
```
mmikolajcz, line 28: Since both implementations are using the same code, maybe it would be better to move it to the same file as regular sort and do something like that to avoid code duplication?
```
OutputVector translate_argsort(NodeContext& context) {
    return {translate_sort(context)[1]}
}
```

PiotrKrzem: Done!

mmikolajcz (approved):

mvafin (changes requested):

src/frontends/pytorch/src/op/sort.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_sort(NodeContext& context) {
+    const auto input_tensor = context.get_input(0);
```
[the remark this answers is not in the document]
PiotrKrzem: Done!

mvafin (approved):

PR #15784, opened 2023-02-17
ReverseInputChannelsFusion - no reverse input channels -> return

mateusztabaka:
ReverseInputChannelsFusion - return early if there is no reverse input channels

Ticket: 98067


src/core/include/openvino/pass/manager.hpp
```diff
@@ -60,7 +60,7 @@ class OPENVINO_API Manager {
         return pass;
     }
 
-    void run_passes(std::shared_ptr<Model>);
```
itikhono, line 63: please add the description

mateusztabaka: done

src/core/src/pass/manager.cpp
```diff
@@ -132,9 +133,12 @@ void ov::pass::Manager::run_passes(shared_ptr<ov::Model> func) {
...
             cout << setw(7) << pass_timer.get_milliseconds() << "ms " << pass->get_name() << "\n";
         }
+        needs_validate = function_changed;
     }
     if (profile_enabled) {
         cout << "passes done in " << overall_timer.get_milliseconds() << "ms\n";
     }
     NGRAPH_SUPPRESS_DEPRECATED_END
+
+    return function_changed;
```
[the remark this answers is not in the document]
mateusztabaka: done

src/core/src/pass/manager.cpp
```diff
@@ -138,9 +140,13 @@ void ov::pass::Manager::run_passes(shared_ptr<ov::Model> func) {
         if (profile_enabled) {
             cout << setw(7) << pass_timer.get_milliseconds() << "ms " << pass->get_name() << "\n";
         }
+        function_changed = function_changed || pass_applied;
+        needs_validate = needs_validate || pass_applied;
```
itikhono, line 144: looks correct, but if there is a mistake, it will significantly affect performance. Could you double-check perfomance for couple of models (locally or via dl bench)

mateusztabaka: tests shared offline

itikhono (approved):

rkazants (commented):
can we have a test please?

PR #15786, opened 2023-02-17
Remove WA for myriad plugin

olpipi:
### Details:
 - Remove WA for myriad plugin in tests because it is not longer needed

### Tickets:
 - CVS-104079


ilya-lavrenov (approved):

iefode:
@olpipi We have the same w/a for ov_cache. Please remove

PR #15792, opened 2023-02-17
Fix tile shape inference when repeats got dynamic shape

praasz:
### Details:
Fix tile shape calculation in case when repeats input has got dynamic shape.

### DIFF
--- a/src/core/shape_inference/include/tile_shape_inference.hpp
+++ b/src/core/shape_inference/include/tile_shape_inference.hpp
@@ -21,7 +21,8 @@ std::vector<T> shape_infer(const Tile* op,
     NODE_VALIDATION_CHECK(op, input_shapes.size() == 2);
 
     const auto& repeats_shape = input_shapes[1];
-    NODE_VALIDATION_CHECK(op, repeats_shape.rank().compatible(1), "Tile repeats must be of rank 1");
+    const auto& repeats_rank = repeats_shape.rank();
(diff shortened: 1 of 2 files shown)


src/core/shape_inference/include/tile_shape_inference.hpp
```diff
@@ -51,9 +52,10 @@ std::vector<T> shape_infer(const Tile* op,
                        rep_it,
                        std::back_inserter(output_shape),
                        std::multiplies<TDim>());
-    } else if (arg_rank.is_static() && repeats_shape[0].is_static()) {
-        // unknown repeats but shape is 1-D static, any dim can be repeated (add missing dimension)
-        output_shape.resize(std::max<size_t>(arg_rank.get_length(), repeats_shape[0].get_length()));
+    } else if (arg_rank.is_static()) {
+        // unknown repeats any dim can be repeated (add missing dimension)
+        const auto r = (repeats_rank.is_static() && repeats_shape[0].is_static()) ? repeats_shape[0].get_length() : 0;
+        output_shape.resize(std::max<size_t>(arg_rank.get_length(), r));
```
jane-intel, line 58: This is incorrect behavior. Please see the specification. repeats could be longer than data rank and it will extend output rank.

praasz: Changed to return dynamic shape.

ilyachur (approved):

PR #15795, opened 2023-02-17
[CPU] Internal dynamism support for extensions

maxnick:
### Details:
This PR introduce support for extension operations with so called internal dynamism, i.e. when the output shapes can only be defined after performing the computations, for example NMS operation.

### Tickets:
 - 99106
 - 81379

### DIFF
--- a/src/core/dev_api/shape_util.hpp
+++ b/src/core/dev_api/shape_util.hpp
@@ -16,7 +16,7 @@ namespace util {
  * @return 2-D shape with {0, SIZE_MAX}
  */
 OPENVINO_DEPRECATED("This function is deprecated and will be removed soon.")
-Shape make_dynamic_shape();
+OPENVINO_API Shape make_dynamic_shape();
 
 /**
  * @brief Check if Shape is marked as dynamic.
@@ -25,6 +25,6 @@ Shape make_dynamic_shape();
  * @return True if shape is dynamic otherwise false.
  */
 OPENVINO_DEPRECATED("This function is deprecated and will be removed soon.")
-bool is_dynamic_shape(const Shape& s);
+OPENVINO_API bool is_dynamic_shape(const Shape& s);
 }  // namespace util
 }  // namespace ov
--- a/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.hpp
+++ b/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.hpp
@@ -9,14 +9,21 @@
 #include <openvino/core/node.hpp>
 
 #include "static_shape.hpp"
+#include "shape_inference_status.hpp"
 
 namespace ov {
 namespace intel_cpu {
 
 class IShapeInferCommon {
 public:
-    virtual std::vector<StaticShape> infer(const std::vector<StaticShape>& input_shapes,
-                                           const std::map<size_t, HostTensorPtr>& constant_data) = 0;
+    struct ShapeInferResult {
+        std::vector<StaticShape> shapes;
+        ShapeInferStatus status;
+    };
+
+public:
+    virtual ShapeInferResult infer(const std::vector<StaticShape>& input_shapes,
+                                   const std::map<size_t, HostTensorPtr>& constant_data) = 0;
 
     // infer may generate padding as by-product, these APIs is designed to retrieve them back
     virtual const ov::CoordinateDiff& get_pads_begin() = 0;
@@ -29,7 +36,7 @@ class IStaticShapeInfer : public IShapeInferCommon {
 public:
     using IShapeInferCommon::infer;
 
-    virtual std::vector<StaticShape> infer(
+    virtual ShapeInferResult infer(
         const std::vector<StaticShape>& input_shapes,
         const std::map<size_t, std::reference_wrapper<const Tensor>>& constant_data) = 0;
 };
--- a/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.hpp
+++ b/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.hpp
@@ -16,13 +16,13 @@ namespace intel_cpu {
 
 class IShapeInferCommon {
 public:
-    struct ShapeInferResult {
+    struct Result {
         std::vector<StaticShape> shapes;
         ShapeInferStatus status;
     };
 
 public:
-    virtual ShapeInferResult infer(const std::vector<StaticShape>& input_shapes,
+    virtual Result infer(const std::vector<StaticShape>& input_shapes,
                                    const std::map<size_t, HostTensorPtr>& constant_data) = 0;
 
     // infer may generate padding as by-product, these APIs is designed to retrieve them back
@@ -36,7 +36,7 @@ class IStaticShapeInfer : public IShapeInferCommon {
 public:
     using IShapeInferCommon::infer;
 
-    virtual ShapeInferResult infer(
+    virtual Result infer(
         const std::vector<StaticShape>& input_shapes,
         const std::map<size_t, std::reference_wrapper<const Tensor>>& constant_data) = 0;
 };
--- a/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_cpu.hpp
+++ b/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_cpu.hpp
@@ -8,6 +8,8 @@
 #include <cpu_memory.h>
 #include <openvino/core/node.hpp>
 
+#include "shape_inference_status.hpp"
+
 namespace ov {
 namespace intel_cpu {
 /**
@@ -18,6 +20,11 @@ class IShapeInfer {
 public:
     using port_mask_t = uint32_t;
 
+    struct ShapeInferResult {
+        std::vector<VectorDims> dims;
+        ShapeInferStatus status;
+    };
+
 public:
     ~IShapeInfer() = default;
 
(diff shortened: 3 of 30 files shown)


src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_status.hpp
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+namespace ov {
+namespace intel_cpu {
+
+enum class ShapeInferStatus {
+    update, // shapes were successfully calculated
```
praasz, line 11: `update` -> `success` as shape `infer` return value not update it?

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.hpp
```diff
@@ -20,8 +21,14 @@ void shape_inference(ov::Node* op,
...
 public:
-    virtual std::vector<StaticShape> infer(const std::vector<StaticShape>& input_shapes,
-                                           const std::map<size_t, HostTensorPtr>& constant_data) = 0;
+    struct ShapeInferResult {
+        std::vector<StaticShape> shapes;
+        ShapeInferStatus status;
+    };
+
+public:
+    virtual ShapeInferResult infer(const std::vector<StaticShape>& input_shapes,
```
praasz, line 30: As the infer result status has got only two states now (not sure if there is a plan to expand it), maybe there is a chance to use empty vector as skipped status?

In template implementation of shape infer it looks like it always returns not empty vector or throw exception if inference fails. The empty  vector could be interpreted as infer has been skipped. Then return type of this function doesn't need to be changed, and changes will be less impacting especially on classes where this status has no special meaning and they always return `updated`.

Other solution (to expand interface not modify it) can be add additional function to this interface e.g. `get_status()` which provide last infer status, for most classes it can be compile time value as it never changes.
Then `entryBase` can implement this getter with default (updated) value and only in required classes override it.

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_cpu.hpp
```diff
@@ -18,6 +20,11 @@ class IShapeInfer {
 public:
```
praasz, line 20: Maybe this interface can inherit from `IShapeInferCommon` (or some new Iface which got common members for both interfaces) or vice versa?

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.hpp
maxnick: Thank you for your suggestions.

>In template implementation of shape infer it looks like it always returns not empty vector or throw exception if inference fails. The empty vector could be interpreted as infer has been skipped. Then return type of this function doesn't need to be changed, and changes will be less impacting especially on classes where this status has no special meaning and they always return `updated`.

Of course, this is the first idea that came to my mind. But, what if the empty vector is a result of some error? In that case the program can not distinguish whether the empty vector is `skip` or we got unexpectedly empty data. Thus, we will have hardly detecting error. The sing is on the `IShapeInferCommon` hierarchy level we do not have control over the template `shape_infer` functions, and in the case of using an empty vector as a `skip` indication, we may just "misrepresent" an error as an valid result. But with the approach proposed in the PR we clearly indicate what we expect. Also, since this "complexity" is located on this level is think this is not a big problem.

>Other solution (to expand interface not modify it) can be add additional function to this interface e.g. get_status() which provide last infer status, for most classes it can be compile time value as it never changes.
Then `entryBase` can implement this getter with default (updated) value and only in required classes override it.

In fact, this approach does not reduce the complexity, but makes the interface less convenient, since the user should remember about the state that should be read with `get_status()`. Yes, in this proposal the user will also have to check the return status, but he receive it together with the resulting shape, so this is less likely that he forget to check it, so slightly more convenient.

But, to be frank, I do not completely like the solution either, but given the options we have, I think the proposed solution is better than others. Also in the CPU plugin I introduced the same `Result` type and it can be used to optimize the shape infer call. We can simply skip the computations if the input data have not been changed from the previous call.
If my reasoning is not good enough, further discussion is always welcome (:

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_cpu.hpp
maxnick: Unfortunately, we had to implement this interface, since the `IShapeInferCommon` use incompatible static shape representation and conversations take too much time. I mean `VectorDims` <-> `StaticShape`, map of host tensors <-> map of memory object pointers. So to implement optimized shape inference operations, optimized here means CPU plugin specifics and constraints, we introduced this interface, fully compatible with the plugin, and just wrapped `IShapeInferCommon` as more generic. But, I see that nobody else use `IShapeInferCommon` so m.b. the entire subsytem may adopt the CPU plugin specific `IShapeInfer `, but under the hood we will have to convert `VectorDims` <-> `StaticShape` and input maps anyway.

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_status.hpp
maxnick: Changed, thank you!

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.hpp
```diff
@@ -20,8 +21,14 @@ void shape_inference(ov::Node* op,
 
 class IShapeInferCommon {
 public:
-    virtual std::vector<StaticShape> infer(const std::vector<StaticShape>& input_shapes,
-                                           const std::map<size_t, HostTensorPtr>& constant_data) = 0;
+    struct ShapeInferResult {
```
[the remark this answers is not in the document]
maxnick: I renamed the struct, thank you for your suggestion!

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.hpp
praasz: Change interface is also fine, just was thinking about other solutions.
For me it looks like the `std::optional` from c++17 could be useful here.

praasz (approved):

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.hpp
maxnick: Completely agree with you. But we cannot use c++17 so far.

src/core/src/tensor_conversion_util.cpp
```diff
@@ -18,7 +18,7 @@ Tensor wrap_tensor(const ngraph::HostTensorPtr& t) {
     } else if (p_shape.is_static()) {
         return {et, p_shape.to_shape(), t->get_data_ptr()};
     } else {
-        return {et, make_dynamic_shape()};
+        return {et, ov::Shape(p_shape.rank().get_length())};
```
praasz, line 21: Why is this changed?
It provides correct conversion between dynamic host tensor and ov::tensor in evaluates and  this WA fix a lot of issues until old tensor will be removed.


@ilyachur can you have a look?

praasz (changes requested):

src/core/src/tensor_conversion_util.cpp
maxnick: Obviously this changes do not work. Need to discuss working with `ov::Tensor`  in the case of dynamics shapes.
The main problem is that `set_shape` call to the `ov::Tensor` instance created from "dynamic" `ov::Shape`, which can not really represent a dynamic shape and that is why we use `ov::PartialShape`, can not be performed since in that case the tensor object is in the "uninitialized" state. But, what should the `evaluate` implementation do in that case? That is the question. Because following the interface of the `ov::Tensor` it seems that this is enough to simply call `set_shape`, but it turns out that it throws if the tensor is not initialized. In other words, `ov::Tensor` is not designed to have "undefined" state, but we need this to at least express operations with "internal" dynamism.

praasz: There is assumption that if `ov::Tensor` is create as "dynamic", `set_shape` will allow change to any shape. This is already used in [softsign evaluate](https://github.com/openvinotoolkit/openvino/blob/27ea9eab329cd7db547f5aeb24bfca7e549ef489/src/core/src/op/softsign.cpp#L95), and there is a case when this evaluate is called with "dynamic" output tensor and reshaped.

Maybe `set_shape` WA should be improved?

maxnick: It looks like I mixed up `ov::Tensor` and `runtime::Tensor`, since the letter being initialized from "dynamic" `ov::Shape` can not perform `set_shape` call, because an exception is thrown. But, `ov::Tensor` performs `set_shape` properly, even though being instantiated from the result of the `ov::util::make_dynamic_shape()` call.  So I adopted the `ov::util::make_dynamic_shape()` utility on the plugin level.

src/core/dev_api/shape_util.hpp
```diff
@@ -25,6 +25,6 @@ Shape make_dynamic_shape();
  * @return True if shape is dynamic otherwise false.
  */
 OPENVINO_DEPRECATED("This function is deprecated and will be removed soon.")
-bool is_dynamic_shape(const Shape& s);
+OPENVINO_API bool is_dynamic_shape(const Shape& s);
```
praasz, line 28: can you check if export of this required. I think is used in core only. 
If not required maybe is better not export as also marked as deprecated.

praasz (approved):

src/core/dev_api/shape_util.hpp
maxnick: It is used by the CPU plugin as well, so this is the reason why they are exported. I also though about making these utils header only, but the usage of a static variable changed my mind.

maxnick:
@ilyachur , do you agree with moving the shape utils to the dev api?

maxnick:

ilyachur (approved):
Core part LGTM

dmitry-gorokhov (approved):
Would be nice to provide a little bit broader explanation for "skip" status and use-cases there it is expected to be applied.

maxnick:
> Would be nice to provide a little bit broader explanation for "skip" status and use-cases there it is expected to be applied.

@dmitry-gorokhov , the code comments have been extended.

PR #15798, opened 2023-02-17
[GPU] Added shape agnostic kernels for GatherElements and Tile

Lyamin-Roman:
### Tickets:
 - *95104*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/gather_elements.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/gather_elements.cpp
@@ -65,24 +65,36 @@ struct gather_elements_impl : typed_primitive_impl_ocl<gather_elements> {
         params.inputs.push_back(convert_data_tensor(impl_param.get_input_layout(1)));
         return {params, optional_params};
     }
+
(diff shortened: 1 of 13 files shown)


andrew-k-park (approved):

src/plugins/intel_gpu/src/graph/impls/ocl/gather_elements.cpp
```diff
@@ -69,24 +69,36 @@ struct gather_elements_impl : typed_primitive_impl_ocl<gather_elements> {
         params.inputs.push_back(convert_data_tensor(impl_param.get_input_layout(1)));
         return {params, optional_params};
     }
+
+    void update_dispatch_data(const kernel_impl_params& impl_param) override {
+       auto kernel_params = get_kernel_params(impl_param);
```
andrew-k-park, line 74:
```suggestion
       auto kernel_params = get_kernel_params(impl_param, true);
```
To set `is_shape_agnostic` of kernel param, add default parameter `bool is_shape_agnositc = false` to `get_kernel_params` and propagate it as parameter of `get_default_params`

vladimir-paramuzov (approved):

PR #15800, opened 2023-02-18
[GPU] Fix OV_GPU_DumpGraphs option

tuxedcat:
### Details:
 - 

### Tickets:
 -


src/plugins/intel_gpu/src/graph/program.cpp
```diff
@@ -1260,6 +1260,10 @@ void program::dump_program(const char* stage,
                            bool with_full_info,
                            std::function<bool(program_node const&)> const& filter) const {
     std::string path = get_dir_path(_config);
+    GPU_DEBUG_GET_INSTANCE(debug_inst);
+    GPU_DEBUG_IF(debug_inst->dump_graphs.size()) {
+        path = debug_inst->dump_graphs;
+    }
```
isanghao, line 1266: can we make it happen with `ExecutionConfig::apply_debug_options` function? It has other config, too.

isanghao (changes requested):

src/plugins/intel_gpu/src/graph/program.cpp
tuxedcat: Used apply_user_properties() because apply_debug_options() is protected function.

isanghao (approved):
This would work. Let's merge this for now. However, I think better approach is to call `apply_debug_options` from `ExecutionConfig::set_default` because it is required for test case as well. How do you think? @vladimir-paramuzov

PR #15810, opened 2023-02-20
[AUTO] Failed to get the ov::model_name when loading network to target device is not ready

yangwang201911:
### Details:
 - Prioritize to get ov::model_name from CompiledModel(ExecutableNetwork) of acceleration device (CPU) since the target device CompiledModel(ExecutableNetwork) may not be ready.
 - Add the lock to guarantee the concurrency safe 

### Tickets:
 - 96206

### DIFF
--- a/src/plugins/auto/auto_executable_network.cpp
+++ b/src/plugins/auto/auto_executable_network.cpp
@@ -172,6 +172,7 @@ IE::Parameter AutoExecutableNetwork::GetMetric(const std::string& name) const {
         }
         return decltype(ov::available_devices)::value_type {exeDevices};
     } else if (name == ov::model_name) {
+        std::lock_guard<std::mutex> lock(_autoSContext->_confMutex);
         if (_autoSchedule->_loadContext[CPU].isEnabled && _autoSchedule->_loadContext[CPU].isAlready)
             return _autoSchedule->_loadContext[CPU].executableNetwork->GetMetric(name);
         return _autoSchedule->_loadContext[ACTUALDEVICE].executableNetwork->GetMetric(name);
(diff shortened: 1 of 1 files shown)


src/plugins/auto/auto_executable_network.cpp
```diff
@@ -172,6 +172,8 @@ IE::Parameter AutoExecutableNetwork::GetMetric(const std::string& name) const {
         }
         return decltype(ov::available_devices)::value_type {exeDevices};
     } else if (name == ov::model_name) {
+        if (_autoSchedule->_loadContext[CPU].isEnabled && _autoSchedule->_loadContext[CPU].isAlready)
+            return _autoSchedule->_loadContext[CPU].executableNetwork->GetMetric(name);
```
songbell, line 176: add lock and cases here

yangwang201911: add lock first and will implement the test case in another PR.

src/plugins/auto/auto_executable_network.cpp
```diff
@@ -172,6 +172,9 @@ IE::Parameter AutoExecutableNetwork::GetMetric(const std::string& name) const {
         }
         return decltype(ov::available_devices)::value_type {exeDevices};
     } else if (name == ov::model_name) {
+        std::lock_guard<std::mutex> lock(_autoSContext->_confMutex);
+        if (_autoSchedule->_loadContext[CPU].isEnabled && _autoSchedule->_loadContext[CPU].isAlready)
```
wangleis, line 176: Will merge this PR as hot fix.

Please use `execution_device` property directly to get property from underlying device in following PR.

wangleis (approved):

PR #15814, opened 2023-02-20
[PyOV] Shared memory improvements for Tensor and Constant classes

jiwaszki:
### Details:
 - Added new constructors for `Constant`.
 - New property `data` for `Constant` that reflects shared memory effects. Name and concept is borrowed from `Tensor` class.
 - Added test cases.
 - Cleaned common code and unified it.
 
### Tickets:
 - *96644*


src/bindings/python/src/pyopenvino/graph/ops/constant.cpp
```diff
@@ -27,6 +27,38 @@ std::vector<size_t> _get_byte_strides(const ov::Shape& s) {
     return byte_strides;
 }
 
+std::vector<size_t> _get_strides(const ov::op::v0::Constant& self) {
```
p-wysocki, line 30: Is it possible to rewrite this into a template?

jiwaszki: Not really, as the type (assume T in the template) is only known at the runtime after Constant creation. This is why I created this function instead of using `_get_byte_strides`.

src/bindings/python/src/pyopenvino/graph/ops/constant.cpp
```diff
@@ -151,4 +195,26 @@ void regclass_graph_op_Constant(py::module m) {
             throw std::runtime_error("Unsupported data type!");
         }
     });
+
+    constant.def_property_readonly(
+        "data",
+        [](ov::op::v0::Constant& self) {
+            auto ov_type = self.get_element_type();
+            auto dtype = Common::ov_type_to_dtype().at(ov_type);
+            if (ov_type.bitwidth() < 8) {
```
akuporos, line 204: may be move `8` to some const variable?
just not to have magic number

src/bindings/python/src/pyopenvino/core/common.hpp
```diff
@@ -28,14 +30,45 @@ const std::map<ov::element::Type, py::dtype>& ov_type_to_dtype();
...
 
-ov::Tensor tensor_from_pointer(py::array& array, const ov::Shape& shape, const ov::element::Type& ov_type);
+// Helpers for numpy arrays
+namespace array_helpers {
 
-ov::Tensor tensor_from_numpy(py::array& array, bool shared_memory);
+inline bool is_contiguous(const py::array& array);
 
-ov::PartialShape partial_shape_from_list(const py::list& shape);
+inline ov::element::Type get_ov_type(const py::array& array);
```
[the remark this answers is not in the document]
jiwaszki: This function is only defined one time and used internally in common part (under specific namespace)... or maybe I am missing a point. It does not matter after all, changed in the commit.

p-wysocki (approved):

akuporos (approved):

mvafin (approved):
Please create a ticket to remove API for pointer

PR #15815, opened 2023-02-20
[CPU] Fix issue about overwriting register names

xuchen-intel:
### Details:
 - *Fix issue about overwriting register names in https://github.com/openvinotoolkit/openvino/pull/15423.*
### Tickets:
 - *101222*

### DIFF
--- a/src/plugins/intel_cpu/src/emitters/utils.cpp
+++ b/src/plugins/intel_cpu/src/emitters/utils.cpp
@@ -27,39 +27,41 @@ template void RegPrinter::print<unsigned char, Reg16>(jit_generator &h, Reg16 re
 template void RegPrinter::print<char, Reg8>(jit_generator &h, Reg8 reg, const char *name);
 template void RegPrinter::print<unsigned char, Reg8>(jit_generator &h, Reg8 reg, const char *name);
 
-void RegPrinter::print_reg_fp32(const char *name, int val) {
-    std::stringstream ss;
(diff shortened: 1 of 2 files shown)


src/plugins/intel_cpu/src/emitters/utils.cpp
```diff
@@ -131,16 +131,19 @@ const char * RegPrinter::get_name(REG_T reg, const char *name) {
     if (name == nullptr) {
         return reg_name;
     } else {
+        constexpr size_t cnt = 1024;
```
chenhu-wang, line 134: It seems like 1024 is an experimental or guess number to avoid flush. I am wondering if it is possible to use an shared mutex across all print,  lock inside print function to serialize each print pipeline. Until previous print finish, we start new print.

xuchen-intel: Thanks Chenhu for the commet! I've tried to use mutex mechanism inside print function (also tried that in the scope of h.call and get_name functions), it doesn't fix the issue. It seems not an issue of synchronism. Further discussion is always welcomed.

chenhu-wang:
functionally LGTM.

chenhu-wang (approved):

xuchen-intel:
@dmitry-gorokhov Hi Dmitry! Updated the PR according to our offline dicussion. Please feel free to comment. Thanks!

dmitry-gorokhov (approved):

PR #15822, opened 2023-02-20
Add use_legacy_frontend option into layer tests

dkrupnov:
### Details:
 - add use_legacy_frontend for layer tests


tests/layer_tests/conftest.py
```diff
@@ -89,6 +94,12 @@ def use_new_frontend(request):
     return request.config.getoption('use_new_frontend')
 
 
+@pytest.fixture(scope="session")
+def use_legacy_frontend(request):
+    """Fixture function for command-line option."""
+    return request.config.getoption('use_legacy_frontend')
```
rnugmanx, line 100: How it will go through tests?
Now I see that "if":
https://github.com/openvinotoolkit/openvino/blob/master/tests/layer_tests/common/layer_test_class.py#L57

It says that we will use legacy fe only in case of use_new_frontend=False

dkrupnov: in configs we determine which option will be used - use_new_frontend or use_legacy_frontend

tests/layer_tests/conftest.py
```diff
@@ -89,6 +94,12 @@ def use_new_frontend(request):
     return request.config.getoption('use_new_frontend')
```
rkazants, line 96: @dkrupnov, can you also please correct files:
https://github.com/openvinotoolkit/openvino/blob/master/tests/layer_tests/common/layer_test_class.py#L57
https://github.com/openvinotoolkit/openvino/blob/master/.ci/azure/linux.yml#L509 (add `use_legacy_frontend`)
https://github.com/openvinotoolkit/openvino/blob/master/.ci/azure/linux.yml#L516 (add `use_legacy_frontend`)

PR #15827, opened 2023-02-20
PushConstantToSubgraph - use vector<bool> for remove_inputs_mask

mateusztabaka:
When MultiSubGraphOp has more than 32 inputs, int is not enough to track inputs to be removed.

Ticket: 103248


mvafin (approved):
This resolves the issue. Looks good to me.

itikhono (approved):

PR #15828, opened 2023-02-20
【PaddlePaddle Hackathon 4】add paddle flip op

Asthestarsfalll:
### Details:
add flip operation in Paddle front end

> Reverse the order of a n-D tensor along given axis in axis.

### Tickets:
 - *ticket-id*

### Reference:
https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/flip_en.html
https://github.com/PaddlePaddle/Paddle/blob/release/2.4/python/paddle/tensor/manipulation.py#L1327
https://github.com/PaddlePaddle/Paddle2ONNX/blob/develop/paddle2onnx/legacy/op_mapper/tensor.py#L2018


Asthestarsfalll:
Sorry, I am not familiar with rebase -_-

Asthestarsfalll:
@ceciliapeng2011 Hi, Al CI have been passed, Could you review this PR?

src/frontends/paddle/src/op/flip.cpp
```diff
@@ -0,0 +1,44 @@
...
+namespace paddle {
+namespace op {
+NamedOutputs flip(const NodeContext& node) {
+    const auto data_node = node.get_input("X");
+    const auto axes = node.get_attribute<std::vector<int32_t>>("axis");
+    const auto input_shape = data_node.get_partial_shape().get_shape();
+    const auto dims = static_cast<int32_t>(data_node.get_partial_shape().rank().get_length());
+    const auto dtype = data_node.get_element_type();
+    // for zero-dim input
+    PADDLE_OP_CHECK(node, (dims > 0), "Input dims must be greater than 0");
```
ceciliapeng2011, line 19: Is this restriction asked by paddlepaddle or this op conversion? Would you please try to build a scaler "X" input model? If     we cannot create such a paddle model, we can remove this check here.

src/frontends/paddle/src/op/flip.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "openvino/frontend/paddle/node_context.hpp"
+
+namespace ov {
+namespace frontend {
+namespace paddle {
+namespace op {
+NamedOutputs flip(const NodeContext& node) {
+    const auto data_node = node.get_input("X");
+    const auto axes = node.get_attribute<std::vector<int32_t>>("axis");
+    const auto input_shape = data_node.get_partial_shape().get_shape();
```
ceciliapeng2011, line 15: If it is a dynamic shape model (i.e. some dimension in the input shape is unknown until inference time), the solution is not feasible, and there would be an exception to call get_shape().
So, could you please -
1. add this static shape check, as the first step.
2. the second step, to investigate how to support dynamic shape input.

src/frontends/paddle/tests/test_models/gen_scripts/generate_flip.py
```diff
@@ -0,0 +1,60 @@
...
+    data_type = 'int64'
+    axis = 1
+    x = np.random.randint(-1, 3, (5, 3, 1, 1)).astype(data_type)
+    flip("flip_4", x, axis)
+
+    data_type = 'float32'
+    axis = -1
+    x = np.random.randn(1).astype(data_type)
+    flip("flip_5", x, axis)
+
```
ceciliapeng2011, line 57: Can have "bool" type tested?

src/frontends/paddle/src/op/flip.cpp
Asthestarsfalll: Use `VariadicSplit` will  be able to support dynamic shape. I will update after passing tests.

src/frontends/paddle/tests/test_models/gen_scripts/generate_flip.py
Asthestarsfalll: I tested bool type before, but it occurs error that openvino test do not support bool output

src/frontends/paddle/src/op/flip.cpp
Asthestarsfalll: Paddle flip do not support zero-dim input, I will remove this.

src/frontends/paddle/src/op/flip.cpp
Asthestarsfalll: Hi, I think it hard to support dynamic shape input, cause we need to split tensor into  several ones which have one dimension in specified axis. 
But if there has some op supports automatically spliting tensor into one-dim-tensor, instead of specifing split numbers(for example, the distinction of torch.split and torch.chunk). It will become eazier.

mvafin: As far as I can see, you really need shape dimensions to be static only on `axes` dimensions. I recommend removing `get_shape()` call here and instead call `axes[idx].get_length()` in the loop below where needed. Also please add `PADDLE_OP_CHECK` for `axes[idx].is_static()` for each call of `get_length()`.

To support dynamic shape on axis you could try using `Range` with -1 step to get indices and `Gather` on that indices

src/frontends/paddle/src/op/flip.cpp
```diff
@@ -0,0 +1,44 @@
...
+    const auto dims = static_cast<int32_t>(data_node.get_partial_shape().rank().get_length());
+    const auto dtype = data_node.get_element_type();
+    // for zero-dim input
+    PADDLE_OP_CHECK(node, (dims > 0), "Input dims must be greater than 0");
+
+    Output<Node> temp = data_node;
+    std::vector<Output<Node>> temp_split_out;
+    int32_t axis;
+    for (size_t idx = 0; idx < axes.size(); idx++) {
```
itikhono, line 24: is it possible to use Reverse or ReverseSequence openvino operators?

rkazants: Exactly, so we do not need to manipulate with input shape and support this operation for dynamic shape input as well

Asthestarsfalll: > Exactly, in this case we do not need to manipulate with input shape and we will be able to support this operation for dynamic shape input as well

Hi, I find that `flip` is the same as [reverse](https://github.com/openvinotoolkit/openvino/blob/master/src/frontends/paddle/src/op/reverse.cpp).
So do I still need to add flip?

rkazants: Sure, just to re-implement the translator using `ReverseSequence`. `Reverse` looks deprecated some time ago: https://docs.openvino.ai/latest/openvino_docs_ops_opset10.html#doxid-openvino-docs-ops-opset10

Asthestarsfalll: Hi, I have update the codes, but I found it's still hard to support dynamic shape input.
Because `ReverseSequence` need the  length to reverse in specified seq_axis.
This requires input shape to being static.

src/frontends/paddle/src/op/flip.cpp
```diff
@@ -0,0 +1,46 @@
...
+        is_1dim = true;
+    }
+    for (size_t idx = 0; idx < axes.size(); idx++) {
+        axis = axes[idx];
+        axis += axis < 0 ? dims : 0;
+        shape_len = input_shape[axis].get_length();
+        batch_index = axis == 0 ? 1 : 0;
+        auto batch_len = input_shape[batch_index].get_length();
+        const auto seq_length =
+            default_opset::Constant::create(element::i64, {static_cast<uint32_t>(batch_len)}, {shape_len});
```
mvafin, line 35: You do not really need to restrict static shape on axis. You could get the seq_len by something like this: `X->ShapeOf->Gather(axis)\Gather(batch_index)->Broadcast()`

src/frontends/paddle/src/op/flip.cpp
```diff
@@ -0,0 +1,46 @@
...
+namespace paddle {
+namespace op {
+NamedOutputs flip(const NodeContext& node) {
+    const auto data_node = node.get_input("X");
+    const auto axes = node.get_attribute<std::vector<int32_t>>("axis");
+    auto input_shape = data_node.get_partial_shape();
+    PADDLE_OP_CHECK(node, (input_shape.rank().is_static()), "flip not support dynamic rank!");
+    const auto dims = static_cast<int32_t>(data_node.get_partial_shape().rank().get_length());
+    const auto dtype = data_node.get_element_type();
+    bool is_1dim = false;
```
mvafin, line 19:
```suggestion
    bool is_1dim = dims == 1;
```

src/frontends/paddle/src/op/flip.cpp
```diff
@@ -0,0 +1,46 @@
...
+    int32_t axis, shape_len, batch_index;
+    Output<Node> temp = data_node;
+    const auto uns_axes = default_opset::Constant::create(ov::element::i64, {1}, {1});
+    if (dims == 1) {
```
mvafin, line 23:
```suggestion
    if (is_1dim) {
```

src/frontends/paddle/src/op/flip.cpp
```diff
@@ -0,0 +1,46 @@
...
+        temp = std::make_shared<default_opset::Unsqueeze>(temp, uns_axes);
+        input_shape = temp.get_partial_shape();
+        is_1dim = true;
```
mvafin, line 26:
```suggestion
```

mvafin (changes requested):

mvafin:
This conversion seems overcomplicated with `ReverseSequence`. You could also look into using `Slice` with step -1 I think it should be simpler.

Asthestarsfalll:
> This conversion seems overcomplicated with `ReverseSequence`. You could also look into using `Slice` with `step=-1`, `start=-1`, `stop=-INT_MIN`, `axes=axes`. I think it should be simpler.

Yes! `Slice` is much simpler.
Test results:

rkazants (approved):
One comment left. Other LGTM.

src/frontends/paddle/src/op/flip.cpp
```diff
@@ -0,0 +1,37 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "default_opset.hpp"
+#include "openvino/frontend/paddle/node_context.hpp"
+
+#define out(a) std::cout << a << std::endl;
```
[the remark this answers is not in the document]
Asthestarsfalll: done.

mvafin (approved):
Looks good from my side, but would like to see approve from @ceciliapeng2011

Asthestarsfalll:
@ceciliapeng2011 PTAL

Asthestarsfalll:
@ceciliapeng2011 PTAL

Asthestarsfalll:
@OpenVINO-dev-contest PTAL

yuxu42:
@ceciliapeng2011 could you please review?

alanzhai219:
I have a question here. `flip` op is an alias of `reverse`, right? If so, why not try `Reverse` to map it?
refer to https://github.com/openvinotoolkit/openvino/pull/17525.

xczhai:
@Asthestarsfalll 
I have a question here. `flip` is an alias of `reverse`, right? If so, why not map it using `Reverse` IR.
refer to https://github.com/openvinotoolkit/openvino/pull/17525

itikhono:
> @Asthestarsfalll I have a question here. `flip` is an alias of `reverse`, right? If so, why not map it using `Reverse` IR. refer to #17525

As I can see, Reverse op was deleted in opset2. We don't have this operator in opset2+ 
It's better to fix `reverse` op convertor (translator) to use ReverseSequence op or similar approach like in this PR if possible

itikhono (approved):

alanzhai219:
> > @Asthestarsfalll I have a question here. `flip` is an alias of `reverse`, right? If so, why not map it using `Reverse` IR. refer to #17525
> 
> As I can see, Reverse op was deleted in opset2. We don't have this operator in opset2+ It's better to fix `reverse` op convertor (translator) to use ReverseSequence op or similar approach like in this PR if possible

PD `reverse` op executes the same operation. It is better to unify the two mappings instead of two separate implement.

xczhai:
> > @Asthestarsfalll I have a question here. `flip` is an alias of `reverse`, right? If so, why not map it using `Reverse` IR. refer to #17525
> 
> As I can see, Reverse op was deleted in opset2. We don't have this operator in opset2+ It's better to fix `reverse` op convertor (translator) to use ReverseSequence op or similar approach like in this PR if possible

PD reverse op executes the same operation. It is better to unify the two mappings instead of two separate implement.

itikhono:
> > > @Asthestarsfalll I have a question here. `flip` is an alias of `reverse`, right? If so, why not map it using `Reverse` IR. refer to #17525
> > 
> > 
> > As I can see, Reverse op was deleted in opset2. We don't have this operator in opset2+ It's better to fix `reverse` op convertor (translator) to use ReverseSequence op or similar approach like in this PR if possible
> 
> PD reverse op executes the same operation. It is better to unify the two mappings instead of two separate implement.

Yes, I agree. But the current pd reverse op translator uses OV Reverse op that was deleted from officially supported opsets (from opset2 and later). We need to fix reverse op translator in this case

xczhai:
> > > > @Asthestarsfalll I have a question here. `flip` is an alias of `reverse`, right? If so, why not map it using `Reverse` IR. refer to #17525
> > > 
> > > 
> > > As I can see, Reverse op was deleted in opset2. We don't have this operator in opset2+ It's better to fix `reverse` op convertor (translator) to use ReverseSequence op or similar approach like in this PR if possible
> > 
> > 
> > PD reverse op executes the same operation. It is better to unify the two mappings instead of two separate implement.
> 
> Yes, I agree. But the current pd reverse op translator uses OV Reverse op that was deleted from officially supported opsets (from opset2 and later). We need to fix reverse op translator in this case

@Asthestarsfalll Hi, it is much appreciated if you can help unify `reverse` and `flip` implements together.

Asthestarsfalll:
> @Asthestarsfalll Hi, it is much appreciated if you can help unify `reverse` and `flip` implements together.

I will update code latter.

ceciliapeng2011 (approved):

ceciliapeng2011 (approved):

PR #15830, opened 2023-02-20
[PyOV] Fix deprecation warning in tests

akuporos:
### Details:
 - Some overloadings of `register_pass` are deprecated and should be removed in this release, so I updated tests
 -  17 warnings -> 0 warnings for `tests` folder.
 -  24 warnings -> 0 warnings for `tests_compatibility` folder.

### Tickets:
 - 104388

### DIFF
--- a/src/bindings/python/tests/test_graph/test_manager.py
+++ b/src/bindings/python/tests/test_graph/test_manager.py
@@ -8,7 +8,7 @@ import os
 import numpy as np
 import pytest
 
-import openvino.runtime.opset8 as ops
+import openvino.runtime.opset10 as ops
 from openvino.runtime import Core, Model
 from openvino.runtime.passes import Manager, Serialize, ConstantFolding, Version
 from tests.test_graph.util import count_ops_of_type
--- a/src/bindings/python/tests/test_graph/test_manager.py
+++ b/src/bindings/python/tests/test_graph/test_manager.py
@@ -8,17 +8,26 @@ import os
 import numpy as np
 import pytest
 
-import openvino.runtime.opset8 as ov
-from openvino.runtime import Model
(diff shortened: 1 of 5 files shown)


src/bindings/python/tests/test_graph/test_manager.py
```diff
@@ -8,17 +8,26 @@
 import numpy as np
 import pytest
 
-import openvino.runtime.opset8 as ov
-from openvino.runtime import Model
+import openvino.runtime.opset8 as ops
+from openvino.runtime import Core, Model
```
jiwaszki, line 12:
```suggestion
import openvino.runtime.opset10 as ops
```
Bump the opset?

jiwaszki (approved):

src/bindings/python/tests/test_graph/test_manager.py
p-wysocki: I don't think we have opset11 yet https://github.com/openvinotoolkit/openvino/tree/master/src/core/include/openvino/opsets

p-wysocki (approved):

src/bindings/python/tests/test_graph/test_manager.py
akuporos: good catch!

src/bindings/python/tests/test_graph/test_manager.py
```diff
@@ -8,17 +8,26 @@
 import numpy as np
 import pytest
 
-import openvino.runtime.opset8 as ov
-from openvino.runtime import Model
+import openvino.runtime.opset8 as ops
```
akuporos, line 11:
```suggestion
import openvino.runtime.opset10 as ops
```

PR #15835, opened 2023-02-21
【PaddlePaddle Hackathon 4】add paddle linspace op

Patrick-Star125:
### Details:
add linspace operation in Paddle front end

### Reference:
- https://www.paddlepaddle.org.cn/documentation/docs/en/2.2/api/paddle/linspace_en.html#linspace
- https://github.com/PaddlePaddle/Paddle2ONNX/blob/develop/paddle2onnx/legacy/op_mapper/tensor.py#L1384

### Unit-Test
image


src/frontends/paddle/tests/test_models/gen_scripts/generate_linspace.py
```diff
@@ -0,0 +1,62 @@
...
+                  outputs=[outs[0]], target_dir=sys.argv[1])
+
+    return outs[0]
+
+
+def main():
+    start = random.randint(0, 10)
+    stop = random.randint(0, 10)
+    num = random.randint(1, 10)
+    linspace("linspace_1", start, stop, num, "int32")
```
ceciliapeng2011, line 48: Is it allowed if start is smaller than stop?

src/frontends/paddle/tests/test_models/gen_scripts/generate_linspace.py
```diff
@@ -0,0 +1,62 @@
...
+
+    start = random.uniform(0, 10)
+    stop = random.uniform(0, 10)
+    num = random.randint(1, 10)
+    linspace("linspace_2", start, stop, num, "float32")
+
```
ceciliapeng2011, line 54: start and stop could be tensor of shape rank 1. Is there any difference in the serialized protobuf model w.r.t. a scaler input? If it is, please add a test case when the input is a tensor.

ceciliapeng2011: and do you cover/consider the cases when num is 1 or 0?

ceciliapeng2011 (changes requested):

Patrick-Star125:
Hi, the source codes of paddle linspace API illustrate that the input scalar(int/float) will be transferred in to tensor.

image

I wonder if it is needed to provide the supporting of scalar input on linspace.cpp. Same question about unit-test samples also exist.

src/frontends/paddle/src/op/linspace.cpp
```diff
@@ -0,0 +1,38 @@
...
+    auto const_one = std::make_shared<default_opset::Constant>(start.get_element_type(), Shape{}, 1);
+    Output<Node> num_minus_one = std::make_shared<default_opset::Subtract>(num, const_one);
+    num_minus_one = std::make_shared<default_opset::Convert>(num_minus_one, start.get_element_type());
+    Output<Node> delta = std::make_shared<default_opset::Subtract>(stop, start);
+    delta = std::make_shared<default_opset::Divide>(delta, num_minus_one);
+
+    // generate a range of numbers [0, 1, ..., num)
+    auto const_zero = std::make_shared<default_opset::Constant>(start.get_element_type(), Shape{}, 0); 
+    auto const_num = std::make_shared<default_opset::Squeeze>(num);
+    auto range0_n = std::make_shared<default_opset::Range>(const_zero, const_num, const_one, start.get_element_type());
```
OpenVINO-dev-contest, line 28: I just wonder can we use ```range``` for a direct output ? so we do not need ```Multiply``` and ```Add```, just a guess

src/frontends/paddle/tests/test_models/gen_scripts/generate_linspace.py
Patrick-Star125: test sample has been added

src/frontends/paddle/tests/test_models/gen_scripts/generate_linspace.py
Patrick-Star125: test sample has been added, but num=0 is not permitted by paddle.

float64 is not supported yet, I am working on this error.
image
I am a bit confused because float64 would have been converted to float32 in the first part of my code.

src/frontends/paddle/src/op/linspace.cpp
Patrick-Star125: great idea, it is important to simplify the operator code. but the behavior of `Range` is subtly different from `Linspace` even though their input parameters are similar with each other. Using `Range` for a direct output maybe require the overal re-design of `linspace`.

OpenVINO-dev-contest: > subtly

Thanks, could you clarify what's gap between them? can we convert the ```num``` in Linspace into ```step``` in Range

OpenVINO-dev-contest: Thanks, could you clarify what's gap between them? can we convert the num in Linspace into step in Range

Patrick-Star125: `step` can be calculated by `num`, which has already implemented in line 30 of `linspace.cpp`. The key point is the behavior of `Range`. For instance, same input (start: [0] stop: [1] num: [4]/step: [0.33]), and the result will be:
`Range`: [0, 0.33, 0.66]
`Linspace`: [0, 0.33, 0.66, 1]
And the input of Range must be scalar, which means <Squeeze> must be taken before range operator.
image
I think it will become more complicate after adjusting the code.

OpenVINO-dev-contest: Great thanks for your illustration, since its a advanced task or our target is to find out a solution with best inference performance. Could you help to try both solution with [benchmark_app](https://github.com/openvinotoolkit/openvino/tree/master/tools/benchmark_tool) to compare their performance ?
any suggestion from your perspective @ceciliapeng2011 ?

OpenVINO-dev-contest: Can ```start``` and ```stop``` be negative here, and can you put this example ( (start: [0] stop: [1] num: [4]/step: [0.33])) into your test case, thanks?

Patrick-Star125: OK，test samples had been added

ceciliapeng2011 (changes requested):
Would you please rebase and resolve the conflicts?

Patrick-Star125:
conflicts resoved

OpenVINO-dev-contest:
> ### Details:
> add linspace operation in Paddle front end
> 
> ### Reference:
> * https://www.paddlepaddle.org.cn/documentation/docs/zh/api/paddle/linspace_cn.html#linspace
> * https://github.com/PaddlePaddle/Paddle2ONNX/blob/develop/paddle2onnx/legacy/op_mapper/tensor.py#L1384

Hello could you also post a English version of Paddle Op description, thanks

src/frontends/paddle/src/op/linspace.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "default_opset.hpp"
+#include "openvino/frontend/paddle/node_context.hpp"
+#include "openvino/frontend/paddle/visibility.hpp"
+
+namespace ov {
+namespace frontend {
+namespace paddle {
+namespace op {
+NamedOutputs linspace(const NodeContext& node) {
+    auto start = node.get_input("Start");
```
OpenVINO-dev-contest, line 14: How about the inputs are variables, but not a tensor. e.g int or float
image

Patrick-Star125:
> Hi, the source codes of paddle linspace API illustrate that the input scalar(int/float) will be transferred in to tensor.
> 
> 
> 
> I wonder if it is needed to provide the supporting of scalar input on linspace.cpp. Same question about unit-test samples also exist.

Hi, like I metioned before, all scalars will be transformed into tensor in advance, the [source code](https://github.com/PaddlePaddle/Paddle/blob/d8b8c2d8326a63adca349a32d6a4c97c8c0a3e7e/paddle/fluid/operators/linspace_op.cc#L64) from paddle can prove that the only attribute input is 'dtype'.

OpenVINO-dev-contest:
> > Hi, the source codes of paddle linspace API illustrate that the input scalar(int/float) will be transferred in to tensor.
> > 
> > I wonder if it is needed to provide the supporting of scalar input on linspace.cpp. Same question about unit-test samples also exist.
> 
> Hi, like I metioned before, all scalars will be transformed into tensor in advance, the [source code](https://github.com/PaddlePaddle/Paddle/blob/d8b8c2d8326a63adca349a32d6a4c97c8c0a3e7e/paddle/fluid/operators/linspace_op.cc#L64) from paddle can prove that the only attribute input is 'dtype'.

Sorry I missed it. but shall we consider the ```dtype``` is not in default? 

> > Hi, the source codes of paddle linspace API illustrate that the input scalar(int/float) will be transferred in to tensor.
> > 
> > I wonder if it is needed to provide the supporting of scalar input on linspace.cpp. Same question about unit-test samples also exist.
> 
> Hi, like I metioned before, all scalars will be transformed into tensor in advance, the [source code](https://github.com/PaddlePaddle/Paddle/blob/d8b8c2d8326a63adca349a32d6a4c97c8c0a3e7e/paddle/fluid/operators/linspace_op.cc#L64) from paddle can prove that the only attribute input is 'dtype'.

Sorry, I missed it, but shall we consider the ```dtype``` when it is not in default in some cases. for example input type is fp32 and dtype is int32. I don't know whether it may happen ?

Patrick-Star125:
> Sorry, I missed it, but shall we consider the dtype when it is not in default in some cases. for example input type is fp32 and dtype is int32. I don't know whether it may happen ?

The python frontend of paddle could ensure that the input type and attr `dtype` is same, so I think there is no need for extra examination or adjustment. [related code](https://github.com/PaddlePaddle/Paddle/blob/release/2.2/python/paddle/fluid/layers/tensor.py#L1512)

OpenVINO-dev-contest:
> The python frontend of paddle could ensure that the input type and attr `dtype` is same, so I think there is no need for extra examination or adjustment. [related code](https://github.com/PaddlePaddle/Paddle/blob/release/2.2/python/paddle/fluid/layers/tensor.py#L1512)

Can you update the uni-test with these case example ? for example input type is fp32 and dtype is int32. just ensure the Paddle front end will do conversion for us.

Patrick-Star125:
> Can you update the uni-test with these case example ? for example input type is fp32 and dtype is int32. just ensure the Paddle front end will do conversion for us.

sorry, I misunderstood the function of `dtype`, adjustment on code and test samples have been conducted now.

src/frontends/paddle/src/op/linspace.cpp
```diff
@@ -0,0 +1,57 @@
...
+    auto stop = node.get_input("Stop");
+    auto num = node.get_input("Num");
+    std::unordered_map<int, element::Type> umap{{2, element::i32}, {3, element::i64}, {5, element::f32}};
+    const auto dtype = node.get_attribute<int>("dtype", 5);
```
OpenVINO-dev-contest, line 18: can we use ```auto dtype = node.get_attribute<ov::element::Type>("dtype");``` to get the the type ?

Patrick-Star125: Done

OpenVINO-dev-contest: Pls add a screenshot of test results, thanks

OpenVINO-dev-contest (approved):

ceciliapeng2011 (approved):

PR #15837, opened 2023-02-21
[Snippets] Added broadcast support for FQ weights calculation

a-sidorova:
### Details:
 - *Fixed broadcasting in FQ decomposition. Saved previous calculations for cases with the same shapes or scalar. Otherwise refence is used for broadcasting calculations*

### Tickets:
 - *100808*


a-sidorova:
@v-Golubev may I ask you to review the changes? Thank you!

src/common/snippets/src/pass/fq_decomposition.cpp
```diff
@@ -205,48 +207,100 @@ bool ngraph::snippets::pass::FakeQuantizeDecomposition::getScalesAndShifts(
     if (!input_low_constant || !input_high_constant || !output_low_constant || !output_high_constant)
         return false;
 
+    const auto input_low_pshape = input_low_constant->get_output_partial_shape(0);
+    const auto input_high_pshape = input_high_constant->get_output_partial_shape(0);
+    const auto output_low_pshape = output_low_constant->get_output_partial_shape(0);
+    const auto output_high_pshape = output_high_constant->get_output_partial_shape(0);
```
v-Golubev, line 213: These nodes are Constants so we can safely get their `Shape` instead of `PartialShape` and avoid additional conversions below
```suggestion
    const auto input_low_shape = input_low_constant->get_shape();
    const auto input_high_shape = input_high_constant->get_shape();
    const auto output_low_shape = output_low_constant->get_shape();
    const auto output_high_shape = output_high_constant->get_shape();
```

src/common/snippets/src/pass/fq_decomposition.cpp
```diff
@@ -205,48 +207,100 @@ bool ngraph::snippets::pass::FakeQuantizeDecomposition::getScalesAndShifts(
...
+        for (size_t i = 0; i < input_size; i++) {
+            float il = input_low[input_low.size() == 1 ? 0 : i];
+            float ih = input_high[input_high.size() == 1 ? 0 : i];
+
+            isc[i] = (levels - 1) / (ih - il);
+            ish[i] = -il * isc[i];
+        }
+        cl = input_low;
+        ch = input_high;
+    } else {  // NUMPY broadcasting
```
v-Golubev, line 237: looks like semantically we can always use NUMPY broadcasting. Why do we leave legacy computation logic for some specific cases?
The same question about output scales/shifts computation

src/common/snippets/src/pass/fq_decomposition.cpp
a-sidorova: Sure, done. Thanks!

src/common/snippets/src/pass/fq_decomposition.cpp
a-sidorova: Because `ngraph::runtime::reference::autobroadcast_binop()` is expensive call. I'd like to not increase model compilation time.
Moreover, the case "NUMPY broadcasting" is rare in real models. I mean the situation when weights of FQ are scalar or have equal shapes is base - I haven't seen models with first case, only in test cases. So for that case we have optimal calculations. For other rare cases - common computations. Thank you!

v-Golubev: Ok, thanks for the explanation. Let's leave this logic as is.

v-Golubev (approved):
LGTM

a-sidorova:
@dmitry-gorokhov take a look please. It should fix sporadic failed tests on master as well

src/common/snippets/src/pass/fq_decomposition.cpp
a-sidorova: Added in 487198f95d917a0cddf520341a60b6ebfd2b687d

maxnick (approved):

dmitry-gorokhov (approved):

PR #15840, opened 2023-02-21
【PaddlePaddle Hackathon 4】add paddle silu op

AndPuQing:
### Details:

$$
silu=\frac{x}{1+e^{-x}}
$$

add silu op in Paddle front end

### Tickets:
 - *ticket-id*

### Reference:

https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/nn/functional/silu_en.html


OpenVINO-dev-contest:
hi @AndPuQing since there is no test screen shot in this PR, so we will prioritize the review on another one: https://github.com/openvinotoolkit/openvino/pull/15874

Any comments ?

AndPuQing:
> hi @AndPuQing since there is no test screen shot in this PR, so we will prioritize the review on another one: #15874
> 
> Any comments ?

Ok, it seems its implementation is better. I will close the PR.

PR #15842, opened 2023-02-21
[GPU] Fallback to kernel caching in the case of dynamic models

e-ddykim:
### Details:
 - For dynamic shape models, kernel caching is enabled instead of model caching.
 - Now, model caching is the default caching policy instead of kernel caching.

### Tickets:
 - 97997

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/plugin/plugin.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/plugin/plugin.hpp
@@ -18,7 +18,7 @@ namespace intel_gpu {
 class Plugin : public InferenceEngine::IInferencePlugin {
     struct impl;
     std::shared_ptr<impl> _impl;
-    bool isModelCachingEnabled = false;
+    bool isModelCachingEnabled = true;
 
     std::string default_device_id = "0";
     // key: device_id, value: cldnn device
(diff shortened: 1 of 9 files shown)


yeonbok (approved):

src/plugins/intel_gpu/include/intel_gpu/plugin/plugin.hpp
```diff
@@ -18,7 +18,7 @@ namespace intel_gpu {
 class Plugin : public InferenceEngine::IInferencePlugin {
     struct impl;
     std::shared_ptr<impl> _impl;
-    bool isModelCachingEnabled = false;
+    bool isModelCachingEnabled = true;
```
vladimir-paramuzov, line 21: Seems like current usage of this flag is incorrect. The flag is stored in the plugin instance, but it depends on the last loaded IR. So I see two issues at least:
1. Not thread safe. Seems like modification of the flag is not guarded with mutex, thus parallel call of compile_model will cause data race
2. Plugin response depends on the models loading sequence:
```cpp
core.compile_model(static_model);
core.compile_model(dynamic_model);
core.get_property("GPU", ov::device::capabilities); // doesn't have IMPORT_EXPORT

core.compile_model(dynamic_model);
core.compile_model(static_model);
core.get_property("GPU", ov::device::capabilities); // have IMPORT_EXPORT
```

If caching for dynamic models can't be enabled quickly, I think we can serialize IR after transformations for dynamic models (similar to CPU plugin behavior) and import will load that IR.

e-ddykim: Thank you for your review. I'll fix it in another PR.

PR #15845, opened 2023-02-21
【PaddlePaddle Hackathon 4】add paddle softshrink op

AndPuQing:
### Details:

add softshrink op in paddle front end

$$
softshrink(x)=\left\{\begin{aligned}
x-\text { threshold, } & \text { if } x>\text { threshold } \\
x+\text { threshold, } & \text { if } x<-\text { threshold } \\
0, & \text { otherwise }
\end{aligned}\right.
$$

### Tickets:
 - *ticket-id*

### Reference

https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/nn/functional/softshrink_en.html

image


src/frontends/paddle/src/op/softshrink.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+    std::shared_ptr<ngraph::Node> adjusted_inputs =
+        default_opset::Constant::create(input_element_type, Shape{data.get_shape()}, {0});
+    std::shared_ptr<ngraph::Node> adjusted_inputs_below_neg_lambda =
+        std::make_shared<default_opset::Multiply>(std::make_shared<default_opset::Add>(data, positive_lambda),
+                                                  values_below_neg_lambda);
+    std::shared_ptr<ngraph::Node> adjusted_inputs_above_pos_lambda =
+        std::make_shared<default_opset::Multiply>(std::make_shared<default_opset::Add>(data, negative_lambda),
+                                                  values_above_pos_lambda);
+    adjusted_inputs = std::make_shared<default_opset::Add>(adjusted_inputs, adjusted_inputs_below_neg_lambda);
```
OpenVINO-dev-contest, line 39: I just wonder can we use ```select``` here ? 
https://github.com/openvinotoolkit/openvino/blob/master/docs/ops/condition/Select_1.md

AndPuQing: Yes, it can be implemented using select, which looks more concise than the original implementation, but I wonder if matrix multiplication will be more efficient than select ?

OpenVINO-dev-contest: My suggestion is to use benchmark_app to test the performance of these 2 single node graph, and find out a best implementation ?

src/frontends/paddle/src/op/softshrink.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "default_opset.hpp"
+#include "openvino/frontend/paddle/node_context.hpp"
+namespace ov {
+namespace frontend {
+namespace paddle {
+namespace op {
+NamedOutputs softshrink(const NodeContext& node) {
+    auto data = node.get_input("X");
+    auto lambda = node.get_attribute<float>("lambda");
```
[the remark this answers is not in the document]
AndPuQing: According to paddle source code  https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/nn/functional/activation.py#L1210 
I believe that threshold always had a value, but to address potential risks, a default value was still set and unit tests were created."

src/frontends/paddle/src/op/softshrink.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+    const auto input_element_type = data.get_element_type();
+    const auto positive_lambda = default_opset::Constant::create(input_element_type, Shape{}, {lambda});
+    const auto negative_lambda = default_opset::Constant::create(input_element_type, Shape{}, {-lambda});
```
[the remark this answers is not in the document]
AndPuQing: Actually, I don't think the input x will be unsigned. It seems that Paddle doesn't have a corresponding type of tensor and there is no corresponding unit test.

It may not be necessary to use a common lambda since we need to compare x with negative lambda in order to use it.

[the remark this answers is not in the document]
liubo-intel: > Actually, I don't think the input x will be unsigned. It seems that Paddle doesn't have a corresponding type of tensor and there is no corresponding unit test.
Hi, @AndPuQing : do you have more detail information about  'Paddle doesn't have a corresponding type of tensor' you mentioned? because I think only by 'no corresponding unit test' information, we can't draw the conclusion that there will be no unsigned type cases for this op.
or maybe we can use 'PADDLE_OP_CHECK' to check 'input_element_type.is_signed()' condition and leave the code only handle the signed cases if we are not quite sure about its data type ranges.
> It may not be necessary to use a common lambda since we need to compare x with negative lambda in order to use it.
yes, common lambda is used for handle unsigned type cases. if none such type, 'common lambda' is not necessary here.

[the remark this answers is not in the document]
AndPuQing: Well, for the potential paddle to register the unsighed type of kernel, I have made a special treatment for the unsighed type, only select the part of x>lambda, and replace the part of x<lambda with 0

liubo-intel (approved):

OpenVINO-dev-contest (approved):

src/frontends/paddle/src/op/softshrink.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+        output = std::make_shared<default_opset::Select>(values_above_pos_lambda, negative_node, data);
+        output = std::make_shared<default_opset::Select>(values_below_neg_lambda, positive_node, output);
+
+        std::shared_ptr<ngraph::Node> zero_mask =
+            std::make_shared<default_opset::LogicalOr>(values_below_neg_lambda, values_above_pos_lambda);
+
+        output = std::make_shared<default_opset::Select>(zero_mask, output, zero_node);
+    } else {
+        // Passing -lambd to unsigned type constant will cause an overflow.
```
ceciliapeng2011, line 41: Refer to Paddle spec, input x could be float32 and float64 only. How can a test case trigger this else branch?

liubo-intel: @ceciliapeng2011 : the other Paddle 'softshrink' api defined here : https://www.paddlepaddle.org.cn/documentation/docs/zh/api/paddle/nn/Softshrink_cn.html#softshrink
looks like there is no such limitation for input x type. To cover this api, I think we may still need the branch to handle '-lambd to unsigned type' cases.

liubo-intel: Hi, @AndPuQing : about @ceciliapeng2011 's concern, could you please have a try to add such unsigned type test case by using `paddle.nn.Softshrink()` api?
if such test case is difficult to generate from paddle side, let's disable such unsigned input support(delete the else branch) this time, and add an input type check `PADDLE_OP_CHECK(node, input_element_type.is_signed(), "softshrink only supports singed input data type");`  for this op.

AndPuQing: I test the unsighed tensor input on the paddlepaddle delelop version, there is the result:

```python
>>>import paddle
>>>a = paddle.to_tensor([1],dtype='uint8')
>>>a
Out[3]:
Tensor(shape=[1], dtype=uint8, place=Place(gpu:0), stop_gradient=True,
       [1])
>>>paddle.nn.functional.softshrink(a)
RuntimeError: (NotFound) The kernel with key (GPU, Undefined(AnyLayout), uint8) of kernel `softshrink` is not registered and fail to fallback to CPU one. Selected wrong DataType `uint8`. Paddle support following DataTypes: float64, float16, float32, bfloat16.
  [Hint: Expected kernel_iter != iter->second.end(), but received kernel_iter == iter->second.end().] (at /paddle/paddle/phi/core/kernel_factory.cc:259)
```

So, I think just add dtype check currently.

ceciliapeng2011 (approved):

PR #15852, opened 2023-02-21
[GPU] Skip set_selected_impl() of post_optimize_weight when target generic layer is already created

wilson-seok:
### Details:
 - Skip set_selected_impl() of pot_optimize_weight wthen target generic layer is already created

### Tickets:
 - 103595


wilson-seok:
No exec_graph diff on fp32/fp16/int8 model.

yeonbok (approved):

PR #15855, opened 2023-02-21
【PaddlePaddle Hackathon 4】add paddle one_hot_v2 op 

AndPuQing:
### Details:

add one_hot_v2 op support

### Tickets:
 - *ticket-id*

### reference

https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/nn/functional/one_hot_en.html#one-hot

https://docs.openvino.ai/latest/openvino_docs_ops_sequence_OneHot_1.html

image

### DIFF
--- a/src/frontends/paddle/tests/test_models/gen_scripts/generate_one_hot_v2.py
+++ b/src/frontends/paddle/tests/test_models/gen_scripts/generate_one_hot_v2.py
@@ -1,0 +1,44 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+# one_hot_v2 paddle model generator
+#
+import numpy as np
+from save_model import saveModel
+import sys
+
+
+def one_hot_v2(name: str, x, num_classes):
+    import paddle
+
+    paddle.enable_static()
+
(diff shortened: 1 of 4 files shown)


src/frontends/paddle/tests/test_models/gen_scripts/generate_one_hot_v2.py
```diff
@@ -0,0 +1,44 @@
...
+
+    return outs[0]
+
+
+def main():
+    test_cases = ["int32", "int64"]
+
+    for test in test_cases:
+        data_x = np.array([0, 1, 2, 3, 4, 5]).astype(test)
+        one_hot_v2("one_hot_v2_" + test, data_x, 6)
```
OpenVINO-dev-contest, line 40: hi @AndPuQing how about "depth_tensor" as a input tensor, but not a attribute ?

AndPuQing: There will be problems in the conversion, and its static derivation shape will be [x, -1] (indicating dynamic), which will cause an openvino shape check error. 

The problem is on https://github.com/PaddlePaddle/Paddle/blob/release/2.4/paddle/fluid/operators/one_hot_op.cc#L50

After that, I will try to propose a pr to the paddle repo to solve this problem

PR #15859, opened 2023-02-21
【PaddlePaddle Hackathon 4】add paddle one_hot_v2 op

Patrick-Star125:
### Details:
add one_hot_v2 operation in Paddle front end

### Reference:
- https://www.paddlepaddle.org.cn/documentation/docs/en/2.2/api/paddle/nn/functional/one_hot_en.html#one-hot

### unit-test

image


Patrick-Star125:
@OpenVINO-dev-contest  seems like parameter `depth_tensor` had been deprecated, should I keep `PADDLE_OP_CHECK` or just ignore this input.
image
image

OpenVINO-dev-contest:
> @OpenVINO-dev-contest seems like parameter `depth_tensor` had been deprecated, should I keep `PADDLE_OP_CHECK` or just ignore this input. image image

Could confirm it with Paddle‘s team, I still see this ```parameter``` in its official documents
https://github.com/PaddlePaddle/Paddle/blob/release/2.4/python/paddle/nn/functional/input.py#L109

Patrick-Star125:
I leaved a [issue](https://github.com/PaddlePaddle/Paddle/issues/50812) about this question, I will comment on this pr as soon as paddle team reply.

Patrick-Star125:
@OpenVINO-dev-contest I understand that only when the shape of num_class is 0-D, it is valid as input. I have added a corresponding test sample.

ceciliapeng2011 (approved):

src/frontends/paddle/src/op/one_hot_v2.cpp
```diff
@@ -0,0 +1,31 @@
...
+namespace paddle {
+namespace op {
+NamedOutputs one_hot_v2(const NodeContext& node) {
+    auto data = node.get_input("X");
+    Output<Node> depth;
+    if (node.has_input("depth_tensor")) {
+        auto depth_value = node.get_input("depth_tensor");
+        depth = std::make_shared<default_opset::Squeeze>(depth_value);
+    } else {
+        auto depth_value = node.get_attribute<int>("depth");
```
OpenVINO-dev-contest, line 19: Hi can we leave it as ```const```, and also set a default value for the attributes

src/frontends/paddle/tests/test_models/gen_scripts/generate_one_hot_v2.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# one_hot_v2 paddle model generator
+#
+import paddle
+import numpy as np
+from save_model import saveModel
+import sys
+
+
+def one_hot_v2_1(name: str, x, num_classes):
```
OpenVINO-dev-contest, line 13: Since there are some duplicated code, can we combine these 2 functions together? You can refer to ```top_k_v2```

OpenVINO-dev-contest:
> ### Details:
> add one_hot_v2 operation in Paddle front end
> 
> ### Reference:
> * https://www.paddlepaddle.org.cn/documentation/docs/zh/2.2/api/paddle/nn/functional/one_hot_cn.html
> 
> ### unit-test
> image

Pls update the URL with English version of OP description, thanks

Patrick-Star125:
URL updated

src/frontends/paddle/src/op/one_hot_v2.cpp
Patrick-Star125: 1.const had been added
2.input without num_class is not permitted in paddle, so I think it is presumably not needful to set a default value. If it is necessary, the default number of classes could be set as one greater than the largest class value in the input tensor, according to the [pytorch implement](https://pytorch.org/docs/stable/generated/torch.nn.functional.one_hot.html?highlight=one_hot#torch.nn.functional.one_hot).

src/frontends/paddle/src/op/one_hot_v2.cpp
```diff
@@ -0,0 +1,39 @@
...
+namespace op {
+NamedOutputs one_hot_v2(const NodeContext& node) {
+    auto data = node.get_input("X");
+    Output<Node> depth;
+    if (node.has_input("depth_tensor")) {
+        auto depth_value = node.get_input("depth_tensor");
+        depth = std::make_shared<default_opset::Squeeze>(depth_value);
+    } else {
+        const auto depth_value = node.get_attribute<int>("depth", -1);
+        if (depth_value == -1) {
```
OpenVINO-dev-contest, line 20: Why should we consider the ```depth``` as -1 or a negative, is it a available value in Paddle‘s model ? if yes, could you help to add a new case in uni-test ?

Patrick-Star125: Like I mentioned before, input without `num_class` is not permitted in Paddle‘s model and negative value is also not available. So I think it is better not to set a default value for `num_class`. If it must has a default value, I would recommend `-1`, same as the [pytorch implement](https://pytorch.org/docs/stable/generated/torch.nn.functional.one_hot.html?highlight=one_hot#torch.nn.functional.one_hot).
> num_classes ([int](https://docs.python.org/3/library/functions.html#int)) – Total number of classes. If set to -1, the number of classes will be inferred as one greater than the largest class value in the input tensor.

OpenVINO-dev-contest: Thanks. Could you help to confirm if paddle will throw an exception when we set the ```num_class``` as -1/negative ? If Paddle will reject this case, i think we can remove the default value in OV's frontend.

Patrick-Star125: It is not valid when set the `num_class` as -1/negative
image
I have removed default value as well.

OpenVINO-dev-contest (approved):

xczhai:
could you help update the doc? https://github.com/openvinotoolkit/openvino/blob/master/docs/MO_DG/prepare_model/Supported_Frameworks_Layers.md

Patrick-Star125:
So I need to add note at the end of paddle limitation?

> one_hot_v2 attribute `num_class` is a non-default variable.

I'm not sure whether it is a good description.

xczhai:
@Patrick-Star125 How about passing input with N-dims?

Patrick-Star125:
sure, done

src/frontends/paddle/tests/test_models/gen_scripts/generate_one_hot_v2.py
```diff
@@ -0,0 +1,43 @@
...
+
+def main():
+    # int32
+    data = np.array([1, 1, 3, 0]).astype("int32")
+    num_classes = 4
+    one_hot_v2("one_hot_v2_1", data, num_classes, is_tensor=False)
+    # int64
+    data = np.array([4, 1, 3, 3]).astype("int64")
+    num_classes = np.array([5]).astype("int32")
+    one_hot_v2("one_hot_v2_2", data, num_classes, is_tensor=True)
```
xczhai, line 39: Does this implement support N-dim inputs?

xczhai (changes requested):

src/frontends/paddle/tests/test_models/gen_scripts/generate_one_hot_v2.py
Patrick-Star125: now it support N-dim inputs, same as [doc](https://github.com/openvinotoolkit/openvino/blob/master/docs/ops/sequence/OneHot_1.md) description. test sample had been added.

xczhai: I mean doc should mark the `restriction`. If the implement can support N-dim input, you should remove `restriction`. Otherwise, the `restriction` should be clarified.

Patrick-Star125: I see, restrictions of input and attrs are same with paddle (non-default attr `num_class`). doc modification removed.

xczhai (approved):

PR #15866, opened 2023-02-21
[GPU] Enable crop for shape agnostic kernel

yeonbok:
### Details:
 - Enabled crop for shape agnostic kernel
### Tickets:
 - 95104

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/crop.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/crop.cpp
@@ -26,19 +26,52 @@ struct crop_impl : typed_primitive_impl_ocl<crop> {
     }
 
 public:
-    static kernel_params_t get_kernel_params(const kernel_impl_params& impl_param) {
-        auto params = get_default_params<kernel_selector::eltwise_params>(impl_param);
+    static kernel_params_t get_kernel_params(const kernel_impl_params& impl_param, bool is_shape_agnostic = false) {
+        const auto& primitive = impl_param.typed_desc<crop>();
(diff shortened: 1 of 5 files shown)


e-ddykim (approved):
It looks good to me.

src/plugins/intel_gpu/src/graph/impls/ocl/crop.cpp
```diff
@@ -26,19 +26,62 @@ struct crop_impl : typed_primitive_impl_ocl<crop> {
...
         return {params, optional_params};
     }
+        void update_dispatch_data(const kernel_impl_params& impl_param) override {
+            auto kernel_params = get_kernel_params(impl_param, true);
+            auto runtime_offset = convert_data_tensor(impl_param.get_input_layout(), impl_param.input_offsets[0]).GetFirstElementOffset();
+            kernel_selector::ScalarDescriptor s;
+            s.t = kernel_selector::ScalarDescriptor::Types::UINT32;
+            s.v.u32 = runtime_offset;
+
+            if (_kernel_data.kernels[0].params.scalars.empty())
```
sshlyapn, line 52: Shouldn't we throw an exception in that case? Because proper kernel already should be configured to have scalars, isn't it?

yeonbok: In shape agnostic kernel's creation I didnt added initial param scalar value. 
I intended to add that value at the initial run of the shape agnostic kernel. 
However we can change the behavior as well, 
- set at creation => use it

yeonbok: Applied.

PR #15868, opened 2023-02-22
[GPU] Disable converting gather8 to 7 pass because GPU plugin suppors gather8:nagative-index feature

kelvinchoi-intel:
### Details:
 - *Disable ConvertGather8ToGather7 pass for GPU plugin*

### Tickets:
 - *99348*


src/plugins/intel_gpu/src/plugin/ops/gather.cpp
```diff
@@ -46,6 +48,43 @@ void CreateGatherOpBase(Program& p, const std::shared_ptr<T>& op, const int64_t
...
+
+    // WA for NMS->Gather construction. NMS fills part of the output blob by the -1 if these values
+    // must not be taken into account.
+    // CPU also uses this like of WA.
+    if (support_neg_ind) {
+        const auto& rti = op->get_rt_info();
+        const auto& reverse = rti.find("dontReverseIndices");
+        if (reverse != rti.end()) {
+            support_neg_ind = false;
+        }
```
yeonbok, line 79: BTW, currently its like :
if support_neg_ind = false 
    then the -1 is to be used as is in the gather_ref.cl kernel. 

Then won't there be an error with accessing to indices[-1]  ? 
Shouldn't we do like, if support_neg_ind = false and the indices is -1, just return without doing anything?

yeonbok (approved):

PR #15875, opened 2023-02-22
[TF FE] Support conversion of models with non-standard extensions in the path

rkazants:
**Details:** It adds support of conversion models with non-standard extensions in the path. Also, now the frontend checks more carefully supported formats.

**Ticket:** 103793


tools/mo/unit_tests/moc_tf_fe/conversion_basic_models.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2018-2023 Intel Corporation
+# Copyright (C) 2018-2022 Intel Corporation
```
popovaan, line 1: Looks like wrong change

src/frontends/tensorflow/src/graph_iterator_proto.hpp
```diff
@@ -88,29 +88,40 @@ class GraphIteratorProto : public GraphIterator {
         }
     }
 
-    /// Set iterator to the start position
+    /// \brief Check if the input file is supported
+    template <typename T>
+    static bool is_supported(const std::basic_string<T>& path) {
+        std::ifstream pb_stream(path, std::ios::in | std::ifstream::binary);
+        auto graph_def = std::make_shared<::tensorflow::GraphDef>();
+        return pb_stream && pb_stream.is_open() && graph_def->ParsePartialFromIstream(&pb_stream);
```
gkrivor, line 96: Nervous place. But np if it isn't too heavy operation

gkrivor (approved):

tools/mo/unit_tests/moc_tf_fe/conversion_basic_models.py
rkazants, line 1:
```suggestion
# Copyright (C) 2018-2023 Intel Corporation
```

rkazants: Fixed. Thanks)

src/frontends/tensorflow/src/graph_iterator_proto.hpp
rkazants: we have tests + newly added:)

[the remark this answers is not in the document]
rkazants: this is partial parsing, a sort of light parsing to check that this is compatible format.

popovaan (approved):

itikhono (approved):

PR #15881, opened 2023-02-22
[PT FE] Add prim::device and prim::type, improvements to aten::to, aten::eq, aten::ne

mmikolajcz:
### Details:
 - *Add `prim::device` operator*
 - *Add `prim::type` operator, that is used in combination with `prim::device`*
 - *Add transformation allowing to check string equality*
 - *Fix `aten::to` scenarios where it is used without any dtype values*

### Tickets:
 - *99232*

### Comment:
- `prim::device` might still appear in unconverted operations due to issues with operators that should consume it as one of inputs.
- This ticket is continuation of https://github.com/openvinotoolkit/openvino/pull/15563


src/frontends/pytorch/src/op/to.cpp
```diff
@@ -24,22 +24,44 @@ OutputVector translate_to(NodeContext& context) {
...
+            // Cast only to device without changing dtype. Return input node unchanged.
+            return {context.get_input(0)};
+        }
+        if (fw_node && fw_node->get_op_type() == "prim::Constant") {
+            // Device param can be set using constant.
+            if (context.get_input_type(dtype_idx).is<type::Tensor>() == false) {
+                // Cast only to device without changing dtype. Return input node unchanged.
+                return {context.get_input(0)};
+            }
+        }
```
sikorsl1, line 39: Why do we need two distinct if's, if in both cases we do the same? Is it possible to simplify it?

src/frontends/pytorch/src/op/to.cpp
```diff
@@ -24,22 +24,44 @@ OutputVector translate_to(NodeContext& context) {
...
         memory_format_idx = 4;
+        auto node = context.get_input_from_visible_context(dtype_idx).get_node_shared_ptr();
+        auto fw_node = std::dynamic_pointer_cast<PtFrameworkNode>(node);
+        if (fw_node && fw_node->get_op_type() == "prim::device") {
+            // Cast only to device without changing dtype. Return input node unchanged.
+            return {context.get_input(0)};
+        }
+        if (fw_node && fw_node->get_op_type() == "prim::Constant") {
+            // Device param can be set using constant.
+            if (context.get_input_type(dtype_idx).is<type::Tensor>() == false) {
```
sikorsl1, line 35: Why "== false" instead of just "!context.get_input_type(dtype_idx).is<type::Tensor>()" ?

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -10,6 +10,7 @@
 
 import warnings
 import torch
+from torch.onnx._internal import jit_utils
```
tomdol, line 13: I understand that this import provides the functionality required by the decoder but I'm a little worried that you're importing something internal. Can you please double check if there are any remarks in the imported module that it's not recommended to import it or that it's something that might change in the future without a warning?

src/frontends/pytorch/src/op/device.cpp
```diff
@@ -0,0 +1,26 @@
...
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_device(NodeContext& context) {
+    auto decoder = context.get_decoder();
+    auto node = std::make_shared<PtFrameworkNode>(context.get_decoder(), context.inputs());
+    auto attrs = node->get_attrs();
+    auto device = decoder->get_device();
+    attrs["string_value"] = device;
```
tomdol, line 18:
```suggestion
    attrs["string_value"] = decoder->get_device();
```
...and remove the line above

src/frontends/pytorch/src/transforms/string_equality_replacer.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace pass {
+
+using namespace ov::pass;
+using namespace ov::op;
+
+std::string get_str_val(std::shared_ptr<PtFrameworkNode> fw_node) {
```
tomdol, line 25:
```suggestion
std::string get_str_val(const std::shared_ptr<PtFrameworkNode>& fw_node) {
```

src/frontends/pytorch/src/transforms/string_equality_replacer.cpp
```diff
@@ -0,0 +1,86 @@
...
+    if (!fw_node) {
+        return nullptr;
```
tomdol, line 27: Does this statement even compile? I think it should either return an empty string or throw an exception in case the required conditions are not met by the node passed in.

tomdol (changes requested):

tests/layer_tests/pytorch_tests/test_to.py
```diff
@@ -109,7 +109,42 @@ def forward(self, x, y):
 
         return aten_to(), ref_net, "aten::to"
 
+    @pytest.mark.parametrize(
+        ("use_trace"),
```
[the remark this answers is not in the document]
mmikolajcz: Thanks, removed

tests/layer_tests/pytorch_tests/test_device.py
```diff
@@ -0,0 +1,74 @@
...
+
+import numpy as np
+import pytest
+import torch
+from pytorch_layer_test_class import PytorchLayerTest
+
+
+class TestDevice(PytorchLayerTest):
+    def _prepare_input(
+        self,
```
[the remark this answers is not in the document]
mmikolajcz: Formatter did this due to dangling semicolon, fixed :)

src/frontends/pytorch/src/op/to.cpp
mmikolajcz: I was thinking about creating additional method for it or moving it outside loop, however, this check would be applicable only for cases where input size is 6 or 8, not 5. I think this way it's bit simpler.

src/frontends/pytorch/src/op/to.cpp
mmikolajcz: Fixed

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
mmikolajcz: Removed import and changed implementation to use methods provided by torch object.

src/frontends/pytorch/src/op/device.cpp
mmikolajcz: Done

src/frontends/pytorch/src/transforms/string_equality_replacer.cpp
mmikolajcz: Done

src/frontends/pytorch/src/transforms/string_equality_replacer.cpp
mmikolajcz: Changed to `FRONT_END_OP_CONVERSION_CHECK`.

tomdol (approved):

src/frontends/pytorch/src/transforms/string_equality_replacer.cpp
```diff
@@ -0,0 +1,84 @@
...
+using namespace ov::pass;
+using namespace ov::op;
+
+std::string get_str_val(const std::shared_ptr<PtFrameworkNode> fw_node) {
+    FRONT_END_OP_CONVERSION_CHECK(fw_node, "Framework node does not exist.");
+    auto attrs = fw_node->get_attrs();
+    if (attrs.find("string_value") != attrs.end()) {
+        return attrs.at("string_value");
+    }
+    return fw_node->get_decoder()->as_string();
```
mvafin, line 31: Why do we need this here?

mvafin (commented):
General comment: Can we use `Decoder::as_string` for obtaining value of `prim::device` instead of creating a new decoder API for this?

mmikolajcz:
> General comment: Can we use `Decoder::as_string` for obtaining value of `prim::device` instead of creating a new decoder API for this?

It would require modifying current implementation a bit, but it would be possible, plus due to similarity to prim::Constant this would allow to make it more unified. I've pushed those changes in latest commit, if previous implementation would be better I can revert it.

src/frontends/pytorch/src/transforms/string_equality_replacer.cpp
```diff
@@ -0,0 +1,82 @@
...
+            return false;
+        }
+        std::string lhs = get_str_val(lhs_node);
+
+        auto rhs_node =
+            std::dynamic_pointer_cast<PtFrameworkNode>(pattern_map.at(framework_node_rhs).get_node_shared_ptr());
+        if (!rhs_node) {
+            return false;
+        }
+        std::string rhs = get_str_val(rhs_node);
```
mvafin, line 56: If input to `Equal` or `NotEqual` is not a string, but some other `PtFrameworkNode` operation this will throw. We should avoid throwing in transformations, better to return `false`.

mmikolajcz: Done

src/frontends/pytorch/src/transforms/string_equality_replacer.cpp
```diff
@@ -0,0 +1,83 @@
...
+        auto rhs_attrs = rhs_node->get_attrs();
+        if (rhs_attrs.find("string_value") == rhs_attrs.end()) {
+            return false;
+        }
+        std::string rhs = rhs_attrs.at("string_value");
+
+        auto equal_node = pattern_map.at(equal_op).get_node_shared_ptr();
+        if (auto equal = std::dynamic_pointer_cast<v1::Equal>(equal_node)) {
+            auto const_result = v0::Constant::create(element::boolean, Shape{}, {lhs == rhs});
+            copy_runtime_info({lhs_node, rhs_node}, const_result);
```
mvafin, line 62:
```suggestion
            copy_runtime_info({lhs_node, rhs_node, equal_node}, const_result);
```

src/frontends/pytorch/src/transforms/string_equality_replacer.cpp
```diff
@@ -0,0 +1,83 @@
...
+            replace_node(equal_node, const_result);
+            return true;
+        };
+        auto not_equal_node = pattern_map.at(not_equal_op).get_node_shared_ptr();
+        if (auto equal = std::dynamic_pointer_cast<v1::NotEqual>(not_equal_node)) {
+            auto const_result = v0::Constant::create(element::boolean, Shape{}, {lhs != rhs});
+            copy_runtime_info({lhs_node, rhs_node}, const_result);
```
mvafin, line 69:
```suggestion
            copy_runtime_info({lhs_node, rhs_node, not_equal_node}, const_result);
```

mvafin (approved):

PR #15887, opened 2023-02-22
[GPU] Fix output format not changing at runtime

andrew-k-park:
### Details:
 - Fix dep reorder node's output format to be saved to the current reorder node's `output_format` during removing dep reorder node in `remove_redundant_reorders` opt pass

### Tickets:
 - 103951


yeonbok:
Please add a unittest.

andrew-k-park:
> Please add a unittest.

Added `remove_redundant_reorders` pass TC for ov_gpu_unit_tests

yeonbok (approved):

PR #15888, opened 2023-02-22
【PaddlePaddle Hackathon 4】add paddle set_value op

Asthestarsfalll:
### Details:
add paddle `set_value` operation in Paddle front end

`set_value` method will be called when slicing like this:
```python
x: Tensor 
v: Union[int, float, Tensor]
x[3: 4: 1, :, 5: 6: 2] = v
```
For above example, 3 means begin, 4 means end and 1 means step.

### Tickets:
 - *ticket-id*

### Reference:
There is not documents for set_value
https://github.com/PaddlePaddle/Paddle2ONNX/blob/develop/paddle2onnx/legacy/op_mapper/tensor.py#L26
https://github.com/PaddlePaddle/Paddle/blob/4383494f012b6613aa65496e7892ae3f0052ddd9/paddle/fluid/operators/set_value_op.cc
https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/variable_index.py#L641

### Remark
The Paddle2ONNX version only support step=1 and non-tenosr input (start/step/end), so is this PR.
It's able to support tensor input and arbitrary step, but the procedure will become much more complex.

Beside, there may has a potential numerical error when data type of input is interger


OpenVINO-dev-contest (commented):
hi @Asthestarsfalll thanks for your contribution. did you enable a node like that ?
image

i think these attributes can be inputs as well.

Asthestarsfalll:
> hi @Asthestarsfalll thanks for your contribution. did you enable a node like that ? image
> 
> i think these attributes can be inputs as well.

@OpenVINO-dev-contest 
The Paddle2ONNX version only support step=1 and non-tensor input (start/step/end), so is this PR.
It's able to support tensor input and arbitrary step, but the procedure will become much more complex.
I'm wondering whether the tensor inputs (start/step/end) are commonly used, I have rarely seen that it's used in this kind of way

OpenVINO-dev-contest:
> > hi @Asthestarsfalll thanks for your contribution. did you enable a node like that ? image
> > i think these attributes can be inputs as well.
> 
> @OpenVINO-dev-contest The Paddle2ONNX version only support step=1 and non-tensor input (start/step/end), so is this PR. It's able to support tensor input and arbitrary step, but the procedure will become much more complex. I'm wondering whether the tensor inputs (start/step/end) are commonly used, I have rarely seen that it's used in this kind of way

Since there is a real model case, I think we should try to enable it. Paddle2ONXX is just a reference for us, maybe we can find another way to solve it.  Its a long term hackathon, we have enough time to try it.

Asthestarsfalll:
> Since there is a real model case from PPOCR, I think we should try to enable it. Paddle2ONXX is just a reference for us, maybe we can find another way to solve it. Its a long term hackathon, we have enough time to try it.

I have already implemented a demo for tensor input, but it seems more complex. I will update as long as I find some other easy ways.
And do we need to support arbitrary steps?

Asthestarsfalll:
> Since there is a real model case from PPOCR, I think we should try to enable it. Paddle2ONXX is just a reference for us, maybe we can find another way to solve it. Its a long term hackathon, we have enough time to try it.

I have already implemented a demo for tensor input, but it seems more complex. I will update as long as I find some other easy ways.
And do we need to support arbitrary steps?

OpenVINO-dev-contest:
> > Since there is a real model case from PPOCR, I think we should try to enable it. Paddle2ONXX is just a reference for us, maybe we can find another way to solve it. Its a long term hackathon, we have enough time to try it.
> 
> I have already implemented a demo for tensor input, but it seems more complex. I will update as long as I find some other easy ways. And do we need to support arbitrary steps?

I don't see any reason that we can not support arbitrary steps. But we can start it from step=1, right ?

Asthestarsfalll:
@ceciliapeng2011 
Hi, I have update the code to support tensor input.
May you give some suggestions so that I can better support arbitrary steps?

Test result:

Asthestarsfalll:
@OpenVINO-dev-contest 
Any suggestion about current implementation?

Asthestarsfalll:
It seems that I can use ScatterNDUpdate/ScatterUpdate/ScatterElementsUpdate to simplify the codes

OpenVINO-dev-contest:
> @OpenVINO-dev-contest Any suggestion about current implementation?

Maybe you can refer to ```slice```, it has same inputs as ```set_value```

Asthestarsfalll:
> > @OpenVINO-dev-contest Any suggestion about current implementation?
> 
> Maybe you can refer to `slice`, it has same inputs as `set_value`

`Slice` is getting values from input, but `set_value` need to write value into input.
So only the openvino ops which function as seting value can do this, such as `Scatter` and `Select`.

OpenVINO-dev-contest:
> > > @OpenVINO-dev-contest Any suggestion about current implementation?
> > 
> > 
> > Maybe you can refer to `slice`, it has same inputs as `set_value`
> 
> `Slice` is getting values from input, but `set_value` need to write value into input. So only the openvino ops which function as seting value can do this, such as `Scatter` and `Select`.

Definitely, that's why we need your help to support ```set_value```, it's more complicated than ```slice```.

Asthestarsfalll:
As far as I can see, use `ScatterNDUpdate/ScatterUpdate/ScatterElementsUpdate` is overcomplicated as well:

 1.  `ScatterElementsUpdate` requires `updates` has the same shape as input `data`, and it will update the whole axis, but set_value only need update some.
 2.  `ScatterUpdate` is neither feasible.
 3.  `ScatterNDUpdate`  seems to be able complete the task, while we need to take a lot of effort to build `indices`.

Above ops all require `updates`  to have target shape, instead of taking a single value and automatically broadcast.
This means we still need to calculate target shape of `updates` and broadcast manually.

In current implementation of set_value, the conversion is as following procedure:

 1.  Calculate target shape of `updates`. Calculate padding starts and ends.
 2.  Broadcast `updates` to target shape.
 3.  Pad `updates` to the shape of `input` with `MAX_VALUE` in corresponding data type.
 4.  Get mask for those positons which are not equal to `MAX_VALUE`.
 5.  Use `Select`  to update `updates` into `input`.

For arbitrary `steps`, the conversion is as following:

 1.  Calculate **pseudo target shape** of `updates` which assume that all `setps` are 1. Calculate padding starts and ends.Record those `steps` which are not equal to 1.
 2.  Broadcast `updates` to pseudo target shape.
 3.  Split `updates` along the axis whose step is not equal 1.  Insert some `Constant` full with`MAX_VALUE` at corresponding positons, then concatenate.
 4.  Pad `updates` to the shape of `input` with `MAX_VALUE` in corresponding data type.
 5.  Get mask for those positons which are not equal to `MAX_VALUE`.
 6. Use `Select`  to update `updates` into `input`.

meiyang-intel:
Hi @Asthestarsfalll . Maybe this is an easier way to implement. The idea is to use ScatterNDUpdate to implement. About to generate the indices. It can be handled in two cases.
1. in the case of no input 'starts' 'steps' 'ends'. Because 'starts' 'steps' 'ends' are not inputs and are static. So it is easier to generate indices.
2. in the case of input 'starts' 'steps' 'ends'. That may be much complicated. I have an immature idea is to use 'Range' and 'TensorIterator' to generate indices from input 'starts' 'steps' 'ends'. Because 'Range' only support scalar input, so need 'TensorIterator' to iterate the 'starts' 'steps' 'ends' and use 'Range' in each dim of 'start' 'steps' 'ends'.

Asthestarsfalll:
> Hi @Asthestarsfalll . Maybe this is an easier way to implement. The idea is to use ScatterNDUpdate to implement. About to generate the indices. It can be handled in two cases.
> 
> 1. in the case of no input 'starts' 'steps' 'ends'. Because 'starts' 'steps' 'ends' are not inputs and are static. So it is easier to generate indices.
> 2. in the case of input 'starts' 'steps' 'ends'. That may be much complicated. I have an immature idea is to use 'Range' and 'TensorIterator' to generate indices from input 'starts' 'steps' 'ends'. Because 'Range' only support scalar input, so need 'TensorIterator' to iterate the 'starts' 'steps' 'ends' and use 'Range' in each dim of 'start' 'steps' 'ends'.

If I have an accurate comprehension of `ScatterNDUpdate`, I think we will take a lot of effort to generate `indices`.
For example, if there exists a 4-dim tensor:
```python
inputs = paddle.randn((16, 256, 512, 512))
```
Then we set vlues like this:
```python
inputs[8:] = 0
```
So wee need to generate `indices`: 
```python
[[8], [9], [10], [11], [12], [13], [14], [15]]
```
Then we set vlues like this:
```python
inputs[8:, 128:] = 0
```
We need to generate `indices`:
```python
[[[0  , 128], [0  , 129], ... , [0  , 255]],
 .....,
 [[127, 128], [127, 129], ... , [127, 255]]]
```
Further more, if we wants to set value along all 4 axes, the `indices` will become much more complicated.
Maybe it's not wise to use `Range` and `TensorIterator` to generate `indices` ?

meiyang-intel:
> > Hi @Asthestarsfalll . Maybe this is an easier way to implement. The idea is to use ScatterNDUpdate to implement. About to generate the indices. It can be handled in two cases.
> > 
> > 1. in the case of no input 'starts' 'steps' 'ends'. Because 'starts' 'steps' 'ends' are not inputs and are static. So it is easier to generate indices.
> > 2. in the case of input 'starts' 'steps' 'ends'. That may be much complicated. I have an immature idea is to use 'Range' and 'TensorIterator' to generate indices from input 'starts' 'steps' 'ends'. Because 'Range' only support scalar input, so need 'TensorIterator' to iterate the 'starts' 'steps' 'ends' and use 'Range' in each dim of 'start' 'steps' 'ends'.
> 
> If I have an accurate comprehension of `ScatterNDUpdate`, I think we will take a lot of effort to generate `indices`. For example, if there exists a 4-dim tensor:
> 
> ```python
> inputs = paddle.randn((16, 256, 512, 512))
> ```
> 
> Then we set vlues like this:
> 
> ```python
> inputs[8:] = 0
> ```
> 
> So wee need to generate `indices`:
> 
> ```python
> [[8], [9], [10], [11], [12], [13], [14], [15]]
> ```
> 
> Then we set vlues like this:
> 
> ```python
> inputs[8:, 128:] = 0
> ```
> 
> We need to generate `indices`:
> 
> ```python
> [[[0  , 128], [0  , 129], ... , [0  , 255]],
>  .....,
>  [[127, 128], [127, 129], ... , [127, 255]]]
> ```
> 
> Further more, if we wants to set value along all 4 axes, the `indices` will become much more complicated. Maybe it's not wise to use `Range` and `TensorIterator` to generate `indices` ?

yes, you're right

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,120 @@
...
+    const auto concat_node = std::make_shared<default_opset::Concat>(node, 0);
+    const auto mask = std::make_shared<default_opset::Less>(concat_node, zero);
+    const auto res = std::make_shared<default_opset::Add>(concat_node, dim);
+    return std::make_shared<default_opset::Select>(mask, res, concat_node);
+}
+
+NamedOutputs set_value(const NodeContext& node) {
+    const auto input_node = node.get_input("Input");
+    auto value_node = node.get_input("ValueTensor");
+    const auto input_shape = input_node.get_partial_shape().get_shape();
```
meiyang-intel, line 33: If use the get_shape(), that means that the shape is static. Can use the 'ShapeOf ' to support dynamic shape

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,120 @@
...
+    Shape value_shape(input_shape);
+
+    if (node.has_input("StartsTensorList") && node.has_input("StepsTensorList") && node.has_input("EndsTensorList")) {
+        const auto dim_node = default_opset::Constant::create(element::i64, {input_shape.size()}, input_shape);
+        auto starts = handle_minus_index(node.get_ng_inputs("StartsTensorList"), dim_node);
+        auto ends = handle_minus_index(node.get_ng_inputs("EndsTensorList"), dim_node);
+        const auto steps = node.get_ng_inputs("StepsTensorList");
+        std::vector<int64_t> needed_input_dim;
+        for (size_t i = 0; i < input_shape.size(); i++) {
+            if (axes[i] == static_cast<int64_t>(i))
```
meiyang-intel, line 48: That may be no problem if used in '_setitem_impl_'. What if 'set_value' is not used with '_setitem_impl_' and axes is like this [0, 2]?

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,120 @@
...
+    const auto dims = static_cast<int64_t>(input_node.get_partial_shape().rank().get_length());
+    const auto dtype = input_node.get_element_type();
+    const auto axes = node.get_attribute<std::vector<int64_t>>("axes");
+
+    Output<Node> padding_starts_node, padding_ends_node, value_target_shape;
+    Shape value_shape(input_shape);
+
+    if (node.has_input("StartsTensorList") && node.has_input("StepsTensorList") && node.has_input("EndsTensorList")) {
+        const auto dim_node = default_opset::Constant::create(element::i64, {input_shape.size()}, input_shape);
+        auto starts = handle_minus_index(node.get_ng_inputs("StartsTensorList"), dim_node);
```
meiyang-intel, line 43: I think it should not be 'dim_node', it should be [input_shape[axes[i]] for i in len(starts)]

Asthestarsfalll: Yes, you are right. Test cases are not comprehensive enough

src/frontends/paddle/src/op/set_value.cpp
Asthestarsfalll: Sorry, I don't understand what you mean

meiyang-intel: I mean all the testcase use the operator[] to allocate the 'set_value' in paddle. I see that the axes should be like a list [0, 1, 2, 3] in https://github.com/PaddlePaddle/Paddle/blob/develop/python/paddle/fluid/variable_index.py#L641 . What if it is not used with operator[] and set the axes value like [0, 2]?

Asthestarsfalll: I think we must use operator[] to call `_setitem_impl`. 
[0, 2] is illegal, because we need to specify start and end index in every axis.

Asthestarsfalll: In fact, `axes` means that there are indexes on these axes. 
For example
```python
a[1:2, 2:3, 3:4]
```
The axes are [0, 1, 2]
```python
a[1:2, :, 3:4]
```
The axes are [0, 2]

src/frontends/paddle/src/op/set_value.cpp
Asthestarsfalll: done

Asthestarsfalll:
@meiyang-intel 
I have refactor codes to spport dynamic shape.
Test results:

Asthestarsfalll:
> Hi @Asthestarsfalll , what about the progress of arbitrary steps

I haven't found any simple way to implement this.
I am wondering that how pytorch and tensorflow  frontends implement set_value?
From my perscpective, the funciton of set_value should be supported at openvino backend instead of frontends.

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // get padding ends
+    padding_ends_node =
+        default_opset::Constant::create(element::i64, {static_cast<size_t>(dims)}, std::vector<int64_t>(dims));
+    const auto ends_update_node = std::make_shared<default_opset::Subtract>(spec_dim_node, ends);
+    padding_ends_node =
+        std::make_shared<default_opset::ScatterNDUpdate>(padding_ends_node, axes_node, ends_update_node);
+
+    // get target value shape
+    Output<Node> value_shape_update_node = std::make_shared<default_opset::Add>(ends_update_node, starts);
+    value_shape_update_node = std::make_shared<default_opset::Subtract>(spec_dim_node, value_shape_update_node);
```
OpenVINO-dev-contest, line 81: I just wonder can we simply use ```ends -starts ``` to get the ```value_shape_update_node```.
Meanwhile pls update the mapping code with some explanation comments on its process which can help other reviewers to understand your codes easily. Thanks e.g https://github.com/openvinotoolkit/openvino/blob/master/src/frontends/paddle/src/op/slice_ops.hpp#L39

Asthestarsfalll: > I just wonder if we can simply use ends - starts  to get the value_shape_update_node.

No, we can't.
For example:
```python
A: Tensor  # shape (10, 10, 10)
A[2:5, :, 2:5] = value
```
The starts and ends are `[2, 2]` and `[5, 5]`，and the shape of value can be `(3, 1, 3)` or `(3, 10, 3)`.
So we have to make sure it is broadcasted to the correct shape `(3, 10, 3)` instead of `[5-2, 5-2]` .

OpenVINO-dev-contest: yes, then we can use ```ScatterNDUpdate``` to update the input shape with ```[5-2, 5-2]```?

meiyang-intel:
> 

Find that it's implement with StridedSliceAssign in tensorflow. FYI
https://github.com/tensorflow/tensorflow/blob/e9c74b5ce7ffba58465e33269e5eecf32fb0e751/tensorflow/python/ops/array_ops.py#L1347
https://github.com/tensorflow/tensorflow/blob/e9c74b5ce7ffba58465e33269e5eecf32fb0e751/tensorflow/python/ops/array_ops.py#L1324
https://github.com/tensorflow/tensorflow/blob/e9c74b5ce7ffba58465e33269e5eecf32fb0e751/tensorflow/core/kernels/strided_slice_op.cc#L286

src/frontends/paddle/src/op/set_value.cpp
Asthestarsfalll: > yes, then we can use `ScatterNDUpdate` to update the input shape with `[5-2, 5-2]`?

Oh yes, you are right.
It's been so long that I forgot what I wrote. 🤡

Asthestarsfalll:
> Find that it's implement with StridedSliceAssign in tensorflow. FYI https://github.com/tensorflow/tensorflow/blob/e9c74b5ce7ffba58465e33269e5eecf32fb0e751/tensorflow/python/ops/array_ops.py#L1347 https://github.com/tensorflow/tensorflow/blob/e9c74b5ce7ffba58465e33269e5eecf32fb0e751/tensorflow/python/ops/array_ops.py#L1324 https://github.com/tensorflow/tensorflow/blob/e9c74b5ce7ffba58465e33269e5eecf32fb0e751/tensorflow/core/kernels/strided_slice_op.cc#L286


Thank you for the information.
But what I mean is that in openvino is how pytorch and tensorflow implement the set_value。
It seems that they do not implement this.

src/frontends/paddle/src/op/set_value.cpp
Asthestarsfalll: > I just wonder if we can simply use `ends - starts ` to get the `value_shape_update_node`. Meanwhile pls update the mapping code with some explanation comments on its process which can help other reviewers to understand your codes easily. Thanks e.g https://github.com/openvinotoolkit/openvino/blob/master/src/frontends/paddle/src/op/slice_ops.hpp#L39

done.

Asthestarsfalll:
@meiyang-intel 
Hi, It's able to use `Concat`  `Reshape` and `Flatten` to support arbitrary steps.
For exaxmple:
```python
A: Tensor # shape: [10, 10, 10]
A[::2, :, ::2] = 1
```

1. `target_value_shape` is [5, 10, 5]. 
2. Reshape it to [5, 1, 10, 5] for step in axis 0 is not equl to 1.
3. If the step is less than 0, reverse elments of value along this axis.
4. Create a constant filled with maxum_value in shape[5,1, 10, 5]
5. Concatenate them, the shape is [5, 2, 10, 5]
6. Reshpe it to [10, 10, 5]
7. Do the same things with axis 2.

Finaly, the shape of value will be [10, 10, 10].

Pseudocode：
```python
for i in axes:
    if steps[i] != 1:
        if steps[i] < 0:
            value  = reverse_by_axis(value, axis=i)
        new_shape = value_shape.insert(i, 1)
        constant = create_constant(new_shape)
        value = concat([value, constant], axis=i)
        value = value.flatten(i-1, i)
        value_shape = value.shape()
```

Do you have any suggestion about this solution?

meiyang-intel:
> @meiyang-intel Hi, It's able to use `Concat` `Reshape` and `Flatten` to support arbitrary steps. For exaxmple:
> 
> ```python
> A: Tensor # shape: [10, 10, 10]
> A[::2, :, ::2] = 1
> ```
> 
> 1. `target_value_shape` is [5, 10, 5].
> 2. Reshape it to [5, 1, 10, 5] for step in axis 0 is not equl to 1.
> 3. If the step is less than 0, reverse elments of value along this axis.
> 4. Create a constant filled with maxum_value in shape[5,1, 10, 5]
> 5. Concatenate them, the shape is [5, 2, 10, 5]
> 6. Reshpe it to [10, 10, 5]
> 7. Do the same things with axis 2.
> 
> Finaly, the shape of value will be [10, 10, 10].
> 
> Pseudocode：
> 
> ```python
> for i in axes:
>     if steps[i] != 1:
>         if steps[i] < 0:
>             value  = reverse_by_axis(value, axis=i)
>         new_shape = value_shape.insert(i, 1)
>         constant = create_constant(new_shape)
>         value = concat([value, constant], axis=i)
>         value = value.flatten(i-1, i)
>         value_shape = value.shape()
> ```
> 
> Do you have any suggestion about this solution?

1. In the case of steps[i] < 0, should be careful that the concat sequence may be different. for example
`value = [1, 2]; step = -2`. then the mask of value may be `[max, 2, max, 1]`
2. should delete some element in some condition like
`input_shape = [5]; value = [1, 2]; step = 3`. then the mask of value may be `[1, max, max, 2, max]`

Asthestarsfalll:
> > @meiyang-intel Hi, It's able to use `Concat` `Reshape` and `Flatten` to support arbitrary steps. For exaxmple:
> > ```python
> > A: Tensor # shape: [10, 10, 10]
> > A[::2, :, ::2] = 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 1. `target_value_shape` is [5, 10, 5].
> > 2. Reshape it to [5, 1, 10, 5] for step in axis 0 is not equl to 1.
> > 3. If the step is less than 0, reverse elments of value along this axis.
> > 4. Create a constant filled with maxum_value in shape[5,1, 10, 5]
> > 5. Concatenate them, the shape is [5, 2, 10, 5]
> > 6. Reshpe it to [10, 10, 5]
> > 7. Do the same things with axis 2.
> > 
> > Finaly, the shape of value will be [10, 10, 10].
> > Pseudocode：
> > ```python
> > for i in axes:
> >     if steps[i] != 1:
> >         if steps[i] < 0:
> >             value  = reverse_by_axis(value, axis=i)
> >         new_shape = value_shape.insert(i, 1)
> >         constant = create_constant(new_shape)
> >         value = concat([value, constant], axis=i)
> >         value = value.flatten(i-1, i)
> >         value_shape = value.shape()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do you have any suggestion about this solution?
> 
> 1. In the case of steps[i] < 0, should be careful that the concat sequence may be different. for example
>    `value = [1, 2]; step = -2`. then the mask of value may be `[max, 2, max, 1]`
> 2. should delete some element in some condition like
>    `input_shape = [5]; value = [1, 2]; step = 3`. then the mask of value may be `[1, max, max, 2, max]`

I can't understand the usage of `TensorIterator`

Asthestarsfalll:
@meiyang-intel 
Hi, How can I get value from a tensor scalar by a bool mask?
for example:
```
mask = [true, false, true]
value = [0, 1, 2]

# How can I get following scalar?
target = [0, 2]
```
If this operation can be ealiy implemented, I think the procedure will become simpler.
> ```python
> A: Tensor # shape: [10, 10, 10]
> A[::2, :, ::2] = 1
> ```
> 
> 1. `target_value_shape` is [5, 10, 5].
> 2. Reshape it to [5, 1, 10, 5] for step in axis 0 is not equl to 1.
> 3. If the step is less than 0, reverse elments of value along this axis.
> 4. Create a constant filled with maxum_value in shape[5,1, 10, 5]
> 5. Concatenate them, the shape is [5, 2, 10, 5]
> 6. Reshpe it to [10, 10, 5]
> 7. Do the same things with axis 2.

This makes above solution get rid of `for loop` and `TensorIterator`.
We just need use `Unsqueeze` to expand `value` to corresponding shape, in this example  is (5, 1, 10, 5, 1)
Then we can pad `value` to shape (5, 2, 10, 5, 2)  at only one time

meiyang-intel:
> @meiyang-intel Hi, How can I get value from a tensor scalar by a bool mask? for example:
> 
> ```
> mask = [true, false, true]
> value = [0, 1, 2]
> 
> # How can I get following scalar?
> target = [0, 2]
> ```
> 
> If this operation can be ealiy implemented, I think the procedure will become simpler.
> 
> > ```python
> > A: Tensor # shape: [10, 10, 10]
> > A[::2, :, ::2] = 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 1. `target_value_shape` is [5, 10, 5].
> > 2. Reshape it to [5, 1, 10, 5] for step in axis 0 is not equl to 1.
> > 3. If the step is less than 0, reverse elments of value along this axis.
> > 4. Create a constant filled with maxum_value in shape[5,1, 10, 5]
> > 5. Concatenate them, the shape is [5, 2, 10, 5]
> > 6. Reshpe it to [10, 10, 5]
> > 7. Do the same things with axis 2.
> 
> This makes above solution get rid of `for loop` and `TensorIterator`. We just need use `Unsqueeze` to expand `value` to corresponding shape, in this example is (5, 1, 10, 5, 1) Then we can pad `value` to shape (5, 2, 10, 5, 2) at only one time

Hi @Asthestarsfalll , I didn't find out an easy way to implement this. Maybe we can abort this solution because  there may be performance risk. I advise to follow the method on https://github.com/PaddlePaddle/Paddle2ONNX/blob/61bc345edd7eaf346f0f14ded92a915f1e148270/paddle2onnx/mapper/tensor/set_value.cc . Although there is limitation 'steps_.size() == 1', I think it's OK.

Asthestarsfalll:
> Hi @Asthestarsfalll , I didn't find out an easy way to implement this. Maybe we can abort this solution because there may be performance risk. I advise to follow the method on https://github.com/PaddlePaddle/Paddle2ONNX/blob/61bc345edd7eaf346f0f14ded92a915f1e148270/paddle2onnx/mapper/tensor/set_value.cc . Although there is limitation 'steps_.size() == 1', I think it's OK. 

Which solution has performance risk? The method of supporting arbitrary steps or the initial implementation of set_value？
The method of paddle2onnx means that we only support the cases like this:
```python
a[start:end:step] = b
a[:, start:end:step] = b
a[..., start:end:step] = b
```
but do not  support:
```python
a[start1:end1:step1, start2:end2:step2] = b
```
The initial implementation supports:
```python
a[start1:end1:1, start2:end2:1] = b
```

meiyang-intel:
> > Hi @Asthestarsfalll , I didn't find out an easy way to implement this. Maybe we can abort this solution because there may be performance risk. I advise to follow the method on https://github.com/PaddlePaddle/Paddle2ONNX/blob/61bc345edd7eaf346f0f14ded92a915f1e148270/paddle2onnx/mapper/tensor/set_value.cc . Although there is limitation 'steps_.size() == 1', I think it's OK.
> 
> Which solution has performance risk? The method of supporting arbitrary steps or the initial implementation of set_value？ The method of paddle2onnx means that we only support the cases like this:
> 
> ```python
> a[start:end:step] = b
> a[:, start:end:step] = b
> a[..., start:end:step] = b
> ```
> 
> but do not support:
> 
> ```python
> a[start1:end1:step1, start2:end2:step2] = b
> ```
> 
> The initial implementation supports:
> 
> ```python
> a[start1:end1:1, start2:end2:1] = b
> ```

I mean the solution we discussed before to support arbitrary step. It is complicated and may use the loop or TensorIterator.
I think it's better to follow the current implement of 'set_value' in Paddle2ONNX for now https://github.com/PaddlePaddle/Paddle2ONNX/blob/61bc345edd7eaf346f0f14ded92a915f1e148270/paddle2onnx/mapper/tensor/set_value.cc . Althrough it doesn't support step.size() > 1 below:
```python
a[start1:end1:step1, start2:end2:step2] = b
```

meiyang-intel:
Hi @Asthestarsfalll , after internal discuss and thanks to @rkazants , there is another solution to implement arbitrary 'steps'.
1. Generate a tensor of indices [0, 1, ..., num_elem - 1] and reshape to the same shape as input.
2. Apply StridedSlice to this indices tensor
3. flatten the input, indices tensor, and values.
4. Apply Scatter operation and reshape back to the original shape
I think this is not much complicated and can support the arbitrary 'steps'. What do you think?

Asthestarsfalll:
> Hi @Asthestarsfalll , after internal discuss and thanks to @rkazants , there is another solution to implement arbitrary 'steps'.
> 
> 1. Generate a tensor of indices [0, 1, ..., num_elem - 1] and reshape to the same shape as input.
> 2. Apply StridedSlice to this indices tensor
> 3. flatten the input, indices tensor, and values.
> 4. Apply Scatter operation and reshape back to the original shape
>    I think this is not much complicated and can support the arbitrary 'steps'. What do you think?

Yes! I think it's more succinct. 
And I implement this, but occurr an issue that `"StridedSlice node doesn't have compiled executor!"`.
It's that my development environment cause this?
Anyway, I'll submit the code first.

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,157 @@
...
+    // steps = [2, 1]
+    // Our process is:
+    // 1. Get axes [2, 3], get shape of input [5, 6, 7, 8, 9], select dimension from shape by axes: [7, 8].
+    // 2. Get starts [2, -4] and ends [3, -1]. Process minus starts and ends. starts: [2, 4], ends: [7, 7].
+    // 3. Calculate starts_node, ends_node and steps_node
+    //    1. Create `starts node` filled with 0. Update `starts` to `starts_node` according to axes.
+    //    starts node[axes[i]] = starts[i] for i in axes.size
+    //    starts node: [0, 0, 0, 0, 0] -> [0, 0, 2, 4, 0].
+    //    2. Create `ends_node` filled with 0. Update `dim - ends` to `ends_node` according to axes.
+    //    ends_node[axes[i]] = input_shape[axes[i]] - ends[i] for i in axes.size
```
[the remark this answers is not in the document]
Asthestarsfalll: This comment is for padding method, I will update this

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,157 @@
...
+    }
+    else PADDLE_OP_CHECK(node, (false), "Invalid arguments!");
+
+    // 3.1 get starts node
+    const auto zero_dim_node =
+        default_opset::Constant::create(element::i64, {static_cast<size_t>(dims)}, std::vector<int64_t>(dims));
+    starts_node = std::make_shared<default_opset::ScatterNDUpdate>(zero_dim_node, axes_node, starts);
+
+        // 3.2 get ends node
+        const auto ends_update_node = std::make_shared<default_opset::Subtract>(spec_dim_node, ends);
```
meiyang-intel, line 105: Some lines are not aligned. Recommend to use build command 'make clang_format_fix_all' to automatically fix the code format issue.

src/frontends/paddle/src/op/set_value.cpp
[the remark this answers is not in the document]
meiyang-intel: Not only in the comment but also the 'ends_update_node' in line 105.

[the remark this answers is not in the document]
Asthestarsfalll: > Not only in the comment but also the 'ends_update_node' in line 105.

yes, I have updated just now

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,155 @@
...
+    const auto zero_node = default_opset::Constant::create(element::i64, Shape{}, {0});
+    const auto one_node = default_opset::Constant::create(element::i64, Shape{}, {1});
+    const auto dim_node = default_opset::Constant::create(element::i64, Shape{}, {dims});
+    const auto reshape_flatten = default_opset::Constant::create(ov::element::i64, {1}, {-1});
+
+    // get positive starts ends and steps
+    if (node.has_input("StartsTensorList") && node.has_input("StepsTensorList") && node.has_input("EndsTensorList")) {
+        starts = handle_minus_index(node.get_ng_inputs("StartsTensorList"), spec_dim_node);
+        ends = handle_minus_index(node.get_ng_inputs("EndsTensorList"), spec_dim_node);
+        steps = handle_minus_index(node.get_ng_inputs("StepsTensorList"), spec_dim_node);
```
meiyang-intel, line 86: Does the 'steps' need to 'handle_minus_index'?

src/frontends/paddle/tests/test_models/gen_scripts/generate_set_value.py
```diff
@@ -0,0 +1,148 @@
...
+    data = np.random.randint(0, 5, shape).astype(dtype)
+    value = np.random.randint(0, 2, (1, )).astype(dtype)
+    starts = generate_data([-4], np.int64)
+    ends = generate_data([-1], np.int64)
+    steps = generate_data([1], np.int64)
+
+    def set_value7(x, value, *slice):
+        x[build_slice(*slice)] = value
+        return x
+
```
meiyang-intel, line 144: Can you add some test case with step > 1?

Asthestarsfalll: done, but I can't test since there are some issue about `StridedSlice`

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,156 @@
...
+                                                                                   std::vector<int64_t>(dims),
+                                                                                   std::vector<int64_t>(dims));
+
+    // flatten input, upadte_value and sliced_range_node
+    input_node = std::make_shared<default_opset::Reshape>(input_node, reshape_flatten, true);
+    sliced_range_node = std::make_shared<default_opset::Reshape>(sliced_range_node, reshape_flatten, true);
+    value_node = std::make_shared<default_opset::Reshape>(value_node, reshape_flatten, true);
+
+    // update value to input according to sliced_range_node
+    input_node = std::make_shared<default_opset::ScatterUpdate>(input_node, sliced_range_node, value_node, zero_node);
```
meiyang-intel, line 146: From the error log of ScatterUpdate, the shape of input/slice/value is incorrect. Recommend to use ScatterNDUpdate to update value

Asthestarsfalll: The shape error is due to the slice operation, which will return the maximum number in int32
for example:
```python
x[::2] = 1
```
so start, end and step are 0, 214748364, 2

Asthestarsfalll: It's weird that if start or end are minus, it will throw `StridedSlice node doesn't have compiled executor!`. But I have converted them to positive in fucntion `handle_minus_index`.

meiyang-intel: Please check the spec of https://docs.openvino.ai/latest/openvino_docs_ops_movement_ScatterUpdate_3.html . Need to reshape the indice and value if want to use the ScatterUpdate. Use ScatterNDUpdate may be easier here.
About the StridedSlice error, it is another issue. May need to debug

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,156 @@
...
+    if (node.has_input("StartsTensorList") && node.has_input("StepsTensorList") && node.has_input("EndsTensorList")) {
+        starts = handle_minus_index(node.get_ng_inputs("StartsTensorList"), spec_dim_node);
+        ends = handle_minus_index(node.get_ng_inputs("EndsTensorList"), spec_dim_node);
+        steps = std::make_shared<default_opset::Concat>(node.get_ng_inputs("StepsTensorList"), 0);
+    } else if (node.has_attribute("starts") && node.has_attribute("steps") && node.has_attribute("ends")) {
+        starts = handle_minus_index(node.get_attribute<std::vector<int64_t>>("starts"), spec_dim_node);
+        ends = handle_minus_index(node.get_attribute<std::vector<int64_t>>("ends"), spec_dim_node);
+        auto step_vec = node.get_attribute<std::vector<int64_t>>("steps");
+        for (size_t i = 0; i < step_vec.size(); i++)
+            PADDLE_OP_CHECK(node, (step_vec[i] == 1), "Elements of steps must be 1");
```
meiyang-intel, line 92: Remove the constraints.

src/frontends/paddle/src/op/set_value.cpp
Asthestarsfalll: done, but only the cases whose all params(starts, ends) are positive can pass the tests

Asthestarsfalll:
The bug appears to be related to the processing of input. When I don't process the input (starts, ends), most of the unit tests pass.However, the remaining examples require us to process the input.

meiyang-intel:
> The bug appears to be related to the processing of input. When I don't process the input (starts, ends), most of the unit tests pass.However, the remaining examples require us to process the input.

There is input data dependency in strided_slice when starts or ends is negative. https://github.com/openvinotoolkit/openvino/blob/175e169537f5acd5cb58022ec8b6e3b65b38295d/src/plugins/intel_cpu/src/nodes/strided_slice.cpp#L389 . Seems the strided_slice doesn't handle correctly in the case of data dependency and that cause the bug. Will continue to check.

Asthestarsfalll:
I tried to use `Mod` (modulus operation) to tackle this issue, but it had no effect. Do you have any other suggestions?

meiyang-intel:
> @meiyang-intel I think `shapeHasDataDependency` is mainly related to `Select` operation, so I tried to use `Mod` (modulus operation) to tackle this issue, but it had no effect. Do you have any other suggestions?

Hi, @Asthestarsfalll , sorry for the delay reply. I have a workaround to pass this issue but it cause other issue. It may take some time to have a solution to fix this issue. I think we can first merge this PR by removing that testcase which has issue and adding some constraints like not support negative starts/ends. After this issue fixed we will add these testcase and remove the constraints.

Asthestarsfalll:
unittests:

Asthestarsfalll:
> > @meiyang-intel I think `shapeHasDataDependency` is mainly related to `Select` operation, so I tried to use `Mod` (modulus operation) to tackle this issue, but it had no effect. Do you have any other suggestions?
> 
> Hi, @Asthestarsfalll , sorry for the delay reply. I have a workaround to pass this issue but it cause other issue. It may take some time to have a solution to fix this issue. I think we can first merge this PR by removing that testcase which has issue and adding some constraints like not support negative starts/ends. After this issue fixed we will add these testcase and remove the constraints.

Hi, @meiyang-intel 
What is the workaround?
I have updated code, but I have no ideo how to judge if tensor inpust have negative numbers.

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,180 @@
...
+    const auto axes_node = default_opset::Constant::create(element::i64, {axes.size(), 1}, axes);
+    const auto spec_dim_node = std::make_shared<default_opset::GatherND>(input_shape, axes_node);
+    const auto zero_node = default_opset::Constant::create(element::i64, Shape{}, {0});
+    const auto one_node = default_opset::Constant::create(element::i64, Shape{}, {1});
+    const auto dim_node = default_opset::Constant::create(element::i64, Shape{}, {dims});
+    const auto reshape_flatten = default_opset::Constant::create(ov::element::i64, {1}, {-1});
+    const auto slice_shape = default_opset::Constant::create(ov::element::i64, {1, 1}, {-1});
+
+    // get positive starts ends and steps
+    if (node.has_input("StartsTensorList") && node.has_input("StepsTensorList") && node.has_input("EndsTensorList")) {
```
meiyang-intel, line 97: If have no good solution to check the negative number in starts/ends input tensor, can remove the support of starts/ends inputs currently. Will the testcase use the starts/ends input or attribute?

ceciliapeng2011:
@Asthestarsfalll Could you please locally test your PR with template plugin and upload the test result screenshot? Once it's approved, could you please disable the failing test cases and submit your commits? We will address the failures in a separate ticket and enable your test cases.

To test locally with TEMPLATE plugin -
1. change https://github.com/openvinotoolkit/openvino/blob/5b6bab263667a754bf33f26408a19d8322b60514/src/frontends/tests/frontend/shared/src/op_fuzzy.cpp#L68 to "TEMPLATE"
2. build with template plugin enabled with cmake option "-DENABLE_TEMPLATE=ON"

Asthestarsfalll:
> 1. change https://github.com/openvinotoolkit/openvino/blob/5b6bab263667a754bf33f26408a19d8322b60514/src/frontends/tests/frontend/shared/src/op_fuzzy.cpp#L68
>     to "TEMPLATE"

How can I exactly do this?

meiyang-intel (approved):

Asthestarsfalll:
Is there anything I need to do?

ceciliapeng2011 (approved):

src/frontends/paddle/src/op/set_value.cpp
```diff
@@ -0,0 +1,180 @@
+// // Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "default_opset.hpp"
+#include "openvino/frontend/paddle/node_context.hpp"
+
+#define INT_MAX 2147483647
```
ilya-lavrenov, line 8: please, remove and use https://en.cppreference.com/w/c/types/limits

Asthestarsfalll: I will modify this by next PR

ilya-lavrenov: Already fixed here https://github.com/openvinotoolkit/openvino/pull/17890

PR #15891, opened 2023-02-22
Review scatter elements update class for shape inference aspects

praasz:
### Details:
 - Review interval shape and label propagation.
 - Review template implementation of `shape_infer` function.
 - Add evaluate upper, lower and labels.

### Tickets:
 - 97254
 - 97284
 - 97333
 - 97382
 - 97534
 - 97583


tomdol (approved):

src/core/src/op/scatter_elements_update.cpp
```diff
@@ -51,17 +52,9 @@ void op::v3::ScatterElementsUpdate::validate_and_infer_types() {
...
-    std::vector<ov::PartialShape> output_shapes = {ov::PartialShape()};
-    std::vector<ov::PartialShape> input_shapes = {data, indices, updates, axis};
-
-    shape_infer(this, input_shapes, output_shapes);
-    set_output_type(0, data_et, output_shapes[0]);
-    if (output_shapes[0].is_dynamic())
+    const auto output_shape = shape_infer(this, get_node_input_partial_shapes(*this)).front();
+    set_output_type(0, data_et, output_shape);
+    if (output_shape.is_dynamic())
         set_input_is_relevant_to_shape(0);
```
jane-intel, line 58: FYI, this is the exact method that I was pointing to when we discussed the port_map. I find it irrelevant here.

jane-intel (approved):

src/core/src/op/scatter_elements_update.cpp
praasz: Sure, I've just left it as it was done before.

PR #15895, opened 2023-02-22
【PaddlePaddle Hackathon 4】add silu op mapping in Paddle frontend

jinyouzhi:
### Details:
- add silu operation in Paddle front end
- map silu to swish because of similar equivalent formula

### Reference:
https://www.paddlepaddle.org.cn/documentation/docs/zh/api/paddle/nn/Silu_cn.html#silu

### UT:
image


akladiev:

akladiev:

PR #15920, opened 2023-02-23
rework the config and better follow new api

songbell:
### Details:
 - rework the internal config logic to 1) better apply to api 2.0, 2) clearer logic for internal use

### Tickets:
 - *ticket-id*


src/plugins/auto/utils/plugin_config.hpp
```diff
@@ -0,0 +1,256 @@
...
+            metrics.push_back(METRIC_KEY(SUPPORTED_CONFIG_KEYS));
+            metrics.push_back(METRIC_KEY(OPTIMIZATION_CAPABILITIES));
+            return metrics;
+        }();
+        auto multi_supported_metrics = supported_metrics;
+        return pluginName == "AUTO" ? supported_metrics : multi_supported_metrics;
+    }
+
+    bool isSupportedDevice(const std::string& deviceName) const {
+        if (deviceName.empty())
```
songbell, line 210: @yangwang201911 is it a must that we need to check SupportDevice?

src/plugins/auto/plugin.cpp
```diff
@@ -452,12 +478,8 @@ IExecutableNetworkInternal::Ptr MultiDeviceInferencePlugin::LoadNetworkImpl(cons
...
         autoSContext->_LogTag = _LogTag;
-        auto tmpiter = fullConfig.find(ov::intel_auto::device_bind_buffer.name());
-        if (tmpiter != fullConfig.end() && tmpiter->second == PluginConfigParams::YES)
-            autoSContext->_bindBuffer = true;
-        auto tmpiter_enableStartupFallback = fullConfig.find(ov::intel_auto::enable_startup_fallback.name());
-        if (tmpiter_enableStartupFallback != fullConfig.end() && tmpiter_enableStartupFallback->second == PluginConfigParams::NO)
-            autoSContext->_startupfallback = false;
+        autoSContext->_bindBuffer = loadConfig.get_property(ov::intel_auto::device_bind_buffer);
+        autoSContext->_startupfallback = loadConfig.get_property(ov::intel_auto::enable_startup_fallback);
         return std::make_shared<AutoExecutableNetwork>(autoSContext, std::make_shared<AutoSchedule>());
```
songbell, line 483: @WeldonWangwang please help review the change

src/plugins/auto/plugin.cpp
```diff
@@ -526,8 +548,8 @@ IExecutableNetworkInternal::Ptr MultiDeviceInferencePlugin::LoadNetworkImpl(cons
                 sStreamNums = exec_net->GetMetric(ov::num_streams.name()).as<std::string>();
                 sThreadNums = exec_net->GetMetric(ov::inference_num_threads.name()).as<std::string>();
             } else if (deviceName.find("GPU") != std::string::npos) {
-                sStreamNums = exec_net->GetConfig(PluginConfigParams::KEY_GPU_THROUGHPUT_STREAMS).as<std::string>();
-                sThreadNums = exec_net->GetConfig(GPUConfigParams::KEY_GPU_MAX_NUM_THREADS).as<std::string>();
+                sStreamNums = exec_net->GetConfig(ov::num_streams.name()).as<std::string>();
+                sThreadNums = exec_net->GetConfig(ov::compilation_num_threads.name()).as<std::string>();
             }
```
songbell, line 553: @wgzintel please help review this change

src/plugins/auto/plugin_config.cpp
```diff
@@ -0,0 +1,108 @@
...
+    for (auto& kv : properties) {
+        auto& name = kv.first;
+        auto& val = kv.second;
+        if (is_supported(kv.first)) {
+            OPENVINO_ASSERT(property_validators.at(name)->is_valid(val),
+                    "[AUTO]", "Invalid value for property ", name,  ": ", val.as<std::string>());
+            internal_properties[name] = val;
+            // when user call set_property to set some config to plugin, we also respect this and pass through the config in this case
+            user_properties[name] = val;
+        } else {
```
songbell, line 50: take key perf_count for example
case1: core.set_property(auto, perf_count)
          when auto compile model to CPU, auto should pass the above property through
case 2, compile_model(auto, perf_count)
           later when auto compile model to CPU, auto should pass the above property through
case 3, not set property to auto and no perf count config set with compiled model with auto
           later when auto compile model to CPU, auto should **NOT** pass the above property through
is my understanding corrent? @yangwang201911  @wangleis

wangleis: @songbell perf_count is HW independent property. So in case 3, AUTO will pass default perf_count in AUTO to HW plugin. Your understanding for case 3 is right for HW related property.

src/plugins/auto/utils/plugin_config.hpp
```diff
@@ -0,0 +1,256 @@
...
+        } catch (ov::Exception&) {
+            return false;
+        }
+    }
+};
+
+class PerformanceModeValidator : public BaseValidator {
+public:
+    bool is_valid(const ov::Any& v) const override {
+        auto mode = v.as<ov::hint::PerformanceMode>();
```
yangwang201911, line 54: should remove below logic as v.as<T>() will check if the value is the valid.

src/plugins/auto/plugin_config.cpp
```diff
@@ -0,0 +1,105 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+#include "utils/plugin_config.hpp"
+
+namespace MultiDevicePlugin {
+const std::set<std::string> PluginConfig::_availableDevices = {"AUTO", "CPU", "GPU", "TEMPLATE", "NVIDIA", "VPUX", "MULTI", "HETERO", "mock"};
```
songbell, line 7: @yangwang201911 is this list correct for now?

src/plugins/auto/plugin.cpp
```diff
@@ -52,11 +52,52 @@ namespace {
...
     }
+    int MapPriorityValues(ov::hint::Priority priority) {
+        switch (priority) {
+            case ov::hint::Priority::HIGH:
+                return 0;
+            case ov::hint::Priority::MEDIUM:
+                return 1;
+            case ov::hint::Priority::LOW:
+                return 2;
+            default:
```
wangleis, line 63: Could we have common definition about model priority in whole OpenVINO?

src/tests/unit/auto/auto_default_perf_hint_test.cpp
```diff
@@ -295,7 +295,7 @@ class AutoDefaultPerfHintTest : public ::testing::TestWithParam<ConfigParams> {
         std::vector<std::string> configKeys = {"SUPPORTED_CONFIG_KEYS", "NUM_STREAMS"};
         ON_CALL(*core, GetMetric(_, StrEq(METRIC_KEY(SUPPORTED_CONFIG_KEYS)), _)).WillByDefault(Return(configKeys));
 
-        ON_CALL(*core, GetConfig(_, StrEq(GPU_CONFIG_KEY(MAX_NUM_THREADS)))).WillByDefault(Return(12));
+        ON_CALL(*core, GetConfig(_, StrEq(ov::compilation_num_threads.name()))).WillByDefault(Return(12));
```
wangleis, line 298: we still need to keep 1.0 config kay test case now.

src/plugins/auto/plugin_config.cpp
yangwang201911: I think so.

src/plugins/auto/plugin.cpp
WeldonWangwang: Agree

src/plugins/auto/plugin_config.cpp
```diff
@@ -0,0 +1,115 @@
...
+        std::make_tuple(ov::hint::allow_auto_batching, true),
+        std::make_tuple(ov::auto_batch_timeout, 1000),
+        // Legacy API properties
+        std::make_tuple(exclusive_asyc_requests, false),
+        // RO for register only
+        std::make_tuple(ov::device::full_name),
+        std::make_tuple(ov::device::capabilities),
+        std::make_tuple(ov::supported_properties));
+}
+void PluginConfig::register_property_impl(const ov::AnyMap::value_type& property, ov::PropertyMutability mutability, BaseValidator::Ptr validator) {
```
riverlijunjie, line 39: Property definition itself has contained Mutability, why add additional Mutability parameter?

src/plugins/auto/plugin.cpp
```diff
@@ -52,11 +52,52 @@ namespace {
         }
         return METRIC_VALUE(FP32);
     }
+    int MapPriorityValues(ov::hint::Priority priority) {
```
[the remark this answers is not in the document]
WeldonWangwang: Similarly questions, do we have any rules for naming style? Snake case or camel case?

yangwang201911 (approved):

WeldonWangwang (approved):

src/plugins/auto/plugin.cpp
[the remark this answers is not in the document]
songbell: will update, I think recent changes in iplugin interface seems switching to snake case, while old iplugin is with camel, I suppose we will need to follow new iplugin in future.

src/plugins/auto/plugin.cpp
```diff
@@ -52,11 +52,52 @@ namespace {
...
+                return 0;
+            case ov::hint::Priority::MEDIUM:
+                return 1;
+            case ov::hint::Priority::LOW:
+                return 2;
+            default:
+                return 1;
+        }
+    }
+    std::map<std::string, std::string> convert_to_string_map(ov::AnyMap& properties) {
```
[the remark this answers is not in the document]
songbell: this is to minimize the changes to internal APIs, and also, old plugin and icore implementation are with old string APIs, this can be removed when plugin switched to 2.0 also

wgzintel (approved):

src/plugins/auto/plugin.cpp
```diff
@@ -52,11 +52,52 @@ namespace {
...
 }  // namespace
 
 std::mutex MultiDeviceInferencePlugin::_mtx;
 std::map<unsigned int, std::list<std::string>> MultiDeviceInferencePlugin::_priorityMap;
 
+ov::AnyMap MultiDeviceInferencePlugin::preprocess_config(const std::map<std::string, std::string>& orig_config) const {
+    ov::AnyMap properties = ov::AnyMap(orig_config.begin(), orig_config.end());
+    for (auto& property : properties) {
+        // for model_priority, the values need to be converted
+        if (property.first == ov::hint::model_priority.name() && !IsNewAPI()) {
```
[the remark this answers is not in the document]
songbell: fixed, actually after debug I think the api check is not needed here, hence remove it

src/plugins/auto/plugin_config.cpp
songbell: original Property has been expanded to anyMap, and mutability here is for all the supported config/metric query

wangleis (approved):

PR #15926, opened 2023-02-24
Fix warnings for IE

panhaiqi:
### Details:
 - *Fix warnings for IE*

### Tickets:
 - *93719*


ilyachur:
Were all warnings already fixed? Or don't we catch their in our CI?

panhaiqi:
> Were all warnings already fixed? Or don't we catch their in our CI?

Actually I didn't find warnings when I built this repo in local machine. And It seems that there is no warning in CI?

src/inference/CMakeLists.txt
```diff
@@ -4,10 +4,6 @@
 
 set (TARGET_NAME "inference_engine")
 
-if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-    # TODO: Add IMPLEMENT_OPENVINO_API to inference_engine_obj
```
ilya-lavrenov, line 8: maybe we need to use `IMPLEMENT_OPENVINO_API ` instead of `IMPLEMENT_INFERENCE_ENGINE_API` here:
https://github.com/openvinotoolkit/openvino/blob/713b37cb251782333040f9798ada0989bc9b8823/src/inference/include/openvino/runtime/common.hpp#L18-L29
?

src/inference/CMakeLists.txt
```diff
@@ -138,7 +134,7 @@ ie_faster_build(${TARGET_NAME}_obj
 )
 
 target_compile_definitions(${TARGET_NAME}_obj PRIVATE
-    IMPLEMENT_INFERENCE_ENGINE_API
+    IMPLEMENT_OPENVINO_API
```
ilya-lavrenov, line 137: I think we need completely drop `IMPLEMENT_INFERENCE_ENGINE_API` then from OpenVINO code

panhaiqi: > IMPLEMENT_INFERENCE_ENGINE_API

@ilya-lavrenov I just found IMPLEMENT_INFERENCE_ENGINE_API in common.hpp, ie_api.h and CMakeLists.txt, any other places should be fixed?

ilya-lavrenov: I suppose `grep` knows better than me all the places where `IMPLEMENT_INFERENCE_ENGINE_API` should be removed :)

panhaiqi: > I suppose `grep` knows better than me all the places where `IMPLEMENT_INFERENCE_ENGINE_API` should be removed :)

grep_IMPLEMENT_INFERENCE_ENGINE_API
It seems that there are 3 places in openvino code. @ilya-lavrenov

ilya-lavrenov: yes

src/inference/include/openvino/runtime/common.hpp
```diff
@@ -19,7 +19,7 @@
 #    define OPENVINO_RUNTIME_API_C(...) OPENVINO_EXTERN_C __VA_ARGS__
 #    define OPENVINO_RUNTIME_API
 #else
-#    ifdef IMPLEMENT_INFERENCE_ENGINE_API  // defined if we are building the OpenVINO runtime DLL (instead of using it)
+#    ifdef IMPLEMENT_OPENVINO_API  // defined if we are building the OpenVINO runtime DLL (instead of using it)
 #        define OPENVINO_RUNTIME_API_C(...) OPENVINO_EXTERN_C OPENVINO_CORE_EXPORTS __VA_ARGS__ OPENVINO_CDECL
 #        define OPENVINO_RUNTIME_API        OPENVINO_CORE_EXPORTS
 #    else
```
ilyachur, line 25: Please also fix:
```
#    endif  // IMPLEMENT_INFERENCE_ENGINE_API
```

src/inference/include/ie/ie_api.h
```diff
@@ -15,11 +15,7 @@
 #    define INFERENCE_ENGINE_API_CLASS(...) __VA_ARGS__
 #else
 #    if defined(_WIN32)
-#        ifdef IMPLEMENT_INFERENCE_ENGINE_API
-#            define INFERENCE_ENGINE_API(...)       extern "C" __declspec(dllexport) __VA_ARGS__ __cdecl
```
ilya-lavrenov, line 19: why?

akladiev:

PR #15929, opened 2023-02-24
[LoadTime][MO] flush fp32 subnormals to zero at offline phase

pavel-esir:
**Details:**
We already had flushing subnormals to zero but at online phase.
image

If we do it at offline phase this expensive procedure will not be run at online. After getting rid of subnormals during IR generation there are no extra copying at [src/plugins/intel_cpu/src/nodes/input.cpp#L358-L364 ](https://github.com/pavel-esir/openvino/blob/master/src/plugins/intel_cpu/src/nodes/input.cpp#L358-L364), this improved LoadTime on CPU.

On distilbert-base-cased LoadTIme improved from 220 ms -> 160 ms on ADL 12th Gen Intel(R) Core(TM) i9-12900K. 
FP16 ReadTime + LoadTime = 160 + 50 ~= 210 ms (almost comparable with fp32)
FP32 ReadTime + LoadTime 85 + 110 ~195 ms

**Important**: FP16 subnormals was not changed: they are kept in IR and they are important. FP16 subnormal values (~1.0e-8 -- 1.0e-5) are not as small as FP32 (~1.0e-45 -- 1.0e-38), we do not flush them to zero, they will be upcasted to FP32 for CPU.

 **Tickets:**
 - xxx-104573, xxx-91865


slyalin:
@dmitry-gorokhov, @vladimir-paramuzov, @AlexKoff88, please accept the change -- it may have impact on accuracy if somewhere such values are really used (but we don't know about such places). If you have some feedback, please share. I suppose we will have detailed perf/accuracy comparison eventually.

AlexKoff88:
I am ok if it is only for FP32.

AlexKoff88 (approved):

src/plugins/intel_cpu/src/nodes/input.cpp
```diff
@@ -355,6 +355,7 @@ void Input::cloneBlobIfRequired() {
     if (weightCache) {
         MemoryPtr ptr = *weightCache->findOrCreate(blobKey(), cloneBlob);
         memoryPtr = std::const_pointer_cast<const Memory>(ptr);
+    // todo: at offline phase we always flush subnormals to zero, probably hasSubnormals can be removed?
```
rkazants, line 358: why should we have this comment here?
I think we should rather to say:
// read_model scenario with directly loaded original model still requires this subnormal normalization

src/bindings/python/src/pyopenvino/core/offline_transformations.cpp
```diff
@@ -141,4 +142,13 @@ void regmodule_offline_transformations(py::module m) {
...
         py::arg("model"));
+
+    m_offline_transformations.def(
+        "apply_flush_fp32_subnormals_to_zero",
+        [](std::shared_ptr<ov::Model> model) {
+            ov::pass::Manager manager;
+            manager.register_pass<ov::pass::FlushFP32SubnormalsToZero>();
+            manager.run_passes(model);
+        },
+        py::arg("model"));
```
rkazants, line 153: not needed here to create a new Python routine, just update existing `m_offline_transformations` with this transformation as we did for `SmartReshape` and `CF`

src/bindings/python/tests/test_transformations/test_offline_api.py
```diff
@@ -339,3 +340,26 @@ def test_convert_gru_to_tensor_iterator():
...
+                                  weights,
+                                  strides=[1, 1], pads_begin=[0, 0], pads_end=[0, 0], dilations=[1, 1])
+
+    result = ov.opset10.result(conv)
+    model = Model([result], [parameter])
+
+    apply_flush_fp32_subnormals_to_zero(model)
+
+    assert np.all(weights.data[0, 2:3] != subnormal_val)
+    assert np.all(weights.data[0, 2:3] == 0.0)
```
rkazants, line 365: can we have more simple test case please with lower rank tensor please?:)
`assert np.all(weights.data[0, 2:3] == 0.0)` is hard to recall what it means for 4D tensor weights.

rkazants (changes requested):

src/plugins/intel_cpu/src/nodes/input.cpp
pavel-esir: done

src/bindings/python/src/pyopenvino/core/offline_transformations.cpp
pavel-esir: done

src/bindings/python/tests/test_transformations/test_offline_api.py
pavel-esir: done

src/common/transformations/tests/common_optimizations/flush_fp32_subnormals_to_zero_test.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+namespace {
+union FloatIntUnion {
+    uint32_t u;
+    float f;
+};
+FloatIntUnion maximum_subnorm_val = {0x007fffff};  // = 2^−126 * (1 - 2^−23) ~= 1.1754942107e-38f
+FloatIntUnion minimum_subnorm_val = {0x00000001};  // = 2^−149 ~= 1.4012984643e-45f
+FloatIntUnion minimum_norm_val = {0x00800000};     // = 2^−126 ~= 1.1754943508-38f
+}  // namespace
```
rkazants, line 28: please use limits, all of these values are defined: https://en.cppreference.com/w/cpp/types/numeric_limits/denorm_min
please replace all hard-coded numbers in PR.

src/common/transformations/src/transformations/flush_fp32_subnormals_to_zero.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+        union {
+            float maximum_subnorm_val;
+            std::uint32_t uint_maximum_subnorm_val;
+        };
+        uint_maximum_subnorm_val = 0x007fffff;  // = 2^−126 * (1 - 2^−23) ~= 1.175e-38f
+
+        bool has_denormals = false;
+        for (size_t i = 0; i < size; ++i) {
+            if (std::abs(data[i]) <= maximum_subnorm_val && data[i] != 0.0f) {
```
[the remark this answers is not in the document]
pavel-esir: thanks! used fpclasify

src/common/transformations/include/transformations/flush_fp32_subnormals_to_zero.hpp
```diff
@@ -0,0 +1,30 @@
...
+namespace pass {
+
+class TRANSFORMATIONS_API FlushFP32SubnormalsToZero;
+
+}  // namespace pass
+}  // namespace ov
+
+/* @ingroup ie_transformation_common_api
+ * @brief FlushFP32SubnormalsToZero flushes f32 subnormals to zero.
+ * This is read/write expensive transformation, threfore should be run at offline phase.
```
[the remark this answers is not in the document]
pavel-esir: done

src/common/transformations/tests/common_optimizations/flush_fp32_subnormals_to_zero_test.cpp
```diff
@@ -0,0 +1,382 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <gtest/gtest.h>
+
+#include <ngraph/function.hpp>
+#include <ngraph/opsets/opset8.hpp>
+#include <ngraph/pass/manager.hpp>
```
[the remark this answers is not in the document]
pavel-esir: done

src/common/transformations/tests/common_optimizations/flush_fp32_subnormals_to_zero_test.cpp
pavel-esir: Unfortunately there are all numbers, except `subnormal_max`. Can we leave this test as is?

rkazants (approved):

rkazants:
Need approval from plugin teams before merge.
@vladimir-paramuzov, @dmitry-gorokhov, please take a look.

src/common/transformations/tests/common_optimizations/flush_fp32_subnormals_to_zero_test.cpp
```diff
@@ -0,0 +1,377 @@
...
+#include <gtest/gtest.h>
+
+#include <openvino/core/model.hpp>
+#include <openvino/opsets/opset10.hpp>
+#include <openvino/pass/manager.hpp>
+#include <transformations/flush_fp32_subnormals_to_zero.hpp>
+#include <transformations/init_node_info.hpp>
+
+#include "common_test_utils/ngraph_test_utils.hpp"
+#include "ngraph/pass/visualize_tree.hpp"
```
rkazants, line 14: not needed header. please remove it.

rkazants, line 14:
```suggestion
```

pavel-esir: ouh, sorry for that 🤦

PR #15930, opened 2023-02-24
keep Const + DecompressionConvert for CPU

pavel-esir:
### Details:
 - We deliver Const+DecompressionConvert pattern to CPU in order to try to CF it there and/or join it other processing to reduce amount of memory access. We are doing that in order to get rid of slow ngraph constant fold and try to CF on CPU with optimized kernels and parallelism. Also on CPU plugin side dealing with this Const+Convert patterns can be joined with other processing to overcome memory boundness of this ops. 
 - If we succeed with dealing with this pattern on CPU side we expect a significant LoadTime improvement. Since the main bottleneck of FP16 LoadModel is slow ngraph evaluate during CF of this exact pattern Const+Decompression Convert.
 - **IMPORTANT:** traces show increase on the CPU side during GraphCreate, therefore transformation were commented with #todo xxx-105060. These markings should be activated after CPU will be ready to deal with unfolded Const+Decompression 

### Tickets:
 - xxx-104574, xxx-91865


maxnick:
@eshoguli , could you please take a look?

slyalin:
@itikhono, it looks like what we need as the first step to postpone f16 constants decompression. To our discussion about how to move forward with generic transformations.

@dmitry-gorokhov, the same statement for you. I believe it was assigned to @eshoguli because we assume that most of the issues will arise in LPT. Is that correct?

@eshoguli, do you have any updates on that?

akladiev:

src/common/transformations/include/transformations/mark_decompression_convert_constant_folding.hpp
```diff
@@ -0,0 +1,50 @@
...
+class TRANSFORMATIONS_API EnableDecompressionConvertConstantFolding;
+class TRANSFORMATIONS_API DisableDecompressionConvertConstantFolding;
+class TRANSFORMATIONS_API KeepConstAndDecompression;
+
+}  // namespace pass
+}  // namespace ov
+
+/**
+ * @ingroup ie_transformation_common_api
+ * @brief Disables ConstantFolding for Convert operation in compressed function.
```
rkazants, line 23:
```suggestion
 * @brief Enables ConstantFolding for Convert operation in compressed function.
```

src/common/transformations/src/transformations/mark_decompression_convert_constant_folding.cpp
```diff
@@ -0,0 +1,69 @@
+// Copyright (C) 2018-2022 Intel Corporation
```
rkazants, line 1:
```suggestion
// Copyright (C) 2018-2023 Intel Corporation
```

src/common/transformations/src/transformations/mark_decompression_convert_constant_folding.cpp
```diff
@@ -0,0 +1,69 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "transformations/mark_decompression_convert_constant_folding.hpp"
+
```
rkazants, line 6: Please consider to place fp16 compression transformations and related sources into the directory `fp16_compression` like it is done for Transpose Sinking

rkazants:
Don't we have a case when some transformation will no be applied due to remained compressed FP16 sub-graphs?

rkazants:
I think we should also run DLB scope to validate these changes

src/common/transformations/src/transformations/mark_decompression_convert_constant_folding.cpp
pavel-esir: done

src/common/transformations/include/transformations/mark_decompression_convert_constant_folding.hpp
pavel-esir: done

dmitry-gorokhov (approved):
LGTM

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -1029,6 +1030,10 @@ bool fuse_type_to_constant(const std::shared_ptr<ngraph::Node>& node,
                            const precisions_map& precisions,
                            const std::vector<Input<Node>>& consumers) {
     auto from = node->get_element_type();
+    // Consts marked with disable_constant_folding should be kept in f16 until they reach to the CPU plugin
```
dmitry-gorokhov, line 1033: I believe that will be required not only for CPU plugin, so I would propose to remove any plugin specification here.

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -1029,6 +1030,10 @@ bool fuse_type_to_constant(const std::shared_ptr<ngraph::Node>& node,
...
+    if (from == element::f16 && constant_folding_is_disabled(node))
```
dmitry-gorokhov, line 1034: Do we really need to check precision is equal f16? Isn't it enough to check `constant_folding_is_disabled(node)` condition?

src/common/transformations/src/transformations/convert_precision.cpp
pavel-esir: done

src/common/transformations/src/transformations/convert_precision.cpp
pavel-esir: done

rkazants:
here we do not have e2e test run. Did we run them?

src/core/src/bound_evaluate.cpp
```diff
@@ -247,7 +248,7 @@ bool ov::could_propagate(const Output<Node>& output, std::vector<Node*>& result)
             bool can_add = true;
             size_t arg_count = node->get_input_size();
 
-            if (arg_count == 0 && !is_type<op::v0::Constant>(node)) {
+            if ((arg_count == 0 && !is_type<op::v0::Constant>(node)) || is_decompression(node->shared_from_this())) {
```
jane-intel, line 251: Convert operations with decompression attribute will be left untouched (won't be inserted in the result vector) which would result in absence of partial values in the output of Convert. This means that some operations will have insufficient shape propagation due to lack of constant data. Which also may result in non-optimal subgraphs due to transformations being disabled for unknown shapes.

jane-intel (changes requested):

src/core/src/bound_evaluate.cpp
pavel-esir: i see. i will try to check if the node is in shape subgraphs and if so do not disable evaluate for them

pavel-esir: Done. Now i do not skip evaluate and CF for shape path. Just to clarify only thing is that if there is e.g. the following graph
 ```mermaid
graph TD;
    ShapeOf-->subgraph_for_start;
    subgraph_for_start-->Range;
    stop-->Range;
    Range-->Convert_to_FP32;
    Convert_to_FP32-->MatMul;
    input_2-->MatMul;
```

All decompression converts inside `subgraph_for_start` will be evaluaated and folded. But evaluate for `Convert_to_FP32` will not be called, it will not be folded. This is not a bug but a feature: we don't want to convert such big size data tensors with slow ngraph evaluate, we want to do it on the plugin side with fast kernels.

File: src/common/transformations/src/transformations/convert_precision.cpp [hunk 7]
```diff
@@ -1029,6 +1030,10 @@ bool fuse_type_to_constant(const std::shared_ptr<ngraph::Node>& node,
                            const precisions_map& precisions,
                            const std::vector<Input<Node>>& consumers) {
     auto from = node->get_element_type();
+    // Consts marked with disable_constant_folding should be kept in f16 until they reach to the plugin
+    if (constant_folding_is_disabled(node))
+        return false;
```
jane-intel, line 1035: Are we sure this only affects CPU and other plugins get the same graph as before?

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
```diff
@@ -255,7 +258,10 @@ void Transformations::PreLpt(const std::vector<ov::element::Type>& defaultPrecis
     }
     CPU_REGISTER_PASS_COMMON(manager, ov::pass::Validate);
     CPU_REGISTER_PASS_COMMON(manager, ov::pass::RefConvertI64ToI32);
+
+    // todo: remove this conversion line when xxx-105060 is ready and activate conversion at the end of the pipeline
```
jane-intel, line 262: Are we sure that the following common transformations and LPT would deliver same graph after we move ConvertPrecision down? It doesn't seem obvious from the code. Could you share your thought process?

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
```diff
@@ -197,6 +197,9 @@ void Transformations::PreLpt(const std::vector<ov::element::Type>& defaultPrecis
     ov::pass::Manager manager;
     manager.set_per_pass_validation(false);
     CPU_REGISTER_PASS_COMMON(manager, ov::pass::InitNodeInfo);
+    CPU_REGISTER_PASS_COMMON(manager, ov::pass::MarkShapeOfSubgraphs);
+    // todo: should be uncommented when xxx-105060 is ready
+     CPU_REGISTER_PASS_COMMON(manager, ov::pass::KeepConstAndDecompression);
```
jane-intel, line 202: What should be uncommented?

jane-intel:
Maybe file renaming was a bad idea
image

pavel-esir:
> Maybe file renaming was a bad idea image

i already opened a PR to update cuda plugin as well. https://github.com/openvinotoolkit/openvino_contrib/pull/665 and have run products config to merge both of these PR when review is finished.

I will respond to other comments shortly

src/common/transformations/src/transformations/convert_precision.cpp
pavel-esir: for constants now i use a separate `rt_info` to ensure we don't interfere with other transformations markings

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
pavel-esir: Registering `KeepConstAndDecompression` should be uncommented when CPU will be ready to deal with unfolded graphs. 
On this commit i left if uncommented by mistake. We agreed to merge mechanism in this PR without activating it yet

pavel-esir:
CI is green https://openvino-ci.toolbox.iotg.sclab.intel.com/blue/organizations/jenkins/private-ci%2F1source_innersource%2Fframeworks.ai.openvino.ci.product-configs/detail/PR-2363/9/pipeline

jane-intel (approved):
Based on the provided test reports I approve this PR

PR #15943, opened 2023-02-24
Check Node in Model creation

akuporos:
### Details:
 - Found that check for creating `Model` from `shared_ptr<Node> == nullptr` was missed (that called when a user tries to create it in python as `Model(None, [])` that caused segfault)
 - 
### Tickets:
 - 102965

### DIFF
--- a/src/core/include/openvino/core/model.hpp
+++ b/src/core/include/openvino/core/model.hpp
@@ -536,9 +536,6 @@ private:
     /// model and registers them, otherwise checks all the Parameters are registered.
     void prerequirements(bool detect_variables, bool detect_parameters);
 
-    // Check that a Node argument for ctor isn't nullptr.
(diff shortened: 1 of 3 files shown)


src/core/include/openvino/core/model.hpp
```diff
@@ -536,6 +536,9 @@ class OPENVINO_API Model : public std::enable_shared_from_this<Model> {
     /// model and registers them, otherwise checks all the Parameters are registered.
     void prerequirements(bool detect_variables, bool detect_parameters);
 
+    // Check that a Node argument for ctor isn't nullptr.
+    static const std::shared_ptr<Node>& verify_node(const std::shared_ptr<Node>& node);
+
```
ilyachur, line 541: I don't think that we need to extend public API
```suggestion
```

akuporos: done

ilyachur (approved):

PR #15957, opened 2023-02-26
[PT FE] Add aten::Sort

PiotrKrzem:
### Details:
 - Add sort implementation for PyTorch frontend
 - Based on [this previous draft PR](https://github.com/openvinotoolkit/openvino/pull/15761)

### Tickets:
 - 101081

### DIFF
--- a/src/frontends/pytorch/src/op/sort.cpp
+++ b/src/frontends/pytorch/src/op/sort.cpp
@@ -1,0 +1,52 @@
+// Copyright (C) 2018-2022 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <climits>
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
+#include "utils.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+#define mark(...) context.mark_node(__VA_ARGS__)
+#define shared    std::make_shared
+
+using namespace opset10;
+
+OutputVector translate_sort(NodeContext& context) {
+    const auto input_tensor = context.get_input(0);
+    bool stable;
+    int64_t dim;
+    bool descending;
+
+    if (context.get_input_size() == 4) {
+        stable = context.const_input<bool>(1);
+        FRONT_END_OP_CONVERSION_CHECK(stable == false, "Stable sorting in aten::sort is not yet supported.");
+        dim = context.const_input<int64_t>(2);
+        descending = context.const_input<bool>(3);
+    } else {
+        dim = context.const_input<int64_t>(1);
+        descending = context.const_input<bool>(2);
+    }
+    auto mode = descending ? ov::op::TopKMode::MAX : ov::op::TopKMode::MIN;
+
+    auto zero_axis = mark(Constant::create(element::i64, Shape({1}), {0}));
+    auto dim_axis = mark(Constant::create(element::i64, Shape({1}), {dim}));
+    auto shape = mark(shared<ShapeOf>(input_tensor));
+    auto k_values_node = mark(shared<Gather>(shape, dim_axis, zero_axis));
+    auto k_values = mark(shared<Squeeze>(k_values_node));
+
+    auto topk = mark(shared<TopK>(input_tensor, k_values, dim, mode, ov::op::TopKSortType::SORT_VALUES));
+    auto indices = mark(shared<Convert>(topk->output(1), element::i64));
+    return {topk->output(0), indices};
+};
+
+}  // namespace op
+}  // namespace pytorch
+}  // namespace frontend
+}  // namespace ov
--- a/tests/layer_tests/pytorch_tests/test_sort.py
+++ b/tests/layer_tests/pytorch_tests/test_sort.py
@@ -1,0 +1,59 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import numpy as np
+import torch
+import pytest
+
+from pytorch_layer_test_class import PytorchLayerTest
+
+
+class TestSortConstants(PytorchLayerTest):
+    def _prepare_input(self):
+        return (self.input_tensor,)
+
+    def create_model(self, dim, descending, stable):
+        class aten_sort(torch.nn.Module):
+            def __init__(self, dim, descending, stable) -> None:
+                torch.nn.Module.__init__(self)
+                self.stable = stable
+                self.dim = dim
+                self.descending = descending
+
+            def forward(self, input_tensor):
+                if self.stable is not None:
+                    return torch.sort(input_tensor, 
+                        stable = self.stable,
+                        dim = self.dim, 
+                        descending = self.descending
+                    )[0]
+                else:
+                    return torch.sort(input_tensor, 
+                        dim = self.dim, 
+                        descending = self.descending
+                    )[0]
+
+        ref_net = None
+        return aten_sort(dim, descending, stable), ref_net, "aten::sort"
+
+    @pytest.mark.parametrize("input_tensor", [
+        np.array([1, 2, 4, 6, 5]),
+        np.array([0, 1] * 9),
+        np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]]),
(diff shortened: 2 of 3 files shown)


src/frontends/pytorch/src/op/sort.cpp
```diff
@@ -0,0 +1,52 @@
+// Copyright (C) 2018-2022 Intel Corporation
```
rkazants, line 1:
```suggestion
// Copyright (C) 2018-2023 Intel Corporation
```

src/frontends/pytorch/src/op/sort.cpp
```diff
@@ -0,0 +1,52 @@
...
+        FRONT_END_OP_CONVERSION_CHECK(stable == false, "Stable sorting in aten::sort is not yet supported.");
+        dim = context.const_input<int64_t>(2);
+        descending = context.const_input<bool>(3);
+    } else {
+        dim = context.const_input<int64_t>(1);
+        descending = context.const_input<bool>(2);
+    }
+    auto mode = descending ? ov::op::TopKMode::MAX : ov::op::TopKMode::MIN;
+
+    auto zero_axis = mark(Constant::create(element::i64, Shape({1}), {0}));
```
rkazants, line 38:
```suggestion
    auto zero_axis = mark(Constant::create(element::i32, Shape{1}, 0));
```

src/frontends/pytorch/src/op/sort.cpp
```diff
@@ -0,0 +1,52 @@
...
+    auto dim_axis = mark(Constant::create(element::i64, Shape({1}), {dim}));
```
rkazants, line 39:
```suggestion
    auto dim_axis = mark(Constant::create(element::i64, Shape{1}, dim));
```

src/frontends/pytorch/src/op/sort.cpp
```diff
@@ -0,0 +1,52 @@
...
+    if (context.get_input_size() == 4) {
+        stable = context.const_input<bool>(1);
+        FRONT_END_OP_CONVERSION_CHECK(stable == false, "Stable sorting in aten::sort is not yet supported.");
+        dim = context.const_input<int64_t>(2);
+        descending = context.const_input<bool>(3);
+    } else {
+        dim = context.const_input<int64_t>(1);
+        descending = context.const_input<bool>(2);
+    }
+    auto mode = descending ? ov::op::TopKMode::MAX : ov::op::TopKMode::MIN;
```
rkazants, line 36: may be the following helps to make it shorter

```
using namespace ov::op;
```

tests/layer_tests/pytorch_tests/test_sort.py
```diff
@@ -0,0 +1,59 @@
...
+    @pytest.mark.parametrize("dim", [
+        0,
+        -1
+    ])
+    @pytest.mark.parametrize("descending", [
+        True,
+        False 
+    ])
+    @pytest.mark.parametrize("stable", [
+        False,
```
rkazants, line 54: please also test stable case. Now you need to keep it xfail and refer to the JIRA ticket. We will have to move this translator to the new version of `TopK-11` with stable sorting and switch on this case after.

src/frontends/pytorch/src/op/sort.cpp
PiotrKrzem: Applied!

src/frontends/pytorch/src/op/sort.cpp
PiotrKrzem: Similarly as in the previous comment, same story, segfault even with &dim, so I kept the {dim}

src/frontends/pytorch/src/op/sort.cpp
PiotrKrzem: Good idea, applied!

tests/layer_tests/pytorch_tests/test_sort.py
PiotrKrzem: Added the test!

src/frontends/pytorch/src/op/sort.cpp
```diff
@@ -0,0 +1,49 @@
...
+namespace pytorch {
+namespace op {
+
+#define mark(...) context.mark_node(__VA_ARGS__)
+#define shared    std::make_shared
+
+using namespace opset10;
+using namespace ov::op;
+
+OutputVector translate_sort(NodeContext& context) {
```
mvafin, line 18: This code is very similar to aten::sort from #15769 Can we merge two implementations?

src/frontends/pytorch/src/op/sort.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/opsets/opset10.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+#define mark(...) context.mark_node(__VA_ARGS__)
+#define shared    std::make_shared
```
mvafin, line 13: Please remove these defines, I don't think this is a good practice

src/frontends/pytorch/src/op/sort.cpp
```diff
@@ -0,0 +1,49 @@
...
+    auto mode = descending ? TopKMode::MAX : TopKMode::MIN;
+
+    auto zero_axis = mark(Constant::create(element::i32, Shape{1}, {0}));
+    auto dim_axis = mark(Constant::create(element::i64, Shape{1}, {dim}));
+    auto shape = mark(shared<ShapeOf>(input_tensor));
+    auto k_values_node = mark(shared<Gather>(shape, dim_axis, zero_axis));
+    auto k_values = mark(shared<Squeeze>(k_values_node));
+
+    auto topk = mark(shared<TopK>(input_tensor, k_values, dim, mode, ov::op::TopKSortType::SORT_VALUES));
+    auto indices = mark(shared<Convert>(topk->output(1), element::i64));
```
tomdol, line 42: You don't need to convert the second output to i64, TopK can be configured to produce i64 out of the box. You just need to construct TopK with non-default last argument https://github.com/openvinotoolkit/openvino/blob/master/src/core/include/openvino/op/topk.hpp#L152 setting it to element::i64. Then the return statement of this translation function can look like this:

`return topk->outputs();`

tests/layer_tests/pytorch_tests/test_sort.py
```diff
@@ -0,0 +1,76 @@
...
+                    )[0]
+
+        ref_net = None
+        return aten_sort(dim, descending, stable), ref_net, "aten::sort"
+
+    @pytest.mark.parametrize("input_tensor", [
+        np.array([1, 2, 4, 6, 5]),
+        np.array([0, 1] * 9),
+        np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]]),
+        np.array([[[1, 2, 3], [4, 5, 6], [7, 8, 9]], [[5, 2, 4], [4, 9, 0], [7, 7, 9]], [[5, 2, 4], [4, 9, 0], [7, 7, 9]]])
```
tomdol, line 43: I think you could try adding more tests for some corner cases. By that I mean 1D tensor with some repeating elements, N-D tensor with repeating slices and soring over an axis different than the last one. The Unique operator tests might give you some ideas https://github.com/tomdol/openvino/blob/master/src/plugins/template/tests/functional/op_reference/unique.cpp

tomdol (changes requested):

src/frontends/pytorch/src/op/sort.cpp
PiotrKrzem: Oki, extracted the common denominator into a sort_elements utils function

src/frontends/pytorch/src/op/sort.cpp
PiotrKrzem: Removed

src/frontends/pytorch/src/op/sort.cpp
PiotrKrzem: Great, applied, thank you!

tests/layer_tests/pytorch_tests/test_sort.py
PiotrKrzem: I added a couple more, hopefully that is enough!

tomdol (approved):

PR #15973, opened 2023-02-27
[PyOV] Remove `setuptools` upperbound

p-wysocki:
### Details:
 - Remove upperbound
 - Currently for debug purposes, to see if the bug is still present

### Tickets:
 - 101371

### DIFF
--- a/src/bindings/python/wheel/requirements-dev.txt
+++ b/src/bindings/python/wheel/requirements-dev.txt
@@ -1,3 +1,3 @@
-setuptools>=53.0.0,!=65.7.0
+setuptools>=65.6.1
 wheel>=0.38.1
 patchelf; sys_platform == 'linux' and platform_machine == 'x86_64'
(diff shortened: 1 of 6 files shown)


src/bindings/python/wheel/requirements-dev.txt
```diff
@@ -1,3 +1,3 @@
-setuptools>=53.0.0,<=65.7.0
+setuptools>=65.6.1
```
evgenytalanin-intel, line 1: Should we really have this lower bound here? I suspect PEP440-compliant stuff will work with older setuptools as well

p-wysocki: It will work with older `setuptools`, that's correct. My reason for bumping lower bound was aligning both wheels' requirements between each other (https://github.com/openvinotoolkit/openvino/pull/15971).

However, the issue this PR addresses can be solved in a much simpler way, by hardcoding POT version to `0.0.0`, which is being done in https://github.com/openvinotoolkit/openvino/pull/16021. I think this PR may need to be closed, if POT version does not matter for `openvino_dev` wheel we can indeed hardcode it. I was under the impression it has to be unique and reflect the date, commit and branch, as it was before.

PR #15980, opened 2023-02-27
[GPU] Fix Kernel clone to use copy constructor instead of clCreateKernel

yeonbok:
### Details:
 - Fix Kernel clone to use copy constructor instead of clCreateKernel

### Tickets:
 - 104925

### DIFF
--- a/src/plugins/intel_gpu/src/runtime/ocl/ocl_ext.hpp
+++ b/src/plugins/intel_gpu/src/runtime/ocl/ocl_ext.hpp
@@ -967,8 +967,7 @@ public:
     KernelIntel(const Kernel &other, const UsmHelper& usmHelper) : Kernel(other), _usmHelper(usmHelper) { }
 
     KernelIntel clone() const {
-        Kernel cloned_kernel(this->getInfo<CL_KERNEL_PROGRAM>(), this->getInfo<CL_KERNEL_FUNCTION_NAME>().c_str());
-        return KernelIntel(cloned_kernel, _usmHelper);
+        return KernelIntel(*this, _usmHelper);
     }
 
     cl_int setArgUsm(cl_uint index, const UsmMemory& mem) {


src/plugins/intel_gpu/src/runtime/ocl/ocl_ext.hpp
```diff
@@ -967,8 +967,7 @@ class KernelIntel : public Kernel {
     KernelIntel(const Kernel &other, const UsmHelper& usmHelper) : Kernel(other), _usmHelper(usmHelper) { }
 
     KernelIntel clone() const {
-        Kernel cloned_kernel(this->getInfo<CL_KERNEL_PROGRAM>(), this->getInfo<CL_KERNEL_FUNCTION_NAME>().c_str());
-        return KernelIntel(cloned_kernel, _usmHelper);
+        return KernelIntel(*this, _usmHelper);
```
vladimir-paramuzov, line 970: As far as I remember, this change will lead to usage of the same kernel handle from multiple streams and data races. So I think we need to have more careful handling for cached kernels in dynamic case.  

Maybe we can just replace unique_ptr with shared_ptr for `primitive_inst::_impl` ? In that case we can just take shared ptr from impls cache w/o any kind of clone

yeonbok: Thanks for the comment, and I would like to know more exactly.
Seeing the code, Kernel's owner is still the initial impl, and isnt this just increasing reference count for the kernel?

vladimir-paramuzov: @yeonbok  kernel is initially owned by impl created for program_node which is then cloned for each stream in primitive_inst c-tor:
```
, _impl(node.get_selected_impl() ? node.get_selected_impl()->clone() : nullptr)
```
So if you change kernel's clone() impl, then multiple implementation instances will refer to the same kernel object as copy ctor just increase ref count while current version does deep copy. And then an attempt to call set_arguments for an impl from multiple streams will cause data race

yeonbok: Thanks for elaborating. I see, the argument field will be shared within streams :( 
Then, "using shared_ptr for same impl and not clone" from multiple streams" <= Isnt this with the same issue? 
Currently impl cache is also being shared b/w streams.. 
Maybe we need to still clone from impl_cache if the stream id > 1 ?

vladimir-paramuzov: > Maybe we need to still clone from impl_cache if the stream id > 1 ?

I guess we need to try impls cache with 2 levels: per-stream and per-model. So when we trigger async compilation it adds impl to global (per model) cache. When primitive_inst tries to find cached impl, it does something like that:
1. Check local (per-stream) cache
2. In case of local cache hit - get ptr w/o kernel clone (ideal case)
3. In case of local cache miss - go to global cache
4. In case of global cache hit - clone kernel to local cache and then use it
5. In case of global cache miss - use shape agnostic impl or wait for compilation completion

akladiev:

akladiev:

PR #15989, opened 2023-02-28
Fix warning: sequence point strict aliasing

xipingyan:
### Details:
 - *Remove: ie_add_compiler_flags(-Wno-sequence-point)*
 - *Remove: ie_add_compiler_flags(-Wno-strict-aliasing)*
 - *Fix corresponding warnings*

### Tickets:
 - *104795*

### DIFF
--- a/src/plugins/intel_cpu/src/utils/blob_dump.cpp
+++ b/src/plugins/intel_cpu/src/utils/blob_dump.cpp
@@ -176,7 +176,9 @@ void BlobDumper::dumpAsTxt(std::ostream &stream) const {
             for (size_t i = 0; i < data_size; i++) {
                 int i16n = blob_ptr[desc.getElementOffset(i)];
                 i16n = i16n << 16;
-                float fn = *(reinterpret_cast<const float *>(&i16n));
+                float fn;
+                assert(sizeof(fn) == sizeof(i16n));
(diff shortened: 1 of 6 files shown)


src/plugins/intel_cpu/src/utils/blob_dump.cpp
```diff
@@ -176,7 +176,9 @@ void BlobDumper::dumpAsTxt(std::ostream &stream) const {
             for (size_t i = 0; i < data_size; i++) {
                 int i16n = blob_ptr[desc.getElementOffset(i)];
                 i16n = i16n << 16;
-                float fn = *(reinterpret_cast<const float *>(&i16n));
+                float fn;
```
EgorDuplensky, line 179: Can't we just work with bloat16_t type instead?

xipingyan: Sorry @EgorDuplensky , I am not sure, if dump bfloat16_t type, it will not need to convert.
Hi @maxnick , I see you modified it. What's your idea? If we dump it with bf16 directly, or is it still readable ?

maxnick: `bfloat16_t ` has `operator float()` so I expect the number implicitly converted to float type and dumped normally.

maxnick (approved):

maxnick (approved):

PR #15996, opened 2023-02-28
Fix MO IR Reader for Eye op

mvafin:
### Details:
 - *Eye op has `output_type` attribute, we need to read it as numpy type*

### Tickets:
 - *104163*


rkazants (dismissed):

mvafin:
Fix is incomplete

PR #16003, opened 2023-02-28
[GNA] Fix crash in GNA plugin when using POT

marcinkusm:
### Details:
 - added extra checks for condition fixing infinity loop

### Tickets:
 - 104500

### DIFF
--- a/src/plugins/intel_gna/src/frontend/model_quantizer.hpp
+++ b/src/plugins/intel_gna/src/frontend/model_quantizer.hpp
@@ -148,13 +148,28 @@ private:
                 inf_loop_count++;
             } else {
                 if (inf_loop_count > 0 &&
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gna/src/frontend/model_quantizer.hpp
```diff
@@ -148,13 +148,28 @@ class ModelQuantizer {
                 inf_loop_count++;
             } else {
                 if (inf_loop_count > 0 &&
-                    (inf_loop_history.size() % inf_loop_pattern.size() == 0 || sf.allLayersProcessed()) &&
```
marcinkusm, line 151: if we decide to add only extra condition check to prevent crash "(inf_loop_history.size() > inf_loop_pattern.size())" 
we can see exception with infinite loop.

marcinkusm:
Please let me know if PR is acceptable

mryzhov (approved):

dorloff (approved):

PR #16015, opened 2023-03-01
Runtime fallback to other devices

WeldonWangwang:
### Details:
 - for auto plugin, once inference failed from the underline hardware plugins, we will switch to the next candidate device(if any) to continue the inference workloads

### Tickets:
 - 100082

### Test Cases:

|Test Case|ov::intel_auto::enable_runtime_fallback|GPU.0|GPU.1|VPUX|CPU|except LoadNetwork num|
|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
|1|YES|infer failed|infer failed| | |2|
|2|YES|infer failed|infer success| | |2|
|3|YES|infer success|infer failed| | |1|
|4|YES|infer success|infer success| | |1|
|5|YES|infer success| | |infer success|2|
|6|YES|infer failed| | |infer success|2|
|7|YES|infer success| | |infer failed|2|
|8|YES|infer failed| | |infer failed|2|
|9|YES|infer success|infer success|infer success| |1|
|10|YES|infer failed|infer success|infer success| |2|
|11|YES|infer failed|infer failed|infer success| |3|
|12|YES|infer failed|infer failed|infer failed| |3|
|13|YES|infer success|infer success| |infer success|2|
|14|YES|infer failed|infer success| |infer success|2|
|15|YES|infer failed|infer failed| |infer success|2|
|16|YES|infer success|infer success| |infer failed|2|
|17|YES|infer failed|infer success| |infer failed|3|
|18|YES|infer failed|infer failed| |infer failed|3|
|19|NO|infer failed|infer failed| | |1|
|20|NO|infer failed|infer success| | |1|
|21|NO|infer success|infer failed| | |1|
|22|NO|infer success|infer success| | |1|
|23|NO|infer success| | |infer success|2|
|24|NO|infer failed| | |infer success|2|
|25|NO|infer success| | |infer failed|2|
|26|NO|infer failed| | |infer failed|2|
|27|NO|infer success|infer success|infer success| |1|
|28|NO|infer failed|infer success|infer success| |1|
|29|NO|infer failed|infer failed|infer success| |1|
|30|NO|infer failed|infer failed|infer failed| |1|
|31|NO|infer success|infer success| |infer success|2|
|32|NO|infer failed|infer success| |infer success|2|
|33|NO|infer failed|infer failed| |infer success|2|
|34|NO|infer success|infer success| |infer failed|2|
|35|NO|infer failed|infer success| |infer failed|2|
|36|NO|infer failed|infer failed| |infer failed|2|


src/inference/include/openvino/runtime/properties.hpp
```diff
@@ -142,7 +142,19 @@ class Property : public util::BaseProperty<T, mutability_> {
...
+            if (std::is_unsigned<U>::value) {
+                if (value < 0)
+                    throw ov::Exception{"Expecting positive value for unsigned type"};
+            }
+            return value;
+        }
+
+        template <typename U,
+                  typename std::enable_if<!std::is_same<typename std::decay<U>::type, std::string>::value &&
+                                              !std::is_convertible<V, std::string>::value && !std::is_signed<V>::value,
```
songbell, line 157: you don't need this, can remove

WeldonWangwang: Okay

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,75 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
-                    IE::Task t;
-                    do {
-                        _inferPipelineTasks.try_pop(t);
-                    } while (t && ScheduleToWorkerInferRequest(std::move(t)));
-                    do {
-                        _inferPipelineTasksDeviceSpecific[device]->try_pop(t);
-                    } while (t && ScheduleToWorkerInferRequest(std::move(t), device));
+                    if (workerRequestPtr->_exceptionPtr != nullptr) {
+                        if (workerRequestPtr->_testExec->count == 1) {
+                            auto capturedTask = std::move(workerRequestPtr->_task);
```
songbell, line 84: does this flag means we only try once after failure? I think we don't need to pose this restriction here in infer request?

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,75 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+                            capturedTask();
+                        } else {
+                            workerRequestPtr->_reload = true;
+                            auto capturedTask = std::move(workerRequestPtr->_testExec->_task);
+                            workerRequestPtr->_testExec->count++;
+                            selectOtherDevice(workerRequestPtr->_deviceName);
```
songbell, line 90: what if selectOtherDevice fails, throw exception, or hang unexpectedly

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,75 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+                            _inferPipelineTasksDeviceSpecific[device]->try_pop(t);
+                        } while (t && ScheduleToWorkerInferRequest(std::move(t), device));
+                    }
                 }
             });
     }
 }
 
+void AutoSchedule::selectOtherDevice(std::string currentDeviceName) {
+    std::unique_lock<std::mutex> lck(_mutex);
```
songbell, line 120: we should implement this to be called only once, like if many infer requests fails, we don;t need to load fallback device for several times.

src/plugins/auto/auto_schedule.cpp
```diff
@@ -176,14 +227,19 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
             _loadContext[CPU].isEnabled = false;
         }
     }
+    _loadContext[FALLBACKDEVICE].usefeature = false;
```
songbell, line 230: this is for a switch from config for use to control right? are we gonna to introduce a property?

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,75 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+                        if (workerRequestPtr->_testExec->count == 1) {
+                            auto capturedTask = std::move(workerRequestPtr->_task);
+                            capturedTask();
+                        } else {
+                            workerRequestPtr->_reload = true;
+                            auto capturedTask = std::move(workerRequestPtr->_testExec->_task);
+                            workerRequestPtr->_testExec->count++;
+                            selectOtherDevice(workerRequestPtr->_deviceName);
+                            if (_loadContext[FALLBACKDEVICE].isAlready) {
+                                workerRequestPtr->_startTimes.pop_back();
```
songbell, line 92: why remove starttime here

src/plugins/auto/multi_schedule.cpp
```diff
@@ -154,8 +165,19 @@ void MultiSchedule::GenerateWorkers(const std::string& device,
                 IdleGuard<NotBusyWorkerRequests> idleGuard{workerRequestPtr, *idleWorkerRequestsPtr};
                 workerRequestPtr->_exceptionPtr = exceptionPtr;
                 {
-                    auto capturedTask = std::move(workerRequestPtr->_task);
-                    capturedTask();
+                    if (workerRequestPtr->_exceptionPtr != nullptr) {
+                        if (workerRequestPtr->_reload) {
```
songbell, line 169: multi will be merged with auto logic soon, probably we can leave multi unchanged for now, just focus on auto

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Because if reload success, the original request will only has start time, it will assert in https://github.com/openvinotoolkit/openvino/blob/0860db0dc3e366e6cb54c03943de1ef3710a9afa/src/plugins/auto/multi_schedule.cpp#L241

src/plugins/auto/multi_schedule.cpp
WeldonWangwang: Okay.

src/plugins/auto/auto_schedule.cpp
songbell: then I think we should fill in a endtime properly for this infer request, isn't it?

WeldonWangwang: Agree

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Yes, if there is no other devices in _devicePriorities and can retry in many times, it will hang here. I think it' better to retry once after selectOtherDevice.

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: if there is no other devices in selectOtherDevice , it will retry once and end.
if load failed in selectOtherDevice, it will hang here.

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: If re-infer and failed again, I think it need to retry another devices again, so just add some lock here but it can be called in many times.

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Yes, this is a switch to _loadContext[FALLBACKDEVICE] feature, Add this switch because I want to use the original _loadContext[i].task. But _loadContext[FALLBACKDEVICE] can not use 'future'

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Modified the logic that notify the condition_variable and added some exception handling

src/tests/unit/auto/plugin/mock_infer_request.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <gmock/gmock.h>
+#include "ie_icore.hpp"
+#include "plugin.hpp"
+#include <iostream>
+
+using namespace MockMultiDevicePlugin;
+namespace MockMultiDevice {
+
+struct mockRequestExecutor : public IE::ITaskExecutor {
+public:
```
songbell, line 15: I think this is same as ImmediateExecutor?

WeldonWangwang: Yes, will replace it with ImmediateExecutor. Thanks.

src/plugins/auto/auto_schedule.cpp
songbell: I mean this situation, say if we have infer_request_1_gpu and infer_request_2_gpu both failed in separate threads, no matter which one triggered the selectOtherDevice, we only expect this select to be called only once, and the other one just need to re-trigger the inference. (all these operations need to be thread-safe)

songbell: also, some other points may need consider, say we have fallbackdevice successfully enabled, shall we release the resources of actualdevice?

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,98 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
-                    } while (t && ScheduleToWorkerInferRequest(std::move(t)));
-                    do {
-                        _inferPipelineTasksDeviceSpecific[device]->try_pop(t);
-                    } while (t && ScheduleToWorkerInferRequest(std::move(t), device));
+                    auto stopRetryAndContinue = [workerRequestPtr]() {
+                        auto capturedTask = std::move(workerRequestPtr->_task);
+                        capturedTask();
+                    };
+                    if (workerRequestPtr->_exceptionPtr != nullptr && _autoSContext->_runtimeFallback) {
+                        if (workerRequestPtr->_testExec->count == 1) {
```
songbell, line 87: btw, when will this if be triggered, if needed, able to use bool?

src/plugins/auto/auto_schedule.cpp
```diff
@@ -116,6 +189,10 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
     // loadContext[ACTUALDEVICE] is always enabled,
     // when there is CPU and there are more than two devices, loadContext[CPU] is enabled
     _loadContext[ACTUALDEVICE].isEnabled = true;
+    if (_autoSContext->_runtimeFallback) {
+        _loadContext[FALLBACKDEVICE].isEnabled = true;
+        _loadContext[FALLBACKDEVICE].usefeature = false;
```
songbell, line 194: do you mean usefuture?
is it able to use future also for the fallback task

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Added the result of `selectOtherDevice()`, so rm the count flag now.

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Added the lock to whole `selectOtherDevice()`, other infer request will need to wait and will get the result of previous state that show if load successfully.

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Yes, Added call_once func to release ACTUALDEVICE when FALLBACKDEVICE is ready.

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,98 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+                            try {
+                                selectOtherDeviceFlag = selectOtherDevice(workerRequestPtr->_deviceName);
+                            } catch (const IE::Exception&) {
+                                selectOtherDeviceFlag = false;
+                            }
+                            if (selectOtherDeviceFlag) {
+                                workerRequestPtr->_reload = true;
+                                workerRequestPtr->_testExec->count = 0;
+                                // Add end time to current workerRequest
+                                workerRequestPtr->_endTimes.push_back(std::chrono::steady_clock::now());
```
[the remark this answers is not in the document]
WeldonWangwang: AUTO need to summery the duration of each infer request. so we need to add endtime or rm starttime of failed inferrquest. Need to discuss which is better.

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Remove this flag now.

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Consider some scenarios where switching to fallbackdevice and reinfer failed, so it need _loadContext[FALLBACKDEVICE] load again, if we want use the _loadContext[i].task again, if use promise, _loadContext[i].task cannot be ran again.

WeldonWangwang: Replaced `usefeature ` with `usePromise`

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,113 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+                                            return d.deviceName.find(currentDeviceName) != std::string::npos;});
+        if (CurrentDeviceIter != _autoSContext->_devicePriorities.end()) {
+            if (_autoSContext->_devicePriorities.size() == 1) {
+                LOG_INFO_TAG("No other devices in _devicePriorities");
+                return false;
+            }
+            _autoSContext->_devicePriorities.erase(CurrentDeviceIter);
+        } else {
+            LOG_INFO_TAG("Alreay selected fallback device");
+            return _loadContext[FALLBACKDEVICE].isReloadSuccess ? true : false;
```
songbell, line 162: at the time of return, the fallback device may not finish loading, this infer request will fail still

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,113 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+        }
+        _loadContext[FALLBACKDEVICE].isLoadSuccess = false;
+        _loadContext[FALLBACKDEVICE].workName = "";
+        _loadContext[FALLBACKDEVICE].isReloadSuccess = false;
+        _loadContext[FALLBACKDEVICE].deviceInfo =
+            _autoSContext->_plugin->SelectDevice(_autoSContext->_devicePriorities,
+                                                    _loadContext[FALLBACKDEVICE].networkPrecision,
+                                                    _autoSContext->_modelPriority);
+        _executor->run(_loadContext[FALLBACKDEVICE].task);
```
songbell, line 171: can you help check if runAndWait() can simply your implementation?
void ITaskExecutor::runAndWait(const std::vector<Task>& tasks) {
    run_and_wait(tasks);
}

src/plugins/auto/auto_schedule.cpp
songbell: probably need to check the load finished here: _cond.wait(lck, []{return loadfinished;});

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,113 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+        std::unique_lock<std::mutex> lck(_mutex);
+        _cond.wait(lck);
+        if (_loadContext[FALLBACKDEVICE].isAlready) {
```
songbell, line 174: btw, with the _fallbackMutex, is it possible that other threads may even run to reach this line?

WeldonWangwang: Other threads can not reach this line with _fallbackMutex.

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,113 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+            std::call_once(_firstReleaseActualDevice, [currentDeviceName, this]() {
+                releaseActualdevice(currentDeviceName);
+            });
+        }
```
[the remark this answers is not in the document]
WeldonWangwang: Okay

src/tests/unit/auto/auto_runtime_fallback_test.cpp
```diff
@@ -0,0 +1,281 @@
...
+            .Times(loadNetworkNum);
+    }
+
+    std::shared_ptr<InferenceEngine::IExecutableNetworkInternal> exeNetwork;
+
+    ASSERT_NO_THROW(exeNetwork = plugin->LoadExeNetworkImpl(cnnNet, config));
+
+    std::shared_ptr<IInferRequestInternal> infer_request;
+    ASSERT_NO_THROW(infer_request = exeNetwork->CreateInferRequest());
+    ASSERT_NO_THROW(infer_request->StartAsync());
```
songbell, line 257: when disable ENABLE_RUNTIME_FALLBACK, to me, some exceptions should be expected, while I did not see this situation been tested?
some test scenarios maybe:
1) runtime_fallback enabled
    a. helper loaded first, helper failed
    b. helper loaded first, actual failed
    c. actual loaded first, actual failed
    d. helper been disabled, actual failed
    d. cumulative situation(TBD)
2) runtime_fallback disabled
    exceptions will be seen no matter which failed

wangleis: In CTPUT situation, if inference on one device fail, AUTO remove this device from worker list and infer current input on another device.

For example, AUTO load network to GPU.1, GPU.0, VPUX and CPU. If inference on VPUX fail, AUTO remove VPUX from worker list, current input in pipeline will infer in next worker.

wangleis (approved):

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Yes, replace it with runAndWait() and have tested.

src/tests/unit/auto/auto_runtime_fallback_test.cpp
WeldonWangwang: Added these cases.

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,101 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+                    }
                 }
             });
     }
 }
 
+bool AutoSchedule::selectOtherDevice(std::string currentDeviceName) {
+    {
+        std::lock_guard<std::mutex> lock(_autoSContext->_fallbackMutex);
+        bool isCPUHELP = false;
```
riverlijunjie, line 125:
```suggestion
        bool isCPUHelp = false;
```

src/plugins/auto/auto_executable_network.cpp
```diff
@@ -17,8 +17,12 @@ AutoExecutableNetwork::AutoExecutableNetwork(AutoScheduleContext::Ptr& context,
 }
 
 std::shared_ptr<IE::RemoteContext> AutoExecutableNetwork::GetContext() const {
-    _autoSchedule->WaitActualNetworkReady();
-    return _autoSchedule->_loadContext[ACTUALDEVICE].executableNetwork->GetContext();
+    if (_autoSchedule->_loadContext[FALLBACKDEVICE].isAlready) {
+        return _autoSchedule->_loadContext[FALLBACKDEVICE].executableNetwork->GetContext();
+    } else {
+        _autoSchedule->WaitActualNetworkReady();
+        return _autoSchedule->_loadContext[ACTUALDEVICE].executableNetwork->GetContext();
```
riverlijunjie, line 24: Is there any race condition in case of one thread to call GetContext and other thread is handling runtime fallback?

WeldonWangwang: Added lock here and selectOtherDevice() to avoid this. Please review.

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Fixed it.

src/plugins/auto/auto_schedule.hpp
```diff
@@ -18,6 +18,8 @@ struct AutoLoadContext {
     std::atomic<bool> isEnabled = {false};
     std::atomic<bool> isAlready = {false};
     std::atomic<bool> isLoadSuccess = {false};
+    std::atomic<bool> isReloadSuccess = {false};
+    std::atomic<bool> ableReload = {false};
```
songbell, line 22: do we really need this flag？ looks like without this, a future/promise will be enabled, but this seems will not be harmful

src/plugins/auto/auto_schedule.hpp
```diff
@@ -18,6 +18,8 @@ struct AutoLoadContext {
     std::atomic<bool> isEnabled = {false};
     std::atomic<bool> isAlready = {false};
     std::atomic<bool> isLoadSuccess = {false};
+    std::atomic<bool> isReloadSuccess = {false};
```
songbell, line 21: do we really need this flag？ can't we use isLoadSuccess for loadContext[FALLBACKDEVICE]

src/plugins/auto/auto_schedule.cpp
songbell: so I suppose the call_once is not needed?

songbell: also, for optimization, I want to see if this thread can share common code with cpu_help release, with the target release device name as parameter

src/plugins/auto/auto_schedule.hpp
WeldonWangwang: This flag to show if load fallback device ready, isLoadSuccess is set before GenerateWorkers, so add this flag to show the fallback device is ready after GenerateWorkers, just like isAlready flag. but if we use isAlready flag, if the infer failed again in fallback device, it can't reset isAlready to false because there maybe other infer still work in the pipeline, these infer request will use isAlready  flag in ScheduleToWorkerInferRequest(). so must keep isAlready always ok if use fallback device.

riverlijunjie (approved):
LGTM

src/plugins/auto/auto_schedule.hpp
songbell: I see, it's about the future can only be used once. 
So why not just check the FALLBACKDEVICE enum, disable future/promise when the context is AutoLoadContextIndex::FALLBACKDEVICE

WeldonWangwang: Yes, this will be the latest modification :)

src/plugins/auto/auto_schedule.cpp
```diff
@@ -218,14 +301,34 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
...
+        _loadContext[ACTUALDEVICE].isLoadSuccess = false;
+        _loadContext[ACTUALDEVICE].isAlready = false;
+        _loadContext[ACTUALDEVICE].future.wait();
+        // clean up ACTUALDEVICE infer requests
+        size_t destroynum = 0;
+        std::pair<int, WorkerInferRequest*> worker;
+        while (_idleWorkerRequests[currentDeviceName].try_pop(worker)) {
+            destroynum++;
+        }
+        if (destroynum == _workerRequests[currentDeviceName].size()) {
```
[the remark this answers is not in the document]
songbell: so I guess  we have to wait for all workers to complete (probably STATUS_ONLY in failed situation?) before the while loop.

src/plugins/auto/auto_schedule.cpp
```diff
@@ -301,6 +406,23 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
...
+                    static_cast<int>(std::thread::hardware_concurrency()) /* max possible #streams*/,
+                    0 /*default threads per stream, workaround for ticket 62376*/,
+                    IE::IStreamsExecutor::ThreadBindingType::NONE});
+        for (auto&& device : _autoSContext->_devicePriorities) {
+            // initialize containers before run async task
+            _idleWorkerRequests[device.deviceName];
+            _workerRequests[device.deviceName];
+            _inferPipelineTasksDeviceSpecific[device.deviceName] = nullptr;
+        }
+        _executor->run(_loadContext[ACTUALDEVICE].task);
```
songbell, line 425: the _passthroughExeNet is enabled for single device from memory copy saving perspective (to pass through infer requests directly to hardware infer requests), we shall not break this rule. with this implementation, extra memcpy will be introduced.

src/plugins/auto/auto_schedule.cpp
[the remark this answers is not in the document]
WeldonWangwang: Yes, we can use STATUS_ONLY, added it.

src/plugins/auto/auto_schedule.cpp
```diff
@@ -301,6 +406,23 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
...
         };
         _executor->run(std::move(recycleTask));
+    } else if (_autoSContext->_devicePriorities.size() != 1 && !isCumulative && _autoSContext->_runtimeFallback) {
+        _firstLoadFuture = _firstLoadPromise.get_future();
+        // will not wait for loading accelerator network,
+        // so the executor can't be destroyed before finished the task,
+        // so use executor as a member of AutoSchedule.
+        _executor = _autoSContext->_plugin->executorManager()->getIdleCPUStreamsExecutor(
+                    IE::IStreamsExecutor::Config{"AutoDeviceAsyncLoad",
+                    static_cast<int>(std::thread::hardware_concurrency()) /* max possible #streams*/,
```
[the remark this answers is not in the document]
WeldonWangwang: Any suggestions? It just keep the same func as AUTO generate executor.

src/plugins/auto/auto_schedule.cpp
```diff
@@ -79,25 +79,101 @@ void AutoSchedule::GenerateWorkers(const std::string& device,
...
+        if (currentDeviceName == "CPU_HELP") {
+            currentDeviceName = "CPU";
+            isCPUHelp = true;
+            WaitActualNetworkReady();
+        }
+        const auto CurrentDeviceIter = std::find_if(_autoSContext->_devicePriorities.begin(), _autoSContext->_devicePriorities.end(),
+                                            [=](const DeviceInformation& d) -> bool {
+                                            return d.deviceName.find(currentDeviceName) != std::string::npos;});
+        if (CurrentDeviceIter != _autoSContext->_devicePriorities.end()) {
+            if (_autoSContext->_devicePriorities.size() == 1) {
```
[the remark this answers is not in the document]
WeldonWangwang: currentDeviceName seems can not be find in _devicePriorities if _autoSContext->_devicePriorities.size() == 0?

src/plugins/auto/multi_schedule.cpp
```diff
@@ -54,10 +54,11 @@ Pipeline MultiSchedule::GetPipeline(const IInferPtr& syncInferRequest, WorkerInf
             }
         });
     } else {
+        MultiImmediateExecutor::Ptr _FirstExecutor = std::make_shared<MultiImmediateExecutor>();
         pipeline = {
             // if the request is coming with device-specific remote blobs make sure it is scheduled to the specific device only:
             Stage {
-                /*TaskExecutor*/ std::make_shared<IE::ImmediateExecutor>(), /*task*/ [this, &syncInferRequest]() {
+                /*TaskExecutor*/ _FirstExecutor, /*task*/ [this, &syncInferRequest]() {
```
songbell, line 61: let's keep in mind another AR here after guozhong's PR to integrate multi to auto. @WeldonWangwang @wgzintel 
let's add one extra step in this first stage,  disable runtime fallback when remote blob is involved.
like -d AUTO:GPU.0, GPU.1 and the input is remote blob for GPU.0, if fallback to GPU.1, the inference cannot infer anyway, so just disable it and let the exception throw.

src/plugins/auto/auto_schedule.hpp
```diff
@@ -52,17 +54,21 @@ class AutoSchedule : public MultiSchedule {
...
     void WaitFirstNetworkReady();
     void TryToLoadNetWork(AutoLoadContext& context, const std::string& modelPath, const IE::CNNNetwork& network);
+    bool selectOtherDevice(std::string currentDeviceName);
+    IE::Task releaseActualdeviceTask;
 
 private:
     IE::IStreamsExecutor::Ptr                _executor;
     mutable std::once_flag                   _oc;
     std::once_flag                           _firstLoadOC;
+    std::once_flag                           _firstReleaseActualDevice;
```
songbell, line 71: can remove

src/plugins/auto/auto_schedule.hpp
```diff
@@ -52,17 +54,21 @@ class AutoSchedule : public MultiSchedule {
...
     bool ScheduleToWorkerInferRequest(IE::Task, DeviceName preferred_device = "") override;
-    static bool RunPipelineTask(IE::Task& inferPipelineTask, NotBusyPriorityWorkerRequests& idleWorkerRequests, const DeviceName& preferred_device);
+    static bool RunPipelineTask(IE::Task& inferPipelineTask, NotBusyPriorityWorkerRequests& idleWorkerRequests,
+                                const DeviceName& preferred_device);
     DeviceMap<NotBusyPriorityWorkerRequests> _idleWorkerRequests;
 
 private:
     void WaitFirstNetworkReady();
     void TryToLoadNetWork(AutoLoadContext& context, const std::string& modelPath, const IE::CNNNetwork& network);
+    bool selectOtherDevice(std::string currentDeviceName);
```
songbell, line 64:
```suggestion
    bool selectOtherDevice(const std::string& currentDeviceName);
```

src/plugins/auto/auto_schedule.hpp
WeldonWangwang: Removed

src/plugins/auto/multi_schedule.hpp
```diff
@@ -16,12 +16,14 @@
 
 namespace MultiDevicePlugin {
 struct ThisRequestExecutor : public IE::ITaskExecutor {
-    explicit ThisRequestExecutor(WorkerInferRequest** ptr): _workptrptr{ptr} {}
+    explicit ThisRequestExecutor(WorkerInferRequest** ptr, MultiImmediateExecutor::Ptr test = nullptr): _workptrptr{ptr}, _testexecutor(test) {}
```
songbell, line 19: change all these variables start with test to some other meaningful name, like _fallbackExecutor, _fallbackExec

src/plugins/auto/multi_schedule.hpp
```diff
@@ -54,14 +56,14 @@ class MultiSchedule : public Schedule, public IE::ITaskExecutor {
...
     DeviceMap<std::vector<WorkerInferRequest>>                _workerRequests;
-    mutable std::mutex                                        _mutex;
     std::atomic_size_t                                        _numRequestsCreated = {0};
     MultiScheduleContext::Ptr                                 _multiSContext;
     SoExecNetwork                                             _passthroughExeNet;
     Time                                                      _cpuHelpReleaseTime;
     unsigned int                                              _cpuHelpInferCount = 0;
     double                                                    _cpuHelpFps = 0.0;
     std::string                                               _LogTag;
+    MultiImmediateExecutor::Ptr                               _FirstExecutor;
```
songbell, line 66: inital to lower case

songbell (approved):

src/plugins/auto/auto_schedule.cpp
```diff
@@ -218,7 +310,11 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
                         });
                     }
                 }
-                contextPtr->promise.set_value();
+                try {
+                    contextPtr->promise.set_value();
```
songbell, line 314: something minor you can have a try if want to get rid of the try block, before trying to load for second failure, clear the promise:
contextPtr->promise = {};
and in loading:
contextPtr->future = contextPtr->promise.get_future().share();

PR #16018, opened 2023-03-01
[GPU] Add clDNN shape agnostic kernels usage as an initial impls for dGPU

sshlyapn:
### Details:
 - Use clDNN's shape agnostic kernels for dGPU as default dynamic impls
 - Add LRU weights caching

### Tickets:
 - 104960, 105028

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/runtime/lru_cache.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/runtime/lru_cache.hpp
@@ -34,7 +34,7 @@ public:
     /**
      * @brief Get the least recently used element with key and value pair in the cache
      *
-     * @return Value
+     * @return std::pair<Key, Value>
      */
     std::pair<Key, Value> get_lru_element() const {
         if (_lru_data_list.size()) {
@@ -45,11 +45,24 @@ public:
     }
 
     /**
+     * @brief Get the most recently used element with key and value pair in the cache
+     *
+     * @return std::pair<Key, Value>
+     */
+    std::pair<Key, Value> get_recent_element() const {
+        if (_lru_data_list.size()) {
+            return _lru_data_list.front();
+        } else {
+            return std::make_pair(Key(), Value());
+        }
+    }
+
+    /**
      * @brief Add new value with associated key into the LRU cache
      *
      * @param key if same key is existed in the cache, the value of key is updated new entry.
      * @param value
-     * @return true, if cache is full and lease recently used entry are removed to add new entry.
+     * @return true, if cache is full and least recently used entry are removed to add new entry.
      * @return false Otherwise
      */
     bool add(const Key& key, const Value& value) {
@@ -124,6 +137,15 @@ public:
     }
 
     /**
+     * @brief Return whether the cache is full or not
+     *
+     * @return true, if cache is full, false otherwise
+     */
+    size_t is_full() const {
+        return _lru_data_list.size() == _capacity;
+    }
+
+    /**
      * @brief Get the all keys object
      *
      * @return std::vector<Key>
@@ -154,7 +176,7 @@ private:
     }
 
     /**
-     * @brief Pop n lease recently used cache data.
+     * @brief Pop n least recently used cache data.
      *
      * @param n number of data to be popped
      */
--- a/src/plugins/intel_gpu/include/intel_gpu/runtime/lru_cache.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/runtime/lru_cache.hpp
@@ -45,19 +45,6 @@ public:
     }
 
     /**
-     * @brief Get the most recently used element with key and value pair in the cache
-     *
-     * @return std::pair<Key, Value>
-     */
-    std::pair<Key, Value> get_recent_element() const {
-        if (_lru_data_list.size()) {
-            return _lru_data_list.front();
-        } else {
-            return std::make_pair(Key(), Value());
-        }
-    }
-
-    /**
      * @brief Add new value with associated key into the LRU cache
      *
      * @param key if same key is existed in the cache, the value of key is updated new entry.
--- a/src/plugins/intel_gpu/src/graph/include/fully_connected_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/fully_connected_inst.h
@@ -53,8 +53,10 @@ public:
     typed_primitive_inst(network& network, fully_connected_node const& node);
 
     memory::ptr weights_memory() const {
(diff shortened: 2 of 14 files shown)


src/plugins/intel_gpu/src/graph/include/primitive_inst.h
```diff
@@ -260,6 +261,9 @@ class primitive_inst {
 
     std::vector<memory::cptr> _intermediates_memory;
 
+    static const size_t _weights_cache_capacity = 2;
```
vladimir-paramuzov, line 264: I think that capacity should be based on global mem size. We need to ensure that (1 + capacity)*total_weights_size < global_mem_size. Otherwise, single weights copy must be stored in the cache. 
`+1` for capacity is needed as we store original constant too

src/plugins/intel_gpu/src/graph/include/primitive_inst.h
```diff
@@ -260,6 +261,9 @@ class primitive_inst {
...
+    LruCache<size_t, memory::ptr> _reordered_weights_cache;
```
vladimir-paramuzov, line 265: Use `layout` as a key to avoid hash collisions

src/plugins/intel_gpu/src/graph/include/fully_connected_inst.h
```diff
@@ -53,7 +53,11 @@ class typed_primitive_inst<fully_connected> : public typed_primitive_inst_base<f
     typed_primitive_inst(network& network, fully_connected_node const& node);
 
     memory::ptr weights_memory() const {
-        return is_dynamic() && _impl_params->reordered_weights != nullptr ? _impl_params->reordered_weights : dep_memory_ptr(1);
+        if (is_dynamic() && _reordered_weights_cache.size() != 0) {
+            return _reordered_weights_cache.get_recent_element().second;
```
vladimir-paramuzov, line 57: Reason why recent is used here is not that  obvious to my mind. What do you think about something like that:
```cpp
// in update_weights
impl_params.weights_layout = requested_weights_layout


if (is_dynamic()) {
   auto dyn_weights = _reordered_weights_cache.get(_impl_params.weights_layout);
   return dyn_weights ? dyn_weights : dep_memory(1)
} else { ... }

```

src/plugins/intel_gpu/src/graph/include/primitive_inst.h
vladimir-paramuzov, line 265: I think we need to share weights across streams somehow to reduce memory consumption

sshlyapn: Yeah, agree, do you mind to implement this in separate PR?

src/plugins/intel_gpu/src/graph/include/primitive_inst.h
vladimir-paramuzov: For now we can have internal property so we can expose it later if needed

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -688,6 +692,39 @@ void network::add_default_output_chains() {
...
+        }
+
+        return weights_size;
+    };
+
+    size_t total_weights_size = 0;
+    for (auto node : _program->get_processing_order())
+        total_weights_size += get_weights_size(*node);
+
+    total_weights_size *= _config.get_property(ov::streams::num);
```
yeonbok, line 713: Don't we share the weights within streams? I think it could be done simply if we move the weight cache to node?

sshlyapn: @yeonbok  yes, it is a good idea, but it looks like we need to cover some other topics to implement this in the most effective way:
1) Pre-compiled reorders kernels sharing over streams (move in_memory_kernels_cache from network to program OR use existing program's implementations cache)
2) Provide safe cache's entry access (LRU's methods _get()_ (which may update existing cache entry and cause some problem for other streams by cached object replacemnt), _add()_) from multiple streams
3) Memory reuse related difficulties
And maybe something else...

yeonbok: I agree to have a separate work for the improvement. BTW about memory reuse related difficulties, why there would be such a concern? The memories for weights (constant) will not be reused.

src/plugins/intel_gpu/include/intel_gpu/runtime/lru_cache.hpp
```diff
@@ -44,12 +44,25 @@ class LruCache {
         }
     }
 
+    /**
+     * @brief Get the most recently used element with key and value pair in the cache
+     *
+     * @return std::pair<Key, Value>
+     */
+    std::pair<Key, Value> get_recent_element() const {
```
vladimir-paramuzov, line 52: Looks like a duplicate of get_lru_element

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -688,6 +692,39 @@ void network::add_default_output_chains() {
...
+
+    if (total_weights_size != 0) {
+        const size_t required_weights_cache_capacity = 3;
+        const size_t max_device_mem_size = _engine.get_device_info().max_global_mem_size;
+        const size_t max_weights_cache_capacity = max_device_mem_size / total_weights_size;
+
+        if (max_weights_cache_capacity >= required_weights_cache_capacity) {
+            _weights_cache_capacity = required_weights_cache_capacity;
+        } else if (max_weights_cache_capacity >= 1) {
+            _weights_cache_capacity = max_weights_cache_capacity;
```
vladimir-paramuzov, line 723: Can't it be simplified to
```cpp
_weights_cache_capacity = std:min(max_weights_cache_capacity, required_weights_cache_capacity);
```
?

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -688,6 +692,39 @@ void network::add_default_output_chains() {
...
+    size_t total_weights_size = 0;
+    for (auto node : _program->get_processing_order())
+        total_weights_size += get_weights_size(*node);
+
+    total_weights_size *= _config.get_property(ov::streams::num);
+
+    if (total_weights_size != 0) {
+        const size_t required_weights_cache_capacity = 3;
+        const size_t max_device_mem_size = _engine.get_device_info().max_global_mem_size;
+        const size_t max_weights_cache_capacity = max_device_mem_size / total_weights_size;
```
vladimir-paramuzov, line 718: Ideally, this logic should respect memory required for activations and memory for other constants

PR #16019, opened 2023-03-01
Fix for topological sort

olpipi:
### Details:
 - Change order of element sent to topological sort to avoid reordering of inputs or outputs

### Tickets:
 - CVS-104357

### DIFF
--- a/src/core/include/openvino/core/graph_util.hpp
+++ b/src/core/include/openvino/core/graph_util.hpp
@@ -227,9 +227,9 @@ std::vector<std::shared_ptr<Node>> topological_sort(T root_nodes) {
     std::unordered_map<Node*, uint8_t /*is_visited*/> nodes_visited;
     std::vector<std::shared_ptr<Node>> result;
 
-    for (auto node = root_nodes.rbegin(); node != root_nodes.rend(); ++node ) {
+    for (auto node = root_nodes.rbegin(); node != root_nodes.rend(); ++node) {
         nodes_to_do.push(node->get());
-    } 
+    }
 
     while (nodes_to_do.size() > 0) {
         Node* node = nodes_to_do.top();
--- a/src/core/include/openvino/core/graph_util.hpp
+++ b/src/core/include/openvino/core/graph_util.hpp
@@ -227,9 +227,10 @@ std::vector<std::shared_ptr<Node>> topological_sort(T root_nodes) {
     std::unordered_map<Node*, uint8_t /*is_visited*/> nodes_visited;
     std::vector<std::shared_ptr<Node>> result;
 
-    for (auto& node : root_nodes) {
-        nodes_to_do.push(node.get());
-    }
+    for (auto node = root_nodes.rbegin(); node != root_nodes.rend(); ++node ) {
+        nodes_to_do.push(node->get());
+    } 
+
     while (nodes_to_do.size() > 0) {
         Node* node = nodes_to_do.top();
         if (nodes_done.count(node) == 0) {
--- a/src/core/include/openvino/core/graph_util.hpp
+++ b/src/core/include/openvino/core/graph_util.hpp
@@ -8,7 +8,6 @@
 #include <functional>
 #include <list>
 #include <memory>
-#include <stack>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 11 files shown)


src/core/include/openvino/core/graph_util.hpp
```diff
@@ -226,9 +226,10 @@ std::vector<std::shared_ptr<Node>> topological_sort(T root_nodes) {
     std::unordered_set<Node*> nodes_done;
     std::vector<std::shared_ptr<Node>> result;
 
-    for (auto& node : root_nodes) {
-        nodes_to_do.push(node.get());
-    }
+    for (auto node = root_nodes.rbegin(); node != root_nodes.rend(); ++node ) {
```
ilyachur, line 229: Why do you need the reverse order here?

olpipi: nodes_to_do is a std::stack. It reverse order of inputs and outputs

src/core/tests/pass/serialization/deterministicity.cpp
```diff
@@ -8,31 +8,35 @@
...
 #include "common_test_utils/common_utils.hpp"
 #include "common_test_utils/file_utils.hpp"
+#include "openvino/opsets/opset1.hpp"
 #include "openvino/pass/serialize.hpp"
 #include "openvino/util/file_util.hpp"
 #include "read_ir.hpp"
 #include "util/test_common.hpp"
 
-class SerializationDeterministicityTest : public ov::test::TestsCommon {
+class DeterministicityCommon {
```
ilyachur, line 17: Why did you do this change? Why doesn't previous approach work?

src/core/include/openvino/core/graph_util.hpp
```diff
@@ -227,9 +227,10 @@ std::vector<std::shared_ptr<Node>> topological_sort(T root_nodes) {
     std::unordered_map<Node*, uint8_t /*is_visited*/> nodes_visited;
     std::vector<std::shared_ptr<Node>> result;
 
-    for (auto& node : root_nodes) {
-        nodes_to_do.push(node.get());
+    for (auto node = root_nodes.rbegin(); node != root_nodes.rend(); ++node) {
```
ilyachur, line 230: Do we really need to have stack for nodes_to_do?

src/core/tests/pass/serialization/deterministicity.cpp
olpipi: I want to create new parameterized tests and I need some methods from SerializationDeterministicityTest. So I moved them to DeterministicityCommon class and inherit SerializationDeterministicityTest and new SerializationDeterministicityInputOutputTest from it.

src/core/tests/pass/serialization/deterministicity.cpp
```diff
@@ -8,31 +8,35 @@
...
+    std::string xmlFileName;
 
-    void SetUp() override {
+    void SetupFileNames() {
         std::string filePrefix = CommonTestUtils::generateTestFilePrefix();
         m_out_xml_path_1 = filePrefix + "1" + ".xml";
         m_out_bin_path_1 = filePrefix + "1" + ".bin";
         m_out_xml_path_2 = filePrefix + "2" + ".xml";
         m_out_bin_path_2 = filePrefix + "2" + ".bin";
+        xmlFileName = filePrefix + "_TestModel.xml";
```
vurusovs, line 31: The field is used only in one test. It would be better to replace it from common part directly to the test

src/core/tests/pass/serialization/deterministicity.cpp
```diff
@@ -130,3 +145,150 @@ TEST_F(SerializationDeterministicityTest, ModelWithConstants) {
...
+    EXPECT_EQ(input0Name, expected1->input(0).get_node()->get_friendly_name());
+    EXPECT_EQ(input1Name, expected1->input(1).get_node()->get_friendly_name());
+    EXPECT_EQ(output0Name, expected1->output(0).get_node()->get_friendly_name());
+    EXPECT_EQ(output1Name, expected1->output(1).get_node()->get_friendly_name());
+    EXPECT_EQ(input0Name, expected2->input(0).get_node()->get_friendly_name());
+    EXPECT_EQ(input1Name, expected2->input(1).get_node()->get_friendly_name());
+    EXPECT_EQ(output0Name, expected2->output(0).get_node()->get_friendly_name());
+    EXPECT_EQ(output1Name, expected2->output(1).get_node()->get_friendly_name());
+
+    std::ifstream initial_xml(xmlFileName, std::ios::in | std::ios::binary);
```
vurusovs, line 283: The part below is look like a test on `Serialization`. Is it not covered by serialization tests?

vurusovs (changes requested):

vurusovs:
`Topological_sort` part LGTM

olpipi:
Need to check if CI passes before review

src/core/include/openvino/core/graph_util.hpp
```diff
@@ -222,14 +222,17 @@ void replace_nodes(const std::shared_ptr<Model>& f,
...
     std::unordered_map<Node*, uint8_t /*is_visited*/> nodes_visited;
     std::vector<std::shared_ptr<Node>> result;
 
+    std::vector<Node*> root_nodes_vector;
     for (auto& node : root_nodes) {
-        nodes_to_do.push(node.get());
+        root_nodes_vector.push_back(node.get());
     }
+    std::reverse(root_nodes_vector.begin(), root_nodes_vector.end());
+    std::stack<Node*, std::vector<Node*>> nodes_to_do(root_nodes_vector);
```
vurusovs, line 234: Additional vector leads to extra copies. In this case `deque` or reverse filling of `stack` were preferable

PR #16029, opened 2023-03-01
[PT FE] Fix for prim::Constant tensor with int/float values

mmikolajcz:
### Details:
 - *Fix issue with translating prim::Constant with values of type float/int*

### Tickets:
 - *104921*


mvafin:
Could you add tests for this? There are tests for decoder in `tests/layer_tests/py_frontend_tests/test_torch_decoder.py`

mmikolajcz:
> Could you add tests for this? There are tests for decoder in `tests/layer_tests/py_frontend_tests/test_torch_decoder.py`

Added test cases that reproduce behavior that previously caused issues.

PR #16035, opened 2023-03-02
[PT FE] Add aten::Chunk implementation

PiotrKrzem:
### Details:
 - Add chunk implementation for PyTorch frontend

### Tickets:
 - 101084


src/frontends/pytorch/src/transforms/prim_list_unpack_replacer.cpp
```diff
@@ -73,14 +73,51 @@ PrimListUnpackReplacer::PrimListUnpackReplacer() {
...
         if (auto chunk = cast_fw_node(input_node, "aten::chunk")) {
-            // Using number of ListUnpack outputs instead of 1st input to chunk.
-            // TODO: confirm it works for all cases
-            auto split = std::make_shared<opset10::Split>(chunk->get_input_source_output(0),
-                                                          chunk->get_input_source_output(2),
-                                                          list_unpack->get_output_size());
-
-            copy_runtime_info({list_unpack, input_node}, split);
-            replace_node(list_unpack, split);
+            auto input_tensor = chunk->get_input_source_output(0);
```
tomdol, line 76: As a rule of thumb all those local variables should be `const auto` instead.

mvafin (approved):

PR #16037, opened 2023-03-02
Take VPUX out of AUTO default candidate device list

yangwang201911:
### Details:
 - Add device blacklist for AUTO plugin.
 - Update the logic to parse out the device candidate list from the inputting config MULTI_DEVICE_PRIORITIES.

### Tickets:
 - 100585


src/plugins/auto/utils/config.hpp
```diff
@@ -304,5 +298,6 @@ struct PluginConfig {
     std::map<std::string, std::string> _keyConfigMap;
     const std::set<std::string> _availableDevices =
         {"AUTO", "CPU", "GPU", "TEMPLATE", "MYRIAD", "VPUX", "MULTI", "HETERO", "mock"};
+    const std::set<std::string> _deviceBlacklist = {"GNA", "VPUX"};
```
riverlijunjie, line 301: I don't think blacklist is must.  Any clarification about black list?

peterchen-intel: _availableDevices will be removed CVS-90102. Due to the readiness and device support strategy, GNA/VPUX are not in AUTO automatic selection list now, user needs to have them in ov::device::priorities list explicitly if they want to use any of  them.

src/plugins/auto/plugin_config.cpp
```diff
@@ -5,6 +5,7 @@
 
 namespace MultiDevicePlugin {
 const std::set<std::string> PluginConfig::_availableDevices = {"AUTO", "CPU", "GPU", "TEMPLATE", "NVIDIA", "VPUX", "MULTI", "HETERO", "mock"};
+const std::set<std::string> PluginConfig::_deviceBlacklist = {"VPUX", "GNA"};
```
wangleis, line 8: Please add description about blacklist in code

wangleis (approved):

wangleis (approved):

src/plugins/auto/plugin_config.cpp
yangwang201911: Updated.

src/plugins/auto/plugin_config.cpp
```diff
@@ -5,6 +5,10 @@
 
 namespace MultiDevicePlugin {
 const std::set<std::string> PluginConfig::_availableDevices = {"AUTO", "CPU", "GPU", "TEMPLATE", "NVIDIA", "VPUX", "MULTI", "HETERO", "mock"};
+// AUTO/MULTI will disable the device appeared in blacklist if
```
songbell, line 8:
```suggestion
// AUTO will enable the back list if
```

src/plugins/auto/plugin_config.cpp
```diff
@@ -5,6 +5,10 @@
...
+// 1.No device condidate list passing to AUTO/MULTI.(eg. core.compile_model(model, "AUTO", configs);)
```
songbell, line 9:
```suggestion
// 1.No device priority been passed to AUTO/MULTI.(eg. core.compile_model(model, "AUTO", configs);)
```

src/plugins/auto/plugin_config.cpp
```diff
@@ -5,6 +5,10 @@
...
+// 2.No valid device parsed out from device condidate list (eg. core.compile_model(model, "AUTO:-CPU,-GPU", configs);).
```
songbell, line 10:
```suggestion
// 2.No valid device parsed out from device priority (eg. core.compile_model(model, "AUTO:-CPU,-GPU", configs);).
```

src/tests/unit/auto/get_device_list.cpp
```diff
@@ -95,20 +94,21 @@ TEST_P(GetDeviceListTest, GetDeviceListTestWithExcludeList) {
...
-    ConfigParams {"-GPU.0,GPU,CPU",
-        "GPU.1,CPU"},
-    ConfigParams {"-GPU.0,GPU",
-        "GPU.1"},
-    ConfigParams {"-GPU.0", "CPU,GPU.1,VPUX"},
-    ConfigParams {"-GPU.0,-GPU.1", "CPU,VPUX"},
-    ConfigParams {"-GPU.0,-CPU", "GPU.1,VPUX"}
-};
-
+    ConfigParams{" ", " "},
```
songbell, line 97: where will the exception happen for this case?

src/tests/unit/auto/get_device_list.cpp
```diff
@@ -95,20 +94,21 @@ TEST_P(GetDeviceListTest, GetDeviceListTestWithExcludeList) {
...
+    ConfigParams{"", "CPU,GPU.0,GPU.1"},
+    ConfigParams{"CPU, ", "CPU, "},
+    ConfigParams{" ,CPU", " ,CPU"},
```
songbell, line 100: same question as above

src/tests/unit/auto/get_device_list.cpp
yangwang201911: No exception happened here as this test case is just to parse out the valid device from the device priorities rather than is to check if loadNetwork() works. Following the current logic of Auto::GetDevicelist(), device with name " " will be parsed out as vaild device because it was specified by user.

src/tests/unit/auto/get_device_list.cpp
yangwang201911: refer to above commont.

src/plugins/auto/plugin.cpp
```diff
@@ -854,84 +854,73 @@ void MultiDeviceInferencePlugin::RegisterPriority(const unsigned int& priority,
...
+        devicesToBeDeleted.resize(std::distance(devicesToBeDeleted.begin(), iterDel));
+        const auto& iterMerge =
+            std::remove_if(devicesToBeMerged.begin(), devicesToBeMerged.end(), [](const std::string& item) {
+                return item.front() == '-';
+            });
+        devicesToBeMerged.resize(std::distance(devicesToBeMerged.begin(), iterMerge));
+        for (auto&& device : devicesToBeDeleted)
+            LOG_INFO_TAG("remove %s from device candidate list", device.c_str());
+        auto isAnyDev = [](std::string& device, const std::vector<std::string>& devices) {
+            auto iter = std::find_if(devices.begin(), devices.end(), [device](const std::string& devItem) {
```
songbell, line 885: something like this?
return std::find(devices.begin(), devices.end(),device) != devices.end())

src/plugins/auto/plugin.cpp
```diff
@@ -854,84 +854,73 @@ void MultiDeviceInferencePlugin::RegisterPriority(const unsigned int& priority,
...
-    auto deviceListConfig = config.find(ov::device::priorities.name());
-    if (deviceListConfig->second.empty()) {
-        for (auto&& device : deviceList) {
-            // filter out the supported devices
-            if (!_pluginConfig.isSupportedDevice(device))
-                continue;
-            allDevices += device + ",";
-        }
-    } else {
+    auto deviceListConfig = config.find(MultiDeviceConfigParams::KEY_MULTI_DEVICE_PRIORITIES);
```
songbell, line 857:
```suggestion
    auto deviceListConfig = config.find(ov::device::priorities.name());
```

songbell (approved):

src/plugins/auto/plugin.cpp
```diff
@@ -854,84 +854,73 @@ void MultiDeviceInferencePlugin::RegisterPriority(const unsigned int& priority,
...
-                for (auto&& iter : deviceVec) {
-                    for (auto&& viter : deviceList) {
-                        if (viter.find(iter) != std::string::npos && deviceNeedToMerge(iter))
-                            mergedList.push_back(std::move(viter));
+            for (auto&& device : devicesToBeMerged) {
+                if (!isAnyDev(device, deviceList)) {
+                    // Add user specified device into candidate list
+                    devicesMerged.push_back(device);
+                } else {
+                    // Update device name if supported device with id existd
```
riverlijunjie, line 902:
```suggestion
                    // Update device name if supported device with id existed
```

src/tests/unit/auto/get_device_list.cpp
riverlijunjie, line 97: Do we need add test for:
ConfigParams{"", " "},

yangwang201911: I think this is necessary for the test case of inputting device priorities like "AUTO: ".

riverlijunjie (approved):

src/tests/unit/auto/get_device_list.cpp
```diff
@@ -79,40 +88,63 @@ class GetDeviceListTest : public ::testing::TestWithParam<ConfigParams> {
...
+                                         Params{"GPU", "GPU"},
+                                         Params{"GPU.0", "GPU.0"},
+                                         Params{"GPU,GPU.0", "GPU,GPU.0"},
+                                         Params{"CPU", "CPU"},
+                                         Params{" ,CPU", " ,CPU"},
+                                         Params{" ,GPU", " ,GPU"},
+                                         Params{"GPU, ", "GPU, "},
+                                         Params{"CPU,GPU", "CPU,GPU"},
+                                         Params{"CPU,-GPU", "CPU"},
+                                         Params{"CPU,-GPU,GPU.0", "CPU,GPU.0"},
```
wangleis, line 131: `-GPU` mean remove all GPU devices. Why we still have GPU.0 in result?

src/tests/unit/auto/get_device_list.cpp
```diff
@@ -79,40 +88,63 @@ class GetDeviceListTest : public ::testing::TestWithParam<ConfigParams> {
...
+                                         Params{"CPU,GPU,-GPU.0", "CPU"},
```
wangleis, line 132: Do we have `GPU.1` in this test platform? If yes, should we have `GPU.1` in result?

src/tests/unit/auto/get_device_list.cpp
```diff
@@ -79,40 +88,63 @@ class GetDeviceListTest : public ::testing::TestWithParam<ConfigParams> {
...
+                                         Params{"CPU,GPU,-GPU.1", "CPU"},
```
wangleis, line 133: Why do we not have `GPU.0` in result?

src/tests/unit/auto/get_device_list.cpp
yangwang201911: For this case, the available device list is just {"CPU", "GPU"}.

src/tests/unit/auto/get_device_list.cpp
yangwang201911: same reason.

src/tests/unit/auto/get_device_list.cpp
yangwang201911: I think It may be reasonable to keep "GPU.0" because user has specified the "GPU.0" in the device priority. "-GPU"  means to remove the rest sub-device of GPU like "GPU.1".

src/tests/unit/auto/get_device_list.cpp
wangleis: If there is only one GPU card in system, from my understanding, the result should be "CPU, GPU.0" or "CPU, GPU".

src/plugins/auto/plugin_config.cpp
```diff
@@ -5,6 +5,10 @@
 
 namespace MultiDevicePlugin {
 const std::set<std::string> PluginConfig::_availableDevices = {"AUTO", "CPU", "GPU", "TEMPLATE", "NVIDIA", "VPUX", "MULTI", "HETERO", "mock"};
+// AUTO will enable the back list if
```
peterchen-intel, line 8: back list -> blocklist

yangwang201911: Updated.

src/tests/unit/auto/get_device_list.cpp
```diff
@@ -79,40 +88,72 @@ class GetDeviceListTest : public ::testing::TestWithParam<ConfigParams> {
...
+                                         Params{"CPU,GPU", "CPU,GPU"},
+                                         Params{"CPU,-GPU", "CPU"},
+                                         Params{"CPU,-GPU,GPU.0", "CPU,GPU.0"},
+                                         Params{"CPU,-GPU,GPU.1", "CPU,GPU.1"},
+                                         Params{"CPU,GPU,-GPU.0", "CPU"},
+                                         Params{"CPU,GPU,-GPU.1", "CPU,GPU"},
+                                         Params{"CPU,GPU.0,GPU", "CPU,GPU"},
+                                         Params{"CPU,GPU,GPU.0", "CPU,GPU"},
+                                         Params{"CPU,GPU,GPU.1", "CPU,GPU,GPU.1"},
+                                         Params{"CPU,GPU.1,GPU", "CPU,GPU.1,GPU"},
```
wangleis, line 141: When there is `GPU.1` in device list, I assume `GPU` is class name. 

I guess the result should be `CPU,GPU.1,GPU.0`

yangwang201911: The available device list of the platform in this case is {"CPU", "GPU"} rather than {"CPU","GPU.0","GPU.1"}. "GPU.0" should not be existed in the candidate device list.

PR #16045, opened 2023-03-02
stridedslice skip execution with 0 dims

zhangYiIntel:
### Details:
 - *Skip execution when i/o has 0 dims*

### Tickets:
 - *CVS-105044*


src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -310,15 +310,19 @@ void StridedSlice::prepareParams() {
             dstMemory.push_back(getChildEdgeAt(i)->getMemoryPtr());
         }
     }
-
-    execPtr = std::make_shared<StridedSliceCommonExecutor>(attrs, srcMemory, dstMemory, errorPrefix);
+    if (srcMemory[DATA_ID]->GetShape().hasZeroDims() || dstMemory[0]->GetShape().hasZeroDims())
```
dmitry-gorokhov, line 313: We do have isExecutable() method in Node API which was designed for such use-cases. Can we deal with the issue by its implementaion adjustment?

zhangYiIntel: Good idea, I will try it, thx!

src/plugins/intel_cpu/src/nodes/strided_slice.h
```diff
@@ -127,6 +127,7 @@ class StridedSlice : public Node {
     bool isConstantInput[AXES_ID + 1] = {false};
     bool shapeHasDataDependency = false;
     bool hasConstAttrInputs = true;
+    bool skipExecution = false;
```
dmitry-gorokhov, line 130: The variable is not required anymore

dmitry-gorokhov (approved):

PR #16065, opened 2023-03-02
[VPUX] - benchmark app issue fix - PERFORMANCE_HINT: UNDEFINED

DariaMityagina:
Ticket: E-71637

With the latest OV master benchmark_app output looks like this:

image

It should be like this:
image

The `performance_hint` parameter is ignored.

These changes seem to cause this behavior:
- https://github.com/openvinotoolkit/openvino/commit/a088d1ab7dd92d471f392003df05d3daa8c903cd#diff-e7a536b4318b2270a4ee67f723b8ff57e9a5aed0928c11fa46e5a0b02db6300fL609-L613
- https://github.com/openvinotoolkit/openvino/commit/a088d1ab7dd92d471f392003df05d3daa8c903cd#diff-f311f9345d31892350d1e310ba8f0119806af62f514f7936f632df7fa6a32995L325

I tested the changes from this PR in our Net-Validation CI job. But feel free to point out any other OV check to confirm the fix.

### DIFF
--- a/samples/cpp/benchmark_app/main.cpp
+++ b/samples/cpp/benchmark_app/main.cpp
@@ -851,7 +851,7 @@ int main(int argc, char* argv[]) {
             if (!modelStream.is_open()) {
                 throw std::runtime_error("Cannot open model file " + FLAGS_m);
             }
-            compiledModel = core.import_model(modelStream, device_name, {});
+            compiledModel = core.import_model(modelStream, device_name, device_config);
             modelStream.close();
 
             auto duration_ms = get_duration_ms_till_now(startTime);
--- a/tools/benchmark_tool/openvino/tools/benchmark/main.py
+++ b/tools/benchmark_tool/openvino/tools/benchmark/main.py
@@ -442,7 +442,7 @@ def main():
             next_step()
 
             start_time = datetime.utcnow()
-            compiled_model = benchmark.core.import_model(args.path_to_model)
+            compiled_model = benchmark.core.import_model(args.path_to_model, benchmark.device, device_config)
             duration_ms = f"{(datetime.utcnow() - start_time).total_seconds() * 1000:.2f}"
             logger.info(f"Import model took {duration_ms} ms")
             if statistics:


samples/cpp/benchmark_app/main.cpp
```diff
@@ -606,6 +606,10 @@ int main(int argc, char* argv[]) {
                 device_nstreams.erase(device);
             }
         }
+        for (auto&& item : config) {
+            core.set_property(item.first, item.second);
```
nkogteva, line 610: It breaks virtual plugins. `ov::device::properties` are disallowed in `set_property`

nkogteva (changes requested):

samples/cpp/benchmark_app/main.cpp
DariaMityagina: Thanks for the review!
Sorry, didn't get the idea. Could you please describe it more broadly?
Breaks virtual plugins how?

nkogteva: > Thanks for the review! Sorry, didn't get the idea. Could you please describe it more broadly? Breaks virtual plugins how?

Such plugins as AUTO/MUTLI/HETERO allows to set device properties via `ov::device::properties`. You mentioned commit from PR https://github.com/openvinotoolkit/openvino/pull/15175 which disallows setting such properties via core. They must be set in config of compile_model. Are you sure that problem is not in VPUX?

DariaMityagina: > Are you sure that problem is not in VPUX?

Pretty much yes. Because the only change I made when I started CI was to update the OV commit to a recent commit on the master branch. And I got the results that are shown in the PR description.

Okay, I will see what I can do here.
Thanks!

DariaMityagina: Discussed this PR offline with @yangwang201911. 
Thanks @yangwang201911!
I will modify this PR shortly.

DariaMityagina: Okay, it turned out to be very simple :)
The problem only occurred when running a benchmark_app that used .blob files.
```
./benchmark_app -m model.blob ...
```
This is due to the fact that when switching to `device_config`, the `import_model` call remained unchanged, without the `device_config` parameter.

DariaMityagina: The CI check is now in progress. But the fix worked locally.

DariaMityagina:
@nkogteva, @ilya-lavrenov, @ilyachur hello!
Could you please take a look at this PR?

DariaMityagina:
> Is it applicable for python BA?

Let me check.

DariaMityagina:
> Is it applicable for python BA?

@nkogteva done, thanks!

DariaMityagina:
FYI: pip-conflicts_windows_vs2019_release failure - know issue. It is not caused by my changes.
```
[2023-03-09T16:15:50.287Z] [2023-03-09 17:15:49,977] [13000] helper.cmd_exec INFO: ERROR: Cannot install scikit-learn>=0.24.1 and scikit-learn>=1.2.0 because these package versions have conflicting dependencies.
```

nkogteva (approved):

yangwang201911 (approved):

ilyachur (approved):

PR #16068, opened 2023-03-03
Fix C1083 issue in CI

riverlijunjie:
### Details:
 - CI build failure: output_layers_handling_in_transformations_for_concat_multi_channel.cpp fatal error C1083: Cannot open compiler generated file: '': Invalid argument

### Tickets:
 - 97736


riverlijunjie:
Close it due to it doesn't work to solve Window path maximum length issue:
https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

PR #16069, opened 2023-03-03
optimize shape infer of stridedslice

xczhai:
### Details:
 - *optimize the shape infer of stridedslice*
 - *not include ellipsis_axis for now*

### Tickets:
 - *100507*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/strided_slice.cpp
+++ b/src/plugins/intel_cpu/src/nodes/strided_slice.cpp
@@ -115,7 +115,7 @@ public:
         if (const auto Slice_op = ov::as_type_ptr<const ov::op::v8::Slice>(m_op)) {
             return std::make_shared<NgraphShapeInfer>(make_shape_inference(m_op), m_port_mask);
         } else if (const auto StridedSlice_op = ov::as_type_ptr<const ov::op::v1::StridedSlice>(m_op)) {
-            const auto ellipsis_mask = StridedSlice_op->get_ellipsis_mask();
+            const auto& ellipsis_mask = StridedSlice_op->get_ellipsis_mask();
             if (std::count(ellipsis_mask.begin(), ellipsis_mask.end(), 1)) {
                 return std::make_shared<NgraphShapeInfer>(make_shape_inference(m_op), m_port_mask);
             } else {
--- a/src/plugins/intel_cpu/src/nodes/strided_slice.cpp
+++ b/src/plugins/intel_cpu/src/nodes/strided_slice.cpp
@@ -39,57 +39,75 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
     return true;
 }
 
-inline int calculate_output_shape(const int& startIdx, const int& stopIdx, const int& step, const int& len) {
-    auto start = (startIdx > len - 1) ? len - 1 : ((startIdx % len + len) % len);
-    auto stop = (stopIdx > len - 1)   ? len     : ((stopIdx % len + len) % len);
-    return (stop - start + step - 1) / step;
+// keep it in mind:
+// The calculate_output_shape function is designed based on the premise that the len parameter is a positive integer greater than 0.
+// If the len parameter is equal to 0, the output result will be 0 directly, and there is no need to call the function separately.
+inline int calculate_output_shape(int64_t startIdx, int64_t stopIdx, int64_t step, int64_t len) {
+    if (startIdx < 0) {
+        startIdx = startIdx < -len ? 0 : startIdx + len;
+    } else {
+        startIdx = startIdx > (len - 1) ? len : startIdx;
+    }
+
+    if (stopIdx < 0) {
+        stopIdx = stopIdx < -len ? 0 : stopIdx + len;
+    } else {
+        stopIdx = stopIdx > (len - 1) ? len : stopIdx;
+    }
+
+    if (step < 0) {
+        step = -step;
+    }
+    return (stopIdx - startIdx + step - 1) / step;
 }
 
 class StridedSliceShapeInfer : public ShapeInferEmptyPads {
 public:
-    StridedSliceShapeInfer(const std::shared_ptr<ov::Node>& op,
-                           const IShapeInfer::port_mask_t& port_mask,
-                           const std::set<int64_t>& begin_mask,
-                           const std::set<int64_t>& end_mask,
-                           const std::set<int64_t>& new_axis_mask,
-                           const std::set<int64_t>& shrink_axis_mask)
-    : m_op(op), m_port_mask(port_mask), m_outputShape(op->get_output_partial_shape(0).rank().get_length(), 1),
-      m_begin_mask_set(begin_mask), m_end_mask_set(end_mask), m_new_axis_mask_set(new_axis_mask), m_shrink_axis_mask_set(shrink_axis_mask) {}
+    StridedSliceShapeInfer(const IShapeInfer::port_mask_t& port_mask,
+                           size_t output_size,
+                           std::unordered_set<int64_t> begin_mask,
+                           std::unordered_set<int64_t> end_mask,
+                           std::unordered_set<int64_t> new_axis_mask,
+                           std::unordered_set<int64_t> shrink_axis_mask)
+    : m_port_mask(port_mask), m_outputShape(output_size, 1),
+      m_begin_mask_set(std::move(begin_mask)),
+      m_end_mask_set(std::move(end_mask)),
+      m_new_axis_mask_set(std::move(new_axis_mask)),
+      m_shrink_axis_mask_set(std::move(shrink_axis_mask)) {}
 
-    std::vector<VectorDims> infer(
+    Result infer(
         const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
         const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
         const VectorDims& shapeIn = input_shapes[0].get();
         const VectorDims& shapeBegin = input_shapes[1].get();
-        auto beginPtr = reinterpret_cast<int *>(data_dependency.at(1)->GetPtr());
-        auto endPtr = reinterpret_cast<int *>(data_dependency.at(2)->GetPtr());
-        auto stridePtr = reinterpret_cast<int *>(data_dependency.at(3)->GetPtr());
+        if (data_dependency.at(1)->GetDataType() != dnnl::memory::data_type::s32) {
+            IE_THROW(Unexpected) << "The data type of begin/end/stride is NOT s32, which is unexpected!";
+        }
+        auto beginPtr = reinterpret_cast<int32_t *>(data_dependency.at(1)->GetPtr());
+        auto endPtr = reinterpret_cast<int32_t *>(data_dependency.at(2)->GetPtr());
+        auto stridePtr = reinterpret_cast<int32_t *>(data_dependency.at(3)->GetPtr());
 
         for (auto i = 0, new_idx = 0; i < shapeIn.size(); ++i) {
-            bool is_new_axis = m_new_axis_mask_set.count(i);
-            bool is_shrink_axis = m_shrink_axis_mask_set.count(i);
-            bool is_begin_axis = m_begin_mask_set.count(i);
-            bool is_end_axis = m_end_mask_set.count(i);
-
-            if (is_new_axis) {
+            if (m_new_axis_mask_set.count(i)) {
+                // deal with new_axis_mask
                 m_outputShape[new_idx] = 1;
                 m_outputShape[new_idx+1] = shapeIn[i];
                 new_idx+=2;
-                continue;
-            }
-            if (is_shrink_axis) {
-                continue;
-            }
-            if (i<shapeBegin[0]) {
-                auto begin = is_begin_axis ? 0 : beginPtr[i];
-                auto end  = is_end_axis ? shapeIn[i] : endPtr[i];
-                m_outputShape[new_idx] = calculate_output_shape(begin, end, stridePtr[i], shapeIn[i]);
+            } else if (m_shrink_axis_mask_set.count(i)) {
+                // deal with shrink_axis_mask
             } else {
-                m_outputShape[new_idx] = shapeIn[i];
+                // deal with begin_mask and end_mask
+                if ((i >= shapeBegin[0]) || (shapeIn[i] == 0)) {
+                    m_outputShape[new_idx] = shapeIn[i];
+                } else {
+                    auto begin = m_begin_mask_set.count(i) ? 0 : beginPtr[i];
+                    auto end  = m_end_mask_set.count(i) ? shapeIn[i] : endPtr[i];
+                    m_outputShape[new_idx] = calculate_output_shape(begin, end, stridePtr[i], shapeIn[i]);
+                }
+                new_idx += 1;
             }
-            new_idx += 1;
         }
-        return {m_outputShape};
+        return {{m_outputShape}, ShapeInferStatus::success};
     }
 
     port_mask_t get_port_mask() const override {
@@ -97,19 +115,18 @@ public:
     }
 
 private:
-    const std::shared_ptr<ov::Node> m_op;
     const IShapeInfer::port_mask_t m_port_mask;
     VectorDims m_outputShape;
     // set
-    std::set<int64_t> m_begin_mask_set;
-    std::set<int64_t> m_end_mask_set;
-    std::set<int64_t> m_new_axis_mask_set;
-    std::set<int64_t> m_shrink_axis_mask_set;
+    const std::unordered_set<int64_t> m_begin_mask_set;
+    const std::unordered_set<int64_t> m_end_mask_set;
+    const std::unordered_set<int64_t> m_new_axis_mask_set;
+    const std::unordered_set<int64_t> m_shrink_axis_mask_set;
 };
 
 class StridedSliceShapeInferFactory : public ShapeInferFactory {
 public:
-    StridedSliceShapeInferFactory(const std::shared_ptr<ov::Node>& op, const IShapeInfer::port_mask_t& port_mask)
+    StridedSliceShapeInferFactory(const std::shared_ptr<ov::Node>& op, IShapeInfer::port_mask_t port_mask)
     : m_op(op), m_port_mask(port_mask) {}
     ShapeInferPtr makeShapeInfer() const override {
         if (const auto Slice_op = ov::as_type_ptr<const ov::op::v8::Slice>(m_op)) {
@@ -120,8 +137,8 @@ public:
                 return std::make_shared<NgraphShapeInfer>(make_shape_inference(m_op), m_port_mask);
             } else {
                 auto vec_to_set = [](const std::vector<int64_t>& vec){
-                    std::set<int64_t> to_set;
-                    for (auto i=0; i<vec.size(); ++i) {
+                    std::unordered_set<int64_t> to_set;
+                    for (auto i = 0; i < vec.size(); ++i) {
                         if (vec[i] == 1) {
                             to_set.emplace(i);
                         }
@@ -129,7 +146,8 @@ public:
                     return to_set;
                 };
                 return std::make_shared<StridedSliceShapeInfer>(
-                    m_op, m_port_mask,
+                    m_port_mask,
+                    m_op->get_output_partial_shape(0).rank().get_length(),
                     vec_to_set(StridedSlice_op->get_begin_mask()),
                     vec_to_set(StridedSlice_op->get_end_mask()),
                     vec_to_set(StridedSlice_op->get_new_axis_mask()),
@@ -139,6 +157,7 @@ public:
             IE_THROW(NotImplemented) << "not Slice or StridedSlice";
         }
     }
+
 private:
     const std::shared_ptr<ov::Node> m_op;
     const IShapeInfer::port_mask_t m_port_mask;
--- a/src/plugins/intel_cpu/src/nodes/strided_slice.cpp
+++ b/src/plugins/intel_cpu/src/nodes/strided_slice.cpp
@@ -65,7 +65,7 @@ public:
         auto endPtr = reinterpret_cast<int *>(data_dependency.at(2)->GetPtr());
         auto stridePtr = reinterpret_cast<int *>(data_dependency.at(3)->GetPtr());
 
-        for (auto i=0, new_idx=0; i<shapeIn.size(); ++i) {
+        for (auto i = 0, new_idx = 0; i < shapeIn.size(); ++i) {
             bool is_new_axis = m_new_axis_mask_set.count(i);
             bool is_shrink_axis = m_shrink_axis_mask_set.count(i);
             bool is_begin_axis = m_begin_mask_set.count(i);
--- a/src/plugins/intel_cpu/src/nodes/strided_slice.cpp
+++ b/src/plugins/intel_cpu/src/nodes/strided_slice.cpp
@@ -40,15 +40,18 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
     return true;
 }
 
+namespace {
+
+constexpr IShapeInfer::port_mask_t port_mask = PortMask(/*BEGIN_ID*/1, /*END_ID*/2, /*STRIDE_ID*/3, /*AXES_ID*/4);
+
 class StridedSliceShapeInfer : public ShapeInferEmptyPads {
 public:
-    StridedSliceShapeInfer(const IShapeInfer::port_mask_t& port_mask,
-                           size_t output_size,
+    StridedSliceShapeInfer(size_t output_size,
                            std::unordered_set<int64_t> begin_mask,
                            std::unordered_set<int64_t> end_mask,
                            std::unordered_set<int64_t> new_axis_mask,
                            std::unordered_set<int64_t> shrink_axis_mask)
-    : m_port_mask(port_mask), m_outputShape(output_size, 1),
+    : m_outputShape(output_size, 1),
       m_begin_mask_set(std::move(begin_mask)),
       m_end_mask_set(std::move(end_mask)),
       m_new_axis_mask_set(std::move(new_axis_mask)),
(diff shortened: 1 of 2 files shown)


src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+class StridedSliceShapeInfer : public ShapeInferEmptyPads {
+public:
+    StridedSliceShapeInfer(const std::shared_ptr<ov::Node>& op,
+                           const IShapeInfer::port_mask_t& port_mask,
+                           const std::set<int64_t>& begin_mask,
+                           const std::set<int64_t>& end_mask,
+                           const std::set<int64_t>& new_axis_mask,
+                           const std::set<int64_t>& shrink_axis_mask)
+    : m_op(op), m_port_mask(port_mask), m_outputShape(op->get_output_partial_shape(0).rank().get_length(), 1),
+      m_begin_mask_set(begin_mask), m_end_mask_set(end_mask), m_new_axis_mask_set(new_axis_mask), m_shrink_axis_mask_set(shrink_axis_mask) {}
```
maxnick, line 57:
```suggestion
    StridedSliceShapeInfer(const std::shared_ptr<ov::Node>& op,
                           IShapeInfer::port_mask_t port_mask,
                           std::unordered_set<int64_t> begin_mask,
                           std::unordered_set<int64_t> end_mask,
                           std::unordered_set<int64_t> new_axis_mask,
                           std::unordered_set<int64_t> shrink_axis_mask)
    : m_op(op), m_port_mask(port_mask), m_outputShape(op->get_output_partial_shape(0).rank().get_length(), 1),
      m_begin_mask_set(std::move(begin_mask)),
      m_end_mask_set(std::move(end_mask)),
      m_new_axis_mask_set(std::move(new_axis_mask)),
      m_shrink_axis_mask_set(std::move(shrink_axis_mask)) {}
```
It is better to use hash table than a tree.
unordered_set has move constructor, then it is better to pass by value to avoid extra copy.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+
+class StridedSliceShapeInferFactory : public ShapeInferFactory {
+public:
+    StridedSliceShapeInferFactory(const std::shared_ptr<ov::Node>& op, const IShapeInfer::port_mask_t& port_mask)
+    : m_op(op), m_port_mask(port_mask) {}
+    ShapeInferPtr makeShapeInfer() const override {
+        if (const auto Slice_op = ov::as_type_ptr<const ov::op::v8::Slice>(m_op)) {
+            return std::make_shared<NgraphShapeInfer>(make_shape_inference(m_op), m_port_mask);
+        } else if (const auto StridedSlice_op = ov::as_type_ptr<const ov::op::v1::StridedSlice>(m_op)) {
+            const auto ellipsis_mask = StridedSlice_op->get_ellipsis_mask();
```
maxnick, line 118:
```suggestion
            const auto& ellipsis_mask = StridedSlice_op->get_ellipsis_mask();
```
Since used as read only

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+    // set
+    std::set<int64_t> m_begin_mask_set;
+    std::set<int64_t> m_end_mask_set;
+    std::set<int64_t> m_new_axis_mask_set;
+    std::set<int64_t> m_shrink_axis_mask_set;
+};
+
+class StridedSliceShapeInferFactory : public ShapeInferFactory {
+public:
+    StridedSliceShapeInferFactory(const std::shared_ptr<ov::Node>& op, const IShapeInfer::port_mask_t& port_mask)
```
maxnick, line 112:
```suggestion
    StridedSliceShapeInferFactory(const std::shared_ptr<ov::Node>& op, IShapeInfer::port_mask_t port_mask)
```
Maybe I'm missing something, but why it is better to pass 32bit value by reference?

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+                m_outputShape[new_idx+1] = shapeIn[i];
+                new_idx+=2;
+                continue;
+            }
+            if (is_shrink_axis) {
+                continue;
+            }
+            if (i<shapeBegin[0]) {
+                auto begin = is_begin_axis? 0 : beginPtr[i];
+                auto end  = is_end_axis? shapeIn[i] : endPtr[i];
```
maxnick, line 85:
```suggestion
                auto begin = is_begin_axis ? 0 : beginPtr[i];
                auto end  = is_end_axis ? shapeIn[i] : endPtr[i];
```

maxnick (changes requested):

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+                m_outputShape[new_idx] = 1;
+                m_outputShape[new_idx+1] = shapeIn[i];
+                new_idx+=2;
+                continue;
+            }
+            if (is_shrink_axis) {
+                continue;
+            }
+            if (i<shapeBegin[0]) {
+                auto begin = is_begin_axis? 0 : beginPtr[i];
```
praasz, line 84: I think is not correct when strides is negative (reverse). the begin should be dimension size. Also for end the default value it should be to first element (include it).
Check the template implementation [here](https://github.com/praasz/openvino/blob/e6a65f406d0362d5077db8a389ac2d555f167e80/src/core/shape_inference/include/strided_slice_shape_inference.hpp#L171). Is probably to complex for what you need but logic is already tested. It can looks complex but this implementation takes into account that dimension, start, stop can have interval values.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
     return true;
 }
 
+inline int calculate_output_shape(const int& startIdx, const int& stopIdx, const int& step, const int& len) {
+    auto start = (startIdx > len - 1) ? len - 1 : ((startIdx % len + len) % len);
+    auto stop = (stopIdx > len - 1)   ? len     : ((stopIdx % len + len) % len);
```
praasz, line 44: Is not try to to normalize and clip the start/stop to dimension like [here](https://github.com/openvinotoolkit/openvino/blob/b8348cda2e992b2e163a2c326013e59688efbb23/src/core/shape_inference/include/slice_shape_inference_utils.hpp#L164)
maybe re-use these functions, they use simple operation like compare,add no `%`.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+                           const std::set<int64_t>& shrink_axis_mask)
+    : m_op(op), m_port_mask(port_mask), m_outputShape(op->get_output_partial_shape(0).rank().get_length(), 1),
+      m_begin_mask_set(begin_mask), m_end_mask_set(end_mask), m_new_axis_mask_set(new_axis_mask), m_shrink_axis_mask_set(shrink_axis_mask) {}
+
+    std::vector<VectorDims> infer(
+        const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
+        const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
+        const VectorDims& shapeIn = input_shapes[0].get();
+        const VectorDims& shapeBegin = input_shapes[1].get();
+        auto beginPtr = reinterpret_cast<int *>(data_dependency.at(1)->GetPtr());
```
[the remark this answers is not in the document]
xczhai: > Can you give short explanation why cast to `int*` is safe here? These inputs can have any integer precision so the conversion to int is already done somewhere? As int can have different bit length so how e.g. int64_t is handled when int is less than 64-bit?

I check it that the data type is dnnl::memory::data_type::s32 while the begin/end/strides is passed in i64 data type. It is converted somewhere but I cant find this place. It is much appreciated if you know it.
To make it more clear, I replace int with int32_t.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
     return true;
 }
 
+inline int calculate_output_shape(const int& startIdx, const int& stopIdx, const int& step, const int& len) {
```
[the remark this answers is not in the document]
xczhai: > There is no advantage of using references

fix it. yes, you are right after I re-read the item 20 in Effective C++. 
- Prefer pass-by-reference-to-const over pass-by-value. It’s typically
more efficient and it avoids the slicing problem.
- The rule doesn’t apply to `built-in types` and `STL iterator` and `function
object types`. For them, pass-by-value is usually appropriate.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
xczhai: > Maybe I'm missing something, but why it is better to pass 32bit value by reference?

reply as above. accept your advice.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+        }
+        return {m_outputShape};
+    }
+
+    port_mask_t get_port_mask() const override {
+        return m_port_mask;
+    }
+
+private:
+    const std::shared_ptr<ov::Node> m_op;
```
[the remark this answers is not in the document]
xczhai: > is not used

It is used to fetch the output length. I move it out of this class and pass the output length value instead.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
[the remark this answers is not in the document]
maxnick: At the moment, due to the convert precision transformation, CPU plugin works only with s32 as a typical integer type (except of course quantization, where we use s8 and u8). But, this might be changed in near future, so it is probably better to check the type.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
xczhai: > It is better to use hash table than a tree. unordered_set has move constructor, then it is better to pass by value to avoid extra copy.

accept and refresh it.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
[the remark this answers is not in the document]
xczhai: > At the moment, due to the convert precision transformation, CPU plugin works only with s32 as a typical integer type (except of course quantization, where we use s8 and u8). But, this might be changed in near future, so it is probably better to check the type at least to throw if the type is unexpected.

Yes. I add a data type check for begin/end/stride here.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+        const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
+        const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
+        const VectorDims& shapeIn = input_shapes[0].get();
+        const VectorDims& shapeBegin = input_shapes[1].get();
+        auto beginPtr = reinterpret_cast<int *>(data_dependency.at(1)->GetPtr());
+        auto endPtr = reinterpret_cast<int *>(data_dependency.at(2)->GetPtr());
+        auto stridePtr = reinterpret_cast<int *>(data_dependency.at(3)->GetPtr());
+
+        for (auto i=0, new_idx=0; i<shapeIn.size(); ++i) {
+            bool is_new_axis = m_new_axis_mask_set.count(i);
```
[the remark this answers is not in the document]
xczhai: > Move acquire this bool flags into `if( m_new_axis_mask_set.count(i))` as getting all masks is not always required.

A very good comment. Accept your suggestion and postpone the definition.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+            bool is_shrink_axis = m_shrink_axis_mask_set.count(i);
+            bool is_begin_axis = m_begin_mask_set.count(i);
+            bool is_end_axis = m_end_mask_set.count(i);
+
+            if (is_new_axis) {
+                m_outputShape[new_idx] = 1;
+                m_outputShape[new_idx+1] = shapeIn[i];
+                new_idx+=2;
+                continue;
```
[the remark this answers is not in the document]
xczhai: > It looks like all cases in are excluding. Maybe convert this several `ifs` into one `if, else if, ... , else` to show explicitly that each case is excluding and `continue` use will be not required.

These attributes have different priority and `exclusive` property. If one attribute is set to `True`, other attribute will be ignored. The process should stop the current loop immediately and start the next loop. 
So it makes sense to use `continue` here, I think.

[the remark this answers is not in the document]
xczhai: > It looks like all cases in are excluding. Maybe convert this several `ifs` into one `if, else if, ... , else` to show explicitly that each case is excluding and `continue` use will be not required.

Sure. update it.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +39,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
     return true;
 }
 
+inline int calculate_output_shape(const int& startIdx, const int& stopIdx, const int& step, const int& len) {
+    auto start = (startIdx > len - 1) ? len - 1 : ((startIdx % len + len) % len);
+    auto stop = (stopIdx > len - 1)   ? len     : ((stopIdx % len + len) % len);
+    return (stop - start + step - 1) / step;
```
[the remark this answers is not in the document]
xczhai: > If step is negative you can get negative dimension is that ok here?
> 
> Check this [sliced dim calc](https://github.com/openvinotoolkit/openvino/blob/b8348cda2e992b2e163a2c326013e59688efbb23/src/core/shape_inference/include/slice_shape_inference_utils.hpp#L172). I think it could be possible to extract this part of code to smaller function and use in template code and here. The `lb` and `ub` is pre-processed start, stop and this part take into account the step direction and avoid using `/` in some cases (but maybe there is possibility to improve it).
> 
> If it is possible to have common code then you get for free extra tests from `type_prop`, because this code is used in `slice` and `strided slice` and tested with dynamic and static shapes.

I compare both functions in performance and coverage. 
- `calculate_output_shape` performs better in most cases but is not faster too much than `get_sliced_value`. 
- `get_sliced_value` seems safe and verified.
- code reuse is better to maintain instead of inventing one.
@praasz @maxnick Do you have other comments for this?

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +40,110 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+      m_shrink_axis_mask_set(std::move(shrink_axis_mask)) {}
+
+    Result infer(
+        const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
+        const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
+        const VectorDims& shapeIn = input_shapes[0].get();
+        const VectorDims& shapeBegin = input_shapes[1].get();
+        if (data_dependency.at(1)->GetDataType() != dnnl::memory::data_type::s32) {
+            IE_THROW(Unexpected) << "The data type of begin/end/stride is NOT s32, which is unexpected!";
+        }
```
maxnick, line 64: Two comments:
1. It is better to use `data_dependency.at(1)->getDesc().getPrecision() != Precision::I32` to avoid extra dependency on dnnl entities.
2. It looks like we need additional checks for the other 2 componentis, don't we?

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +40,110 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+        auto stridePtr = reinterpret_cast<int32_t *>(data_dependency.at(3)->GetPtr());
+
+        for (auto i = 0, new_idx = 0; i < shapeIn.size(); ++i) {
+            if (m_new_axis_mask_set.count(i)) {
+                // deal with new_axis_mask
+                m_outputShape[new_idx] = 1;
+                m_outputShape[new_idx+1] = shapeIn[i];
+                new_idx+=2;
+            } else if (m_shrink_axis_mask_set.count(i)) {
+                // deal with shrink_axis_mask
```
maxnick, line 76: If this is something we need to do later, plaese add TODO statement to the comment. But how does it work now?

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +40,110 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+    }
+
+private:
+    const std::shared_ptr<ov::Node> m_op;
+    const IShapeInfer::port_mask_t m_port_mask;
+};
+
 StridedSlice::StridedSlice(const std::shared_ptr<ov::Node>& op, const GraphContext::CPtr context) :
-        Node(op, context, NgraphShapeInferFactory(op, PortMask(1, 2, 3, 4))) {
+        Node(op, context, StridedSliceShapeInferFactory(op, PortMask(1, 2, 3, 4))) {
```
maxnick, line 146: Why we have to pass the port mask? It should be a static field of `StridedSliceShapeInfer`, I mean static constexpr. So it can be passed to the `NgraphShapeInfer` constructor and returned from the `get_port_mask` call.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
alanzhai219: > If this is something we need to do later, plaese add TODO statement to the comment. But how does it work now?

There is nothing to do if the `shrink_axis` is `true`. Just go head without doing anything.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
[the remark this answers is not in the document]
praasz: I'm not sure  if performance difference between `calculate_output_shape` vs. `get_sliced_valur` can be ignored, but `get_sliced_value` takes into account more corner cases and so I think will be slower always.

If `get_sliced_value` will be little refactored then maybe performance gap can be reduced:
- change parameters to be passed by value.
- the bool flags around lines 143:150 are not always required so could be moved inside `if is_norm_dim_max` as this case I think will be less common. These limit flags are not required, when dimension has no maximum value (or infinite), but it requires refactor if statement for `lb`,`ub` calculation which can be tricky :).

praasz (approved):

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
xczhai: > Two comments:
> 
> 1. It is better to use `data_dependency.at(1)->getDesc().getPrecision() != Precision::I32` to avoid extra dependency on dnnl entities.
> 2. It looks like we need additional checks for the other 2 componentis, don't we?

apply them.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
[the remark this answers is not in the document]
xczhai: > I'm not sure if performance difference between `calculate_output_shape` vs. `get_sliced_valur` can be ignored, but `get_sliced_value` takes into account more corner cases and so I think will be slower always.
> 
> If `get_sliced_value` will be little refactored then maybe performance gap can be reduced:
> 
> * change parameters to be passed by value.
> * the bool flags around lines 143:150 are not always required so could be moved inside `if is_norm_dim_max` as this case I think will be less common. These limit flags are not required, when dimension has no maximum value (or infinite), but it requires refactor if statement for `lb`,`ub` calculation which can be tricky :).

I profile the changes but get very little improvement. And the code becomes not tidy. So I think it is not worth to invest time here.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
maxnick: Then it is better to replace the last else with `else if (!m_shrink_axis_mask_set.count(i))` since this condition is checked anyway, and remove this line.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
maxnick: `StridedSliceShapeInferFactory` do not need to have those parameter in the parameter list at all. I propose to do the following:
1. Introduce a static field in the `StridedSliceShapeInfer` as `static constexpr port_mask = PortMask(1, 2, 3, 4);`
2. Reuse this field in the `get_port_mask()` implementation
3. Pass this value to the `NgraphShapeInfer` inside the factory.
This simple action will reduce the size of the shape inference class and the shape infer factory size by `sizeof(port_mask_t)` and remove the unnecessary complexity from the factory.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +40,113 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+class StridedSliceShapeInferFactory : public ShapeInferFactory {
+public:
+    StridedSliceShapeInferFactory(const std::shared_ptr<ov::Node>& op, IShapeInfer::port_mask_t port_mask)
+    : m_op(op), m_port_mask(port_mask) {}
+    ShapeInferPtr makeShapeInfer() const override {
+        if (const auto Slice_op = ov::as_type_ptr<const ov::op::v8::Slice>(m_op)) {
+            return std::make_shared<NgraphShapeInfer>(make_shape_inference(m_op), m_port_mask);
+        } else if (const auto StridedSlice_op = ov::as_type_ptr<const ov::op::v1::StridedSlice>(m_op)) {
+            const auto& ellipsis_mask = StridedSlice_op->get_ellipsis_mask();
+            if (std::count(ellipsis_mask.begin(), ellipsis_mask.end(), 1)) {
```
maxnick, line 118:
```suggestion
            if (std::any_of(ellipsis_mask.begin(), ellipsis_mask.end(), [](int64_t x){ return x == 1; })) {
```
&Theta; (n) -> O(n)

xczhai: > 

Good point. Apply it.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
xczhai: > `StridedSliceShapeInferFactory` do not need to have those parameter in the parameter list at all. I propose to do the following:
> 
> 1. Introduce a static field in the `StridedSliceShapeInfer` as `static constexpr port_mask = PortMask(1, 2, 3, 4);`
> 2. Reuse this field in the `get_port_mask()` implementation
> 3. Pass this value to the `NgraphShapeInfer` inside the factory.
>    This simple action will reduce the size of the shape inference class and the shape infer factory size by `sizeof(port_mask_t)` and remove the unnecessary complexity from the factory.

modified these port_mask in func and class.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
xczhai: > Then it is better to replace the last else statement with `else if (!m_shrink_axis_mask_set.count(i))` since this condition is checked anyway, and remove this line with the comment since the purpose will be clear from the code.

sure. re-arrange these codes.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -38,8 +40,109 @@ bool StridedSlice::isSupportedOperation(const std::shared_ptr<const ov::Node>& o
...
+    const std::unordered_set<int64_t> m_new_axis_mask_set;
+    const std::unordered_set<int64_t> m_shrink_axis_mask_set;
+};
+
+class StridedSliceShapeInferFactory : public ShapeInferFactory {
+public:
+    StridedSliceShapeInferFactory(const std::shared_ptr<ov::Node>& op)
+    : m_op(op) {}
+    ShapeInferPtr makeShapeInfer() const override {
+        static constexpr IShapeInfer::port_mask_t port_mask = PortMask(1, 2, 3, 4);
```
maxnick, line 111: It is more logical to keep this mask in one place:
```cpp
class StridedSliceShapeInfer {
...
public:
static constexpr IShapeInfer::port_mask_t port_mask = PortMask(1, 2, 3, 4);
...


port_mask_t get_port_mask() const override {
    return port_mask;
}
...
};
```
And access just as follows:
```cpp
return std::make_shared<NgraphShapeInfer>(make_shape_inference(m_op), StridedSliceShapeInfer::port_mask );
...
```

xczhai: @maxnick 
Thanks for your advice. Appy your suggestion to `keep it in one place`.
But I didn't define the `port_mask` in the `StridedSliceShapeInfer`. The reason is the `port_mask` is not only called by `StridedSlice` shape infer. If Slice` shape infer also access `port_mask` from `StridedSlice` shape infer it in the near future, it seems a bit strange. So I define it at the top as other nodes do.

src/plugins/intel_cpu/src/nodes/strided_slice.cpp
```diff
@@ -19,6 +21,8 @@ namespace ov {
 namespace intel_cpu {
 namespace node {
 
+static constexpr IShapeInfer::port_mask_t port_mask = PortMask(/*BEGIN_ID*/1, /*END_ID*/2, /*STRIDE_ID*/3, /*AXES_ID*/4);
```
maxnick, line 24: I did not notice this before. Could you please move this constant and all the shape inference related classes into an anonymous namespace, so that they will not be accessible outside the translation unit. In that case the `static` qualifier may be omitted.
```cpp
namespace {
constexpr IShapeInfer::port_mask_t port_mask = PortMask(/*BEGIN_ID*/1, /*END_ID*/2, /*STRIDE_ID*/3, /*AXES_ID*/4);
class StridedSliceShapeInfer : public ShapeInferEmptyPads {
...
};

class StridedSliceShapeInferFactory : public ShapeInferFactory {
...
};
} // namespace
```

PR #16072, opened 2023-03-03
[ShapeInference] EmbeddingBag-Offsets/Packed-Sum shape infer improvements

mitruska:
### Details:
 - Add shape_infer function for EmbeddingBagPackedSum (Base)
 - Register EmbeddingBagPackedSum shape_infer for CPU
 - Update shape_infer function for EmbeddingBagOffsetsSum
 - Add tests

### Tickets:
 - 97220, 97221


src/core/shape_inference/include/embeddingbag_packed_shape_inference.hpp
```diff
@@ -0,0 +1,57 @@
...
+namespace op {
+namespace util {
+
+template <class TShape>
+std::vector<TShape> shape_infer(const ov::op::util::EmbeddingBagPackedBase* op,
+                                const std::vector<TShape>& input_shapes) {
+    const auto input_size = input_shapes.size();
+    NODE_VALIDATION_CHECK(op, input_size == 2 || input_size == 3);
+
+    static constexpr int EMB_TABLE = 0;
```
praasz, line 21:
```suggestion
    constexpr size_t EMB_TABLE = 0;
```
There is not much benefit from (compiler handle optimization) and if possible make it unsigned.

src/core/tests/type_prop/embeddingbag_offsetssum.cpp
```diff
@@ -231,6 +266,24 @@ TEST(type_prop, ebos_fail_indices_1d) {
...
 }
 
+TEST(type_prop, ebos_fail_emb_table_0d) {
+    auto emb_table = make_shared<op::Parameter>(element::f32, Shape{});
+    auto indices = make_shared<op::Parameter>(element::i64, Shape{4});
+    auto offsets = make_shared<op::Parameter>(element::i64, Shape{3});
+    auto per_sample_weights = make_shared<op::Parameter>(element::f32, Shape{4});
+    auto default_index = make_shared<op::Parameter>(element::i64, Shape{});
+
+    try {
```
praasz, line 276: Detail: include common assertion utils and use `OV_EXPECT_THROW` here?

src/core/tests/type_prop/embeddingbag_packedsum.cpp
```diff
@@ -115,6 +162,21 @@ TEST(type_prop, ebps_fail_indices_1d) {
     }
 }
 
+TEST(type_prop, ebps_fail_emb_table_0d) {
+    auto emb_table = make_shared<op::Parameter>(element::f32, Shape{});
+    auto indices = make_shared<op::Parameter>(element::i64, Shape{3, 4});
+    auto per_sample_weights = make_shared<op::Parameter>(element::f32, Shape{3, 4});
+
+    try {
```
praasz, line 170: Use `OV_EXPECT_THROW` here?

src/core/shape_inference/include/embeddingbag_packed_shape_inference.hpp
mitruska: Ok, done

src/core/tests/type_prop/embeddingbag_offsetssum.cpp
mitruska: Done

src/core/tests/type_prop/embeddingbag_packedsum.cpp
mitruska: Done

src/core/shape_inference/include/embeddingbag_packed_shape_inference.hpp
```diff
@@ -0,0 +1,57 @@
...
+    if (input_size == 3) {
+        NODE_VALIDATION_CHECK(op,
+                              input_shapes[PER_SAMPLE_WEIGHTS].rank().compatible(2),
+                              "PER_SAMPLE_WEIGHTS must be 2D.");
+
+        NODE_VALIDATION_CHECK(op,
+                              TShape::merge_into(indices_shape, input_shapes[PER_SAMPLE_WEIGHTS]),
+                              "INDICES and PER_SAMPLE_WEIGHTS shape must be same.");
+    }
+    TShape output_shape;
```
praasz, line 37: This shape calculation can be same for offset and packed shape inference only require point correctly batch shape for calculations.

praasz, line 37:
```suggestion
 auto output_shape = emb_table_shape;
```
If you want use emb_table_shape as output copy it here. then for static rank set only dim0, else case is not required as output shape will dynamic.

src/core/shape_inference/include/embeddingbag_packed_shape_inference.hpp
```diff
@@ -0,0 +1,57 @@
...
+    const auto& emb_table_shape = input_shapes[EMB_TABLE];
+    if (emb_table_shape.rank().is_static()) {
+        NODE_VALIDATION_CHECK(op, emb_table_shape.size() > 0, "EMB_TABLE can't be a scalar.");
+        output_shape = emb_table_shape;
```
praasz, line 41: This can be removed.

src/core/shape_inference/include/embeddingbag_packed_shape_inference.hpp
```diff
@@ -0,0 +1,57 @@
...
+        output_shape[0] = indices_shape.rank().is_static() ? indices_shape[0] : Dimension::dynamic();
+    } else {
```
praasz, line 43: this else case can be removed if output shape set from emb shape.

src/core/shape_inference/include/embeddingbag_packed_shape_inference.hpp
mitruska: My first assumption was that there is no need to create a helper just for those few lines, 
but eventually I've added a common util for embeddings output shape.
It can be also reused for [EmbeddingSegmentsSum](https://github.com/openvinotoolkit/openvino/pull/16119).

src/core/shape_inference/include/embeddingbag_packed_shape_inference.hpp
mitruska: Updated approach is used in the common `copy_shape_and_update_dim` util.

src/core/shape_inference/include/embeddingbag_packed_shape_inference.hpp
mitruska: Updated approach is used in the common `copy_shape_and_update_dim` util.

src/core/shape_inference/include/embedding_shape_infer_utils.hpp
```diff
@@ -0,0 +1,44 @@
...
+namespace embedding {
+/**
+ * \brief Return a copy of the source shape with the first dimension replaced by provided dim
+ *
+ * \tparam TShape        Shape type
+ * \tparam TDim          Dimension type
+ *
+ * \param op             Pointer to operator.
+ * \param shape_src      The shape to be copied
+ * \param dim            The replacement dimension to overwirite the first one
```
praasz, line 20: It can be named as `batch_shape` as it may have batch dimension required for output shape.

src/core/shape_inference/include/embedding_shape_infer_utils.hpp
```diff
@@ -0,0 +1,44 @@
...
+namespace util {
+namespace embedding {
+/**
+ * \brief Return a copy of the source shape with the first dimension replaced by provided dim
+ *
+ * \tparam TShape        Shape type
+ * \tparam TDim          Dimension type
+ *
+ * \param op             Pointer to operator.
+ * \param shape_src      The shape to be copied
```
praasz, line 19: It can be renamed to `emb_table_shape`. this function is specific for embedding operator names don't need to be so generic.

src/core/shape_inference/include/embedding_shape_infer_utils.hpp
```diff
@@ -0,0 +1,44 @@
...
+template <class TShape, class TDim = typename TShape::value_type>
+TShape copy_shape_and_update_dim(const ov::Node* op, const TShape& shape_src, const TDim& dim) {
+    if (shape_src.rank().is_static()) {
+        NODE_VALIDATION_CHECK(op, shape_src.size() > 0, "EMB_TABLE can't be a scalar.");
+        auto shape_copy = shape_src;
+        shape_copy[0] = dim;
+        return shape_copy;
+    }
+    return PartialShape::dynamic();
+}
```
praasz, line 34:
```suggestion
template <class TShape>
TShape out_shape_infer(const ov::Node* op, const TShape& emb_table_shape, const TShape& batch_shape) {
    auto out_shape = emb_table_shape;
    if (emb_table_shape.rank().is_static()) {
        NODE_VALIDATION_CHECK(op, emb_table_shape.size() > 0, "EMB_TABLE can't be a scalar.");
        shape_copy[0] = batch_shape.rank().is_static() ? batch_shape[0] : Dimension::dynamic());
    }
    return out_shape;
}
```
Just one function version (maybe name it out_shape_calc)

praasz (approved):

src/core/shape_inference/include/embedding_shape_infer_utils.hpp
mitruska: The dimension is called "num_segments" for [EmbeddingSegmentsSum_3](https://github.com/openvinotoolkit/openvino/blob/master/docs/ops/sparse/EmbeddingSegmentsSum_3.md), so I decided to keep this name more generic.

src/core/shape_inference/include/embedding_shape_infer_utils.hpp
mitruska: The second version is intended to be used for [EmbeddingSegmentsSum](https://github.com/openvinotoolkit/openvino/pull/16119) as there the dimension is taken from the input value, not from the other shape.

praasz: In mentioned PR takes input as shape also after modifications.

src/core/shape_inference/include/embedding_shape_infer_utils.hpp
mitruska: Ok, done

src/core/shape_inference/include/embedding_shape_infer_utils.hpp
mitruska: As reusage of the helper for the EmbeddingSegmentsSum can lead to redundant checks,
I've updated the code to keep the only one version.

tomdol (approved):

PR #15782, opened 2023-02-17
[GPU] Add data broadcasting for OneDNN binary ops for Gemm primitive

sshlyapn:
### Details:
 - Add broadcast for proper binary_add handling

### Tickets:
 - *103387*

PR #16074, opened 2023-03-03
Broadcast for post ops enable enable onednngemm

hyunback:
This PR's purpose is for checking full models performance.
Based on https://github.com/openvinotoolkit/openvino/pull/15782 and enabling oneDNN gemm(matmul) from support multiple users and non constant input.
Fix multiple batch issue.

### Tickets:
 - *103387, 104821*

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/runtime/utils.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/runtime/utils.hpp
@@ -173,14 +173,9 @@ inline bool any_not_zero(const std::vector<T> vec) {
     return std::any_of(vec.begin(), vec.end(), [](const T& val) { return val != 0; });
 }
 
-template <typename T, typename P>
-constexpr bool one_of(const T& val, P option) {
-    return val == option;
-}
-
-template <typename T, typename P, typename... Args>
-constexpr bool one_of(const T& val, P option, Args... options) {
-    return val == option || one_of(val, options...);
+template <typename T>
+inline bool one_of(const T& val, const std::vector<T>& vec) {
+    return std::any_of(vec.begin(), vec.end(), [&val](const T& v) { return v == val; });
 }
 
 // Helpers to get string for types that have operator<< defined
--- a/src/plugins/intel_gpu/include/intel_gpu/runtime/utils.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/runtime/utils.hpp
@@ -173,6 +173,16 @@ inline bool any_not_zero(const std::vector<T> vec) {
     return std::any_of(vec.begin(), vec.end(), [](const T& val) { return val != 0; });
 }
 
+template <typename T, typename P>
+constexpr bool one_of(const T& val, P option) {
+    return val == option;
+}
+
+template <typename T, typename P, typename... Args>
+constexpr bool one_of(const T& val, P option, Args... options) {
+    return val == option || one_of(val, options...);
+}
+
 // Helpers to get string for types that have operator<< defined
 template <typename T>
 inline std::string to_string(const T& v) {
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
@@ -7,12 +7,19 @@
 #include "layout_optimizer.h"
 #include "intel_gpu/graph/program.hpp"
 #include "intel_gpu/runtime/debug_configuration.hpp"
+#include "intel_gpu/runtime/utils.hpp"
 #include "program_helpers.h"
 #include "binary_convolution_inst.h"
 #include "mvn_inst.h"
 #include "to_string_utils.h"
 #include "pooling_inst.h"
 #include "reshape_inst.h"
+
+#ifdef ENABLE_ONEDNN_FOR_GPU
+#include "gemm_inst.h"
+#include "broadcast_inst.h"
+#include <impls/onednn/utils.hpp>
+#endif
 
 #include <vector>
 #include <memory>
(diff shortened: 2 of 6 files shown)


isanghao:
No perf regression, including bert: ../log/2023-03-03--08-15.report --ref ../log/2023-03-02--00-01.report

sshlyapn (approved):

src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
```diff
@@ -958,4 +965,46 @@ void reorder_inputs::run(program& p, layout_optimizer& lo, reorder_factory& rf)
...
+                    auto data_layout = data.get_output_layout();
+                    auto data_dims = onednn::convert_gemm_tensor(data_layout.get_tensor(),
+                                                                 cldnn::format::dimension(data_layout.format),
+                                                                 false);
+
+                    if (gemm_dims[0] == data_dims[0])
+                        continue;
+
+                    auto broadcast_size = data_layout.get_shape();
+                    broadcast_size[0] = gemm_dims[0] / broadcast_size[1];
```
isanghao, line 997: @sshlyapn 
we have two cases of broadcasting: `[b,1,y,x] -> [b,f,y,x]` and `[1,f,y,x] -> [b,f,y,x]`
I think it is handling only one of these two.
What about getting `b` and `f` sizes from `gemm_layout`?

src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
```diff
@@ -958,4 +965,46 @@ void reorder_inputs::run(program& p, layout_optimizer& lo, reorder_factory& rf)
...
+    // * Gemm fused op shape: (1,f,y,x) -> OneDNN shape: (1*f,y,x)
+    // If batch dimension of gemm output is not equal to 1, then OneDNN will not be able to broadcast fused op data
+    // correctly and we need to do it manually
+#ifdef ENABLE_ONEDNN_FOR_GPU
+    for (auto& node : p.get_processing_order()) {
+        if (node->is_type<gemm>() && node->get_preferred_impl_type() == impl_types::onednn) {
+            for (const auto& fused_prim : node->get_fused_primitives()) {
+                if (fused_prim.is_type<eltwise>() &&
+                    one_of(fused_prim.typed_desc<eltwise>()->mode, eltwise_mode::sum, eltwise_mode::sub, eltwise_mode::prod)) {
+                    auto& data = node->get_dependency(fused_prim.dep_start_idx);
```
isanghao, line 981: `data` usually means data type node... What about `dep`?

src/plugins/intel_gpu/include/intel_gpu/runtime/utils.hpp
```diff
@@ -173,6 +173,16 @@ inline bool any_not_zero(const std::vector<T> vec) {
     return std::any_of(vec.begin(), vec.end(), [](const T& val) { return val != 0; });
 }
 
+template <typename T, typename P>
```
isanghao, line 176: just a question: why do we have separate type of `T` and `P`? Aren't they same? @sshlyapn

vladimir-paramuzov: Looks like a copy-paste from onednn. 
Maybe just use stl based imp insteadl?

```cpp
template <typename T>
inline bool one_of(const T& val, const std::vector<T>& vec) {
    return std::any_of(vec.begin(), vec.end(), [&val](const T& v) { return v == val; });
}
```

src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
sshlyapn: I haven't seen such cases, but yes, it looks like we can handle second case too with using `gemm_layout.get_shape()` as `broadcast_size`

isanghao (changes requested):
Discussed with OneDNN team and they said that gemm input can be ND. In that case, we can just feed our dimension information to OneDNN without broadcasting or dimension folding.

src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
hyunback: Applied

src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
hyunback: It means just dependency

src/plugins/intel_gpu/include/intel_gpu/runtime/utils.hpp
hyunback: Applied

src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
```diff
@@ -958,4 +965,44 @@ void reorder_inputs::run(program& p, layout_optimizer& lo, reorder_factory& rf)
...
+                    const auto prim_id = "broadcast:" + data.id() + "_broadcasted" + std::to_string(idx++);
+                    auto broadcast_prim = std::make_shared<cldnn::broadcast>(prim_id, cldnn::input_info(data.id()), gemm_layout.get_shape(), ov::AxisSet{});
+
+                    auto& broadcast_node = p.get_or_create(broadcast_prim);
+                    p.add_intermediate(broadcast_node, *node, fused_prim.dep_start_idx, true);
+                    broadcast_node.recalc_output_layouts(false);
+                }
+            }
+        }
+    }
```
isanghao, line 1006: sorry for belated comment, but...
Discussed with OneDNN team and they said that gemm input can be ND. In that case, we can just feed our dimension information to OneDNN without broadcasting or dimension folding.

src/plugins/intel_gpu/tests/fusions/gemm_fusion_test.cpp
```diff
@@ -275,6 +279,103 @@ INSTANTIATE_TEST_SUITE_P(fusings_gpu, gemm_2in_scale, ::testing::ValuesIn(std::v
...
+        input_layout("input1", get_input_layout(p, 1)),
+        data("add_data", get_mem(add_data_layout, 1.0f/p.kernel.count())),
+        gemm("gemm_prim", { input_info("input0"), input_info("input1") }, data_types::f32),
+        eltwise("add_prim", { input_info("gemm_prim"), input_info("add_data") }, eltwise_mode::sub, p.default_type),
+        reorder("reorder_bfyx", input_info("add_prim"), p.default_format, data_types::f32)
+    );
+
+    tolerance = default_tolerance(p.default_type);
+    execute(p);
+}
```
isanghao, line 372: I think we can parameterize these three test cases.

hyunback: applied

src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
hyunback: Let's update it next PR

src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
```diff
@@ -958,4 +965,44 @@ void reorder_inputs::run(program& p, layout_optimizer& lo, reorder_factory& rf)
...
+                    auto data_dims = onednn::convert_gemm_tensor(data_layout.get_tensor(),
+                                                                 cldnn::format::dimension(data_layout.format),
+                                                                 false);
+
+                    if (gemm_dims[0] == data_dims[0])
+                        continue;
+
+                    static size_t idx = 0;
+                    const auto prim_id = "broadcast:" + data.id() + "_broadcasted" + std::to_string(idx++);
+                    auto broadcast_prim = std::make_shared<cldnn::broadcast>(prim_id, cldnn::input_info(data.id()), gemm_layout.get_shape(), ov::AxisSet{});
```
isanghao, line 998: note: it will broadcast spatial axis too, and we don't want that. but I think it is OK at this moment because we will remove this code.

isanghao (approved):

PR #16075, opened 2023-03-03
[GPU] Minor fixes for dynamic model

andrew-k-park:
### Details:
 - Fix removing redundant reshapes is skipped so that only the current reshape is not removed if only the input size of dependency reshape is not 1 among the sequence of reshapes in `prepare_primitive_fusing` opt pass
 - Fix to set kernel arguments correctly when only the input tensor of `gather` is dynamic

### Tickets:
 - 104936


e-ddykim (approved):
LGTM

vladimir-paramuzov (approved):

PR #16079, opened 2023-03-03
Andreib/2022.3 myriad plugin obs

andrei-boji-intel:
Revert to using the OBS build firmware for MyriadX plugin.
Changed the dependencies to support the new Firmware version and hash values.


DariaMityagina (approved):
LGTM 👍 
Thanks!

Maxim-Doronin (approved):

PR #16092, opened 2023-03-03
[PT FE]: add translation for aten::cumsum

eaidova:
### Details:
 - *support operator aten::cumsum*

### DIFF
--- a/src/frontends/pytorch/src/op/cumsum.cpp
+++ b/src/frontends/pytorch/src/op/cumsum.cpp
@@ -1,0 +1,30 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "openvino/frontend/pytorch/node_context.hpp"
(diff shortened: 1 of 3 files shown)


src/frontends/pytorch/src/op/cumsum.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include "openvino/op/cum_sum.hpp"
+#include "utils.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_cumsum(NodeContext& context) {
+    // aten::cumsum(Tensor self, int dim, *, ScalarType? dtype=None, Tensor out=None)
```
mvafin, line 16: Please also add check that `out` input is None, or try to support the case when `out` is not None.

eaidova: added

mvafin (approved):

PR #16100, opened 2023-03-05
[PT FE] Support set/get element type, shape and value in PyTorch FE InputModel

mvafin:
### Details:
 - *Implement `get_inputs`, `get_outputs`, `get_place_by_tensor_name`, `set_partial_shape`, `get_partial_shape`, `set_element_type`, `get_element_type`, `set_tensor_value` to allow specifying input element type, partial shape and value before conversion*

### Tickets:
 - *102339*

### DIFF
--- a/src/frontends/pytorch/src/input_model.cpp
+++ b/src/frontends/pytorch/src/input_model.cpp
@@ -69,7 +69,8 @@ Place::Ptr InputModel::get_place_by_tensor_name(const std::string& tensor_name) 
 }
 
 void InputModel::set_partial_shape(const Place::Ptr& place, const ov::PartialShape& shape) {
-    FRONT_END_GENERAL_CHECK(place && place->is_input(), "Provided place is invalid, only inputs are supported for setting shape.");
+    FRONT_END_GENERAL_CHECK(place && place->is_input(),
+                            "Provided place is invalid, only inputs are supported for setting shape.");
     auto pytorch_place = std::dynamic_pointer_cast<pytorch::Place>(place);
     FRONT_END_GENERAL_CHECK(pytorch_place, "Only place produced by PyTorch Frontend is supported");
     auto it = m_descriptors.find(pytorch_place->get_tensor_index());
@@ -91,7 +92,8 @@ ov::PartialShape InputModel::get_partial_shape(const Place::Ptr& place) const {
 }
 
 void InputModel::set_element_type(const Place::Ptr& place, const ov::element::Type& type) {
-    FRONT_END_GENERAL_CHECK(place && place->is_input(), "Provided place is invalid, only inputs are supported for setting element type.");
+    FRONT_END_GENERAL_CHECK(place && place->is_input(),
+                            "Provided place is invalid, only inputs are supported for setting element type.");
     auto pytorch_place = std::dynamic_pointer_cast<pytorch::Place>(place);
     FRONT_END_GENERAL_CHECK(pytorch_place, "Only place produced by PyTorch Frontend is supported");
     auto it = m_descriptors.find(pytorch_place->get_tensor_index());
@@ -113,7 +115,8 @@ ov::element::Type InputModel::get_element_type(const Place::Ptr& place) const {
 }
 
 void InputModel::set_tensor_value(const Place::Ptr& place, const void* value) {
-    FRONT_END_GENERAL_CHECK(place && place->is_input(), "Provided place is invalid, only inputs are supported for setting tensor value.");
+    FRONT_END_GENERAL_CHECK(place && place->is_input(),
+                            "Provided place is invalid, only inputs are supported for setting tensor value.");
     auto pytorch_place = std::dynamic_pointer_cast<pytorch::Place>(place);
     FRONT_END_GENERAL_CHECK(pytorch_place, "Only place produced by PyTorch Frontend is supported");
     auto it = m_descriptors.find(pytorch_place->get_tensor_index());
--- a/src/frontends/pytorch/src/input_model.cpp
+++ b/src/frontends/pytorch/src/input_model.cpp
@@ -4,8 +4,8 @@
 
 #include "input_model.hpp"
 
-#include "utils.hpp"
 #include "place.hpp"
+#include "utils.hpp"
 
 namespace ov {
 namespace frontend {
--- a/src/frontends/pytorch/src/input_model.cpp
+++ b/src/frontends/pytorch/src/input_model.cpp
@@ -62,12 +62,14 @@ std::vector<ov::frontend::Place::Ptr> InputModel::get_outputs() const {
 
 Place::Ptr InputModel::get_place_by_tensor_name(const std::string& tensor_name) const {
     auto place_it = m_name_to_place.find(tensor_name);
-    FRONT_END_GENERAL_CHECK(place_it != m_name_to_place.end(), "Couldn't find Place for provided tensor name.");
-    return place_it->second;
+    if (place_it != m_name_to_place.end()) {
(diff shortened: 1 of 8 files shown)


tests/layer_tests/pytorch_tests/test_rsub.py
```diff
@@ -83,17 +83,17 @@ def forward2(self, lhs, rhs:int):
         return aten_rsub(lhs_type, rhs_type), ref_net, "aten::rsub"
 
     @pytest.mark.parametrize(("lhs_type", "rhs_type"),
-                             [[torch.int32, "int"],
-                            #  [torch.int32, "float"], fp64 produce ov error of eltwise constant fold
-                              [torch.int64, "int"],
-                            #  [torch.int64, "float"], fp64 produce ov error of eltwise constant fold
-                              [torch.float32, "int"],
-                              [torch.float32, "float"],
+                             [[torch.int32, np.int32],
```
eaidova, line 86: not sure that it is correct, rsub operator injected into model only if the second operand is a native python scalar, so it is the reason why I used native python dtypes in test

tests/layer_tests/py_frontend_tests/test_torch_frontend.py
```diff
@@ -0,0 +1,70 @@
...
+        model = torch.jit.script(model)
+        model.eval()
+        model = torch.jit.freeze(model)
+        print(model.inlined_graph)  # will help debugging
+        return model
+
+
+def test_pytorch_fe_set_input_shape():
+    from openvino.frontend.pytorch.decoder import TorchScriptPythonDecoder
+    from openvino.frontend.pytorch.py_pytorch_frontend import _Type as DecoderType
```
eaidova, line 27: how it should be used? I can not see it in test...

eaidova (approved):

src/frontends/pytorch/src/input_model.cpp
```diff
@@ -0,0 +1,127 @@
...
+        auto place_it = m_name_to_place.find(std::to_string(output_idx));
+        FRONT_END_GENERAL_CHECK(place_it != m_name_to_place.end(), "Couldn't find Place for output.");
+        res.push_back(place_it->second);
+    }
+    return res;
+}
+
+Place::Ptr InputModel::get_place_by_tensor_name(const std::string& tensor_name) const {
+    auto place_it = m_name_to_place.find(tensor_name);
+    FRONT_END_GENERAL_CHECK(place_it != m_name_to_place.end(), "Couldn't find Place for provided tensor name.");
```
rkazants, line 65: I think we should return `nullptr` according to the common Frontend API spec.:

```
    /// \brief Returns a tensor place by a tensor name following framework conventions, or
    /// nullptr if a tensor with this name doesn't exist.
    /// \param tensor_name Name of tensor
    /// \return Tensor place corresponding to specified tensor name or nullptr if not exists
    virtual Place::Ptr get_place_by_tensor_name(const std::string& tensor_name) const;

````

File: src/frontends/pytorch/src/input_model.cpp [hunk 4]
```diff
@@ -0,0 +1,127 @@
...
+    FRONT_END_GENERAL_CHECK(pytorch_place, "Only place produced by PyTorch Frontend is supported");
+    auto it = m_descriptors.find(pytorch_place->get_tensor_index());
+    if (it != m_descriptors.end()) {
+        it->second.m_pshape = shape;
+    }
+}
+
+ov::PartialShape InputModel::get_partial_shape(const Place::Ptr& place) const {
+    auto pytorch_place = std::dynamic_pointer_cast<pytorch::Place>(place);
+    FRONT_END_GENERAL_CHECK(pytorch_place, "Only place produced by PyTorch Frontend is supported");
```
rkazants, line 81: I think more user-friendly message should be that you support setting shapes to only model inputs. This messages does not tell anything to the user:)
The same comment for other methods

src/frontends/pytorch/src/input_model.cpp
```diff
@@ -0,0 +1,127 @@
...
+void InputModel::set_tensor_value(const Place::Ptr& place, const void* value) {
+    FRONT_END_GENERAL_CHECK(place->is_input(), "Setting value is supported for input only.");
+    auto pytorch_place = std::dynamic_pointer_cast<pytorch::Place>(place);
+    FRONT_END_GENERAL_CHECK(pytorch_place, "Only place produced by PyTorch Frontend is supported");
+    auto it = m_descriptors.find(pytorch_place->get_tensor_index());
+    if (it != m_descriptors.end()) {
+        auto el_type = it->second.m_type;
+        auto p_shape = it->second.m_pshape;
+        FRONT_END_GENERAL_CHECK(el_type.is_static() && p_shape.is_static(),
+                                "Shape and type must be statically defined before calling set_tensor_value");
```
rkazants, line 118: Is it possible to deduce type from the model without requesting it?

src/frontends/pytorch/src/place.cpp
```diff
@@ -0,0 +1,47 @@
...
+    FRONT_END_GENERAL_CHECK(im, "PyTorch Place requires PyTorch InputModel class.");
+    const auto& inputs = im->m_model_decoder->inputs();
+    const auto& outputs = im->m_model_decoder->outputs();
+    auto in_it = std::find(inputs.begin(), inputs.end(), tensor_index);
+    if (in_it != inputs.end()) {
+        m_is_input = true;
+        const auto& debug_name = im->m_model_decoder->get_input_debug_name(std::distance(inputs.begin(), in_it));
+        if (debug_name != m_names.at(0)) {
+            m_names.push_back(debug_name);
+        }
```
rkazants, line 30: what is `debug_name`? Interesting name:)
May be it makes sense to put a comment about this logic.

rkazants:
Also, will you clean-up MO code for PT FE in the separate PR?

eaidova:
> Also, will you clean-up MO code (from that WA) for PT FE in the separate PR?

yes, it is my responsibility. I'm working on that

tests/layer_tests/pytorch_tests/test_rsub.py
mvafin: It is still python scalar in pytorch model. This type goes to np.array function because of that we need to specify 32bit type otherwise it will be 64bit.

src/frontends/pytorch/src/input_model.cpp
mvafin: Are you talking about reverse infer? We do not do any conversion on this point, so we cannot deduce it now, it is possible to do it after conversion, but it is not guaranteed and it seems to be out of scope of this API. If the model is traced, then type and shape will exist in descriptors map, so in that case everything is expected, type will also be known if it was annotated in initial model even if scripting is used.

src/frontends/pytorch/src/place.cpp
mvafin: debug_name is the name from pytorch. It stores name of the tensor obtained from name of the variable, it can be useful for user, so we want to provide him with this name

tests/layer_tests/pytorch_tests/test_rsub.py
mvafin: True, but we currently decided to convert python native types to 32bit in frontend, because this will happen any way and we want to do it earlier, cannot say it is the best approach, but it works fine for now.

src/frontends/pytorch/src/input_model.cpp
mvafin: Agree that for the case of `nullptr` it is misleading

rkazants (approved):

PR #16114, opened 2023-03-06
Review pooling classes for shape inference aspects

praasz:
### Details:
 - Review for interval dimension and label propagation.
 - Adaptive pooling preserve interval values and labels on output shape input.
 - Make common shape inference for pooling operators.
 - Add dimension util for shape inference.

### Tickets:
 - 97211
 - 97327
 - 97376
 - 97425
 - 97474
 - 97528


tomdol (approved):

src/core/shape_inference/include/avg_pool_shape_inference.hpp
```diff
@@ -0,0 +1,61 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include <ngraph/util.hpp>
```
ilyachur, line 7: Can we use only new headers?
Also please use "" instead of <> for OV headers

src/core/shape_inference/include/dimension_util.hpp
```diff
@@ -0,0 +1,82 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+#pragma once
+
+#include <cstdint>
+#include <ngraph/util.hpp>
```
ilyachur, line 7: The same

src/core/shape_inference/include/max_pool_shape_inference.hpp
```diff
@@ -0,0 +1,58 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+// #include "adaptive_pool_shape_inference.hpp"
```
ilyachur, line 7: Can we remove unused header?

src/core/shape_inference/include/avg_pool_shape_inference.hpp
praasz: Yes, this old util header is not required here.

src/core/shape_inference/include/dimension_util.hpp
praasz: Corrected.

ilyachur (approved):

PR #16116, opened 2023-03-06
[GPU] Change lws to avoid synchronization issue in nonzero_count

ahnyoung-paul:
### Details:
 - *Set 1 work group to avoid synchronization issue for summation of nonzero count*
 - *...*

### Tickets:
 - *102467*

### DIFF
--- a/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/count_nonzero_ref.cl
+++ b/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/count_nonzero_ref.cl
@@ -4,47 +4,37 @@
 
 #include "include/batch_headers/common.cl"
 
-#define INPUT0_GET_INDEX1(idx_order) INPUT0_GET_INDEX(idx_order)
-
 KERNEL (count_nonzero_ref)(
     OPTIONAL_SHAPE_INFO_ARG
     const __global INPUT0_TYPE* input,
     volatile __global OUTPUT_TYPE* output)
 {
-    const uint gdim0 = (uint)get_global_id(0);
-    const uint gdim1 = (uint)get_global_id(1);
-    const uint gdim2 = (uint)get_global_id(2);
+    const uint local_idx = get_local_id(0);
+    const uint num_work_items = get_global_size(0);
+    const uint data_size = DATA_SIZE;
+    const uint items_num = data_size / num_work_items;
+    const uint leftovers = data_size - (items_num * num_work_items);
 
-    if (gdim0 == 0 && gdim1 == 0 && gdim2 == 0) {
+    if (local_idx == 0)
         output[0] = 0;
-    }
     barrier(CLK_GLOBAL_MEM_FENCE);
 
-    #if INPUT0_DIMS == 6
-        #define INPUT_ORDER b,f,w,z,y,x
-        const uint x = gdim0 % INPUT0_SIZE_X;
-        const uint y = gdim0 / INPUT0_SIZE_X;
-        const uint z = gdim1 % INPUT0_SIZE_Z;
-        const uint w = gdim1 / INPUT0_SIZE_Z;
-    #elif INPUT0_DIMS == 5
-        #define INPUT_ORDER b,f,z,y,x
-        const uint x = gdim0;
-        const uint y = gdim1 % INPUT0_SIZE_Y;
-        const uint z = gdim1 / INPUT0_SIZE_Y;
-    #elif INPUT0_DIMS == 4
-        #define INPUT_ORDER b,f,y,x
-        const uint x = gdim0;
-        const uint y = gdim1;
-    #endif
-    const uint f = gdim2 % INPUT0_FEATURE_NUM;
-    const uint b = gdim2 / INPUT0_FEATURE_NUM;
+    uint workitem_nonzero_count = 0;
+    uint iter = 0;
+    for (; iter < items_num; iter++) {
+        uint count = (input[num_work_items * iter + local_idx] == INPUT0_VAL_ZERO) ? 0 : 1;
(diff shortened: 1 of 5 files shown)


src/plugins/intel_gpu/src/kernel_selector/kernels/non_zero/count_nonzero_kernel_ref.cpp
```diff
@@ -58,8 +56,8 @@ CommonDispatchData CountNonzeroKernelRef::SetDefault(const count_nonzero_params&
                        {Tensor::DataChannelName::FEATURE, Tensor::DataChannelName::BATCH}};
     }
 
-    dispatchData.lws =
-        GetOptimalLocalWorkGroupSizes(dispatchData.gws, params.engineInfo, in_layout, out_layout, dims_by_gws);
+    // Set 1 work group to avoid synchornization issue for summation of nonzero counting.
+    dispatchData.lws = dispatchData.gws;
```
yeonbok, line 60: What if we exceed # available work group size? 
Should be changed in a way like : 
- Divide data by params.engineInfo.maxWorkGroupSize
- e.g., if max # of wgs is 512 and the total # of data is 1024, each work item should handle 2 data. 
-

ahnyoung-paul: Thank you for your comments. I will update it.

ahnyoung-paul: Update unit tests

ahnyoung-paul: Update kernel. please review it again

src/plugins/intel_gpu/tests/test_cases/non_zero_gpu_test.cpp
```diff
@@ -53,6 +53,82 @@ TEST(test_count_non_zero, 5d_fp16_1_3_2_1_2) {
...
+
+    for (size_t idx = 0; idx < input_shapes.size(); idx++) {
+        const size_t input_length = input_shapes[idx];
+
+        ov::PartialShape p_shape  = { 1, static_cast<long int>(input_length) };
+        auto in_layout = layout(p_shape, data_types::i32, format::bfyx);
+        auto input_mem = engine.allocate_memory(in_layout);
+
+        std::vector<int32_t> in_data(input_length);
+        std::fill(in_data.begin(), in_data.end(), 1);
```
yeonbok, line 118: This data is too simple to verify nonzero. 
* Could you fix this test to test more random data? 
e.g., 
1) fill random number (including floating point test) to the inputs 
2) add 0s to random positions 

* Also could you add test for blocked format too? 

*We dont need to validate such a many shapes. I think, some 2~3 shapes which are small than 256, larger than 1024, and nod divisable to power of two. We'll need to test 
1) more random numbers distributions (which contains definitly zeros) 
2) at least one float test and one integer test 
3) bfyx, bfzyx, bs_yx_fsv16  (here the feature shape should be > 16)

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/count_nonzero_ref.cl
```diff
@@ -4,47 +4,37 @@
...
-        const uint y = gdim1;
-    #endif
-    const uint f = gdim2 % INPUT0_FEATURE_NUM;
-    const uint b = gdim2 / INPUT0_FEATURE_NUM;
+    uint workitem_nonzero_count = 0;
+    uint iter = 0;
+    for (; iter < items_num; iter++) {
+        uint count = (input[num_work_items * iter + local_idx] == INPUT0_VAL_ZERO) ? 0 : 1;
+        workitem_nonzero_count += count;
+    }
```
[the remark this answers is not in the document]
ahnyoung-paul: I checked there is no performance drop but I changed the local indexing as your comments.
Please review it again. thanks

src/plugins/intel_gpu/tests/test_cases/non_zero_gpu_test.cpp
ahnyoung-paul: I added bfyx, bfzyx, and bfzwyx format test but does not include bs_yx_fsv16 format because current nonzero_count does not support fsv16 format.

src/plugins/intel_gpu/tests/test_cases/non_zero_gpu_test.cpp
```diff
@@ -15,6 +15,19 @@
...
+    static std::default_random_engine generator(random_seed);
+    int k = 8; // 1/k is the resolution of the floating point numbers
+    std::uniform_int_distribution<int> distribution(k * min, k * max);
+    std::vector<T> v(input_size);
+    for (size_t i = 0; i < input_size; ++i) {
+        v[i] = (T)distribution(generator);
+        v[i] /= k;
+    }
+    return v;
+}
```
yeonbok, line 29: Can we ensure that the generated values have "0"s definitely? 
How about 
1) generate with original utility generate_random
2) select random number of "# of zeros" // count_zero (e.g., 10~100 zeros)
3) generate random "indices"  to set input mem to be zero
4) set the value of those indices to zero

ahnyoung-paul: Update random generation input function
- Use original random generation util function
- Set minimum number of zero inputs
- Generate random indices to set zero in input buffer

yeonbok (approved):

PR #16122, opened 2023-03-06
[GNA] Update GNA library version to 3.5.0.1906

rjeziers:
### Details:
- Changed GNA library version from 3.0.0.1910 to 3.5.0.1906

### Tickets:
 - 94946


sirzabek (approved):

sirzabek (commented):
Shouldn't we change the default target to 3.5 as well?

dorloff:
> Shouldn't we change the default target to 3.5 as well?

I don't think so - the new HW is not released yet.

dorloff (approved):

PR #16130, opened 2023-03-07
[GPU] Changed impls cache key type to avoid hash collisions

vladimir-paramuzov:
### Details:
 - `hash_combine` has high rate of hash collisions, thus updated ImplementationsCache to use `kernel_impl_params` as key instead of `size_t` (hash) and declared comparison + hash operators for impl params.

### Tickets:
 - *103415*


yeonbok (approved):

p-durandin (approved):

src/plugins/intel_gpu/include/intel_gpu/graph/program.hpp
```diff
@@ -249,8 +250,14 @@ struct program {
     std::pair<int64_t/*const alloc*/, int64_t/*general alloc*/> get_estimated_device_mem_usage();
 
     void remove_kernel(kernel_id id);
-    void calc_nodes_hash();
 
+    struct ImplHasher {
```
sshlyapn, line 254: Does it not make sense to make this structure a part of kernel_impl_params.hpp?

vladimir-paramuzov: Makes sense. Will update it in a next PR

PR #16131, opened 2023-03-07
[CPU] Ngram node fusion

v-Golubev:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


v-Golubev:
@rkazants could you please review a transformation part? Thanks

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/ngram.cpp
```diff
@@ -0,0 +1,223 @@
+// Copyright (C) 2018-2023 Intel Corporation
```
rkazants, line 1: Also, I would recommend to you to add one more test that will prevent the common transformations to break the pattern that we expect for Ngram now.
You can add TF FE unit-test with extracted sub-graph of `Ngram` in pbtxt and pass it into core read and compile and see if your pattern will apply. Because now we have some risk that some previous transformation will break it.

dmitry-gorokhov:
@nshchego Please take a look from CPU implementation perspective

src/plugins/intel_cpu/src/nodes/ngram.h
```diff
@@ -0,0 +1,49 @@
...
+#include <string>
+#include <vector>
+
+namespace ov {
+namespace intel_cpu {
+namespace node {
+
+class Ngram : public Node {
+public:
+    Ngram(const std::shared_ptr<ngraph::Node>& op, const GraphContext::CPtr context);
```
nshchego, line 19: - Use `ov` instead of `ngraph` wherever possible.
- Use reference GraphContext::CPtr &context.

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/ngram.cpp
```diff
@@ -0,0 +1,223 @@
...
+static std::shared_ptr<ov::Node> getReshape(const std::shared_ptr<ov::Node> data,
+                                            const std::vector<int64_t>& requested_shape,
+                                            const ov::element::Type& prc) {
+    auto requested_shape_node = ov::opset1::Constant::create(prc, {requested_shape.size()}, requested_shape);
+    return std::make_shared<ov::opset1::Reshape>(data, requested_shape_node, true);
+}
+
+static std::shared_ptr<ov::Model> initNgram(std::vector<ov::PartialShape>& input_shapes, const size_t k) {
+    const auto data_et = ov::element::f32;
+    const auto idces_et = ov::element::i32;
```
nshchego, line 46: Could you parameterize precisions to support others in future?

src/plugins/intel_cpu/src/nodes/ngram.h
v-Golubev: Done, thanks

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/ngram.cpp
v-Golubev: Good idea, thanks for the suggestion. Done

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/ngram.cpp
v-Golubev: The test is added, thanks for the suggestion. Could you please take a look?

src/plugins/intel_cpu/src/nodes/ngram.cpp
```diff
@@ -0,0 +1,164 @@
...
+        size_t dstWindowBias = 0;
+
+        const size_t niter = batchLenghts[batchIdx + 1] - batchLenghts[batchIdx];
+        const size_t srcBatchBias = batchLenghts[batchIdx] * windowStride;
+        const size_t dstBatchBias = batchLenghts[batchIdx] * windowStride * k;
+        for (size_t i = 0; i < niter; ++i) {
+            const size_t curLeftPad = leftPad >= i ? leftPaddingSize - i * windowStride : 0;
+            const size_t curRightPad = rightPad >= niter - 1 - i ? rightPaddingSize - (niter - 1 - i) * windowStride : 0;
+
+            memset(dstData + dstBatchBias + dstWindowBias, 0, curLeftPad * sizeof(float));
```
nshchego, line 139: Are the curLeftPad and curRightPad comparable with destination blob size? Perhaps it is cheaper to fill all blob by zero once before parallel loop?

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/ngram.cpp
```diff
@@ -0,0 +1,232 @@
...
+namespace CPUSubgraphTestsDefinitions {
+
+typedef std::tuple<
+    std::vector<InputShape>,
+    ElementType,
+    ElementType,
+    size_t
+> NgramTestParams;
+
+static std::shared_ptr<ov::Node> getStridedSlice(const std::shared_ptr<ov::Node> data,
```
nshchego, line 30: Please use references for shared_ptr as well, due to their copying produce additional routine.

src/plugins/intel_cpu/src/ngraph_transformations/op/ngram.hpp
```diff
@@ -0,0 +1,35 @@
...
+#pragma once
+
+#include <openvino/core/node.hpp>
+#include <openvino/op/op.hpp>
+
+namespace ov {
+namespace intel_cpu {
+/**
+ * The operation flattens embedding tensor of token vectors and traverse it by a sliding window of size times the original embedding sizes.
+ * Inputs:
```
nshchego, line 14: Please add allowed data types for Inputs/Outputs and classifier Required/Optional for Inputs.

src/plugins/intel_cpu/src/nodes/ngram.cpp
```diff
@@ -0,0 +1,164 @@
...
+}
+
+void Ngram::execute(dnnl::stream strm) {
+    auto* srcData = reinterpret_cast<const float*>(getParentEdgeAt(0)->getMemoryPtr()->GetPtr());
+    auto* srcIndices = reinterpret_cast<const int*>(getParentEdgeAt(1)->getMemoryPtr()->GetPtr());
+    auto* dstData = reinterpret_cast<float*>(getChildEdgeAt(0)->getMemoryPtr()->GetPtr());
+
+    std::vector<size_t> batchLenghts{0};
+    batchLenghts.reserve(numIdces + 1);
+    for (size_t i = idcesStride; i < idcesShapeSize; i += idcesStride) {
```
nshchego, line 117: Looks like that it's easy to support INT64 here. Just need to dub this loop. Could you please do so?

v-Golubev: In terms of semantic, I64 indices makes sense only in the case when number of embeddings more than I32::max. I don't think that we can meet such cases in real scenario. Given the above, do we really need I64 support here?
If yes, could you please clarify what means "Just need to dub this loop"? Didn't get it.

nshchego: We are going to add optional support of INT64 and to avoid excessive Convert i32->i64 better to add native support in node.
The possible solution:
`
auto mem = getParentEdgeAt(1)->getMemoryPtr();
if (mem->getDesc().getPrecision() == Precision::I64) {
    auto* srcIndices = reinterpret_cast<const int64_t*>(mem->GetPtr());
    for (size_t i = idcesStride; i < idcesShapeSize; i += idcesStride) {
        if (srcIndices[i - idcesStride] != srcIndices[i]) {
            batchLenghts.push_back(i / idcesStride);
        }
    }
} else if (mem->getDesc().getPrecision() == Precision::I32) {
    auto* srcIndices = reinterpret_cast<const int32_t*>(mem->GetPtr());
    for (size_t i = idcesStride; i < idcesShapeSize; i += idcesStride) {
        if (srcIndices[i - idcesStride] != srcIndices[i]) {
            batchLenghts.push_back(i / idcesStride);
        }
    }
}`
or use conditional compilation switch.

v-Golubev: Thank you for the detailed explanation! Will follow your suggestion

src/plugins/intel_cpu/src/nodes/ngram.cpp
v-Golubev: The ratio of the number of zero and non-zero elements depends on the `k` attribute and the input shape. However, quite often a significant part is occupied by paddings (up to 50-60%). So I changed the logic: all blob is filled by zero before parallel loop,

src/plugins/intel_cpu/src/ngraph_transformations/op/ngram.hpp
v-Golubev: Added, thanks

dmitry-gorokhov (approved):

PR #16140, opened 2023-03-07
[PT FE] Fix failing translation of aten::index_put_

mmikolajcz:
### Details:
 - *Rewrite current translation into transformation*
 - *Add test for previously failing case from models*
 - *Add deepcopy to inputs to PyTorch tests due to issues with models modifying test data*

### Tickets:
 - *105141*

### Changes from previous implementation:
Added ability to parse unconverted `prim::ListConstruct` framewok nodes as input for `indices`.
Previous implementation required `prim::ListConstruct` to return list items concatenated into single tensor with static rank and static first dimension during translation to read length of list.
Issue was that translator of `prim::ListConstruct` wasn't able to convert `indices` input from detectron, causing failure at check for static rank.
Most significant change was implemented [around line 68](https://github.com/openvinotoolkit/openvino/pull/16140/files#diff-f7a418f5c894e763fce7c6314e751de0f5ad1e97e9657e31af262522c8613e21R68), where length of list and Tensors inside are either accessed directly from unconverted `prim::ListConstruct` if it is available, or in else block, where they use approach that was used previously in translation.


src/frontends/pytorch/src/op/index_put_.cpp
```diff
@@ -24,101 +12,13 @@ namespace op {
...
-    if (accumulate) {
-        auto zeros = generate_zeros_with_convertlike(context, input_shape, input);
-        result = context.mark_node(std::make_shared<v3::ScatterNDUpdate>(zeros, index, values));
-        result = context.mark_node(std::make_shared<v1::Add>(input, result));
-    } else {
-        result = context.mark_node(std::make_shared<v3::ScatterNDUpdate>(input, index, values));
-    }
-
-    return {result};
+    // Pass as PtFrameworkNode to register as `inplace_op`. Conversion to OV operators is done as transformation.
```
mvafin, line 16: In fact this is not required, regular fw node creation should detect the fact that the operation is modifying input if it has "!" in schema. But it is okay to leave this translation here. Also I would prefer that custom translator is not marked as `inplace_op`, it seems much less obvious what is happen during translation.

mvafin:
> ### Details:
> * _Rewrite current translation into transformation_
> * _Add test for previously failing case from models_
> * _Add deepcopy to inputs to PyTorch tests due to issues with models modifying test data_
> 
> ### Tickets:
> * _105141_

Could you also add more details what is changed in transformation vs the previous implementation

src/frontends/pytorch/src/op/index_put_.cpp
mmikolajcz: Removing it from op_table was my first idea, however, it seem to fail. Creating simple fw_node was the simplest way I though for fixing it, didn't known that there is more generic approach, I'll check it and look what went wrong.

mvafin: The generic approach may fail if schema doesn't contain "!". It makes sense to have translator in that case.

mmikolajcz: I've checked that schema contains "!" and code works correctly, issue with number of outputs throw me out, however, now I see that it may be intended [since it add one additional output](https://github.com/openvinotoolkit/openvino/blob/master/src/frontends/pytorch/src/utils.cpp#L172). Should then transformation replace both outputs with the same node?

mvafin: Ah yeah, that is the downside of generic approach, in your case it would be better to have separate translator to not handle one extra output

mmikolajcz: Yeah, since there is also out-place version of aten::index_put_, I guess that adding dummy translation and using inplace_op might be more generic since then transformation should be able to handle both uses

mvafin (approved):

PR #16142, opened 2023-03-07
[TF FE] Support Multi-output Body Graph nodes with named ports

rkazants:
**Details:** This is a corner case because body graph nodes have named output ports. This allows to support custom RetinaNet model.

**Ticket:** 97993

### DIFF
--- a/src/frontends/common/include/openvino/frontend/decoder.hpp
+++ b/src/frontends/common/include/openvino/frontend/decoder.hpp
@@ -50,6 +50,7 @@ public:
 
 class FRONTEND_API DecoderBase {
 public:
+    using OpTypeByName = std::unordered_map<std::string, std::string>;
     /// \brief Get attribute value by name
     ///
     /// \param name Attribute name
@@ -67,6 +68,17 @@ public:
     virtual void get_input_node(size_t input_port_idx,
                                 std::string& producer_name,
                                 size_t& producer_output_port_index) const = 0;
+
+    /// \brief Get a producer name and its output port index
+    ///
+    /// \param input_port_idx              Input port index by which data is consumed
+    /// \param producer_name               A producer name
+    /// \param producer_output_port_index  Output port index from which data is generated
+    /// \param op_type_by_name             Map of operation name to their types
+    virtual void get_input_node(size_t input_port_idx,
+                                std::string& producer_name,
+                                size_t& producer_output_port_index,
+                                const OpTypeByName& op_type_by_name) const = 0;
 
     /// \brief Get operation type
     virtual const std::string& get_op_type() const = 0;
--- a/src/frontends/tensorflow/src/decoder_proto.cpp
+++ b/src/frontends/tensorflow/src/decoder_proto.cpp
@@ -285,16 +285,53 @@ size_t DecoderProto::get_input_size() const {
     return m_node_def->input_size();
 }
 
-void DecoderProto::get_input_node(size_t input_port_idx,
-                                  std::string& producer_name,
-                                  size_t& producer_output_port_index) const {
-    // Body graph nodes may have two colons `:`, for example,
-    // producer_name:z:2 means that producer operation name is `producer_name`
-    // and output port is 2
-    std::string producer_port_name = m_node_def->input(static_cast<int>(input_port_idx));
+void parse_producer_name(const std::string& producer_port_name,
+                         std::string& producer_name,
+                         size_t& producer_output_port_index,
+                         const DecoderBase::OpTypeByName& op_type_by_name) {
+    using OutputPortIdxMax = std::unordered_map<std::string, int>;
+    // create a table of operation type and its output ports
+    // for which we specify output port indices manually
+    // it is mainly affects multiple output operations
+    // extract this information from tensorflow/core/ops/*.cc files
+    const OutputPortIdxMax output_port_idx_map = {
+        {"TopK:indices", 1},
+        {"TopKV2:indices", 1},
+        {"CTCGreedyDecoder:decoded_values", 1},
+        {"CTCGreedyDecoder:decoded_shape", 2},
+        {"CTCGreedyDecoder:log_probability", 3},
+        {"CTCGreedyDecoder:log_probability", 3},
+        {"FusedBatchNorm:batch_mean", 1},
+        {"FusedBatchNorm:batch_variance", 2},
+        {"FusedBatchNormV2:batch_mean", 1},
+        {"FusedBatchNormV2:batch_variance", 2},
+        {"FusedBatchNormV3:batch_mean", 1},
+        {"FusedBatchNormV3:batch_variance", 2},
+    };
+    // Body graph nodes may have two colons `:` input names, for example,
+    // `TopKV2Name:indices:0` means that producer operation name is `TopKV2Name`
+    // the middle name is output port name of the producer `indices` that means
+    // the second output port of TopKV2 is used.
+    // The first output port of TopKV2 is described as `TopKV2Name:values:0`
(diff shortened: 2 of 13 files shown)


src/frontends/common/include/openvino/frontend/decoder.hpp
```diff
@@ -50,6 +50,7 @@ class IDecoder {
 
 class FRONTEND_API DecoderBase {
 public:
+    using OpTypeByName = std::unordered_map<std::string, std::string>;
```
gkrivor, line 53: Please, add description

src/frontends/tensorflow/src/decoder_proto.cpp
```diff
@@ -285,16 +285,53 @@ size_t DecoderProto::get_input_size() const {
...
 }
 
-void DecoderProto::get_input_node(size_t input_port_idx,
-                                  std::string& producer_name,
-                                  size_t& producer_output_port_index) const {
-    // Body graph nodes may have two colons `:`, for example,
-    // producer_name:z:2 means that producer operation name is `producer_name`
-    // and output port is 2
-    std::string producer_port_name = m_node_def->input(static_cast<int>(input_port_idx));
+void parse_producer_name(const std::string& producer_port_name,
```
gkrivor, line 288: Don't see documentation, check please

src/frontends/tensorflow/src/decoder_proto.hpp
```diff
@@ -18,6 +18,11 @@ namespace ov {
 namespace frontend {
 namespace tensorflow {
 
+void parse_producer_name(const std::string& producer_port_name,
```
gkrivor, line 21: Here is also no documentation

src/frontends/tensorflow/src/decoder_proto.cpp
rkazants: below you can see a couple of comments what this function is doing. And step-by-step it explains this in detail

gkrivor: It doesn't describe a function definition, overall functionality and expected parameters.

src/frontends/tensorflow/src/decoder_proto.cpp
```diff
@@ -285,16 +285,53 @@ size_t DecoderProto::get_input_size() const {
...
+    // The first output port of TopKV2 is described as `TopKV2Name:values:0`
     auto first_colon = producer_port_name.find_first_of(":");
     auto last_colon = producer_port_name.find_last_of(":");
-    if (first_colon != std::string::npos && last_colon != std::string::npos) {
+    if (first_colon != std::string::npos && first_colon < last_colon) {
+        // we have at least two colons producer_name:output_port_name:port_idx
+        producer_name = producer_port_name.substr(0, first_colon);
+        auto port_id = producer_port_name.substr(last_colon + 1);
+        auto port_name = producer_port_name.substr(first_colon + 1, last_colon - first_colon - 1);
+        FRONT_END_GENERAL_CHECK(!port_id.empty() && std::all_of(port_id.begin(), port_id.end(), ::isdigit),
```
gkrivor, line 323: Why it may contain only digits? Please, refer on documentation it could only decimal indeces.

src/frontends/tensorflow/src/decoder_proto.cpp
```diff
@@ -285,16 +285,53 @@ size_t DecoderProto::get_input_size() const {
...
+    // the second output port of TopKV2 is used.
+    // The first output port of TopKV2 is described as `TopKV2Name:values:0`
     auto first_colon = producer_port_name.find_first_of(":");
     auto last_colon = producer_port_name.find_last_of(":");
-    if (first_colon != std::string::npos && last_colon != std::string::npos) {
+    if (first_colon != std::string::npos && first_colon < last_colon) {
+        // we have at least two colons producer_name:output_port_name:port_idx
+        producer_name = producer_port_name.substr(0, first_colon);
+        auto port_id = producer_port_name.substr(last_colon + 1);
+        auto port_name = producer_port_name.substr(first_colon + 1, last_colon - first_colon - 1);
```
gkrivor, line 322: I don't see a check for empty string, is it expected? what should it do in such case?

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -130,6 +131,7 @@ void TranslateSession::translate_graph(const ov::frontend::InputModel::Ptr& inpu
     for (const auto& operation_place : operation_places) {
         auto operation_decoder = operation_place->get_decoder();
         auto operation_name = operation_place->get_names()[0];
+        op_type_by_name[operation_name] = operation_decoder->get_op_type();
```
gkrivor, line 134: Why only first item is using? Why dont you use "for(name : get_names) { ... }" ?

If it is a limitation - could you provide a description

src/frontends/tensorflow/tests/test_models/models_pbtxt/partitioned_call2.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import tensorflow.compat.v1 as tf
+
```
gkrivor, line 5: is it necessary to use double line breaks between blocks? why should it mean?

src/frontends/tensorflow_common/include/helper_ops/internal_operation.hpp
```diff
@@ -35,6 +35,14 @@ class DecoderFake : public ov::frontend::DecoderBase {
                                       "Internal error: the get_input_node method of the fake node decoder is invoked.");
     }
 
+    void get_input_node(size_t input_port_idx,
+                        std::string& producer_name,
+                        size_t& producer_output_port_index,
+                        const OpTypeByName& op_type_by_name) const override {
+        FRONT_END_OP_CONVERSION_CHECK(false,
```
gkrivor, line 42: Not sure I understand why you use macros FRONT_END_OP_CONVERSION_CHECK with overall meaning "not implemented" in this place, but FRONT_END_NOT_IMPLEMENTED later.

src/frontends/tensorflow/tests/test_models/models_pbtxt/partitioned_call2.py
rkazants: this is PyCharm formatting that we follow to.

src/frontends/tensorflow/src/translate_session.cpp
rkazants: operation has only one name. You mix up with tensor names. No limitation is here

src/frontends/tensorflow/src/decoder_proto.cpp
rkazants: in case of empty string there will be no first_colon  and last_colon to be endpos. Not valid comment

src/frontends/tensorflow/src/decoder_proto.cpp
rkazants: described above in the comment

src/frontends/tensorflow/src/decoder_proto.hpp
rkazants: Not a blocker for merge because this is internal function. If this is critical, I can fix it in separate PR.

mvafin (approved):
This solution has issues with extensibility, for example if user will implement extension for operation that have multiple outputs he cannot modify the map that stores such operation names. Better would be to implement hybrid conversion rules that can return named outputs  and indexed outputs. This is out of scope of this bugfix, ticket for this task is created: 105588.

PR #16150, opened 2023-03-08
Changed throw ov::Exception to macro

ilyachur:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


src/plugins/intel_gpu/src/graph/include/primitive_type_base.h
```diff
@@ -55,7 +55,7 @@ struct primitive_type_base : primitive_type {
                                         "[GPU] Original name: ", p->origin_op_name, "\n"
                                         "[GPU] Original type: ", p->origin_op_type_name, "\n"
                                         "[GPU] Reason: ", e.what());
-            throw ov::Exception(ss.str());
+            OPENVINO_UNREACHABLE(ss.str());
```
vladimir-paramuzov, line 58: Can we have a macro with better name for such cases? "Unreachable" sounds misleading.
OPENVINO_THROW maybe?

ilya-lavrenov (approved):
agreed to rename in a separate PR

src/plugins/intel_gpu/src/graph/include/primitive_type_base.h
ilyachur: Will do it in the separate PR :)

PR #16151, opened 2023-03-08
Remove Myriad public headers

ilyachur:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


ilya-lavrenov (approved):
Original idea was to deprecate and remove these headers together with API 1.0
But maybe we can remove them directly.

PR #16168, opened 2023-03-09
[CPU] Gather node shape infer

xuchen-intel:
### Details:
 - *CPU plugin specific shape infer for Gather node.*

### Tickets:
 - *100511*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/gather.cpp
+++ b/src/plugins/intel_cpu/src/nodes/gather.cpp
@@ -44,14 +44,16 @@ bool Gather::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, std
 
 class GatherShapeInfer : public ShapeInferEmptyPads {
 public:
-    GatherShapeInfer(bool isAxisInputConst, int axis, int batchDims) : m_isAxisInputConst(isAxisInputConst),
-                     m_axis(axis), m_batchDims(batchDims) {}
+    GatherShapeInfer(bool isAxisInputConst, bool isIndicesScalar, int axis, int batchDims) : m_isAxisInputConst(isAxisInputConst),
+                     m_isIndicesScalar(isIndicesScalar), m_axis(axis), m_batchDims(batchDims) {}
     Result infer(const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
                  const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
         static constexpr size_t GATHER_DATA = 0, GATHER_INDICES = 1, GATHER_AXIS = 2;
 
         const auto& input_shape = input_shapes[GATHER_DATA].get();
-        const auto& indices_shape = input_shapes[GATHER_INDICES].get();
+
+        // Use VectorDims{} instead of {1} for Scalar
+        const auto& indices_shape = m_isIndicesScalar ? VectorDims{} : input_shapes[GATHER_INDICES].get();
 
         if (!m_isAxisInputConst) {
             if (data_dependency.at(GATHER_AXIS)->getDesc().getPrecision() != Precision::I32) {
@@ -80,6 +82,7 @@ public:
 
 private:
     bool m_isAxisInputConst = false;
+    bool m_isIndicesScalar = false;
     int m_axis = 0;
     int m_batchDims = 0;
 };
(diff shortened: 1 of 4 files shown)


src/plugins/intel_cpu/src/nodes/gather.cpp
```diff
@@ -41,8 +42,64 @@ bool Gather::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, std
...
+class GatherShapeInfer : public ShapeInferEmptyPads {
+public:
+    GatherShapeInfer(bool isAxisInputConst, int axis, int batchDims, IShapeInfer::port_mask_t port_mask) :
+                     m_isAxisInputConst(isAxisInputConst), m_axis(axis), m_batchDims(batchDims), m_port_mask(port_mask) {}
+    Result infer(const std::vector<std::reference_wrapper<const VectorDims>>& input_shapes,
+                 const std::unordered_map<size_t, MemoryPtr>& data_dependency) override {
+        static constexpr size_t GATHER_DATA = 0, GATHER_INDICES = 1, GATHER_AXIS = 2;
+
+        auto input_shape = input_shapes[GATHER_DATA].get();
+        auto indices_shape = input_shapes[GATHER_INDICES].get();
```
maxnick, line 54:
```suggestion
        const auto& input_shape = input_shapes[GATHER_DATA].get();
        const auto& indices_shape = input_shapes[GATHER_INDICES].get();
```
To avoid copy

src/plugins/intel_cpu/src/nodes/gather.cpp
```diff
@@ -41,8 +42,64 @@ bool Gather::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, std
...
+
+        if (!m_isAxisInputConst)
+            m_axis = reinterpret_cast<const int32_t *>(data_dependency.at(GATHER_AXIS)->GetPtr())[0];
```
maxnick, line 57: I think it won't be superfluous to check the memory precision, to make sure it is I32 and throw if it isn't.

src/plugins/intel_cpu/src/nodes/gather.cpp
```diff
@@ -41,8 +42,64 @@ bool Gather::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, std
...
+        if (m_batchDims < 0)
+            m_batchDims += indices_shape.size();
+
+        VectorDims output_shape;
+        output_shape.reserve(input_shape.size() + indices_shape.size() - m_batchDims - 1);
+        output_shape.insert(output_shape.end(), input_shape.begin(), input_shape.begin() + m_axis);
+        output_shape.insert(output_shape.end(), indices_shape.begin() + m_batchDims, indices_shape.end());
+        output_shape.insert(output_shape.end(), input_shape.begin() + m_axis + 1, input_shape.end());
+
+        return {{output_shape}, ShapeInferStatus::success};
```
maxnick, line 70:
```suggestion
        return {{std::move(output_shape)}, ShapeInferStatus::success};
```
To avoid copy.

src/plugins/intel_cpu/src/nodes/gather.cpp
```diff
@@ -41,8 +42,64 @@ bool Gather::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, std
...
+private:
+    bool m_isAxisInputConst = false;
+    int m_axis = 0;
+    int m_batchDims = 0;
+    port_mask_t m_port_mask;
+};
+
+class GatherShapeInferFactory : public ShapeInferFactory {
+public:
+    GatherShapeInferFactory(std::shared_ptr<ov::Node> op, IShapeInfer::port_mask_t port_mask) : m_op(op), m_port_mask(port_mask) {}
```
maxnick, line 85:
```suggestion
    GatherShapeInferFactory(std::shared_ptr<ov::Node> op) : m_op(op), m_port_mask(port_mask) {}
```
There is no need to pass the port mask as a parameter. We can simply return `PortMask(GATHER_AXIS)` in the `get_port_mask()` call.

src/plugins/intel_cpu/src/nodes/gather.cpp
xuchen-intel: Applied. Thanks Maksim!

src/plugins/intel_cpu/src/nodes/gather.cpp
xuchen-intel: Applied. Thanks Maksim!

src/plugins/intel_cpu/src/nodes/gather.cpp
xuchen-intel: Good point! Applied. Thanks Maksim!

src/plugins/intel_cpu/src/nodes/gather.cpp
xuchen-intel: Applied. Thanks Maksim!

src/plugins/intel_cpu/src/nodes/gather.cpp
```diff
@@ -41,8 +42,76 @@ bool Gather::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, std
     return true;
 }
 
+class GatherShapeInfer : public ShapeInferEmptyPads {
```
maxnick, line 45: could you please put all the shape inference related code to an anonymous namespace, so that it is not visible outside this translation unit?

xuchen-intel: Applied. Thanks Maksim!

maxnick (approved):

xuchen-intel:
> @xuchen-intel , please make CI green.

Hi @maxnick, CI is green now.

PR #16169, opened 2023-03-09
[GPU] Transformed IR serialization for dynamic models

e-ddykim:
### Details:
 - In the case of dynamic models, transformed IR will be serialized.

### Tickets:
 - 105212


src/common/transformations/include/transformations/op_conversions/convert_shapeof3.hpp
```diff
@@ -22,3 +23,9 @@ class ov::pass::ConvertShapeOf3 : public ov::pass::MatcherPass {
     OPENVINO_RTTI("ConvertShapeOf3", "0");
     ConvertShapeOf3();
 };
+
```
yeonbok, line 26: I think we can add this as gpu specific transform instead of common transform?

e-ddykim: I moved it from common to GPU plugin as you reviewed. Thank you.

ahnyoung-paul (approved):

andrew-k-park (approved):

PR #16170, opened 2023-03-09
[GPU] Shape agnostic concat kernel + refactoring

vladimir-paramuzov:
### Details:
 - Shape agnostic concat ref kernel impl
 - Moved kernel_impl_params method's impl to correct cpp file
 - Added time logging for ngraph transformations
 - Moved kernels compilation + kernels init + cleanup to separate pass

### Tickets:
 - *95104*


src/plugins/intel_gpu/src/graph/impls/ocl/concatenation.cpp
```diff
@@ -72,11 +72,38 @@ struct concatenation_impl : typed_primitive_impl_ocl<concatenation> {
 
         return {params, optional_params};
     }
+
+    void update_dispatch_data(const kernel_impl_params& impl_param) override {
+        auto kernel_params = get_kernel_params(impl_param);
```
yeonbok, line 77: auto kernel_params = get_kernel_params(impl_param, true);

vladimir-paramuzov: Looks like that's not enough to enable empty tensor support for shape agnostic multi-stage kernels. 

For example for 2 stages in concat kernel with empty first input (case from unit tests), `get_kernels_source()` returned 2 kernels as the impl shape agnostic and `skip_execution` was equal to `false` for both kernels. Then `update_kernels_list_to_skip` marked first kernel to be skipped due to gws=0, thus `set_arguments()` and `execute()` tried to skip it, but since index of the kernel taken from `_kernels` was independently incremented after actual kernel execution, second stage actually ran first kernel. 

So I removed `skip_execution` check in `get_kernels_source()` and made `set_arguments()/execute()` method use same idx for `_kernels` and `_kernel_data.kernels`

yeonbok: Oh I got it, having difference indexing for kernel and kernel data is problematic for "multi staged shape agnostic kernels.." I see..

yeonbok (approved):

src/plugins/intel_gpu/src/kernel_selector/kernels/concatenation/concatenation_kernel_base.cpp
```diff
@@ -47,12 +47,38 @@ bool ConcatenationKernelBase::Validate(const Params& p, const optional_params&)
 }
 
 JitConstants ConcatenationKernelBase::GetJitConstants(const concatenation_params& params) const {
-    JitConstants jit = MakeBaseParamsJitConstants(params);
+    auto& inputs = params.original_input_layouts;
+    bool is_dynamic = std::any_of(inputs.begin(), inputs.end(), [](const DataTensor& t) { return t.is_dynamic(); }) ||
+                      std::any_of(params.outputs.begin(), params.outputs.end(), [](const DataTensor& t) { return t.is_dynamic(); });
```
yeonbok, line 52: Can't we use params.is_shape_agnostic?

src/plugins/intel_gpu/src/kernel_selector/kernels/concatenation/concatenation_kernel_base.cpp
```diff
@@ -47,12 +47,38 @@ bool ConcatenationKernelBase::Validate(const Params& p, const optional_params&)
...
+        for (size_t i = 0; i < params.original_input_layouts.size(); i++) {
+            if (params.original_input_layouts[i].is_dynamic() || params.original_input_layouts[i].LogicalSize() == 0)
+                out_offset++;
+        }
+
+        jit.AddConstant(MakeJitConstant("INPUT0", params.inputs[0], in_offset));
+        jit.AddConstant(MakeJitConstant("OUTPUT", params.outputs[0], out_offset));
+
+        jit.AddConstant(MakeJitConstant("IS_DYNAMIC", 1));
+        jit.AddConstant(MakeJitConstant("OPTIONAL_SHAPE_INFO_ARG", "__global const int* shape_info,"));
```
yeonbok, line 78: Isnt this generated by common kernel_base.cpp?

vladimir-paramuzov: Common tensor definitions are not generated for that case:
```cpp
JitConstants jit = MakeBaseParamsJitConstants(params, !is_dynamic); // !is_dynamic disables it
```

src/plugins/intel_gpu/src/kernel_selector/kernels/concatenation/concatenation_kernel_base.cpp
vladimir-paramuzov: Probably we can. Will refactor it in separate PR.

PR #16178, opened 2023-03-09
Introduce IRemoteContext

ilyachur:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


src/inference/src/dev/itensor.cpp
```diff
@@ -0,0 +1,490 @@
...
+            return params;
+        } else {
+            return {};
+        }
+    }
+
+    std::shared_ptr<ie::Blob> blob;
+};
+
+class TensorRemoteBlob : public ie::RemoteBlob {
```
riverlijunjie, line 324:
```suggestion
class RemoteBlobTensor : public ie::RemoteBlob {
```

src/inference/src/dev/itensor.cpp
```diff
@@ -0,0 +1,490 @@
...
+public:
+    TensorRemoteBlob(const std::shared_ptr<ITensor>& tensor_)
```
riverlijunjie, line 326: Is it an empty/fake blob to compatible with legacy RemoteBlob? We can see many of its method return null.

src/inference/src/dev/itensor.cpp
```diff
@@ -0,0 +1,490 @@
...
+                       return ie::TensorDesc{ie::details::convertPrecision(element_type),
+                                             shape,
+                                             ie::BlockingDesc{shape, blk_order, 0, dim_offset, blk_strides}};
+                   }(),
+                   static_cast<T*>(tensor_->data()),
+                   tensor_->get_byte_size()},
+            tensor{tensor_} {}
+    catch (const std::exception& ex) {
+        throw ov::Exception(ex.what());
+    } catch (...) {
```
riverlijunjie, line 413: It will be better if we don't use `catch(...)` clause.

src/inference/src/dev/itensor.cpp
```diff
@@ -0,0 +1,490 @@
...
+
+AnyMap ITensor::get_properties() const {
+    return {};
+}
+
+Coordinate ITensor::get_offsets() const {
+    return {get_shape().size(), 0};
+}
+
+class ViewTensor : public ITensor {
```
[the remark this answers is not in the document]
ilyachur: @riverlijunjie hi, for all ITensor things please comment PR: https://github.com/openvinotoolkit/openvino/pull/16048
It will be merged first :)

src/inference/src/dev/itensor.cpp
ilyachur: It is a specific of InferenceEngine Remte blob

PR #16182, opened 2023-03-09
[GPU] Use 4dim directly for onednn in gemm

hyunback:
We are collaping n-dim into 3d for onednn gemm, But it is not necessary.
Until 4d run jit:gemm and 5/6d run ocl:ref kernel, it will be fixed by 3.2
So only 4d directly use gemm without collaping 3d. And 5/6d collaping 4d.

### Tickets:
 - *105534*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/onednn/gemm_onednn.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/onednn/gemm_onednn.cpp
@@ -81,26 +81,15 @@ protected:
         const auto& in0_l = in_layouts[0];
         const auto& in1_l = in_layouts[1];
 
-        size_t in0_batched_size = in0_l.count() / (in0_l.spatial(0) * in0_l.spatial(1));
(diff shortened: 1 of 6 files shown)


src/plugins/intel_gpu/src/graph/impls/onednn/gemm_onednn.cpp
```diff
@@ -301,16 +301,6 @@ struct gemm_onednn : typed_primitive_onednn_impl<gemm> {
...
-        bool full_tensor_or_per_tensor = true;
-        for (auto prim : arg.get_fused_primitives()) {
-            if (prim.input_layout.is_static() && prim.output_layout.is_static()) {
-                full_tensor_or_per_tensor &=
-                    prim.input_layout.count() == prim.output_layout.count() || prim.input_layout.count() == 1;
-            }
-        }
-        if (!full_tensor_or_per_tensor) {
-            IE_THROW() << "Unimplemented: per channel binary post-operation is not supported for onednn gemm. Refer PR(#15353) message.";
-        }
```
isanghao, line 313: Did you check GemmFusingTest?

isanghao (approved):

PR #16183, opened 2023-03-09
[TF FE] Support EmptyTensorList and TensorListPushBack operations

rkazants:
**Details:** Support EmptyTensorList and TensorListPushBack operations. Add unit-test to check the structure of the generated IR.

**Ticket:** 103871


src/frontends/tensorflow_common/src/op/tensor_list_operations.cpp
```diff
@@ -122,6 +122,38 @@ OutputVector translate_tensor_list_set_item_op(const NodeContext& node) {
...
+    // because we want to cover a case of dynamic rank tensor list
+    // the real shape of the tensor elements will be restored by TensorListStack operations
+    auto new_tensor_shape = make_shared<Constant>(element::i32, Shape{1}, -1);
+    tensor = make_shared<Reshape>(tensor, new_tensor_shape, false);
+    auto tensor_shape = make_shared<ShapeOf>(tensor, element::i32);
+
+    // reshape the tensor list to the shape [num_elements, -1]
+    // that is because in the first iteration we have empty constant of a shape [0,0]
+    Output<Node> num_elements = make_shared<ShapeOf>(input_handle, element::i32);
+    auto zero_const = make_shared<Constant>(element::i32, Shape{1}, 0);
```
gkrivor, line 141: I remember we have such constants (zero_const, one_const, two_const, minus_one_const?) in another translators. I saw it in convolution, for example. I didn't check other places.

If it is a common constant I suggest to move it to common constant.

src/frontends/tensorflow/tests/test_models/gen_scripts/empty_tensor_list.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2018-2023 Intel Corporation
```
gkrivor, line 1: Just a question: is it some kind of model generator which you suggested to use for Saved Model tests?

src/frontends/tensorflow_common/src/op/tensor_list_operations.cpp
```diff
@@ -122,6 +122,38 @@ OutputVector translate_tensor_list_set_item_op(const NodeContext& node) {
...
+    // unsqueeze tensor to be inserted into the list
+    tensor = make_shared<Unsqueeze>(tensor, zero_const);
+
+    // insert the tensor into the end
+    auto updated_list = make_shared<Concat>(OutputVector{input_handle, tensor}, 0);
+
+    set_node_name(node.get_name(), updated_list);
+    return {updated_list};
+}
+
```
gkrivor, line 156: Is this empty line under the same rule as you mentioned in my PR?
https://github.com/openvinotoolkit/openvino/pull/15762/files/bc2c01480670b6aca098b388d097774a7c392f76#diff-49c58017be48c28cb7c7825bdf5cdda5df0bb01db66b4f7d729ec8e1c7cfac7b

src/frontends/tensorflow/tests/test_models/gen_scripts/empty_tensor_list.py
rkazants: yes, correct. It generates test model.

rkazants: @gkrivor, absolutely, you can use the same approach for SM format.

PR #16185, opened 2023-03-09
Fix documentation (md and inline) for C++ and Python spech samples

dorloff:
### Details:
 - Update and correct documentation for C++ and Python speech samples

### Tickets:
 - 94314

### DIFF
--- a/samples/cpp/speech_sample/README.md
+++ b/samples/cpp/speech_sample/README.md
@@ -56,9 +56,9 @@ network.
 
 Several execution modes are supported via the `-d` flag:
 
-- `CPU` - All calculation are performed on CPU device using CPU Plugin.
-- `GPU` - All calculation are performed on GPU device using GPU Plugin.
-- `VPUX` - All calculation are performed on VPUX device using VPUX Plugin.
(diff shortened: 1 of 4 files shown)


nkogteva (approved):

mdeisher (commented):
I have not found that the port number is required with the -i option.
For -o and -r, do we support npz too?
HPP, Line 17 should be "Path(s) to input file(s)."
Line 94:  do we really want to keep "dynamic" as an option since it is not implemented.
Line 97 should be "Number of weight bits" or "Weight resolution in bits"

dorloff:
> I have not found that the port number is required with the -i option.

Do you mean you think it's not required? I double checked, it is required.

> For -o and -r, do we support npz too?

Yes.

> HPP, Line 17 should be "Path(s) to input file(s)."

Fixed, thanks.

> Line 94: do we really want to keep "dynamic" as an option since it is not implemented.

Indeed, it makes no sense to keep it, removed.

> Line 97 should be "Number of weight bits" or "Weight resolution in bits"

FIxed, thanks.

PR #16186, opened 2023-03-09
[CPU]Fix crash issue: RuntimeError: Primitive descriptor was not found for…

xipingyan:
### Details:
 - *Choose first PrimitiveDescriptor if it can't match PrimitiveDescriptor*

### Tickets:
 - *104143*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/matmul.cpp
+++ b/src/plugins/intel_cpu/src/nodes/matmul.cpp
@@ -624,7 +624,7 @@ void MatMul::prepareParams() {
         primitive_desc_iterator itpd = desc.createPrimitiveDescriptorIterator(engine, key.attr);
         matmul::primitive_desc prim_desc;
 
-        auto itpdFirst = itpd;
+        auto itpd_first = itpd;
         while (static_cast<bool>(itpd))  {
             impl_desc_type impl_type = parse_impl_name(itpd.impl_info_str());
 
@@ -633,7 +633,10 @@ void MatMul::prepareParams() {
                 break;
             }
             if (!itpd.next_impl()) {
-                prim_desc = itpdFirst.get();
+                // implType(AMX) will be not in list when input dynamic shape for AMX.
+                // Bcause AMX will be available when it assumes batch 16 for dynamic shape in compile model.
+                // However if we only pass batch 1 in runtime, AMX will be not avaiable(Performance is low).
+                prim_desc = itpd_first.get();
                 break;
             }
         }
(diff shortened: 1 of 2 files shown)


src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -624,15 +624,18 @@ void MatMul::prepareParams() {
         primitive_desc_iterator itpd = desc.createPrimitiveDescriptorIterator(engine, key.attr);
         matmul::primitive_desc prim_desc;
 
+        auto itpdFirst = itpd;
```
usstq, line 627: suggest to change `itpdFirst` into  `itpd_first`

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -624,15 +624,18 @@ void MatMul::prepareParams() {
...
             impl_desc_type impl_type = parse_impl_name(itpd.impl_info_str());
 
             if (impl_type == key.implType) {
                 prim_desc = itpd.get();
                 break;
             }
-            if (!itpd.next_impl())
-                return matmul();
+            if (!itpd.next_impl()) {
+                prim_desc = itpdFirst.get();
```
usstq, line 636: Can we add a comment about why `impl_type` is missing in candidate list?

luo-cheng2021:
Could we add a testcase to cover this?

xipingyan:
> Could we add a testcase to cover this?

Done, Adding new param for matmul single layer test. @luo-cheng2021 @dmitry-gorokhov

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -624,15 +624,21 @@ void MatMul::prepareParams() {
...
             if (impl_type == key.implType) {
                 prim_desc = itpd.get();
                 break;
             }
-            if (!itpd.next_impl())
-                return matmul();
+            if (!itpd.next_impl()) {
+                // implType(AMX) will be not in list when input dynamic shape for AMX.
+                // Bcause AMX will be available when it assumes batch 16 for dynamic shape in compile model.
+                // However if we only pass batch 1 in runtime, AMX will be not available(Performance is low).
```
maxnick, line 638:
```suggestion
                // implType(AMX) may not be in the list when the input shapes are dynamic in the case of AMX.
                // AMX is available on the compile model stage when the dummy shape is used.
                // However, if we pass batch 1 at runtime, AMX will not be available (due to the oneDNN implementation selection algorithm).
```

EgorDuplensky: My version would be 😄:
"In case of dynamic shapes an implementation type chosen as optimal for a primitive_desc with undefined input shapes, is not necessarily available for the primitive_desc with defined shape.
Example: brgemm_avx512_amx is available for a primitive with undefined input shapes but not available for primitive_desc with input batch 1"

xipingyan: @EgorDuplensky gave the root cause. It will be great.

luo-cheng2021 (approved):

usstq (approved):

dmitry-gorokhov (approved):

PR #16204, opened 2023-03-10
Add internal functions for throw exceptions

ilyachur:
### Details:
 - *item1*
 - *...*

### Tickets:
 - 103685


src/frontends/paddle/include/openvino/frontend/paddle/node_context.hpp
```diff
@@ -19,7 +19,7 @@ using NamedInputs = std::map<InPortName, OutputVector>;
 
 /// Keep necessary data for a single node in the original FW graph to facilitate
 /// conversion process in the rules code.
-class NodeContext : public ov::frontend::NodeContext {
+class PADDLE_API NodeContext : public ov::frontend::NodeContext {
```
rkazants, line 22: Do we really need to expose this externally? Ideally, think - no

rkazants, line 22:
```suggestion
class NodeContext : public ov::frontend::NodeContext {
```

ilya-lavrenov: should we move it to `src` folder then?

ilyachur: Yes in this case we need to remove this API from public API

src/core/include/openvino/core/node.hpp
```diff
@@ -541,25 +541,13 @@ class OPENVINO_API NodeValidationFailure : public ov::AssertFailure {
...
-                          node->input_values().size(),
-                          " argument",
-                          (node->input_values().size() == 1 ? "" : "s"),
-                          " but got ",
-                          new_args.size());
-}
+OPENVINO_API void check_new_args_count(const Node* node, OutputVector new_args);
 
-}  // namespace ov
+[[noreturn]] OPENVINO_API void throw_on_node_validation_failure(const ov::CheckLocInfo& check_loc_info,
```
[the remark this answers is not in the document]
ilyachur: I got the difference ~ 0.4 Mb on M1 Mac.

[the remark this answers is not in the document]
ilya-lavrenov: M1 is not first priority platform for OV and we optimize primarily for Windows, then Linux

What I want to understand: whether noreturn attribute adds extra binary size with compare to usual function

rkazants (approved):
Looks good from FE perspective.

PR #16216, opened 2023-03-10
split evaluate_map.cpp to small files

zhaixuejun1993:
### Details:
 - split evaluate_map.cpp

### Tickets:
 - 103527


zhaixuejun1993:
This is for CI help bug fix. Can close now.

src/plugins/template/backend/ops/if.hpp
```diff
@@ -0,0 +1,169 @@
+// Copyright (C) 2018-2023 Intel Corporation
```
ilya-lavrenov, line 1: we need to split cpp files into multiple per-operation cpp file. 
splitting up in header files will not help, because compilation time will not decrease

zhaixuejun1993: > we need to split cpp files into multiple per-operation cpp file. splitting up in header files will not help, because compilation time will not decrease

Yes, I know, the next step will split the implementation from hpp to cpp.

zhaixuejun1993:
The CI `openvino-lin` failed, but can not reproduce in local. Anybody can help? @ilya-lavrenov @ilyachur 
image

ilya-lavrenov:
Try to compile with clang and static openvino build.
```bash
CC=clang CXX=clang++ cmake -DBUILD_SHARED_LIBS=OFF 
```

zhaixuejun1993:
> ```shell
> -DBUILD_SHARED_LIBS=OFF
> ```

Thanks Ilya, reproduced with static library

zhaixuejun1993:
> Try to compile with clang and static openvino build.
> 
> ```shell
> CC=clang CXX=clang++ cmake -DBUILD_SHARED_LIBS=OFF 
> ```

Thanks! Reproduced with static library

zhaixuejun1993:
**Using the following script double check to make sure all OPs are exist after split**
```
#!/bin/bash

# step1: get all evaluate func (note: 'evaluates_map.cpp' from original master)
grep "bool evaluate(const" ./evaluates_map.cpp > step_1.txt

# step2: get the full name of OP
sed 's/^.*<//' step_1.txt > step_2.txt
sed 's/>.*//' step_2.txt  > step_3.txt

# Check the OP one by one (note: 'ops_evaluates.hpp' from split repo, which extern all evaluate methods of OPs)
cat step_3.txt | while read line
do
        grepOutput="$(grep $line /home/xuejun-sv/xuejun/CVS-103527/original/openvino/src/plugins/template/backend/ops/ops_evaluates.hpp)"
        if [ ! -z "$grepOutput" ]; then
                echo -e "\n [OP] $line ---------------> SUCCESS"
                echo ${grepOutput}
        else
                echo -e "\n [OP] $line ---------------> FAILED"
        fi
done
```

ilya-lavrenov (approved):

ilyachur (approved):

PR #16223, opened 2023-03-10
[GPU] softmax_gpu_items_class_optimized kernel optimizations

ytorzuk-altran:
### Details:
 - *intel_sub_group_read/intel_sub_group_read are incorporated*

### Tickets:
 - *[CVS-91171](https://jira.devtools.intel.com/browse/CVS-91171)*

### DIFF
--- a/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/softmax_gpu_items_class_optimized.cl
+++ b/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/softmax_gpu_items_class_optimized.cl
@@ -3,11 +3,36 @@
 //
 
 #include "include/batch_headers/common.cl"
+#include "include/batch_headers/fetch_data.cl"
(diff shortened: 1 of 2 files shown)


wilson-seok (approved):

akladiev:

yeonbok:
@wilson-seok  Is this PR verified with target network & benchmark w.r.t the perf as well as accuracy? 
@ytorzuk-altran Which unittest is used for this kernel? Do we have any test with unaligned data and triggers this kernel?

wilson-seok:
> @wilson-seok Is this PR verified with target network & benchmark w.r.t the perf as well as accuracy? @ytorzuk-altran Which unittest is used for this kernel? Do we have any test with unaligned data and triggers this kernel?

This kernel optimization doesn't have specific target model. Instead it has general performance improvement. I verified performance in internal model zoo. I will check accuracy also.

PR #16226, opened 2023-03-11
[GPU] Fix concat to use ngraph shape infer and crop to return original rank 

yeonbok:
### Details:
 - Fix concat to use ngraph shape infer
 - Fix crop to return shape of original rank


ahnyoung-paul (approved):

src/plugins/intel_gpu/include/intel_gpu/runtime/tensor.hpp
```diff
@@ -652,6 +654,19 @@ struct tensor {
         return offset;
     }
 
+    /// @brieef Returns tensor as Partial shape of requested rank
```
andrew-k-park, line 657: nit: @brief

andrew-k-park (approved):
Overall looks good to me except for minor comment

PR #16229, opened 2023-03-11
fix core dump issue for ENABLE_DEBUG_CAPS

xufang-lisa:
### Details:
 - *when openvino is compiled with option -DENABLE_DEBUG_CAPS=ON, run ./ov_cpu_func_tests  in a docker contain with a memory limit of 8G, then a crash error will appear for test case "smoke_MemoryTest/MemoryTest.CompareWithRefs/transformation=LOW_LATENCY_V2_iteration_count=3_IS=(3)_netPRC=FP32_trgDev=HETERO:CPU)"*


dmitry-gorokhov:
@EgorDuplensky please take a look

EgorDuplensky (approved):

PR #16243, opened 2023-03-13
[PT FE]: improve integration into mo.convert_model

eaidova:
### Details:
 - *productization pytorch frontend in mo.convert_model*

### Tickets:
 - *CVS-103615*

### DIFF
--- a/tests/layer_tests/pytorch_tests/test_add.py
+++ b/tests/layer_tests/pytorch_tests/test_add.py
@@ -8,40 +8,40 @@ import torch
 from pytorch_layer_test_class import PytorchLayerTest
 
 
-# @pytest.mark.parametrize('alpha', (-0.5, 0, 0.5, 1, 2))
-# @pytest.mark.parametrize('input_rhs', (np.random.randn(2, 5, 3, 4).astype(np.float32),
-#                                        np.random.randn(
-#                                            1, 5, 3, 4).astype(np.float32),
-#                                        np.random.randn(1).astype(np.float32)))
-# class TestAdd(PytorchLayerTest):
+@pytest.mark.parametrize('alpha', (-0.5, 0, 0.5, 1, 2))
+@pytest.mark.parametrize('input_rhs', (np.random.randn(2, 5, 3, 4).astype(np.float32),
+                                       np.random.randn(
+                                           1, 5, 3, 4).astype(np.float32),
+                                       np.random.randn(1).astype(np.float32)))
+class TestAdd(PytorchLayerTest):
 
-#     def _prepare_input(self):
-#         return (np.random.randn(2, 5, 3, 4).astype(np.float32), self.input_rhs)
+    def _prepare_input(self):
+        return (np.random.randn(2, 5, 3, 4).astype(np.float32), self.input_rhs)
 
-#     def create_model(self, alpha, op_type):
-#         class aten_add(torch.nn.Module):
+    def create_model(self, alpha, op_type):
+        class aten_add(torch.nn.Module):
 
-#             def __init__(self, alpha, op) -> None:
-#                 super().__init__()
-#                 self.alpha = alpha
-#                 self.forward = self.forward1 if op == "add" else self.forward2
+            def __init__(self, alpha, op) -> None:
+                super().__init__()
+                self.alpha = alpha
+                self.forward = self.forward1 if op == "add" else self.forward2
 
-#             def forward1(self, lhs, rhs):
-#                 return torch.add(lhs, rhs, alpha=self.alpha)
+            def forward1(self, lhs, rhs):
+                return torch.add(lhs, rhs, alpha=self.alpha)
 
-#             def forward2(self, lhs, rhs):
-#                 return lhs.add_(rhs, alpha=self.alpha)
+            def forward2(self, lhs, rhs):
+                return lhs.add_(rhs, alpha=self.alpha)
 
-#         ref_net = None
+        ref_net = None
 
-#         return aten_add(alpha, op_type), ref_net, f"aten::{op_type}"
+        return aten_add(alpha, op_type), ref_net, f"aten::{op_type}"
 
-#     @pytest.mark.nightly
-#     @pytest.mark.precommit
-#     @pytest.mark.parametrize("op_type", ["add", "add_"])
-#     def test_add(self, ie_device, precision, ir_version, alpha, input_rhs, op_type):
-#         self.input_rhs = input_rhs
-#         self._test(*self.create_model(alpha, op_type), ie_device, precision, ir_version)
+    @pytest.mark.nightly
+    @pytest.mark.precommit
+    @pytest.mark.parametrize("op_type", ["add", "add_"])
+    def test_add(self, ie_device, precision, ir_version, alpha, input_rhs, op_type):
+        self.input_rhs = input_rhs
+        self._test(*self.create_model(alpha, op_type), ie_device, precision, ir_version)
 
 
 class TestAddTypes(PytorchLayerTest):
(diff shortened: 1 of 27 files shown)


tools/mo/openvino/tools/mo/moc_frontend/extractor.py
```diff
@@ -132,9 +132,10 @@ def get_port(match, match_starts_with_name, input_model, framework):
     if len(found_places) == 0:
         raise_no_node(node_name)
 
-    # Check that there is no collision, all found places shall point to same data
-    if not all([n.is_equal_data(found_places[0]) for n in found_places]):
-        raise_node_name_collision(node_name, found_place_names)
+    if framework != "pytorch":
+        # Check that there is no collision, all found places shall point to same data
+        if not all([n.is_equal_data(found_places[0]) for n in found_places]):
+            raise_node_name_collision(node_name, found_place_names)
```
rkazants, line 138: why is it so for PT - one tensor name for several tensor places?

tools/mo/openvino/tools/mo/moc_frontend/extractor.py
```diff
@@ -187,11 +188,14 @@ def fe_input_user_data_repack(
     """
     _input_shapes = []
     _input_names = []
+    model_inputs = input_model.get_inputs()
+    if framework == "pytorch":
+        model_inputs = model_inputs[1:]
```
rkazants, line 193: several questions:
1. what is the first input?
2. why can we exclude this input in PT FE internally if it is not used at all? So MO code should not care of it.

tests/layer_tests/pytorch_tests/test_add.py
```diff
@@ -8,40 +8,40 @@
...
-                                       np.random.randn(
-                                           1, 5, 3, 4).astype(np.float32),
-                                       np.random.randn(1).astype(np.float32)))
-class TestAdd(PytorchLayerTest):
+# @pytest.mark.parametrize('alpha', (-0.5, 0, 0.5, 1, 2))
+# @pytest.mark.parametrize('input_rhs', (np.random.randn(2, 5, 3, 4).astype(np.float32),
+#                                        np.random.randn(
+#                                            1, 5, 3, 4).astype(np.float32),
+#                                        np.random.randn(1).astype(np.float32)))
+# class TestAdd(PytorchLayerTest):
```
rkazants, line 16: please return tests:)

tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
```diff
@@ -50,8 +52,96 @@ def get_pytorch_decoder(model, input_shape, example_inputs):
         scripted = model
     f_model = torch.jit.freeze(scripted)
     decoder = TorchScriptPythonDecoder(f_model)
+    input_signature = align_input_parameters(f_model, input_signature, inputs, input_shape, args)
+        
     return decoder, input_signature
 
+def align_input_parameters(f_model, input_signature, inputs, input_shape, args):
```
rkazants, line 59: can you please provide description for this function?

tools/mo/openvino/tools/mo/__init__.py
```diff
@@ -1,4 +1,5 @@
 # Copyright (C) 2018-2023 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-from .convert import convert_model, InputCutInfo, LayoutMap
+from .convert import convert_model
+from openvino.tools.mo.utils.arg_wrappers import InputCutInfo, LayoutMap
```
rkazants, line 5: please do not change it

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
```diff
@@ -215,7 +218,6 @@ def shape_to_array(shape: PartialShape):
 
     ngraph_function = moc_front_end.convert(input_model)
 
-    # TO DO: remove as part of PyTorch frontend productization CVS-103615
     if argv.framework == "pytorch":
-        pytorch_process_after_convert(argv, ngraph_function)
+        resolve_input_signature(argv, ngraph_function)
```
rkazants, line 222: not sure why signature (input order, etc.) resolving can not be handled under the hood of PT FE? Looks not correct.

tools/mo/openvino/tools/mo/moc_frontend/extractor.py
eaidova: the first input is always self, it does not have shape and type and invisible for users, it is removed from model after conversion.

tools/mo/openvino/tools/mo/__init__.py
eaidova: if I do not move them , I have round dependencies, I want use these parameters wrappers for provide additional parameters hints based on example input.

eaidova: Also my opinion, it is not good place for storing data types

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
eaidova: this info does not provided inside model and from discussion with @mvafin , I know that it is impossible to add tensor names to InputModel

tools/mo/openvino/tools/mo/moc_frontend/extractor.py
slyalin: We should filter inputs on PT FE side. There are two possible solutions: short term (1) apply the same hack on PT FE side (just remove the first input), long term (2) restore model topology and report back the right number of inputs. To avoid blocking at this moment we can apply (1) and plan (2) as a separate task.

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
slyalin: Looks like we can (and we should) move Pytorch specific code, like this one, to PT FE with extending Decoder API. Decoder is what connects original model with the conversion core, that means that we can provide any auxiliary information for model conversion pre- and post-processing via decoder. The problem which is specific for PT FE is that this pre- and post-processing should happens in Python. This is not an issue for us, because we can extend Decoder with pre- and post-processing methods with appropriate arguments and call them from fe.convert method implemented in C++. Actually it is also related to tracing/scripting model -- it also can be a part of the FE, not general MO code. @eaidova , do you feel confident at changing API in this way or you prefer someone else to sketch it for you?

slyalin (changes requested):
Move PT specific code to PT FE via extended Decoder. We can discuss how to quickly do that or I can provide a draft if you wish.

tools/mo/openvino/tools/mo/moc_frontend/extractor.py
rkazants: Exactly, now the solution from PT FE is incomplete.

rkazants (changes requested):
Still a lot of PT FE specific code

tools/mo/openvino/tools/mo/moc_frontend/extractor.py
eaidova: moved to frontend side

tools/mo/openvino/tools/mo/__init__.py
eaidova: moved back

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
eaidova: @slyalin, scripting moved to decoder, input signature adding if applicable too, please review

tests/layer_tests/pytorch_tests/test_add.py
eaidova: return

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
rkazants: can we hide resolving of the input signature in `moc_front_end.convert()` method? Can we provide specific implementation of Python part of `convert` method for PT FE? I think input signature resolving should be a part of conversion.

eaidova: no, we need to know it before conversion, because user can provide named input information via inputs which will be failed if input names from signature will be not available from input_model

rkazants (approved):
one comment left. Other looks good to me.

tools/mo/openvino/tools/mo/moc_frontend/shape_utils.py
```diff
@@ -99,4 +103,26 @@ def parse_input_shapes(argv):
...
+    if isinstance(input_info, tuple) and len(input_info) == 4 and isinstance(input_info[0], str):
+        if input_info.shape is not None:
+            return [parse_single_shape(input_info.shape, False)]
+    if isinstance(input_info, (list, tuple)):
+        input_shapes = []
+        for inp in input_info:
+            if inp.shape is not None:
+                input_shapes.append(parse_single_shape(inp.shape, False))
+        return input_shapes or None
+    return None
```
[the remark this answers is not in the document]
eaidova: these changes need for allowing prepare example inputs from input parameter, now only input_shapes are accepted, if @popovaan agrees to cover it later or as part of her PR, I can remove these changes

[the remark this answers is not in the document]
eaidova: removed

eaidova:
@slyalin do you have any additional comments?

slyalin (approved):

PR #16249, opened 2023-03-13
[CPU][Snippets]fix candidate merged node's subgraph inputs have common subgraph input

chenhu-wang:
### Details:
 - *fix unite two body parameters from two input subgraphs when the two input subgraphs are connected to the same parent node/subgraph, but does not respect which output port the two edges link*

### Tickets:
 - *99399*

### DIFF
--- a/src/common/snippets/src/pass/collapse_subgraph.cpp
+++ b/src/common/snippets/src/pass/collapse_subgraph.cpp
@@ -389,8 +389,11 @@ TokenizeSnippets::TokenizeSnippets() {
                             size_t current_input_index = body_parameters.size();
                             for (size_t p_ind = 0; p_ind <  body_parameters.size(); p_ind++) {
                                 const auto & p = body_parameters[p_ind];
+                                // unite two body parameters from two input subgraphs only if:
+                                // 1. two input subgraphs are connected to the same parent node/subgraph,
+                                // 2. and connected to the same output port of this parent node/subgraph.
                                 if (p->get_friendly_name() == found->get_node_shared_ptr()->get_friendly_name() &&
-                                    external_inputs[p_ind] == *found) {  // or replace only when common parent subgraph have one output
+                                    external_inputs[p_ind] == *found) {
                                     current_input_index = p_ind;
                                     break;
                                 }
--- a/src/common/snippets/src/pass/collapse_subgraph.cpp
+++ b/src/common/snippets/src/pass/collapse_subgraph.cpp
@@ -389,7 +389,8 @@ TokenizeSnippets::TokenizeSnippets() {
                             size_t current_input_index = body_parameters.size();
                             for (size_t p_ind = 0; p_ind <  body_parameters.size(); p_ind++) {
                                 const auto & p = body_parameters[p_ind];
-                                if (p->get_friendly_name() == found->get_node_shared_ptr()->get_friendly_name()) {
+                                if (p->get_friendly_name() == found->get_node_shared_ptr()->get_friendly_name() &&
+                                    external_inputs[p_ind] == *found) {  // or replace only when common parent subgraph have one output
                                     current_input_index = p_ind;
                                     break;
                                 }
--- a/src/tests/functional/plugin/shared/include/snippets/edge_replace.hpp
+++ b/src/tests/functional/plugin/shared/include/snippets/edge_replace.hpp
@@ -11,14 +11,7 @@ namespace test {
 namespace snippets {
 
 typedef std::tuple<
-        InputShape,                  // Input 0 Shape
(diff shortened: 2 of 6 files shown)


src/tests/functional/plugin/shared/include/snippets/edge_replace.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+typedef std::tuple<
+        InputShape,                  // Input 0 Shape
+        InputShape,                  // Input 1 Shape
+        InputShape,                  // Input 2 Shape
+        InputShape,                  // Input 3 Shape
+        InputShape,                  // Input 4 Shape
+        InputShape,                  // Input 5 Shape
+        InputShape,                  // Input 6 Shape
+        InputShape,                  // Input 7 Shape
```
a-sidorova, line 21: Do we really need the separate 8 shapes? :smile: Can we just use `std::vector`?

Moreover, you can use `ov::PartialShape` insetad of `InputShape` to use `init_input_shapes(static_partial_shapes_to_test_representation(input_shapes));` in `SetUp()`. What do you think?

src/common/snippets/src/pass/collapse_subgraph.cpp
```diff
@@ -376,7 +376,8 @@ TokenizeSnippets::TokenizeSnippets() {
                             size_t current_input_index = body_parameters.size();
                             for (size_t p_ind = 0; p_ind <  body_parameters.size(); p_ind++) {
                                 const auto & p = body_parameters[p_ind];
-                                if (p->get_friendly_name() == found->get_node_shared_ptr()->get_friendly_name()) {
+                                if (p->get_friendly_name() == found->get_node_shared_ptr()->get_friendly_name() &&
+                                    external_inputs[p_ind] == *found) {  // or replace only when common parent subgraph have one output
```
a-sidorova, line 380: It'd be great if you add comment that we check for the same output ports of parent.
I mean that we can unite body parameters only if they're connected to the same parent of Subgraph (it's already been) and to the same output port of this parent (it's added in the PR). What do you think? It's just to remember what happens here.

Thanks!

src/tests/ngraph_helpers/snippets_ngraph_functions/include/subgraph_simple.hpp
```diff
@@ -227,6 +227,21 @@ class BroadcastSelectFunction : public SelectFunction {
...
+/// yolo-v3 subgraph to cover edge replacement.
+//                subgraph
+//                /     |
+//           subgraph  subgraph
+//                \     /
+//              eltwise node
+class EdgeReplaceFunction : public SnippetsFunctionBase {
+public:
+    explicit EdgeReplaceFunction(const std::vector<PartialShape>& inputShapes) : SnippetsFunctionBase(inputShapes) {
+        NGRAPH_CHECK(input_shapes.size() == 8, "Got invalid number of input shapes");
```
IvanNovoselov, line 240: Please note that we don't have to reproduce the entire subgraph from the real model here. It would be really help to reduce maintenance costs if you could simplify the subgraph as much as possible, but in a way that it still covers the introduced changes. 
For example, you can use minimalistic one-operation subgraphs and replace unnecessary Parameter inputs with scalar Constants. This way we won't need 8 inputs.

src/common/snippets/src/pass/collapse_subgraph.cpp
IvanNovoselov: Agree, a comment is definitely needed here.

chenhu-wang: comments is added! thanks.

src/tests/ngraph_helpers/snippets_ngraph_functions/include/subgraph_simple.hpp
chenhu-wang: Simplified.
common parent could be node that have multiple output, now use a split node instead of a big subgraph to represent it.
one block is enough to reproduce, removed two parallel blocks.
reduced input numbers with constant internally, as this issue is not related input values.
Thank you!

src/tests/functional/plugin/shared/include/snippets/edge_replace.hpp
chenhu-wang: yeap, the number of input is reduced with a simplified reproduceble use case, and partial shape is used now. Thank you!

a-sidorova (approved):

a-sidorova:
@dmitry-gorokhov may I ask you to take a look please? Thanks!

dmitry-gorokhov (approved):

PR #16253, opened 2023-03-13
Fix unet3d mlperf dump

hyunback:
convolutionbackpropdata:conv3d_transpose in unet3d_mlperf doesn't have bias(),

### Tickets:
 - *106024*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/deconvolution.cpp
+++ b/src/plugins/intel_gpu/src/graph/deconvolution.cpp
@@ -230,8 +230,10 @@ std::string deconvolution_inst::to_string(deconvolution_node const& node) {
 
     ss_weights << node.weights().id();
     ss_weights << ", count: " << node.weights().get_output_layout().count();
-    ss_biases << node.bias().id();
-    ss_biases << ", count: " << node.bias().get_output_layout().count();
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gpu/src/graph/deconvolution.cpp
```diff
@@ -230,8 +230,10 @@ std::string deconvolution_inst::to_string(deconvolution_node const& node) {
 
     ss_weights << node.weights().id();
     ss_weights << ", count: " << node.weights().get_output_layout().count();
-    ss_biases << node.bias().id();
-    ss_biases << ", count: " << node.bias().get_output_layout().count();
+    if (node.get_dependencies().size() == 3) {
```
isanghao, line 233: What about it has another fused layer such as `prelu`? I think we need to check directly whether it has bias or not.
I think correct approach is to check with `bias_term()` function..

isanghao (changes requested):

src/plugins/intel_gpu/src/graph/deconvolution.cpp
hyunback: "program_node& bias() const { return get_dependency(2);}" in deconvolution_inst.h
Size 3 is always bias in deconv, but for readability I agree

isanghao (approved):

PR #16258, opened 2023-03-13
disable mixed precision inference for quantized IRs

pavel-esir:
### Details:
 - Accidentally quantized models were marked by mixed precision inference transformations, because of that redundant Reorders were inserted around MVNs/NormalizeL2/etc. and this caused significant perf degradation on GPU. Need to disable mixed precision marking for Quantized IRs, on GPU they should run as is, without mixed precision inference.
 
image

### Tickets:
 - xxx-105969, xxx-105064


src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -312,8 +312,15 @@ bool ov::pass::ConvertPrecision::run_on_model(const std::shared_ptr<ngraph::Func
         return false;
 
     bool has_fp16_compression = m_precisions.count(element::f32) > 0 && m_precisions[element::f32] == element::f16;
+    bool has_quantize_nodes = false;
+    for (const auto& op : f->get_ops()) {
+        if (std::dynamic_pointer_cast<opset8::FakeQuantize>(op)) {
```
rkazants, line 317:
```suggestion
        if (as_typ_ptr<opset8::FakeQuantize>(op)) {
```

rkazants (approved):

rkazants:
@KodiaqQ, @AlexKoff88, please review it before merge

AlexKoff88:
cc'ed @vladimir-paramuzov.

src/common/transformations/src/transformations/convert_precision.cpp
pavel-esir: these lines have been removed, but applied change dynamic_pointer_cast -> as_type_ptr in other modified places

pavel-esir:
Closing, since solution here has not general. Implemented better solution with selective turning off marking here https://github.com/openvinotoolkit/openvino/pull/16785

PR #16276, opened 2023-03-14
[PT FE]: enable dtype in softmax and constant filling, extend logical…

eaidova:
… ops support

### Details:
 - *extend aten::softmax with dtype support*
 - *support aten::__or__*
 - *fix input signature for aten::empty*
 - *support aten::baddbmm (add matmul for batched matrix)*

P.S. this list of ops required to support bloom model conversion, tested locally that after these changes it successfully converted
### Tickets:
 - *TBD*

### DIFF
--- a/src/frontends/pytorch/src/op/bitwise_not.cpp
+++ b/src/frontends/pytorch/src/op/bitwise_not.cpp
@@ -1,0 +1,29 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/op/logical_not.hpp"
+#include "utils.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_bitwise_not(NodeContext& context) {
+    num_inputs_check(context, 1, 2);
+    auto x = context.get_input(0);
+    FRONT_END_OP_CONVERSION_CHECK(x.get_element_type().compatible(element::boolean),
+                                  "aten::bitwise_not suppored only for boolean input");
+    auto not_x = context.mark_node(std::make_shared<ov::op::v1::LogicalNot>(x));
+    if (!context.input_is_none(1)) {
+        context.mutate_input(1, not_x);
+    }
+    return {not_x};
+};
+
+}  // namespace op
+}  // namespace pytorch
+}  // namespace frontend
+}  // namespace ov
--- a/src/frontends/pytorch/src/op/softmax.cpp
+++ b/src/frontends/pytorch/src/op/softmax.cpp
@@ -12,11 +12,15 @@ namespace frontend {
 namespace pytorch {
 namespace op {
 
+using namespace ov::op;
 OutputVector translate_softmax(NodeContext& context) {
-    num_inputs_check(context, 2, 2);
+    num_inputs_check(context, 2, 3);
     auto x = context.get_input(0);
     auto axis = context.const_input<int64_t>(1);
-    return {context.mark_node(std::make_shared<ov::op::v8::Softmax>(x, axis))};
+    if (!context.input_is_none(2)) {
+        x = apply_dtype(context, 2, x);
+    }
+    return {context.mark_node(std::make_shared<v8::Softmax>(x, axis))};
 };
 
 }  // namespace op
--- a/src/frontends/pytorch/src/utils.cpp
+++ b/src/frontends/pytorch/src/utils.cpp
@@ -142,6 +142,21 @@ element::Type convert_dtype(int64_t pt_type) {
     return TORCH_TO_OV_TYPE.at(pt_type);
 };
 
+Output<Node> apply_dtype(const NodeContext& context, size_t dtype_port, const Output<Node>& input_tensor) {
+    if (std::dynamic_pointer_cast<opset10::Constant>(
+            context.get_input_from_visible_context(dtype_port).get_node_shared_ptr())) {
+        auto dtype = convert_dtype(context.const_input<int64_t>(dtype_port));
+        return context.mark_node(std::make_shared<opset10::Convert>(input_tensor, dtype));
+    } else if (const auto& fw_node =
+                   cast_fw_node(context.get_input(static_cast<int>(dtype_port)).get_node_shared_ptr(), "prim::dtype")) {
+        auto out_tensor = fw_node->input_value(0);
+        return context.mark_node(std::make_shared<opset10::ConvertLike>(input_tensor, out_tensor));
+    } else {
+        FRONT_END_OP_CONVERSION_CHECK(false, "Couldn't get dtype input");
+    }
+    return input_tensor;
+};
+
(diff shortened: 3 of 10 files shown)


src/frontends/pytorch/src/op/bitwise_not.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/op/logical_not.hpp"
+#include "utils.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+OutputVector translate_bitwise_not(NodeContext& context) {
```
rkazants, line 14:
```suggestion
OutputVector translate_bitwise_not(const NodeContext& context) {
```

rkazants: I remember Maxim should have added this. Please check

src/frontends/pytorch/src/op/bitwise_not.cpp
```diff
@@ -0,0 +1,29 @@
...
+    num_inputs_check(context, 1, 2);
+    auto x = context.get_input(0);
+    FRONT_END_OP_CONVERSION_CHECK(x.get_element_type().is_dynamic() || x.get_element_type() == element::boolean,
+                                  "aten::bitwise_not suppored only for boolean input");
```
rkazants, line 18:
```suggestion
    FRONT_END_OP_CONVERSION_CHECK(x.get_element_type().compatible(element::boolean),
                                  "aten::bitwise_not suppored only for boolean input");
```

src/frontends/pytorch/src/utils.cpp
```diff
@@ -142,6 +142,22 @@ element::Type convert_dtype(int64_t pt_type) {
     return TORCH_TO_OV_TYPE.at(pt_type);
 };
 
+Output<Node> apply_dtype(const NodeContext& context, int dtype_port, Output<Node> input_tensor) {
```
rkazants, line 145:
```suggestion
Output<Node> apply_dtype(const NodeContext& context, int dtype_port, const Output<Node>& input_tensor) {
```

src/frontends/pytorch/src/utils.hpp
```diff
@@ -41,6 +41,9 @@ std::shared_ptr<Node> get_axes_range(const NodeContext& context, int input_id);
 std::shared_ptr<Node> numel(const NodeContext& context, const Output<Node>& x);
 
 element::Type convert_dtype(int64_t dtype_value);
+
+Output<Node> apply_dtype(const NodeContext& context, int dtype_port, Output<Node> input_tensor);
```
rkazants, line 45:
```suggestion
Output<Node> apply_dtype(const NodeContext& context, int dtype_port, const Output<Node>& input_tensor);
```

src/frontends/pytorch/src/op/bitwise_not.cpp
eaidova: it is still not merged due to segfault issue in tests

eaidova: for now, input mutation does not work if I will use const context

src/frontends/pytorch/src/op/bitwise_not.cpp
eaidova: updated, thanks

src/frontends/pytorch/src/utils.cpp
eaidova: updated

src/frontends/pytorch/src/utils.cpp
```diff
@@ -142,6 +142,22 @@ element::Type convert_dtype(int64_t pt_type) {
...
+        auto dtype = convert_dtype(context.const_input<int64_t>(dtype_port));
+        return context.mark_node(std::make_shared<opset10::Convert>(input_tensor, dtype));
+
+    } else if (const auto& fw_node = cast_fw_node(context.get_input(dtype_port).get_node_shared_ptr(), "prim::dtype")) {
+        auto out_tensor = fw_node->input_value(0);
+        return context.mark_node(std::make_shared<opset10::ConvertLike>(input_tensor, out_tensor));
+
+    } else {
+        FRONT_END_OP_CONVERSION_CHECK(false, "Couldn't get dtype input");
+    }
```
rkazants, line 157: why can't we check if `context.get_input(dtype_port).get_element_type().is_static()` and use `Convert` if it is static? And use `ConvertLike` in all other cases?

Can you please see (may be it will be helpful) `ov::Output<ov::Node> create_same_type_const_scalar(const ov::Output<ov::Node>& same_type_output, T value)` approach from TF FE? Can we apply the same?

src/frontends/pytorch/src/utils.cpp
```diff
@@ -142,6 +142,22 @@ element::Type convert_dtype(int64_t pt_type) {
     return TORCH_TO_OV_TYPE.at(pt_type);
 };
 
+Output<Node> apply_dtype(const NodeContext& context, int dtype_port, const Output<Node>& input_tensor) {
+    if (std::dynamic_pointer_cast<opset10::Constant>(
+            context.get_input_from_visible_context(dtype_port).get_node_shared_ptr())) {
+        auto dtype = convert_dtype(context.const_input<int64_t>(dtype_port));
+        return context.mark_node(std::make_shared<opset10::Convert>(input_tensor, dtype));
+
```
rkazants, line 150:
```suggestion
```

src/frontends/pytorch/src/utils.cpp
```diff
@@ -142,6 +142,22 @@ element::Type convert_dtype(int64_t pt_type) {
...
+    } else if (const auto& fw_node = cast_fw_node(context.get_input(dtype_port).get_node_shared_ptr(), "prim::dtype")) {
+        auto out_tensor = fw_node->input_value(0);
+        return context.mark_node(std::make_shared<opset10::ConvertLike>(input_tensor, out_tensor));
+
```
rkazants, line 154:
```suggestion
```

rkazants (approved):
left some comments.
Other looks good to me.

Also, please fix the build failure due to type check

src/frontends/pytorch/src/utils.cpp
eaidova: it works slightly different, dtype can be represented as operation torch.dtype(tensor), which returns dtype of input tensor or can be represented by native torch dtype e.g. torch.float32, which turns to enum constant int and should be mapped to openvino dtype. As we do not have operations which extract dtype from node, we are trying to resolve torch.dtype with replacement convertlike with its input.

However, maybe I do not fully understand your suggestion, maybe it required offline discussion

src/frontends/pytorch/src/op/softmax.cpp
```diff
@@ -5,18 +5,23 @@
 #include "openvino/op/softmax.hpp"
 
 #include "openvino/frontend/pytorch/node_context.hpp"
+#include "pt_framework_node.hpp"
```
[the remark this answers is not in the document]
eaidova: removed, thanks

mvafin (approved):
Please resolve conflict

PR #16286, opened 2023-03-14
[CPU] enable brgconv512 when exists binary post ops. 

luo-cheng2021:
### Details:
 - *enable brgconv512 when exists binary post ops. *
 - *...*

### Tickets:
 - *105755*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/conv.cpp
+++ b/src/plugins/intel_cpu/src/nodes/conv.cpp
@@ -960,17 +960,6 @@ void Convolution::addLegacyZeroPoints(dnnl::primitive_attr& attr) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1650,26 +1639,13 @@ void Convolution::appendZeroPointsArgs() {
 
 // Due to performance issue, brgconv will only be enabled by default:
 // 1, static shape(dynamic shape may change weights layout if the input shape changes and cause performance issue: 86948)
-// 2, support amx except having input zero point.
-// 3, support avx512 without legacy postops/per channel zero point when avx512
+// 2, hw supports avx512+
```
usstq, line 1642: `2,` should be removed?

luo-cheng2021: still have the limitation yet, after avx2 enabled it will change to avx2+.

usstq (approved):

PR #16289, opened 2023-03-14
[GPU] Fix segmentation fault for dynamic model

andrew-k-park:
### Details:
 - Follow up segmentation fault issue from https://github.com/openvinotoolkit/openvino/pull/16158
 - Allow eltwise to raise up through dependency squeeze(ND->0D) in `prepare_primitive_fusing_through` opt pass

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing_through.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing_through.cpp
@@ -42,14 +42,18 @@ void prepare_primitive_fusing_through::run(program& p) {
                 node->get_output_layout().data_type != node->get_dependency(0).get_output_layout().data_type)
                 return false;
 
-            // Not to fuse reshape after Reduce changing the order of un-reduced axes. It is expected to be optimized out.
-            if (node->is_type<reshape>() && node->get_dependencies().front().first->is_type<reduce>())
-                return false;
+            if (node->is_type<reshape>()) {
+                // Not to fuse reshape after Reduce changing the order of un-reduced axes. It is expected to be optimized out.
+                if (node->get_dependencies().front().first->is_type<reduce>())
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing_through.cpp
```diff
@@ -42,14 +42,18 @@ void prepare_primitive_fusing_through::run(program& p) {
...
+                // Not to fuse reshape after Reduce changing the order of un-reduced axes. It is expected to be optimized out.
+                if (node->get_dependencies().front().first->is_type<reduce>())
+                    return false;
+
+                // Not to raise up target node through reshape where the size of dimension is changed (e.g. Squeeze, Unsqueeze)
+                // except for squeezing ND tensor to 0D tensor
+                auto input_pshape = node->get_dependency(0).get_output_layout().get_partial_shape();
+                auto output_pshape = node->get_output_layout().get_partial_shape();
+                if (output_pshape.size() != 0 && input_pshape.size() != output_pshape.size())
+                    return false;
```
yeonbok, line 55: Should we move through reshape if input rank != output_rank && output rank == 0 ?

andrew-k-park: Yes, The purpose is to allow eltwise to raise up through reshape only when squeezing to 0D so that there is no unfused eltwise with 0D output tensor.

yeonbok: What happens if there is Eltwise with 0D output tensor?

andrew-k-park:
Closed this PR because the issue has already been fixed since the following PR (https://github.com/openvinotoolkit/openvino/pull/16319) was merged

PR #16291, opened 2023-03-14
TransposeSinking refactoring: part 2 (class names, folders, file names)

itikhono:
### Details:
All TransposeSinking transformations and tests were moved to transformations/transpose_sinking folder
Introduced ov::pass::transpose_sinking namespace
Transformation classes were renamed: TransposeSinking* -> TS* 
Transformation files were renamed: transpose_sinking_* -> ts_*

  before: ______________________________________________after: _______________________________
image image


evkotov (approved):

src/frontends/tensorflow_lite/src/frontend.cpp
```diff
@@ -268,7 +268,7 @@ void FrontEnd::normalize(const std::shared_ptr<ov::Model>& function) const {
     manager.register_pass<ov::frontend::tensorflow_lite::pass::TFLQuantizeResolver>();
     manager.register_pass<ov::frontend::tensorflow_lite::pass::Rfft2dSimplifier>();
     manager.register_pass<ov::pass::TransposeSinking>();
-    manager.register_pass<ov::pass::TransposeSinkingGeneral>();
+    manager.register_pass<ov::pass::transpose_sinking::TSGeneral>();
```
rkazants, line 271: To me `ov::pass::TransposeSinking` looks better. You can include all particular TS passes into `transpose_sinking` namespace and remain the main entry on `pass` namespace.

rkazants (commented):
Left one comment about naming and using additional namespace only for particular TS passes.

Other looks good to me

src/frontends/tensorflow_lite/src/frontend.cpp
itikhono: unfortunately, we already have ov::pass::TransposeSinking (old common transformation), see line 270
We will be able to replace it in the next Release

rkazants: I see what about `TransposeSinkingExtended` or `TransposeSinkingAdvanced` names?

itikhono: I believe we can leave both options:
ov::pass::TransposeSinkingGeneral (the alias)
ov::pass::transpose_sinking::TSGeneral

rkazants (approved):

src/frontends/tensorflow_lite/src/frontend.cpp
itikhono: it will be fixed in the next PR

PR #16300, opened 2023-03-15
[GPU] Need to exclude fused mem_dep from shape_infer_dep

kelvinchoi-intel:
### Details:
 - *Fused mem_dep doesn't be considered as shape_infer_dep*

### Tickets:
 - *100964*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/include/program_node.h
+++ b/src/plugins/intel_gpu/src/graph/include/program_node.h
@@ -91,6 +91,9 @@ public:
                 if (u->get_dependencies().size() <= dep_idx) {
                     continue;
                 }
+                if (u->is_fused_dep(dep_idx)) {
+                    continue;
+                }
                 if (u->get_dependency(dep_idx).get_unique_id() == unique_id) {
                     return true;
                 }
@@ -98,6 +101,8 @@ public:
         }
         return false;
     }
+
+    bool is_fused_dep(size_t dep_idx) const;
 
     std::map<size_t, memory::ptr> get_const_memory_deps() const;
 
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -190,6 +190,10 @@ void primitive_inst::update_shape() {
         }
         auto& dep = _node->get_dependency(i);
         auto dep_id = dep.id();
+        // exclude fused node from memory_deps
+        if (_node->is_fused_dep(i)) {
+            break;
+        }
         // Events may be not created for in-order queue, so take them for OOO queue only
         if (_network.has_event(dep.id()) && queue_type == QueueTypes::out_of_order) {
             dependencies_events.push_back(_network.get_primitive_event(dep_id));
--- a/src/plugins/intel_gpu/src/graph/program_node.cpp
+++ b/src/plugins/intel_gpu/src/graph/program_node.cpp
@@ -378,6 +378,16 @@ bool program_node::has_padded_dependency() const {
     });
 }
 
+bool program_node::is_fused_dep(size_t dep_idx) const {
+    for (auto fused : get_fused_primitives()) {
+        if (dep_idx >= fused.dep_start_idx) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
 std::map<size_t, memory::ptr> program_node::get_const_memory_deps() const {
     std::map<size_t, memory::ptr> mem_deps;
     for (auto& i : get_shape_infer_dependencies()) {
@@ -385,6 +395,12 @@ std::map<size_t, memory::ptr> program_node::get_const_memory_deps() const {
         if (i >= get_dependencies().size())
             continue;
 
+        // exclude fused dependency
+        if (is_fused_dep(i)) {
+            continue;
+        }
+
+        // constant type only
         auto& dep = get_dependency(i);
         if (dep.is_type<data>()) {
             mem_deps.insert({i, dep.as<data>().get_attached_memory_ptr()});
--- a/src/plugins/intel_gpu/tests/test_cases/deconvolution_gpu_test.cpp
+++ b/src/plugins/intel_gpu/tests/test_cases/deconvolution_gpu_test.cpp
@@ -7,6 +7,7 @@
 #include <intel_gpu/primitives/input_layout.hpp>
 #include <intel_gpu/primitives/deconvolution.hpp>
 #include <intel_gpu/primitives/crop.hpp>
+#include <intel_gpu/primitives/eltwise.hpp>
 #include <intel_gpu/primitives/reorder.hpp>
 #include <intel_gpu/primitives/data.hpp>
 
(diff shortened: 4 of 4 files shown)


src/plugins/intel_gpu/src/graph/program_node.cpp
```diff
@@ -378,14 +378,40 @@ bool program_node::has_padded_dependency() const {
...
+
+    return false;
+}
+
+std::map<size_t, memory::ptr> program_node::get_const_memory_deps(bool exclude_fused_dep) const {
     std::map<size_t, memory::ptr> mem_deps;
     for (auto& i : get_shape_infer_dependencies()) {
         // Some primitives may have flexible count of deps (e.g. reshape), thus allow skipping some deps
         if (i >= get_dependencies().size())
             continue;
```
vladimir-paramuzov, line 403: Can't the same to be achieved by 
```
if (i >= get_primitive()->input_size())
    continue;
```

?
get_dependencies() returns all deps including fused ones, while get_primitive()->input_size() inputs count of original primitive.

src/plugins/intel_gpu/src/graph/program_node.cpp
```diff
@@ -378,14 +378,40 @@ bool program_node::has_padded_dependency() const {
...
+            if (compare_rt == 0) {
+                return true;
+            } else if (compare_rt < 0) {
+                auto cropped_dep_id = dep_id.substr(0, fused_dep.first.size());    // in case of adding '_reorder_'
+                if (fused_dep.first.compare(cropped_dep_id) == 0)
+                    return true;
+            }
+        }
+    }
+
```
yeonbok, line 394: I think you can simply use fused_primitive_desc::dep_start_idx

yeonbok:
Please add a unittest which checks whther the fused input is being regarded as impl_param->mem_dep.
1) Create some topology with  Deconvolution  (e.g, [1, 3, 2, 2] = [1,3,4,4])+ Eltwise (which does not same spatial shape with Deconv output, broadcasting output, e.g,, [1, 3, 1, 1]) 
2) Do shape infer for Deconv

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -190,6 +190,10 @@ void primitive_inst::update_shape() {
         }
         auto& dep = _node->get_dependency(i);
         auto dep_id = dep.id();
+        // should not be deps of fused prims
+        if (_node->is_fused_dep(dep_id)) {
+            continue;
```
yeonbok, line 195: If i is fused dep, we can break this loop.

src/plugins/intel_gpu/src/graph/program_node.cpp
kelvinchoi-intel: I think this is better idea than using fused_primitive_desc::dep_start_idx

src/plugins/intel_gpu/src/graph/program_node.cpp
kelvinchoi-intel: done by using get_primitive()->input_size()

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
kelvinchoi-intel: done

src/plugins/intel_gpu/src/graph/program_node.cpp
```diff
@@ -378,13 +378,21 @@ bool program_node::has_padded_dependency() const {
...
-std::map<size_t, memory::ptr> program_node::get_const_memory_deps() const {
+std::map<size_t, memory::ptr> program_node::get_const_memory_deps(bool exclude_fused_dep) const {
     std::map<size_t, memory::ptr> mem_deps;
     for (auto& i : get_shape_infer_dependencies()) {
         // Some primitives may have flexible count of deps (e.g. reshape), thus allow skipping some deps
         if (i >= get_dependencies().size())
             continue;
 
+        // exclude fused dependency
+        if (exclude_fused_dep) {
```
vladimir-paramuzov, line 389: Do we really need this flag? Can't it always be =true?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -190,6 +190,10 @@ void primitive_inst::update_shape() {
         }
         auto& dep = _node->get_dependency(i);
         auto dep_id = dep.id();
+        // exclude fused node from memory_deps
+        if (i >= _node->get_primitive()->input_size()) {
```
vladimir-paramuzov, line 194: I think line 188 can be modified instead: `_node->get_dependencies().size()` -> `_node->get_primitive()->input_size()`

src/plugins/intel_gpu/src/graph/program_node.cpp
kelvinchoi-intel: get_kernel_impl_params() has params->memory_deps = get_const_memory_deps();
I think 'exclude_fused_dep' should be false at this time.

src/plugins/intel_gpu/src/graph/program_node.cpp
kelvinchoi-intel: I've changed to use fused_primitive_desc::dep_start_idx instead of get_primitive()->input_size() because get_primitive() doesn't represent original prim.

src/plugins/intel_gpu/src/graph/program_node.cpp
kelvinchoi-intel: I've changed to use fused_primitive_desc::dep_start_idx instead of get_primitive()->input_size() because get_primitive() doesn't represent original prim.

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
kelvinchoi-intel: I've changed to use fused_primitive_desc::dep_start_idx instead of get_primitive()->input_size() because get_primitive() doesn't represent original prim.

src/plugins/intel_gpu/src/graph/program_node.cpp
vladimir-paramuzov: Why? fused deps are not supposed to be used via `memory_deps`, so to my this check be removed.

yeonbok: I think it can be true always.

kelvinchoi-intel: Deleted the 'if' condition

src/plugins/intel_gpu/src/graph/include/program_node.h
```diff
@@ -99,7 +99,9 @@ struct program_node {
         return false;
     }
 
-    std::map<size_t, memory::ptr> get_const_memory_deps() const;
+    bool is_fused_dep(size_t dep_idx) const;
```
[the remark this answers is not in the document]
kelvinchoi-intel: done

src/plugins/intel_gpu/src/graph/program_node.cpp
```diff
@@ -378,13 +378,29 @@ bool program_node::has_padded_dependency() const {
     });
 }
 
+bool program_node::is_fused_dep(size_t dep_idx) const {
+    for (auto fused : get_fused_primitives()) {
+        if (dep_idx >= fused.dep_start_idx) {
```
vladimir-paramuzov, line 383: Does it work correctly for dep_idx = 0 and some fused op w/o dependencies? I assume that dep_start_idx will be also 0 in that case.

yeonbok: I think we can utilize total_num_deps too.

kelvinchoi-intel: local_desc.dep_start_idx = fused_node.get_dependencies().size();
Initial value of dep_start_idx comes from fused_node's dep size and it can't be 0.

PR #16305, opened 2023-03-15
Simple fix of redundant const-weight reordering for brgconv node in dynamic model

usstq:
### Details:
 - Remove WA on 86948 (the WA which stops trying brgconv in dynamic mode)
 - Copy weight reordering design from fullyconnect node into conv:

### Tickets:
 - CVS-86948

### DIFF
--- a/src/plugins/intel_cpu/src/node.cpp
+++ b/src/plugins/intel_cpu/src/node.cpp
@@ -841,6 +841,52 @@ void Node::prepareMemory(dnnl::primitive_desc_iterator& itpd) {
     Node::prepareMemory(intDescs);
 }
 
+
+MemoryPtr Node::prepareWeightMemory(DnnlMemoryDescPtr weightDesc) {
+    if (!getParentEdgeAt(1)->getParent()->isConstant())
+        IE_THROW() << "Weight input is not const for node " << getName() << ".";
+    auto blob = getParentEdgeAt(1)->getMemoryPtr();
+    if (!blob)
+        IE_THROW() << "Cannot get const weights blob for node " << getName() << ".";
+
+    auto constDnnlMemOutDesc = blob->GetDescWithType<DnnlMemoryDesc>();
+    auto weightSrcDesc = constDnnlMemOutDesc->getDnnlDesc();
+    weightSrcDesc = weightSrcDesc.reshape(weightDesc->getDnnlDesc().dims());
+    auto create = [&] () {
+        auto newSrcDesc = DnnlExtensionUtils::makeDescriptor(weightSrcDesc);
+
+        Memory srcMemory{ getEngine() };
+        srcMemory.Create(newSrcDesc, blob->GetData());
+
+        MemoryPtr _ptr = std::make_shared<Memory>(getEngine());
+        _ptr->Create(weightDesc);
+        node::Reorder::reorderData(srcMemory, *_ptr, context->getParamsCache());
+
+        return _ptr;
+    };
+
+    MemoryPtr ptr;
+    const auto& format = weightDesc->serializeFormat();
+    auto itr = privateWeightCache.find(format);
+    if (privateWeightCache.end() != itr) {
+        ptr = itr->second;
+    } else {
+        auto weightCache = context->getWeightsCache();
+        if (weightCache != nullptr) {
+            const std::string string_hash = getName() + "_" + format
+                                            + "_" + std::to_string(blob->GetSize())
+                                            + "_" + std::to_string(reinterpret_cast<uint64_t>(blob->GetData()));
+
+            ptr = *weightCache->findOrCreate(string_hash, create);
+        } else {
+            ptr = create();
+        }
+        privateWeightCache[format] = ptr;
+    }
+
+    return ptr;
+}
+
 bool Node::isInPlace() {
     if (inplace == InPlaceType::Unknown) {
         auto selected_pd = getSelectedPrimitiveDescriptor();
--- a/src/plugins/intel_cpu/src/node.h
+++ b/src/plugins/intel_cpu/src/node.h
@@ -620,6 +620,15 @@ protected:
     void prepareMemory(const std::vector<DnnlMemoryDescPtr>& intDescs);
     void prepareMemory(dnnl::primitive_desc_iterator& itpd);
 
+    // we cannot account per-NUMA weightCache for caching weights because:
+    //   1.it may not exist(in single stream configuration)
+    //   2.it only holds weak references, the life-cycle of cached item
+    //     is still under control of strong references outside of cache.
+    // privateWeightCache is for holding strong references to weight copies
+    // of same raw data but with different layouts.
+    std::unordered_map<std::string, MemoryPtr> privateWeightCache;
+    MemoryPtr prepareWeightMemory(DnnlMemoryDescPtr weightDesc);
+
     bool isDynamic = false;
 
     bool isInputTensorAtPortEmpty(size_t port) const;
--- a/src/plugins/intel_cpu/src/nodes/conv.cpp
+++ b/src/plugins/intel_cpu/src/nodes/conv.cpp
@@ -1602,22 +1602,19 @@ void Convolution::appendZeroPointsArgs() {
 
 void Convolution::initTryBrgconvFlag() {
     // Due to performance issue, brgconv will only be enabled by default:
-    // 1, static shape(dynamic shape may change weights layout if the input shape changes and cause performance issue: 86948)
-    // 2, support amx except having input zero point.
-    // 3, support avx512 without legacy postops/per channel zero point when avx512
-    if (!isDynamicNode()) {
-        if (dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx512_core_amx)) {
-            shouldTryBrgconv = true;
-        } else if (dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx512_core)) {
-            shouldTryBrgconv = true;
-            // should remove after binary postops performance issue resolved
-            // heuristics: if it's  avx512 ISA model && it doesn't have binary post ops or per channel zero point.
-            dnnl::primitive_attr attr;
-            setPostOps(attr, outputStaticShape(), false);
-            const auto& ops = attr.get_post_ops();
-            if (ops.get()->find(dnnl::impl::primitive_kind::binary) != -1) {
-                shouldTryBrgconv = false;
-            }
+    // 1, support amx except having input zero point.
+    // 2, support avx512 without legacy postops/per channel zero point when avx512
+    if (dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx512_core_amx)) {
+        shouldTryBrgconv = true;
+    } else if (dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx512_core)) {
+        shouldTryBrgconv = true;
+        // should remove after binary postops performance issue resolved
+        // heuristics: if it's  avx512 ISA model && it doesn't have binary post ops or per channel zero point.
+        dnnl::primitive_attr attr;
+        setPostOps(attr, outputStaticShape(), false);
+        const auto& ops = attr.get_post_ops();
+        if (ops.get()->find(dnnl::impl::primitive_kind::binary) != -1) {
+            shouldTryBrgconv = false;
         }
     }
 }
--- a/src/plugins/intel_cpu/src/nodes/conv.cpp
+++ b/src/plugins/intel_cpu/src/nodes/conv.cpp
@@ -49,6 +49,8 @@ struct ConvKey {
     dnnl::primitive_attr attr;
     impl_desc_type implType;
 
+    bool constWeight;
+
     size_t hash() const;
     bool operator==(const ConvKey& rhs) const;
 };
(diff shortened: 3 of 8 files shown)


src/plugins/intel_cpu/src/node.cpp
```diff
@@ -834,6 +834,51 @@ void Node::prepareMemory(dnnl::primitive_desc_iterator& itpd) {
     Node::prepareMemory(intDescs);
 }
 
+MemoryPtr Node::prepareWeightMemory(DnnlMemoryDescPtr weightDesc) {
+    if (!getParentEdgeAt(1)->getParent()->isConstant())
+        IE_THROW() << "Weight input is not const for node " << getName() << ".";
+    auto blob = getParentEdgeAt(1)->getMemoryPtr();
```
maxnick, line 840: `blob` -> `edgeMem`

src/plugins/intel_cpu/src/node.h
```diff
@@ -620,6 +620,15 @@ class Node {
     void prepareMemory(const std::vector<DnnlMemoryDescPtr>& intDescs);
     void prepareMemory(dnnl::primitive_desc_iterator& itpd);
 
+    // we cannot rely on per-NUMA weightCache for caching weights because:
+    //   1.it may not exist(in single stream configuration)
+    //   2.it only holds weak references, the life-cycle of cached item
+    //     is still under control of strong references outside of cache.
+    // privateWeightCache is for holding strong references to constant weight
+    // copies of same content with different layouts.
+    std::unordered_map<std::string, MemoryPtr> privateWeightCache;
```
maxnick, line 629: `privateWeightCache` should probably be moved to the `private` section, to prevent uncontrolled access.

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1368,10 +1376,22 @@ void Convolution::prepareParams() {
...
 
     auto engine = getEngine();
     auto builder = [&engine](const ConvKey& key) -> executorPtr {
+        // remove the requirement on weight memory layout to let primitive
+        // report the best layout for weight to be reordered dynamically at runtime
+        auto src_dt = key.inp0->getDataType();
+        auto wdt = src_dt;
+        if (src_dt == dnnl_s8 || src_dt == dnnl_u8) {
+            wdt = memory::data_type::s8;
+        }
```
maxnick, line 1390: Shouldn't we retrive the data from the `key.inp1`? Of course if we've set the proper precision on the compilation stage (see previous comment).

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1159,6 +1162,11 @@ bool Convolution::isPossibleToSkipInitConfig(const dnnl::primitive_desc &desc) c
 }
 
 std::shared_ptr<MemoryDesc> Convolution::getSrcMemDesc(dnnl::primitive_desc_iterator &primitive_desc_it, size_t idx) {
+    if (idx == 1) {
+        // report original plain layout for weight since it needs to be reordered dynamically at runtime
+        return std::make_shared<CpuBlockedMemoryDesc>(getOriginalInputPrecisionAtPort(idx),
+                                                      Shape(getInputShapeAtPort(idx).getStaticDims()));
```
maxnick, line 1168: Shouldn't we extract the precision from the actual weights? In my understanding only the layout should be replaced.

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1481,9 +1500,21 @@ void Convolution::prepareParams() {
...
         primArgs[DNNL_ARG_SRC] = srcMemPtr->GetPrimitive();
-        primArgs[DNNL_ARG_WEIGHTS] = wghMemPtr->GetPrimitive();
         primArgs[DNNL_ARG_DST] = dstMemPtr->GetPrimitive();
 
+        if (key.constWeight) {
+            // const weight preparation/reordering needs to be done once at next execution
+            // when the input weight data is guaranteed to be ready (considering possible const-folding
+            // subgraphs inserted between constant weight node and conv)
+            auto it = primArgs.find(DNNL_ARG_WEIGHTS);
+            if (it == primArgs.end() || it->second.get_desc() != execPtr->getWeightDesc()) {
```
maxnick, line 1510: Unfortunately, `get_desc()` is a costly operation since it does memory descriptor clone under the hood. So to reduce the impact, it does make sense to consider the approach with using `prevExecPtr` to read the memory descriptor.

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1534,6 +1567,11 @@ void Convolution::execute(dnnl::stream strm) {
         IE_THROW() << "Can't execute Convolution node with name: " << getName() << ", because executor is not compiled";
     }
 
+    if (pendingConstWeightReorder) {
+        primArgs[DNNL_ARG_WEIGHTS] = prepareWeightMemory(DnnlExtensionUtils::makeDescriptor(execPtr->getWeightDesc()))->GetPrimitive();
+        pendingConstWeightReorder = false;
+    }
```
maxnick, line 1573: What is the reason of putting this check to `execute` instead of `prepareParams()`?

src/plugins/intel_cpu/src/nodes/conv.cpp
usstq: In some model, the weight is provided in s32 or other precision which is un-supported by oneDNN, and CPU node will use f32 as inference precision in such case, so oneDNN primitive also requires f32 for weight precision.

if we report `f32+plain_layout` here, CPU graph will insert reorders which converts `s32+plain_layout` into `f32+plain_layout`, this only performed once for constant weight, but it only produces a temporary memory format, since after `prepareParam` stage, the layout is going to change again, so here I report the original precision specified in ngraph, to prevent the extra reorder nodes and temporary memories, one reorder in prepareWeightMemory() or DnnlExecutor() should be enough in all cases, right?

src/plugins/intel_cpu/src/nodes/conv.cpp
usstq: because we want to reorder only once from the original constant ngraph node(see previous reply), so we have to derive the actual precision here based on oneDNN's requirement.

src/plugins/intel_cpu/src/nodes/conv.cpp
usstq: Good idea, will change!

src/plugins/intel_cpu/src/nodes/conv.cpp
usstq: we cannot be 100% sure that the constant data is ready to access at `prepareParam()` stage, considering that there may be a constant folding subgraph between the constant input node and `this` Conv node, which are executed after `CreatePrimitives()`, check this code snippets:

```cpp
void Graph::InitGraph() {
  ...
  CreatePrimitives();      // prepareParam() is invoked
  ...
  ExecuteConstantNodesOnly(); // const-folding subgraph is executed
```

 and `prepareParam()` is called inside `Node::createPrimitive()`.

src/plugins/intel_cpu/src/node.h
usstq: Yes, will move it

src/plugins/intel_cpu/src/node.cpp
usstq: OK, will change

src/plugins/intel_cpu/src/nodes/conv.cpp
maxnick: Ok. That does make sense. However, on lines: https://github.com/openvinotoolkit/openvino/blob/934224b14977c1363f87d0355c316873b8e0ea2d/src/plugins/intel_cpu/src/nodes/conv.cpp#L885-L887 we are doing pretty the same check. So I would propose to keep this check in one place to remove this coupling between those two code frames. To this end we can simply extract this check to a separate method.

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1481,9 +1500,21 @@ void Convolution::prepareParams() {
...
         primArgs[DNNL_ARG_SRC] = srcMemPtr->GetPrimitive();
-        primArgs[DNNL_ARG_WEIGHTS] = wghMemPtr->GetPrimitive();
         primArgs[DNNL_ARG_DST] = dstMemPtr->GetPrimitive();
 
+        if (key.constWeight) {
+            // const weight preparation/reordering needs to be done once at next execution
+            // when the input weight data is guaranteed to be ready (considering possible const-folding
+            // subgraphs inserted between constant weight node and conv)
+            auto it = primArgs.find(DNNL_ARG_WEIGHTS);
+            if (it == primArgs.end() || prevExecPtr->getWeightDesc() != execPtr->getWeightDesc()) {
```
maxnick, line 1510: It does make sense to add `prevExecPtr` null check to prevent nullptr dereferencing.

src/plugins/intel_cpu/src/nodes/conv.cpp
maxnick: Ok. Got it.

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/conv_eltwise.cpp
```diff
@@ -16,7 +16,7 @@ INSTANTIATE_TEST_SUITE_P(smoke_Snippets_ConvAdd, ConvEltwise,
         ::testing::Values(convInputShape),
         ::testing::Values(convInputShape),
         ::testing::Values(std::shared_ptr<ov::Node> (std::make_shared<ov::op::v1::Add>())), // non-tokenizable
-        ::testing::Values(6), // num nodes = 6: Convert + Convolution + 4 Reorders on Convs in&outs
+        ::testing::Values(5), // num nodes = 6: Convert + Convolution + 3 Reorders on Convs in&outs
```
maxnick, line 19:
```suggestion
        ::testing::Values(5), // num nodes = 5: Convert + Convolution + 3 Reorders on Convs in&outs
```
???

src/plugins/intel_cpu/src/nodes/conv.cpp
maxnick: That does make sense. Agree. This is another reason to refactor how we work with oneDNN.

maxnick:
@usstq , https://github.com/openvinotoolkit/openvino/pull/16372 has been merged, could you please rebase?

src/plugins/intel_cpu/src/nodes/conv.cpp
maxnick, line 1510: After the rebase to the new version of the DnnlExecutor, `prevExecPtr->getWeightDesc() != execPtr->getWeightDesc()` will have the semantics of comparison of two pointers, so please change this check to `!execPtr->getWeightDesc()->isCompatible(*(prevExecPtr->getWeightDesc()))`

src/plugins/intel_cpu/src/nodes/conv.cpp
usstq: OK, will change

src/plugins/intel_cpu/src/nodes/conv.cpp
usstq: OK, will add `|| !prevExecPtr ||` into the condition so when there is no valid `prevExecPtr` (which is rare) we will also reorder the constant weight.

src/plugins/intel_cpu/tests/functional/shared_tests_instances/snippets/conv_eltwise.cpp
usstq: Sorry, will change!

src/plugins/intel_cpu/src/nodes/conv.cpp
usstq: OK, will change!

usstq:
@maxnick review comments applied, please review again, Thx!

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1497,12 +1527,14 @@ Convolution::ConvolutionExecutor::ConvolutionExecutor(const dnnl::convolution_fo
...
                                                                 const dnnl::memory::desc& outMemDesc,
-                                                                const dnnl::engine& engine) : DnnlExecutor(pd) {
+                                                                const dnnl::engine& engine,
+                                                                bool constWeight) : DnnlExecutor(pd) {
     if (inMemDesc != getDnnlSrcDesc()) {
         inputReorders.insert({DNNL_ARG_SRC, IntermReorder(inMemDesc, getDnnlSrcDesc(), engine)});
     }
 
-    if (weightMemDesc != getDnnlWeightDesc()) {
+    if (weightMemDesc != getDnnlWeightDesc() && !constWeight) {
```
maxnick, line 1536: It is better to swap these two conditions, since `!constWeight` is likely to be false and it is by far cheaper that the comparison of two memory descs.

usstq: Great idea, will fix the order

usstq:
@maxnick  order is changed, please review again, Thanks!

dmitry-gorokhov (approved):

PR #16307, opened 2023-03-15
[GPU] change activation position in reorder_data_bfyx_to_blocked_format kernel

wilson-seok:
### Details:
 - Change activation position in reorder_data_bfyx_to_blocked_format kernel. Activation in load function causes ambiguous call error of fmax/fmin when input/output data type are different.

### Tickets:
 - 103092


isanghao (approved):
This might have issue for ordering of operation between activation and precision-conversion. But it was also issue from original implementation. So let's go this way for now.

PR #16011, opened 2023-02-28
Improve detectron2 support

mvafin:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*

PR #16312, opened 2023-03-15
[PT FE] Add aten::new_empty

mmikolajcz:
### Details:
 - *Add aten::new_empty based on draft from https://github.com/openvinotoolkit/openvino/pull/16011*
 - *Add test cases for aten::new_empty*
 - *Fix sporadic issues with existing tests for aten::empty*
 - *Fix issues with aten::full_like, aten::ones_like and aten::zeros_like where specified dtype is none and it should default to dtype of input.*

### Tickets:
 - *101109*

### DIFF
--- a/src/frontends/pytorch/src/op/full.cpp
+++ b/src/frontends/pytorch/src/op/full.cpp
@@ -185,23 +185,6 @@ OutputVector translate_empty(NodeContext& context) {
     }
     return {empty};
 };
-
-OutputVector translate_new_empty(NodeContext& context) {
-    // aten::new_empty(Tensor self, SymInt[] size, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None,
-    // bool? pin_memory=None) -> Tensor
-    num_inputs_check(context, 2, 6);
-    auto input = context.get_input(0);
-    auto sizes = context.get_input(1);
-    auto value = context.mark_node(v0::Constant::create(element::f32, Shape{}, {0}));
-    int dtype_id = 2;
-    Output<Node> empty;
-    if (!context.input_is_none(dtype_id)) {
-        empty = base_translate_full_with_convert(context, sizes, value, dtype_id);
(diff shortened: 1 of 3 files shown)


src/frontends/pytorch/src/op/full.cpp
```diff
@@ -186,7 +186,23 @@ OutputVector translate_empty(NodeContext& context) {
     return {empty};
 };
 
+OutputVector translate_new_empty(NodeContext& context) {
+    // aten::new_empty(Tensor self, SymInt[] size, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None,
+    // bool? pin_memory=None) -> Tensor
+    num_inputs_check(context, 2, 6);
+    auto input = context.get_input(0);
+    auto sizes = context.get_input(1);
+    auto value = context.mark_node(v0::Constant::create(element::f32, Shape{}, {0}));
```
tomdol, line 195: Should it be a constant with a single value or an empty constant?

tomdol (approved):

src/frontends/pytorch/src/op/full.cpp
mmikolajcz: It should be constant scalar with value 0, similarly like aten::empty.
Now when I'm looking into it, it should behave exactly like aten::new_zeros and aten::zeros, however, I think there is an easy to fix bug in current implementation of those op that fail tests, I'll try that and try to remove duplication if possible

PiotrKrzem (approved):

src/frontends/pytorch/src/op/full.cpp
mmikolajcz: My bad, while aten::new_zero and aten::new_empty seem to have same scheme and can be covered by the same translation, there seem to be a difference in schema between aten::zero and aten::empty.

mvafin (approved):

PR #16322, opened 2023-03-15
[VPUX] - failures : [ ERROR ] Source and destination tensors shapes and byte sizes are expected to be equal for data copying

DariaMityagina:
**Description:** After updating the `master` commit of Openvino, a new error appeared on the models that were fine before (during benchmark_app run):
`[ ERROR ] Source and destination tensors shapes and byte sizes are expected to be equal for data copying`

**Ticket:** E-72838

### DIFF
--- a/src/inference/src/dev/iremote_tensor.cpp
+++ b/src/inference/src/dev/iremote_tensor.cpp
@@ -62,7 +62,7 @@ public:
     }
 
     const Shape& get_shape() const override {
-        m_shape = blob->getTensorDesc().getBlockingDesc().getBlockDims();
+        m_shape = blob->getTensorDesc().getDims();
         return m_shape;
     }
 


DariaMityagina:
I tested the fix using our CI.
See the comments on the ticket for details.

OrestChura (approved):

src/inference/src/dev/iremote_tensor.cpp
```diff
@@ -62,7 +62,7 @@ class BlobTensor : public ITensor {
     }
 
     const Shape& get_shape() const override {
-        m_shape = blob->getTensorDesc().getBlockingDesc().getBlockDims();
+        m_shape = blob->getTensorDesc().getDims();
```
ilyachur, line 65: It is incorrect change. So OpenVINO tensors don't have layouts, but blob has, so it means that:
```
IE::Blob (dims = {1, 3, 2, 2}, Layout = "NCHW") == ov::Tensor(shape = {1, 3, 2, 2})
but:
IE::Blob (dims = {1, 3, 2, 2}, Layout = "NHWC") == ov::Tensor(shape = {1, 2, 2, 3})
```
It is why `get_shape()` returns block dims which are reordered for the blob layout.
This behaviour is aligned with host tensors.

DariaMityagina:
The tests fail.
@ilyachur could you please advise on what could have gone wrong?
Since we had perfectly running tests before these changes:
- https://github.com/openvinotoolkit/openvino/commit/a5ec5f547635426d0e8f3eeed4fcf421fdaa8ec0
- https://github.com/openvinotoolkit/openvino/commit/95faa573ed0c18c6c1e4b05aa0e1ecabcdcac958

PR #16337, opened 2023-03-16
[GNA] Pwl rework new algorithm

marcinkusm:
### Details:
    * added implementation based on inheritance and composition
    * updated tests
    * added option to choose PWL approximation mode with two options.
    * updated speech_sample

### Tickets:
 - *111218*, *93631*


akladiev:

marcinkusm:
This PR is still valid, changes under construction.

src/plugins/intel_gna/src/transformations/pwl/activation_representation_configuration.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "activation_representation.hpp"
+#include "boundaries_fq_hanlder.hpp"
+#include "openvino/core/node.hpp"
+
+namespace ov {
+namespace intel_gna {
+namespace pass {
+namespace pwl {
+
```
marcinkacprzak, line 15: Although your classes and methods are short and quite self-descriptive, it would be good to provide at lease short descriptions.

src/plugins/intel_gna/tests/unit/transformations/gna_pwl.cpp
```diff
@@ -9,19 +9,34 @@
...
+}
+
 namespace {
 
 template <typename T>
 struct Function {};
 
 template <>
-struct Function<ngraph::opset9::Sigmoid> {
+struct Function<ov::opset11::Sigmoid> {
```
marcinkacprzak, line 39: It would be better to use "using namespace ov::opset11;" and then Function<Sigmoid>. It would be easier in the future to change the opset to the newest for the whole file.

src/inference/include/ie/gna/gna_config.hpp
```diff
@@ -199,13 +199,11 @@ INFERENCE_ENGINE_DEPRECATED("The config key is deprected and will be removed")
 DECLARE_GNA_CONFIG_KEY(PWL_UNIFORM_DESIGN);
 
 /**
- * @brief The option to allow to specify the maximum error percent that the optimized algorithm finding
- * will use to find PWL functions.
- * By default (in case of NO value set), 1.0 value is used.
- * @deprecated The config key is deprecated and will be removed in a future release.
+ * @brief The option to set PWL Approximation mode. Allow to spacify how accurate should be approximation
```
mkwapul, line 202: cosmetic: spacify -> specify

samples/cpp/speech_sample/speech_sample.hpp
```diff
@@ -125,8 +125,9 @@ static const char layout_message[] =
 ;
 
 /// @brief message for PWL max error percent
-static const char pwl_max_error_percent_message[] = "Optional. The maximum percent of error for PWL function."
-                                                    "The value must be in <0, 100> range. The default value is 1.0.";
+static const char pwl_approximation_mode_message[] =
```
mkwapul, line 128: I wonder if constexpr would be ok?

src/plugins/intel_gna/src/transformations/pwl/configuration_log.cpp
```diff
@@ -0,0 +1,74 @@
...
+constexpr const double ConfigurationLog::kAccuracyErrorPercentage;
+constexpr const double ConfigurationLog::kBreakBound;
+constexpr const size_t ConfigurationLog::kMaxIterations;
+
+double ConfigurationLog::FunctionLog::get_value(double x) const {
+    return std::log(x);
+}
+
+double ConfigurationLog::FunctionLog::get_first_derivative(double x) const {
+    return 1.0 / x;
```
mkwapul, line 32: it's public function: what assures x>0?

src/plugins/intel_gna/src/transformations/pwl/configuration_tanh.cpp
```diff
@@ -0,0 +1,73 @@
...
+double ConfigurationTanh::FunctionTanh::get_first_derivative(double x) const {
+    return 1.0 - tanh(x) * tanh(x);
+}
+
+std::shared_ptr<ActivationRepresentationConfiguration> ConfigurationTanh::create_config(PWLApproximationMode mode) {
+    return std::make_shared<ConfigurationTanh>(mode, std::make_shared<BoundariesFQHandlerImpl>());
+}
+
+ConfigurationTanh::ConfigurationTanh(PWLApproximationMode mode, std::shared_ptr<BoundariesFQHandler> boundaries_handler)
+    : ActivationRepresentationConfigurationBase(std::move(boundaries_handler)),
```
mkwapul, line 40: are you going to avoid passing param through reference, that's why move semantics?

If yes, then not convinced if taking control over rvalue parameter passing is desirable here.
will copy elision handle it on it's own? - you are passing rvalue at the beginning of call chain. If not in C++11, then in C++17?... but we won't fix code when we'll be in C++17.

Decision factors:
1. is this code running under performance path? IMO not - it's at configuration phase.
2. Code readability/simplicity/maintainability might also be taken into account.
3. is similar solution observable in other OV parts?

src/plugins/intel_gna/src/transformations/pwl/segment_generator.cpp
```diff
@@ -0,0 +1,250 @@
...
+    bool same_epsilon = false;
+    double Delta;
+    double epsilon_final = 0.0;
+    double max_epsilon = 0.0;
+    double max_epsilon_prev;
+    double min_epsilon;
+    double sgn = (negative) ? -1.0 : 1.0;
+    int j;
+
+    // Figure 4:  Box #1
```
mkwapul, line 126: is there some ASCII art comment refers to?

src/plugins/intel_gna/src/descriptions/gna_flags.hpp
```diff
@@ -8,7 +8,6 @@
 
 #include "gna/gna_config.hpp"
 #include "openvino/runtime/intel_gna/properties.hpp"
-
```
mkwapul, line 11: ditto

src/plugins/intel_gna/src/transformations/pwl/base_segment_precision_validator.cpp
```diff
@@ -0,0 +1,52 @@
...
+        if (val > max_val)
+            max_val = val;
+        if (val < min_val)
+            min_val = val;
+    }
+    return std::abs(max_val - min_val);
+}
+
+double BaseSegmentPrecisionValidator::calculate_error_percentage(double epsilon, double range) const {
+    return (100.0 * std::fabs(epsilon) / range);
```
rjeziers, line 47:
```suggestion
    return (100.0 * std::abs(epsilon) / range);
```

src/plugins/intel_gna/src/transformations/pwl/configuration_log.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+std::shared_ptr<ActivationRepresentationConfiguration> ConfigurationLog::create_config(PWLApproximationMode mode) {
+    // prvious implementation didn't used does not use fake_quantize at all for Log
```
rjeziers, line 36:
```suggestion
    // previous implementation didn't use fake_quantize at all for Log
```

src/plugins/intel_gna/src/transformations/pwl/configuration_exp.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "openvino/opsets/opset11.hpp"
+#include "openvino/pass/pattern/op/wrap_type.hpp"
+
+namespace ov {
+namespace intel_gna {
+namespace pass {
+namespace pwl {
+
+const std::string ConfigurationExp::kActivationName = "Exp";
+const double ConfigurationExp::kLowerBound = -std::log2(std::numeric_limits<int16_t>::max());
```
rjeziers, line 18: Could you clarify why such values for kLowerBound and kUpperBound were chosen? Using log2 and log10 looks like an accident to me, please add a comment if it's intentional

src/plugins/intel_gna/src/transformations/pwl/segment.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+#pragma once
+
+namespace ov {
+namespace intel_gna {
+namespace pass {
+namespace pwl {
+struct Segment {
+    Segment() = default;
+    Segment(double im, double ib, double ialpha, double ibeta = 0) : m(im), b(ib), alpha(ialpha), beta(ibeta) {}
```
rjeziers, line 13:
```suggestion
    Segment(double im, double ib, double ialpha, double ibeta = 0.0) : m(im), b(ib), alpha(ialpha), beta(ibeta) {}
```

src/plugins/intel_gna/src/transformations/pwl/segment_generator.cpp
```diff
@@ -0,0 +1,250 @@
...
+    j = 0;
+    Delta = 1.0;
+
+    for (int i = 0; i < segments_num; i++) {
```
rjeziers, line 130: This will trigger a warning - comparison of signed and unsigned value.
Please also check other loops.

src/plugins/intel_gna/tests/unit/transformations/gna_pwl.cpp
```diff
@@ -9,19 +9,34 @@
...
+#include "openvino/runtime/intel_gna/properties.hpp"
 #include "transformations/pwl_approximation.hpp"
 
+using namespace ov::intel_gna;
 using namespace ov::intel_gna::common;
 
+constexpr double kAccuracyMaxErrorPrecentage = 0.09;
+constexpr double kPerformanceMaxErrorPrecentage = 1.0;
+
+double get_max_error_percentage(const PWLApproximationMode& mode) {
```
mkwapul, line 26: in fact we don't need const& here, passing by value might spare us 4B? + a bit more elegant?

src/plugins/intel_gna/tests/unit/transformations/gna_pwl.cpp
```diff
@@ -76,117 +91,182 @@ struct Function<ngraph::opset9::Power> {
...
 template <typename T>
 class GnaPWlTestsFixture {
 public:
     template <typename U = T, typename Enable<U>::type = 0>
-    GnaPWlTestsFixture(const ngraph::Shape& input_shape,
+    GnaPWlTestsFixture(const ov::Shape& input_shape,
                        double lower_bound,
                        double upper_bound,
-                       double max_error_percent);
+                       const ov::intel_gna::PWLApproximationMode& mode,
```
mkwapul, line 108: ditto

src/inference/include/openvino/runtime/intel_gna/properties.hpp
```diff
@@ -264,12 +264,47 @@ inline std::istream& operator>>(std::istream& is, PWLDesignAlgorithm& pwl_design
...
- * @brief The option to allow to specify the maximum error percent that the optimized algorithm finding
- * will be used to find PWL functions.
- * By default (in case of NO value set), 1.0 value is used.
+ * @brief Enum to define pwl approximation mode.
  * @ingroup ov_runtime_gna_prop_cpp_api
  */
-static constexpr Property<float> pwl_max_error_percent{"GNA_PWL_MAX_ERROR_PERCENT"};
+enum class PWLApproximationMode {
+    PERFORMANCE = 1,  //!<  Optimize for max performance, may apply properties which slightly affect accuracy
+    ACCURACY = 2,     //!<  Optimize for max accuracy
```
mkwapul, line 272: cosmetic: do we need '= 2' here?

src/plugins/intel_gna/src/gna_plugin_config.cpp
```diff
@@ -212,21 +212,10 @@ void Config::UpdateFromMap(const std::map<std::string, std::string>& config) {
...
-            } catch (std::invalid_argument&) {
-                THROW_GNA_EXCEPTION << "Invalid value of PWL max error percent";
-            } catch (std::out_of_range&) {
-                THROW_GNA_EXCEPTION << "Unsupported PWL error percent value: " << value
-                                    << ", should be greater than 0 and less than 100";
-            }
-            gnaFlags.pwlMaxErrorPercent = max_error;
-            OPENVINO_SUPPRESS_DEPRECATED_END
+        } else if (key == GNA_CONFIG_KEY(PWL_APPROXIMATION_MODE) || key == ov::intel_gna::pwl_approximation_mode) {
+            ov::intel_gna::PWLApproximationMode pwl_approximation_mode =
```
mkwapul, line 216: any reason to use this proxy variable?

src/plugins/intel_gna/tests/unit/transformations/gna_pwl.cpp
```diff
@@ -76,117 +91,182 @@ struct Function<ngraph::opset9::Power> {
...
+    std::enable_if<std::is_same<T, ov::opset11::Sigmoid>::value || std::is_same<T, ov::opset11::Tanh>::value ||
+                       std::is_same<T, ov::opset11::SoftSign>::value || std::is_same<T, ov::opset11::Log>::value ||
+                       std::is_same<T, ov::opset11::Exp>::value,
                    int>;
 template <typename T>
-using EnableWithExtraArg = std::enable_if<std::is_same<T, ngraph::opset9::Power>::value, int>;
+using EnableWithExtraArg = std::enable_if<std::is_same<T, ov::opset11::Power>::value, int>;
 
 template <typename T>
 class GnaPWlTestsFixture {
```
mkwapul, line 102: why W (capital) while 'l' (lower)?

samples/cpp/speech_sample/speech_sample.hpp
marcinkusm: I have used the same convention as for other variables in the file.

src/plugins/intel_gna/src/transformations/pwl/configuration_log.cpp
marcinkusm: This is inherited from old code and it assumes that +infinity/-infinity  will be assigned.

https://en.cppreference.com/w/cpp/language/operator_arithmetic

"dividing a non-zero number by ±0.0 gives the correctly-signed infinity and [FE_DIVBYZERO](https://en.cppreference.com/w/cpp/numeric/fenv/FE_exceptions) is raised"

src/plugins/intel_gna/src/transformations/pwl/configuration_tanh.cpp
marcinkusm: I want to have copy of shared_ptr - not copy of reference.
In such case it does not matter if I pass by const reference and assign to new object or I  pass by value and move.
Event more second option is better in case you create rvalue( create share_ptr as argument during invocation of method).

src/plugins/intel_gna/src/transformations/pwl/segment_generator.cpp
marcinkusm: This is left from old implementation and it is pointing to the figure of document where algorithm is explained.
Overall method was not modified beside some naming convention.

src/plugins/intel_gna/src/transformations/pwl/base_segment_precision_validator.cpp
marcinkusm: don't see the reason
https://en.cppreference.com/w/cpp/numeric/math/fabs

src/plugins/intel_gna/src/transformations/pwl/configuration_exp.cpp
marcinkusm: I am not sure.
this parameters where used in previous version of algorithm.
I suspect that this is related to limitations when conversion from flaots to integers happening further in backend.

src/plugins/intel_gna/tests/unit/transformations/gna_pwl.cpp
marcinkusm: yes I agree, but in many places in the code we use that convention, this is why I wanted to align, may change

src/inference/include/openvino/runtime/intel_gna/properties.hpp
marcinkusm: Theoretically not, but what is the problem with giving exact value?
Especially that it aligns with convention of other declarations in the file.

src/plugins/intel_gna/src/gna_plugin_config.cpp
marcinkusm: It left after after refactoring, will fix it.

src/inference/include/openvino/runtime/intel_gna/properties.hpp
mkwapul: having initial value only, enum's new values do not require to maintain them - less error prone. If C++ provides feature of autoincrement of enum values, why not use it.
By the way, explicit values are usable for bitfields for example.
I'm not insisting on though. please choose whatever you prefer.

src/plugins/intel_gna/tests/unit/transformations/gna_pwl.cpp
mkwapul: .... then, there is still ConfigurationSigmoid::create_config(PWLApproximationMode mode) left then.
choose whatever way better for you then.

src/plugins/intel_gna/src/transformations/pwl/configuration_tanh.cpp
mkwapul: >In such case it does not matter if I pass by const reference and assign to new object or I pass by value and move.
copy elision is such a compiler intelligence that for rvalues, it eliminates all the move constructor calls on the way. it does not require 'std::move' for it  - you just depend on compiler - code is cleaner, you don't justify compiler. The question is if it'll be enabled here (ie prerequisites satisfied). I'll try to check it during the day.

PR #16351, opened 2023-03-17
[GPU] Prevent memory reset at runtime allocation for dynamic shape, fix wrong padding handling

yeonbok:
### Details:
 - Prevent memory reset at runtime allocation for dynamic shape
 - Added mem reset when reusing memory for padded tensor  
 - Fixed wrong offset w.r.t padding in shape agnostic kernel 

### Tickets:
 - 104928

### DIFF
--- a/src/plugins/intel_gpu/src/graph/include/convolution_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/convolution_inst.h
@@ -129,7 +129,7 @@ public:
 
     bool need_reset_input_memory() const override {
         auto input_layout = _deps[0].first->_impl_params->get_output_layout(0);
-        if (input_layout.data_padding.lower_size() != tensor(0) || input_layout.data_padding.upper_size() != tensor(0)) {
+        if (input_layout.data_padding) {
             return true;
         }
         return false;
@@ -138,7 +138,7 @@ public:
     bool need_reset_output_memory() const override {
         bool res = parent::need_reset_output_memory();
         auto output_layout = _impl_params->get_output_layout(0);
-        if (output_layout.data_padding.lower_size() != tensor(0) || output_layout.data_padding.upper_size() != tensor(0)) {
+        if (output_layout.data_padding) {
             return true;
         }
         return res;
--- a/src/plugins/intel_gpu/src/graph/include/convolution_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/convolution_inst.h
@@ -127,6 +127,23 @@ public:
     static layout calc_output_layout(convolution_node const& node, kernel_impl_params const& impl_param);
     static std::string to_string(convolution_node const& node);
 
+    bool need_reset_input_memory() const override {
+        auto input_layout = _deps[0].first->_impl_params->get_output_layout(0);
+        if (input_layout.data_padding.lower_size() != tensor(0) || input_layout.data_padding.upper_size() != tensor(0)) {
+            return true;
+        }
+        return false;
+    }
+
+    bool need_reset_output_memory() const override {
+        bool res = parent::need_reset_output_memory();
+        auto output_layout = _impl_params->get_output_layout(0);
+        if (output_layout.data_padding.lower_size() != tensor(0) || output_layout.data_padding.upper_size() != tensor(0)) {
+            return true;
+        }
+        return res;
+    }
+
 public:
     typed_primitive_inst(network& network, convolution_node const& node);
 
--- a/src/plugins/intel_gpu/src/graph/include/deconvolution_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/deconvolution_inst.h
@@ -60,6 +60,23 @@ public:
     static layout calc_output_layout(deconvolution_node const& node, kernel_impl_params const& impl_param);
     static std::string to_string(deconvolution_node const& node);
 
+    bool need_reset_input_memory() const override {
+        auto input_layout = _deps[0].first->_impl_params->get_output_layout(0);
+        if (input_layout.data_padding.lower_size() != tensor(0) || input_layout.data_padding.upper_size() != tensor(0)) {
+            return true;
+        }
+        return false;
+    }
+
+    bool need_reset_output_memory() const override {
+        bool res = parent::need_reset_output_memory();
+        auto output_layout = _impl_params->get_output_layout(0);
+        if (output_layout.data_padding.lower_size() != tensor(0) || output_layout.data_padding.upper_size() != tensor(0)) {
+            return true;
+        }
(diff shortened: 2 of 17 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -261,7 +261,7 @@ void primitive_inst::realloc_if_needed() {
         GPU_DEBUG_TRACE_DETAIL << id() << ": realloc output memory. "
                                <<  " Current buffer_size=" << max_output_layout_size
                                <<  " Requested buffer_size=" << actual_layout.count() << std::endl;
-        _outputs = allocate_outputs(&updated_params);
+        _outputs = allocate_outputs(&updated_params, false);
```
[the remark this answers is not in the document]
yeonbok: Ah, ..yes, I agree. However regarding padding, I think we'll need to reset whenever the shape is changed. 
So in the future I think we'll need to fix such kernels with padding not to use the padded area as is but set inside the kernel :( 

But for now, I will update it to reset mem for such primitives with padding.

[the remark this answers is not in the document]
yeonbok: Anyway.. Seems that we'll have an potential issue regarding padding + reuse mem even though we reset the mem here. 
e.g., If se smaller shape than before, mem will be reused => padding will not be reset. Which means we need to realloc & reset mem always for padded layers.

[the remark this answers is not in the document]
yeonbok: @vladimir-paramuzov  Can I use "NeedPaddedInput" of convolution kernel for this purpose?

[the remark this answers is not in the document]
yeonbok: @vladimir-paramuzov  Please review this commiit added for the comments : 
https://github.com/openvinotoolkit/openvino/pull/16351/commits/64c237a7514c49629d5c0f2d0cb8cab3bdc1d29a

src/plugins/intel_gpu/src/graph/include/deconvolution_inst.h
```diff
@@ -60,6 +60,23 @@ class typed_primitive_inst<deconvolution> : public typed_primitive_inst_base<dec
     static layout calc_output_layout(deconvolution_node const& node, kernel_impl_params const& impl_param);
     static std::string to_string(deconvolution_node const& node);
 
+    bool need_reset_input_memory() const override {
+        auto input_layout = _deps[0].first->_impl_params->get_output_layout(0);
```
sshlyapn, line 64: Do we really need this for deconvolution? As I can see, only convolution and binary_convolution use physical paddings for calculation, isn't?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -238,40 +238,43 @@ void primitive_inst::update_shape() {
...
-        return;
+        return ev;
 
     bool can_reuse_buffer = _outputs[0] && actual_layout.count() <= max_output_layout_size;
 
     if (can_reuse_buffer) {
         GPU_DEBUG_TRACE_DETAIL << id() << ": reuse previously allocated output buffer" << std::endl;
         _outputs[0] = _network.get_engine().reinterpret_buffer(*_outputs[0], actual_layout);
+        if (need_reset_output_memory()) {
+            ev = _outputs[0]->fill(_network.get_stream());
```
sshlyapn, line 261: It looks like for USM buffers fill() method has a blocking call, did you try to change this behavior? Maybe it'll improve performance a little

src/plugins/intel_gpu/src/graph/include/convolution_inst.h
```diff
@@ -127,6 +127,23 @@ class typed_primitive_inst<convolution> : public typed_primitive_inst_base<convo
     static layout calc_output_layout(convolution_node const& node, kernel_impl_params const& impl_param);
     static std::string to_string(convolution_node const& node);
 
+    bool need_reset_input_memory() const override {
+        auto input_layout = _deps[0].first->_impl_params->get_output_layout(0);
+        if (input_layout.data_padding.lower_size() != tensor(0) || input_layout.data_padding.upper_size() != tensor(0)) {
```
vladimir-paramuzov, line 132: We have `operator bool()` defined for data padding, so this can be simplified

src/plugins/intel_gpu/tests/test_cases/empty_tensor_gpu_test.cpp
```diff
@@ -51,6 +51,7 @@ TEST_P(test_empty_tensor, concat_two_inputs) {
 
     std::vector<int32_t> nonzero_input_with_all_zero(p.nonzero_input_layout.count());
     std::fill(nonzero_input_with_all_zero.begin(), nonzero_input_with_all_zero.end(), 0);
+    get_test_stream().finish();
```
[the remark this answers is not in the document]
yeonbok: oh here is mistake. Will remove.

src/plugins/intel_gpu/src/graph/include/deconvolution_inst.h
yeonbok: I am not sure exact kernel impl of deconv, but if there is predefined output size, it may need. We can optimize out this afterwards if it is known to be not needed.

src/plugins/intel_gpu/src/graph/include/convolution_inst.h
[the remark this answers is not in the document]
yeonbok: Applied

yeonbok: Applied

vladimir-paramuzov (approved):

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
sshlyapn: I meant forcing of blocking call `const bool blocking = true;` for `gpu_usm::fill()` function

yeonbok: I see, I didnt recognized that it was set as blocking call. Now I am curious that why dont we get rid of that blocking true from the fill and  let the sync using the dependencies? hmm...  I will investigate the performance regarding those points in the upcoming works. Thanks for letting me know.

sshlyapn: @yeonbok, there were some problems related to synchronization, if I remember correctly. But I think we can try to disable it now

PR #16359, opened 2023-03-17
[CPU][TESTS] Use run_on_model instead of run_on_function

EgorDuplensky:


dmitry-gorokhov (approved):

EgorDuplensky:
@dmitry-gorokhov I think we can force-merge it after the build is verified

PR #16362, opened 2023-03-17
[PT FE]: support mixed precision in floor divide

eaidova:
### Details:
 - *ifix aten::floor_divide behaviour for mixed precision*

### Tickets:
 - *CVS-106323*

### DIFF
--- a/src/frontends/pytorch/src/op/floordiv.cpp
+++ b/src/frontends/pytorch/src/op/floordiv.cpp
@@ -15,7 +15,6 @@ OutputVector translate_floordiv(NodeContext& context) {
     num_inputs_check(context, 2, 2);
     auto x = context.get_input(0);
     auto y = context.get_input(1);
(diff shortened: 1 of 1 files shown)


src/frontends/pytorch/src/op/floordiv.cpp
```diff
@@ -15,6 +15,7 @@ OutputVector translate_floordiv(NodeContext& context) {
     num_inputs_check(context, 2, 2);
     auto x = context.get_input(0);
     auto y = context.get_input(1);
+    align_eltwise_input_types(context, x, y, true);
```
mvafin, line 18: Since we don't have test for this operation I think we better not change it, because we do not know if this is right change. Maybe we even need to remove this operation. It was needed before when we were experimenting with older pytorch, not clear how we can test it. So do not change for now.

PR #16371, opened 2023-03-17
[GNA] Enable GNA 3.5 part 1 - Fix crash for SWFP32 mode

marcinkusm:
### Details:
* add enforcing of Pooling2D for FP32
* fixed Unit Tests for GNA 3.5 version where Conv1D is converted to Conv2D
* set GNA 3.5 as default target

### Tickets:
 - *105916*


src/plugins/intel_gna/tests/unit/gna_extra_pwl_segments_tests.cpp
```diff
@@ -166,15 +166,16 @@ const std::vector<size_t> kPadEnd1D = {0, 0};
...
-const PWLExtraSegmentsParamsWithConv kConvolution1DReluWithoutPoolParams = {kConvolutionParams1D,
-                                                                            kPrecision32,
-                                                                            ngraph::helpers::ActivationTypes::Relu,
-                                                                            2,
-                                                                            false};
+const PWLExtraSegmentsParamsWithConv kConvolution1DReluWithoutPoolParams = {
+    kConvolutionParams1D,
+    kPrecision32,
+    ngraph::helpers::ActivationTypes::Relu,
+    4,  // For GNA_3_5 Conv1d is treated as Conv2D
```
tadamowicz, line 173: it can't be fixed becase it depends on compile target, so fo current branch (defaut targe is gna 3.0) it will fail

src/plugins/intel_gna/src/common/gna_target.hpp
```diff
@@ -24,7 +24,7 @@ enum class DeviceVersion {
     GNAEmbedded3_5 = 0x35e,
     GNA3_6 = 0x36e,
     GNA4_0 = 0x40e,
-    Default = GNA3_0
+    Default = GNA3_5
```
tadamowicz, line 27: revert this

PR #16372, opened 2023-03-17
[CPU] Use Dnnl executor to avoid extra dnnl primitve desc query

maxnick:
### Details:
Since v3.1 it was observed that early inefficient dnnl::primitive_desc query became even slower. It was decided to store input, output and scratchpad memory descriptors in the DnnlExecutor to avoid dnnl::primitive_desc query calls at runtime.

### Tickets:
 - 99126

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/conv.cpp
+++ b/src/plugins/intel_cpu/src/nodes/conv.cpp
@@ -1495,8 +1495,7 @@ void Convolution::prepareParams() {
 
         Node::appendPostOpArgs(*pAttrLocal, primArgs, convPostOpsArgs[preferLegacyPostOps]);
 
-        auto pd = execPtr->getPrimitiveDesc();
-        auto scratchpadMem = getScratchPadMem(pd);
+        auto scratchpadMem = getScratchPadMem(execPtr->getScratchPadDesc());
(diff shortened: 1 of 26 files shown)


src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1513,9 +1512,7 @@ Convolution::ConvolutionExecutor::ConvolutionExecutor(const dnnl::convolution_fo
                                                                 const dnnl::memory::desc& inMemDesc,
                                                                 const dnnl::memory::desc& weightMemDesc,
                                                                 const dnnl::memory::desc& outMemDesc,
-                                                                const dnnl::engine& engine) {
-    execPrim = dnnl::convolution_forward(pd);
-
+                                                                const dnnl::engine& engine) : DnnlExecutor(pd) {
     if (inMemDesc != pd.src_desc()) {
```
usstq, line 1516: should we change it to `getSrcDesc()` since `pd.src_desc()` is slow?
also `pd.weights_desc()` and `pd.dst_desc()` in following lines

maxnick: Good point. Changed accordingly. Thank you!

maxnick:

dmitry-gorokhov (approved):

PR #16394, opened 2023-03-20
[MO] remove deprecated: data_type, disable_nhwc_to_nchw, tensorflow_use_custom_operations_config

pavel-esir:
### Details:
 - `data_type` is deprecated `compress_to_fp16` arguments should be used instead
 - `disable_nhwc_to_nchw` is deprecated, automatic layout detection is used instead
 - `tensorflow_use_custom_operations_config` is deprecated, `transformations_config` should be used instead

### Tickets:
 - xxx-100890


pavel-esir:
Some of the mo_argtests still use data_type. Also some layer tests fail because of eltwise comparison mismatch: probably because for fp16 uses lower threshold which intended for fp32, i'll take a careful look

rkazants (approved):

rkazants:
@pavel-esir, has some conflicts with tests. Please check and recover this PR.

PR #16398, opened 2023-03-20
Fix MO Reader for Squeeze without axes

mvafin:
### Details:
 - *Legacy MO never had Squeeze without second input, while frontend can generate such Squeeze. Fix MO Reader to support this case.*

### Tickets:
 - *104158*

### DIFF
--- a/tools/mo/openvino/tools/mo/utils/ir_reader/internal_ops/squeeze.py
+++ b/tools/mo/openvino/tools/mo/utils/ir_reader/internal_ops/squeeze.py
@@ -21,5 +21,5 @@ class SqueezeInternal(Squeeze):
             assert is_fully_defined(
                 input_shape), 'Squeeze dimensions are not defined for op "{}"'.format(node_name)
             output_shape = [s for s in shape_array(input_shape).tolist() if s != 1]
-            node.out_port(0).data.set_shape(output_shape)
+            node.out_port(0).data.set_shape(shape_array(output_shape))
 
--- a/tools/mo/openvino/tools/mo/utils/ir_reader/internal_ops/squeeze.py
+++ b/tools/mo/openvino/tools/mo/utils/ir_reader/internal_ops/squeeze.py
@@ -3,12 +3,23 @@
 
 from openvino.tools.mo.graph.graph import Node
 from openvino.tools.mo.ops.squeeze import Squeeze
+from openvino.tools.mo.front.common.partial_infer.utils import shape_array, dynamic_dimension, shape_delete, is_fully_defined, \
+    undefined_shape_of_rank
 
 
 class SqueezeInternal(Squeeze):
     @staticmethod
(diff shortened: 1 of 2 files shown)


tools/mo/openvino/tools/mo/utils/ir_reader/internal_ops/squeeze.py
```diff
@@ -3,12 +3,23 @@
...
+            # preserve initial axis value
+            node.in_port(1).data.set_value(axis_value)
+        else:
+            # Squeeze without axes provided
+            node_name = node.soft_get('name', node.id)
+            input_shape = node.in_port(0).data.get_shape()
+            assert is_fully_defined(
+                input_shape), 'Squeeze dimensions are not defined for op "{}"'.format(node_name)
+            output_shape = [s for s in shape_array(input_shape).tolist() if s != 1]
+            node.out_port(0).data.set_shape(output_shape)
```
mvafin, line 24:
```suggestion
            node.out_port(0).data.set_shape(shape_array(output_shape))
```

rkazants (approved):

tools/mo/openvino/tools/mo/utils/ir_reader/internal_ops/squeeze.py
```diff
@@ -3,12 +3,23 @@
...
-        Squeeze.infer(node)
-        # preserve initial axis value
-        node.in_port(1).data.set_value(axis_value)
+        if node.is_in_port_connected(1):
+            axis_value = node.in_port(1).data.get_value()
+            Squeeze.infer(node)
+            # preserve initial axis value
+            node.in_port(1).data.set_value(axis_value)
+        else:
+            # Squeeze without axes provided
```
rkazants, line 18: Am I correct that TF FE is a source of such IRs with `Squeeze` nodes without axis? And we need to avoid such `Squeeze` nodes. What is downside of such IRs? May be we need to introduce new `Squeeze` specification with this limitation?

mvafin: Downside of such IRs is in dynamic cases. If input shape is not known the output rank is not known.

rkazants: May be we should consider to have the second input as mandatory?

mvafin: I don't think that would be a good idea. I suppose it was introduced as optional, because some frameworks might have it as optional. So to comply with such frameworks we better have it optional too. But avoid this in frontends if possible

rkazants: I see that ONNX needs such feature.

PR #16404, opened 2023-03-20
Interpolate op bugfix follow-up

tomdol:
### Details:
 - Follow up of the fix from this PR https://github.com/openvinotoolkit/openvino/pull/16365

### Tickets:
 - 106294

### DIFF
--- a/src/core/tests/visitors/op/interpolate.cpp
+++ b/src/core/tests/visitors/op/interpolate.cpp
@@ -81,8 +81,7 @@ TEST(attributes, interpolate_op4) {
 TEST(attributes, interpolate_op11) {
     NodeBuilder::get_ops().register_factory<opset11::Interpolate>();
     const auto img = make_shared<op::Parameter>(element::f32, Shape{1, 3, 32, 32});
-    const auto scales = op::v0::Constant::create(element::f32, {2}, {2.0, 2.0});
-    const auto axes = op::v0::Constant::create(element::i32, {2}, {2, 3});
+    const auto scales = op::v0::Constant::create(element::f32, {4}, {1.0, 1.0, 2.0, 2.0});
 
     op::v11::Interpolate::InterpolateAttrs attrs;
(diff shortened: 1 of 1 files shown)


praasz (approved):

src/core/tests/visitors/op/interpolate.cpp
```diff
@@ -81,8 +81,7 @@ TEST(attributes, interpolate_op4) {
 TEST(attributes, interpolate_op11) {
     NodeBuilder::get_ops().register_factory<opset11::Interpolate>();
     const auto img = make_shared<op::Parameter>(element::f32, Shape{1, 3, 32, 32});
-    const auto scales = op::v0::Constant::create(element::f32, {2}, {2.0, 2.0});
-    const auto axes = op::v0::Constant::create(element::i32, {2}, {2, 3});
+    const auto scales = op::v0::Constant::create(element::f32, {4}, {1.0, 1.0, 2.0, 2.0});
```
mitruska, line 84: Why not keep the `axes` input provided or both test cases if it makes difference?

Also originally this test was supposed to fail, because of `scales` size smaller than the input rank,
such cases of inputs shapes and count should be primarily tested by interpolate "type_prop" tests: https://github.com/openvinotoolkit/openvino/blob/master/src/core/tests/type_prop/interpolate.cpp

tomdol: The test was not supposed to fail, it's the attributes visitor test and the op should be created successfully in order to test that. There's a spot in the shape inference code that needs to be tweaked in order to disallow the behavior which sometimes occurs when the axes input is not provided. This will be done in a separate PR.

tomdol: Will get this fixed in a separate PR. This test can now stay as it verifies the attribute visitor behavior, the bug that will be fixed (and additionally tested) later is located in the shape inference code.

PR #16412, opened 2023-03-20
[GPU] Unique-10 operation implementation.

mhnap-lohika:
### Details:
 - *Implementation of new Unique-10 operation in GPU plugin*

### Tickets:
 - *CVS-94709*


wilson-seok:
@mhnap-lohika Please resolve gpu func test failures.
image

yeonbok:
BTW, why didnt you do like what is done in nonzero? 
In nonzero, we intentionally do count nonzero first so that we could use only required memory. 
The internal kernel optimization can be done separately, but the basic primitive construction should be consistent for all kernels. 
In the future we'll need to optimize the memory not to use all the input shape, then we'll need reimplement this....

src/tests/functional/plugin/gpu/single_layer_tests/dynamic/unique.cpp
```diff
@@ -0,0 +1,170 @@
...
+                int32_t range = std::accumulate(targetInputStaticShapes[0].begin(),
+                                                targetInputStaticShapes[0].end(),
+                                                1,
+                                                std::multiplies<size_t>());
+                tensor = utils::create_and_fill_tensor(funcInput.get_element_type(),
+                                                       targetInputStaticShapes[0],
+                                                       range,
+                                                       -range / 2,
+                                                       1);
+            }
```
yeonbok, line 114: I think using only random data for this kind of operation is not sufficient. Could you please add a gpu unittest with some duplicated data? Or could you add such data with dupilication here?

mhnap-lohika: I have copied the same generation function from CPU tests, but I agree that it is better to have a deterministic test for such an operation. I will add not random unit tests.

mhnap-lohika:
> BTW, why didnt you do like what is done in nonzero? In nonzero, we intentionally do count nonzero first so that we could use only required memory. The internal kernel optimization can be done separately, but the basic primitive construction should be consistent for all kernels. In the future we'll need to optimize the memory not to use all the input shape, then we'll need reimplement this....

Hmm, I don't fully understand why this implementation differs from nonzero. As I saw, nonzero is also divided into two separate kernels and the second shape is get from the first kernel run in "calc_output_layouts". I have implemented similar logic here. The difference is that I don't do parallel algorithm for now, cause it's not so simple for this operation and requires more time, and as I see, reference and CPU implementations do the same way. I have thoughts that it would make sense to parallelize by batch, but the documentation not stated that we can use batch for such a purpose, so I leave the same logic. Please correct me if I misunderstood something.

mhnap-lohika:
> > BTW, why didnt you do like what is done in nonzero? In nonzero, we intentionally do count nonzero first so that we could use only required memory. The internal kernel optimization can be done separately, but the basic primitive construction should be consistent for all kernels. In the future we'll need to optimize the memory not to use all the input shape, then we'll need reimplement this....
> 
> Hmm, I don't fully understand why this implementation differs from nonzero. As I saw, nonzero is also divided into two separate kernels and the second shape is get from the first kernel run in "calc_output_layouts". I have implemented similar logic here. The difference is that I don't do parallel algorithm for now, cause it's not so simple for this operation and requires more time, and as I see, reference and CPU implementations do the same way. I have thoughts that it would make sense to parallelize by batch, but the documentation not stated that we can use batch for such a purpose, so I leave the same logic. Please correct me if I misunderstood something.

Oh, I think I got what you meant. So the problem is that I allocate the maximum shape for all outputs in the first kernel. And then allocate the correct ones in the second kernel. But nonzero use only one scalar to save the nonzero count in the first kernel and then correct shape in the second, so there is less memory used. I see.. 
I can do the same, but it would require running almost the same algorithm in the first kernel but only saving unique count and do not save other outputs. And then run the whole algorithm in the second kernel. So I thought it would be worse on performance. And also we can have "sorted" attribute and in such case I still would need an additional buffer to sort all data. 
But yeah, this is possible to do it in such a way, just not sure that is better. Interesting to hear more thoughts from you, thank you!

wilson-seok:
@mhnap-lohika Unique-10 functional test has still failures. Please check it.

src/tests/functional/plugin/gpu/single_layer_tests/dynamic/unique.cpp
mhnap-lohika: Added unit tests

yeonbok:
>> I can do the same, but it would require running almost the same algorithm in the first kernel but only saving unique count and do not save other outputs. And then run the whole algorithm in the second kernel. So I thought it would be worse on performance. And also, we still would need an additional buffer to run algorithms and find unique count.

Actually in the count non zero it only need to count, which can be paralallezed. 
Also, it is just a memory read operation, which is same overhead as reshape. 
However the strategy of nonzero has further effect for reducing the memory.

mhnap-lohika:
Hello @yeonbok !

I have implemented the new approach that you proposed.

Now, in the first kernel, I use only 1 output with a scalar shape equal to 1, as we have in the nonzero op, instead of 5 outputs as it was before. In the second kernel, we still have 4 outputs, as our documentation states, but now we have only 2 inputs instead of 5 as it was previously.

But note that we still need to create an additional buffer to save intermediate algorithm results with the same shape as input in the first kernel. You can find this code [here](https://github.com/openvinotoolkit/openvino/pull/16412/commits/b16bdf862fb04d84c4a24b403008ff5709cb0970#diff-8a8ad046a12d39530496d36ec1c040e3a0f3064547d7dfbdef413d26c0dc2a6aR166).

So, to sum up, we reduced memory only for 3 outputs (and inputs accordingly) out_indices, out_rev_indices, and out_counts.

These changes will be more noticeable for the flattened case than for the axis case. As in the flattened case, all three outputs have the same shape as the input. But in the axis case, these three outputs have shapes equal only to one axis.

Also worth noticing that the whole algorithm complexity grows up a lot. Previously, we ran the algorithm once in the first kernel and copied all outputs in the second one. Now, we need to run it in both the first and second kernels.

Moreover, there have added more additional complexity on the second kernel [here](https://github.com/openvinotoolkit/openvino/pull/16412/commits/b16bdf862fb04d84c4a24b403008ff5709cb0970#diff-5aa18d52b7b3ce6ea99728ffd45b62280175c85dbde48a049fdb35db5692ee18R176). As now out_unique_elements output have smaller shape than input, and we cannot save sorted results there and run "deduplicate" algorithm as it was before. So, we first found all unique elements by running the "unique" algorithm and only after sort all data. But, after sorting, we need to adjust the out_rev_indices output to have correct values. We could change this behavior to the previous one by creating one more additional buffer that would have an input shape and would contain sorted data. But as it would increase memory and we tried to reduce it, I didn't implement it in what way.

Also, I have changed only the logic to work correctly, so we can decide whether to keep the current implementation or revert to the previous one. But, if we agreed to keep the current implementation, I would need to refactor the code to change namings and comments.

What you think about this?

yeonbok:
Hi @mhnap-lohika, Thanks for the investigation. I think we can do some optimization for the intermeidate buffer w.r.t the total input size in the future, so that it does not need global internal memory but to utilize local memory (if it fits).

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/unique_ref.cl
```diff
@@ -0,0 +1,65 @@
+// Copyright (C) 2023 Intel Corporation
```
yeonbok, line 1: random spot: 
Please let the kernel names and file names to be aligned.

mhnap-lohika:
Hi @yeonbok! So should we keep the current more memory-safe approach? If yes, I will rename all usage of the "unique" kernel to "unique_count" and "unique_reshape" to "unique_gather".

mhnap-lohika:
@wilson-seok @yeonbok There is some strange bug when the input shape rank is greater than 4 for the axis case. I have managed to fix this behavior for the flattened case in [this commit](https://github.com/openvinotoolkit/openvino/pull/16412/commits/236d01b3695c41fa115e860ca5d3a3e6c0f51809). The problem is that for some reason I got crashes or std::bad_alloc exceptions during "reorder_inputs" pass. I notice that this problem goes away if I set output_format for the first output equal to "bfyx", but not "bfzyx" even if the input layout has "bfzyx" format. But then came other problems, the kernel cannot compile because input format != output format, and many defines rely on what logic. I tried manually fix the output format in ocl to be equal to the input format, and after this change kernel compiles, but for some reason first output seems not to have all the needed data after the kernels run. I debug actual data in the cl kernel, and all data are correctly set there, but still differ during comparing with reference results in tests. I added [this commit](https://github.com/openvinotoolkit/openvino/pull/16412/commits/236640d2f0e9d5b1f8dcbbf9482763badd7fde66) so you can see that it helps a bit but is still insufficient to fix this issue fully. Maybe there is some bug for dynamic shape flow when the rank is greater than 4, or maybe there is some bug in my "calc_output_layouts" code, I'm not sure. Maybe you will have some ideas. For now I added those tests to skip config.

yeonbok:
> @wilson-seok @yeonbok There is some strange bug when the input shape rank is greater than 4 for the axis case. I have managed to fix this behavior for the flattened case in [this commit](https://github.com/openvinotoolkit/openvino/pull/16412/commits/236d01b3695c41fa115e860ca5d3a3e6c0f51809). The problem is that for some reason I got crashes or std::bad_alloc exceptions during "reorder_inputs" pass. I notice that this problem goes away if I set output_format for the first output equal to "bfyx", but not "bfzyx" even if the input layout has "bfzyx" format. But then came other problems, the kernel cannot compile because input format != output format, and many defines rely on what logic. I tried manually fix the output format in ocl to be equal to the input format, and after this change kernel compiles, but for some reason first output seems not to have all the needed data after the kernels run. I debug actual data in the cl kernel, and all data are correctly set there, but still differ during comparing with reference results in tests. I added [this commit](https://github.com/openvinotoolkit/openvino/pull/16412/commits/236640d2f0e9d5b1f8dcbbf9482763badd7fde66) so you can see that it helps a bit but is still insufficient to fix this issue fully. Maybe there is some bug for dynamic shape flow when the rank is greater than 4, or maybe there is some bug in my "calc_output_layouts" code, I'm not sure. Maybe you will have some ideas. For now I added those tests to skip config.

Hi @mhnap-lohika Appologies for the belated response. Your comment is a bit confusing and currently the linked commits are not existing :) Could you please give me some reproducer of the initial issue you experienced so that I could check from my side? 
(Also about the statement: " the kernel cannot compile because input format != output format, and many defines rely on what logic" It is strange because many kernels are supporting different in/out formats.)

mhnap-lohika:
> > @wilson-seok @yeonbok There is some strange bug when the input shape rank is greater than 4 for the axis case. I have managed to fix this behavior for the flattened case in [this commit](https://github.com/openvinotoolkit/openvino/pull/16412/commits/236d01b3695c41fa115e860ca5d3a3e6c0f51809). The problem is that for some reason I got crashes or std::bad_alloc exceptions during "reorder_inputs" pass. I notice that this problem goes away if I set output_format for the first output equal to "bfyx", but not "bfzyx" even if the input layout has "bfzyx" format. But then came other problems, the kernel cannot compile because input format != output format, and many defines rely on what logic. I tried manually fix the output format in ocl to be equal to the input format, and after this change kernel compiles, but for some reason first output seems not to have all the needed data after the kernels run. I debug actual data in the cl kernel, and all data are correctly set there, but still differ during comparing with reference results in tests. I added [this commit](https://github.com/openvinotoolkit/openvino/pull/16412/commits/236640d2f0e9d5b1f8dcbbf9482763badd7fde66) so you can see that it helps a bit but is still insufficient to fix this issue fully. Maybe there is some bug for dynamic shape flow when the rank is greater than 4, or maybe there is some bug in my "calc_output_layouts" code, I'm not sure. Maybe you will have some ideas. For now I added those tests to skip config.
> 
> Hi @mhnap-lohika Appologies for the belated response. Your comment is a bit confusing and currently the linked commits are not existing :) Could you please give me some reproducer of the initial issue you experienced so that I could check from my side? (Also about the statement: " the kernel cannot compile because input format != output format, and many defines rely on what logic" It is strange because many kernels are supporting different in/out formats.)

Hi @yeonbok! Sorry, I forgot that commit links expire when I rebase PR. You can easily reproduce this issue that I talked about by reverting [Added some tests to skip config.](https://github.com/openvinotoolkit/openvino/pull/16412/commits/4bb62e303308a2ada136adc33bfc95a1fdbd66b2) commit.

yeonbok:
>> Hi @yeonbok! Sorry, I forgot that commit links expire when I rebase PR. You can easily reproduce this issue that I talked about by reverting [Added some tests to skip config.](https://github.com/openvinotoolkit/openvino/pull/16412/commits/4bb62e303308a2ada136adc33bfc95a1fdbd66b2) commit.

Hi, I checked the root cause and which was due to some missing internal behavior for multiple output. 
1) reorder input tries to add reorder b/w unique node w/ multiple output result and result node 
2) Here, original result node had dep to multiple output with dep_idx
3) However add_intermediate function for this functionality was not handling the original dep_idx. So all of the added reorder node were point to the first output of the unique node while the original result node keeps to find n-th output of the newly added reorder. 
4) Let me figure out how to fix this.

yeonbok:
@mhnap-lohika  I made a fix for the initial issue of build failure:
https://github.com/yeonbok/openvino/commit/7924fa4b3e01515b14c9d6bebf856d4182d3fcf7

You can cherry pick above change. 

However after this fix, the func test result was not passing as follows: 
Could you please check?

abs_max < abs_threshold && rel_max < rel_threshold
         abs_max: 6
                 coordinate 1; abs errors count 5; abs mean 2.85714; abs threshold 1
         rel_max: 6
                 coordinate 1; rel errors count 0; rel mean 1.72619; rel threshold 1.79769e+308
Reset PluginCache. OV Core use count: 1
[  FAILED  ] smoke_static/UniqueLayerDynamicGPUTest.CompareWithRefs/IS=([])_TS={(7.2.3.5.4)}_axis=0_sorted=True_dataPrc=f16, where GetParam() = ({ ({}, { { 7, 2, 3, 5, 4 } }) }, (false, 0), true, 4-byte object <04-00 00-00>) (115 ms)

mhnap-lohika:
> @mhnap-lohika I made a fix for the initial issue of build failure: [yeonbok@7924fa4](https://github.com/yeonbok/openvino/commit/7924fa4b3e01515b14c9d6bebf856d4182d3fcf7)
> 
> You can cherry pick above change.
> 
> However after this fix, the func test result was not passing as follows: Could you please check?
> 
> abs_max < abs_threshold && rel_max < rel_threshold abs_max: 6 coordinate 1; abs errors count 5; abs mean 2.85714; abs threshold 1 rel_max: 6 coordinate 1; rel errors count 0; rel mean 1.72619; rel threshold 1.79769e+308 Reset PluginCache. OV Core use count: 1 [ FAILED ] smoke_static/UniqueLayerDynamicGPUTest.CompareWithRefs/IS=([])_TS={(7.2.3.5.4)}_axis=0_sorted=True_dataPrc=f16, where GetParam() = ({ ({}, { { 7, 2, 3, 5, 4 } }) }, (false, 0), true, 4-byte object <04-00 00-00>) (115 ms)

@yeonbok Thank you! Your fix seems to be working. I now don't get the std::bad_alloc exception. But yes, now some tests are failing. I have checked what data are set in the kernel and what data we compare, seems that there are still some bugs with multiple outputs. Because I see that in the kernel, we set the correct data, but the data is not correct in comparing results. Moreover seems like data for the 3rd and 4th outputs are just copied from the 2nd output. Here is a small example:

Second output expected and actual data:
i:0 ex:1 ad: 1
i:1 ex:6 ad: 6
i:2 ex:3 ad: 3
i:3 ex:2 ad: 2
i:4 ex:0 ad: 0
i:5 ex:4 ad: 4
i:6 ex:5 ad: 5
(correct)

Third output expected and actual data:
i:0 ex:4 ad: 1
i:1 ex:0 ad: 6
i:2 ex:3 ad: 3
i:3 ex:2 ad: 2
i:4 ex:5 ad: 0
i:5 ex:6 ad: 4
i:6 ex:1 ad: 5
(incorrect) 


And one more example:
Second output expected and actual data:
i:0 ex:0 ad: 0
i:1 ex:1 ad: 1
i:2 ex:2 ad: 2
i:3 ex:3 ad: 3
i:4 ex:4 ad: 4
i:5 ex:5 ad: 5
i:6 ex:6 ad: 6
(correct)

Third output expected and actual data:
i:0 ex:0 ad: 0
i:1 ex:1 ad: 1
i:2 ex:2 ad: 2
i:3 ex:3 ad: 3
i:4 ex:4 ad: 4
i:5 ex:5 ad: 5
i:6 ex:6 ad: 6
(correct)

Fourth output expected and actual data:
i:0 ex:1 ad: 0
i:1 ex:1 ad: 1
i:2 ex:1 ad: 2
i:3 ex:1 ad: 3
i:4 ex:1 ad: 4
i:5 ex:1 ad: 5
i:6 ex:1 ad: 6
(incorrect) 

But in kernel, I see such logs:
out_counts[0]=1
out_counts[1]=1
out_counts[2]=1
out_counts[3]=1
out_counts[4]=1
out_counts[5]=1
out_counts[6]=1

yeonbok:
I got the reason.. the cached reorder is used regardless the dep_idx. Will provide additional fix soon.

yeonbok:
https://github.com/yeonbok/openvino/commit/8081318bdcdbd7ece6170e3492296fb9a3e20112 Additional fix : please cherry pick this.

mhnap-lohika:
@yeonbok Thank you! Yes, now all tests are successfully passed with your fixes. I have also pushed them in this PR.

yeonbok (approved):
LGTM but please make CI pass.

wilson-seok:
@mhnap-lohika We still have build error in openvino-win. Please check it.
image

And have you tried failed unit/functional test cases from ov CI on your local?

wilson-seok:
@mhnap-lohika Win build error still happens. And please check failed gpu unit/functional test on your local.

mhnap-lohika:
Hello @wilson-seok! I managed to reproduce unit test failures on my local machine. I notice that the issue comes from those two cherry-picked commits, because when I revert them, all tests are passing correctly. @yeonbok Could you please take a look at how we can fix these commits?

wilson-seok:
@mhnap-lohika Please rebase to get fix for current test failure.

PR #16422, opened 2023-03-21
[TF FE] Fix leftovers from code review

rkazants:
**Details:** Fix leftovers from code review

**Ticket:** TBD

### DIFF
--- a/src/frontends/tensorflow_common/src/op/strided_slice.cpp
+++ b/src/frontends/tensorflow_common/src/op/strided_slice.cpp
@@ -27,9 +27,10 @@ OutputVector translate_strided_slice_op(const NodeContext& node) {
             return vector<int64_t>{};
         }
         size_t max_length = sizeof(mask) * CHAR_BIT;
-        vector<int64_t> vec{};
+        vector<int64_t> vec;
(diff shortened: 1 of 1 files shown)


ilya-lavrenov (approved):

src/frontends/tensorflow_common/src/op/strided_slice.cpp
```diff
@@ -27,9 +27,10 @@ OutputVector translate_strided_slice_op(const NodeContext& node) {
             return vector<int64_t>{};
         }
         size_t max_length = sizeof(mask) * CHAR_BIT;
-        vector<int64_t> vec{};
+        vector<int64_t> vec;
```
itikhono, line 30: I believe you can allocate this vec
`vector<int64_t> vec(max_length, 0);`
and use it 
`vec[i] = 1;`
After this,  the resizing in line 49 is unnecessary ( sizeof(mask) * CHAR_BIT is equal for each mask)
For mask == 0:
`if (mask == 0) return vector<int64_t>(max_length, 0)`

itikhono (approved):

PR #16430, opened 2023-03-21
flush subnorms to zero by recreating constants

pavel-esir:
### Details:
 - In order to turn on mmap allocator https://github.com/openvinotoolkit/openvino/pull/12673 for constants read our from IR no in place changes for should be done. Therefore changes should be done by recreating constants. Flush to zero should be corrected for such recreating.
 - No significant changes in performance occur because most of the constants are normal and are not recreated.

### Tickets:
 - xxx-104573

### DIFF
--- a/src/common/transformations/src/transformations/flush_fp32_subnormals_to_zero.cpp
+++ b/src/common/transformations/src/transformations/flush_fp32_subnormals_to_zero.cpp
@@ -36,14 +36,28 @@ ov::pass::FlushFP32SubnormalsToZero::FlushFP32SubnormalsToZero() {
         bool has_subnormals = false;
         for (size_t i = 0; i < size; ++i) {
             if (fpclassify(std::abs(data[i])) == FP_SUBNORMAL) {
-                data[i] = 0.0f;
                 has_subnormals = true;
+                break;
             }
         }
-        if (has_subnormals)
-            return true;
+        if (!has_subnormals)
+            return false;
 
-        return false;
+        auto new_constant = std::make_shared<ov::opset8::Constant>(ov::element::f32, node->get_shape());
+        auto* dst_data = const_cast<float*>(new_constant->get_data_ptr<float>());
+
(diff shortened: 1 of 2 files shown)


vurusovs (approved):

ilya-lavrenov (approved):

src/common/transformations/src/transformations/flush_fp32_subnormals_to_zero.cpp
```diff
@@ -36,14 +36,28 @@ ov::pass::FlushFP32SubnormalsToZero::FlushFP32SubnormalsToZero() {
...
                 has_subnormals = true;
+                break;
             }
         }
-        if (has_subnormals)
-            return true;
+        if (!has_subnormals)
+            return false;
+
+        auto new_constant = std::make_shared<ov::opset8::Constant>(ov::element::f32, node->get_shape());
```
itikhono, line 46: I believe you need to use type of the old constant
auto new_constant = std::make_shared<ov::opset8::Constant>(old_constant->get_element_type(), node->get_shape());

itikhono: I see we have a check in the transformation
        if (node->get_output_element_type(0) != element::f32)
            return false;

src/common/transformations/src/transformations/flush_fp32_subnormals_to_zero.cpp
```diff
@@ -36,14 +36,28 @@ ov::pass::FlushFP32SubnormalsToZero::FlushFP32SubnormalsToZero() {
...
+        auto* dst_data = const_cast<float*>(new_constant->get_data_ptr<float>());
```
itikhono, line 47: it's ok for this PR, but it is not recommended to use `const_cast`
I believe it's better to create ov::Tensor and use the corresponding ctor of Constant

PR #16434, opened 2023-03-21
Add Conversion and Op Extension to Pytorch frontend

mateusztabaka:
Tickets: 98766 and 98767


src/frontends/common/include/openvino/frontend/node_context.hpp
```diff
@@ -48,6 +48,11 @@ class FRONTEND_API NodeContext {
         FRONT_END_NOT_IMPLEMENTED(get_input);
     }
 
+    /// \brief Returns const input with a given index
+    virtual Any const_input(int idx) const {
+        FRONT_END_NOT_IMPLEMENTED(const_input);
+    }
+
```
rkazants, line 55: not sure that we need this method in the common place. Is this really needed?

src/bindings/python/requirements_test.txt
```diff
@@ -35,6 +35,7 @@ pytest-html==1.19.0
 py>=1.9.0
 radon
 retrying
+torch
```
ilya-lavrenov, line 38: can we test extensions in frontend tests?
we want to keep python tests as framework independent as possible

src/frontends/common/include/openvino/frontend/node_context.hpp
mateusztabaka: Code in frontend/common that handles OpExtension refers to base NodeContext class https://github.com/openvinotoolkit/openvino/pull/16434/files#diff-a441f7a33d0e530dc1b19988cda0dd386a5cde256bf34871dc30217b9de8ba10R243. That's why I thought it would be good to have it here.

src/bindings/python/requirements_test.txt
mateusztabaka: I can see that we have frontend tests for onnx, tf, paddle in this directory so I thought it'd be a good place. Also the tests I did serve as a python bindings tests for pytorch frontend.

ilya-lavrenov: but we don't have dependency on TF, PADDLE.
If you managed to write tests without adding torch dependency, I'm totally OK to keep it here.

The idea to avoid more and more dependencies because of specific use cases.

mateusztabaka: ok, I moved those tests of out src/bindings/python/tests.

src/frontends/pytorch/include/openvino/frontend/pytorch/node_context.hpp
```diff
@@ -54,6 +54,8 @@ class NodeContext : public frontend::NodeContext {
         return m_tensor_map->at(input);
     }
 
+    Any const_input(int index) const override;
```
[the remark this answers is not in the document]
mateusztabaka: This version returns ov::Any so it can be used by visitor in common frontend extension code:
https://github.com/openvinotoolkit/openvino/pull/16434/files#diff-a441f7a33d0e530dc1b19988cda0dd386a5cde256bf34871dc30217b9de8ba10R243

It's analogous to NodeContext::get_attribute_as_any.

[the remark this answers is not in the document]
mvafin: cant we use `const_input<Any>(index)` instead of creating a new method? Or maybe we can have a different name for it? e.g. `const_input_as_any`?

jiwaszki (approved):
Looks OK on Python side.

src/frontends/pytorch/include/openvino/frontend/pytorch/node_context.hpp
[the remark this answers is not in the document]
mateusztabaka: const_input_as_any looks good to me. I renamed that function.

src/frontends/common/include/openvino/frontend/node_context.hpp
```diff
@@ -48,6 +48,11 @@ class FRONTEND_API NodeContext {
         FRONT_END_NOT_IMPLEMENTED(get_input);
     }
 
+    /// \brief Returns const input with a given index
```
itikhono, line 51: /// \brief Returns values from Constant input with the given index as ov::Any. Throws an exception if the input cannot be represented as Constant.

src/frontends/pytorch/include/openvino/frontend/pytorch/node_context.hpp
```diff
@@ -54,6 +54,8 @@ class NodeContext : public frontend::NodeContext {
         return m_tensor_map->at(input);
     }
 
+    Any const_input_as_any(int index) const override;
```
itikhono, line 57: should we use size_t?

src/frontends/pytorch/src/node_context.cpp
```diff
@@ -180,6 +180,55 @@ std::string NodeContext::const_input<std::string>(size_t index) const {
...
+    const T* ptr = reinterpret_cast<const T*>(constant->get_data_ptr());
+    const auto& shape = constant->get_shape();
+    if (is_scalar(shape)) {
+        return ptr[0];
+    }
+    return std::vector<T>(ptr, ptr + shape_size(shape));
+}
+}  // namespace
+
+Any NodeContext::const_input_as_any(int index) const {
```
itikhono, line 195: @mvafin could you take a look?
I'm not sure that we actually need this method.
There is a data field in python representation of ov::Constant which can be used instead of this
https://github.com/openvinotoolkit/openvino/blob/5e835e327b235b17e605de7f18eaf49a7c806bc7/src/bindings/python/src/pyopenvino/graph/ops/constant.cpp#L193

src/bindings/python/src/pyopenvino/frontend/pytorch/extension.cpp
```diff
@@ -0,0 +1,78 @@
...
+            m,
+            "OpExtensionPytorch",
+            py::dynamic_attr());
+
+    ext.def(py::init([](const std::string& fw_type_name,
+                        const std::map<std::string, size_t>& attr_names_map,
+                        const std::map<std::string, py::object>& attr_values_map) {
+                std::map<std::string, ov::Any> any_map;
+                for (const auto& it : attr_values_map) {
+                    any_map[it.first] = Common::utils::py_object_to_any(it.second);
```
itikhono, line 57: Is it correct to throw exception in the ctor in python?
py_object_to_any can throw exception in case of the py_object can not be representes as c++ type

src/frontends/common/include/openvino/frontend/extension/op.hpp
```diff
@@ -198,6 +218,85 @@ class OpConversionFunction {
...
+                                      const std::map<std::string, size_t>& attr_names_map = {},
+                                      const std::map<std::string, ov::Any>& attr_values_map = {})
+        : m_context(context),
+          m_attr_names_map(attr_names_map),
+          m_attr_values_map(attr_values_map) {}
+
+    void on_adapter(const std::string& name, ValueAccessor<void>& adapter) override {
+        auto p_value = m_attr_values_map.find(name);
+
+        if (p_value != m_attr_values_map.end()) {
```
itikhono, line 233: it looks like the logic is incorrect
the value will override the mapping, I'm fixing this in PR
file: https://github.com/openvinotoolkit/openvino/pull/15294/ 
src/frontends/common/include/openvino/frontend/extension/op.hpp

we agreed with Sergey L. that if both the value and name mapping are provided, this is an issue and the behavior is undefined

tests/layer_tests/py_frontend_tests/test_torch_frontend.py
mvafin, line 1: Please add test of operation that have `string` input and `None` input.

src/frontends/pytorch/src/node_context.cpp
mvafin: I think we need this method to be able to get values of inputs, especially if those inputs are strings or nones. I would recommend rename it to the generic name that can be used in other frontends too, for example `get_input_const_value`, @itikhono do you agree?

itikhono: it looks like we can't re-use Constant class functionality.
`get_values_from_const_input` maybe?

tests/layer_tests/py_frontend_tests/test_torch_frontend.py
mateusztabaka: done

src/frontends/common/include/openvino/frontend/extension/op.hpp
mateusztabaka: done

src/frontends/pytorch/src/node_context.cpp
mateusztabaka: done

src/frontends/common/include/openvino/frontend/node_context.hpp
mateusztabaka: done

itikhono (approved):

mvafin (approved):

slyalin:
@andrei-kochin, there is no any tests for `OpExtension` and looks like it doesn't work in some trivial scenarios.

PR #16435, opened 2023-03-21
Support float64 data type as input of benchmark_app

jade-cho:


samples/cpp/benchmark_app/remote_tensors_filling.cpp
```diff
@@ -40,6 +40,8 @@ void fill_buffer_random(void* inputBuffer,
 void fill_buffer(void* inputBuffer, size_t elementsNum, const ov::element::Type& type) {
     if (type == ov::element::f32) {
         fill_buffer_random<float, float>(inputBuffer, elementsNum);
+    } else if (type == ov::element::f64) {
+        fill_buffer_random<double, double>(inputBuffer, elementsNum);
     } else if (type == ov::element::f16) {
         fill_buffer_random<short, short>(inputBuffer, elementsNum);
```
ilyachur, line 46: I think this also should be changed to:
```suggestion
        fill_buffer_random<ov::float16, float>(inputBuffer, elementsNum);
```

jade-cho: Changed

isanghao (approved):

PR #16443, opened 2023-03-21
Fix sanitizer out-of-memory error

olpipi:
### Details:
 - Change max array element check in cnpy to avoid out-of-memory error in sanitizer tests

### Tickets:
 - CVS-105669

### DIFF
--- a/thirdparty/cnpy/cnpy.h
+++ b/thirdparty/cnpy/cnpy.h
@@ -18,17 +18,25 @@
 #include<memory>
 #include<stdint.h>
 #include<numeric>
+#include <sys/sysinfo.h>
 
 namespace cnpy {
 
     struct NpyArray {
+        unsigned long GetFreeMemorySize() {
+            struct sysinfo info{};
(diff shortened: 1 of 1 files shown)


thirdparty/cnpy/cnpy.h
```diff
@@ -18,17 +18,25 @@
...
+        }
+
         NpyArray(const std::vector<size_t>& _shape, size_t _word_size, bool _fortran_order) :
             shape(_shape), word_size(_word_size), fortran_order(_fortran_order)
         {
             num_vals = 1;
             for(size_t i = 0;i < shape.size();i++) num_vals *= shape[i];
             if (word_size &&
-                num_vals > std::vector<char>().max_size() / word_size)
+                num_vals > (0.9 * GetFreeMemorySize() / sizeof(char) / word_size))
```
olpipi, line 39: @ilyachur, should we keep this coefficient 0.9 here? I'd recommend to keep it to avoid very rare returns of out-of-memory error

PR #16447, opened 2023-03-21
Review PSROIPooling class for shape inference aspects

praasz:
### Details:
 - Review interval shape and label propagation.
 - Add template implementation of `shape_infer`.
 - Use new shape infer in CPU plugin.
 - Add test with static shapes.

### Tickets:
 - 97244
 - 97294
 - 97343
 - 97392


src/core/shape_inference/include/psroi_pooling_shape_inference.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include "dimension_util.hpp"
+#include "openvino/op/roi_pooling.hpp"
+#include "roi_pooling_shape_inference.hpp"
+
+namespace ov {
+namespace op {
+namespace psroi_pooling {
+namespace validate {
+template <class TROIPooling, class TShape>
+void feat_input_shape(const TROIPooling* op, const TShape feat_shape) {
```
mitruska, line 19: Is this `TROIPooling` template function currently used with other type than the `v1::PSROIPooling` or the template is used to enable reusage in the future?

mitruska (approved):

src/core/shape_inference/include/psroi_pooling_shape_inference.hpp
praasz: For future use with next versions. But if some base will be created could be changed  to be base class instead.

PR #16460, opened 2023-03-21
Move memory tests from core to template plugin tests

olpipi:
### Details:
 - Rewrite and extend memory tests to use template plugin. Move tests to ov_template_functional_tests
 - Fix creation INTExecutable inside InferRequest 
 - Fix VariableState::set_state to set real value to VariableValue

### Tickets:
 - CVS-93935


src/plugins/template/tests/functional/op_eval/memory.cpp
ilyachur, line 1: I expect that in these tests should be rewritten and use template plugin instead of `evaluate()` method.

olpipi:
I wrote tests that use template plugin and testing assign/readvalue functionality.
But some tests (which use reset and manual value set) cannot be written using template plugin. I kept them.

src/plugins/template/tests/functional/op_eval/memory.cpp
```diff
@@ -21,14 +21,6 @@ using namespace ngraph::opset7;
 
 OPENVINO_SUPPRESS_DEPRECATED_START
```
ilyachur, line 22: Why do we still have this file? I thought that all tests should be moved to: https://github.com/openvinotoolkit/openvino/tree/master/src/plugins/template/tests/functional/op_reference

olpipi:
ReferenceReadValueAssignAddModifyLayerTest is disabled due to bug in ov::IVariableState::set_state. It changes tensor ptr only in VariableState but not in VariableValue. So it does not really update var value.

src/plugins/template/src/sync_infer_request.cpp
```diff
@@ -53,7 +53,8 @@ ov::template_plugin::InferRequest::InferRequest(const std::shared_ptr<const ov::
                               "_WaitPipline"),
     };
 
-    m_executable = get_template_model()->get_template_plugin()->m_backend->compile(get_template_model()->m_model);
+    auto clonned_model = get_template_model()->m_model->clone();
```
ilyachur, line 56: Why do you need to clone the model one more time?!

src/plugins/template/src/variable_state.hpp
```diff
@@ -12,8 +12,15 @@ namespace template_plugin {
...
 public:
     VariableState(const std::string& name, const ov::Tensor& tensor) : ov::IVariableState(name) {
-        set_state(tensor);
+        m_state = tensor;
     }
+    void set_state(const ov::Tensor& state) override {
+        OPENVINO_ASSERT(state.get_shape() == m_state.get_shape(), "Wrong tensor shape.");
+        OPENVINO_ASSERT(state.get_element_type() == state.get_element_type(), "Wrong tensor type.");
+        OPENVINO_ASSERT(state.get_byte_size() == state.get_byte_size(), "Blob size of tensors are not equal.");
+        std::memcpy(m_state.data(), state.data(), state.get_byte_size());
```
ilyachur, line 21: Why don't you change the tensor?

ilyachur (approved):

PR #16461, opened 2023-03-21
[ONNX FE] Implementation of ONNX STFT op

mbencer:
Ticket: 92907
Spec: https://github.com/onnx/onnx/blob/main/docs/Operators.md#STFT


src/frontends/onnx/frontend/src/op/stft.cpp
```diff
@@ -0,0 +1,125 @@
...
+                     ngraph::op::is_constant(frame_step_node.get_node_shared_ptr()) &&
+                         ov::shape_size(frame_step_node.get_shape()) <= 1,
+                     "frame_step input must be a scalar or Shape{1} constant.");
+    const auto frame_step =
+        ov::as_type_ptr<default_opset::Constant>(frame_step_node.get_node_shared_ptr())->cast_vector<int64_t>()[0];
+    const auto signal_param_shape = signal.get_partial_shape();
+    CHECK_VALID_NODE(node,
+                     signal_param_shape.rank().is_static() && signal_param_shape.rank().get_length() == 3 &&
+                         signal_param_shape.is_static(),
+                     "Shape of signal input must be static with the rank equal to 3.");
```
mitruska, line 36: If the whole shape is required to be static there is no need to check `signal_param_shape.rank().is_static()` and `signal_param_shape.is_static()` separately, it can be just `signal_param_shape.is_static()` because static shape must have static rank.
```suggestion
    CHECK_VALID_NODE(node,
                     signal_param_shape.is_static() && signal_param_shape.size() == 3,
                     "Shape of signal input must be static with the rank equal to 3.");
```

src/frontends/onnx/frontend/src/op/stft.cpp
```diff
@@ -0,0 +1,125 @@
...
+    const auto window_node_provided = ng_inputs.size() > 2 && !ngraph::op::is_null(ng_inputs[2]);
+    if (window_node_provided) {  // window input provided
+        if (ng_inputs[2].get_partial_shape().rank().is_static()) {
+            CHECK_VALID_NODE(node,
+                             ng_inputs[2].get_partial_shape().rank().get_length() == 1,
+                             "The rank of window input must be 1D.");
+            if (ng_inputs[2].get_partial_shape()[0].is_static()) {
+                CHECK_VALID_NODE(node,
+                                 ng_inputs[2].get_partial_shape()[0].get_length() == frame_length,
+                                 "The rank of window input must be 1D.");
```
mitruska, line 58: The message about 1D rank is from the check above, something like that can be more informative:
```suggestion
                                 "The dimension of window shape must be equal to `frame_length`.");
```

mitruska (approved):

artkuli (approved):

tomdol (approved):

src/frontends/onnx/frontend/src/op/stft.cpp
mbencer: that's right, thanks, done

PR #16463, opened 2023-03-21
Interpolate v11 usage in ONNX FE

tomdol:
### Details:
 - Switch of the ONNX upsample & resize operators to the new Interpolate version
 - Bump of the default_opset to 11 and TopK-related corrections

### Tickets:
 - 104327


tomdol:
The PR is blocked and the CI is expected to fail until https://github.com/openvinotoolkit/openvino/pull/16448 is merged

src/frontends/onnx/frontend/src/op/resize.cpp
```diff
@@ -177,8 +148,7 @@ OutputVector resize(const onnx_import::Node& node) {
         attrs.coordinate_transformation_mode = Transform_mode::ASYMMETRIC;
     }
 
-    const auto output_shape = calculate_output_shape_based_on_scales(data, scales);
-    return {std::make_shared<default_opset::Interpolate>(data, output_shape, scales, attrs)};
```
mitruska, line 181: Even the [ONNX Resize-10](https://onnx.ai/onnx/operators/onnx__Resize.html#resize-10) has scales input, it looks like it was previously converted by usage of ShapeCalcMode::SIZES, with output_shape input created based on the scales.
In this PR besides the switch to the Interpolate-11, the conversion was also updated to use ShapeCalcMode::SCALES,
so it can be the reason of resize10 test failure. Neverthless the scales mode should give correct result as well.

Is the `onnx_resize10_down_scales_const_nearest` test failing:
- with Interpolate-11 and the ShapeCalcMode::SIZES and custom output shape input (the logic as before)? 
- with Interpolate-4 but with the ShapeCalcMode::SCALES mode?

mbencer (approved):

src/frontends/onnx/frontend/src/op/resize.cpp
tomdol: In both cases the inference is executed with v4 implementation since v11 is currently always downgraded to v4. When SIZES mode is used together with the `calculate_output_shape_based_on_scales` helper the test passes. Switching to SCALES and feeding the scales input directly to the Interpolate op makes the test fail.

mitruska: I recommend to keep the version with the `SIZES` mode to keep backward compatibility, and change the mode to `SCALES` separately after resolving the issue with failing test results. 
Switching to the Interpolate-11 is a one story, removal of this workaround is another.

mitruska (approved):

PR #16467, opened 2023-03-22
Fix issue in reducing HW with small channel size in npsc layout

xuchen-intel:
### Details:
 - *Fix issue in reducing HW with small channel size in npsc layout.*

### Tickets:
 - *106448*


xuchen-intel:
@dmitry-gorokhov Hi Dmitry! Could you please take a look? Minor fix. Added protection for vector size 0.

PR #16473, opened 2023-03-22
fix error with setting hint

songbell:
### Details:
 - problem with setting hint with benchmark_app.py
    [ ERROR ] __init__(): incompatible constructor arguments. The following argument types are supported:
    openvino._pyopenvino.properties.hint.PerformanceMode(value: int)
    
### Tickets:
 - *ticket-id*

### DIFF
--- a/tools/benchmark_tool/openvino/tools/benchmark/main.py
+++ b/tools/benchmark_tool/openvino/tools/benchmark/main.py
@@ -117,7 +117,7 @@ def main():
                         logger.warning(f"No device {device} performance hint is set.")
                         perf_hint = properties.hint.PerformanceMode.UNDEFINED
                     else:
-                        perf_hint = properties.hint.PerformanceMode(args.perf_hint.upper())
+                        perf_hint = args.perf_hint.upper()
                 else:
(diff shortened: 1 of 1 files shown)


tools/benchmark_tool/openvino/tools/benchmark/main.py
```diff
@@ -117,7 +117,7 @@ def get_performance_hint(device) -> properties.hint.PerformanceMode:
                         logger.warning(f"No device {device} performance hint is set.")
                         perf_hint = properties.hint.PerformanceMode.UNDEFINED
                     else:
-                        perf_hint = properties.hint.PerformanceMode(args.perf_hint.upper())
+                        perf_hint = args.perf_hint.upper()
```
songbell, line 120: python3.7 seems construct the enum from a string is not supported in pybind. I'm going to try a higher version.

songbell: @jiwaszki Hi, Jan, can you confirm this? currently in my env, setting hint to benchmark_app.py will report error as in description.

Wovchena:
Well, I suppose this PR isn't needed anymore

PR #16480, opened 2023-03-22
[CPU] Added callback for AddTransformation

v-Golubev:
### Details:
This PR introduces a callback for LP AddTransformation which disables propagation of DQ operations through biases. Example is below:
Old order: `Conv->Bias->DQ`
New order: `Conv->DQ->Bias`

### Tickets:
 - *106503*


v-Golubev:
These changes have been merged into master in [#16781](https://github.com/openvinotoolkit/openvino/pull/16781)

PR #16482, opened 2023-03-22
[GPU] Update tuning params of shape agnostic version of fully_connected_bf_tiled kernel for dGPUs

sshlyapn:
### Details:
 - These changes improve performance of shape-agnostic FC kernel in dGPU

### Tickets:
 - 104973

### DIFF
--- a/src/plugins/intel_gpu/src/kernel_selector/kernels/fully_connected/fully_connected_kernel_bf_tiled.cpp
+++ b/src/plugins/intel_gpu/src/kernel_selector/kernels/fully_connected/fully_connected_kernel_bf_tiled.cpp
@@ -201,12 +201,21 @@ FullyConnected_bf_tiled::GetAutoTuneParams(const fully_connected_params& params,
         max_tile_ofm *= 2;
 
     if (params.is_shape_agnostic) {
+        // Use special tuning params for Gen12HP dGPUs, since these parameters demonstrate higher performance
+        // due to better HW utilization (reduced TILE_OFM parameter) and better assembler kernel's code
+        // generation (extended TILE_K parameter) for both FP16 and FP32 data types
         if (dtype == Datatype::F16) {
             // tune_params(tile_b, tile_ofm, tile_ifm, tile_k, dispatch_bsv, dispatch_fsv, exec_options)
-            selector.Case(tune_params(8,  std::min(max_tile_ofm, 2u), 1, 2, 1,  1, EXE_MODE_AGE_BASED));
+            if (params.engineInfo.supports_immad)
+                selector.Case(tune_params(8, 1, 1, 4, 1, 1, EXE_MODE_AGE_BASED));
+            else
+                selector.Case(tune_params(8,  std::min(max_tile_ofm, 2u), 1, 2, 1,  1, EXE_MODE_AGE_BASED));
         } else if (dtype == Datatype::F32) {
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gpu/src/kernel_selector/kernels/fully_connected/fully_connected_kernel_bf_tiled.cpp
```diff
@@ -203,10 +203,16 @@ FullyConnected_bf_tiled::GetAutoTuneParams(const fully_connected_params& params,
     if (params.is_shape_agnostic) {
         if (dtype == Datatype::F16) {
             // tune_params(tile_b, tile_ofm, tile_ifm, tile_k, dispatch_bsv, dispatch_fsv, exec_options)
-            selector.Case(tune_params(8,  std::min(max_tile_ofm, 2u), 1, 2, 1,  1, EXE_MODE_AGE_BASED));
+            if (params.engineInfo.supports_immad)
```
vladimir-paramuzov, line 206: 1. Please add some comment which explains the motivation of different behavior.
2. Shouldn't it depend on EUs count rather then available ISA? What is more efficient on dGPU with 128 EUs?

src/plugins/intel_gpu/src/kernel_selector/kernels/fully_connected/fully_connected_kernel_bf_tiled.cpp
```diff
@@ -201,12 +201,21 @@ FullyConnected_bf_tiled::GetAutoTuneParams(const fully_connected_params& params,
...
 
     if (params.is_shape_agnostic) {
+        // Use special tuning params for Gen12HP dGPUs, since these parameters demonstrate higher performance
+        // due to better HW utilization (reduced TILE_OFM parameter) and better assembler kernel's code
+        // generation (extended TILE_K parameter) for both FP16 and FP32 data types
         if (dtype == Datatype::F16) {
             // tune_params(tile_b, tile_ofm, tile_ifm, tile_k, dispatch_bsv, dispatch_fsv, exec_options)
-            selector.Case(tune_params(8,  std::min(max_tile_ofm, 2u), 1, 2, 1,  1, EXE_MODE_AGE_BASED));
+            if (params.engineInfo.supports_immad)
+                selector.Case(tune_params(8, 1, 1, 4, 1, 1, EXE_MODE_AGE_BASED));
```
yeonbok, line 210: Is this trend same for various shapes?

src/plugins/intel_gpu/src/kernel_selector/kernels/fully_connected/fully_connected_kernel_bf_tiled.cpp
sshlyapn: I have added some additional info
As for dGPU with 128 EUs - it also has some performance gain. That is why I would like to apply this tuning params for whole Gen12HP generation. Maybe other platforms will get some gain from this tuning as well, but I haven't checked it yet

src/plugins/intel_gpu/src/kernel_selector/kernels/fully_connected/fully_connected_kernel_bf_tiled.cpp
sshlyapn: Yes, I have checked this for several shapes and both FP16 and FP32 precisions from different bert models (including bert-large, bert-base and some other)

yeonbok (approved):

PR #16500, opened 2023-03-23
[CPU] enable f16 inference precision

usstq:
### Details:
 - Replace `enforceBF16` in config with `inferencePrecision`
 - Add f16 support in CPU config property INFERENCE_PRECISION_HINT 
 - Add f16 support in `jit_convert_truncation_emitter`, `jit_convert_saturation_emitter`, `jit_load_emitter` & `jit_store_emitter`
 - Brings f16 related oneDNN impls back
 - Add f16 support in conv/deconv/eltwise/fullyconnect/matmul/mvn/pad/pooling/softmax/subgraph

### oneDNN fork PR

https://github.com/openvinotoolkit/oneDNN/pull/197

### Tickets:
 - CVS-92294


akladiev:

yuxu42:
@usstq can we go on for the PR review?

wenjiew:
Is this PR ready for review? Thanks!

usstq:
@wenjiew Yes, I think it's ready!

yuxu42:
@luo-cheng2021 @tiger100256-hu Could you please review? Thanks!

src/plugins/intel_cpu/src/utils/debug_capabilities.cpp
```diff
@@ -18,6 +18,13 @@
 #include "snippets/op/subgraph.hpp"
 #include <ie_ngraph_utils.hpp>
 
+namespace dnnl {
+namespace impl {
+std::ostream &operator<<(std::ostream &ss, const primitive_attr_t *attr);
```
tiger100256-hu, line 23: the implementation of this two operator seem to be in oneDNN, maybe it will be better that  also put declaration in onednn.

src/plugins/intel_cpu/src/utils/debug_capabilities.h
```diff
@@ -62,6 +62,10 @@ class PrintableVector {
     int maxsize;
 };
 
+template<typename T>
+PrintableVector<T> printable(const std::vector<T>& values, int maxsize = 80) {
```
tiger100256-hu, line 66: which file is using this function **printable**

luo-cheng2021 (approved):
LGTM. We'd better to link the ticket about adding testcases to cover the fp16 function.

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1598,24 +1596,89 @@ bool Graph::InsertNode(NodePtr parent, NodePtr child, NodePtr node, int parentPo
...
+ * important debugging tools for accuracy issues
+ * define F16SET as comma separated list of node types for which F16 is allowed to be enabled
+ * define F16CNT as number of nodes totally allowed to enable f16
+ * increasing/decreasing F16CNT until accuracy issue happens/disappears
+ * from the log we can spot the first node having issue when enabled f16
+ */
+struct FP16Debug {
+    std::string safe_getenv(const char* name, const char* default_value = "") {
+        std::string value = default_value;
+        const char* p = getenv(name);
```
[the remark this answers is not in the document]
usstq: changed

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -495,16 +496,34 @@ void Convolution::getSupportedDescriptors() {
...
         out_candidate = std::make_shared<DnnlBlockedMemoryDesc>(getOutputShapeAtPort(0), outputDataType, nspc);
         createDescriptor({ in_candidate }, { out_candidate });
         return;
     }
 
-    inputDataType = (getOriginalInputPrecisionAtPort(0) == Precision::BF16
-                     && !(isDepthWise() && ndims == 5)) ? memory::data_type::bf16 : memory::data_type::f32;
-    outputDataType = (getOriginalOutputPrecisionAtPort(0) == Precision::BF16
-                      && !(isDepthWise() && ndims == 5)) ? memory::data_type::bf16 : memory::data_type::f32;
+    auto getSupportedDataType = [&](InferenceEngine::Precision originalPrec) {
```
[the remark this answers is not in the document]
usstq: changed into [this, ndims]

src/plugins/intel_cpu/src/utils/debug_capabilities.cpp
usstq: they are actually internal functions not exposed by oneDNN in any headers, but I found it's useful for debugging purpose, when you failed to construct some primitive, you can check attr to see if something is wrong, I prefer not to add more patches into oneDNN and just use it silently.

src/plugins/intel_cpu/src/config.cpp
```diff
@@ -324,8 +342,7 @@ void Config::updateProperties() {
     IE_SUPPRESS_DEPRECATED_START
         _config.insert({ PluginConfigParams::KEY_DUMP_EXEC_GRAPH_AS_DOT, dumpToDot });
     IE_SUPPRESS_DEPRECATED_END;
-
-    if (enforceBF16) {
+    if (inferencePrecision == ov::element::bf16) {
         _config.insert({ PluginConfigParams::KEY_ENFORCE_BF16, PluginConfigParams::YES });
     } else {
```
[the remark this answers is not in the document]
usstq: that means we also need to add a new config param called `ENFORCE_F16`, and it looks like that 
both `ENFORCE_BF16` & `ENFORCE_F16` are redundant and maybe conflict with another framework level config param `INFERENCE_PRECISION_HINT`,  and since `INFERENCE_PRECISION_HINT` is widely supported by all plugins, and `ENFORCE_xxx` config only appears in CPU plugin, so I choose not to add another CPU specific config

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1580,24 +1578,89 @@ bool Graph::InsertNode(NodePtr parent, NodePtr child, NodePtr node, int parentPo
...
+                        Type::FullyConnected, // conv / bert nets
+                        Type::RNNCell,        // recurent nets
+                        Type::RNNSeq,         // recurent nets
+                        Type::MatMul,         // bert nets
+                        Type::ROIPooling,     // object detection nets
+                        Type::Interpolate))    // super resolution nets
+                    continue;   // stop at significant nodes
+            }
+
+            if (inferPrec == InferenceEngine::Precision::FP16) {
```
tiger100256-hu, line 1652:
```suggestion
          else  if (inferPrec == InferenceEngine::Precision::FP16) {
```

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
```diff
@@ -541,6 +541,10 @@ void Transformations::Lpt(const bool hasINT16orINT32Levels, const std::vector<ov
         },
         ngraph::pass::low_precision::ConvolutionBackpropDataTransformation);
 
+    lptManager.get_pass_config()->set_callback<ngraph::pass::low_precision::AddTransformation>(
```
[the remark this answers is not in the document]
usstq: Sorry, my fault, will remove! Thanks for pointing out!

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -518,14 +537,15 @@ void Convolution::getSupportedDescriptors() {
                 // for input of inplace tensor precision) to FP32. This will add reorder for that in-place tensor
                 // bofore the fused convolution. This behaviour might be more correct regarding expected markup
                 // of the graph but performance of first and second approaches might be different. Need to verify
-                outputDataType = eltwisePrecision == Precision::BF16 ? memory::data_type::bf16 : memory::data_type::f32;
+                outputDataType = getSupportedDataType(eltwisePrecision);
```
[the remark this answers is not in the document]
usstq: Yes, I did it on purpose, because 5D depthwise convolution on bf16/f16 have performance issue, and we fallback to f32 in general (when no EltwiseAdd fused), and we should keep this behavior when there is EltwiseAdd fused.

src/plugins/intel_cpu/src/graph.cpp
usstq: OK,

tiger100256-hu (approved):

yuxu42:
@dmitry-gorokhov Could you please take a review? Thanks!

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -247,11 +247,20 @@ void FullyConnected::getSupportedDescriptors() {
         if (one_of(outputDataType , memory::data_type::u8, memory::data_type::s8)) {
             outputDataType = memory::data_type::bf16;
         }
+    } else if (inputDataType == memory::data_type::f16) {
+        // f16 input only supports f16/f32 output, even if FQ is fused as post-ops
+        if (!one_of(outputDataType , memory::data_type::f32, memory::data_type::f16)) {
```
dmitry-gorokhov, line 252: Same. Lets clarify if oneDNN will handle this and add ticket.

src/plugins/intel_cpu/src/nodes/eltwise.cpp
```diff
@@ -715,11 +720,11 @@ struct jit_uni_eltwise_generic : public jit_uni_eltwise_kernel, public jit_gener
 
             switch (dst_prc) {
                 case Precision::FP32:
-                    if (src_prc != Precision::FP32 && src_prc != Precision::BF16)
+                    if (src_prc != Precision::FP32 && src_prc != Precision::BF16 && src_prc != Precision::FP16)
```
dmitry-gorokhov, line 723: is_float()?

src/plugins/intel_cpu/src/nodes/deconv.cpp
```diff
@@ -442,15 +442,17 @@ void Deconvolution::getSupportedDescriptors() {
         if (InferenceEngine::Precision::BF16 == outPrecision)
             outPrecision = InferenceEngine::Precision::FP32;
     } else {
-        if (!one_of(inPrecision, InferenceEngine::Precision::FP32, InferenceEngine::Precision::BF16))
+        if (!one_of(inPrecision, InferenceEngine::Precision::FP32, InferenceEngine::Precision::BF16, InferenceEngine::Precision::FP16))
```
dmitry-gorokhov, line 445: is_float()?

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1580,24 +1578,87 @@ bool Graph::InsertNode(NodePtr parent, NodePtr child, NodePtr node, int parentPo
     return true;
 }
 
+#ifdef CPU_DEBUG_CAPS
```
dmitry-gorokhov, line 1581: I am wondering if we can do small changes for this debug capability:
1. Is it possbile to extend it for BF16 markup as well?
2. Can we move it into src/utils (into existing or new file)?
3. Can we align env variable naming with other CPU debug variables?
4. It would be nice to add small document descrbing how this itility can be used for debugging (docs/debug_capabilities)

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1612,7 +1673,7 @@ void Graph::EnforceBF16() {
     // starting from output nodes
     for (const auto& entry : outputNodesMap) {
         const auto& node = entry.second;
-        if (node->getOriginalInputPrecisionAtPort(0) == Precision::BF16)
+        if (node->getOriginalInputPrecisionAtPort(0) == inferPrec)
```
dmitry-gorokhov, line 1676: Please fix comment above this code block (`Skip BF16 enforcement for tail`)

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1622,26 +1683,52 @@ void Graph::EnforceBF16() {
             continue;
 
         if (node->getType() != Type::Input && node->getType() != Type::Output) {
+            // FP16 is only implemented on limited types of node.
```
dmitry-gorokhov, line 1686: BF16 is implemented for limited numeber of  nodes as well. The expection is the node itself will fallback on FP32 precision in case lower precision is not suppotred. 
Could you please clarify why we need different behavior for FP16?

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1580,24 +1578,87 @@ bool Graph::InsertNode(NodePtr parent, NodePtr child, NodePtr node, int parentPo
...
+        inferPrec = InferenceEngine::Precision::BF16;
+        break;
+    case ov::element::f16:
+        inferPrec = InferenceEngine::Precision::FP16;
+        break;
+    default:
+        break;
+    }
+
+    if (inferPrec == InferenceEngine::Precision::FP32)
```
dmitry-gorokhov, line 1630: Maybe just do return in `default:` case in switch block above?

wenjiew:
@usstq Let us address the review comments. Thanks!

src/plugins/intel_cpu/src/config.cpp
```diff
@@ -256,10 +269,15 @@ void Config::readProperties(const std::map<std::string, std::string> &prop) {
     // apply execution mode after all the params are handled to prevent possible conflicts
     // when both execution_mode and inference_precision are specified
     if (!inferencePrecisionSetExplicitly) {
-        if (executionMode == ov::hint::ExecutionMode::PERFORMANCE && (mayiuse(avx512_core_bf16))) {
-            enforceBF16 = true;
+        if (executionMode == ov::hint::ExecutionMode::PERFORMANCE) {
+            if (mayiuse(avx512_core_bf16))
+                inferencePrecision = ov::element::bf16;
+            else if (mayiuse(avx512_core_amx_fp16) || mayiuse(avx512_core_fp16))
```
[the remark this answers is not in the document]
usstq: OK, agree

src/plugins/intel_cpu/src/config.cpp
```diff
@@ -197,15 +204,21 @@ void Config::readProperties(const std::map<std::string, std::string> &prop) {
...
-                    enforceBF16 = true;
+                    inferencePrecision = ov::element::bf16;
                 } else {
                     IE_THROW() << "Platform doesn't support BF16 format";
                 }
+            } else if (val == "f16") {
+                if (mayiuse(avx512_core_fp16) || mayiuse(avx512_core_amx_fp16)) {
+                    inferencePrecision = ov::element::f16;
+                } else {
+                    IE_THROW() << "Platform doesn't support FP16 format";
```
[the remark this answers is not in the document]
usstq: Agree, moreover in cases the precision hint is not support on target platform, we should just keep `inferencePrecision` unchanged (which is initialized in constructor) and keep `inferencePrecisionSetExplicitly` un-touched, so later on the logic in line 269 will start work to choose the best infer precision based on `executionMode` instead, is that OK?

src/plugins/intel_cpu/src/emitters/x64/jit_conversion_emitters.cpp
```diff
@@ -110,6 +110,11 @@ void jit_convert_truncation_emitter::emit_isa(const std::vector<size_t> &in_vec_
             if (one_of(output_type, ov::element::i32, ov::element::i8, ov::element::u8))
                 h->uni_vcvttps2dq(vmm_dst, vmm_dst);
             break;
+        case ov::element::f16:
+            // to be exact, vcvtph2ps belongs to AVX512VL/AVX512F
+            assert(dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx512_core_fp16));
+            h->vcvtph2ps(vmm_dst, Ymm(vmm_src.getIdx()));
+            break;
```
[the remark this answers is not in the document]
usstq: Thanks for pointing out, will change!

src/plugins/intel_cpu/src/emitters/x64/jit_load_store_emitters.cpp
```diff
@@ -665,7 +679,7 @@ void jit_store_emitter::emit_isa(const int in_vec_idx, const Xbyak::Reg64 &reg_d
     if (src_prc_ != dst_prc_) {
         switch (src_prc_) {
             case Precision::FP32:
-                if ((dst_prc_ != Precision::FP32) && (dst_prc_ != Precision::BF16)) {
+                if ((dst_prc_ != Precision::FP32) && (dst_prc_ != Precision::BF16) && (dst_prc_ != Precision::FP16)) {
```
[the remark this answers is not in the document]
usstq: Good, will change

src/plugins/intel_cpu/src/utils/debug_capabilities.h
```diff
@@ -120,12 +126,12 @@ std::ostream & operator<<(std::ostream & os, const PrintableVector<T>& vec) {
     return os;
 }
 
-static inline std::ostream& write_all_to_stream(std::ostream& os) {
+static inline std::ostream& _write_all_to_stream(std::ostream& os) {
```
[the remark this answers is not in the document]
usstq: It conflict with `ov::write_all_to_stream()`  in openvino/src/core/include/openvino/core/except.hpp

src/plugins/intel_cpu/src/nodes/softmax.cpp
```diff
@@ -81,7 +81,9 @@ void SoftMax::getSupportedDescriptors() {
         return;
 
     InferenceEngine::Precision precision = getOriginalInputPrecisionAtPort(0);
-    if (precision != InferenceEngine::Precision::FP32 && precision != InferenceEngine::Precision::BF16)
+    if (precision != InferenceEngine::Precision::FP32 &&
```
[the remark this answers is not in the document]
usstq: Yes, will change

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -91,7 +91,7 @@ bool MVNKey::operator==(const MVNKey& rhs) const {
 
 // some utility functions
 static inline bool isFloatCompatible(Precision prc) {
-    return Precision::FP32 == prc || Precision::BF16 == prc;
+    return Precision::FP32 == prc || Precision::BF16 == prc || Precision::FP16 == prc;
```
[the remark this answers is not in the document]
usstq: OK

src/plugins/intel_cpu/src/nodes/eltwise.cpp
usstq: OK

src/plugins/intel_cpu/src/nodes/deconv.cpp
usstq: OK

src/plugins/intel_cpu/src/graph.cpp
usstq: OK

src/plugins/intel_cpu/src/graph.cpp
usstq: OK, will change

src/plugins/intel_cpu/src/graph.cpp
usstq: OK, I didn't realize that, maybe this can be removed, let me do some tests!

src/plugins/intel_cpu/src/graph.cpp
usstq: OK, changed

usstq:
I found some regression in performance after rebase, debugging...

src/plugins/intel_cpu/src/config.cpp
```diff
@@ -51,8 +51,15 @@ Config::Config() {
     }
 #endif
 
-    if (!mayiuse(avx512_core_bf16))
-        enforceBF16 = false;
+    if (mayiuse(avx512_core_bf16)) {
```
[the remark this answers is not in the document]
dmitry-gorokhov: Then why not to use `ov::element::f32` as safe default option for each CPU architecture: x86/ARM now and RV64 in future? This default sounds very logical for me since actual inference precision value might be only changed by user (by inference_precision and execution_mode properties handled in readProperties()).
Finally this logic is not aligned with initialization code inside readProperties() which confuses a lot.

src/plugins/intel_cpu/src/nodes/eltwise.cpp
```diff
@@ -697,6 +698,10 @@ struct jit_uni_eltwise_generic : public jit_uni_eltwise_kernel, public jit_gener
                     vpmovzxwd(vmm_src, op);
                     uni_vpslld(vmm_src, vmm_src, 16);
                     break;
+                case Precision::FP16:
+                    assert(mayiuse(x64::avx512_core_fp16));
```
dmitry-gorokhov, line 702: BTW, from my understanding conversion instructions like `vcvtph2ps` don't actually require `avx512_core_fp16` ISA, but just F16C + AVX512VL/AVX512F (or pure avx2)  which is available on all modern intel CPUs. Given that we can relax ISA limitation for all operations that uses only FP32<->FP16 conversion and keep real math in FP32 (like Eltwise, MVN, Interpolate etc). By doing that we can enable FP16 tests for such layers in precommit already now. 
What do you think? Sounds like worth to add in follow-up PR.

src/plugins/intel_cpu/docs/debug_capabilities/infer_prc.md
```diff
@@ -0,0 +1,8 @@
+# Inference precision debug
+
+When developing/debugging accuracy issue related to inference precision feature, you can define following environment variables to control the type & number of nodes for which the specified inference precision is enforced:
+
+ - `OV_INFER_PRC_TYPES` : comma separated list of node types for which infer-precision is enforced
```
dmitry-gorokhov, line 5: Please align with other debug caps naming: should be `OV_CPU_INFER_PRC_TYPES` and `OV_CPU_INFER_PRC_CNT`

src/plugins/intel_cpu/src/nodes/deconv.cpp
dmitry-gorokhov: Still there

src/plugins/intel_cpu/src/config.cpp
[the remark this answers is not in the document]
dmitry-gorokhov: Yes, I believe this is the best behavior.

dmitry-gorokhov:
@usstq Could you please create PR with corresponding changed in oneDNN fork? I would also ask you to have only one commit for FP16 enabling there.

dmitry-gorokhov:
@usstq We also need to create ticket for FP16 signle layer tests (for Convolutions, Matmuls etc) enabling activities once GNR will be available.

dmitry-gorokhov:
@usstq Please also check binary size impact. We need to understand the change caused by FP16 instances.

src/plugins/intel_cpu/src/config.cpp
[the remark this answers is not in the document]
usstq: Agree, will change

src/plugins/intel_cpu/docs/debug_capabilities/infer_prc.md
usstq: OK

src/plugins/intel_cpu/src/nodes/deconv.cpp
usstq: Sorry, changed

src/plugins/intel_cpu/src/nodes/eltwise.cpp
usstq: OK, let me remove these assert and test can be added in follow-up PR

usstq: BTW, one exception is `vcvtss2sh/vcvtsh2ss` used in  `load_scalar/store_scalar` which requires AVX512-FP16, Can we changed them into using `vcvtps2ph/vcvtph2ps` instead?

usstq:
> @usstq We also need to create ticket for FP16 signle layer tests (for Convolutions, Matmuls etc) enabling activities once GNR will be available.

OK, will do.

> @usstq Could you please create PR with corresponding changed in oneDNN fork? I would also ask you to have only one commit for FP16 enabling there.

Done: https://github.com/openvinotoolkit/oneDNN/pull/197

> @usstq Please also check binary size impact. We need to understand the change caused by FP16 instances.

libopenvino_intel_cpu_plugin.so has been increased from 47813192 to 48566856, by 736KB, relative increase 1.58%

usstq:
@dmitry-gorokhov  I have fixed fp16 brgconv issue and validated locally, and recent review comments also have been addressed, please review again, Thanks!

usstq:
@dmitry-gorokhov I just found that MHA node will throw exception when enforce fp16:
```

void MHA::initSupportedPrimitiveDescriptors() {
    if (!supportedPrimitiveDescriptors.empty())
        return;

    for (auto idx : {0, 1, 2, 3}) {
        inputPrecisions.push_back(getOriginalInputPrecisionAtPort(idx));
        if (!one_of(inputPrecisions[idx], Precision::FP32, Precision::BF16, Precision::I8))
            THROW_ERROR << "doesn't support " << inputPrecisions[idx].name() << " precision on " << idx <<  " input port";
    }
```

dmitry-gorokhov:
> @dmitry-gorokhov I just found that MHA node will throw exception when enforce fp16, should I change it's behaviour by fallback to FP32 automatically
> 
> ```
> 
> void MHA::initSupportedPrimitiveDescriptors() {
>     if (!supportedPrimitiveDescriptors.empty())
>         return;
> 
>     for (auto idx : {0, 1, 2, 3}) {
>         inputPrecisions.push_back(getOriginalInputPrecisionAtPort(idx));
>         if (!one_of(inputPrecisions[idx], Precision::FP32, Precision::BF16, Precision::I8))
>             THROW_ERROR << "doesn't support " << inputPrecisions[idx].name() << " precision on " << idx <<  " input port";
>     }
> ```

Yes. MHA behavior should be updated.

src/plugins/intel_cpu/src/nodes/eltwise.cpp
dmitry-gorokhov: For Eltwise it is save. 
Not sure about load/store_emitters. Would ask @chenhu-wang to comment.

usstq: OK, that's enough, load/store_emitters actually already using vector version `vcvtps2ph/vcvtph2ps` with mask to handle variable length load/store

usstq:
@dmitry-gorokhov MHA's behavior is changed and avx512_fp16 assertions are totally removed, I validated following models  on local machine using FP16 infer precision and found no regression in accuracy & performance:
 - bert-large-uncased-whole-word-masking-squad-onnx
 - resnet_v1.5_50
 - distilbert-base-uncased-sst-2
 - vit-base-16-224

src/plugins/intel_cpu/src/nodes/eltwise.cpp
dmitry-gorokhov: You are still using `avx512_core_fp16` check which, from my understanding, is available on SPR only. In other words `avx512_core_fp16` is not equal `f16c + avx512f + avx512vl`. So to enable single layer tests in precommit we need to relax isa limitation.

dmitry-gorokhov (approved):
Merging the PR as long as major functionality is completed.
There are two follow-ups:
1. Enable FP16 single layer tests on HW with AVX512 support.
2. Clarify with oneDNN on lacking functionality

src/plugins/intel_cpu/src/nodes/eltwise.cpp
usstq: Yes,  indeed, some avx512_core_fp16 checks are still there, sorry I didn't remove them completely, and we can do that in single layer tests PR.

usstq:
> Merging the PR as long as major functionality is completed. There are two follow-ups:
> 
> 1. Enable FP16 single layer tests on HW with AVX512 support.
> 2. Clarify with oneDNN on lacking functionality

OK, no problem, will follow-up these tasks

src/plugins/intel_cpu/src/nodes/eltwise.cpp
chenhu-wang: Just a reminder that sse4 do not support f16<-> f32 convert instructions. Instead of report exceptions or assert, there should be some alignment, in property(precision hint) reading stage or createSupportedPrimitive() stage of nodes.

PR #16501, opened 2023-03-23
TopK 11 exposed to Python

tomdol:
### Details:
 - Python bindings for TopK v11

### Tickets:
 - 101500


akuporos (approved):

jiwaszki (approved):

src/bindings/python/src/compatibility/ngraph/opset11/ops.py
```diff
@@ -75,3 +75,33 @@ def interpolate(
...
+    :param index_element_type: Type of output tensor with indices.
+    :param stable: Specifies whether the equivalent elements should maintain
+                   their relative order from the input tensor during sorting.
+    :return: The new node which performs TopK
+    """
+    return _get_node_factory_opset11().create(
+        "TopK",
+        as_nodes(data, k),
+        {"axis": axis, "mode": mode, "sort": sort, "index_element_type": index_element_type, "stable": stable},
+    )
```
slyalin, line 107: @tomdol, just for my education please comment why do we need to add it twice: here and in `openvino`. Is it possible to implement once and import in the second place?

tomdol: I think it was the initial design of how legacy API and the new API will expose ops to Python. Totally separate, easy to remove at some point. I'm not entirely sure though, perhaps @jiwaszki can comment.

jiwaszki: @tomdol is right. It was made this way so everything would be simple to manage (read: remove in future) and fits in complex project build systems. In addition, opsets from API 1.0 `compatibility` and current 2.0 `runtime` are not working with each other -- as well as most of other objects. So it makes sense to separate them.

jiwaszki: @slyalin did I answer your question?

slyalin: @jiwaszki,

When are we going to delete the compatibility part?

Not sure that I can realize the real complexity of all distribution channels, but doesn't it look like we always distribute both `compatibility` and `openvino` parts together so we technically we can import `openvino` parts in `compatibility` parts as we do in C++ openvino and ngraph (legacy) headers where operations are defined. Does it make sense to add new ops only in the new openvino module and avoid doing it in the legacy part? I'm just trying to better understand, probably my current understanding is oversimplified.

jiwaszki: > When are we going to delete the compatibility part?

Exactly at the same time when old API is being removed.

> so we technically we can import openvino parts in compatibility parts as we do in C++ openvino and ngraph (legacy) headers where operations are defined.

Short answer: no.
Longer answer:
1) All classes/objects are different in old and new APIs, due to underlaying C++ code. That's why one operator (which is in fact a child of many different classes, i.e. `Node`) is not equal to the same one in other API.
2) pybind is not able to register the same class two times. That is why if you inspect a code you will be able to see `py::module_local` in many places. Limitation of the solution itself.
3) Underlaying C++ code, with addition of NodeFactory layer does not provide a way to produce two "different" objects from the same call. They need to be understood by whole module, not only by some Python end-point code.
4) This split is much easier to maintain and understand than "aliasing of everything". Additionally it provides clear split between APIs. Now API 1.0 is in it's own place, API 2.0 in it's own. There is no conflicts of names, no accidental uses of objects from different API. There are more benefits than disadvantages that come from this simple, yet effective, approach.

I hope this answers your questions in detail.

PR #16514, opened 2023-03-23
[INSTALL] Fix setupvars (installation for MacOS) and build python 

iefode:
### Details:
 - *Fix find Cython to build OV/IE Python under MacOS*
 - *Undate `setupvars.sh` to work under MacOS*

### Tickets:
 - *96897*


scripts/setupvars/setupvars.sh
```diff
@@ -146,8 +146,8 @@ check_python_version () {
     fi
 } 
 
-python_version_to_check="$python_version"
-if [ -z "$python_version" ]; then
+python_version_to_check="$python_version_major"
+if [ -z "$python_version_to_check" ]; then
```
ilya-lavrenov, line 150: looks like `python_version_major` is not defined yet here
maybe we need to revert the change?

scripts/setupvars/setupvars.sh
```diff
@@ -121,7 +121,9 @@ check_python_version () {
         "${PYTHON_VERSION_MAJOR}.${MAX_SUPPORTED_PYTHON_VERSION_MINOR} (64-bit) from https://www.python.org/downloads/"
         return 0
     fi
-    python_bitness=$(python"$python_version" -c 'import sys; print(64 if sys.maxsize > 2**32 else 32)')
+    if command -v python"$python_version" > /dev/null 2>&1; then
```
ilya-lavrenov, line 124: Suggest:
```bash
   if command -v python"$python_version" > /dev/null 2>&1; then
         python_interp=python"$python_version"
   else
         python_interp=python"$python_version_major"
   fi
   python_bitness=$("$python_interp" -c 'import sys; print(64 if sys.maxsize > 2**32 else 32)')
```

scripts/setupvars/setupvars.sh
```diff
@@ -147,8 +149,9 @@ check_python_version () {
 } 
 
 python_version_to_check="$python_version"
-if [ -z "$python_version" ]; then
-    python_version_to_check="3"
+local_py_version="$( python3 -c 'import sys; print(".".join(map(str, sys.version_info[0:2])))' )"
```
ilya-lavrenov, line 152: who said that `python3` exists ?
Line 157 exactly required to check it

scripts/setupvars/setupvars.sh
```diff
@@ -149,9 +152,10 @@ check_python_version () {
 } 
 
 python_version_to_check="$python_version"
-local_py_version="$( python3 -c 'import sys; print(".".join(map(str, sys.version_info[0:2])))' )"
-if [ "$local_py_version" = "$python_version_to_check" ]; then
-        python_version_to_check="$PYTHON_VERSION_MAJOR"
+if [ -z "$python_version_to_check"  ]; then
+    python_version_to_check="3"
+else
+    python_version_to_check="$python_version_major"
```
ilya-lavrenov, line 158: is `python_version_major` really defined?
I suppose it's defined only after `check_python_version ` call which happens below.

iefode: @ilya-lavrenov I am sure that the source code looked so strange and as a w/a. I decided to change it after review. Please check.
Enjoy :)

scripts/setupvars/setupvars.sh
```diff
@@ -144,14 +149,10 @@ check_python_version () {
             echo "[setupvars.sh] WARNING: OpenVINO Python environment does not set properly"
         fi
     fi
-} 
-
-python_version_to_check="$python_version"
-if [ -z "$python_version" ]; then
-    python_version_to_check="3"
-fi
+}
```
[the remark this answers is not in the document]
ilya-lavrenov: ping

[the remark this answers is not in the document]
iefode: pong.
It's still installation problem on MacOS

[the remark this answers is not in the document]
iefode: Please review. I reverted such changes and replaced var by pre-defined python version. Anyway the py vesion will be checked in the validator. May be, is not the best way to compare py_version_to_check with empty string?

scripts/setupvars/setupvars.sh
```diff
@@ -4,12 +4,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 abs_path () {
-    path=$(eval echo "$1")
-    directory=$(dirname "$path")
-    echo "$(cd "$directory" || exit; pwd -P)/$(basename "$path")";
+    script_path=$(eval echo "$1")
+    directory=$(dirname "$script_path")
+    echo "$(cd "$directory" || exit; pwd -P)/$(basename "$script_path")";
```
ilya-lavrenov, line 9:
```suggestion
    echo "$(cd "$directory" || exit; pwd -P)";
```

ilya-lavrenov (approved):

PR #16517, opened 2023-03-23
[CPU] Optimize TBB usage in the parallel dynamic shapes processing

maxnick:
### Details:
Parallel execution of the updateShapes and updateDynParams stages was revised. Now each task is pinned to a particular slot inside the task area. It helps to mitigate overheads on threads switching. Also OMP version the parallel tasks run was introduced.

### Tickets:
 - 99129


maxnick:
@dmitry-gorokhov , the PR is ready for review. Could you please take a look when you have time?

dmitry-gorokhov (approved):

PR #16520, opened 2023-03-23
[TF FE] Post leftovers to support the MUSE model in SavedModel format

rkazants:
**Details:** It contains tests imitating a case with Tokenizer extension and raised problems: setting custom type for body graph Parameter via conversion extension, named ports for `RaggedTensorToSparse` and `Unique` operations.

**Ticket:** TBD

### DIFF
--- a/src/frontends/tensorflow/src/translate_session.cpp
+++ b/src/frontends/tensorflow/src/translate_session.cpp
@@ -66,7 +66,16 @@ void TranslateSession::inject_body_model(std::shared_ptr<ov::Model> body_model,
                             "inputs and arguments to the function " +
                                 operation_type + " do not match.");
     for (size_t param_ind = 0; param_ind < body_parameters.size(); ++param_ind) {
+        auto orig_type = body_parameters[param_ind]->get_element_type();
         body_parameters[param_ind]->output(0).replace(ov_inputs[param_ind]);
+        if (auto ext_parameter = as_type_ptr<ov::opset8::Parameter>(ov_inputs[param_ind].get_node_shared_ptr())) {
+            // save type of a Parameter as converted in the body
+            // this is important if the external conversion extension is applied to body graph node
+            // with setting its own type
+            if (orig_type != element::dynamic) {
+                ext_parameter->set_element_type(orig_type);
+            }
+        }
     }
     for (const auto& result_node : body_model->get_results()) {
         ov_outputs.push_back(result_node->input_value(0));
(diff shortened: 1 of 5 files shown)


gkrivor (approved):

src/frontends/tensorflow/tests/convert_tricky_models.cpp
```diff
@@ -38,6 +42,36 @@ shared_ptr<Model> convert_model(const string& model_path) {
...
+    if (auto parameter = as_type_ptr<ov::opset10::Parameter>(strings.get_node_shared_ptr())) {
+        parameter->set_partial_shape(ov::PartialShape{Dimension()});
+        parameter->set_element_type(ov::element::u8);
+        parameter->validate_and_infer_types();
+    }
+
+    row_splits = make_shared<ConvertLike>(row_splits, strings);
+    auto const_one = make_shared<Constant>(row_splits.get_element_type(), Shape{}, 1);
+    Output<Node> mul = make_shared<Multiply>(row_splits, const_one);
+    auto const_two = make_shared<Constant>(ov::element::u8, Shape{}, 2);
```
itikhono, line 64: row_splits.get_element_type() can be different from u8, in case of strings is Constant with other type.
maybe it's better to check this or use strings.get_element_type

itikhono (approved):

slyalin (approved):

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -66,7 +66,16 @@ void TranslateSession::inject_body_model(std::shared_ptr<ov::Model> body_model,
                             "inputs and arguments to the function " +
                                 operation_type + " do not match.");
     for (size_t param_ind = 0; param_ind < body_parameters.size(); ++param_ind) {
+        auto orig_type = body_parameters[param_ind]->get_element_type();
         body_parameters[param_ind]->output(0).replace(ov_inputs[param_ind]);
+        if (auto ext_parameter = as_type_ptr<ov::opset8::Parameter>(ov_inputs[param_ind].get_node_shared_ptr())) {
```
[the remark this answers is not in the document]
rkazants: Yeah, this is annoying thing from which I want to get rid of in the separate PR. Actually, I am afraid to use opset10 for Parameter because everywhere in TF FE we use opset8 for Parameter.

src/frontends/tensorflow/tests/convert_tricky_models.cpp
rkazants: `const_two` is used for addition with strings for which we re-write its type to `u8` so they should be aligned.

PR #16524, opened 2023-03-24
[TF FE] Added MetaGraph file format

gkrivor:
### Details:
 - Add MetaGraph support

### Tickets:
 - 106762


rkazants (changes requested):
Please add tests for both SM and MetaGraph formats

gkrivor:
Tests for this functionality are available in e2e coverage.

rkazants (commented):
It is definitely good to have a few tests for both SM and MetaGraph. I remain some comments among them please add some checks for memories and use of smart pointers.

Three files remains to review for me.

src/frontends/tensorflow/src/input_model.cpp
```diff
@@ -56,7 +56,7 @@ class InputModel::InputModelTFImpl {
     InputModelTFImpl(const GraphIterator::Ptr& graph_iterator,
```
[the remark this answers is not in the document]
gkrivor: Good question. It is declared in "input_model.hpp" as an incomplete class (to make a weak link between entities). And because we don't explicitly use it anywhere (we don't count usage as an pointer) - we don't need to include full declaration.

src/frontends/tensorflow/src/variables_index.cpp
```diff
@@ -0,0 +1,456 @@
...
+                                "Cannot retrieve uncompressed block length");
+        std::string uncompressed_string;
+        uncompressed_string.reserve(uncompressed_length);
+        snappy::Uncompress(data.data(), data.size(), &uncompressed_string);
+        data.resize(uncompressed_length);
+        std::copy(uncompressed_string.begin(), uncompressed_string.end(), data.begin());
+        block_size = uncompressed_length;
+    }
+#endif
+    uint32_t numRestarts = smReadFixed<uint32_t>(data.data() + block_size - sizeof(uint32_t));
```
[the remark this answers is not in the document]
gkrivor: Not sure it is necessary, data.size() is always >= 5 bytes. block_size is unsigned. 5 > 4 (sizeof(uint32_t))

src/frontends/tensorflow/src/variables_index.cpp
```diff
@@ -0,0 +1,456 @@
...
+        node_dictionary[global_name] = node.second;
+    }
+}
+
+void VariablesIndex::map_assignvariable(const std::shared_ptr<::tensorflow::GraphDef> graph_def,
+                                        std::map<std::string, std::string>& variables_map) {
+    std::map<std::string, PtrNode*> nodes;
+
+    for (const auto& node : graph_def->node()) {
+        nodes[node.name()] = new PtrNode(node, nodes);
```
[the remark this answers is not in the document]
gkrivor: Yep, why not, on next PR iteration.

src/frontends/tensorflow/tests/test_models/gen_scripts/generate_saved_model_variables.py
```diff
@@ -0,0 +1,22 @@
...
+    self.var1 = tf.Variable(123.0)
+  @tf.function(input_signature=[tf.TensorSpec([], tf.float32)])
+  def mutate(self, new_v):
+    self.var1.assign(new_v)
+  @tf.function(input_signature=[tf.TensorSpec([1], tf.float32)])
+  def __call__(self, x):
+    return {'test_output_name': x * self.var1}
+
+module = AddVariable()
+tf.saved_model.save(module, os.path.join(sys.argv[1], "saved_model_variables"))
```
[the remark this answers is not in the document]
gkrivor: Updated file, not sure I understand original idea of change.

tools/mo/openvino/tools/mo/front/tf/loader.py
```diff
@@ -334,6 +334,10 @@ def convert_to_pb(argv: argparse.Namespace):
             isinstance(argv.input_model, str):
         return None
 
+    # Saved Model format and MetaGraph format is supported without freezing
+    if argv.saved_model_dir or argv.input_meta_graph:
```
mvafin: Is this really needed here? Can you just not call `convert_to_pb` for meta and sm?

src/frontends/tensorflow/src/meta.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <fstream>
+#include <string>
+
+#include "graph_iterator_meta.hpp"
+#include "openvino/core/type/element_type.hpp"
+#include "tensor_bundle.pb.h"
+#include "trackable_object_graph.pb.h"
+
+#ifdef ENABLE_SNAPPY_COMPRESSION
+#    include "snappy.h"
```
mvafin: How is snappy used in this file? Is this include really needed?

src/frontends/tensorflow/src/input_model.cpp
```diff
@@ -394,6 +394,35 @@ ov::frontend::Place::Ptr InputModel::InputModelTFImpl::get_place_by_tensor_name(
...
+                } else {
+                    operation_name = alt_name.first;
+                }
+                break;
+            }
+        }
+    }
+
+    if (m_saved_model_output_names.get()) {
+        for (const auto& alt_name : *m_saved_model_output_names) {
```
mvafin, line 413: nitpick: these two for loops are very similar, can be extracted to a function called twice

tools/mo/openvino/tools/mo/front/tf/loader.py
gkrivor: I don't know how it works here, copy/pasted from original solution :) If you know better way - could you point on it?

rkazants: Checkpoint is remained. Once checkpoint is implemented, we will remove `convert_to_pb` at all.

rkazants: So I am leaning to hide this inside this function

src/frontends/tensorflow/src/meta.cpp
gkrivor: Good catch. It is artefact when code was in saved_model.cpp. Now it is used only in variables_index.cpp. I'll remove it in both files. I think it'll be useless in future too.

tools/mo/openvino/tools/mo/front/tf/loader.py
```diff
@@ -335,7 +335,7 @@ def convert_to_pb(argv: argparse.Namespace):
         return None
 
     # Saved Model format and MetaGraph format is supported without freezing
-    if argv.saved_model_dir or argv.input_meta_graph:
+    if not argv.use_legacy_frontend and (argv.saved_model_dir or argv.input_meta_graph):
```
rkazants: your previous version was correct

tools/mo/unit_tests/mo/front/tf/convert_to_pb_test.py
```diff
@@ -154,13 +155,15 @@ def test_meta_format_session_clearing(self):
...
+            self.argv.use_legacy_frontend = true
             path_to_pb = convert_to_pb(self.argv)
             self.assertTrue(os.path.exists(path_to_pb), "The auxiliary .pb is not generated")
             self.assertTrue(os.path.getsize(path_to_pb) != 0, "The auxiliary .pb is empty")
 
             self.argv.input_meta_graph = os.path.join(tmp_dir, 'model2.meta')
             self.argv.output_dir = tmp_dir
             self.argv.input_model = None
+            self.argv.use_legacy_frontend = true
             path_to_pb = convert_to_pb(self.argv)
```
rkazants, line 167: not needed. Just expect that `convert_to_pb` returns `None`.

tools/mo/openvino/tools/mo/moc_frontend/pipeline.py
```diff
@@ -34,10 +34,11 @@ def moc_pipeline(argv: argparse.Namespace, moc_front_end: FrontEnd):
...
             input_model = moc_front_end.load(argv.input_model)
-        elif argv.saved_model_dir:
-            input_model = moc_front_end.load(argv.saved_model_dir)
-        elif argv.input_meta_graph:
-            input_model = moc_front_end.load(argv.input_meta_graph)
+        elif not argv.use_legacy_frontend:
+            if argv.saved_model_dir:
+                input_model = moc_front_end.load(argv.saved_model_dir)
+            elif argv.input_meta_graph:
+                input_model = moc_front_end.load(argv.input_meta_graph)
```
rkazants: not needed change

src/frontends/tensorflow/src/input_model.cpp
gkrivor: It is place I want to be reviewed by @rkazants. It is needed to fix issue with names. And yes it might be replaced by a function/lambda if it is correct change for names.

tools/mo/openvino/tools/mo/front/tf/loader.py
gkrivor: ok

tools/mo/unit_tests/mo/front/tf/convert_to_pb_test.py
gkrivor: can/should I check for legacy frontend? I want to leave this check for legacy frontend.

rkazants: no, the legacy does not use `convert_to_pb` at all. Just leave this test with the same input parameters and check that `convert_to_pb` returns None.

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -444,6 +433,22 @@ void TranslateSession::translate_graph(const ov::frontend::InputModel::Ptr& inpu
         }
     }
 
+    if (saved_model_inputs.get() && saved_model_inputs->size() > 0) {
+        for (auto param : params) {
+            if (!apply_saved_model_names(param, saved_model_inputs)) {
```
rkazants, line 438: please add a comment that unused placeholders are not mapped to user's input names

rkazants (approved):

PR #16541, opened 2023-03-24
[MO] Use direct support of non-frozen formats in MO

rkazants:
**Details:** Remove WA to freeze models before submitting them into TF FE. Now it removes WA for MetaGraph and SavedModel formats. For checkpoints and text formats it still exists.

**Ticket:** TBD


rkazants:
No longer needed. Merged with other PR

PR #16547, opened 2023-03-25
[CPU] Element-wise precision selection fix

eshoguli:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


src/plugins/intel_cpu/src/nodes/eltwise.cpp
```diff
@@ -194,17 +219,19 @@ struct jit_uni_eltwise_generic : public jit_uni_eltwise_kernel, public jit_gener
             std::set<std::vector<element::Type>> prcs = get_supported_precisions(eltwise_data_[i].algo);
             std::set<std::vector<element::Type>> prcs_intersect = {};
 
-            // to support previous functionality
-            if (!std::all_of(
+            // to support previous functionality with one precision
```
nshchego, line 222: Please remove this comment or replace it on something like this: "Looking for precisions intersection between eltwise nodes."

src/plugins/intel_cpu/src/nodes/eltwise.cpp
```diff
@@ -194,17 +219,19 @@ struct jit_uni_eltwise_generic : public jit_uni_eltwise_kernel, public jit_gener
...
-                continue;
-            }
-
-            std::set_intersection(supported_precision_intersection.begin(), supported_precision_intersection.end(),
-                                  prcs.begin(), prcs.end(), std::inserter(prcs_intersect, prcs_intersect.begin()));
+                [](const std::vector<element::Type>& precisions) {
+                    return std::all_of(
+                        precisions.begin(),
+                        precisions.end(),
+                        [&precisions](const element::Type precision) { return precision == precisions[0]; });
```
nshchego, line 231: Use reference instead
`const element::Type &precision`

src/plugins/intel_cpu/src/nodes/eltwise.cpp
```diff
@@ -157,6 +160,28 @@ class EltwiseShapeInferFactory : public ShapeInferFactory {
     }
 };
 
+void set_intersection(
+    const std::set<std::vector<element::Type>>& precisions1,
```
nshchego, line 164: Please add indent to distinguish fn arguments from fn body.

eshoguli: done, thanks

src/plugins/intel_cpu/src/nodes/eltwise.cpp
eshoguli: updated, but text is more sutable

src/plugins/intel_cpu/src/nodes/eltwise.cpp
eshoguli: done, thanks

src/plugins/intel_cpu/src/nodes/eltwise.cpp
```diff
@@ -157,88 +160,175 @@ class EltwiseShapeInferFactory : public ShapeInferFactory {
...
-struct jit_uni_eltwise_generic : public jit_uni_eltwise_kernel, public jit_generator {
-    DECLARE_CPU_JIT_AUX_FUNCTIONS(jit_uni_eltwise_generic)
 
-    explicit jit_uni_eltwise_generic(const jit_eltwise_params& jep,
-                                     const std::vector<Eltwise::EltwiseData>& eltwise_data,
-                                     const std::vector<ov::intel_cpu::Type>& ops_list,
-                                     const dnnl::post_ops& post_ops)
-    : jit_uni_eltwise_kernel(jep), jit_generator(jit_name()), eltwise_data_(eltwise_data), ops_list_(ops_list), post_ops_(post_ops) {}
+InferenceEngine::Precision eltwise_precision_helper::get_precision(const size_t inputs_number,
+                                                                   const InferenceEngine::Precision(&src_prc)[MAX_ELTWISE_INPUTS],
```
nshchego, line 188: Does it work here: `const InferenceEngine::Precision *src_prc`

nshchego: ok

nshchego (approved):

src/plugins/intel_cpu/src/nodes/eltwise.cpp
```diff
@@ -157,88 +160,175 @@ class EltwiseShapeInferFactory : public ShapeInferFactory {
...
 
+void set_intersection(const std::set<std::vector<element::Type>>& precisions1,
+                      const std::set<std::vector<element::Type>>& precisions2,
+                      std::set<std::vector<element::Type>>& intersection) {
+                      std::map<element::Type, size_t> intersection_types;
+
+    for (auto it1 = precisions1.begin(); it1 != precisions1.end(); ++it1) {
+        for (auto it2 = precisions2.begin(); it2 != precisions2.end(); ++it2) {
+            const auto& it1_precisions = *it1;
+            // all element types are equal
```
dmitry-gorokhov, line 171: How this is guaranteed? Is it just an assumtion that all eltwise emitters work only with equal precisions on inputs? If so I would suggest to add aditional check that only sets with equal precisions should be used in intersection computation

dmitry-gorokhov (approved):

PR #16552, opened 2023-03-26
support Ecore only in streams calculation

wangleis:
### Details:
 - *Support Ecore only cases in streams calculation*

### Tickets:
 - *ticket-id*


src/plugins/intel_cpu/src/cpu_streams_calculation.cpp
```diff
@@ -28,8 +28,14 @@ std::vector<std::vector<int>> get_streams_info_table(const int input_streams,
     if (1 == input_streams) {
         stream_info[NUMBER_OF_STREAMS] = 1;
         int limit_threads = (input_threads == 0) ? model_prefer_threads : input_threads;
-        if ((proc_type_table[0][EFFICIENT_CORE_PROC] > 0) &&
-            ((limit_threads == 0) || (limit_threads > proc_type_table[0][MAIN_CORE_PROC]))) {
+        if (proc_type_table[0][ALL_PROC] == proc_type_table[0][EFFICIENT_CORE_PROC]) {
```
riverlijunjie, line 31: When `ALL_PROC` is equal to `EFFICIENT_CORE_PROC` ?

wangleis: Use Ecore only on hybrid platform

sunxiaoxia2022 (approved):

PR #16556, opened 2023-03-26
[TF FE] Add layer test for Bucketize

rkazants:
**Details:** Add layer test for Bucketize

**Ticket:** 51023

### DIFF
--- a/tests/layer_tests/tensorflow_tests/test_tf_Bucketize.py
+++ b/tests/layer_tests/tensorflow_tests/test_tf_Bucketize.py
@@ -4,90 +4,43 @@
 import numpy as np
 import pytest
 import tensorflow as tf
-from common.layer_test_class import check_ir_version
 from common.tf_layer_test_class import CommonTFLayerTest
-
-from unit_tests.utils.graph import build_graph
 
 
 class TestBucketize(CommonTFLayerTest):
-    def create_bucketize_net(self, input_shape, input_type, boundaries_size, ir_version,
-                             use_new_frontend):
-        """
-            Tensorflow net:                     IR net:
-                 Input            =>      Input        Boundaries
-                   |                           \       /
-               Bucketize                       Bucketize
-           {attrs: boundaries}
-        """
+    def _prepare_input(self, inputs_info):
+        assert 'input' in inputs_info, "Test error: inputs_info must contain `input`"
+        input_shape = inputs_info['input']
+        input_type = self.input_type
+        inputs_data = {}
+        input_data = np.random.randint(-20, 20, input_shape).astype(input_type)
+        inputs_data['input'] = input_data
+        return inputs_data
 
+    def create_bucketize_net(self, input_shape, input_type, boundaries_size):
+        self.input_type = input_type
         tf.compat.v1.reset_default_graph()
         with tf.compat.v1.Session() as sess:
-            x = tf.compat.v1.placeholder(input_type, input_shape, 'Input')
-            constant_value = np.arange(-boundaries_size * 5, boundaries_size * 5, 10,
-                                       dtype=np.float32)
-            # TODO: Bucketize is not tested here. Need to re-write the test
+            input = tf.compat.v1.placeholder(input_type, input_shape, 'input')
+            # generate boundaries list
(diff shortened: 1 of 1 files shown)


tests/layer_tests/tensorflow_tests/test_tf_Bucketize.py
```diff
@@ -4,90 +4,43 @@
...
-from common.layer_test_class import check_ir_version
 from common.tf_layer_test_class import CommonTFLayerTest
 
-from unit_tests.utils.graph import build_graph
-
 
 class TestBucketize(CommonTFLayerTest):
-    def create_bucketize_net(self, input_shape, input_type, boundaries_size, ir_version,
-                             use_new_frontend):
-        """
```
gkrivor, line 16: I may ask to add additional comments with link on original operation, but could you leave such comments in place where it already exist?

tests/layer_tests/tensorflow_tests/test_tf_Bucketize.py
```diff
@@ -4,90 +4,43 @@
...
-               Bucketize                       Bucketize
-           {attrs: boundaries}
-        """
-
+    def _prepare_input(self, inputs_info):
+        assert 'input' in inputs_info, "Test error: inputs_info must contain `input`"
+        input_shape = inputs_info['input']
+        input_type = self.input_type
+        inputs_data = {}
+        input_data = np.random.randint(-20, 20, input_shape).astype(input_type)
```
gkrivor, line 16: We need seriously think about this place. I strongly suggest to refactor original implementation.

It was an exceptional way to set an input in some unusual situations. Now I can see an overloading of _prepate_input in each PR. It looks like strange.

We may just introduce a dictionary to the parent class. Like inputs["name"] = [lval, rval, input_type] with default behavior in case a redefinition is absent for exact input.

tests/layer_tests/tensorflow_tests/test_tf_Bucketize.py
```diff
@@ -4,90 +4,43 @@
...
+        self.input_type = input_type
         tf.compat.v1.reset_default_graph()
         with tf.compat.v1.Session() as sess:
-            x = tf.compat.v1.placeholder(input_type, input_shape, 'Input')
-            constant_value = np.arange(-boundaries_size * 5, boundaries_size * 5, 10,
-                                       dtype=np.float32)
-            # TODO: Bucketize is not tested here. Need to re-write the test
+            input = tf.compat.v1.placeholder(input_type, input_shape, 'input')
+            # generate boundaries list
+            boundaries = np.sort(np.unique(np.random.randint(-200, 200, [boundaries_size]).astype(np.float32))).tolist()
```
gkrivor, line 26: And we need to think about leaving comments near magic numbers. Why did you choose range -200..200? Is it meaningful?..

tests/layer_tests/tensorflow_tests/test_tf_Bucketize.py
rkazants: I would not like to add additional comment with a link to the operation because it is easy to find.

tests/layer_tests/tensorflow_tests/test_tf_Bucketize.py
rkazants: This is not a task of this activity. The most relevant thing is to have code coverage. For now, it does not create difficulties to create a custom function with data generation.

tests/layer_tests/tensorflow_tests/test_tf_Bucketize.py
rkazants: This is a good question, I will add explanation about logic to select boundaries in separate PR.
Let me describe the logic here. I generate boundaries in way that there will be boundaries numbers laying out of range for `input data` and inside this range. This is done to potentially cover all possible cases where buckets can be.

mvafin (approved):

tests/layer_tests/tensorflow_tests/test_tf_Bucketize.py
rkazants: Fixed here: https://github.com/openvinotoolkit/openvino/pull/16619

PR #16558, opened 2023-03-27
Remove set_element_type() from TemporaryReplaceOutputType class

zhaixuejun1993:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/tests/functional/shared_test_classes/src/subgraph/memory_LSTMCell.cpp
+++ b/src/tests/functional/shared_test_classes/src/subgraph/memory_LSTMCell.cpp
@@ -114,8 +114,17 @@ namespace SubgraphTestsDefinitions {
         auto out_cell = tensor_iterator->get_iter_value(C_o, -1);
 
 
-        out_hidden.get_tensor().set_element_type(ngPrc);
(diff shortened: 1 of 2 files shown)


src/tests/functional/shared_test_classes/src/subgraph/memory_LSTMCell.cpp
```diff
@@ -114,8 +114,17 @@ namespace SubgraphTestsDefinitions {
         auto out_cell = tensor_iterator->get_iter_value(C_o, -1);
 
 
-        out_hidden.get_tensor().set_element_type(ngPrc);
-        out_cell.get_tensor().set_element_type(ngPrc);
+        // out_hidden.get_tensor().set_element_type(ngPrc);
+        // out_cell.get_tensor().set_element_type(ngPrc);
+        if (auto out_hidden_parameter = dynamic_cast<ov::op::v0::Parameter*>(out_hidden.get_node())) {
```
ilyachur, line 119: It won't work... Please correct me if I am wrong, but `out-hidden` cannot be a parameter.
In order to avoid this problem, you need to set the right shapes for the model parameters and call validate and infer types to propagate shapes

PR #16571, opened 2023-03-27
[Transforamtions] NonZero horizontal fusion

v-Golubev:
### Details:
Horizontal fusion of equal NonZero layers with the same parent.

### Tickets:
 - *102297*
 - *106820*

### DIFF
--- a/src/common/transformations/include/transformations/common_optimizations/nonzero_fusion.hpp
+++ b/src/common/transformations/include/transformations/common_optimizations/nonzero_fusion.hpp
@@ -1,0 +1,28 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include <memory>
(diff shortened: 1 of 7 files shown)


t-jankowski (approved):

src/common/transformations/include/transformations/common_optimizations/nonzero_fusion.hpp
```diff
@@ -0,0 +1,28 @@
...
+class TRANSFORMATIONS_API NonZeroFusion;
+
+}  // namespace pass
+}  // namespace ov
+
+/**
+ * @ingroup ie_transformation_common_api
+ * @brief NonZeroFusion transformation makes horizontal fusion for equal NonZero layers
+ */
+class ov::pass::NonZeroFusion : public ov::pass::MatcherPass {
```
itikhono, line 24: NonZeroHorizontalFusion would be better.
Usually we use `fusion` when there is a pattern for folding.

PR #16573, opened 2023-03-27
[PT FE]: revert usage mo.convert_model in tests

eaidova:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


tests/layer_tests/pytorch_tests/pytorch_layer_test_class.py
```diff
@@ -181,6 +175,7 @@ def convert_directly_via_frontend(self, model, example_input, trace_model, dynam
                 model = torch.jit.trace(model, example_input)
             else:
                 model = torch.jit.script(model)
+        print(model)
```
mvafin, line 178: do we need this here?

eaidova: fixed

tests/layer_tests/pytorch_tests/pytorch_layer_test_class.py
```diff
@@ -181,6 +174,7 @@ def convert_directly_via_frontend(self, model, example_input, trace_model, dynam
                 model = torch.jit.trace(model, example_input)
             else:
                 model = torch.jit.script(model)
+        print(model.inlined_graph)
```
rkazants, line 177: please remove this print. Not sure that this is needed.

mvafin: This is instead of line 63, this is nice to have thing for debugging test fails. I don't think it affects performance, because it is only printed when tests fail

mvafin (approved):

tests/layer_tests/pytorch_tests/pytorch_layer_test_class.py
vurusovs: If it's required to leave some message, please, use `logging` because `pytest` ignores any prints by default (but may be configured) and stores logs in it's log handlers

PR #16574, opened 2023-03-27
[ie transformations] improve SoftMax fusion for better mixed precision inference

pavel-esir:
### Details: 
- This PR includes 2 changes: Softmax fusion improvements and also more precise mixed precision marking for normalization nodes. Details are below:

1. SoftMax fusion transformation was not fusing simpler pattern (graph below)
 ```mermaid
graph TD;
    input-->Exp;
    Exp-->ReduceSum;
    Exp-->Div;
    ReduceSum-->Div;
```
and combination of Exp and ReduceSum was marked by mixed precision transformations and was inferred in FP32, which slowed performance. After fusing in into SoftMax it was not marked and performance improved without accuracy problems.

2. Normalization ops marking changes. Previously for normalization nodes adjacent nodes are also marked as precision sensitive, but it turns out that this gives no significant accuracy improvements but only slows performance. Only normalization node should be performed in FP32 precision. Graphs below.
**Before**:
 ```mermaid
graph TD;
    input-->Convert_to_FP32;
    Convert_to_FP32-->Add;
    Const-->Add;
    Add-->Slice;
    Slice-->MVN;
    MVN-->Mul;
    Mul-->Convert_to_FP16;
```

**After**: please note that here Convert is placed right before and after normalization node MVN. Lesser nodes are marked and perf improves.
 ```mermaid
graph TD;
    input-->Add;
    Const-->Add;
    Add-->Slice;
    Slice-->Convert_to_FP32;
    Convert_to_FP32-->MVN;
    MVN-->Convert_to_FP16;
    Convert_to_FP16-->Mul;
```

### Tickets:
 - xxx-107115, xxx-105064

### CODE REVIEW

File: src/common/transformations/src/transformations/common_optimizations/softmax_fusion.cpp [hunk 1]
```diff
@@ -8,63 +8,74 @@
...
     auto div_pattern = ngraph::pattern::wrap_type<opset6::Divide>({exp_pattern, reduce_sum_pattern});
 
-    ov::matcher_pass_callback callback = [=](pattern::Matcher& m) {
+    ov::matcher_pass_callback callback = [=](pass::pattern::Matcher& m) {
         if (transformation_callback(m.get_match_root()))
             return false;
 
         const auto& pattern_map = m.get_pattern_value_map();
+        const auto& pshape = pattern_map.at(data_pattern).get_partial_shape();
+        auto rank = pshape.rank().get_length();
```
mvafin, line 38: Please verify that rank is static

src/common/transformations/src/transformations/common_optimizations/softmax_fusion.cpp
```diff
@@ -8,63 +8,74 @@
...
         auto reduce_sum_axes =
             std::dynamic_pointer_cast<opset6::Constant>(pattern_map.at(reduce_sum_axes_pattern).get_node_shared_ptr());
         if (!reduce_sum_axes || shape_size(reduce_sum_axes->get_shape()) != 1)
             return false;
-
-        int64_t reduce_max_axis = reduce_max_axes->cast_vector<int64_t>()[0];
         int64_t reduce_sum_axis = reduce_sum_axes->cast_vector<int64_t>()[0];
-        if (reduce_max_axis < 0 || reduce_sum_axis < 0) {
-            const auto& pshape = pattern_map.at(data_pattern).get_partial_shape();
-            auto rank = pshape.rank().get_length();
```
mvafin, line 46: Rank is only needed when any of axes are negative, could you do that too?

mvafin (changes requested):
Did you verify that both patterns produce correct results? Could you explain why `softmax(x) == softmax(max(x) - x)`

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
```diff
@@ -191,6 +192,7 @@ void TransformationsPipeline::apply(std::shared_ptr<ov::Model> func) {
         type_to_fuse_map empty_fuse_map = {};
         manager.register_pass<ov::pass::Validate>();
         //  call ConvertPrecision with keep_precision_sensitive_in_fp32 = true
+        manager.register_pass<ov::pass::SoftmaxFusion>();
```
mvafin, line 195: Why you add this transformation for GPU pipeline? Is it not added in common pipeline?

src/common/transformations/src/transformations/common_optimizations/softmax_fusion.cpp
pavel-esir: we already ensured that data shape is static in pattern matcher above
https://github.com/openvinotoolkit/openvino/blob/4b10df4f1810327a15df10f6e118e21c0570d0f5/src/common/transformations/src/transformations/common_optimizations/softmax_fusion.cpp#L20

src/common/transformations/src/transformations/common_optimizations/softmax_fusion.cpp
pavel-esir: Done. Now get the rank is obtained only when it's needed (axis < 0).

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
pavel-esir: By default it was called inside `MOCTransformations`, and `MOCTransformations` were called inside `CommonOptimizations`, but `CommonOptimizations` are called in the end of pipeline. Therefore I called it here in advance to get all softmax pattern fused before the `ConvertPrecision` is run.

Do you think we should add a comment why we call this fusing in advance?

mvafin: Yes, please add comment. But still, is it really needed? My expectation was that MOC pipeline was run before GPU pipeline. In `convert_model` or `read_model`

pavel-esir:
> Did you verify that both patterns produce correct results? Could you explain why `softmax(x) == softmax(x - max(x))`

yes, i ensured that `softmax(x) == softmax(max(x) - x)`, algebraically they are equal.
```
softmax(max(x) - x) = exp(x - max(x))/reducesum(exp(x - max(x))) 
= exp(x) * exp(-max(x))/reducesum(exp(x) * exp(-max(x))) -> we can take exp(-max(x)) out of brackets because it's a constant
= exp(x) * exp(-max(x))/[ exp(-max(x))*reducesum(exp(x))] ->we can cancel in the numerator and denominator exp(-max(x))
= exp(x) / [reducesum(exp(x))]
```
also i playes a bit in numpy and got the same results. As far as i understood this -max(x) prevents from overflows. Actually i am surprised why our transformation covered more complex case with -max(x) but lacked the simpler direct definition of Softmax :)

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
rkazants: Honestly, looks like WA here. Can we have the same situation for other fusion transformations which are affected by `ConvertPrecision`? May we should have more general approach to solve such problems like ignoring some `Convert` nodes in the patterns? Or can you remind why `ConvertPrecision` is called after `CommonOptimizations`? 

fyi, @slyalin, @vladimir-paramuzov

rkazants (commented):
Several comments are left. Other looks good to me. Need review from the GPU.

src/common/transformations/src/transformations/common_optimizations/softmax_fusion.cpp
```diff
@@ -8,63 +8,77 @@
...
                 pattern_map.at(exp_pattern).get_node_shared_ptr(),
                 pattern_map.at(reduce_sum_pattern).get_node_shared_ptr(),
                 div,
             },
             softmax);
+        if (exp_input_is_subtract) {
+            copy_runtime_info({pattern_map.at(reduce_max_pattern).get_node_shared_ptr(),
+                               pattern_map.at(sub_pattern).get_node_shared_ptr()},
+                              softmax);
+        }
```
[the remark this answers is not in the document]
pavel-esir: Hmm, i thought that if we call `copy_runtime_info` 2 times separately result will be the same as if we accumulate list of all nodes and then pass to `copy_runtime_info`. But after looking to `rt_info.cpp` i am not so sure. You proposal is safer. I will update

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
pavel-esir: We call `ConvertPrecision` before `CommonOptimization` so that pattern matcher is not broken. Because if we call CommonOptimization earlier then on a bunch of operation there will be Cost+Decompression Convert instead of a single Const. @vladimir-paramuzov confirmed that personally.

So it looks like this w/a is inevitable.

slyalin: @itikhono, please share your opinion.

@mvafin, `read_model` doesn't call MOC. This is one of the desired things -- to call MOC in read_model -- from consistency perspective to align convert_model and read_model behaviors. But, as @mateusztabaka showed in his measurements, we cannot do it immediately without significantly regressing in read_time or without significant investments in the reoragnization of the transformation pipeline.

But even if we call it in read_model, the current convention is that during compile_model we shouldn't rely on that, because we don't really know where the model come from and whether it was modified manually. So it can require a full optimization pipeline to be applied.

ConvertPrecision is called in that place because "Const+Decompress patterns will be instead of just Consts and some pattern matchers potentially could not match". Just because nobody knowns how many transformation should be modified to support Const+Decompress we are calling it in this way. It looks to be much bigger topic that requires more dedicated analysis among all the transformation. Nobody committed to investigate part.

Looks like we only this transformation called twice. So this doesn't look to have major impact. In comparison to the bigger task described above, this is only thing that we can do considering amount of resources involved.

But the picture actually involves more details from plugins side, especially for big models where we are trying to minimize touching big weights in the pipeline. Should be discussed offline.

src/common/transformations/src/transformations/common_optimizations/softmax_fusion.cpp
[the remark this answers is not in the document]
pavel-esir: done

src/common/transformations/src/transformations/common_optimizations/mark_subgraphs_to_keep_in_mixed_precision.cpp
```diff
@@ -337,18 +337,18 @@ bool MarkSugraphsToKeepInMixedPrecision::run_on_model(const shared_ptr<ov::Model
...
     // both Up and Down propagations are needed.
     // Why both of them are needed is explained in comments in passes declarations.
     REGISTER_PASS(manager, PropagateDownMarkToKeepInMixedPrecision)
+
     auto propagate_up = manager.register_pass<BackwardGraphRewrite>();
     ADD_MATCHER(propagate_up, PropagateUpMarkToKeepInMixedPrecision)
 
     // Mark nodes in ShapeOf subgraphs to keep in FP32
     REGISTER_PASS(manager, MarkPrecisionSensitiveShapeOfSubgraphs)
+    REGISTER_PASS(manager, MarkNormalizationOps)
```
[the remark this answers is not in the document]
pavel-esir: If we mark normalization nodes and then propagate this marking down and up to adjacent elementwise nodes we will keep in fp32 more nodes than necessary. If we mark them in the end only after down and up propagation marking only single nod normalization will be marked.

The reason why we had this propagate down and up was that we don't know where to stop precision sensitivity pattern, therefore we marked all neighbors with elementwise, Slice, Split and so on. But with MVN and NormalizeL2 this is not necessary we know exactly when sensitivity starts and when ends.

For comparison, with the old approach we kept more node in fp32 (outlined by red rectangle), with the new only single MVN/NormalizeL2 node is kept in fp32 and it is enough to save accuracy.
image

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
rkazants: Also, why don't we call it for CPU plugin? It seems to be also affected.

rkazants (approved):
Let us discuss the right place/approach for fusing transformations offline

pavel-esir:
Windows for some reason is failing. Merging the latest master

mvafin (approved):

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
itikhono: Ok, let's leave it at that for now.
Could you create a ticket to delete this call and mention the ticket number in this comment?

pavel-esir:
Layer tests for reduce fail because of the requirements. Starting from `onnx==1.13.1` is has `axes` `input` instead of and attribute as in previous versions, and even the latest `onnxruntime==1.14.1` is not aware of that therefore then layer test try to compare with reference onnxruntime it fails. 

To get the green precommit, need to downgrade onnx requirement for layer tests or (mark this test as skipped. The latter is preferable

pavel-esir:
reduce op layer tests just have been skipped here https://github.com/openvinotoolkit/openvino/pull/16730

PR #16586, opened 2023-03-27
[GPU] Fix malfunction in crop static kernel in dynamic shape scenario

yeonbok:
### Details:
 - Static crop kernels are created with wrong params in async compilation in dynamic shape scenario
 - This PR is needed to fix accuracy drop for dynamic shape


yeonbok:
(Corresponding test is to be added soon)

ahnyoung-paul (approved):

PR #16590, opened 2023-03-28
[TF FE] Move to TopK-11 operation and update downgrading TopK transformation

rkazants:
**Details:** Move to TopK-11 operation and updare the downgrading transformation for TopK-11. Even if `stable` mode is set, we need to donwgrade to `TopK-3` to provide backward-compatibility and plugins that does not support stable mode.

**Ticket:** 105421


src/common/transformations/src/transformations/op_conversions/convert_topk11_downgrade.cpp
```diff
@@ -18,11 +18,13 @@ ov::pass::ConvertTopK11ToTopK3::ConvertTopK11ToTopK3() {
 
     const matcher_pass_callback callback = [=](pattern::Matcher& m) {
         const auto topk_v11 = std::dynamic_pointer_cast<opset11::TopK>(m.get_match_root());
-        if (!topk_v11 || topk_v11->get_stable() || transformation_callback(topk_v11)) {
+        if (!topk_v11 || transformation_callback(topk_v11)) {
```
tomdol, line 21: As agreed this check should be modified and take the second output into account

rkazants: We agreed to downgrade anyway, just to avoid risk with affection to different plugins.

tomdol (approved):

mvafin (approved):

PR #16593, opened 2023-03-28
[GPU] Shape agnostic ref kernels implementation for convolution

kelvinchoi-intel:
### Details:
 - *Implement shape agnostic ref kernel for convolution*

### Tickets:
 - *106809*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.h
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.h
@@ -257,13 +257,15 @@ inline kernel_impl_params canonicalize_fused_shapes(const kernel_impl_params& im
 
     for (auto& fd : updated_impl_params.fused_desc) {
         if (fd.is_type<eltwise>() && fd.total_num_deps == 2 && fd.has_outer_dep()) {
-            auto out_pshape = updated_impl_params.output_layouts[0].get_partial_shape();
+            if (updated_impl_params.input_layouts.size() > size_t(fd.outer_dep_start_idx)) {
+                auto out_pshape = updated_impl_params.output_layouts[0].get_partial_shape();
 
-            auto& dep_layout = updated_impl_params.input_layouts[fd.outer_dep_start_idx];
-            auto dep_shape = dep_layout.get_partial_shape();
+                auto& dep_layout = updated_impl_params.input_layouts[fd.outer_dep_start_idx];
+                auto dep_shape = dep_layout.get_partial_shape();
 
-            if (!broadcastable(dep_shape, out_pshape, use_new_shape_infer)) {
-                dep_layout.set_partial_shape(extend_shape_to_rank_from_begin(dep_shape, out_pshape.size()));
+                if (!broadcastable(dep_shape, out_pshape, use_new_shape_infer)) {
+                    dep_layout.set_partial_shape(extend_shape_to_rank_from_begin(dep_shape, out_pshape.size()));
+                }
             }
         }
     }
--- a/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/convolution_gpu_ref.cl
+++ b/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/convolution_gpu_ref.cl
@@ -6,6 +6,7 @@
 #include "include/batch_headers/fetch_weights.cl"
 
 KERNEL(kernel_name)(
+    OPTIONAL_SHAPE_INFO_ARG
     const __global INPUT0_TYPE *conv_input,
     __global OUTPUT_TYPE *output,
     const __global FILTER_TYPE *weights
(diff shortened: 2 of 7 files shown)


src/plugins/intel_gpu/src/kernel_selector/cl_kernels/convolution_gpu_ref.cl
```diff
@@ -62,7 +63,7 @@ KERNEL(kernel_name)(
 
     for (uint k = 0; k < FILTER_IFM_NUM; ++k)
     {
-#if INPUT0_SIZE > 4
+#if INPUT0_DIMS > 4
```
Lyamin-Roman, line 66: Maybe make a equal names for the ranks in static and dynamic cases? Now it's `_DIMS`, but for not to change in all the following agnostic kernels `_SIZE` -> `_DIMS` I suggest leaving one old name `_SIZE` and return the name back in the already changed kernels.

Lyamin-Roman (approved):
In general LGTM, but if this PR is merged after #16690, then there will be conflicts

src/plugins/intel_gpu/src/kernel_selector/kernels/convolution/convolution_kernel_base.cpp
```diff
@@ -77,6 +77,12 @@ JitConstants ConvolutionKernelBase::GetJitConstants(const convolution_params& pa
             mem_consts.AddConstants({MakeJitConstant("BILINEAR_INTERPOLATION_PAD", params.bilinear_interpolation_pad)});
     }
 
+    if (!params.is_shape_agnostic) {
+        if (params.outputs[0].Batch().v == 1) {
+            mem_consts.AddConstant(MakeJitConstant("OUTPUT_BATCH_NUM_IS_1", 1));
```
vladimir-paramuzov, line 82: Can we have something generic instead of that? I think macro like
`${TENSOR_NAME}_IS_DYNAMIC` and/or `${DIM_NAME}_IS_DYNAMIC` can help to handle such cases

src/plugins/intel_gpu/src/kernel_selector/cl_kernels/convolution_gpu_ref.cl
vladimir-paramuzov: looks like `_SIZE` and `_DIMS` macros always have the same value `t.GetDims().size()` regardless tensor type, `_SIZE` macro is just not defined for dynamic tensors. So I'd say that this is a good change, but I think we also need to remove `_SIZE` macro at all from jitter as at the first glance it's used in conv ref kernel only.

kelvinchoi-intel: I've deleted creating _SIZE from jitter because this is only used on conv ref kernel.

src/plugins/intel_gpu/src/kernel_selector/kernels/convolution/convolution_kernel_base.cpp
kelvinchoi-intel: This jitconstant is used to make b = 0 situation in kernel. I've changed the jitter name from "OUTPUT_BATCH_NUM_IS_1" to "SKIP_BATCH" to avoid misunderstanding.

src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.h
```diff
@@ -255,13 +255,15 @@ inline kernel_impl_params canonicalize_fused_shapes(const kernel_impl_params& im
 
     for (auto& fd : updated_impl_params.fused_desc) {
         if (fd.is_type<eltwise>() && fd.total_num_deps == 2) {
-            auto out_pshape = updated_impl_params.output_layouts[0].get_partial_shape();
+            if (updated_impl_params.input_layouts.size() > size_t(fd.outer_dep_start_idx)) {
```
yeonbok, line 258: Now there is fd.has_outer_dep()

vladimir-paramuzov:
@kelvinchoi-intel Please resolve conflicts. 

Also I'm curious about perf. I think in some cases runtime compilation + static optimized kernel may be faster than ref shape agnostic impl. So could you check it?

src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.h
kelvinchoi-intel: Got it! But still line 258 condition is necessary because of line 263.

yeonbok:
> @kelvinchoi-intel Please resolve conflicts.
> 
> Also I'm curious about perf. I think in some cases runtime compilation + static optimized kernel may be faster than ref shape agnostic impl. So could you check it?

1)Actually this work is a base work for next enabling of blocked format. 
This kernel will not be actually used for now, because the layout optimizer is enforcing blocked format (fsv16) for convolution layer if it is dynamic.
We'll check the performance next when we enable blocked format. 

2) But when we check the perf it won't be simply said that which one is better, because if the tensor is small the ref is better, and if the tensor is quite large, compiling opt kernel might be better. 
Also for onednn, compiling new kernel will be better than ref kernel definitely. (We'll need to add such condition in primitive_inst.)

3) Finally we'll need to have to consider for CNNs 
- improve layout selection policy 
- multiple shape agnostic kernels of ref+opt kernels => select best one at runtime (or not to use shape agnostic kernel)

PR #16602, opened 2023-03-28
[GNA] Fixed cases when FQ is not the 1st layer

mryzhov:
### Details:
 - Fixed issue when FQ layer connected to Input via non functional layer (reshape, transpose, squeeze, unsqueeze)

### Tickets:
 - 107336


src/plugins/intel_gna/src/layers/gna_layer_info.hpp
```diff
@@ -349,6 +349,12 @@ class LayerInfo {
     bool isNonValuesChangable() const {
         return isNonFunctional() || isSplit() || isSlice() || isConcat();
     }
+    bool is_gather() const noexcept {
+        return isOfType("gather");
+    }
+    bool is_fq_non_sensitive() const noexcept {
```
sirzabek, line 355: Shouldn't we use isNonFunctional instead and just extend it?

mryzhov: gather and permute are not non-functional layers, it is better to isolate them in separate condition

mryzhov: And isNonFunctional() check a bit complicated and we will not call it at all in the most cases in is_fq_non_sensitive()

evkotov (approved):

sirzabek (approved):

PR #16603, opened 2023-03-28
[GPU] Refactor dimensions jitter and max rank related code

vladimir-paramuzov:
### Details:
 - defined max rank in dedicated function instead of magic const in multiple places
 - used `get_default_format` and `is_default_format` instead of if/elseif sequences
 - added helper for dynamic dimensions code generation


yeonbok (approved):

yeonbok:
LGTM for the baseline idea, but some unittests are failing (smoke_ExperimentalDetectronPriorGridGenerator_f16/xx)

PR #16604, opened 2023-03-28
[TF FE] Support frozen models in text protobuf format aka pbtxt

rkazants:
**Details:** Support frozen models in Text Protobuf format aka pbtxt. Switched off WA in the MO tool.

**Ticket:** TBD


src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -88,7 +89,8 @@ bool FrontEnd::supported_impl(const std::vector<ov::Any>& variants) const {
             // for automatic deduction of the frontend to convert the model
             // we have more strict rule that is to have `.pb` extension in the path
             return true;
-        } else if (GraphIteratorSavedModel::is_supported(model_path)) {
```
mvafin, line 91: Why you delete `SavedModel` check?

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -101,7 +103,8 @@ bool FrontEnd::supported_impl(const std::vector<ov::Any>& variants) const {
...
```
mvafin, line 104: Same

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -101,7 +103,8 @@ bool FrontEnd::supported_impl(const std::vector<ov::Any>& variants) const {
...
+        } else if (GraphIteratorProtoTxt::is_supported(model_path)) {
```
rkazants, line 106: accidentally removed SavedModel support. Need to recover

mvafin (changes requested):

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -93,6 +94,10 @@ bool FrontEnd::supported_impl(const std::vector<ov::Any>& variants) const {
         } else if (GraphIteratorSavedModel::is_supported(model_path)) {
             return true;
         }
```
rkazants, line 96:
```suggestion
```

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -106,6 +110,10 @@ bool FrontEnd::supported_impl(const std::vector<ov::Any>& variants) const {
...
```
rkazants, line 112:
```suggestion
```

src/frontends/tensorflow/src/frontend.cpp
rkazants: recovered

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -116,7 +123,7 @@ bool FrontEnd::supported_impl(const std::vector<ov::Any>& variants) const {
 }
 
 ov::frontend::InputModel::Ptr FrontEnd::load_impl(const std::vector<ov::Any>& variants) const {
-    // TODO: Support other TensorFlow formats: SavedModel, .meta, checkpoint, pbtxt
+    // TODO: Support other TensorFlow formats: SavedModel, .meta, checkpoint
```
gkrivor, line 126: I didn't see passed results with existing models to declare it is supported.

Could you share, please?

src/frontends/tensorflow/src/graph_iterator_proto_txt.hpp
```diff
@@ -0,0 +1,52 @@
...
+        if (!model_exists) {
+            return false;
+        }
+        auto input_stream = std::make_shared<::google::protobuf::io::IstreamInputStream>(&pbtxt_stream);
+        if (!input_stream) {
+            return false;
+        }
+        auto graph_def = std::make_shared<::tensorflow::GraphDef>();
+        auto is_parsed = ::google::protobuf::TextFormat::Parse(input_stream.get(), graph_def.get());
+        return is_parsed;
```
gkrivor, line 47: As I know it doesn't cover pbtxt with checkpoints. As I understand it is most common use case. Add this limitation in the check.

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -92,6 +93,9 @@ bool FrontEnd::supported_impl(const std::vector<ov::Any>& variants) const {
             return true;
         } else if (GraphIteratorSavedModel::is_supported(model_path)) {
             return true;
+        } else if (GraphIteratorProtoTxt::is_supported(model_path)) {
```
gkrivor, line 96: Reduce complexity and keep same logic as your already used, check ends_with(".pbtxt") first. As I can see from code above (for binary protobuf) we doesn't expect "any file which might be interpreted as a text protobuf".

src/frontends/tensorflow/src/graph_iterator_proto_txt.hpp
```diff
@@ -0,0 +1,52 @@
...
+namespace frontend {
+namespace tensorflow {
+
+class GraphIteratorProtoTxt : public GraphIteratorProto {
+public:
+    template <typename T>
+    GraphIteratorProtoTxt(const std::basic_string<T>& path) : GraphIteratorProto() {
+        std::ifstream pbtxt_stream(path, std::ios::in);
+        FRONT_END_GENERAL_CHECK(pbtxt_stream && pbtxt_stream.is_open(), "Model file does not exist");
+        auto input_stream = std::make_shared<::google::protobuf::io::IstreamInputStream>(&pbtxt_stream);
```
gkrivor, line 24: Strange place. Few lines below you check an input (!input_stream - error). Check here is missing.

src/frontends/tensorflow/tests/test_models/gen_wrapper.py
```diff
@@ -19,16 +20,14 @@
...
-        graph_def = tf.GraphDef()
-        text_format.Merge(f.read(), graph_def)
-        tf.import_graph_def(graph_def, name='')
-        tf.io.write_graph(graph_def, os.path.join(sys.argv[2], model_name), model_name + '.pb', False)
+    model_name = os.path.basename(model_pbtxt).split('.')[0]
+    dest_path = os.path.join(out_folder, model_name, model_name + '.pbtxt')
+    os.makedirs(os.path.dirname(dest_path), exist_ok=True)
+    try:
+        shutil.copy(model_pbtxt, dest_path)
+    except shutil.SameFileError:
```
gkrivor, line 29: why don't you overwrite? what if previous file is wrong (copy incomplete, broken, previous version)?

Looks like original implementation overwrites content each run.

gkrivor (changes requested):

src/frontends/tensorflow/src/frontend.cpp
rkazants: Please take a look at TF FE unit-tests: https://dev.azure.com/openvinoci/dldt/_build/results?buildId=528226&view=logs&j=982848f2-7f8d-5e57-730a-8df1e5f991ba&t=c36556a9-9217-5d3d-3f01-b110be4c2de5
Now most of them moved to pbtxt

Also have other MO unit-tests also working with pbtxt file: https://dev.azure.com/openvinoci/dldt/_build/results?buildId=528226&view=logs&j=982848f2-7f8d-5e57-730a-8df1e5f991ba&t=cf9ee82e-e056-500f-81f4-f6a3a79d25db

We have more than 10 tests for pbtxt format

src/frontends/tensorflow/src/graph_iterator_proto_txt.hpp
rkazants: When we introduce checkpoint, we need to update the classes `GraphIteratorProto` and `GraphIteratorProtoTxt` with new constructor accepting checkpoint path.
Now it is quite scalable solution for checkpoint.

src/frontends/tensorflow/src/frontend.cpp
rkazants: Agreed with @slyalin, we want to be unbound from file extension and to check it without this because we want to have MO with minimal options so that user can just specify only path to the model.

src/frontends/tensorflow/tests/test_models/gen_wrapper.py
rkazants: Good question, it turned out that some tests in Jenkins just place tests files to the same directory when src_path and dest_path match. So we don't do anything with it and handle exception from shutil.
If copy is incomplete, we have below logic that mark successful copying. In case of unsuccessful copying, this marker is not created and build procedure failures.

src/frontends/tensorflow/src/graph_iterator_proto_txt.hpp
rkazants: Good catch, let me fix it in the separate PR. Don't thing it is to critical for now.

rkazants: Fixed here: https://github.com/openvinotoolkit/openvino/pull/16619

PR #16610, opened 2023-03-28
Fix sporadic fails when beta==0 in baddmm

mvafin:
### Details:
 - *If beta is 0 torch.baddmm can produce nan in some cases*

### Tickets:
 - *CVS-107053*

### DIFF
--- a/tests/layer_tests/pytorch_tests/test_addmm.py
+++ b/tests/layer_tests/pytorch_tests/test_addmm.py
@@ -94,6 +94,5 @@ class TestBAddBMM(PytorchLayerTest):
     @pytest.mark.nightly
     @pytest.mark.precommit
     def test_baddbmm(self, kwargs_to_prepare_input, alpha, beta, ie_device, precision, ir_version):
-      for i in range(100):
         self._test(*self.create_model(alpha, beta), ie_device, precision, ir_version,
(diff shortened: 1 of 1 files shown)


eaidova (approved):

tests/layer_tests/pytorch_tests/test_addmm.py
```diff
@@ -83,8 +83,14 @@ def forward(self, m0, m1, m2):
 
     ])
     @pytest.mark.parametrize("alpha,beta",
-                             [(1., 1.), (0., 1.), (1., 0.), (1., 2.), (2., 1.), (-5., -6.), (3., 4.), (0.5, 0.75),
-                              (1, 1)])
+                             [  # beta==0 in some cases produce nan in pytorch
```
slyalin, line 86: Sounds weird. The doc https://pytorch.org/docs/stable/generated/torch.baddbmm.html just says that it should work fine. Pytorch bug?

mvafin: Yeah, looks like pytorch bug, because it works in ov

rkazants: looks regular matrices multiplication. How `nan` values can happen here?

PR #16613, opened 2023-03-28
[MO] Fix legacy If

mvafin:
### Details:
 - *Fix legacy If*

### Tickets:
 - *103311*


rkazants (commented):
can we have test for this please?

tools/mo/openvino/tools/mo/ops/If.py
```diff
@@ -278,11 +278,11 @@ def re_numerate_internal_id_and_get_if_id(if_node):
...
         then_graph_nodes = if_node.then_graph.nodes()
-        for idx in range(len(if_node.then_graph.get_op_nodes())):
-            then_graph_nodes[idx]['internal_layer_id'] = idx
+        for node in if_node.then_graph.get_op_nodes():
+            then_graph_nodes[node.id]['internal_layer_id'] = node.id
         else_graph_nodes = if_node.else_graph.nodes()
-        for idx in range(len(if_node.else_graph.get_op_nodes())):
-            else_graph_nodes[idx]['internal_layer_id'] = idx
+        for node in if_node.else_graph.get_op_nodes():
+            else_graph_nodes[node.id]['internal_layer_id'] = node.id
```
rkazants, line 285: don't we break IRs generated by the legacy MO?

mvafin: I double checked. It doesn't break legacy MO. I checked manually created tf model with `If` as well as e2e tests

rkazants (approved):

PR #16615, opened 2023-03-29
Optimize and fix fps number in log

songbell:
### Details:
 - cpu help fps log info is not correct
 - fix a potential threading issue

### Tickets:
 - *ticket-id*


src/plugins/auto/auto_schedule.cpp
```diff
@@ -676,23 +678,23 @@ void AutoSchedule::WaitActualNetworkReady() const {
 bool AutoSchedule::ScheduleToWorkerInferRequest(IE::Task inferPipelineTask, DeviceName preferred_device) {
     std::vector<DeviceInformation> devices;
     // AUTO work mode
+    // Devices that fail infer will be removed from the priority list in the callback, need lock here
+    std::unique_lock<std::mutex> lock(_autoSContext->_fallbackMutex);
```
songbell, line 682: here I remove the lock from if else block out, because all the ACTUALDEVICE, FALLBACK or ctput device info may been changed in the fallback logic, please review @WeldonWangwang

src/plugins/auto/auto_schedule.cpp
```diff
@@ -676,23 +678,23 @@ void AutoSchedule::WaitActualNetworkReady() const {
...
     if (!preferred_device.empty()) {
         if (_pCTPUTLoadContext) {
-            std::lock_guard<std::mutex> lock(_autoSContext->_fallbackMutex);
             devices = _autoSContext->_devicePriorities;
         } else {
             // if the device needed by customer is not ready, need to wait for it
             WaitActualNetworkReady();
             // the preferred_device should be the selected device in AUTO work mode
+            // or, if fallback enabled, the fallback device will not fit in the remote tensor as well
```
wangleis, line 690: How about `preferred_device` for remote tensor cannot work and fallback device is CPU?

src/plugins/auto/multi_schedule.cpp
```diff
@@ -244,16 +244,7 @@ MultiSchedule::~MultiSchedule() {
...
-                LOG_INFO_TAG("CPU_HELP:infer:%ld", _cpuHelpInferCount + count);
-                if (_cpuHelpFps > 0.0) {
-                    LOG_INFO_TAG("CPU_HELP:fps:%lf", _cpuHelpFps);
-                } else if (count >= 1) {
-                    std::chrono::duration<double, std::milli> durtation =
-                        reqAllEndTimes.back() - reqAllStartTimes.front();
-                    LOG_INFO_TAG("CPU_HELP:fps:%lf", count * 1000 / durtation.count());
-                }
-            } else {
+            if (_workerRequest.first != "CPU_HELP") {
```
wangleis, line 247: Could we update fps calculation for CPU_HELP based completed frames instead of totally drop it?

songbell: we re not dop it , the calc is in above code , current master cannot give correct cpu help fps

src/plugins/auto/auto_schedule.cpp
songbell: here is kind of confusing, when feeding a gpu remote tensor, and -d AUTO:vpux,cpu, currently we are throwing an exception directly, because I think with remote tensor, user are explicitly expecting a performance benefit. probably need to understand the real customer need, what's their expectation on this scenario.

yangwang201911 (approved):

src/plugins/auto/auto_schedule.cpp
```diff
@@ -676,23 +678,23 @@ void AutoSchedule::WaitActualNetworkReady() const {
...
     std::vector<DeviceInformation> devices;
     // AUTO work mode
+    // Devices that fail infer will be removed from the priority list in the callback, need lock here
+    std::unique_lock<std::mutex> lock(_autoSContext->_fallbackMutex);
     if (!preferred_device.empty()) {
         if (_pCTPUTLoadContext) {
-            std::lock_guard<std::mutex> lock(_autoSContext->_fallbackMutex);
             devices = _autoSContext->_devicePriorities;
         } else {
             // if the device needed by customer is not ready, need to wait for it
```
yangwang201911, line 687: should we unlock `lock` before calling `WaitActualNetworkReady()`?

songbell: updated:)

src/plugins/auto/auto_schedule.cpp
WeldonWangwang: Okay.

wangleis (approved):

PR #16618, opened 2023-03-29
Resolved failed unit-tests

byungilm:
+ Resolve TCs' failure of oneDNN tests

### Details:
 - Bugfix in convolution_gpu_mmad_b_fs_yx_fsv32 kernel (invalid index calculation)
 - depthwise_conv_b_fs_yx_fsv16 tested a given clDNN kernel. But it could not be applied for oneDNN execution.
   Added a logic for oneDNN execution in the test.
 - Added func to check possible implementation for a certain imple type (e.g. impl_types::onednn)
 - Resolve TCs' failure of oneDNN tests

### Tickets:
 - 107829, 67491

### DIFF
--- a/src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
+++ b/src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
@@ -1286,7 +1286,7 @@ bool layout_optimizer::is_primitive_implemented_for_onednn(program_node& node) {
     if (node.is_type<fully_connected>() || node.is_type<gemm>() || node.is_type<pooling>() ||
         node.is_type<convolution>() || node.is_type<deconvolution>() ||
         node.is_type<reduce>() || node.is_type<reorder>() || node.is_type<concatenation>()) {
-            return true;
+        return true;
     }
 
     return false;
--- a/src/plugins/intel_gpu/tests/test_cases/fully_connected_gpu_test.cpp
+++ b/src/plugins/intel_gpu/tests/test_cases/fully_connected_gpu_test.cpp
@@ -1591,10 +1591,10 @@ INSTANTIATE_TEST_SUITE_P(
     fully_connected_i8_u8_test,
     testing::Combine(
         testing::Values(1, 2),
-        testing::Values(3, 64),
+        testing::Values(16, 64),
         testing::Values(1),
         testing::Values(1),
-        testing::Values(3, 32),
+        testing::Values(16, 32),
         testing::Values(format::bfyx, format::b_fs_yx_fsv4, format::b_fs_yx_fsv16, format::b_fs_yx_fsv32)
     ),
     fully_connected_i8_u8_test::PrintToStringParamName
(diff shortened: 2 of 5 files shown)


src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
```diff
@@ -1279,10 +1280,9 @@ bool layout_optimizer::are_layouts_suitable_for_onednn(program_node& node) {
 }
 
 bool layout_optimizer::is_primitive_implemented_for_onednn(program_node& node) {
-    if (node.is_type<fully_connected>() || node.is_type<gemm>() || node.is_type<pooling>() ||
-        node.is_type<convolution>() || node.is_type<deconvolution>() ||
-        node.is_type<reduce>() || node.is_type<reorder>() || node.is_type<concatenation>()) {
-            return true;
+    if (node.type()->does_possible_implementation_exist(node, *(node.get_kernel_impl_params()), cldnn::impl_types::onednn) ||
+        node.is_type<deconvolution>()) {
```
isanghao, line 1284: why is it necessary to handle deconv separately?

src/plugins/intel_gpu/tests/test_cases/fully_connected_gpu_test.cpp
```diff
@@ -1591,10 +1591,10 @@ INSTANTIATE_TEST_SUITE_P(
     fully_connected_i8_u8_test,
     testing::Combine(
         testing::Values(1, 2),
-        testing::Values(3, 64),
+        testing::Values(16, 64),
         testing::Values(1),
         testing::Values(1),
-        testing::Values(3, 32),
+        testing::Values(16, 32),
```
isanghao, line 1597: could you describe more in the commit message why it is changed?

isanghao (changes requested):
Please see the comments.

src/plugins/intel_gpu/tests/test_cases/fully_connected_gpu_test.cpp
byungilm: As discussed, I added some  description about this change.

src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
byungilm: deconv has onednn impl. But does_possbile_implementation_exist returns true for checkking onednn impl of deconv. I just added condition for now. I will check root cause

src/plugins/intel_gpu/tests/test_cases/fully_connected_gpu_test.cpp
hyunback: NIT: Maybe you change the shape, because some specific shape has the issue in oneDNN side, if true, could you add a comment in failed case? It should be help to meet unexpected error in the later.

hyunback (approved):

src/plugins/intel_gpu/tests/test_cases/fully_connected_gpu_test.cpp
byungilm: At least, I was fine on benchdnn execution. A relevant comment is added to commit message.

isanghao (approved):

PR #16626, opened 2023-03-29
remove deprecated MO args

pavel-esir:
### Details:
- since removing deprecated was reverted in #16555, in this PR we again remove deprecated args, taking into account the latest changes.

### Tickets:
 - xxx-100890, xxx-106810


tests/layer_tests/common/utils/common_utils.py
```diff
@@ -26,6 +26,10 @@ def generate_ir(coverage=False, **kwargs):
             params.extend(("-b", str(value)))
         elif key == "k":
             params.extend(("-k", str(value)))
+        # for FP32 set explicitly compress_to_fp16=False,
```
rkazants, line 29: It looks that we can run with `FP16` or `FP32` mode using two ways by setting `compress_to_fp16` or `precision`? Do we need both ways? Can they contradict to each other?

pavel-esir: They cannot contradict, layer tests have only one `precision` argument, `compress_to_fp16` is internal and it's set based on precision: 
https://github.com/openvinotoolkit/openvino/blob/67afe170d2652f37e7f2a0af3f7fb04825073d11/tests/layer_tests/common/layer_test_class.py#L43-L47

user cannot run layer tests with defined compress_to_fp16 only, it should use precision

rkazants: Got it.

rkazants (approved):

PR #16639, opened 2023-03-29
[Transforamtions] NonZero horizontal fusion: review leftovers

v-Golubev:
Main PR: #16571


v-Golubev:
@itikhono @t-jankowski Could you please confirm that the review comments from the main PR are corrected as expected? Thanks

t-jankowski (approved):

src/common/transformations/src/transformations/common_optimizations/nonzero_horizontal_fusion.cpp
```diff
@@ -24,8 +24,9 @@ ov::pass::NonZeroFusion::NonZeroFusion() {
 
         bool status = false;
         auto replace_if_nodes_match = [&](const ov::Input<ov::Node>& in) {
-            auto cur_nonzero = ov::as_type_ptr<ov::opset10::NonZero>(in.get_node()->shared_from_this());
-            if (cur_nonzero && cur_nonzero->get_output_type() == out_prc) {
+            auto in_node = in.get_node()->shared_from_this();
```
itikhono, line 27: why do you need to call ->shared_from_this()
I believe you can use `in_node` itself without .get()

v-Golubev: Yes, you are right. Corrected, thanks

PR #16648, opened 2023-03-30
Fix ITT build fail

pereanub:
### Details:
OpenVino build fails when enabling the ITT Traces flag with the following message:
_**C:\Users\bpereanu\WORK\OpenVinoSources\openvino\src\plugins\template\src\plugin.cpp(166): error C2653: 'TemplatePlugin': is not a class or namespace name**_

Still waiting for the build to finish, I will update the PR when it finishes and can confirm that this fixes the issue.

### Tickets:
 - *C[VS-107186](https://jira.devtools.intel.com/browse/CVS-107186)*


ilyachur (approved):

pereanub:
The build finished successfully without other errors.

vurusovs (approved):

PR #16652, opened 2023-03-30
[PyOV] Align and bump numpy, further tidy up requirements

p-wysocki:
### Details:
 - After ONNX 1.13.1 has been merged we can allow for `numpy<1.25` (`1.24.2` is the latest release)
 - Attempt to reduce the amount or requirement markers such as `scipy>=1.8; python_version >= '3.8'`
 - Move as many requirements into `constraints.txt` files as possible

### Tickets:
 - 98878


rkazants:
> `numpy` version in `openvino/tools/mo/requirements.txt` differs from the other MO requirement files. It cannot be changed as the base install is not intended to change users environment. Because of that its version cannot be contained in the common `constraints.txt` file.

I meant files like `openvino/tools/mo/requirements_tf.txt`, for TF2. MxNET, etc.

p-wysocki:
> I meant files like openvino/tools/mo/requirements_tf.txt, for TF2, MxNET, etc.

So the `numpy` requirements in `openvino/tools/mo` look like this:

- `openvino/tools/mo/requirements.txt` has `numpy>=1.16.6`
- all files `openvino/tools/mo/requirements_{extras}.txt` have `numpy>=1.16.6,<1.25.0`

The reason is, as you mentioned some time ago, that the default installation should not override users' environment. Since `constraints.txt` is a simple feature with not many options, it's impossible to override its values. **Any `requirements.txt` file which is linked to a given `constraints.txt` has to have the same requirement for that package.**

This means that if we were to simplify `numpy` in `openvino/tools`, we would add `numpy>=1.16.6,<1.25.0` to `openvino/tools/constraints.txt`, which I understand is what you desire. Since the aforementioned `numpy` versions differ between `requirements.txt` files in MO directory, it can have only three possible options:

1. Do not link default installation `openvino/tools/mo/requirements.txt` to `constraints.txt` at all and have it as a completely separate file - as a result we lose the benefit of having versions in a single place.
2. We could add `numpy>=1.16.6` to `constraints.txt` (**lower bound only**) and for every `openvino/tools/mo/requirements_{extras}.txt` add `numpy<1.25.0` (**upper bound only**) - this **should** work (I haven't tested it), but - we're still losing the benefit of having all versions in `constraints.txt`.
3. Keep `numpy` requirements separate for each `openvino/tools/mo/requirements_{extras}.txt`. This is still far from perfect since we have their versions in different files, but in my opinion it's the best option.

I also tested doing this:
image
which in my opinion should only be applying `-c` flag **after** installing `numpy` with relaxed bounds, but it still throws a dependency conflict: `ERROR: Double requirement given: numpy<1.25.0,>=1.16.6 (from -c ../constraints.txt (line 24)) (already in numpy>=1.16.6 (from -r requirements.txt (line 1)), name='numpy')`


@jiwaszki @rkazants @akuporos what do you think?

rkazants:
> > I meant files like openvino/tools/mo/requirements_tf.txt, for TF2, MxNET, etc.
> 
> So the `numpy` requirements in `openvino/tools/mo` look like this:
> 
> * `openvino/tools/mo/requirements.txt` has `numpy>=1.16.6`
> * all files `openvino/tools/mo/requirements_{extras}.txt` have `numpy>=1.16.6,<1.25.0`
> 
> The reason is, as you mentioned some time ago, that the default installation should not override users' environment. Since `constraints.txt` is a simple feature with not many options, it's impossible to override its values. **Any `requirements.txt` file which is linked to a given `constraints.txt` has to have the same requirement for that package.**
> 
> This means that if we were to simplify `numpy` in `openvino/tools`, we would add `numpy>=1.16.6,<1.25.0` to `openvino/tools/constraints.txt`, which I understand is what you desire. Since the aforementioned `numpy` versions differ between `requirements.txt` files in MO directory, it can have only three possible options:
> 
> 1. Do not link default installation `openvino/tools/mo/requirements.txt` to `constraints.txt` at all and have it as a completely separate file - as a result we lose the benefit of having versions in a single place.
> 2. We could add `numpy>=1.16.6` to `constraints.txt` (**lower bound only**) and for every `openvino/tools/mo/requirements_{extras}.txt` add `numpy<1.25.0` (**upper bound only**) - this **should** work (I haven't tested it), but - we're still losing the benefit of having all versions in `constraints.txt`.
> 3. Keep `numpy` requirements separate for each `openvino/tools/mo/requirements_{extras}.txt`. This is still far from perfect since we have their versions in different files, but in my opinion it's the best option.
> 
> I also tested doing this: image which in my opinion should only be applying `-c` flag **after** installing `numpy` with relaxed bounds, but it still throws a dependency conflict: `ERROR: Double requirement given: numpy<1.25.0,>=1.16.6 (from -c ../constraints.txt (line 24)) (already in numpy>=1.16.6 (from -r requirements.txt (line 1)), name='numpy')`
> 
> @jiwaszki @rkazants @akuporos what do you think?

Sorry, I didn't quite catch your answer. Can we put `-c constraints` into `requirements_{extra}.txt` file to get lower and upper bounds versions for other dependencies but `numpy`, for example? Or are you going to fix it in the separate PR and in the scope of other ticket? That is because after this PR we will still have a lot versions there spread among few files.

jiwaszki (approved):

p-wysocki:
@rkazants 
> Can we put -c constraints into requirements_{extra}.txt file to get lower and upper bounds versions for other dependencies but numpy, for example?

That's how it's currently done, assuming I understood you correctly. If you take a look at https://github.com/openvinotoolkit/openvino/blob/master/tools/mo/requirements_kaldi.txt you can see that all the packages are linked to `constraints.txt`, all but `numpy`, which has to be specified individually. In MO, the only version which is spread among files is `tensorflow` (because it differs between `requirements_tf.txt` and `requirements_tf2.txt`) and `numpy` (because it differs between `requirements.txt` and `requirements_{extra}.txt`). Please let me know if this answers your question.

PR #16654, opened 2023-03-30
[GPU] Fix OneDNN primitive attr serialization logic

e-ddykim:
### Details:
 - This PR fixes a bug in OneDNN primitive attr serialization logic.
 - In some cases, memory format tags are missed to be stored.

### Tickets:
 - 107099


e-ddykim:
> Could you add a unittest?

I added unit tests for fc and gemm fusing cases. Thank you.

yeonbok (approved):

PR #16671, opened 2023-03-31
[GPU] Fix synchronization issue from wrong stream in multi-stream mode on dGPU

andrew-k-park:
### Details:
 - Update to store network pointer in `kernel_impl_params` and use mem_lock through stream from network at runtime for `calc_output_layouts`

### Tickets:
 - 105711

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
@@ -7,6 +7,7 @@
 #include "intel_gpu/graph/serialization/binary_buffer.hpp"
 #include "intel_gpu/runtime/engine.hpp"
 #include "intel_gpu/runtime/memory.hpp"
+#include "intel_gpu/runtime/stream.hpp"
 #include "intel_gpu/runtime/utils.hpp"
 #include "intel_gpu/runtime/tensor.hpp"
 #include "intel_gpu/primitives/primitive.hpp"
@@ -21,7 +22,6 @@
 namespace cldnn {
 
 struct program;
-struct network;
 
 
 struct kernel_impl_params {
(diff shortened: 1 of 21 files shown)


src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
```diff
@@ -53,6 +54,7 @@ struct kernel_impl_params {
     size_t primary_input_idx = 0;
 
     memory::ptr reordered_weights = nullptr;
+    network *net = nullptr;
```
yeonbok, line 57: network::ptr can be used.

yeonbok: Or isn't it enough to have stream::ptr?

andrew-k-park: Yes, like 2nd option, `stream::ptr` is just enough. updated as your suggestion.

src/plugins/intel_gpu/src/graph/arg_max_min.cpp
```diff
@@ -96,7 +96,7 @@ std::vector<layout> arg_max_min_inst::calc_output_layouts(arg_max_min_node const
     } else if (constant_mem.count(1)) {
         std::map<size_t, ngraph::HostTensorPtr> const_data;
         auto target_shape_mem = constant_mem.at(1);
-        cldnn::mem_lock<uint8_t, mem_lock_type::read> target_shape_lock(target_shape_mem, impl_param.prog->get_stream());
+        cldnn::mem_lock<uint8_t, mem_lock_type::read> target_shape_lock(target_shape_mem, impl_param.get_stream());
```
sshlyapn, line 99: It looks like the same changes should be applied to `src/plugins/intel_gpu/src/graph/non_max_suppression.cpp` and
`src/plugins/intel_gpu/src/graph/impls/ocl/resample.cpp`

andrew-k-park: Applied to them. thanks!

yeonbok (approved):

e-ddykim (approved):
Looks good to me.

PR #16688, opened 2023-03-31
Add ConvolutionToGroupConvolutionFusion

mateusztabaka:
Fuses Split->series of Conv->Concat to GroupConvolution op.

Ticket: 105170

### DIFF
--- a/src/common/transformations/src/transformations/common_optimizations/convolution_to_group_convolution.cpp
+++ b/src/common/transformations/src/transformations/common_optimizations/convolution_to_group_convolution.cpp
@@ -1,0 +1,164 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <openvino/core/rt_info.hpp>
+#include <openvino/opsets/opset10.hpp>
+#include <openvino/pass/pattern/op/wrap_type.hpp>
+
+#include "itt.hpp"
+#include "transformations/common_optimizations/convolution_to_group_convolution_fusion.hpp"
+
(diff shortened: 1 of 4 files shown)


src/common/transformations/src/transformations/common_optimizations/convolution_to_group_convolution.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include "itt.hpp"
+#include "transformations/common_optimizations/convolution_to_group_convolution_fusion.hpp"
+
+static bool compare_convolutions(const ov::opset10::Convolution* conv1, ov::Node* node) {
+    const auto conv2 = ov::as_type<ov::opset10::Convolution>(node);
+    if (!conv2)
+        return false;
+    return conv1->get_strides() == conv2->get_strides() && conv1->get_pads_begin() == conv2->get_pads_begin() &&
+           conv1->get_pads_end() == conv2->get_pads_end() && conv1->get_dilations() == conv2->get_dilations() &&
+           conv1->get_auto_pad() == conv2->get_auto_pad();
```
itikhono, line 18: should we also compare pads_begin/pads_end attributes?

itikhono (approved):

t-jankowski (approved):

rkazants:
@mateusztabaka, can you please add TF FE unit-test to cover this case? Just create simple sub-graph in pbtxt (imitating the original model), pass it to the whole pipeline and check after compilation that the corresponsing node is created. Otherwise, it is risky to have such transformation because it can be broken by new transformation or some refactoring.

PR #16689, opened 2023-03-31
Bugfix in reorder_bfyx_to_blocked_format kernel

byungilm:
### Details:
 - Fix accuracy issue from check_image_accuracy.py
 - originated from PR#16410

### Tickets:
 - 67491


src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
```diff
@@ -1433,7 +1430,7 @@ impl_types layout_optimizer::get_preferred_impl_type(program_node& node, format
             preferred_impl = impl_types::ocl;
         }
 
-        if (!are_data_types_suitable_for_onednn(node)) {
+        if (input_dt == data_types::i64 || output_dt == data_types::i64) {
```
isanghao, line 1433: could you explain why this caused accuracy issue? Did it choose ocl for target reorder? I think it should not cause any accuracy issue whether it is ocl or onednn..

byungilm: Yes, you are right. I checked that there was an issue in bfyx_to_blocked_format kernel with bs_fs_yx_bsv16_fsvxx format.
I tried to merge this PR first to remove failure of check_image_accuracy. We definitely need to fix the root cause. Thanks.

byungilm: If it is fine to take a little more time that check_image_accuracy is failed. I will update this PR to fix root cause

isanghao (approved):

PR #16690, opened 2023-03-31
[GPU] Added GatherND dynamic support and changed logic for empty tensor support

Lyamin-Roman:
### Details:
 - *Added GatherND calc_output_layouts(..)*
 - *Kernel skip condition changed. Before: gws != 0, now: сustomizable function based on input/output shapes*
 - *Minor fix for gather_nonzero*

### Tickets:
 - *106432*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/softmax.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/softmax.cpp
@@ -59,7 +59,6 @@ struct softmax_impl : typed_primitive_impl_ocl<softmax> {
     void update_dispatch_data(const kernel_impl_params& impl_param) override {
         auto kernel_params = get_kernel_params(impl_param, true);
         (_kernel_data.update_dispatch_data_func)(kernel_params.first, _kernel_data);
-        update_kernels_list_to_skip();
     }
 };
 
--- a/src/plugins/intel_gpu/src/kernel_selector/kernel_selector_common.h
+++ b/src/plugins/intel_gpu/src/kernel_selector/kernel_selector_common.h
@@ -114,6 +114,24 @@ struct KernelData {
 
     bool can_reuse_memory = true;
 
+    static bool SkipKernelExecution(const base_params& params, size_t kernel_id = 0) {
+        for (const auto& input : params.inputs) {
+            if (input.LogicalSize() == 0) {
+                return true;
+            }
+        }
+        for (const auto& output : params.outputs) {
+            if (output.LogicalSize() == 0) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    static bool SkipKernelExecution(const Params& params, size_t kernel_id = 0) {
+        return false;
+    }
+
     template <typename T>
     inline static KernelData Default(const Params& _params, size_t kernel_nums = 1) {
         KernelData kd;
@@ -124,6 +142,10 @@ struct KernelData {
         kd.reorderInput = false;  // for KW
         kd.autoTuneIndex = -1;
         kd.can_reuse_memory = true;
+
+        for (auto& kernel : kd.kernels) {
+            kernel.skip_execution = SkipKernelExecution(orgParams);
+        }
         return kd;
     }
 };
--- a/src/plugins/intel_gpu/src/kernel_selector/kernels/concatenation/concatenation_kernel_base.cpp
+++ b/src/plugins/intel_gpu/src/kernel_selector/kernels/concatenation/concatenation_kernel_base.cpp
@@ -109,6 +109,10 @@ ConcatenationKernelBase::DispatchData ConcatenationKernelBase::SetDefault(const 
     return dispatchData;
 }
 
+bool ConcatenationKernelBase::SkipKernelExecution(const concatenation_params& params, size_t kernel_id) const {
+    return params.inputs[kernel_id].LogicalSize() == 0;
+}
+
 KernelsData ConcatenationKernelBase::GetCommonKernelsData(const Params& params, const optional_params& options) const {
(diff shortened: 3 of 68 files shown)


src/plugins/intel_gpu/src/kernel_selector/kernel_selector_common.h
```diff
@@ -114,6 +114,16 @@ struct KernelData {
 
     bool can_reuse_memory = true;
 
+    template <typename op_params>
+    inline static bool SkipKernelExecution(const op_params& params, size_t kernel_id = 0) {
+        for (const auto& input : params.inputs) {
+            if (input.LogicalSize() == 0) {
```
yeonbok, line 120: why not seeing the output? 
Even though there is inputs, we can skip the kernel if there is no output to be generated by the kernel.

src/plugins/intel_gpu/src/kernel_selector/kernel_selector_common.h
```diff
@@ -124,6 +134,10 @@ struct KernelData {
         kd.reorderInput = false;  // for KW
         kd.autoTuneIndex = -1;
         kd.can_reuse_memory = true;
+
+        for (auto& kernel : kd.kernels) {
+            kernel.skip_execution = SkipKernelExecution<T>(orgParams);
```
yeonbok, line 139: Is this to be set by all kernel generation? What about for shape agnostic kernel?

src/plugins/intel_gpu/src/kernel_selector/kernels/concatenation/concatenation_kernel_base.cpp
```diff
@@ -136,6 +141,7 @@ KernelsData ConcatenationKernelBase::GetCommonKernelsData(const Params& params,
             DispatchData dispatchData = SetDefault(newParams);
             kernel.params.workGroups.global = dispatchData.gws;
             kernel.params.workGroups.local = dispatchData.lws;
+            kernel.skip_execution = KernelData::SkipKernelExecution(prim_params, i);
```
yeonbok, line 144: Setting here is not enough for shape agnostic kernel. This should be updated at runtime for the shape agnostic kernel.

src/plugins/intel_gpu/src/kernel_selector/kernels/concatenation/concatenation_kernel_base.cpp
```diff
@@ -109,6 +109,11 @@ ConcatenationKernelBase::DispatchData ConcatenationKernelBase::SetDefault(const
     return dispatchData;
 }
 
+template<>
+bool KernelData::SkipKernelExecution<concatenation_params>(const concatenation_params& params, size_t kernel_id) {
+    return params.inputs[kernel_id].LogicalSize() == 0;
```
yeonbok, line 114: Why concat need its own function?

src/plugins/intel_gpu/src/graph/impls/ocl/softmax.cpp
```diff
@@ -59,7 +59,6 @@ struct softmax_impl : typed_primitive_impl_ocl<softmax> {
     void update_dispatch_data(const kernel_impl_params& impl_param) override {
         auto kernel_params = get_kernel_params(impl_param, true);
         (_kernel_data.update_dispatch_data_func)(kernel_params.first, _kernel_data);
-        update_kernels_list_to_skip();
```
yeonbok, line 62: (Random spot) Why these are removed? It should be updated at runtime whether to skip or not for shape agnostic kernel.

Lyamin-Roman: Updating skip_execution in update_dispatch_data_func is not equivalent to updating after it?

src/plugins/intel_gpu/src/kernel_selector/kernels/concatenation/concatenation_kernel_base.cpp
Lyamin-Roman: Because in concatenation each kernel has its own input and we must consider them separately

src/plugins/intel_gpu/src/kernel_selector/kernel_selector_common.h
Lyamin-Roman: For agnostic kernels, this value will be overwritten later

src/plugins/intel_gpu/src/kernel_selector/kernel_selector_common.h
Lyamin-Roman: Added check for outputs

src/plugins/intel_gpu/src/kernel_selector/kernel_selector_common.h
```diff
@@ -114,6 +114,21 @@ struct KernelData {
 
     bool can_reuse_memory = true;
 
+    template <typename op_params>
+    static bool SkipKernelExecution(const op_params& params, size_t kernel_id = 0) {
```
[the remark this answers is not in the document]
Lyamin-Roman: It is possible, but then instead of `KernelData::SkipKernelExecution(prim_params);` it will be `KernelData::SkipKernelExecution(dynamic_cast<const base_params&>(prim_params));`
Just shortened it a bit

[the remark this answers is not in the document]
vladimir-paramuzov: dynamic cast is not needed as it's just upcasting to base class ref.
Extensive templates usage impacts lib binary size, so I think in that case single (even non-virtual) function is better

vladimir-paramuzov (approved):

PR #16706, opened 2023-04-03
[TF FE] Add diagnostics capabilities via Framework nodes

rkazants:
**Details:** Now we continue translation despite failures in the existing translators. We just convert such nodes to Framework nodes. For debug capabilities, we need to save conversion failure reason into the FW node and summarize it in the end of conversion pipeline. The summary should contain all such failure messages and a list of unsupported operations, i.e. operations for which translators do not exist.

**Ticket:** TBD


src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -32,24 +32,32 @@ using namespace ov;
...
             auto op_type = fw_node->get_decoder()->get_op_type();
-            unconverted_ops_types.push_back(op_type);
+            auto fw_node_attrs = fw_node->get_attrs();
+            if (fw_node_attrs.find(FrameworkNode::failed_conversion_key) != fw_node_attrs.end()) {
+                // that means the translator is found but the conversion is failed
+                failures[op_type] = fw_node_attrs.at(FrameworkNode::failed_conversion_key);
+            } else if (std::find(unsupported_operations.begin(), unsupported_operations.end(), op_type) ==
+                       unsupported_operations.end()) {
+                // found new unsupported operation
+                unsupported_operations.push_back(op_type);
```
slyalin, line 51: Can we use a set for collecting unsupported operations instead of doing not efficient linear search in `std::vector`? After collecting them in the set, you will move them to `unsupported_operations` in the end. Or just consider replacing the type of the original parameter to set instead of vector.

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -32,24 +32,32 @@ using namespace ov;
 using namespace ov::frontend::tensorflow;
 
 namespace {
-std::vector<std::string> get_unconverted_types_from_model(const std::shared_ptr<Model>& model) {
-    std::vector<std::string> unconverted_ops_types;
+
+void get_unsupported_operations_and_failures(const std::shared_ptr<Model>& model,
+                                             std::vector<std::string>& unsupported_operations,
+                                             std::unordered_map<std::string, std::string>& failures) {
+    unsupported_operations.clear();
```
slyalin, line 39: When the function is called recursively (in case when `MultiSubGraphOp` is discovered below in the code) this will erase all accumulated content with unsupported operations. This makes output incomplete: not all the unsupported operations are reported.

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -195,15 +203,33 @@ ov::frontend::InputModel::Ptr FrontEnd::load_impl(const std::vector<ov::Any>& va
...
+        if (m_telemetry) {
+            // TODO: 105173 support anonymization of exception message in order to send to telemetry
+        }
+        exception_message += "[TensorFlow Frontend] Internal error: conversion is failed for " + failure.first +
+                             " operation with a message:\n" + failure.second + "\n";
+    }
+
+    // append unsupported operations
+    for (const auto& unsupported_operation : unsupported_operations) {
+        if (m_telemetry) {
```
slyalin, line 221: This condition is invariant, please move it out of the loop body.

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -195,15 +203,33 @@ ov::frontend::InputModel::Ptr FrontEnd::load_impl(const std::vector<ov::Any>& va
...
             m_telemetry->send_event("error_cause", "tf_" + unsupported_operation);
         }
+        // TODO 107500: report the full list of unsupported operations
```
slyalin, line 224: It looks to be already reported in the loop above. What do you mean by this TODO?

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -195,15 +203,33 @@ ov::frontend::InputModel::Ptr FrontEnd::load_impl(const std::vector<ov::Any>& va
...
     }
-    FRONT_END_OP_CONVERSION_CHECK(
-        unsupported_operations.size() == 0,
-        "[TensorFlow Frontend] Internal error: No translator found for " + unsupported_operations[0] + " node.");
+    exception_message = (unsupported_operations.size() > 0
+                             ? exception_message + "[TensorFlow Frontend] Internal error: No translator found for " +
+                                   unsupported_operations[0] + " node."
```
slyalin, line 228: Even if there are no messages collected, there can be multiple operations unsupported, but this code reports about the first unsupported operation only.

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -242,12 +243,14 @@ void TranslateSession::translate_graph(const ov::frontend::InputModel::Ptr& inpu
...
+                // remember the root-cause about why the translation was failed
+                ov::op::util::FrameworkNodeAttrs attrs;
+                attrs[FrameworkNode::failed_conversion_key] = ex.what();
                 auto fw_node = std::make_shared<FrameworkNode>(operation_decoder,
                                                                ov_inputs,
                                                                operation_place->get_output_ports().size());
+                fw_node->set_attrs(attrs);
                 set_node_name(operation_name, fw_node);
                 ov_outputs = fw_node->outputs();
             }
```
slyalin, line 256: That means we are collection status of conversion for `std::exception` only missing other exception origins. Unknown exceptions should be gathered as well and kind of "Unknown exception type" is recorded in FW node. It is especially important for user-defined translators that are our of our control.

slyalin (changes requested):

src/frontends/tensorflow/src/frontend.cpp
rkazants: Now some tests rely on the order of unsupported operations. I would rather to fix in the scope of 107500 ticket.

src/frontends/tensorflow/src/frontend.cpp
rkazants: Fixed. Great catch!

src/frontends/tensorflow/src/frontend.cpp
rkazants: Fixed.

src/frontends/tensorflow/src/frontend.cpp
rkazants: Explained in the commend and discussed offline.

src/frontends/tensorflow/src/frontend.cpp
rkazants: Will be addressed in the ticket 107500

src/frontends/tensorflow/src/translate_session.cpp
rkazants: Fixed.

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -195,15 +203,33 @@ ov::frontend::InputModel::Ptr FrontEnd::load_impl(const std::vector<ov::Any>& va
...
+    std::vector<std::string> unsupported_operations;
+    get_unsupported_operations_and_failures(f, unsupported_operations, failures);
+
+    std::string exception_message;
+    for (const auto& failure : failures) {
+        if (m_telemetry) {
+            // TODO: 105173 support anonymization of exception message in order to send to telemetry
+        }
+        exception_message += "[TensorFlow Frontend] Internal error: conversion is failed for " + failure.first +
+                             " operation with a message:\n" + failure.second + "\n";
```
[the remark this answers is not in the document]
rkazants: Let us have the current version just to avoid complex logic that can be not reasonable for such feature. Just reporting the first failure will be likely sufficient to root-cause the problem.

src/frontends/tensorflow/tests/convert_unsupported.cpp
```diff
@@ -75,6 +78,15 @@ shared_ptr<Model> convert_model_partially(const string& model_path) {
 
     return model;
 }
+
+ov::OutputVector incorrect_less_translator(const ov::frontend::NodeContext& node) {
+    // NOTE: pay attention that this is a fake translator for Less operation
+    // only serves for testing purposes
+    // FRONT_END_GENERAL_CHECK(false, "Less expects at least ten inputs.");
```
rkazants, line 85:
```suggestion
```

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -32,24 +32,32 @@ using namespace ov;
...
         if (const auto& fw_node = ov::as_type_ptr<FrameworkNode>(node)) {
             auto op_type = fw_node->get_decoder()->get_op_type();
-            unconverted_ops_types.push_back(op_type);
+            auto fw_node_attrs = fw_node->get_attrs();
+            if (fw_node_attrs.find(FrameworkNode::failed_conversion_key) != fw_node_attrs.end() &&
+                failures.count(op_type) == 0) {
+                // save only the first encountered failure that is more improtant for developer
+                // that means the translator is found but the conversion is failed
+                failures[op_type] = fw_node_attrs.at(FrameworkNode::failed_conversion_key);
+            } else if (std::find(unsupported_operations.begin(), unsupported_operations.end(), op_type) ==
```
itikhono, line 48: could you use `unordered_set`? O(n) is not so good

rkazants: Now some tests rely on the order of unsupported operations. If I start use `set`, the order is lost for them. Proper structuring will be fixed 107500 ticket where I am going to store it in the `unordered_set` and save it in the exception for further communication with MO.

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -32,24 +32,32 @@ using namespace ov;
...
+                       unsupported_operations.end()) {
+                // found new unsupported operation
+                unsupported_operations.push_back(op_type);
```
itikhono, line 51: should we count the number of unsupported_op of each type?

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -195,15 +203,34 @@ ov::frontend::InputModel::Ptr FrontEnd::load_impl(const std::vector<ov::Any>& va
...
+    // TODO 107500: report the full list of unsupported operations
+    // also, communicate with MO for the fallback to the legacy FE
+    // via OpConversionFailure exception that will store all failures and unsupported_operations
+    exception_message = (unsupported_operations.size() > 0
+                             ? exception_message + "[TensorFlow Frontend] Internal error: No translator found for " +
+                                   unsupported_operations[0] + " node."
+                             : exception_message);
+
+    bool is_conversion_successful = ((unsupported_operations.size() == 0) && (failures.size() == 0));
+    FRONT_END_OP_CONVERSION_CHECK(is_conversion_successful, exception_message);
```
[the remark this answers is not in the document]
rkazants: `is_conversion_successful` is really more logical

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -32,24 +32,32 @@ using namespace ov;
...
     for (const auto& node : model->get_ordered_ops()) {
         if (const auto& fw_node = ov::as_type_ptr<FrameworkNode>(node)) {
             auto op_type = fw_node->get_decoder()->get_op_type();
-            unconverted_ops_types.push_back(op_type);
+            auto fw_node_attrs = fw_node->get_attrs();
+            if (fw_node_attrs.find(FrameworkNode::failed_conversion_key) != fw_node_attrs.end() &&
+                failures.count(op_type) == 0) {
+                // save only the first encountered failure that is more improtant for developer
+                // that means the translator is found but the conversion is failed
+                failures[op_type] = fw_node_attrs.at(FrameworkNode::failed_conversion_key);
```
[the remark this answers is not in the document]
rkazants: It can be a long and repeating quite similar message for the same operation. Just do not overload `cout`, propose to print only topologically first operation for which the conversion is failed.

src/frontends/tensorflow/src/frontend.cpp
rkazants: Not useful info for the user. The most important thing is to know that some translators are missed for conversion.

src/frontends/tensorflow_common/src/tf_framework_node.cpp
```diff
@@ -0,0 +1,7 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "tf_framework_node.hpp"
+
+const std::string ov::frontend::tensorflow::FrameworkNode::failed_conversion_key = "failed_conversion_message";
```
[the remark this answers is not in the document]
rkazants: Not sure what advantages will it bring if this key only used to sort out FW nodes which have translators. And there will be no conflicts with other FEs.

slyalin (changes requested):

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -195,15 +203,34 @@ ov::frontend::InputModel::Ptr FrontEnd::load_impl(const std::vector<ov::Any>& va
...
             m_telemetry->send_event("error_cause", "tf_" + unsupported_operation);
         }
     }
-    FRONT_END_OP_CONVERSION_CHECK(
-        unsupported_operations.size() == 0,
-        "[TensorFlow Frontend] Internal error: No translator found for " + unsupported_operations[0] + " node.");
+    // TODO 107500: report the full list of unsupported operations
+    // also, communicate with MO for the fallback to the legacy FE
+    // via OpConversionFailure exception that will store all failures and unsupported_operations
+    exception_message = (unsupported_operations.size() > 0
```
itikhono, line 227: unnesessary coping, please use
if (unsupported_operations.size() > 0)
    exception_message += "[TensorFlow Frontend] Internal error: No translator found for " +
                                   unsupported_operations[0] + " node.";

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -195,15 +203,34 @@ ov::frontend::InputModel::Ptr FrontEnd::load_impl(const std::vector<ov::Any>& va
 std::shared_ptr<ov::Model> FrontEnd::convert(const ov::frontend::InputModel::Ptr& model) const {
     auto f = convert_partially(model);
 
-    auto unsupported_operations = get_unconverted_types_from_model(f);
+    std::unordered_map<std::string, std::string> failures;
+    std::vector<std::string> unsupported_operations;
+    get_unsupported_operations_and_failures(f, unsupported_operations, failures);
+
+    std::string exception_message;
```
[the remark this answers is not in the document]
itikhono: string += will reallocate buffer (like vector) 
stringstring doesn't reallocate because contains linked buffers or something like that.
the string concatenation will be faster with stringstream

src/frontends/tensorflow_common/src/tf_framework_node.cpp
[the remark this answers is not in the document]
itikhono: agree, but you can use `static constexpr const char* failed_conversion_key`
I believe that map[failed_conversion_key] will work without any changes after that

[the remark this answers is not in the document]
itikhono: In any case, this key looks more strict

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -242,12 +243,24 @@ void TranslateSession::translate_graph(const ov::frontend::InputModel::Ptr& inpu
                 auto translator = m_translator_map->at(operation_decoder->get_op_type());
                 NodeContext node_context(operation_decoder, ov_inputs, this);
                 ov_outputs = translator(node_context);
-            } catch (const std::exception&) {
-                // continue translation by replacing with FrameworkNode
-                // in case of any failures in translators due to their limitation
+            } catch (const std::exception& ex) {
+                // remember the root-cause about why the translation was failed
```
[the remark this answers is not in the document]
rkazants: Fixed.

src/frontends/tensorflow/src/frontend.cpp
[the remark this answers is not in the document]
rkazants: Fixed.

src/frontends/tensorflow/src/frontend.cpp
rkazants: Fixed.

itikhono (approved):

src/frontends/tensorflow_common/src/tf_framework_node.cpp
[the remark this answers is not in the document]
rkazants: Fixed.

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -242,12 +243,24 @@ void TranslateSession::translate_graph(const ov::frontend::InputModel::Ptr& inpu
...
+                ov_outputs = fw_node->outputs();
+            } catch (...) {
+                // also, saves unknown exception type
+                ov::op::util::FrameworkNodeAttrs attrs;
+                attrs[FrameworkNode::failed_conversion_key] = "Unknown exception type";
                 auto fw_node = std::make_shared<FrameworkNode>(operation_decoder,
                                                                ov_inputs,
                                                                operation_place->get_output_ports().size());
+                fw_node->set_attrs(attrs);
                 set_node_name(operation_name, fw_node);
```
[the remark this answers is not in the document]
rkazants: Done.

src/frontends/tensorflow_common/include/tf_framework_node.hpp
```diff
@@ -15,6 +15,7 @@ namespace tensorflow {
 
 class FrameworkNode : public ov::op::util::FrameworkNode {
 public:
+    static constexpr const char* failed_conversion_key = "tensorflow::FrameworkNode::failed_conversion_key";
```
slyalin, line 18: Not a blocker for the PR but just for the records, this `failed_conversion_key` looks bulky in comparison to a dedicate field in the class.

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -242,14 +257,20 @@ void TranslateSession::translate_graph(const ov::frontend::InputModel::Ptr& inpu
...
+                                                           operation_place->get_output_ports().size(),
+                                                           operation_name,
+                                                           ex.what());
+            } catch (...) {
+                // save unknown exception type
+                ov_outputs = create_fw_node_with_exception(operation_decoder,
+                                                           ov_inputs,
+                                                           operation_place->get_output_ports().size(),
+                                                           operation_name,
+                                                           "Unknown exception type");
```
slyalin, line 273: There is no test that covers this part.

src/frontends/tensorflow/tests/convert_unsupported.cpp
```diff
@@ -75,6 +78,14 @@ shared_ptr<Model> convert_model_partially(const string& model_path) {
 
     return model;
 }
+
+ov::OutputVector incorrect_less_translator(const ov::frontend::NodeContext& node) {
+    // NOTE: pay attention that this is a fake translator for Less operation
+    // only serves for testing purposes
+    TENSORFLOW_OP_VALIDATION(node, false, "Less expects ten inputs.");
+    return {};
+}
```
slyalin, line 87: To test both known and unknown exception handlers specialize this function by the exception statement. I would recommend to implement translator itself as a lambda function in the generation function that accepts another function as a parameter. That function will be used to generate the exception. Doing so you can build "incorrect" translator with configured exception type.

src/frontends/tensorflow_common/include/tf_framework_node.hpp
rkazants: Agreed to proceed with the current version because this is not clear how to implement this better.

src/frontends/tensorflow/src/translate_session.cpp
rkazants: Added. Thanks!

src/frontends/tensorflow/tests/convert_unsupported.cpp
rkazants: Discussed, agreed that not needed.

PR #16714, opened 2023-04-03
[GPU] Don't reorder weights when can reinterpret

vladimir-paramuzov:
### Details:
 - Avoid redundant weights reordering for formats that can be reinterpreted to each other
 - Optimize out dynamic -> static reshape to avoid recompilation
 - Disable internal einsum decomposition for dynamic shapes and rely on common transform

### DIFF
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -395,11 +395,13 @@ bool primitive_inst::update_impl() {
                     impl->set_kernels(kernels);
                     cache.add(updated_params, impl->clone());
                 });
-                _impl = _dynamic_impl->clone();
-                auto new_impl_params = _impl->canonicalize_shapes(*_impl_params);
-                _impl->update_dispatch_data(new_impl_params);
+                if (!can_be_optimized())  {
+                    _impl = _dynamic_impl->clone();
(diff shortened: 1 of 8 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -715,90 +717,97 @@ event::ptr primitive_inst::update_weights() {
...
+        // Set original patrial shape, because it may be lost during kernel_selector::weights_tensor -> layout conversion
+        expected_layout.set_partial_shape(original_weights_memory->get_layout().get_partial_shape());
+        _impl_params->weights_layout = optional_layout(expected_layout);
+
+        if (_reordered_weights_cache.has(expected_layout)) {
             GPU_DEBUG_PROFILED_STAGE_CACHE_HIT(true);
-            kernel = cache.get(kernel_key);
+            GPU_DEBUG_TRACE_DETAIL << id() << ": reuse weights for " << expected_layout.to_short_string() << std::endl;
+            return nullptr;
+        } else if (original_layout.compatible(expected_layout)) {
```
sshlyapn, line 741: Maybe it would be better to check for compatibility not only original layout, but all other cached layout as well

vladimir-paramuzov: Yeah, that can be a further improvement

sshlyapn (approved):

PR #16718, opened 2023-04-03
[GNA] Fix compilation

dorloff:
### Details:
 - Fix compilation error for GNA plugin

### Tickets:
 - 104500

### DIFF
--- a/src/plugins/intel_gna/src/frontend/model_quantizer.hpp
+++ b/src/plugins/intel_gna/src/frontend/model_quantizer.hpp
@@ -148,7 +148,7 @@ private:
                 inf_loop_count++;
             } else {
                 if (inf_loop_count > 0 &&
-                    (inf_loop_pattern.size() > 0 && inf_loop_history.size() % inf_loop_pattern.size() == 0 ||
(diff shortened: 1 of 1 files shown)


evkotov (approved):

src/plugins/intel_gna/src/frontend/model_quantizer.hpp
```diff
@@ -148,7 +148,7 @@ class ModelQuantizer {
                 inf_loop_count++;
             } else {
                 if (inf_loop_count > 0 &&
```
andrei-kochin, line 150:
```suggestion
                if ((inf_loop_count > 0) &&
```

dorloff:
#16719 is created instead to revert the problematic PR (#16003)

src/plugins/intel_gna/src/frontend/model_quantizer.hpp
dorloff: Thanks @andrei-kochin . We decided that we need to check this more carefully and revert first.

rkazants:
@dorloff. still not

PR #16724, opened 2023-04-04
[GPU] add WA to avoid hang issue.

sungeunk:
Force exit without releasing any resources if CL_OUT_OF_RESOURCE is returned from call OpenCL API.

### Tickets:
 - 106623

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
+++ b/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
@@ -488,7 +488,18 @@ protected:
         }
 
         if (!instance.can_be_optimized()) {
-            _prim.execute(stream.get_onednn_stream(), _args[net_id]);
+            try {
+                _prim.execute(stream.get_onednn_stream(), _args[net_id]);
+            } catch (dnnl::error& err) {
+                /// WA: Force exit. Any opencl api call can be hang after CL_OUT_OF_RESOURCES.
+                if (err.status == dnnl_status_t::dnnl_out_of_memory) {
+                    std::cerr << "[GPU] WA: Force exit. CL_OUT_OF_RESOURCES is returned from call OpenCL API." << std::endl
+                              << "      It could occur hang issue." << std::endl
+                              << "      Please use smaller batches or streams." << std::endl;
+                    std::_Exit(-1);
+                }
+                throw;    // rethrowing dnnl::error if not out_of_memory
+            }
         }
 
         if (_enable_profiling) {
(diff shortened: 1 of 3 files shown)


src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
```diff
@@ -488,7 +488,18 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
...
 
         if (!instance.can_be_optimized()) {
-            _prim.execute(stream.get_onednn_stream(), _args[net_id]);
+            try {
+                _prim.execute(stream.get_onednn_stream(), _args[net_id]);
+            } catch (dnnl::error& err) {
+                /// WA: Force exit. Any opencl api call can be hang after CL_OUT_OF_RESOURCES.
+                if (err.status == dnnl_status_t::dnnl_out_of_memory) {
+                    std::cerr << "[GPU] WA: Force exit. CL_OUT_OF_RESOURCES is returned from call OpenCL API." << std::endl
+                              << "      It could occur hang issue." << std::endl
```
vladimir-paramuzov, line 497: Could you please mention here that the hang is the driver issue and recommend to try with the newer one? Like
`Due to the driver bug any subsequent OpenCL API call will cause application hang, so GPU plugin can't finish correctly. Please try to update the driver or reduce memory consumption (use smaller batch size, less streams, lower precision, etc) to avoid CL_OUT_OF_RESOURCES exception`

src/plugins/intel_gpu/src/runtime/ocl/ocl_stream.cpp
```diff
@@ -304,6 +304,16 @@ event::ptr ocl_stream::enqueue_kernel(kernel& kernel,
     try {
         _command_queue.enqueueNDRangeKernel(kern, cl::NullRange, global, local, dep_events_ptr, set_output_event ? &ret_ev : nullptr);
     } catch (cl::Error const& err) {
+        /// WA: Force exit. Any opencl api call can be hang after CL_OUT_OF_RESOURCES.
+        if (err.err() == CL_OUT_OF_RESOURCES) {
+            std::cerr << "[GPU Plugin] force exit.\n"
```
vladimir-paramuzov, line 309: 1. `[GPU] ... `
2. Does it impact dGPU only? If yes, I think we need to check device type here and just rethrow exception on igpu.

sungeunk: It's dGPU only. In iGPU case, the app will be killed with 'Out of memory' by host kernel side before gpu plugin gets 'CL_OUT_OF_RESOURCES'.

vladimir-paramuzov (approved):

PR #16731, opened 2023-04-04
Remove legacy TransposeSinking transformation

itikhono:
### Details:
 - *Removing legacy TransposeSinking ovtf transformation. We replaced this with TransposeSinkingGeneral

### Tickets:
 - *107399*


src/frontends/tensorflow_common/tests/transpose_sinking.cpp
```diff
@@ -1,645 +0,0 @@
-// Copyright (C) 2018-2023 Intel Corporation
-// SPDX-License-Identifier: Apache-2.0
-//
```
rkazants, line 3: How do you think - is it possible to adopt these tests for the new Transpose Sinking?

rkazants (approved):

src/frontends/tensorflow_common/tests/transpose_sinking.cpp
itikhono: I think so

ilya-lavrenov (approved):

PR #16739, opened 2023-04-04
[CPU] Clean up primitive desc creation in prepareParams

EgorDuplensky:
Small refactoring

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/matmul.cpp
+++ b/src/plugins/intel_cpu/src/nodes/matmul.cpp
@@ -609,10 +609,10 @@ void MatMul::prepareParams() {
     auto engine = getEngine();
 
     auto builder = [&engine](const MatMulKey& key) -> executorPtr {
-        dnnl::matmul::primitive_desc matmul_desc;
+        dnnl::matmul::primitive_desc prim_desc;
 
         if (key.bias) {
-            matmul_desc = matmul::primitive_desc(
+            prim_desc = matmul::primitive_desc(
                 engine,
                 key.inp0->getDnnlDesc(),
                 key.inp1->getDnnlDesc(),
@@ -620,7 +620,7 @@ void MatMul::prepareParams() {
                 key.out->getDnnlDesc(),
                 key.attr);
         } else {
-            matmul_desc = matmul::primitive_desc(
+            prim_desc = matmul::primitive_desc(
                 engine,
                 key.inp0->getDnnlDesc(),
                 key.inp1->getDnnlDesc(),
@@ -628,27 +628,17 @@ void MatMul::prepareParams() {
                 key.attr);
         }
 
-        primitive_desc_iterator itpd = matmul_desc;
-        matmul::primitive_desc prim_desc;
+        auto first_desc = dnnl::matmul::primitive_desc(prim_desc.get());
+        const bool found = DnnlExtensionUtils::find_implementation(prim_desc, key.implType);
 
-        auto itpd_first = itpd;
-        while (static_cast<bool>(itpd))  {
-            impl_desc_type impl_type = parse_impl_name(itpd.impl_info_str());
+        if (found)
+            return std::make_shared<DnnlExecutor>(prim_desc);
 
-            if (impl_type == key.implType) {
-                prim_desc = itpd.get();
-                break;
-            }
-            if (!itpd.next_impl()) {
-                // In case of dynamic shapes an implementation type chosen as optimal for a primitive_desc with
-                // undefined input shapes, is not necessarily available for the primitive_desc with defined shape.
-                // Example: brgemm_avx512_amx (Intel Sapphire Rapids Platform) is available for a primitive with
-                // undefined input shapes but not available for primitive_desc with input batch 1.
-                prim_desc = itpd_first.get();
-                break;
-            }
-        }
-        return std::make_shared<DnnlExecutor>(prim_desc);
+        // In case of dynamic shapes an implementation type chosen as optimal for a primitive_desc with
+        // undefined input shapes, is not necessarily available for the primitive_desc with defined shape.
+        // Example: brgemm_avx512_amx (Intel Sapphire Rapids Platform) is available for a primitive with
+        // undefined input shapes but not available for primitive_desc with input batch 1.
+        return std::make_shared<DnnlExecutor>(first_desc);
     };
 
     auto cache = context->getParamsCache();
--- a/src/plugins/intel_cpu/src/nodes/pooling.cpp
+++ b/src/plugins/intel_cpu/src/nodes/pooling.cpp
@@ -434,30 +434,27 @@ void Pooling::prepareParams() {
                           selected_pd->getImplementationType()};
         auto engine = getEngine();
         auto builder = [&engine](const PoolingKey& key) -> executorPtr {
-            primitive_desc_iterator itpd = createDescriptorHelper(engine,
-                                                                  key.inp->getDnnlDesc(),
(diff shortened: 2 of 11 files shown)


EgorDuplensky:
@maxnick Could you please take a look

src/plugins/intel_cpu/src/nodes/softmax.cpp
```diff
@@ -182,7 +181,7 @@ void SoftMax::prepareParams() {
             key.attr,
             true);
 
-        primitive_desc_iterator itpd = *desc;
+        primitive_desc_iterator itpd = prim_desc;
 
         while (itpd) {
             impl_desc_type impl_type = parse_impl_name(itpd.impl_info_str());
```
EgorDuplensky, line 187: @zhangYiIntel May I ask you to take a look at this?
Will it be correct to re-design this logic the same way it is done for matmul, i.e. just pick first available implementation in case key.implType == ref_any?

src/plugins/intel_cpu/src/nodes/pooling.cpp
```diff
@@ -370,28 +370,23 @@ void Pooling::prepareParams() {
...
-        dnnl::pooling_forward::primitive_desc prim_desc = itpd.get();
-        while (static_cast<bool>(itpd)) {
-            impl_desc_type impl_type = parse_impl_name(itpd.impl_info_str());
-
-            if (impl_type == key.implType) {
-                prim_desc = itpd.get();
-                break;
-            }
-            if (!itpd.next_impl())
-                break;
```
EgorDuplensky, line 393: It seems pooling primitive has a different way of handling the case when a proper implementation is not found.
It just uses the last implementation (reference), when other primitive either return nullptr or pick the first available implementation.
@dmitry-gorokhov @maxnick Does anyone know the reason. Can we pick the first available instead?

EgorDuplensky: I have changed the behavior to use the first available implementation and at least the tests which cover the case with the fallback are passed.

maxnick: Actually no. On L384 the first available implementation is stored, then if nothing matched the impl_type, the stored descriptor (witch corresponds to the first available implementation) is used.

maxnick:
> @maxnick Could you please take a look

The intention of this PR is understandable. But, the thing is little bit more complicated. We use this while loop to match the implementation selected on the compilation stage. We do this intentionally to avoid possible extra reordering. On the compilation stage the variable data layout is selected and propagated through the whole graph in a way to avoid reordering as much as possible. The oneDNN implementations types are stored to ensure the same data layout at runtime. Now, if we always select the first available implementation, we cannot be sure that the layout is the same as stored in the node descriptor, which means that there is a probability of the layout mismatch and insertion of additional reorders, or even exception throwing, if the layout mismatch handling is not implemented for some node. Yes, it is reasonable to assume that in the most cases the layout will be the same, and if not, it is handled with the implicit reordering logic anyway, but it is not obvious what is better to do in case when the first implementation does not match the selected, but the selected still persists in the implementations list.
So the question is, are we sure that this refactoring is at least safe? I mean is it possible that the first implementation always matches the layout assigned to the edge tensor descriptor.

EgorDuplensky:
> > @maxnick Could you please take a look
> 
> The intention of this PR is understandable. But, the thing is little bit more complicated. We use this while loop to match the implementation selected on the compilation stage. We do this intentionally to avoid possible extra reordering. On the compilation stage the variable data layout is selected and propagated through the whole graph in a way to avoid reordering as much as possible. The oneDNN implementations types are stored to ensure the same data layout at runtime. Now, if we always select the first available implementation, we cannot be sure that the layout is the same as stored in the node descriptor, which means that there is a probability of the layout mismatch and insertion of additional reorders, or even exception throwing, if the layout mismatch handling is not implemented for some node. Yes, it is reasonable to assume that in the most cases the layout will be the same, and if not, it is handled with the implicit reordering logic anyway, but it is not obvious what is better to do in case when the first implementation does not match the selected, but the selected still persists in the implementations list. So the question is, are we sure that this refactoring is at least safe? I mean is it possible that the first implementation always matches the layout assigned to the edge tensor descriptor.

Actually there is no code logic changed, just cleaned up code duplications and more declarative style (I have reverted the deep copy change, which would introduce a perf degradation but also did not affect the overall code logic).
May I ask you point out the piece of code you have worries about, maybe I have missed something?

src/plugins/intel_cpu/src/nodes/softmax.cpp
EgorDuplensky: Yeah, that is why I have not touched it. Lets leave it for now.

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -594,46 +594,36 @@ void MatMul::prepareParams() {
...
                 engine,
                 key.inp0->getDnnlDesc(),
                 key.inp1->getDnnlDesc(),
                 key.out->getDnnlDesc(),
                 key.attr);
         }
 
-        primitive_desc_iterator itpd = matmul_desc;
-        matmul::primitive_desc prim_desc;
+        auto initial_desc = dnnl::matmul::primitive_desc(prim_desc.get());
```
maxnick, line 616:
```suggestion
        auto first_desc = dnnl::matmul::primitive_desc(prim_desc.get());
```

maxnick (approved):
In general LGTM. There are a couple of minor comments left.

maxnick:
> > > @maxnick Could you please take a look
> > 
> > 
> > The intention of this PR is understandable. But, the thing is little bit more complicated. We use this while loop to match the implementation selected on the compilation stage. We do this intentionally to avoid possible extra reordering. On the compilation stage the variable data layout is selected and propagated through the whole graph in a way to avoid reordering as much as possible. The oneDNN implementations types are stored to ensure the same data layout at runtime. Now, if we always select the first available implementation, we cannot be sure that the layout is the same as stored in the node descriptor, which means that there is a probability of the layout mismatch and insertion of additional reorders, or even exception throwing, if the layout mismatch handling is not implemented for some node. Yes, it is reasonable to assume that in the most cases the layout will be the same, and if not, it is handled with the implicit reordering logic anyway, but it is not obvious what is better to do in case when the first implementation does not match the selected, but the selected still persists in the implementations list. So the question is, are we sure that this refactoring is at least safe? I mean is it possible that the first implementation always matches the layout assigned to the edge tensor descriptor.
> 
> Actually there is no code logic changed, just cleaned up code duplications and more declarative style used (I have reverted the deep copy change, which would introduce a perf degradation but also did not affect the overall code logic). May I ask you to point out the piece of code you have worries about, maybe I have missed something?

Yep, got the idea. You simply extracted the while loops to a separate method.

EgorDuplensky:
@maxnick I guess we can merge it after a rebase 👀

src/plugins/intel_cpu/src/nodes/pooling.cpp
```diff
@@ -370,30 +370,29 @@ void Pooling::prepareParams() {
...
+            key.kernel,
+            key.effective_pad_begin,
+            key.effective_pad_end,
+            key.effective_dilation,
+            key.data_pad_end,
+            key.attr);
+
+        dnnl::pooling_forward::primitive_desc tmp = prim_desc.get();
+
+        auto initial_desc = dnnl::pooling_forward::primitive_desc(prim_desc.get());
```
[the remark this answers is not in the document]
EgorDuplensky: Done

PR #16741, opened 2023-04-04
[PT FE]:  fix aten::index inconsistent reshape

eaidova:
### Details:
 - *ifix bug in detectron reshape shapes are inconsistent in index op*

### Tickets:
 - *105142*

### DIFF
--- a/src/frontends/pytorch/src/transforms/aten_index_replacer.cpp
+++ b/src/frontends/pytorch/src/transforms/aten_index_replacer.cpp
@@ -47,12 +47,12 @@ std::shared_ptr<Node> flatten(const Output<Node>& value, size_t axis) {
     } else {
         const auto value_shape = std::make_shared<v3::ShapeOf>(value, element::i32);
         const auto value_rank = std::make_shared<v3::ShapeOf>(value_shape, element::i32);
-        const auto axis_node = v0::Constant::create(element::i32, Shape{}, {axis});
-        auto start = v0::Constant::create(element::i32, Shape{}, {0});
-        auto step = v0::Constant::create(element::i32, Shape{}, {1});
+        const auto axis_node = v0::Constant::create(element::i32, Shape{1}, {axis});
+        auto start = v0::Constant::create(element::i32, Shape{1}, {0});
+        auto step = v0::Constant::create(element::i32, Shape{1}, {1});
         const auto first_part_dims = std::make_shared<v8::Slice>(value_shape, start, axis_node, step);
         auto zero = v0::Constant::create(element::i32, {}, {0});
-        auto first_part_dims_length = std::make_shared<ov::op::v1::ReduceProd>(first_part_dims, zero, true);
+        auto first_part_dims_length = std::make_shared<v1::ReduceProd>(first_part_dims, zero, true);
 
         auto remaining_part_length = v0::Constant::create(element::i32, {1}, {-1});
 
@@ -70,7 +70,7 @@ AtenIndexToSelect::AtenIndexToSelect() {
         if (!index_op) {
             return false;
         }
-        auto input_node = index_op->input_value(0).get_node_shared_ptr();
+        auto input_node = index_op->input_value(0);
         auto indicies = index_op->input_value(1).get_node_shared_ptr();
         auto list_indicies = cast_fw_node(indicies, "prim::ListConstruct");
         if (list_indicies) {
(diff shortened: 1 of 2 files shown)


src/frontends/pytorch/src/transforms/aten_index_replacer.cpp
```diff
@@ -1,7 +1,3 @@
-// Copyright (C) 2018-2023 Intel Corporation
-// SPDX-License-Identifier: Apache-2.0
-//
-
```
mvafin, line 4: Please return

src/frontends/pytorch/src/transforms/aten_index_replacer.cpp
```diff
@@ -223,8 +219,9 @@ AtenIndexToSelect::AtenIndexToSelect() {
             }
             auto final_shape = std::make_shared<v0::Concat>(concat_dims, 0);
             gather = std::make_shared<v1::Reshape>(gather, final_shape, false);
-            copy_runtime_info({index_op, input_node, indicies}, gather);
+            copy_runtime_info({index_op, indicies}, gather);
             replace_node(index_op, gather);
+            gather->set_friendly_name(index_op->get_friendly_name());
```
mvafin, line 224: Could you inherit `aten::index` name for other cases too? This really helps to debug problems

mvafin (changes requested):

src/frontends/pytorch/src/transforms/aten_index_replacer.cpp
```diff
@@ -125,31 +125,29 @@ AtenIndexToSelect::AtenIndexToSelect() {
...
                 replace_node(index_op, gather);
                 return true;
             }
             auto adv_idx_count = advanced_ids.size();
-            auto rank = input_node->get_input_partial_shape(0).rank();
-            if (rank.is_dynamic()) {
-                FRONT_END_CHECK_IMPLEMENTED(false, "indexing for tensor with dynamic rank is not implemented ");
-            }
+            auto rank = input_node.get_partial_shape().rank();
+            FRONT_END_CHECK_IMPLEMENTED(rank.is_static(), "indexing for tensor with dynamic rank is not implemented ");
```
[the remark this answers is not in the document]
eaidova: done

src/frontends/pytorch/src/transforms/aten_index_replacer.cpp
eaidova: done

src/frontends/pytorch/src/transforms/aten_index_replacer.cpp
```diff
@@ -108,10 +108,10 @@ AtenIndexToSelect::AtenIndexToSelect() {
                         continue;
                     }
                 }
-                auto id_dtype = ids[i].get_node_shared_ptr()->get_element_type();
+                auto id_dtype = ids[i].get_element_type();
                 if (id_dtype == element::boolean || id_dtype == element::u8) {
-                    auto idx = std::make_shared<ov::op::v0::Convert>(ids[i], element::u8);
-                    auto nonzero = std::make_shared<ov::op::v3::NonZero>(idx);
+                    auto idx = std::make_shared<v0::Convert>(ids[i], element::u8);
+                    Output<ov::Node> nonzero = std::make_shared<v3::NonZero>(idx, element::i32);
```
eaidova, line 114:
```suggestion
                    auto nonzero = std::make_shared<v3::NonZero>(idx, element::i32);
```

src/frontends/pytorch/src/transforms/aten_index_replacer.cpp
```diff
@@ -108,10 +108,10 @@ AtenIndexToSelect::AtenIndexToSelect() {
                         continue;
                     }
                 }
-                auto id_dtype = ids[i].get_node_shared_ptr()->get_element_type();
+                auto id_dtype = ids[i].get_element_type();
                 if (id_dtype == element::boolean || id_dtype == element::u8) {
```
mvafin, line 112: Does this work as expected now? Or is it required for types to be know during conversion (so we have to set them in InputModel)?

eaidova: I'm not sure that we can work without type knowledge at all, but I assume in all cases where bool logic should be, it will be already known (parent op returns bool - e.g. where, or some logical ops eq, le, get e.t.c.)

mvafin (approved):

PR #16743, opened 2023-04-04
[AUTO] Initialize variable / reduce variable copy

peterchen-intel:
### Details:
 - *Initialize variable*
 - *Use reference to reduce variable copy*

### Tickets:
 - *CVS-107195*

### DIFF
--- a/src/plugins/auto/src/auto_executable_network.cpp
+++ b/src/plugins/auto/src/auto_executable_network.cpp
@@ -237,7 +237,7 @@ IE::Parameter AutoExecutableNetwork::GetMetric(const std::string& name) const {
         if (_autoSchedule->_pCTPUTLoadContext) {
             std::vector<std::string> exeDevices = {};
             std::lock_guard<std::mutex> lock(_autoSContext->_confMutex);
-            for (auto & n : _autoSContext->_devicePriorities) {
+            for (auto const & n : _autoSContext->_devicePriorities) {
                 exeDevices.push_back(n.deviceName);
             }
             execution_devices = decltype(ov::execution_devices)::value_type {exeDevices};
--- a/src/plugins/auto/src/auto_executable_network.cpp
+++ b/src/plugins/auto/src/auto_executable_network.cpp
@@ -237,7 +237,7 @@ IE::Parameter AutoExecutableNetwork::GetMetric(const std::string& name) const {
         if (_autoSchedule->_pCTPUTLoadContext) {
             std::vector<std::string> exeDevices = {};
             std::lock_guard<std::mutex> lock(_autoSContext->_confMutex);
-            for (auto n : _autoSContext->_devicePriorities) {
+            for (auto & n : _autoSContext->_devicePriorities) {
                 exeDevices.push_back(n.deviceName);
             }
             execution_devices = decltype(ov::execution_devices)::value_type {exeDevices};
--- a/src/plugins/auto/src/auto_schedule.cpp
+++ b/src/plugins/auto/src/auto_schedule.cpp
@@ -238,12 +238,7 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
             // Generate contexts for loading each device
             _pCTPUTLoadContext.reset(new AutoLoadContext[_nCTputDeviceNums]);
             int idx = 0;
-            /*  To be compatible with C++11 but also fix the scan warning. https://stackoverflow.com/questions/18184096
-                If we limit to use std >= C++14 one day, then we can initilize them in DeviceInformation definition directly
-                DeviceInformation cpuDeviceInformation = {.numRequestsPerDevices = -1, .devicePriority = 0};
-                Failed for Windows error C7555: use of designated initializers requires at least '/std:c++latest'
-                */
-            DeviceInformation cpuDeviceInformation = {std::string(), {}, -1, std::string(), std::string(), 0};
+            DeviceInformation cpuDeviceInformation;
             for (auto& device : validDevices) {
                 if (device.deviceName.find("CPU") == std::string::npos) {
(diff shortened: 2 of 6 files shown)


src/plugins/auto/auto_schedule.cpp
```diff
@@ -247,7 +247,12 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
...
             _pCTPUTLoadContext.reset(new AutoLoadContext[_nCTputDeviceNums]);
             int idx = 0;
-            DeviceInformation cpuDeviceInformation;
+            /*  To be compatible with C++11 but also fix the scan warning. https://stackoverflow.com/questions/18184096
+                If we limit to use std >= C++14 one day, then we can initilize them in DeviceInformation definition directly
+                DeviceInformation cpuDeviceInformation = {.numRequestsPerDevices = -1, .devicePriority = 0};
+                Failed for Windows error C7555: use of designated initializers requires at least '/std:c++latest'
+                */
+            DeviceInformation cpuDeviceInformation = {std::string(), {}, -1, std::string(), std::string(), 0};
             for (auto& device : validDevices) {
```
songbell, line 256: this looks kind of redundant:
1) why can't initialize in struct constructor
2) if change to a pointer helps? DeviceInformation* cpuDeviceInformation;

src/plugins/auto/auto_executable_network.cpp
```diff
@@ -237,7 +237,7 @@ IE::Parameter AutoExecutableNetwork::GetMetric(const std::string& name) const {
         if (_autoSchedule->_pCTPUTLoadContext) {
             std::vector<std::string> exeDevices = {};
             std::lock_guard<std::mutex> lock(_autoSContext->_confMutex);
-            for (auto n : _autoSContext->_devicePriorities) {
+            for (auto & n : _autoSContext->_devicePriorities) {
```
yangwang201911, line 240: `auto&&`

src/plugins/auto/auto_schedule.cpp
yangwang201911, line 256: same as above.

src/plugins/auto/multi_executable_network.cpp
```diff
@@ -108,7 +108,7 @@ IE::Parameter MultiExecutableNetwork::GetMetric(const std::string& name) const {
         };
     } else if (name == ov::device::properties) {
         ov::AnyMap all_devices = {};
-        for (auto network : _multiSContext->_networksPerDevice) {
+        for (auto & network : _multiSContext->_networksPerDevice) {
```
yangwang201911, line 111: same as above.

src/plugins/auto/multi_executable_network.cpp
```diff
@@ -123,7 +123,7 @@ IE::Parameter MultiExecutableNetwork::GetMetric(const std::string& name) const {
         return all_devices;
     } else if (name == ov::optimal_number_of_infer_requests) {
         unsigned int res = 0u;
-        for (auto n : _multiSContext->_networksPerDevice) {
+        for (auto & n : _multiSContext->_networksPerDevice) {
```
yangwang201911, line 126: same as above.

src/plugins/auto/plugin.cpp
```diff
@@ -300,7 +300,7 @@ InferenceEngine::Parameter MultiDeviceInferencePlugin::GetMetric(const std::stri
     } else if (name == METRIC_KEY(OPTIMIZATION_CAPABILITIES)) {
         auto deviceList = GetCore()->GetAvailableDevices();
         std::vector<std::string> capabilities;
-        for (auto device : deviceList) {
+        for (auto & device : deviceList) {
```
yangwang201911, line 303: same as above.

src/plugins/auto/auto_executable_network.cpp
songbell: do we have a reason && is a must? if no, I prefer const auto&

peterchen-intel: https://stackoverflow.com/questions/29859796   Shall we use auto const & ?

yangwang201911: From my understanding, `auto &` works well if it doesn't bind to a temporary object.
`const auto&` could do that, while it doesn't allow to change the binding object.
`auto&&` could bind to both l-value reference and r-value reference, and also allow us to modify the value of the object.

riverlijunjie: If we don't want to modify element value, `const auto&` is a good choice.
`auto&&` maybe more common for lvalue or rvalue reference, but if we don't want to modify its value, `const auto&` is recommended.

src/plugins/auto/common.hpp
```diff
@@ -69,6 +69,15 @@ struct DeviceInformation {
     std::string defaultDeviceID;
     DeviceName uniqueName;
     unsigned int devicePriority;
+    DeviceInformation () : numRequestsPerDevices(-1), devicePriority(0) {}
+    DeviceInformation (DeviceName dn, std::map<std::string, std::string> conf, int nReq, std::string defaultID, DeviceName uName, unsigned int priority) {
```
riverlijunjie, line 73: Where is the second constructor applied?

peterchen-intel: https://github.com/openvinotoolkit/openvino/blob/06cacfe2a70d6ff5/src/tests/unit/auto/exec_network_get_metrics.cpp#L251
https://github.com/openvinotoolkit/openvino/blob/06cacfe2a70/src/plugins/auto/plugin.cpp#L244
Actually, the parameter number can be from 0 and up to 6, so I am adding another commit.

wangleis (approved):

PR #16751, opened 2023-04-05
Fix for slowdown of convert_model() after multiple runs

popovaan:
Root cause analysis: 
Usage of subprocess with torch.jit.freeze() leads to slow down of subprocesses, which leads to slowdown of each convert_model() run for PyTorch.

Solution: 
Subprocesses are used in MO for getting of MO version and for checking of runtime dependencies. It can be moved to singleton class and checked once during first run of convert_model(). So during next runs subprocesses are not used.

Ticket: 107045

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update


mvafin (approved):

slyalin:
Do we really need all these version checks at all? What is the value? We are on the way to merge mo to ov, so can we decide that we will remove the checks after the merge?

tools/mo/openvino/tools/mo/utils/version.py
```diff
@@ -81,3 +85,64 @@ def extract_hash_from_version(full_version: str):
...
+    def __init__(self, cls_name, super_classes, dic):
+        self.__single_instance = None
+        super().__init__(cls_name, super_classes, dic)
+
+    def __call__(cls, *args, **kwargs):
+        if cls.__single_instance is None:
+            cls.__single_instance = super(SingletonMetaClass, cls).__call__(*args, **kwargs)
+        return cls.__single_instance
+
+
```
rkazants, line 100: It looks that we can refactor this file. I mean to make it more clear `simplify_version`, `get_simplified_mo_version`. I believe not all of functions above are needed and we can refactor and incorporate them inside the existing functions of `VersionChecker`. You can consider such refactoring in the next PR. it is up-to-you

tools/mo/unit_tests/mo/convert/version_checker_test_actual.py
```diff
@@ -0,0 +1,41 @@
...
+        for _ in range(100):
+            VersionChecker().check_runtime_dependencies()
+            assert VersionChecker().get_mo_version() == ref_mo_version
+            assert VersionChecker().get_ie_version() == ref_ie_version
+            assert VersionChecker().get_mo_simplified_version() == ref_mo_simplified_version
+            assert VersionChecker().get_ie_simplified_version() == ref_ie_simplified_version
+        loop_time = (datetime.datetime.now() - start_time).total_seconds()
+
+        # Check that time of loop is less than first init, so no actual initialization happens
+        assert loop_time < first_init_time
```
rkazants, line 41: I would say that loop_time / 100 is less than `first_init_time`. Please check

rkazants (approved):

tools/mo/unit_tests/mo/convert/version_checker_test_actual.py
```diff
@@ -0,0 +1,41 @@
...
+
+        # Loop with multiple usages of VersionChecker
+        start_time = datetime.datetime.now()
+        for _ in range(100):
+            VersionChecker().check_runtime_dependencies()
+            assert VersionChecker().get_mo_version() == ref_mo_version
+            assert VersionChecker().get_ie_version() == ref_ie_version
+            assert VersionChecker().get_mo_simplified_version() == ref_mo_simplified_version
+            assert VersionChecker().get_ie_simplified_version() == ref_ie_simplified_version
+        loop_time = (datetime.datetime.now() - start_time).total_seconds()
```
rkazants, line 38:
```suggestion
        loop_time = (datetime.datetime.now() - start_time).total_seconds() / 100.0
```

rkazants: You can consider this fix in the next PR, if I am correct.

tools/mo/unit_tests/mo/convert/version_checker_test_actual.py
popovaan: Both works, as first init time is significantly longer than even whole loop. On my machine I get following time:

first init time: 0.26 sec
loop time: 7.4e-05 sec
loop time / 100: 7.4e-07 sec

So I can change the check to `loop_time / 100`, but initial check looks safe to use too.

rkazants: I am saying that just to avoid sporadic failures in CI, for example

PR #16760, opened 2023-04-05
Fix cycle dependency check in snippets tokenizer

olpipi:
### Details:
 - Fix cycle dependency check in snippets tokenizer

### Tickets:
 - It was discovered in scope of [CVS-104357](https://jira.devtools.intel.com/browse/CVS-104357)
 - Add test


src/tests/functional/plugin/shared/include/snippets/two_inputs_and_outputs.hpp
```diff
@@ -26,6 +26,11 @@ class TwoInputsAndOutputs : public testing::WithParamInterface<ov::test::snippet
     void SetUp() override;
 };
 
+class TwoInputsAndOutputsWithReversedOutputs : public TwoInputsAndOutputs {
```
a-sidorova, line 29: Honestly it'd be great if you add the comment what is verified by this test (why we add almost the same test as above), smth like: The `Result` ops have the specific order in topological sorting. The test checks the correctness of the algorithm for checking possible cyclic dependency in tokenization.

Because after month we'll forget why we have almost the same tests, what is the difference between them :smile:

a-sidorova (approved):

a-sidorova:
FYI @IvanNovoselov

src/tests/functional/plugin/shared/include/snippets/two_inputs_and_outputs.hpp
olpipi: done

dmitry-gorokhov (approved):

PR #16765, opened 2023-04-05
Disable ConstantFolding for ShapeOf subgraph in TS transformation

itikhono:
Disable ConstantFolding for ShapeOf subgraph in TS transformation

### Tickets:
 - *101354, *


src/common/transformations/tests/transpose_sinking/ts_general_test.cpp
```diff
@@ -412,6 +412,29 @@ TEST_F(TransformationTestsF, TSGeneralTestMultipleTypes) {
...
+        auto reshape = std::make_shared<Reshape>(transpose, shape_of, false);
+
+        auto ng_order1 = std::make_shared<Constant>(ov::element::u64, ov::Shape{4}, ov::Shape{0, 3, 1, 2});
+        auto transpose1 = std::make_shared<Transpose>(reshape, ng_order1);
+
+        function = std::make_shared<ov::Model>(transpose1, ov::ParameterVector{X, Shape});
+    }
+    manager.register_pass<TSGeneral>();
+}
+
```
rkazants, line 437: can we have a reference for the comparison?

itikhono: we don't need to create a reference here, the transformation won't affect the original network due to ShapeOf subgraph.
a copy of the original network will be used as a reference

rkazants (approved):

mateusztabaka (approved):

PR #16780, opened 2023-04-06
Fixed command lines in MO docs.

popovaan:
Description:
Corrected command in Convert_EfficientDet_Models.md.
Updated missed data_type=True to 'compress_to_fp16'

Ticket: -

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update

### DIFF
--- a/docs/MO_DG/prepare_model/convert_model/tf_specific/Convert_EfficientDet_Models.md
+++ b/docs/MO_DG/prepare_model/convert_model/tf_specific/Convert_EfficientDet_Models.md
@@ -12,55 +12,13 @@ Converting EfficientDet Model to the IR
 There are several public versions of EfficientDet model implementation available on GitHub. This tutorial explains how to
 convert models from the `repository <https://github.com/google/automl/tree/master/efficientdet>`__  (commit 96e1fee) to the OpenVINO format.
 
-Getting a Frozen TensorFlow Model
-+++++++++++++++++++++++++++++++++
-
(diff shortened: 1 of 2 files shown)


docs/MO_DG/prepare_model/convert_model/tf_specific/Convert_EfficientDet_Models.md
```diff
@@ -50,7 +50,7 @@ referenced in the **"Pretrained EfficientDet Checkpoints"** section of the model
 
 .. code-block:: sh
 
-    mo --runmode=saved_model --model_name=efficientdet-d4  --ckpt_path=efficientdet-d4 --saved_model_dir=savedmodeldir
+    python3 model_inspect.py --runmode=saved_model --model_name=efficientdet-d4  --ckpt_path=efficientdet-d4 --saved_model_dir=savedmodeldir
```
rkazants, line 53: I think we need to re-consider this steps. As I understand, `model_inspect.py` is used to freeze the model from SM format. But we support SM format directly so there is no need to use it and we can skip this step

popovaan: Removed freezing step from this tutorial. 
But currently conversion command works only with 'use_legacy_frontend'.
I created ticket to fix it: 107860

rkazants (approved):

tsavina (approved):

PR #16794, opened 2023-04-06
CompressQuantizeWeights transformation update

mateusztabaka:
Use f32 precision when computing scale and zero point

Ticket: 101825


src/common/offline_transformations/src/compress_quantize_weigths.cpp
```diff
@@ -143,8 +143,17 @@ ngraph::pass::CompressQuantizeWeights::CompressQuantizeWeights() {
                     scale = (output_high - output_low) / (new_output_high - new_output_low)
                     zero_point = new_output_low - output_low / scale
             */
-            const auto& output_low = pattern_value_map.at(output_low_pattern);
-            const auto& output_high = pattern_value_map.at(output_high_pattern);
+            Output<Node> output_low = pattern_value_map.at(output_low_pattern);
+            Output<Node> output_high = pattern_value_map.at(output_high_pattern);
+            const auto& fq_type = fq->get_output_element_type(0);
+            if (fq_type != element::f32) {
+                input_low = std::make_shared<opset8::Convert>(input_low, element::f32);
```
itikhono, line 150: can it cause a significant increase in LoadTime due to repackaging of these constants from low_precision (not fp32) to high_precision (fp32) during ConstantFolding?
do we support fp64 precision? how we handle this case?

mateusztabaka: It's not executed during compile_model also the input intervals are usually small (either scalar or a constant with number of elements equal to number of channels). 

> do we support fp64 precision? how we handle this case?

In theory we can have model with fp64 inputs - I handled that case in the next commit.

PR #16795, opened 2023-04-06
[CPU] Execute constants in order with the create primitives calls

maxnick:
### Details:
We need to safely access constant memory on the prepareParams stage. However, there might be so called constant path, where several constant nodes are chained one after another. Since in the current implementation all the constant nodes are executed after the createPrimitive (and prepareParams) call for all nodes, it might lead to hardly detected errors.
In this PR the calls sequence has changed. We run createPrimitive for all nodes together with the execute call for constant nodes in one pass in topological order. That means that now we always have valid data on constant inputs at the moment of the createPrimitive call.

### Tickets:
 - 106650


usstq (approved):
This is good idea, the execution order guaranteed that by the time of `prepareParam()`, all it's const parents have been executed and data is ready to be accessed, great job!

maxnick:
@dmitry-gorokhov could you please make a final review? Internal validation is in progress.

dmitry-gorokhov (approved):

PR #16797, opened 2023-04-06
[CPU] Convert i64->i32 for Reference node.

nshchego:
### Details:
 - *This transformation adds convert i64->i32 to CPU Reference nodes output.*

### Tickets:
 - **

### DIFF
--- a/src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/ref_convert_i64_i32.cpp
+++ b/src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/ref_convert_i64_i32.cpp
@@ -1,36 +1,48 @@
 // Copyright (C) 2023 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
+//
 
 
 #include "ref_convert_i64_i32.hpp"
 #include <openvino/opsets/opset10.hpp>
-#include "transformations/utils/utils.hpp"
 #include "cpu_types.h"
+#include <openvino/pass/pattern/op/wrap_type.hpp>
 
+#include "itt.hpp"
 
-bool ov::intel_cpu::RefConvertI64ToI32::run_on_model(const std::shared_ptr<ov::Model> &model) {
-    const auto orderedOps = model->get_ordered_ops();
-    for (const auto& op : orderedOps) {
-        if (TypeFromName(op->get_type_name()) == Type::Unknown) {
-            for (auto& output : op->outputs()) {
-                if (output.get_element_type() == ov::element::i64 || output.get_element_type() == ov::element::u64) {
-                    auto targetInputs = output.get_target_inputs();
-                    auto convert = std::make_shared<ov::opset10::Convert>(output, ov::element::i32);
+ov::pass::RefConvertI64ToI32::RefConvertI64ToI32() {
+    MATCHER_SCOPE(RefConvertI64ToI32);
 
-                    auto& rt_info = convert->get_rt_info();
-                    rt_info["convert_i64_i32"] = "";
-                    for (const auto& targetInput : targetInputs) {
-                        targetInput.replace_source_output(convert);
-                    }
+    auto i64_extension = [](const ov::Output<ov::Node>& output) -> bool {
+        auto node = output.get_node_shared_ptr();
+        return ov::intel_cpu::TypeFromName(node->get_type_name()) == ov::intel_cpu::Type::Unknown &&
+               ov::pass::pattern::type_matches_any({ov::element::i64, ov::element::u64})(output);
+    };
 
-                    auto& convertTensor = convert->output(0).get_tensor();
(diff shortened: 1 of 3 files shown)


dmitry-gorokhov:
@v-Golubev Please review the PR

src/plugins/intel_cpu/src/transformation_pipeline.cpp
```diff
@@ -243,6 +244,7 @@ void Transformations::PreLpt(const std::vector<ov::element::Type>& defaultPrecis
         manager.register_pass<ngraph::pass::low_precision::ConvertSubtractConstant>(defaultPrecisions);
     }
     manager.register_pass<ov::pass::Validate>();
+    manager.register_pass<RefConvertI64ToI32>();
```
v-Golubev, line 247: Can we place the new transformation before Validate pass?

src/plugins/intel_cpu/src/ngraph_transformations/ref_convert_i64_i32.cpp
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
```
v-Golubev, line 3:
```suggestion
// Copyright (C) 2023 Intel Corporation
// SPDX-License-Identifier: Apache-2.0
//
```

src/plugins/intel_cpu/src/ngraph_transformations/ref_convert_i64_i32.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "ref_convert_i64_i32.hpp"
+#include <openvino/opsets/opset10.hpp>
+#include "transformations/utils/utils.hpp"
+#include "cpu_types.h"
+
+
+bool ov::intel_cpu::RefConvertI64ToI32::run_on_model(const std::shared_ptr<ov::Model> &model) {
+    const auto orderedOps = model->get_ordered_ops();
+    for (const auto& op : orderedOps) {
+        if (TypeFromName(op->get_type_name()) == Type::Unknown) {
```
v-Golubev, line 14: Do I understand correctly that we insert convert only for extension nodes? Can we reflect this in the transformation description?

src/plugins/intel_cpu/src/ngraph_transformations/ref_convert_i64_i32.cpp
```diff
@@ -0,0 +1,36 @@
...
+            for (auto& output : op->outputs()) {
+                if (output.get_element_type() == ov::element::i64 || output.get_element_type() == ov::element::u64) {
+                    auto targetInputs = output.get_target_inputs();
+                    auto convert = std::make_shared<ov::opset10::Convert>(output, ov::element::i32);
+
+                    auto& rt_info = convert->get_rt_info();
+                    rt_info["convert_i64_i32"] = "";
```
v-Golubev, line 21: Why do we need to set a specific rt_info for such converts?

src/plugins/intel_cpu/src/ngraph_transformations/ref_convert_i64_i32.hpp
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include "openvino/pass/pass.hpp"
+
+namespace ov {
+namespace intel_cpu {
+class RefConvertI64ToI32: public ov::pass::ModelPass {
```
v-Golubev, line 11: Ideally, it would be great to inherit from `MatcherPass`: I suppose this approach will be faster than the current one.
If we have some time to refactor the transformation, please let me know: I will bring you the detailed instructions offline

src/plugins/intel_cpu/src/transformation_pipeline.cpp
nshchego: Validate removes this Convert node and pass ConvertPrecision does not work properly.

src/plugins/intel_cpu/src/ngraph_transformations/ref_convert_i64_i32.cpp
nshchego: Done

v-Golubev (approved):
LGTM

dmitry-gorokhov:
This PR lacks tests. We have example of such tests for Extension op [with internal dynamism](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/tests/functional/single_layer_tests/custom_op_internal_dyn.cpp). Need smt similar to cover this case.
Can be done in separate PR.

dmitry-gorokhov (approved):
Remaining comments doesn't functional part, so agreed to fix in follow-up PR.

PR #15863, opened 2023-02-21
[GPU] Softmax for stable diffusion

ytorzuk-altran:
### Details:
 - *Added single layer test which covers Softmax in stable diffusion case*
 - *Improved inputs generation and absolute error threshold*
 - *Scalar intel_sub_group_read/intel_sub_group_read are incorporated*

### Tickets:
 - *CVS-91171*

PR #16798, opened 2023-04-07
Revert "[GPU] Softmax for stable diffusion (#15863)"

yeonbok:
### Details:
 - This commit has accuracy issue in dynamic shape accuracy for bert. 
 - (Example shape : [1, 306])
 - PR15863 was optimizing kernel but not validated for accuracy enough, and the perf gain was not that critical. 
 - Seems that the original commit needs further investigation for various cases in accuracy perspective.

### DIFF
--- a/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/softmax_gpu_bf.cl
+++ b/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/softmax_gpu_bf.cl
@@ -3,27 +3,11 @@
 //
 
 #include "include/batch_headers/common.cl"
-#include "include/batch_headers/fetch_data.cl"
-#include "include/batch_headers/sub_group_block_read.cl"
-#include "include/batch_headers/sub_group_block_write.cl"
-
-#if SUBGROUP_BLOCK_SIZE == 1
-#define BLOCK_READ(ptr, offset) DT_INPUT_BLOCK_READ(ptr, offset)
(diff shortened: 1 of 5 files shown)


vladimir-paramuzov:
Can we fix it instead of revert?

yeonbok:
> > This commit has accuracy issue in dynamic shape accuracy for bert.
> 
> Can we fix it instead of revert?

According to @wilson-seok and @isanghao the commit also caused perf drops for other networks in benchmark. Also the current observed accuracy issue is not actually for dynamic shape specific one, it happened on the static kernel built on runtime. I think we need to come up with better implementation, I thouhgt to reimplementation with reconsidering the WI size (item size) to be well aligned with the subgroup, but I think it is not feasible during this release, because it will need additional perf benchmark & acc check. All in all, even the original target perf implrovement was not enough to ignore other networks perf drop according to the opinion from @wilson-seok  and @isanghao, so it seems better to revert I think

vladimir-paramuzov:
@yeonbok It gives ~40% perf gain on SD2.1, so I wouldn't want to revert it. If we can reimplement this until CF then OK, but if not - let's fix accuracy issue.

yeonbok:
> @yeonbok It gives ~40% perf gain on SD2.1, so I wouldn't want to revert it. If we can reimplement this until CF then OK, but if not - let's fix accuracy issue.

Oh I see, 40% is a big deal. Let me check

PR #16162, opened 2023-03-08
Interpolate-11 spec + core op

tomdol:
### Details:
 - v11::Interpolate specification (new interpolation modes and changes in the inputs handling)
 - v11::Interpolate core operator + tests
 - common base class for the Interpolate ops
 - changes in the Interpolate shape inference and preparation for a better refactor
 - removal of the deprecated enum values
 - missing opset11 in the ngraph namespace

### Tickets:
 - 104322
 - 104323

PR #16806, opened 2023-04-07
[GPU] Fix Interpolate assert

sshlyapn:
### Details:
 - Relaxed restrictions for Interpolate operation with `SIZES` shapes calculation mode (ShapeCalcMode::SIZES), because scales input may not be fully configured during InterpolateV11 -> InterpolateV4 conversion, but GPU Plugin can handle such cases
 - Added new tests cases
 - Fixed evaluation function for InterpolateV11

### Tickets:
 - 106893

### DIFF
--- a/src/plugins/intel_gpu/src/graph/resample.cpp
+++ b/src/plugins/intel_gpu/src/graph/resample.cpp
@@ -61,7 +61,7 @@ std::vector<layout> resample_inst::calc_output_layouts(resample_node const& /*no
 
     if (((sizes_data.empty() && !memory_deps.count(1)) || !sizes_calc_mod) &&
         ((scales_data.empty() && !memory_deps.count(2)) || sizes_calc_mod)) {
-       return { layout{ShapeType::dynamic(input_rank), input_layout.data_type, input_layout.format} };
+        return { layout{ShapeType::dynamic(input_rank), input_layout.data_type, input_layout.format} };
     }
 
     auto axes_data = desc->axes;
--- a/src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
+++ b/src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
@@ -49,10 +49,9 @@ static void CreateInterpolateOp(Program& p, const std::shared_ptr<ngraph::op::v4
         }
     }
 
-    bool correct_params = !scales_constant;
-    correct_params |= axes.size() == scales.size();
-    correct_params |= attrs.shape_calculation_mode == ov::op::v4::Interpolate::ShapeCalcMode::SIZES;
-    OPENVINO_ASSERT(correct_params, "[GPU] Incorrect axes and scales values for Interpolate operation with id ", op->get_friendly_name());
+    if (attrs.shape_calculation_mode == ov::op::v4::Interpolate::ShapeCalcMode::SCALES && scales_constant) {
+        OPENVINO_ASSERT(axes.size() == scales.size(), "[GPU] Incorrect axes and scales values for Interpolate operation with id ", op->get_friendly_name());
+    }
 
     // TODO shouldn't be all this checking done in ngraph::op::v4::Interpolate?
     auto interpolateMode = attrs.mode;
--- a/src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
+++ b/src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
@@ -49,7 +49,10 @@ static void CreateInterpolateOp(Program& p, const std::shared_ptr<ngraph::op::v4
         }
     }
 
-    OPENVINO_ASSERT(!scales_constant || axes.size() == scales.size(), op->get_friendly_name(), " Incorrect axes and scales should be the same size");
+    bool correct_params = !scales_constant;
+    correct_params |= axes.size() == scales.size();
+    correct_params |= attrs.shape_calculation_mode == ov::op::v4::Interpolate::ShapeCalcMode::SIZES;
+    OPENVINO_ASSERT(correct_params, "[GPU] Incorrect axes and scales values for Interpolate operation with id ", op->get_friendly_name());
 
     // TODO shouldn't be all this checking done in ngraph::op::v4::Interpolate?
     auto interpolateMode = attrs.mode;
--- a/src/plugins/template/backend/evaluates_map.cpp
+++ b/src/plugins/template/backend/evaluates_map.cpp
@@ -4272,7 +4272,7 @@ bool evaluate_interpolate(const shared_ptr<op::v11::Interpolate>& op,
         padded_input_shape.emplace_back(m_attrs.pads_begin[i] + m_attrs.pads_end[i] + input_shape[i].get_length());
     }
 
-    auto axes = get_axes_vector(inputs, inputs[1]->get_shape().size(), axes_port, max_num_of_ports);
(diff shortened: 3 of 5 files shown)


src/plugins/template/backend/evaluates_map.cpp
```diff
@@ -4272,7 +4272,7 @@ bool evaluate_interpolate(const shared_ptr<op::v11::Interpolate>& op,
         padded_input_shape.emplace_back(m_attrs.pads_begin[i] + m_attrs.pads_end[i] + input_shape[i].get_length());
     }
 
-    auto axes = get_axes_vector(inputs, inputs[1]->get_shape().size(), axes_port, max_num_of_ports);
+    auto axes = get_axes_vector(inputs, input_shape.size(), axes_port, max_num_of_ports);
```
mitruska, line 4275: Was there any failing test because of the `inputs[1]->get_shape().size()` default axes size?

There is unresolved discussion about adjusting the new Interpolate-11 spec: https://github.com/openvinotoolkit/openvino/pull/16162#discussion_r1138471720
 because making default `axes` size equal to `data` rank, may lead to bigger size than actual the elements count in the corresponding `sizes_scales` input.
 
 cc: @tomdol

src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
```diff
@@ -49,7 +49,10 @@ static void CreateInterpolateOp(Program& p, const std::shared_ptr<ngraph::op::v4
         }
     }
 
-    OPENVINO_ASSERT(!scales_constant || axes.size() == scales.size(), op->get_friendly_name(), " Incorrect axes and scales should be the same size");
+    bool correct_params = !scales_constant;
+    correct_params |= axes.size() == scales.size();
```
mitruska, line 53: It can fail if the axes are default and scales are provided with smaller size than data rank,
for example `[H, W, C]` data input, scales `[0.5, 0.6]` then default axes `[0, 1, 2]` so `axes.size() != scales.size()`
cc: @tomdol

src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
```diff
@@ -49,7 +49,10 @@ static void CreateInterpolateOp(Program& p, const std::shared_ptr<ngraph::op::v4
...
+    correct_params |= attrs.shape_calculation_mode == ov::op::v4::Interpolate::ShapeCalcMode::SIZES;
```
mitruska, line 54: Shouldn't "scales" input be related to the `ShapeCalcMode::SCALES`?
```suggestion
    correct_params |= attrs.shape_calculation_mode == ov::op::v4::Interpolate::ShapeCalcMode::SCALES;
```

src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
```diff
@@ -49,7 +49,10 @@ static void CreateInterpolateOp(Program& p, const std::shared_ptr<ngraph::op::v4
...
+    OPENVINO_ASSERT(correct_params, "[GPU] Incorrect axes and scales values for Interpolate operation with id ", op->get_friendly_name());
```
mitruska, line 55: Is there any reason why only "scales_constant" input is validated, but there were no such checks for the "sizes_constant" ?

src/plugins/template/backend/evaluates_map.cpp
sshlyapn: Yes, there are several newly added tests which failed w/o this change
For example, in case of _ShapeCalcMode::SCALES_ mode and `[N, C, H, W]` data input with scales_or_sizes = `[1, 1, 0.5, 2.5]`, and empty axes input, then default axes will be calculated as `[0]`, because `inputs[1]->get_shape().size()` returns the rank of 1D tensor. So, this definitely should be fixed. But question is how? As a result of discussion with @tomdol about the behavior of default axes, it was fixed in the same manner as it is calculated for Interpolate_v4

src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
sshlyapn: Is it applicable for Interpolate_v4? According to specs, user should specify axes input explicitly or `[0,...,rank(data) - 1]` will be used as default values. So, shouldn't user in your example specify scales with 3 values or specify axes with 2 explicitly?

src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
sshlyapn: Yeah, you are right. And my idea was to allow further execution in two scenarios: either `axes.size() == scales.size();` and we can verify it (scales_constant exists) or current mode is `ShapeCalcMode::SIZES`, and we can ignore scales related conditions.
Maybe it is better to rewrite the condition in the following way:
```
    if (attrs.shape_calculation_mode == ov::op::v4::Interpolate::ShapeCalcMode::SCALES && scales_constant) {
        OPENVINO_ASSERT(axes.size() == scales.size(), "[GPU] Incorrect axes and scales values for Interpolate operation with id ", op->get_friendly_name());
    }
```
Since this `axes.size() == scales.size()` condition should be checked in case of ShapeCalcMode::SCALES mode and if scales constants are available. What do you think? (Both non-constant sizes and scales are supported by GPU Plugin)

src/plugins/template/backend/evaluates_map.cpp
mitruska: Oh I see, good catch, so the original intention here, was to make number of default axes equal to the elements count in the `scales_sizes` instead of rank:
 Like `inputs[1]->get_shape()[0]` rather than `inputs[1]->get_shape().size()` or `input_shape.size()`
 👇 
`    auto axes = get_axes_vector(inputs, inputs[1]->get_shape()[0], axes_port, max_num_of_ports);
`
or eqivalent
`    auto axes = get_axes_vector(inputs, shape_size(inputs[1]->get_shape()), axes_port, max_num_of_ports);
`

sshlyapn: Got it, thanks, I'll update it

src/plugins/intel_gpu/src/plugin/ops/interpolate.cpp
mitruska: We are trying to clarify that:) There is some space for different interpretation of this restriction, but it was possible to create such v4::Interpolate op, what doesn't mean it was supported by plugins: https://github.com/openvinotoolkit/openvino/blob/8c6910043916d5aa6e5dd17b786eb095854eed62/src/core/tests/visitors/op/interpolate.cpp#L48-L64

sshlyapn: @mitruska  Oh, I see... It's a really tricky thing
Are you okay to merge this PR before this question will be solved? It fixes quite a lot of issues, but doesn't change the behavior dramatically. But if it will be decided to allow patterns like you provided, we can handle it in separate PR

src/plugins/intel_gpu/src/graph/resample.cpp
p-durandin, line 1: Please clean up excessive space

sshlyapn: I just aligned tab with 4 spaces there

jane-intel (approved):
Template part approved.

PR #16814, opened 2023-04-07
[PyOV] Fix models checking and ensure correct destructor calls in tests

jiwaszki:
### Details:
 - Models and their operators were not correctly checked in various test cases. Fixing it with `compare_models` helper method to ensure correctness of models.
 - Removed reference bumping done by incorrect comparisons, allow new mmap feature to work correctly while testing with temporary files.
 - Refactored tests.

### Tickets:
 - *106805*


src/bindings/python/tests/test_graph/test_manager.py
```diff
@@ -153,39 +150,16 @@ def test_serialize_results(prepare_ir_paths):
...
-    pass_manager.register_pass(Serialize(path_to_xml=xml_path, path_to_bin=bin_path))
-    pass_manager.run_passes(func)
+    pass_manager.run_passes(model)
 
     res_model = core.read_model(model=xml_path, weights=bin_path)
-    const = func.get_results()
-    new_const = res_model.get_results()
 
-    assert const == new_const
+    assert compare_functions(model, res_model)[0] is True
```
vurusovs, line 162: If use of `==` isn't correct and user should apply `compare_functions()`, may we document it? From my perspective, it isn't too clear what should I use if it works for `Parameter`, but not for `Result`

vurusovs (changes requested):

src/bindings/python/tests/test_graph/test_manager.py
jiwaszki: Requested rereview. Now solution is purely based on existing public API so additional module is not required to be included in installation configuration/package.

akuporos (approved):

src/bindings/python/tests/test_utils/test_utils.py
```diff
@@ -17,14 +17,100 @@
...
 
-def test_compare_models():
-    try:
-        from openvino.test_utils import compare_models
-        model = get_relu_model()
-        status, _ = compare_models(model, model)
-        assert status
-    except RuntimeError:
-        print("openvino.test_utils.compare_models is not available")  # noqa: T201
+def _compare_models(model_one: Model, model_two: Model, compare_names: bool = True) -> Tuple[bool, str]:  # noqa: C901 the function is too complex
```
vurusovs, line 20: @itikhono @jane-intel please, take a look

vurusovs (approved):

src/bindings/python/tests/test_utils/test_utils.py
jiwaszki: @itikhono @jane-intel please let me know directly if you have any concerns.

PR #16819, opened 2023-04-09
[TF FE] Test ResourceGather operation and fix debug caps

rkazants:
**Details:** This PR fixes the debug or failures reporting capabilities. Namely, it fixes a case when model contains multiple nodes of the same type for which the existing translator failures. In this case, we should report only failure message but we must not report `No translator found`.
Also, in this PR we represent testing a model with `ResourceGather` operation.

**Tickets:** 92818, 89082

### DIFF
--- a/src/frontends/tensorflow/src/frontend.cpp
+++ b/src/frontends/tensorflow/src/frontend.cpp
@@ -38,14 +38,19 @@ void get_unsupported_operations_and_failures(const std::shared_ptr<Model>& model
     for (const auto& node : model->get_ordered_ops()) {
         if (const auto& fw_node = ov::as_type_ptr<FrameworkNode>(node)) {
             auto op_type = fw_node->get_decoder()->get_op_type();
+            // if this operation is encountered among unsupported operations
+            // or conversion failures, skip it
+            if (failures.count(op_type) > 0 ||
+                std::find(unsupported_operations.begin(), unsupported_operations.end(), op_type) !=
+                    unsupported_operations.end()) {
+                continue;
+            }
             auto fw_node_attrs = fw_node->get_attrs();
-            if (fw_node_attrs.find(FrameworkNode::failed_conversion_key) != fw_node_attrs.end() &&
-                failures.count(op_type) == 0) {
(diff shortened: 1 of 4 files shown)


gkrivor (approved):

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -38,14 +38,19 @@ void get_unsupported_operations_and_failures(const std::shared_ptr<Model>& model
     for (const auto& node : model->get_ordered_ops()) {
         if (const auto& fw_node = ov::as_type_ptr<FrameworkNode>(node)) {
             auto op_type = fw_node->get_decoder()->get_op_type();
+            // if this operation is encountered among unsupported operations
+            // or conversion failures, skip it
+            if (failures.count(op_type) > 0 ||
+                std::find(unsupported_operations.begin(), unsupported_operations.end(), op_type) !=
```
itikhono, line 44: please use set instead of vector.
If I remember correctly you said that in the tests the order of "error messages" is hardcoded, I believe you can easily fix the tests if you use search instead of the order.

src/frontends/tensorflow/src/frontend.cpp
```diff
@@ -38,14 +38,19 @@ void get_unsupported_operations_and_failures(const std::shared_ptr<Model>& model
...
+                    unsupported_operations.end()) {
+                continue;
+            }
             auto fw_node_attrs = fw_node->get_attrs();
-            if (fw_node_attrs.find(FrameworkNode::failed_conversion_key) != fw_node_attrs.end() &&
-                failures.count(op_type) == 0) {
+            if (fw_node_attrs.find(FrameworkNode::failed_conversion_key) != fw_node_attrs.end()) {
```
itikhono, line 49: const auto& attr = fw_node_attrs.find(FrameworkNode::failed_conversion_key);
```
if (att != fw_node_attrs.end()) {
    // save only the first encountered failure that is more improtant for developer
    // that means the translator is found but the conversion is failed
    failures[op_type] = *attr;
```

PR #16587, opened 2023-03-28
[Snippets] BF16 enforce in snippets

eshoguli:
### Details:
 - *BF16 enforce in snippets*

### Tickets:
 - *105219*

PR #16822, opened 2023-04-09
[Snippets] BF16 enforce in snippets extended

eshoguli:
### Details:
 - *BF16 enforce in snippets*
 - *based on: https://github.com/openvinotoolkit/openvino/pull/16587 + optimizations*

### Tickets:
 - *105219*

### DIFF
--- a/src/common/snippets/src/pass/propagate_precision.cpp
+++ b/src/common/snippets/src/pass/propagate_precision.cpp
@@ -207,17 +207,85 @@ bool ngraph::snippets::pass::PropagatePrecision::validate_and_infer_types_and_re
         op->validate_and_infer_types();
     }
 
+    // one flag for both options: only one result consumer
+    bool result_consumer_was_handled = false;
     for (size_t i = 0; i < op->get_output_size(); ++i) {
         auto output = op->output(i);
 
         if (output.get_element_type() != op_output_types[i]) {
             was_updated = true;
+            auto inputs_to_handle = output.get_target_inputs();
+            // if at least consumer is result then we should add conversion
+            // in this case we can not remove existing conversion with result consumers
+            const auto least_consumer_is_result = std::any_of(
+                inputs_to_handle.begin(),
+                inputs_to_handle.end(),
+                [](const Input<Node>& input) { return ngraph::is_type<ngraph::op::Result>(input.get_node()); });
+
+            // option #1: if convertion after the operation exists then check if possible to remove it instead add new
+            auto it = inputs_to_handle.begin();
+            while (it != inputs_to_handle.end()) {
+                const auto& child = it->get_node();
+                const auto existing_convert = ov::as_type<ngraph::snippets::op::ConvertSaturation>(child);
+                if (existing_convert == nullptr) {
+                    it++;
+                    continue;
+                }
+
+                const element::Type actual_before = output.get_element_type();
+                const element::Type actual_after = op_output_types[i];
+                const element::Type required_after = existing_convert->output(0).get_element_type();
+                if (can_be_removed(actual_before, actual_after, required_after)) {
+                    const auto convert_target_inputs = existing_convert->output(0).get_target_inputs();
+
+                    // before removal we should check result cunsumers
+                    if (least_consumer_is_result && std::any_of(
+                            convert_target_inputs.begin(),
+                            convert_target_inputs.end(),
+                            [](const Input<Node>& input) { return ngraph::is_type<ngraph::op::Result>(input.get_node()); })) {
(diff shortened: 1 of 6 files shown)


src/common/snippets/src/pass/propagate_precision.cpp
```diff
@@ -212,6 +212,42 @@ bool ngraph::snippets::pass::PropagatePrecision::validate_and_infer_types_and_re
 
         if (output.get_element_type() != op_output_types[i]) {
             was_updated = true;
+
+            // option #1: if convertion after the operation exists then check if possible to remove it instead add new
+            const auto& target_inputs = output.get_target_inputs();
+            // limit by 1 child, while the issue was not resolved
+            // issue #107966
```
a-sidorova, line 219: Honestly I don't understand what is a problem with `output.get_target_inputs().size()`.
Let's to discuss offline

eshoguli: extended

a-sidorova: I see that now the optimization works only these cases when all target inputs are supported `Convert` nodes. But I don't see the problems to support cases still when just some of target inputs (not all) are `Convert` nodes.

Just imagine.
We have `Node_0` with `0`-th output where 3 Nodes are connected: `target_inputs = {Node_1, Convert_0, Node_2}`

We iterate by `target_inputs` and verify each `target_input`: if we can remove Convert, we do it.
`Node_1` - isn't `Convert`
`Convert_0` - is `Convert` and call `can_be_removed` returns `true`. We remove this Convert from `target_inputs`:
```
// output.get_target_inputs().size() is increasing in line below
// as result we need to remove existing child_input manually
 // issue #107966
existing_convert->output(0).replace(output);
output.remove_target_input(target_input);
```
It's safe because we iterate by local set `target_inputs` (it's not changed), not by `node->get_output_target_inputs(0)` which can be changed during the optimization.

As for `Result input tensor renaming`, you can do it before `Convert` removing:
```
existing_convert_target_inputs = existing_convert->get_output_target_inputs(0);
for (const auto& input : existing_convert_target_inputs) {
      if (ngraph::is_type<ngraph::op::Result>(input.get_node())) {
            // Result input tensor name was changed, the name has to be restored
            // issue #107826
            input.get_tensor_ptr()->add_names(output.get_tensor_ptr()->get_names());
            op->set_friendly_name(existing_convert->get_friendly_name());
       }
}
```

And after renaming, you can call `Convert` removing (first code part).

What do you think?  Maybe I missed something.

eshoguli: not actual more: reworked to support more cases and `Result` validation, thaks for comment

iefode (approved):
@itikhono Please take a look to graph comparator changes

src/common/snippets/src/pass/propagate_precision.cpp
```diff
@@ -207,36 +207,113 @@ bool ngraph::snippets::pass::PropagatePrecision::validate_and_infer_types_and_re
...
 
         if (output.get_element_type() != op_output_types[i]) {
             was_updated = true;
+            auto inputs_to_handle = output.get_target_inputs();
+            // if at least consumer is result then we should add conversion
+            // in this case we can not remove existing conversion with result consumers
+            const auto least_consumer_is_result = std::any_of(
+                inputs_to_handle.begin(),
+                inputs_to_handle.end(),
+                [](const Input<Node>& input) { return ngraph::is_type<ngraph::op::Result>(input.get_node()); });
```
a-sidorova, line 223: Seems like we can create lambda function since we call this check several times:
```
auto is_consumer_result = [](const std::set<ov::Input<ov::Node>> consumers) {
    return  std::any_of(inputs_to_handle.begin(), inputs_to_handle.end(),
                [](const Input<Node>& input) { return ngraph::is_type<ngraph::op::Result>(input.get_node()); });
};
```

src/common/snippets/src/pass/propagate_precision.cpp
```diff
@@ -207,36 +207,113 @@ bool ngraph::snippets::pass::PropagatePrecision::validate_and_infer_types_and_re
...
+                    }
+
+                    // output.get_target_inputs().size() is increasing in line below
+                    // as result we need to remove existing child_input manually
+                    // issue #107966
+                    existing_convert->output(0).replace(output);
+                    output.remove_target_input(*it);
+
+                    // use only affected inputs of conversion
+                    for (auto& input : convert_target_inputs) {
```
a-sidorova, line 258: Minor
```suggestion
                    for (const auto& input : convert_target_inputs) {
```

akladiev:

PR #16824, opened 2023-04-09
[CPU] WA: Stop fusing per-OC eltwise into Matmul with input rank >4

usstq:
### Details:
 - oneDNN3.1's brgemm_avx512_amx kernel has limitations on highest ranks supported when binary post-ops is fused into matmul. we workaround it by not fusing per-OC eltwise into 6D matmul

### Tickets:
 - *CVS-84056*


usstq:
@maxnick, can you help to review this PR?

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -456,24 +456,110 @@ std::pair<Shape, Shape> MatMul::makeDummyInputShapes(const Shape& in0, const Sha
...
+        // we can collapse batch dimensions into m using memory::desc::reshape()
+        // to form a equivalent but simpler 2D problem:
+        //  in0:  (bs0*bs1...*m, k)
+        //  in1:  (k, n)
+        //  bias: (1, n)
+        //  out:  (bs0*bs1...*m, n)
+        int rank = in0_dims.size();
+        auto in1_dims = in1.get_dims();
+        auto out_dims = out.get_dims();
+        do {
```
maxnick, line 486: It would be much better to refactor this part using the Guard Clauses. Just replace the `do while` and `break` with return statements:
```cpp
    if (in1_dims.size() != rank || out_dims.size() != rank) return simplified;
    ...
    for (int i = 0; i < rank-2; i++) {
    bs_collapsed *= in0_dims[i];
    if (in1_dims[i] != 1 || out_dims[i] != in0_dims[i])
        return simplified;
    }
    ...
    if (!new_in0) return simplified;
    ...
    /* and so on*/
```

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -456,24 +456,110 @@ std::pair<Shape, Shape> MatMul::makeDummyInputShapes(const Shape& in0, const Sha
     return {Shape(inDims0), Shape(inDims1)};
 }
 
+static bool SimplifyAs2D(dnnl::memory::desc& in0,
```
maxnick, line 459: It does look like the return value is not used.

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -456,24 +456,110 @@ std::pair<Shape, Shape> MatMul::makeDummyInputShapes(const Shape& in0, const Sha
...
+    if (withBiases) {
+        bias = getBiasDescFrom(outDataDesc);
+    }
+
+    SimplifyAs2D(in0, in1, bias, out);
+
+    const auto attr = initPrimitiveAttr(DnnlExtensionUtils::convertToVectorDims(out.get_dims()));
+
+    //if (getName() == "MatMul_298")
+    //    asm("int3");
```
maxnick, line 547: Commented code

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -588,8 +684,21 @@ void MatMul::prepareParams() {
...
 
-    MatMulKey key = {src0TransposedDesc, src1TransposedDesc, dnnlBiasMemDesc,
-                     dstDnnlDesc, *attr, selected_pd->getImplementationType()};
+    memory::desc descIn0 = src0TransposedDesc->getDnnlDesc();
+    memory::desc descIn1 = src1TransposedDesc->getDnnlDesc();
+    memory::desc descBias = dnnlBiasMemDesc ? dnnlBiasMemDesc->getDnnlDesc() : dnnl::memory::desc();
+    memory::desc descOut = dstDnnlDesc->getDnnlDesc();
+
+    SimplifyAs2D(descIn0, descIn1, descBias, descOut);
+
```
maxnick, line 693: This routine will inevitable introduce overheads to `prepareParams` call, but as far as I understand it, it is necessary only  for 6D case, so can we extract this logic to a some routine that can be initialized on the compile stage (since the rank is a constant value) and used polymorphically. The routine (m.b. some functor class) will be initialized on the compile stage since the rank is a constant value. These conversions:
```cpp
{DnnlExtensionUtils::makeDescriptor(descIn0),
                     DnnlExtensionUtils::makeDescriptor(descIn1),
                     dnnlBiasMemDesc ? DnnlExtensionUtils::makeDescriptor(descBias) : nullptr,
                     DnnlExtensionUtils::makeDescriptor(descOut),
                     *attr,
                     selected_pd->getImplementationType()
```
Should also be a part of the routine. So something like that:

```cpp
   std::function<key(...)> init_key;
   
   /*on the prepareParams*/
   if (rank==6) {
     init_key = [](...) {
       SimplifyAs2D //and other code we need
       return MatMulKey{...}
     } else {
        init_key = [](...) {
           //the old init code
           return MatMulKey{...} }
         }
```

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -527,6 +613,19 @@ MemoryDescPtr MatMul::getSrcMemDesc(dnnl::primitive_desc_iterator &primitive_des
...
+    auto desc = primitive_desc_it.dst_desc(idx);
+    return std::make_shared<CpuBlockedMemoryDesc>(
+                        DnnlExtensionUtils::DataTypeToIEPrecision(desc.get_data_type()),
+                        getOutputShapeAtPort(idx)); /* provide initial shapes, so hide transpose effect */
+/*
+    if (getOutputShapeAtPort(idx).isDynamic()) {
+        return DnnlExtensionUtils::makeUndefinedDesc(desc, getOutputShapeAtPort(idx));
+    }
+    return DnnlExtensionUtils::makeDescriptor(desc);
+*/
```
maxnick, line 626: commented code

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -527,6 +613,19 @@ MemoryDescPtr MatMul::getSrcMemDesc(dnnl::primitive_desc_iterator &primitive_des
         return DnnlExtensionUtils::makeDescriptor(desc);
 }
 
+MemoryDescPtr MatMul::getDstMemDesc(dnnl::primitive_desc_iterator &primitive_desc_it, size_t idx) {
+    auto desc = primitive_desc_it.dst_desc(idx);
+    return std::make_shared<CpuBlockedMemoryDesc>(
+                        DnnlExtensionUtils::DataTypeToIEPrecision(desc.get_data_type()),
```
maxnick, line 619: Data type may be retrieved from the port config of the selected node descriptor. (just faster, since querying dnnl::primitive_desc is a very slow operation).

maxnick (changes requested):
Also we have to extend our functional tests to cover this case.

dmitry-gorokhov:
@usstq  @maxnick Just a general thought on the issue. Maybe to avoid unnecessary complication of the plugin code we can just avoid MatMul + Binary post-ops fusion for such particular scenario and wait fix from oneDNN? I understand that might not provide the best possible perf, but still much better that fallback on ref MatMul.

sammysun0711:
> @usstq @maxnick Just a general thought on the issue. Maybe to avoid unnecessary complication of the plugin code we can just avoid MatMul + Binary post-ops fusion for such particular scenario and wait fix from oneDNN? I understand that might not provide the best possible perf, but still much better that fallback on ref MatMul.

@dmitry-gorokhov Thanks for your feedback. As you know, twins as one of popular vision transformer, is widely adopted by many customers and developers, therefore we need BF16-INT8 mixed precision optimization by CPU plugin to be more competitive. 
I know current oneDNN 3.1 has some limitation, now we have following 3 options.
1. Collapse batch dimensions for matmul w/ 6D input in special case as workaround
2. Avoid MatMul + Binary post-ops fusion, waiting for oneDNN fix
3. Fallback on ref Matmul

First of all, option 3 will lead to very low performance, so it is not acceptable. 
Option 2 is worth considering if performance is at least on par with OV 22.3. Could you please kindly share the performance comparison with OV 22.3? 
I think Option 3 will lead to best perf without much overhead in this special case. We can merge this WA first, while still can replace this with more general solution when fix from oneDNN is available. 

What do you think about 3 options above?

usstq:
@sammysun0711 I have added another simpler WA as @dmitry-gorokhov suggested, with an environment variable `MM6D_FUSE_BIN` controlling which WA is used, `0` means do not fuse per-channel Eltwise node into 6D matmul (as @dmitry-gorokhov suggested), `1` means fuse all possible Eltwise, but collapse batch dimensions(the original WA), I tested it in my local machine and observed 5~10% performance drop when `MM6D_FUSE_BIN=0` comparing to `MM6D_FUSE_BIN=1`,  can you test them on your machine to see if it's acceptable ?

usstq:
@maxnick  @dmitry-gorokhov  I've changed the WA as suggested, can you please review again, Thanks!

usstq:
@maxnick  Since performance of `MM6D_FUSE_BIN=0` is acceptable, I updated the PR with Dimitry's WA, please review again, Thanks!

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -204,6 +205,19 @@ MatMul::MatMul(const std::shared_ptr<ngraph::Node>& op, const GraphContext::CPtr
 }
 
 bool MatMul::canFuse(const NodePtr& node) const {
+    // WA for CVS-84056: oneDNN brgemm impl has problem with per-OC binary-postOps for MatMul with 6D inputs
+    if (impl::cpu::x64::mayiuse(impl::cpu::x64::avx512_core)) {
```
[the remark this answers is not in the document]
sammysun0711: @maxnick, I observed same issue on ICX with AVX512, I update info in the tickets.

maxnick (approved):

maxnick:
@usstq , shouldn't we extend our SL tests to check the selected implementation for the 6D case?

src/plugins/intel_cpu/src/nodes/matmul.cpp
```diff
@@ -204,6 +205,19 @@ MatMul::MatMul(const std::shared_ptr<ngraph::Node>& op, const GraphContext::CPtr
...
+        if (auto* eltwiseNode = dynamic_cast<Eltwise*>(node.get())) {
+            if (eltwiseNode->getBroadcastingPolicy() != Eltwise::BroadcastingPolicy::PerTensor) {
+                auto rank = getInputShapeAtPort(0).getRank();
+                if (rank == 6) {
```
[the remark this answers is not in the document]
dmitry-gorokhov: I think we should. However lets do it in separate PR.

dmitry-gorokhov (approved):

dmitry-gorokhov (approved):

PR #9929, opened 2022-01-26
Fix bug in Reduce node concerning model accuracy

xuchen-intel:
### Details:
 - *Fix bug in Reduce node concerning model accuracy*

### Tickets:
 - *76735*

PR #16829, opened 2023-04-10
[CPU] Reduce node asymmetrical precision optimization

xuchen-intel:
### Details:
 - *Generalize to support asymmetrical precision optimization for reducing HW of nspc layout in Reduce node.*

### Tickets:
 - *106653*


src/plugins/intel_cpu/src/nodes/reduce.cpp
```diff
@@ -1983,16 +1984,26 @@ void Reduce::createPrimitive() {
         updateLastInputDims();
     }
 
+    create_reduce_kernel(reduce_kernel, jcp);
+    if (use_aux_kernel) {
+        aux_jcp = jcp;
+        aux_jcp.src_dt = jcp.dst_dt;
```
nshchego, line 1990: That is not clear how the data loading should work here. Please add appropriate test cases.

xuchen-intel: Thanks Nikolay for the comment! In fact test cases (e.g. shape=[2, 19, 2, 9], axes=[2,3], type=Mean, precision=bf16, layout=nhwc) have already been covered in test suite smoke_Reduce_MultiAxis_4D_CPU. Anyway I add some description about how this auxiliary kernel work. Thanks!

xuchen-intel: Thanks Nikolay for the comment! Now I see your concerns. In fact, the code block here https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/nodes/reduce.cpp#L1810-L1819 assures bf16-in-fp32-out will be applied even if precision in SLT is bf16-in-bf16-out. And that's also the reason that the model in the ticket apply bf16-in-fp32-out precision in the first place. And if you are interested about the reason why SLT don't cover different input/output precisions directly, please also refer to the discussion here https://github.com/openvinotoolkit/openvino/pull/9929#issuecomment-1025177761. Further discussions are always welcomed. Thanks!

xuchen-intel: @nshchego According to offline discussion, I've constrained precision restriction to be bf16-in-fp32-out for the usage of auxiliary kernel.

nshchego (approved):

xuchen-intel:
@dmitry-gorokhov Hi Dmitry, could you please also take a look?

dmitry-gorokhov (approved):

PR #16833, opened 2023-04-10
Remove vopset typo

ilyachur:
### Details:
 - Remove incorrect opset version if print the operation:
 `vopset1::...` -> `opset1::...`

### Tickets:
 - *ticket-id*


ilya-lavrenov (approved):

ilyachur:
in 2022.3 release we don't have this error:
https://github.com/openvinotoolkit/openvino/blob/e7cfb8e363a99679fad3be3803c315ce0c465923/src/core/src/node.cpp#L386-L388
because we use `v` prefix for deprecated version

PR #16847, opened 2023-04-10
[GPU] Fix for softmax unaligned leftovers

ytorzuk-altran:
### Details:
 - *https://github.com/openvinotoolkit/openvino/pull/15863 doesn't handle use case where leftovers introduce unaligned memory offset*
 - *This PR aims to disable blocking reads and writes for cases where leftovers cause unaligned memory offset*

### Tickets:
 - *[CVS-91171](https://jira.devtools.intel.com/browse/CVS-91171)*

### DIFF
--- a/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/softmax_gpu_bf.cl
+++ b/src/plugins/intel_gpu/src/kernel_selector/cl_kernels/softmax_gpu_bf.cl
@@ -61,7 +61,7 @@ KERNEL (softmax_gpu_continuous_bfyx)(
     __local INPUT0_TYPE lg_storage[SLM_SIZE];
 
     uint i=0;
-    if (workers_per_data_set > SUB_GROUP_SIZE)
+    if (workers_per_data_set >= SUB_GROUP_SIZE && (leftovers % (16 / sizeof(INPUT0_TYPE))) == 0)
     {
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/kernel_selector/cl_kernels/softmax_gpu_bf.cl
```diff
@@ -61,7 +61,7 @@ KERNEL (softmax_gpu_continuous_bfyx)(
     __local INPUT0_TYPE lg_storage[SLM_SIZE];
 
     uint i=0;
-    if (workers_per_data_set > SUB_GROUP_SIZE)
+    if (workers_per_data_set >= SUB_GROUP_SIZE && (leftovers % (16 / sizeof(INPUT0_TYPE))) == 0)
```
yeonbok, line 64: 1) Having additional calculation in cl kernel will not be good for performance. 
 We'd better have deterministic 
2) I don't see any reason to have a new PR from my PR. Would like to have a review comment to my PR, if you have any opinion,.. if the idea is completely different.

ytorzuk-altran: Most of these calculations should be optimized out by compiler. In static shape mode the whole condition would be optimized out since all variables are either constants or macros. Stable diffusion Softmax doesn't demonstrate any degradation.

PR #16850, opened 2023-04-11
[PyOV] Fix a bug with regex in requirements/constraints parser

p-wysocki:
### Details:
 - Fix a bug noticed by @meiyang-intel at https://github.com/openvinotoolkit/openvino/pull/16808#discussion_r1162372111

### Tickets:
 - NA

### DIFF
--- a/tools/benchmark_tool/setup.py
+++ b/tools/benchmark_tool/setup.py
@@ -67,7 +67,7 @@ def read_requirements(path: str) -> List[str]:
         # get rid of newlines
         line = line.replace('\n', '')
         # if version is specified (non-word chars present)
-        if re.search('\W', line):
+        if re.search('(~|=|<|>|;)', line):
             requirements.append(line)
         # else get version from constraints
         else:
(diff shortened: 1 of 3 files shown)


tools/benchmark_tool/setup.py
```diff
@@ -67,7 +67,7 @@ def read_requirements(path: str) -> List[str]:
         # get rid of newlines
         line = line.replace('\n', '')
         # if version is specified (non-word chars present)
-        if re.search('\W', line):
+        if re.search('(~|=|<|>|;)', line):
             requirements.append(line)
```
rkazants, line 71: Can you please add tests for this parser?

meiyang-intel (approved):

tools/openvino_dev/setup.py
```diff
@@ -252,7 +252,7 @@ def read_requirements(path: str) -> List[str]:
         # get rid of newlines
         line = line.replace('\n', '')
         # if version is specified (non-word chars present)
-        if re.search('\W', line):
+        if re.search('(~|=|<|>|;)', line):
```
rkazants, line 255: can you please avoid duplication of the same parser in different installation scripts? You can do it in separate PR.

PR #16861, opened 2023-04-11
[Transformations] BroadcastTransition transformation

v-Golubev:
### Details:
 - *BroadcastTransition transformation moves Broadcast after elementwise operation. This might allow to restore MatMul + Elemetwise fusing optimization and speedup Broadcast processing itself*

### Tickets:
 - *CVS-107667*


src/common/transformations/src/transformations/common_optimizations/broadcast_transition.cpp
```diff
@@ -0,0 +1,78 @@
...
+ov::pass::BroadcastTransition::BroadcastTransition() {
+    MATCHER_SCOPE(BroadcastTransition);
+    auto bcast_data_m = pass::pattern::any_input();
+    auto target_shape_m = pass::pattern::any_input(pass::pattern::has_static_shape());
+    auto bcast_m = pass::pattern::wrap_type<opset1::Broadcast, opset10::Broadcast>({bcast_data_m, target_shape_m},
+                                                                                   pass::pattern::consumers_count(1));
+    auto eltwise_input_m = pass::pattern::any_input(pass::pattern::has_static_rank());
+    auto eltwise_1 = pass::pattern::wrap_type<op::util::BinaryElementwiseArithmetic>({eltwise_input_m, bcast_m});
+    auto eltwise_2 = pass::pattern::wrap_type<op::util::BinaryElementwiseArithmetic>({bcast_m, eltwise_input_m});
+    auto eltwise_m = std::make_shared<pass::pattern::op::Or>(OutputVector{eltwise_1, eltwise_2});
```
jane-intel, line 26: Our pattern matcher knows which operations are commutative (l[ink](https://github.com/openvinotoolkit/openvino/blob/master/src/core/src/pattern/matcher.cpp#L132)). However, here you also can work with non-commutative operations, like Subtract and Divide. So, this is just an observation.

src/common/transformations/src/transformations/common_optimizations/broadcast_transition.cpp
```diff
@@ -0,0 +1,78 @@
...
+            return false;
+        }
+
+        const auto& eltwise_input = pattern_map.at(eltwise_input_m);
+        const auto& bcast_data = pattern_map.at(bcast_data_m);
+        const auto new_eltwise = eltwise->clone_with_new_inputs({eltwise_input, bcast_data});
+        ov::copy_runtime_info(eltwise, new_eltwise);
+
+        auto target_shape = pattern_map.at(target_shape_m);
+        const auto& target_shape_et = target_shape.get_element_type();
```
jane-intel, line 47: Theoretically Broadcast shape et may be u32, u64 or else and ShapeOf accepts only i32 and i64 variants. However, this is quite rare -- I would say I've never seen it. As an option here we can check if et is within [i32, i64] -- then use ShapeOf's output type attribute. Otherwise, make ShapeOf i64 and insert Convert to the funky type after it.

src/common/transformations/src/transformations/common_optimizations/broadcast_transition.cpp
```diff
@@ -0,0 +1,78 @@
...
+        auto data_shape_path = ov::op::util::make_try_fold<opset10::ShapeOf>(new_eltwise, target_shape_et);
+        ov::copy_runtime_info(eltwise, data_shape_path);
+
+        const size_t target_shape_rank = target_shape.get_partial_shape()[0].get_length();
```
jane-intel, line 51: Is there a way scalar tensors would participate in this pattern?
This code would fail then.

jane-intel (commented):
I like the idea, but I have a minor ask.

src/common/transformations/src/transformations/common_optimizations/broadcast_transition.cpp
v-Golubev: Good point, thanks. I used `Or` exactly to cover non-commutative operations in the pattern. As for me, this is the shortest way to handle both cases.

src/common/transformations/src/transformations/common_optimizations/broadcast_transition.cpp
v-Golubev: I corrected the code per your suggestion and added a test case. Thank you

src/common/transformations/src/transformations/common_optimizations/broadcast_transition.cpp
v-Golubev: Scalar target shape is forbidden for broadcast node. I added a test case to type_prop tests to cover this situation

v-Golubev:
@dmitry-gorokhov could you please take a look? Thanks

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/broadcast_eltwise.cpp
```diff
@@ -0,0 +1,108 @@
...
+};
+
+TEST_P(BroadcastEltwise, smoke_CompareWithRefs) {
+    run();
+
+    const auto model = compiledModel.get_runtime_model();
+    const auto last_node = model->get_result()->get_input_node_shared_ptr(0);
+    const auto& rt_info = last_node->get_rt_info();
+    const auto layerType = rt_info.find("layerType")->second.as<std::string>();
+    EXPECT_EQ(layerType, "Broadcast");
```
dmitry-gorokhov, line 87: Maybe it also worth to check that Broadcast op has equal shapes on input and output. This is kind of mandatory condition for the scenario

src/plugins/intel_cpu/src/nodes/common/tile_broadcast_utils.cpp
```diff
@@ -250,7 +251,10 @@ void TileBroadcastCommon::optimizedExecute(const MemoryPtr& srcMemory, const Mem
     auto srcData = reinterpret_cast<const char *>(srcMemory->GetPtr());
     auto dstData = reinterpret_cast<char *>(dstMemory->GetPtr());
 
-    if (optimizedParams.srcStrides[5] == 0) {
+    if (srcMemory->getStaticDims() == dstMemory->getStaticDims()) {
+        const auto prc = dstMemory->getDesc().getPrecision();
+        cpu_convert(srcData, dstData, prc, prc, optimizedParams.copySize / prc.size());
```
dmitry-gorokhov, line 256: We need to put the comment why cpu_convert is used and not cpu_memcpy. I would also suggest to add TODO: label that will describes our desire to move fast copy logic inside the cpu_memcpu and reuse it everywhere

dmitry-gorokhov (approved):
Minor comments. Can be fixed in separate PR

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/broadcast_eltwise.cpp
v-Golubev: Done in #17168

PR #16863, opened 2023-04-11
[GPU] Fix shape_of shape inference optimization

sshlyapn:
### Details:
 - Move shape_of optimization a little bit higher than shape infer dependencies checking (which prevents it's optimization, because shape_of's shape infer dependencies are not constant) to allow shape_of optimization

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -179,6 +179,10 @@ void primitive_inst::update_shape() {
     if (input_shape_changed)
         set_shape_change();
 
+    // We assume that tensor ranks are static, thus shape_of doesn't need to update anything even if input shape is dynamic
+    if (_node->is_type<shape_of>() && !input_shape_changed)
+        return;
+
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -179,6 +179,10 @@ void primitive_inst::update_shape() {
     if (input_shape_changed)
         set_shape_change();
 
+    // We assume that tensor ranks are static, thus shape_of doesn't need to update anything even if input shape is dynamic
+    if (_node->is_type<shape_of>() && !input_shape_changed)
```
vladimir-paramuzov, line 183: Hmm, wouldn't overloading of `get_shape_infer_dependencies` in shape_of node (which will return empty vector) help to resolve this issue?

yeonbok (approved):

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
sshlyapn: Yeah, it will fix. But wouldn't that be misleading? Since actually shape_of may depend on input in case of dynamic rank (which is not supported, but still...)

PR #16875, opened 2023-04-12
[GPU] Handle unsupported eltwise fusion for onednn gemm in dynamic cases

vladimir-paramuzov:
### Details:
 - Onednn has limited broadcast support for fused eltwise input for gemm primitive, so in static case we insert explicit `broadcast` primitive for that input port. But for dynamic case we can't use same WA as shapes are unknown and we don't know if broadcast is needed or not. Thus we skip WA in reorder_inputs pass for dynamic cases and do similar check in `primitive_inst::is_valid_fusion()` to have unfused execution in case if unsupported broadcast pattern occur
 - Fixes functional issue for https://github.com/openvinotoolkit/openvino_notebooks/tree/main/notebooks/233-blip-visual-language-processing

### DIFF
--- a/src/plugins/intel_gpu/tests/fusions/gemm_fusion_test.cpp
+++ b/src/plugins/intel_gpu/tests/fusions/gemm_fusion_test.cpp
@@ -20,10 +20,8 @@ using namespace ::tests;
 
 namespace {
 struct gemm_test_params {
-    std::vector<tensor> in_shapes;
-    tensor out_shape;
-    tensor kernel;
-    tensor pad;
+    std::vector<ov::PartialShape> in_shapes;
+    ov::PartialShape out_shape;
     data_types data_type_in0;
     data_types data_type_in1;
     data_types data_type_in2;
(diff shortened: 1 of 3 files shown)


src/plugins/intel_gpu/tests/fusions/gemm_fusion_test.cpp
```diff
@@ -349,6 +352,38 @@ TEST_P(gemm_2in_add, eltwise_postop_cached) {
...
+        input_layout("input0", in_layout0),
+        input_layout("input1", in_layout1),
+        data("add_data", get_mem(add_data_layout, 0.5f)),
+        gemm("gemm_prim", { input_info("input0"), input_info("input1") }, data_types::f32, false, false, 1.f, 0.f, in_layout0.get_rank(), in_layout1.get_rank()),
+        eltwise("add_prim", { input_info("gemm_prim"), input_info("add_data") }, p.eltwise_m, p.default_type),
+        reorder("reorder_bfyx", input_info("add_prim"), p.default_format, data_types::f32)
+    );
+
+    tolerance = default_tolerance(p.default_type);
+    execute(p, false);
```
sshlyapn, line 384: It looks like this test should be cached 
```suggestion
    execute(p, false, true);
```

yeonbok (approved):

PR #16887, opened 2023-04-12
[CPU] Correct crop in FQ optimized formula

EgorDuplensky:
### Tickets:
 - 99838
 
 ### Details:
 - Output scale can be negative, so crop high and crop low can be swapped when multiplied.


usstq (approved):
Do I understand correctly that output scale can be negative in some cases, in turn it causes clip postOps failure due to (low > high) ?

Anyway the more robust the better!

EgorDuplensky:
> Do I understand correctly that output scale can be negative in some cases, in turn it causes clip postOps failure due to (low > high) ?
> 
> Anyway the more robust the better!

Exactly so!
I will add a context to the PR message

EgorDuplensky:
@dmitry-gorokhov Could you please take a look?

dmitry-gorokhov (approved):
@EgorDuplensky Please put ticket number related to test coverage here.

PR #16888, opened 2023-04-12
[port from master][ONNX FE] Fix Squeeze v1

mbencer:
Ticket: 105515
Port of https://github.com/openvinotoolkit/openvino/pull/16865


mbencer:
Correct version in https://github.com/openvinotoolkit/openvino/pull/16890

PR #16894, opened 2023-04-12
[PT FE] Support prim::DictConstruct on the output

mvafin:
### Details:
 - *Support `prim::DictConstruct` on the output*

### Tickets:
 - *TBD*

### DIFF
--- a/src/frontends/pytorch/src/transforms/dict_resolver.cpp
+++ b/src/frontends/pytorch/src/transforms/dict_resolver.cpp
@@ -28,7 +28,7 @@ bool DictResolver::run_on_model(const std::shared_ptr<Model>& model) {
                 // inputs must be divisible by 2
                 return false;
             }
-            OutputVector new_outputs;
+            ResultVector new_outputs;
             for (size_t i = 0; i < inputs.size(); i += 2) {
                 auto new_output = inputs.at(i + 1);
                 const auto name_node = inputs.at(i);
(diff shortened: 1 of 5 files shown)


eaidova (approved):

src/frontends/pytorch/src/transforms/dict_resolver.cpp
```diff
@@ -0,0 +1,63 @@
...
+                const auto attrs = fw_node->get_attrs();
+                if (attrs.find("string_value") == attrs.end()) {
+                    // fw node must contain string value
+                    return false;
+                }
+                auto name = attrs.at("string_value");
+                new_output.add_names({name});
+                new_outputs.push_back(new_output);
+            }
+            // TODO: this may break the order of outputs if after dict more outputs exist
```
slyalin, line 49: What's the problem to provide fully functional implementation already now? I realize that when we 'unpacking' `dict` in such a way, this is already something that change the output signature, but looks not to too hard to maintain the remining part of outputs in the same order just to avoid changes without a good reason. As I remember we don't have a convenient way like ov::Model method, and in my experiments when I did such thing I added another method to ov::Model, but we can provide a helper function to make it without changing ov::Model API.

@itikhono, @ilyachur just FYI, a potential gap in ov::Model, or I'm just missing something.

slyalin (changes requested):
Do not change an order of outputs.

src/frontends/pytorch/src/transforms/dict_resolver.cpp
mvafin: There is no `set_results` method for a `Model`, the only way to do it with current api is `remove_result` one by one and do `add_results` after.

PR #16901, opened 2023-04-12
[PT FE] Add aten::all implementation

PiotrKrzem:
### Details:
 - Add aten::all implemetation

### Tickets:
 - 101134


src/frontends/pytorch/src/op/all.cpp
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    bool keep_dims;
+    ov::Output<ov::Node> axes;
+    if (context.get_input_size() == 1) {
+        keep_dims = false;
+        axes = get_axes_range(context, 0);
+    } else {
+        const auto dim = context.const_input<int64_t>(1);
+        keep_dims = !context.input_is_none(2) ? context.const_input<bool>(2) : false;
```
tomdol, line 29: You can initialize `keep_dims` to false in line 22, remove the assignment in line 25 and only update the value here if the third input is not none.

tomdol (approved):

mvafin (approved):

src/frontends/pytorch/src/op/all.cpp
PiotrKrzem: Updated!

PR #16909, opened 2023-04-13
[DOCS] model caching update to GPU

kblaszczak-intel:
Update GPU.md
Update Model_caching_overview.md

### DIFF
--- a/docs/OV_Runtime_UG/supported_plugins/GPU.md
+++ b/docs/OV_Runtime_UG/supported_plugins/GPU.md
@@ -296,11 +296,11 @@ Model Caching helps reduce application startup delays by exporting and reusing
 the compiled model automatically. The cache for the GPU plugin may be enabled 
 via the common OpenVINO ``ov::cache_dir`` property. 
 
-GPU plugin implementation fully supports static models only. For dynamic models,
-kernel caching is used instead and multiple ‘.cl_cache’ files are generated along with the ‘.blob’ file. 
 This means that all plugin-specific model transformations are executed on each ``ov::Core::compile_model()`` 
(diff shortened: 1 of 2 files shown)


docs/OV_Runtime_UG/supported_plugins/GPU.md
```diff
@@ -292,24 +292,27 @@ For more details, see the :doc:`preprocessing API<openvino_docs_OV_UG_Preprocess
...
+via the common OpenVINO ``ov::cache_dir`` property. 
 
-.. note::
-
-   Full model caching support is currently implemented as a preview feature. To activate it, set the OV_GPU_CACHE_MODEL environment variable to 1.
+GPU plugin implementation fully supports static models only. For dynamic models,
+kernel caching is used instead and multiple ‘.cl_cache’ files are generated along with the ‘.blob’ file. 
+This means that all plugin-specific model transformations are executed on each ``ov::Core::compile_model()`` 
+call, regardless of the ``cache_dir`` option. Still, since kernel compilation is a bottleneck in the model 
+loading process, a significant load time reduction can be achieved.
```
e-ddykim, line 303: I think it's more natural for the sentence that begins with `This means that~` to continue with the preceding sentence as below. I also added a namespace `ov` in front of the `cache_dir` for consistency.

```suggestion
This means that all plugin-specific model transformations are executed on each ``ov::Core::compile_model()`` 
call, regardless of the ``ov::cache_dir`` option. Still, since kernel compilation is a bottleneck in the model 
loading process, a significant load time reduction can be achieved.
Currently, GPU plugin implementation fully supports static models only. For dynamic models,
kernel caching is used instead and multiple ‘.cl_cache’ files are generated along with the ‘.blob’ file. 
```

e-ddykim (approved):
Looks good to me!

PR #16911, opened 2023-04-13
[GPU] Fix reshape split for dynamic models

vladimir-paramuzov:
### Details:
 - Propagate all reshape parameters for reshape split pass
 - Fixed dynamic tensor offset for reduce primitive with fused ops
 - Fixes crash & accuracy on SAM predictor
 - Minor improvements to graph dumper:
   - show primitive type
   - print short format of layout when pad is empty

### DIFF
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -846,7 +846,7 @@ memory::ptr primitive_inst::allocate_output(engine& _engine, memory_pool& pool, 
             allocation_type type, bool reusable, bool reset = true) {
         OPENVINO_ASSERT(!layout.is_dynamic() || layout.has_upper_bound(), "[GPU] Can't allocate output for dynamic layout without upper bound");
         // Use layout with max tensor for dynamic shape with upper bound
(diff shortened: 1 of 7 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -846,7 +846,7 @@ memory::ptr primitive_inst::allocate_output(engine& _engine, memory_pool& pool,
             allocation_type type, bool reusable, bool reset = true) {
         OPENVINO_ASSERT(!layout.is_dynamic() || layout.has_upper_bound(), "[GPU] Can't allocate output for dynamic layout without upper bound");
         // Use layout with max tensor for dynamic shape with upper bound
-        auto static_layout = cldnn::layout(layout.data_type, layout.format, layout.get_tensor(), layout.data_padding);
+        auto static_layout = cldnn::layout(layout.get_partial_shape(), layout.data_type, layout.format, layout.data_padding);
```
yeonbok, line 849: Won't this change memory alloc behavior of bounded dynamic shape? 
I made get_tensor to return max tensor but get_partial_shape will still be dynamic shape. We may need get_max_shape handle..

vladimir-paramuzov: added `get_max_shape()`

yeonbok (approved):

PR #15216, opened 2023-01-19
Docs: Update Dynamic Shapes documentation

EdjeElectronics:
This PR rewrites the [Dynamic Shapes](https://docs.openvino.ai/latest/openvino_docs_OV_UG_DynamicShapes.html) page to address issues identified by the Developer Experiences team.

### Details:
- Modify examples that previously showed multiple ways to do the same thing (with `# same as above` shown multiple times in each example), which can be confusing to new users. Examples are simplified to just show one common preferred way of setting dynamic shapes.
- Add example showing how to apply dynamic shapes to models with multiple input layers
- Add example showing how to check if model already has dynamic inputs "out of the box"
- Generally rewrite content to make it more understandable

Note: The newly added examples are only written in Python and C++, I didn't have time to figure out how to do them in C 😺.

### Tickets:
N/A

cc @helena-intel @ryanloney

PR #14793, opened 2022-12-22
Docs: Edits to Performance Hints and Cumulative Throughput documentation

EdjeElectronics:
This PR introduces minor changes to the [Performance Hints](https://docs.openvino.ai/latest/openvino_docs_OV_UG_Performance_Hints.html) and [Automatic Device Selection](https://docs.openvino.ai/latest/openvino_docs_OV_UG_supported_plugins_AUTO.html) pages to:

1. Explain how to use CUMULATIVE_THROUGHPUT with multiple GPUs
2. Make it clear that throughput Performance Hints only affect asynchronous pipelines

### Details:
 - See above

### Tickets:
 - NA

PR #12067, opened 2022-07-07
Docs: Update Devices pages to state improved INT8 performance with 11th & 12th gen devices

EdjeElectronics:
### Details:
 - Small edits to Devices pages to indicate that 11th and 12th generation processors have improved performance with INT8 models. This is a specific request from the OpenVINO product team.

### Tickets:
 - N/A

PR #16913, opened 2023-04-13
[DOCS] Recreation of BDTI PRs - master

msmykx-intel:
Recreation of BDTI PRs for master.

Recreated PRs:

* https://github.com/openvinotoolkit/openvino/pull/15216
* https://github.com/openvinotoolkit/openvino/pull/14793
* https://github.com/openvinotoolkit/openvino/pull/12067


docs/OV_Runtime_UG/supported_plugins/Supported_Devices.md
```diff
@@ -66,17 +66,17 @@ This page shows supported and optimal configurations for each plugin.
...
- FP16 format          Half-precision floating-point format         
- I16 format           2-byte signed integer format                 
- I8 format            1-byte signed integer format                 
- U16 format           2-byte unsigned integer format               
- U8 format            1-byte unsigned integer format               
-===================  =============================================
++------------------+----------------------------------------------+
+| Acronym/Term     |   Description                                |
++==================+==============================================+
+| FP32 format      |   Single-precision floating-point format     |
```
helena-intel, line 72: We support double precision floating point (FP64) and integer precision too. @ilya-lavrenov where should these be added in this document (input/output/model precision)? For integer, should we list both I64 and I64? Do we miss other formats?

@msmykx-intel these tables are tricky to review! Are there rendered previews for these pages?

docs/OV_Runtime_UG/supported_plugins/Supported_Devices.md
```diff
@@ -98,14 +98,14 @@ For example, the CHW value at index (c,h,w) is physically located at index (c\*H
...
+| Plugin           | FP32                     | FP16                     | I8                     |
++==================+==========================+===================================================+ 
+| CPU plugin       | Supported and preferred  | Supported                | Supported              |
+| GPU plugin       | Supported                | Supported and preferred  | Supported              |
+| GNA plugin       | Supported                | Supported                | Not supported          |
+| Arm® CPU plugin  | Supported and preferred  | Supported                | Supported (partially)  |
++------------------+--------------------------+--------------------------+------------------------+
 
 
 For :doc:`Multi-Device <openvino_docs_OV_UG_Running_on_multiple_devices>` and
```
helena-intel, line 111: Not changed in this PR, but can we delete "*Generally, FP16 is preferable as it is most ubiquitous and performant*.". It is not true. It was never "most ubiquitous" but it is also no longer most performant. Model format is now independent of inference precision; on GPU (where FP16 is most performant) inference will be in FP16 by default regardless of model format. FP16 for now has larger memory requirements than FP32. So the only benefit of FP16 is smaller file size, which at the moment comes at a larger memory cost. IMO It's better to not have a blanket recommendation for now. Memory downside of FP16 will be resolved in the next release.

helena-intel (commented):
Thanks @msmykx-intel ! I am glad you are picking this up. I have a few comments for now, will review more later.

docs/OV_Runtime_UG/supported_plugins/Supported_Devices.md
msmykx-intel: Done in https://github.com/openvinotoolkit/openvino/pull/16913/commits/77a9364a34d8b658d98fab05ab314a2c54c12c54

yeonbok:
Only checked ov_dynamic_shapes.md and it looks good.

helena-intel (approved):
Thanks again @msmykx-intel ! Really glad to see these improvements. I have a few suggestions, not blocking. When reviewing I noticed some other things that should be updated (for example that we support more data formats now) but we can address that later.

docs/OV_Runtime_UG/auto_device_selection.md
```diff
@@ -230,6 +240,10 @@ Performance Hints for AUTO
 
 The ``ov::hint::performance_mode`` property enables you to specify a performance option for AUTO to be more efficient for particular use cases. The default hint for AUTO is ``LATENCY``.
 
+The THROUGHPUT and CUMULATIVE_THROUGHPUT hints below only improve performance in an asynchronous inference pipeline. For information on asynchronous inference, see the `Async API documentation <https://docs.openvino.ai/latest/openvino_docs_OV_UG_Infer_request.html#doxid-openvino-docs-o-v-u-g-infer-request>`__ . The following notebooks provide examples of how to set up an asynchronous pipeline:
+
+* :doc:`Image Classification Async Sample <openvino_inference_engine_samples_classification_sample_async_README>`
+* `Notebook - Automatic Device Selection with OpenVINO <https://docs.openvino.ai/latest/notebooks/106-auto-device-with-output.html>`__
```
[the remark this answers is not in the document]
msmykx-intel: Added in https://github.com/openvinotoolkit/openvino/pull/16913/commits/0abbbc401225c583ce63bfb3865cd099c3025e2a

PR #16926, opened 2023-04-13
[GNA] Fix for depthwise convolution network accuracy issue

marcinkacprzak:
### Details:
 - Implemented insertion of Copy layer in case of FakeQuantize found before Concat

### Tickets:
 - 99349

### DIFF
--- a/src/plugins/intel_gna/src/backend/gna_limitations.cpp
+++ b/src/plugins/intel_gna/src/backend/gna_limitations.cpp
@@ -774,7 +774,7 @@ static bool ValidateConcatAxis(const InferenceEngine::CNNLayerPtr layer, std::st
                         0,
                         [](InferenceEngine::CNNLayerPtr ptr) {
                             return LayerInfo(ptr).isNonFunctional() || LayerInfo(ptr).isFakeQuantize() ||
-                                   LayerInfo(ptr).isSplit();
+                                   LayerInfo(ptr).isSplit() || LayerInfo(ptr).isPermute();
                         });
 
                     IE_ASSERT(pre_prev_layer);


src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
```diff
@@ -199,7 +199,8 @@ InsertCopyBeforeConcatLayer::InsertCopyBeforeConcatLayer() {
             // Crop -> Concat, Input -> Split -> Concat
             if ((std::dynamic_pointer_cast<ngraph::op::CropIE>(current_node) && !is_crop_affined(current_node)) ||
                 std::dynamic_pointer_cast<ngraph::opset8::Split>(current_node) ||
-                std::dynamic_pointer_cast<ngraph::opset8::VariadicSplit>(current_node)) {
+                std::dynamic_pointer_cast<ngraph::opset8::VariadicSplit>(current_node) ||
+                std::dynamic_pointer_cast<ngraph::opset9::FakeQuantize>(current_node)) {
```
sirzabek, line 203: Are these the newest opsets available?

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
```diff
@@ -1,21 +1,25 @@
-// Copyright (C) 2022 Intel Corporation
+// Copyright (C) 2023 Intel Corporation
```
sirzabek, line 1:
```suggestion
// Copyright (C) 2022-2023 Intel Corporation
```

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
```diff
@@ -1,21 +1,25 @@
...
 // SPDX-License-Identifier: Apache-2.0
 //
 
 #include <gtest/gtest.h>
 
+#include <functional_test_utils/include/functional_test_utils/precision_utils.hpp>
```
marcinkusm, line 7: cannot we use local include here?

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
```diff
@@ -1,21 +1,25 @@
...
 
 #include "common_test_utils/ngraph_test_utils.hpp"
 #include "ngraph_functions/builders.hpp"
+#include "ngraph_functions/subgraph_builders.hpp"
+#include "openvino/opsets/opset11.hpp"
 #include "ops/copy.hpp"
 #include "transformations/insert_copy_layer.hpp"
 
+using namespace ov::opset11;
+
```
marcinkusm, line 22:
```suggestion
using namespace ov;
using namespace ov::element;
```

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
```diff
@@ -73,42 +77,38 @@ void InsertCopyLayerTest::Run() {
...
+            auto concat = std::make_shared<ov::opset11::Concat>(concat_inputs, m_axis);
+            auto result = std::make_shared<ov::opset11::Result>(concat);
+            m_func = std::make_shared<ov::Model>(ov::ResultVector{result}, ov::ParameterVector{params}, "Concat");
         }
 
         {
-            auto params = std::make_shared<ngraph::opset8::Parameter>(ngraph::element::i64, input_shape);
-            auto add = std::make_shared<ngraph::opset8::Add>(params, params);
+            auto params = std::make_shared<Parameter>(ov::element::i64, input_shape);
+            auto add = std::make_shared<ov::opset11::Add>(params, params);
```
marcinkusm, line 97: `<Add>`

marcinkusm, line 97:
```suggestion
            auto add = std::make_shared<:Add>(params, params);
```

and in other places

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
```diff
@@ -1519,35 +1486,90 @@ TEST(TransformationTests, InsertCopyLayerNFLvsFLSubgraphTestt) {
...
+    ASSERT_TRUE(result.first);
+}
+
+// [Parameter]              [Parameter]
+//     |                         |
+//  [Slice]                   [Slice]
+//     |           =>            |
+//   [FQ]                      [FQ]
+//     |                         |
+// [NonFunc] [Constant]      [NonFunc]
```
rjeziers, line 1532: Why NonFunc instead of Reshape?

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
```diff
@@ -1519,35 +1486,90 @@ TEST(TransformationTests, InsertCopyLayerNFLvsFLSubgraphTestt) {
...
+    std::vector<float> fqIMin{-2.12}, fqIMax{2.12}, fqOMin{-4.24}, fqOMax{4.24};
+    size_t levels = 65535;
+    size_t axis = 0;
+    auto prc = FuncTestUtils::PrecisionUtils::convertIE2nGraphPrc(InferenceEngine::Precision::FP16);
+
+    {
+        auto params = std::make_shared<Parameter>(ov::element::i64, input_shape);
+        auto slice = ngraph::builder::makeSlice(params, {0, 0}, sliced_shape_i64, {1, 1}, {}, prc);
+        auto fq = ngraph::builder::makeFakeQuantize(slice, prc, levels, {}, fqIMin, fqIMax, fqOMin, fqOMax);
+        auto reshape = ov::op::util::reshapeTo(fq->output(0), sliced_shape);
```
rjeziers, line 1553: Why do you create some operations with ngraph::builder, some with std::make_shared, and the reshape is created from ov::op::util? Would it make any functional difference if everything was done, for example, with std::make_shared? If this can be done, it would be more consistent.

src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
mryzhov, line 203: If we change the transformation patterns it is required to update description in the header. 
Please update the header too.

mryzhov (changes requested):
Could you also provide the root cause of the issue in the description. It is not clear what didn't work previously

src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
marcinkacprzak: Yes, changed to opset11.

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
```diff
@@ -73,42 +77,38 @@ void InsertCopyLayerTest::Run() {
...
     void SetUp() override {
-        ngraph::Shape input_shape{10};
+        ov::Shape input_shape{10};
         InsertCopyLayerTest::SetUp();
 
         {
-            auto params = std::make_shared<ngraph::opset8::Parameter>(ngraph::element::i64, input_shape);
-            auto add = std::make_shared<ngraph::opset8::Add>(params, params);
-            ngraph::OutputVector concat_inputs;
+            auto params = std::make_shared<Parameter>(ov::element::i64, input_shape);
```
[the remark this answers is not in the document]
mryzhov: it is not required here, the readably is pretty fine here and we don't change ov::element from version to version like opset

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
marcinkacprzak: It's not needed, actually.

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
marcinkacprzak: Reshape is one of non-functional layers.

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
marcinkacprzak: Ok, I will remove all ov:: from this file.

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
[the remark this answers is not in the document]
marcinkacprzak: @mryzhov, so you suggest to apply only "using namespace ov::opset11" to this file?

src/plugins/intel_gna/tests/unit/transformations/gna_insert_copy_layer.cpp
marcinkacprzak: ngraph helpers do not include appropriate (int this context) method for reshape creation, so the ov::op:util::reshapeTo is used very widely in the code.

PR #16933, opened 2023-04-13
[GPU] Fix nonzero issue in constant propagate

yeonbok:
### Details:
 - Fix gather_nonzero not to be marked as constant.
 - Even though count nonzero is to be turned into a constant, gather nonzero still cannot infer shape at the moment of propagate constant.

### Tickets:
 - 108008

### DIFF
--- a/src/plugins/intel_gpu/src/graph/program.cpp
+++ b/src/plugins/intel_gpu/src/graph/program.cpp
@@ -634,6 +634,10 @@ void program::mark_if_constant(program_node& node) {
         node.is_type<assign>() || node.is_type<read_value>()) {
         return;
     }
+    for (auto i = 0UL; i < node.get_output_layouts().size(); ++i) {
+        if (node.is_dynamic_output_layout(i))
+            return;
+    }
     node.constant = true;
     for (auto& dep : node.get_dependencies()) {
         if (!dep.first->is_constant()) {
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/graph/program.cpp
```diff
@@ -631,7 +632,7 @@ void program::post_optimize_graph(bool is_internal) {
 // mark if the node is constant assuming that all dependencies are marked properly
 void program::mark_if_constant(program_node& node) {
     if (node.get_dependencies().empty() || node.is_type<prior_box>() ||
-        node.is_type<assign>() || node.is_type<read_value>()) {
+        node.is_type<assign>() || node.is_type<read_value>() || node.is_type<gather_nonzero>()) {
         return;
```
e-ddykim, line 636: nit: What do you think to assign false explicitly as below?

```suggestion
        node.is_type<assign>() || node.is_type<read_value>() || node.is_type<gather_nonzero>()) {
        node.constant = false;
        return;
```

e-ddykim: Looks good to me

e-ddykim (approved):
Looks good to me

src/plugins/intel_gpu/src/graph/program.cpp
yeonbok: Because it is original behavior that only to mark "if constant" I think

andrew-k-park (approved):

PR #16936, opened 2023-04-14
[GPU] Fix group axis value for blocking desc

vladimir-paramuzov:
### Details:
 - Group dimension had old axis value which led to format mismatch for onednn path

### Tickets:
 - *108461*


isanghao (approved):
This PR resolved functional issues.

PR #16937, opened 2023-04-14
[MO][TF FE] Support delayed batch setting

rkazants:
**Details:** The delayed batch setting is required for models with dynamic input ranks. The dynamic input ranks can be deduced on the later stages of the model conversion after which it is possible to set the required batch size.

**Ticket:** 102642


pavel-esir (approved):
looks good to me

popovaan (approved):

mvafin (commented):
I think we can set batch when user didn't provide the layout only when this deduced dimension is 1 or dynamic, otherwise we need to fail and ask user to provide layout. Because it is possible that input has shape like [200,200,3] or in case of 3d image [200,200,200,3] where there is obviously no batch, but we will set it, while we shouldn't.

rkazants:
> I think we can set batch when user didn't provide the layout only when this deduced dimension is 1 or dynamic, otherwise we need to fail and ask user to provide layout. Because it is possible that input has shape like [200,200,3] or in case of 3d image [200,200,200,3] where there is obviously no batch, but we will set it, while we shouldn't.

Done

tools/mo/openvino/tools/mo/moc_frontend/layout_utils.py
```diff
@@ -0,0 +1,76 @@
...
+                return default_batch_dim, True
+            from openvino.runtime import Layout  # pylint: disable=no-name-in-module,import-error
+            layout_parsed = Layout(layout)
+            if layout_parsed.has_name(dimension_label):
+                return layout_parsed.get_index_by_name(dimension_label), False
+            else:
+                # if the layout is specified and the required dimension label is not found, the batch is unknown
+                return None, False
+
+    return default_batch_dim, True
```
mvafin, line 76: You made this function universal, as if it will work not only for batch, but  you assume that default dim is batch dim which is not true if someone call this function with 'C' or 'W' or something else.

mvafin (changes requested):
Looks good in general, small comment

tools/mo/openvino/tools/mo/moc_frontend/layout_utils.py
rkazants: Great catch!
Fixed.

mvafin (approved):

PR #16941, opened 2023-04-14
[CPU] Leftovers for Extension convert I64/U64 to I32.

nshchego:
### Details:
 - *Here are some leftovers for commit #16797*

### Tickets:
 - **


src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -262,12 +262,14 @@ void Graph::Replicate(const CNNNetwork &network) {
...
 
-        if (op->get_type_info() == ngraph::op::v0::Result::get_type_info_static()) {
+        if (op->get_type_info() == ov::op::v0::Result::get_type_info_static()) {
             const auto &input = op->input_value(0);
             const auto name = ov::op::util::get_ie_output_name(input);
 
             if (outputsInfo.count(name) != 0) {
                 outputNodesMap[name] = node;
+            } else if (!input.get_names().empty()) {
+                outputNodesMap[*input.get_names().begin()] = node;
```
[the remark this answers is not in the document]
nshchego: Current implementation supports only old API and requires to set legacy name `m_legacy_name` to tensor  via `set_ov_tensor_legacy_name()` in transformation. 
This fix supports a new approach as well.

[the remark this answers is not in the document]
v-Golubev: Agree with @nshchego: in new API, tensors might have several names and these names are stored separately from `m_legacy_name`. So it's good idea to support new API in CPU graph.

But I have a concern about implementation details. Maybe we should add input name only if it's contained in `outputsInfo`? And also check all names, not only first one (until we find the name which is contained in `outputsInfo` or check all names).

v-Golubev (approved):

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
```diff
@@ -254,7 +254,10 @@ void Transformations::PreLpt(const std::vector<ov::element::Type>& defaultPrecis
         CPU_REGISTER_PASS_COMMON(manager, ngraph::pass::low_precision::ConvertSubtractConstant, defaultPrecisions);
     }
     CPU_REGISTER_PASS_COMMON(manager, ov::pass::Validate);
-    CPU_REGISTER_PASS_COMMON(manager, ov::pass::RefConvertI64ToI32);
+    // Common ConvertPrecision pass handles only limited set of opevino operations to match list of precision supported by the plugin.
+    // However if extension operation produces output precsion which is not natively supported this may lead to inconsistency during
+    // element type propagation. This transformation is called before ConvertPrecision pass to align precisions with according supported list.
```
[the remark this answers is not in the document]
nshchego: Done

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -245,11 +245,17 @@ void Graph::Replicate(const CNNNetwork &network) {
...
 
-            if (outputsInfo.count(name) != 0) {
+            for (const auto& name : input.get_names()) {
+                if (outputsInfo.count(name) != 0) {
+                    outputNodesMap[name] = node;
+                }
+            }
+
+            const auto name = ov::op::util::get_ie_output_name(input);
+            if (outputsInfo.count(name) != 0 && outputNodesMap.count(name) == 0) {
```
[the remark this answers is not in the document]
nshchego: I also added a check for nodes.

src/plugins/intel_cpu/src/graph.h
```diff
@@ -253,9 +254,24 @@ class Graph {
...
+        }
+        NodePtr getNodePtrByName(const std::string& name) const;
+        bool hasOutputWithName(const std::string& name) const;
+        void clear() {
+            nodesMap.clear();
+        }
+
+        private:
+            std::unordered_map<NodePtr, std::unordered_set<std::string>> nodesMap;
+    };
```
maxnick, line 270: 1. May be it is a matter of taste, but once the struct has a constructor and a private field, it should be converted to a class. It simply does not correspond to a C struct anymore and has all the features of a typical class.
2. All the requests are made by `name`, so we probably should have a map where `name` is a key. Why can't we just use name->ptr map as before? Having multiple names should not be a problem as long as they map the same pointer, shouldn't it? Actually this is a main question. So far I do not see significant advantages of having `std::unordered_map<NodePtr, std::unordered_set<std::string>>`

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1665,9 +1684,53 @@ void Graph::EnforceBF16() {
...
+        }
+    }
+}
+
+NodePtr Graph::OutputNodesMap::getNodePtrByName(const std::string& name) const {
+    NodePtr outNode;
+    for (auto nIt : nodesMap) {
+        if (nIt.second.find(name) != nIt.second.end()) {
+            outNode = nIt.first;
+            break;
```
maxnick, line 1720: NIT. Why not just return here?

nshchego: Fixed

src/plugins/intel_cpu/src/graph.h
nshchego: 1. Modified
2. The new approach provides the following benefits:
- Eliminates repetitive actions on a single node when iterating by name;
- Prevents cases when node was recreated with additional names in tensor;

src/plugins/intel_cpu/tests/functional/single_layer_tests/custom_op_insert_convert.cpp
```diff
@@ -0,0 +1,121 @@
...
+                  const ov::EvaluationContext& evaluationContext) const override {
+        return evaluate(output_values, input_values);
+    }
+
+    bool has_evaluate() const override {
+        return true;
+    }
+};
+
+class CustomOpCPUTest : public SubgraphBaseTest {
```
maxnick, line 77: May be it does make sense to control the number of convert layers inserted?

maxnick (approved):

PR #16953, opened 2023-04-14
[FP16][IE] exclude MVN and NormalizeL2 from precision sensitive marking

pavel-esir:
**Short description:** 
- We exclude MVN/NormalizeL2 nodes from precision sensitive pattern, since GPU handles f16 ever/underflows by storing accumulators in f32 and this handling is faster that adding convert nodes before and after.
- There was also missing `fuse_type_to` for `Eye` operation, fixed that.

**Detailed description:**
- In order to save accuracy on GPU for fp16 compressed models we normalization ops (MVN, NormalizeL2) to be inferred always in f32 precision [mark_subgraphs_to_keep_in_mixed_precision.cpp#L26-L48 ](https://github.com/openvinotoolkit/openvino/blob/f9bd2d2/src/common/transformations/src/transformations/common_optimizations/mark_subgraphs_to_keep_in_mixed_precision.cpp#L26-L48). But sometimes can lead to performance degradations, even on models which didn't suffer from accuracy problems.

- But in GPU plugin MVN (and probably NormalizeL2) indeed accumulates numbers in FP32 therefore we can safely exclude them from mixed precision pattern.  We tried that previously but had lost accuracy, and the reason of that turned out that models had MVNs with unsupported combinations of arguments and is decomposed to a subgraph [plugins/intel_gpu/src/plugin/transformations_pipeline.cpp#L354-L377.](https://github.com/openvinotoolkit/openvino/blob/f9bd2d2c1e0606eef7737b1405178fd16c59f48e/src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp#L354-L377]). Therefore when we disabled MVN marking this whole subgraph is left in f16 and we end up having accuracy loss.
- After excluding MVN, and NormalizeL2 from pattern matcher we need to resolve f16 problems for subgraphs which appeared because of gpu unsupported MVN. To do that we should firstly decompose `MVN -> ReduceSum + ... + Div with eps` and only after that call `ConvertPrecision f32->f16` with `keep_precision_sensitive_in_fp32=True`.

**Local performance check before this PR**:
69d6ef-current master

**Performance after this PR**:
not include MVN

### Tickets:
 - xxx-103413


rkazants (commented):
I strictly recommend you to start thinking of e2e pre-commit tests for FP16 compression like we have for MO Python API.

It can be a sort of the layer tests (with manually programmable models) for which you get the required sub-graphs. After conversion and compression, execute graph comparator and check for accuracy.

@andrei-kochin

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
```diff
@@ -194,6 +194,8 @@ void TransformationsPipeline::apply(std::shared_ptr<ov::Model> func) {
 
         // fuse softmax patterns so that they will not be marked as precision sensitive in ConvertPrecision
         manager.register_pass<ov::pass::SoftmaxFusion>();
+        // decompose MVNs that sre not supported in GPU, so the they will be marked as precision sensitive in ConvertPrecision
+        manager.register_pass<ov::pass::MVN6Decomposition>();
```
rkazants, line 198: why do you change it? Not clear. I see it is registered in the common optimization.

vladimir-paramuzov (approved):
overall lgtm

isanghao:
Performance looks good.

PR #16959, opened 2023-04-14
Fix Cast util functor when cast from floating point to integer

praasz:
### Details:
 - Fix `Cast` functor when perform cast from floating point to integer when floating point value is out of integer range.
 - Fix sanitizer `float-cast-overflow` runtime error message .

### Tickets:
 - 105002

### DIFF
--- a/src/core/shape_inference/include/shape_infer_type_utils.hpp
+++ b/src/core/shape_inference/include/shape_infer_type_utils.hpp
@@ -29,7 +29,9 @@ struct Cast {
     template <class U,
               typename std::enable_if<std::is_integral<T>::value && std::is_floating_point<U>::value>::type* = nullptr>
     constexpr T operator()(const U u) const {
-        return u < std::numeric_limits<T>::max() ? static_cast<T>(u) : std::numeric_limits<T>::max();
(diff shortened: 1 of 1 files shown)


src/core/shape_inference/include/shape_infer_type_utils.hpp
```diff
@@ -11,18 +11,26 @@ namespace ov {
...
+        class U,
+        typename std::enable_if<!std::is_integral<T>::value || !std::is_floating_point<U>::value>::type* = nullptr>
     constexpr T operator()(const U u) const {
         return static_cast<T>(u);
     }
+
+    template <class U,
+              typename std::enable_if<std::is_integral<T>::value && std::is_floating_point<U>::value>::type* = nullptr>
+    constexpr T operator()(const U u) const {
+        return u < std::numeric_limits<T>::max() ? static_cast<T>(u) : std::numeric_limits<T>::max();
```
ilyachur, line 32: I think we also can get the same issue in case of lower values... Am I wrong?

praasz: Yes you've got right it can be same issue for negative values.
Added additional check.

ilyachur (approved):

PR #16965, opened 2023-04-14
[PyOV] Deprecate PerformanceMode.UNDEFINED and refactor deprecation

jiwaszki:
### Details:
 - Deprecation of `PerformanceMode.UNDEFINED`.
 - Added part of submodules on path to `PerformanceMode.UNDEFINED` to correctly apply the solution.
 - Extended deprecation methods with `name` and `stacklevel` arguments.
 - Added two new decorators `classproperty` and `deprecatedclassproperty` to ensure new Python releases will be compliant. (`classmethod` and `property` combination won't work in future https://docs.python.org/3.11/library/functions.html?highlight=classmethod#classmethod)
 - Added testcase.

Continuation of work from https://github.com/openvinotoolkit/openvino/pull/16563 

Example:
```python
In [1]: import openvino.runtime as ov

In [2]: ov.properties.hint.PerformanceMode.UNDEFINED
<ipython-input-2-8f9140e7222e>:1: DeprecationWarning: PerformanceMode.UNDEFINED is deprecated and will be removed in version 2024.0. Please use actual value instead.
  ov.properties.hint.PerformanceMode.UNDEFINED
Out[2]: <PerformanceMode.UNDEFINED: -1>

In [3]: ov.properties.hint.PerformanceMode.LATENCY
Out[3]: <PerformanceMode.LATENCY: 1>
```

### Tickets:
 - *107175*


ilya-lavrenov (approved):

src/bindings/python/src/openvino/runtime/properties/hint/overloads.py
```diff
@@ -0,0 +1,19 @@
...
+from openvino.utils import deprecatedclassproperty
+
+from openvino._pyopenvino.properties.hint import PerformanceMode as PerformanceModeBase
+
+
+class PerformanceMode(PerformanceModeBase):
+
+    @deprecatedclassproperty(
+        name="PerformanceMode.UNDEFINED",
+        version="2024.0",
```
akuporos, line 14: Shouldn't it be 2023.2?

jiwaszki: CPP counterpart:
https://github.com/openvinotoolkit/openvino/blob/master/src/inference/include/openvino/runtime/properties.hpp#L302

PR #16968, opened 2023-04-14
Fix of crashes of convert_model() when executed for different frameworks

popovaan:
Root cause analysis: 
When convert_model() is executed for different frameworks in same script conversion crashes due to conflicts of transformations classes from different framework.

Solution: 
During class registration exclude from dependency graph all transformation from other frameworks.

Ticket: 107576

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Transformation tests

Documentation:
* [x]  User guide update


tools/mo/openvino/tools/mo/utils/class_registration.py
```diff
@@ -223,7 +216,14 @@ def determined_sort(self):
         return order
 
 
-def get_replacers_order(transform_types: list):
+def need_exclude_class(class_type, excluded_frameworks):
+    for framework in excluded_frameworks:
+        if "." + framework + "." in str(class_type):
+            return True
```
rkazants, line 222: Can it be a case that for some MxNet model we import routine from tf.front in some back transformation? I am afraid that we will get `ImportError`

popovaan: The comment `# skip importing loaders of other frameworks` was not correct here, we import modules earlier in `load_extensions`. So importing modules will be the same, and all imports will work. 

But this part changes how transformation dependency graph is constructed. Dependency graph is network-based graph where node are transformation, it determines what transformations will be actually applied (not only imported). So if some MxNet network has MxNet transformation that uses method from tf.front - it will work. But if MxNet network needs TF tranfsromation to be executed it will not work. In this case I suggest to move such transformations to common group as I did for FakeQuantWithMinMaxVars. It exactly what happed for FakeQuantWithMinMaxVars. 

FakeQuantWithMinMaxVars initially was added to dependency graph for all frameworks, even though it is in TF folder. It happened because dependency graph uses .__subclasses__() field of a class, which dynamically changes when a subclass is imported. And FakeQuantWithMinMaxVars was imported from disable_weights_quantize_value_propagation.py in common part. So it was always added to dependency graph and executed for all frameworks.

In this PR I forbid such cases.

tools/mo/openvino/tools/mo/utils/class_registration.py
```diff
@@ -223,7 +216,14 @@ def determined_sort(self):
...
+    return False
+
+
+def get_replacers_order(transform_types: list, excluded_frameworks: list):
```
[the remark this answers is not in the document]
popovaan: Removed, thanks.

rkazants (commented):
Please add a test where in the same process you sequentially call `convert_model` method for different FWs (TF-legacy, MxNet, ONNX-legacy, caffe). Also you can mix it with new FEs.

rkazants (commented):
Please add test. The changes look good to me

popovaan:
> Please add test. The changes look good to me

As agreed, we can add the test in another PR, similar to this test https://github.com/intel-innersource/frameworks.ai.openvino.tests/pull/934, but in public repo.

rkazants (approved):

PR #16981, opened 2023-04-17
[GPU] Fix proposal sort condition error

kelvinchoi-intel:
### Details:
 - In case proposals size is equal to n, use std::partial_sort as CPU

### Tickets:
 - *65120*


yeonbok:
Please add a unittest.

andrew-k-park (approved):

PR #16987, opened 2023-04-17
[CPU] ReduceMean fix for ACL Executor

alvoron:


alvoron:
RNN, LSTM and GRU tests were enabled since they are not requested any fix to be passed.

ilya-lavrenov (approved):

PR #16988, opened 2023-04-17
[GPU] Fixed OneDNN fc+sum fusion serialization

e-ddykim:
### Details:
 - This PR fixes a bug related with OneDNN fc+sum fusion serialization

### Tickets:
 - 107612


yeonbok (approved):

yeonbok:
There is a failing unittest. Is it relevant to this change....? (It seems not but please check)

src/plugins/intel_gpu/src/graph/graph_optimizer/basic_memory_dependencies.cpp
```diff
@@ -36,7 +36,7 @@ void basic_memory_dependencies::run(program& p) {
         }
 
         if (node->get_preferred_impl_type() == impl_types::onednn
-            && (node->is_type<convolution>() || node->is_type<deconvolution>())) {
+            && (node->is_type<convolution>() || node->is_type<deconvolution>() || node->is_type<fully_connected>())) {
```
isanghao, line 39: what about gemm or reduce? isn't it necessary to check gemm or reduce for such case?
In addition, isn't it necessary to confirm that other node does not have sum post operation?

e-ddykim: I removed the whitelist for sum post op fusion. Thank you.

isanghao:
No perf regression from dgpu

isanghao (approved):

PR #16991, opened 2023-04-17
Removed checks of unsatisfied dependencies in MO

popovaan:
Description: 
Removed checks of unsatisfied dependencies in MO

Ticket: 91273

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Transformation tests

Documentation:
* [x]  User guide update


tools/mo/openvino/tools/mo/utils/versions_checker.py
```diff
@@ -255,25 +326,34 @@ def check_requirements(framework=None, silent=True):
...
+    requirements_list = get_requirements_list_from_file(file_name, env_setup)
 
-    # if MO was run from developer clone requirements are 4 levels out from mo.py
-    requirements_file = os.path.realpath(os.path.join(os.path.dirname(__file__), os.pardir, os.pardir, os.pardir, os.pardir, file_name))
-    if not Path(requirements_file).is_file():
-        # if MO was run from site-packages requirements are at the same level with mo.py file
-        # in site-packages/openvino/tools/mo/requirements.txt
-        requirements_file = os.path.realpath(os.path.join(os.path.dirname(__file__), os.pardir, file_name))
+    # Get constraints list
+    constraints_list = get_requirements_list_from_file("../constraints.txt", env_setup)
```
rkazants, line 334: Please parse requirements file if it has dependency on the constrain file. If yes, you need to get the constraint file name and also parse it.

tools/mo/openvino/tools/mo/utils/versions_checker.py
```diff
@@ -231,6 +235,73 @@ def get_environment_setup(framework):
     return env_setup
```
rkazants, line 237: we also need this parsing functionality for requirements and constrain files in `setup.py`. There is already implemented it there but not quire much well.
Please create the common utils routines for both `setup.py` and MO and use it.

tools/mo/openvino/tools/mo/utils/versions_checker.py
popovaan: Removed checks of unsatisfied dependencies as discussed offline.

tools/mo/openvino/tools/mo/utils/versions_checker.py
popovaan: Removed checks of unsatisfied dependencies as discussed offline.

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -47,8 +47,7 @@
 from openvino.tools.mo.utils.utils import refer_to_faq_msg
 from openvino.tools.mo.utils.telemetry_utils import send_params_info, send_framework_info, send_conversion_result, \
     get_tid
-from openvino.tools.mo.utils.versions_checker import check_requirements, log_not_satisfied_dependencies, \
-    get_environment_setup  # pylint: disable=no-name-in-module
+from openvino.tools.mo.utils.versions_checker import get_environment_setup  # pylint: disable=no-name-in-module
```
slyalin, line 50: @popovaan, please explain the reason we still need to use this method. I see below
`
def check_model_object(argv):
    model = argv['input_model']
    if 'tensorflow' in sys.modules:
        import tensorflow as tf
        env_setup = get_environment_setup("tf")
`
Does it mean we will generate a warning when tensorflow is installed and the version doesn't match the expected one? Does it print warning even if a new FE is used?

rkazants (approved):

mvafin (approved):

PR #16992, opened 2023-04-17
[CPU] Rnn weights repacking

maxnick:
### Details:
Add RNN weights repacking in the case when their layout depends on the input shape.
Also withing this PR we adjust relative thresholds in some SL tests in accordance with the precision. 

### Tickets:
 - 95915
 - 107085

### DIFF
--- a/src/plugins/intel_cpu/src/node.cpp
+++ b/src/plugins/intel_cpu/src/node.cpp
@@ -799,11 +799,13 @@ void Node::prepareMemory(const DnnlMemoryDescPtr& intDesc, size_t indx) {
 
     MemoryPtr ptr;
     auto weightCache = context->getWeightsCache();
-    if (weightCache != nullptr) {
+    if (weightCache != nullptr && memory::format_kind::blocked == intDesc->getDnnlDesc().get_format_kind()) {
+        const auto& format = intDesc->serializeFormat();
         const uint64_t data_hash = weightCache->GetHashFunc().hash(
                 internalBlob->buffer(), internalBlob->byteSize());
 
         const std::string string_hash = name + "_" + std::to_string(indx)
+                                        + "_" + format
                                         + "_" + std::to_string(internalBlob->byteSize())
                                         + "_" + std::to_string(data_hash);
 
--- a/src/plugins/intel_cpu/src/node.cpp
+++ b/src/plugins/intel_cpu/src/node.cpp
@@ -771,6 +771,50 @@ void Node::initDescriptor(const NodeConfig& config) {
     selectedPD->setConfig(updatedConfig);
 }
 
+void Node::prepareMemory(const DnnlMemoryDescPtr& intDesc, size_t indx) {
+    size_t minSize = indx + 1;
+    if (internalBlobMemory.size() < minSize) {
+        internalBlobMemory.resize(minSize);
+    }
+
+    if (minSize > internalBlobs.size()) {
+        IE_THROW() << "Can't prepare memory for internal blob, requested index: " << indx <<
+            " is out of bounds of the internalBlobs vector of size " << internalBlobs.size();
+    }
+
+    const auto &internalBlob = internalBlobs[indx];
+
+    auto create = [&] () {
+        // TODO [DS]: internal blobs should be removed or rewritten using Memory object
+        auto newDesc = MemoryDescUtils::convertToDnnlBlockedMemoryDesc(internalBlob->getTensorDesc());
+
+        Memory memory{ engine };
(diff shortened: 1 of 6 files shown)


maxnick:
@ceciliapeng2011 could you please review?

src/plugins/intel_cpu/src/nodes/rnn.cpp
```diff
@@ -1072,40 +1072,38 @@ void RNN::prepareParams() {
                                                        key.wDescs,
                                                        key.attr);
 
-        return std::make_shared<DnnlExecutor>(descPtr);
+        return descPtr ? std::make_shared<RnnDnnlExecutor>(descPtr) : nullptr;
```
ceciliapeng2011, line 1075: Could it descPtr be NULL?

src/plugins/intel_cpu/src/node.cpp
```diff
@@ -771,46 +771,61 @@ void Node::initDescriptor(const NodeConfig& config) {
...
+    };
 
-        MemoryPtr ptr;
-        auto weightCache = context->getWeightsCache();
-        if (weightCache != nullptr) {
-            const uint64_t data_hash = weightCache->GetHashFunc().hash(
-                    internalBlob->buffer(), internalBlob->byteSize());
+    MemoryPtr ptr;
+    auto weightCache = context->getWeightsCache();
+    if (weightCache != nullptr && memory::format_kind::blocked == intDesc->getDnnlDesc().get_format_kind()) {
```
ceciliapeng2011, line 802: Except blocked and opaque format, could there be other format that we can cache the weights?

src/plugins/intel_cpu/src/node.cpp
```diff
@@ -771,46 +771,61 @@ void Node::initDescriptor(const NodeConfig& config) {
...
-            _ptr->Create(*intDescs[i]);
-            _ptr->SetData(memory);
+        Memory memory{ engine };
+        memory.Create(newDesc, internalBlob->buffer());
 
-            return _ptr;
-        };
+        MemoryPtr _ptr = std::make_shared<Memory>(engine);
+        _ptr->Create(intDesc);
+        node::Reorder::reorderData(memory, *_ptr, context->getParamsCache());
```
ceciliapeng2011, line 796: concern here :

Will this check and subnormals processing be skipped? Since node::Reorder::reorderData replaces where "_ptr->SetData(memory);" once be.
https://github.com/openvinotoolkit/openvino/blob/c186ffdf0dfc927710c9369dc7e193819e2e1b44/src/plugins/intel_cpu/src/cpu_memory.cpp#L78

ceciliapeng2011 (commented):
generally sounds good to me, leaving some small concerns.

src/plugins/intel_cpu/src/nodes/rnn.cpp
maxnick: Yes. Then `nullptr` value explicitly indicates that the builder has failed. This is a part of the multi cache usage protocol. Then the value is checked on L1083. The previous implementation always created an executor notwithstanding whether the primitive is empty or not, which might lead to a nullptr dereferencing, since the L1083 always passed.

src/plugins/intel_cpu/src/node.cpp
maxnick: At the moment oneDNN has four memory desc types: blocked, opaque, winograd, sparce. winograd format is used for winograd convolutions only. Sparce format at the moment is used for convolutions and fully connected nodes. But the `Node::prepareMemory(const std::vector<DnnlMemoryDescPtr>& intDescs)` call is used only for deconvolutions and rnn since they are two node types that still use internal blobs. Therefore, in this method it is enough to cache blocked format tensors only. Obviously, we have to revise the weight cache usage considering the dynamic shapes paradigm to avoid such condition checks.

src/plugins/intel_cpu/src/node.cpp
maxnick: Yep, good point, however the main intention of this PR is rather a bug fix. Thus I would like to preserve the previous behavior as much as possible. With no doubts,  we can optimize this method as you suggest, but I would prefer to do it in a separate PR targeting the next release.

maxnick:
@ceciliapeng2011 I hope I answered your questions. Could you please take a look once again?

ceciliapeng2011 (approved):

maxnick:
@ceciliapeng2011 thank you. @dmitry-gorokhov , could you please make the final review?

dmitry-gorokhov (approved):

PR #17002, opened 2023-04-17
[MO] Fix Interpolate-11 in MO

mvafin:
### Details:
 - *Fix issues with Interpolate-11 created by ONNX FE when reading with MO IR Reader: do not convert input 1 to int64 if opset11, do not recalculate scales if opset11*
 - *Fix TopK-11 output type*

### Tickets:
 - *108632, 108630*


tools/mo/unit_tests/mo/utils/ir_reader/ops_test.py
```diff
@@ -55,6 +57,28 @@ def test_interpolate_11(self):
...
+        graph = TestOps.check_graph_can_save(model, 'interpolate_model')
+        interpolate_node = graph.get_op_nodes(op="Interpolate")[0]
+        self.assertEqual(interpolate_node["version"], "opset11")
+
+    def test_interpolate_11_no_axes(self):
+        data_shape = [6, 12, 10, 24]
+        data_parameter = opset11.parameter(
+            data_shape, name="Data", dtype=np.float32)
+        interpolate = opset11.interpolate(data_parameter, np.int32(
+            [6, 12, 20, 48]), "nearest", "sizes", name="Interpolate_11")
```
rkazants, line 76: should not we specify new sizes for spatial dimensions according to the specification? By default, it should two sizes.

rkazants (commented):
Please check a case with no axes

rkazants (approved):

tools/mo/openvino/tools/mo/ops/interpolate.py
```diff
@@ -151,12 +152,14 @@ def __init__(self, graph: Graph, attrs: dict):
             'pads_end': 0,
 
             'infer': self.infer,
-
+
```
mvafin, line 155:
```suggestion

```

PR #17010, opened 2023-04-18
[GNA] Fix channel multiplier calculation

sirzabek:
### Details:
 - fix channel multiplier calculation/passing for int8 weights quantization

### Tickets:
 - 108530

### DIFF
--- a/src/plugins/intel_gna/src/frontend/quantization.cpp
+++ b/src/plugins/intel_gna/src/frontend/quantization.cpp
@@ -150,7 +150,7 @@ void QuantizeWeights<int8_t>(const QuantizationData& data,
                 if (quantized_weights) {
                     value -= MAX_VAL_1B_WEIGHT;
                 } else {
-                    value = value * (data.scale_factor / channel_multiplier) + rounding_value;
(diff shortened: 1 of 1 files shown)


marcinkusm (approved):

marcinkacprzak (commented):
If levels would be 1, application will crash.

src/plugins/intel_gna/src/frontend/quantization.cpp
```diff
@@ -150,7 +150,7 @@ void QuantizeWeights<int8_t>(const QuantizationData& data,
                 if (quantized_weights) {
                     value -= MAX_VAL_1B_WEIGHT;
                 } else {
-                    value = value * (data.scale_factor / channel_multiplier) + rounding_value;
+                    value = value * (data.scale_factor / ptr_int_biases[row].multiplier) + rounding_value;
```
marcinkacprzak, line 153: Wouldn't it be reasonable to add test for this change?

sirzabek: It would be reasonable, but doesn't seem feasible. This should be covered by accuracy tests.

marcinkacprzak (approved):

PR #17019, opened 2023-04-18
Fix MO IR Reader extender for StridedSlice to support empty masks

mvafin:
### Details:
 - *Fix MO IR Reader extender for `StridedSlice` to support empty begin and end masks*

### Tickets:
 - *108211*

### DIFF
--- a/tools/mo/openvino/tools/mo/utils/ir_reader/extenders/strided_slice_extender.py
+++ b/tools/mo/openvino/tools/mo/utils/ir_reader/extenders/strided_slice_extender.py
@@ -18,8 +18,6 @@ class StridedSlice_extender(Extender):
             if op.has(attr) and op[attr] != '':
                 Extender.attr_to_list(op, attr)
             else:
-                assert attr not in ['begin_mask', 'end_mask'],\
-                    '{} is not defined for the node {}'.format(attr, op.soft_get('name', op.id))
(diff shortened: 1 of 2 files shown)


tools/mo/openvino/tools/mo/utils/ir_reader/extenders/strided_slice_extender.py
```diff
@@ -18,8 +18,6 @@ def extend(op: Node):
             if op.has(attr) and op[attr] != '':
                 Extender.attr_to_list(op, attr)
             else:
-                assert attr not in ['begin_mask', 'end_mask'],\
-                    '{} is not defined for the node {}'.format(attr, op.soft_get('name', op.id))
```
rkazants, line 22: I see that `begin_mask` and `end_mask` are mandatory attributes in the specification. I think some other component erases this attribute

rkazants: If there are such atributes in IR, why do we remove this check?

mvafin: Because empty attributes are not added into the model, so MO IR Reader have extender that adds such attributes as empty array.

PR #17022, opened 2023-04-18
Prepare "memory_optimization_guide.md"

vurusovs:
### Details:
 - From time to time we receives issues with "false positive memory leak" caused by allocator behavior. So I prepare document where describe ways to optimize memory usage and determine real application leaks

### Tickets:
 - [CVS-108277](https://jira.devtools.intel.com/browse/CVS-108277)

### DIFF
--- a/docs/optimization_guide/memory_optimization_guide.md
+++ b/docs/optimization_guide/memory_optimization_guide.md
@@ -4,22 +4,23 @@
 
 .. warning::
 
-   Before applying any of recommendations provided here, note that it may significantly impact first inference latency. 
+   Before applying any of the recommendations provided here, note that it may significantly impact first inference latency. 
 
 The most RAM-consuming OpenVINO stage is model compilation. It may cause several issues:
 
 * Not enough memory to compile a model. To decrease memory requirement, the following options may be applied: 
   
-  * Weights mapping - we introduced memory mapping (using ``mmap``) as a default way to work
-    with weights. For the moment feature is supported by IR frontend.
-    Mapping may be switched by specifying ``ov::enable_mmap(BOOL)`` property for the ``ov::Core``.
-    Because of its "memory on demand" nature, there is no need to store weights fully
-    in RAM, so it decreases memory level required for compilation. Also, ``mmap`` provides
-    extensive memory sharing, so the following compilation of the same model will fetch same
-    memory from RAM instead of additional read from storage
+  * Weights mapping - memory mapping (using ``mmap``) has been introduced as the default way to work
+    with weights. Currently, this feature is supported by the IR frontend.
+    Mapping may be switched by specifying the ``ov::enable_mmap(BOOL)`` property for the ``ov::Core``.
+    Because of its "memory-on-demand" nature, there is no need to store all weights
+    in RAM. Storing just the data that is needed at the moment lowers the amount of memory
+    required for compilation. Moreover, ``mmap`` provides extensive memory sharing, so the
+    consecutive compilation of the same model will fetch the information already stored in RAM
+    instead of reading it one more time from storage.
 
   * Decrease the number of threads for compilation - to change the number of threads, specify
-    ``ov::compilation_num_threads(NUMBER)`` property for the ``ov::Core`` or pass as an additional 
+    the ``ov::compilation_num_threads(NUMBER)`` property for the ``ov::Core`` or pass it as an additional 
     argument to ``ov::Core::compile_model()``
 
 * Not enough memory to recompile a model. If model compilation is successful but one of the following recompilations fails due lack of resources, it may be caused by:
@@ -45,6 +46,6 @@ The most RAM-consuming OpenVINO stage is model compilation. It may cause several
       More details on the two may be found in the 
       `GNU Tunables Manual <https://www.gnu.org/software/libc/manual/html_node/Tunables.html>`__
 
-    * Try another allocator. One of the allocators which operates with memory carefully is ``jemalloc``
+    * Try another allocator. One of the allocators that handles memory carefully is ``jemalloc``
 
 @endsphinxdirective
--- a/docs/optimization_guide/memory_optimization_guide.md
+++ b/docs/optimization_guide/memory_optimization_guide.md
@@ -1,23 +1,50 @@
 # Optimizing memory usage {#openvino_docs_memory_optimization_guide}
 
-Before applying any of provided recomendations, please, note that it may significantly impact first inference latency. 
+@sphinxdirective
+
+.. warning::
+
+   Before applying any of recommendations provided here, note that it may significantly impact first inference latency. 
 
 The most RAM-consuming OpenVINO stage is model compilation. It may cause several issues:
 
-* Not enough memory to compile model. To decrease memory requirement next options may be applied: 
+* Not enough memory to compile a model. To decrease memory requirement, the following options may be applied: 
   
-  * IR mapping - we introduced memory mapping (using `mmap`) as a default way to operate with IR. Because of "memory on demand" nature, there is no need to store weights fully in RAM, so it decreases memory level required for compilation. Also `mmap` provides extensive memory sharing, so next compilation of same model will fetch same memory from RAM instead of additional read from storage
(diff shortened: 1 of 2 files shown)


kblaszczak-intel (commented):
I have added a suggestion with some language and formatting changes. The formatting is required by our recent shift to rst from markdown - that's why we need @sphinxdirective at the top and @endsphinxdirective at the bottom of the article.
Moreover, we need to add the ID of this article to a TOC of its parent article to have it published in the menu. In which section do we want to put it? Is it for Optimize Inference? Maybe under General Optimizations?

ilyachur:
I think this change can be ported to 2023.0 release too

docs/optimization_guide/memory_optimization_guide.md
```diff
@@ -0,0 +1,23 @@
...
+
+* Not enough memory to recompile model. If model compilation was successful, but one of next recompilations failed due lack of resources, it may be caused by:
+
+  * Memory leak - to determine direct leaks tools like `address-sanitizer`, `valgrind` etc. may be used. In case of indirect leaks that can't be catched by tools, peak RAM (VMHWM) may be tracked (as tracking tool tests/stress_tests/memleaks_tests may be used). If you experience significant memory usage increase, please, report to [Github "Issues"](https://github.com/openvinotoolkit/openvino/issues)
+
+  * Memory allocator behavior - each allocator works according to unique strategy and balancing between performance and memory usage. For example, GNU allocator aggressively requests more memory from OS for every next model compilation than was required for very first compilation (such behavior may be determined by tracking actual RAM (VMRSS) after compilation, it will grow until some stable point). To optimize memory pressure, next options are available:
+
+    * Apply `malloc_trim(0)`. The function attempts to release free memory even from threads caches, so it may singificantly decrease and stabilize VMRSS usage
```
vurusovs, line 19:
```suggestion
    * Apply `malloc_trim(0)`. The function attempts to release free memory even from threads caches, so it may significantly decrease and stabilize VMRSS usage
```

docs/optimization_guide/memory_optimization_guide.md
```diff
@@ -0,0 +1,23 @@
+# Optimizing memory usage {#openvino_docs_memory_optimization_guide}
+
+Before applying any of provided recomendations, please, note that it may significantly impact first inference latency. 
+
+The most RAM-consuming OpenVINO stage is model compilation. It may cause several issues:
+
+* Not enough memory to compile model. To decrease memory requirement next options may be applied: 
+  
+  * IR mapping - we introduced memory mapping (using `mmap`) as a default way to operate with IR. Because of "memory on demand" nature, there is no need to store weights fully in RAM, so it decreases memory level required for compilation. Also `mmap` provides extensive memory sharing, so next compilation of same model will fetch same memory from RAM instead of additional read from storage
```
ilyachur, line 9: I propose to rephrase this. That we introduced `mmap` support IR frontend already supports it

ilyachur: I think we also can notice the option `enable_mmap`

docs/optimization_guide/memory_optimization_guide.md
```diff
@@ -0,0 +1,23 @@
...
+
+  * Memory allocator behavior - each allocator works according to unique strategy and balancing between performance and memory usage. For example, GNU allocator aggressively requests more memory from OS for every next model compilation than was required for very first compilation (such behavior may be determined by tracking actual RAM (VMRSS) after compilation, it will grow until some stable point). To optimize memory pressure, next options are available:
+
+    * Apply `malloc_trim(0)`. The function attempts to release free memory even from threads caches, so it may singificantly decrease and stabilize VMRSS usage
+
+    * Use glibc `Tunables`. A few promising options are `glibc.malloc.trim_threshold` and `glibc.malloc.arena_max`. More details may be found here: https://www.gnu.org/software/libc/manual/html_node/Tunables.html
+
+    * Try another allocator. One of the allocators which carefully operates with memory is `jemalloc`
```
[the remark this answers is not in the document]
vurusovs: Done, please, check one more time

docs/optimization_guide/memory_optimization_guide.md
```diff
@@ -0,0 +1,23 @@
+# Optimizing memory usage {#openvino_docs_memory_optimization_guide}
+
+Before applying any of provided recomendations, please, note that it may significantly impact first inference latency.
```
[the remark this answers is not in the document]
vurusovs: Done

docs/optimization_guide/memory_optimization_guide.md
vurusovs: Done, please check

ilyachur (approved):

docs/optimization_guide/memory_optimization_guide.md
```diff
@@ -0,0 +1,50 @@
...
+
+    * Apply ``malloc_trim(0)``. The function attempts to release free memory even from thread
+      caches, so it may signifficantly decrease and stabilize VMRSS usage
+
+    * Use glibc ``Tunables``. A couple of promising options are:
+      ``glibc.malloc.trim_threshold`` and `glibc.malloc.arena_max`. 
+      More details on the two may be found in the 
+      `GNU Tunables Manual <https://www.gnu.org/software/libc/manual/html_node/Tunables.html>`__
+
+    * Try another allocator. One of the allocators which operates with memory carefully is ``jemalloc``
```
kblaszczak-intel, line 48:
```suggestion
.. warning::

   Before applying any of the recommendations provided here, note that it may significantly impact first inference latency. 

The most RAM-consuming OpenVINO stage is model compilation. It may cause several issues:

* Not enough memory to compile a model. To decrease memory requirement, the following options may be applied: 
  
  * Weights mapping - memory mapping (using ``mmap``) has been introduced as the default way to work
    with weights. Currently, this feature is supported by the IR frontend.
    Mapping may be switched by specifying the ``ov::enable_mmap(BOOL)`` property for the ``ov::Core``.
    Because of its "memory-on-demand" nature, there is no need to store all weights
    in RAM. Storing just the data that is needed at the moment lowers the amount of memory
    required for compilation. Moreover, ``mmap`` provides extensive memory sharing, so the
    consecutive compilation of the same model will fetch the information already stored in RAM
    instead of reading it one more time from storage.

  * Decrease the number of threads for compilation - to change the number of threads, specify
    the ``ov::compilation_num_threads(NUMBER)`` property for the ``ov::Core`` or pass it as an additional 
    argument to ``ov::Core::compile_model()``

* Not enough memory to recompile a model. If model compilation is successful but one of the following recompilations fails due lack of resources, it may be caused by:

  * Memory leak - to determine direct leaks, you can use tools like 'address-sanitizer' or
    'valgrind'. In case of indirect leaks, which cannot be caught by tools, peak RAM (VMHWM)
    may be tracked (you can use tests/stress_tests/memleaks_tests as a tracking tool). If you
    experience significant memory usage increase, report it in 
    `Github "Issues" <https://github.com/openvinotoolkit/openvino/issues>`__

  * Memory allocator behavior - each allocator works according to a unique strategy and
    balances between performance and memory usage. For example, the GNU allocator aggressively
    requests from the OS for more memory for consecutive model compilations than was
    required for the first compilation (such behavior may be determined by tracking actual RAM
    (VMRSS) after compilation - it will grow until some stable point). To optimize memory
    pressure, the following options are available:

    * Apply ``malloc_trim(0)``. The function attempts to release free memory even from thread
      caches, so it may signifficantly decrease and stabilize VMRSS usage

    * Use glibc ``Tunables``. A couple of promising options are:
      ``glibc.malloc.trim_threshold`` and `glibc.malloc.arena_max`. 
      More details on the two may be found in the 
      `GNU Tunables Manual <https://www.gnu.org/software/libc/manual/html_node/Tunables.html>`__

    * Try another allocator. One of the allocators that handles memory carefully is ``jemalloc``
```

kblaszczak-intel (approved):
a suggestion with minor adjustments added. Even without it, the article is very nice.

docs/optimization_guide/memory_optimization_guide.md
vurusovs: Applied suggestion as is, thank you!

PR #17026, opened 2023-04-18
Interpolate op cleanup

tomdol:
### Details:
 - Remaining work around the spec and core op

### Tickets:
 - 106374


src/core/shape_inference/include/interpolate_shape_inference.hpp
```diff
@@ -218,6 +218,10 @@ void shape_infer(const Interpolate* op,
         if (op->get_attrs().shape_calculation_mode == Interpolate::ShapeCalcMode::SCALES) {
             std::vector<float> scales;
             if (get_data_as_float<T>(1, op, scales, constant_data)) {
+                NODE_VALIDATION_CHECK(
+                    op,
+                    axes.size() == scales.size(),
+                    "The number of elements in the 'scales_or_sizes' input does not match the number of axes");
```
mitruska, line 224: There should be a test covering this check and expected error.

docs/ops/image/Interpolate_11.md
mitruska, line 1: To reflect all of the restrictions related to the pillow modes implementation described in https://github.com/openvinotoolkit/openvino/pull/16342, 
I would recommend to add a note to `antialias` and `coordinate_transformation_mode` attributes:

For antialias:
**Note**:  In case of `bilinear_pillow` or `bicubic_pillow` the `antialias` attribute is ignored (Pillow-kind antialias is always applied).

For coordinate_transformation_mode:
**Note**:  In case of `bilinear_pillow` or `bicubic_pillow` the `coordinate_transformation_mode` attribute is ignored.

mitruska (approved):

PR #17029, opened 2023-04-18
[ARM CPU] Fix eltwise op tests (Divide)

allnes:


ilya-lavrenov (approved):

src/plugins/intel_cpu/src/transformations/cpu_opset/arm/pass/convert_i32_div.cpp
```diff
@@ -0,0 +1,32 @@
...
+namespace intel_cpu {
+
+ConvertI32Div::ConvertI32Div() {
+    register_matcher(std::make_shared<ngraph::pattern::Matcher>(ngraph::pattern::wrap_type<ngraph::opset1::Divide>(), "ConvertI32Div"),
+         [](ngraph::pattern::Matcher& m) {
+             auto divide = std::dynamic_pointer_cast<ngraph::opset1::Divide>(m.get_match_root());
+             if (!divide) {
+                 return false;
+             }
+             if (divide->get_element_type() != ov::element::i32) {
```
dmitry-gorokhov, line 18: I would propose to use `divide->get_element_type().is_integral_number()` instead. We are going to merge I64 native support, so it would be nice to make this transofmration suitable for upcoming change.

src/plugins/intel_cpu/src/transformations/cpu_opset/arm/pass/convert_i32_div.cpp
```diff
@@ -0,0 +1,32 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include "convert_i32_div.hpp"
+#include <ngraph/opsets/opset1.hpp>
+#include <ngraph/rt_info.hpp>
+
+namespace ov {
+namespace intel_cpu {
+
+ConvertI32Div::ConvertI32Div() {
```
dmitry-gorokhov, line 11: Maybe: `DecomposeIntegerDivide`?
File name should also be changed

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
```diff
@@ -262,6 +263,7 @@ void Transformations::PreLpt(const std::vector<ov::element::Type>& defaultPrecis
     CPU_REGISTER_PASS_ARM(manager, ConvertConv1D);
     CPU_REGISTER_PASS_ARM(manager, ConvertGroupConv1D);
     CPU_REGISTER_PASS_ARM(manager, ConvertGroupConvolution);
+    CPU_REGISTER_PASS_ARM(manager, ConvertI32Div);
```
dmitry-gorokhov, line 266: Lets put the explanation here: `The plugin computes Divide in floating point precision. To preserve correct math for integer division we need to insert explicit Floor operation.`

allnes:
@dmitry-gorokhov done

dmitry-gorokhov (approved):

PR #17039, opened 2023-04-19
[GPU] Improve shape infer performance 

yeonbok:
### Details:
 - For iGPU, using usm_device for shape_infer_dep memory causes memcpy from device to host 
-  This PR prevents such copy to improve shape infer performance in dynamic shape execution for iGPU
-  Also fixed is_shape_infer_dep to use pointer instead of unique id (unique id is not set before compile_graph stage)

### DIFF
--- a/src/plugins/intel_gpu/src/graph/include/program_node.h
+++ b/src/plugins/intel_gpu/src/graph/include/program_node.h
@@ -94,7 +94,7 @@ public:
                 if (u->is_fused_dep(dep_idx)) {
                     continue;
                 }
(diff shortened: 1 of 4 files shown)


vladimir-paramuzov (approved):
Overall LGTM. Maybe we can use usm_host for all allocations on iGPU? Could you check that approach later?

yeonbok:
> Overall LGTM. Maybe we can use usm_host for all allocations on iGPU? Could you check that approach later?

Yes at first I was going to do that, but seems we need some more careful treatment for get_estimated_memory_usage and so on, I will improve after CF.

PR #17049, opened 2023-04-19
Fix ScatterNdUpdate index calculation

olpipi:
### Details:
 - Fix index calculation for copy in ScatterNdUpdate for reference impl and cpu plugin
 - Add test for this corner case

### Tickets:
 - CVS-108289

### DIFF
--- a/src/plugins/intel_gna/tests/functional/pass_tests/convolution_crop_axis_h.cpp
+++ b/src/plugins/intel_gna/tests/functional/pass_tests/convolution_crop_axis_h.cpp
@@ -114,8 +114,9 @@ const std::vector<InferenceEngine::Precision> netPrecisions = {
     InferenceEngine::Precision::FP32,
 };
 
-const std::vector<std::map<std::string, std::string>> configs = {{{"GNA_DEVICE_MODE", "GNA_SW_EXACT"}},
-                                                                 {{"GNA_DEVICE_MODE", "GNA_SW_FP32"}}};
+const std::vector<std::map<std::string, std::string>> configs = {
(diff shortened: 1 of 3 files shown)


src/plugins/intel_cpu/src/nodes/scatter_update.cpp
```diff
@@ -422,6 +422,7 @@ void ScatterUpdate::scatterNDUpdate(uint8_t *indices, uint8_t *update, uint8_t *
             size_t idxValue = getIndicesValue(indices, indicesOffset + i);
             dstOffset += idxValue * srcBlockND[i + 1];
         }
+        dstOffset--;
```
sivanov-work, line 425: shall we out here `assert(dstOffet)` at least?
otherwise comment why we do need extract 1 - would be appreciated

src/core/reference/include/ngraph/runtime/reference/scatter_nd_update.hpp
```diff
@@ -39,7 +39,7 @@ void scatterNdUpdate(const dataType* const inputData,
     for (size_t i = 0; i != num_of_updates; ++i) {
         const auto indices_coord = indices + i * indicesShape.back();
         const auto coord = span(indices_coord, indicesShape.back());
-        const auto out_index = std::inner_product(begin(coord), end(coord), begin(input_data_dim_pading), uint64_t(0));
+        const auto out_index = std::inner_product(begin(coord), end(coord), begin(input_data_dim_pading), uint64_t(0)) - 1;
```
sivanov-work, line 42: To exclude memory corruption on different corner cases - it would be great to put check on 0 before extracting 1
and also comment why we need that operation would be beneficial

PR #17052, opened 2023-04-19
[CPU] Remove legacy dynamic batch processing from the plugin

maxnick:
### Details:
Remove the legacy dynamic batch processing from the plugin. The legacy dynamic batch feature is still supported via processing the network by the upper bound of the batch size. Now the main approach to running networks with a dynamic batch is to use dynamic shapes.


ilya-lavrenov (approved):

maxnick:
@dmitry-gorokhov, could you please take a look?

src/plugins/intel_cpu/src/graph.h
```diff
@@ -190,6 +190,17 @@ class Graph {
...
     }
 
+    /**
+     * @brief This call updates the dynamic batch value
+     * 
+     * @note It is used for backward compatibility with legacy API only.
+     * @param newDynBatch
+     * new dynamic batch value
+     */
+    void setDymBatch(int newDynBatch) {
```
dmitry-gorokhov, line 200: `setDymBatch` -> `setDynBatch`

src/plugins/intel_cpu/src/exec_network.cpp
```diff
@@ -81,14 +81,7 @@ ExecNetwork::ExecNetwork(const InferenceEngine::CNNNetwork &network,
...
-    if (_cfg.isNewApi) {
-        int64_t maxBatchSize = -1;
-        if (canBeExecViaLegacyDynBatch(function, maxBatchSize)) {
-            IE_ASSERT(maxBatchSize > -1);
-            _cfg.batchLimit = maxBatchSize;
-        }
-    } else if (_cfg.batchLimit > 1) {
+    if (_cfg.batchLimit > 1) {
         // check topology for applicability
         if (!CanProcessDynBatch(_network)) {
```
dmitry-gorokhov, line 86: Why `CanProcessDynBatch` is not removed?

src/plugins/intel_cpu/src/nodes/transpose.cpp
```diff
@@ -382,12 +381,7 @@ void Transpose::execute(dnnl::stream strm) {
...
 
-        int MB = 0;
-        if (isDynamicNode()) {
-            MB = srcMemPtr->getStaticDims()[0];
-        } else {
-            MB = batchToProcess();
-        }
+        int MB = srcMemPtr->getStaticDims()[0];
 
         execPtr->exec(this, srcMemPtr, dstMemPtr, MB);
```
dmitry-gorokhov, line 386: it also would be nice to remove `batch` from executor API (here and in other places). However can be done as part of general executors refactoring activity.

maxnick: Yes, it does make sense to handle as a separate activity.

dmitry-gorokhov (approved):

PR #17054, opened 2023-04-19
[LPT] Extending EliminateFakeQuantize transformation

eshoguli:
### Details:
 - *Extending EliminateFakeQuantize transformation*

### Tickets:
 - *extending 98890 & workaround 107875*

### DIFF
--- a/src/common/low_precision_transformations/src/eliminate_fake_quantize.cpp
+++ b/src/common/low_precision_transformations/src/eliminate_fake_quantize.cpp
@@ -9,6 +9,7 @@
 #include <ngraph/ngraph.hpp>
 #include <ngraph/pattern/op/wrap_type.hpp>
 #include "itt.hpp"
+#include "low_precision/network_helper.hpp"
 
 namespace ngraph {
 namespace pass {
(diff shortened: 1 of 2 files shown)


src/common/low_precision_transformations/src/eliminate_fake_quantize.cpp
```diff
@@ -46,25 +47,61 @@ bool EliminateFakeQuantizeTransformation::transform(TransformationContext& conte
...
-            return false;
+            const auto diff = std::fabs(constant_value - value);
+            if ((exact_comparison && (std::fabs(constant_value - value) > std::numeric_limits<float>::epsilon())) ||
+                (diff > max_diff)) {
+                return false;
+            }
+
+            need_to_check_intervals = true;
         }
     }
```
v-Golubev, line 67: I suggest refactoring the code in order to reduce computations. Example:
```
for (const auto constant_value : constant_values) {
    const auto diff = std::fabs(constant_value - value);
    if (diff > std::numeric_limits<float>::epsilon()) {
        if (exact_comparison || diff > max_diff) {
            return false;
        }
        need_to_check_intervals = true;
    }
}
```

v-Golubev (approved):

v-Golubev (approved):

eshoguli:
more suitable version was suggested and merged: https://github.com/openvinotoolkit/openvino/pull/17140

PR #17055, opened 2023-04-19
[GPU] Fix dump graph failure issue in levit-128s model.

hyunback:
1. to_string() in strided_slice always access begin/end/stride param id from dependencies regardless max dependencies. Fix only for vaild.
2. Add an exception in dump_full_node(). It helps below.
- Avoid a dump failure. Usually, graph dump are used during debugging, which reduces unnecessary debugging time due to graph dump failure.
- You can immediately see which node has failed, making it easy to find it.

### Tickets:
 - *108865*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/program_dump_graph.cpp
+++ b/src/plugins/intel_gpu/src/graph/program_dump_graph.cpp
@@ -136,7 +136,16 @@ void close_stream(std::ofstream& graph) { graph.close(); }
 
 std::string get_node_id(const program_node* ptr) { return "node_" + std::to_string(reinterpret_cast<uintptr_t>(ptr)); }
 
-void dump_full_node(std::ofstream& out, const program_node* node) { out << node->type()->to_string(*node); }
+void dump_full_node(std::ofstream& out, const program_node* node) {
+    GPU_DEBUG_GET_INSTANCE(debug_config);
+    try {
+        out << node->type()->to_string(*node);
+    } catch(const std::exception& e) {
+        GPU_DEBUG_IF(debug_config->verbose >= 2) {
+            std::cerr << node->id() << " to_string() error: " << e.what() << '\n';
+        }
+    }
+}
 }  // namespace
 
 std::string get_dir_path(const ExecutionConfig& config) {
--- a/src/plugins/intel_gpu/src/graph/program_dump_graph.cpp
+++ b/src/plugins/intel_gpu/src/graph/program_dump_graph.cpp
@@ -6,6 +6,7 @@
 #include "to_string_utils.h"
 #include "data_inst.h"
 #include "condition_inst.h"
+#include "json_object.h"
 
 #include <algorithm>
 #include <vector>
@@ -141,9 +142,14 @@ void dump_full_node(std::ofstream& out, const program_node* node) {
     try {
         out << node->type()->to_string(*node);
     } catch(const std::exception& e) {
-        GPU_DEBUG_IF(debug_config->verbose >= 2) {
-            std::cerr << node->id() << " to_string() error: " << e.what() << '\n';
-        }
+        auto node_info = std::shared_ptr<json_composite>(new json_composite());
+        node_info->add("id", node->id());
+        node_info->add("error", "failed to make string from descriptor");
+        std::stringstream emtpy_desc;
+        node_info->dump(emtpy_desc);
+        out << emtpy_desc.str();
+
+        GPU_DEBUG_INFO << node->id() << " to_string() error: " << e.what() << '\n';
     }
 }
 }  // namespace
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/graph/strided_slice.cpp
```diff
@@ -114,9 +114,10 @@ std::string strided_slice_inst::to_string(strided_slice_node const& node) {
 
     json_composite strided_slice_info;
     strided_slice_info.add("input id", input.id());
-    strided_slice_info.add("begin_param id", node.get_dependency(1).id());
-    strided_slice_info.add("end_param id", node.get_dependency(2).id());
-    strided_slice_info.add("stride_param id", node.get_dependency(3).id());
+    std::vector<std::string> dependencies_info = {"begin_param id", "end_param id", "stride_param id"};
+    for (size_t i = 0; i < node.get_dependencies().size(); ++i) {
+        strided_slice_info.add(dependencies_info[i], node.get_dependency(i).id());
+    }
```
yeonbok, line 120: stride param is the only optional and others are required input. Isn't it enough to check the last param exists or not and dump it if exists?

hyunback: I'm not sure but levit-128s fp16 is running only 1 param now.
stridedslice:Slice_196 to_string() error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
stridedslice:Slice_201 to_string() error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
stridedslice:Slice_482 to_string() error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
stridedslice:Slice_487 to_string() error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)

hyunback: I'm not sure but levit-128s fp16 is running only 1 param now.
stridedslice:Slice_196 to_string() error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
stridedslice:Slice_201 to_string() error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
stridedslice:Slice_482 to_string() error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
stridedslice:Slice_487 to_string() error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)

yeonbok: Ah I see if they are constant, they are set as attribute instead of actual node..

src/plugins/intel_gpu/src/graph/program_dump_graph.cpp
```diff
@@ -136,7 +136,16 @@ void close_stream(std::ofstream& graph) { graph.close(); }
...
-void dump_full_node(std::ofstream& out, const program_node* node) { out << node->type()->to_string(*node); }
+void dump_full_node(std::ofstream& out, const program_node* node) {
+    GPU_DEBUG_GET_INSTANCE(debug_config);
+    try {
+        out << node->type()->to_string(*node);
+    } catch(const std::exception& e) {
+        GPU_DEBUG_IF(debug_config->verbose >= 2) {
+            std::cerr << node->id() << " to_string() error: " << e.what() << '\n';
+        }
+    }
```
yeonbok, line 147: Instead of just skiping could we add the node to the graph with empty properties?

src/plugins/intel_gpu/src/graph/strided_slice.cpp
e-ddykim, line 120: Could you check if the dependency index for "begin_param id" is 0? It was 1 before this change.

src/plugins/intel_gpu/src/graph/program_dump_graph.cpp
hyunback: Good idea, applied it.

src/plugins/intel_gpu/src/graph/strided_slice.cpp
hyunback: Yes, it was also issue, and I tried, issue was also reproduced based 0.

src/plugins/intel_gpu/src/graph/program_dump_graph.cpp
isanghao: 1. What about:
```
GPU_DEBUG_INFO << node->id() << " to_string() error: " << e.what() << '\n';
```
You don't need `GPU_DEBUG_IF`

2. Instead of "empty" what about adding an error message there too so that user can understand it failed? Such as "failed to make string from descriptor"

isanghao (approved):

isanghao (approved):

yeonbok (approved):

PR #17071, opened 2023-04-19
[ONNX FE] Fixed handling duplicates during graph extraction

mbencer:
Ticket: 103652
Let's say that a ONNX model has outputs A, A, B.
A user cut the model in A place.
Current state:  The model has A, A outputs.
New state: The model has only an A output.
Only, if a user cut the model in A,A places explicity, the model should still have A,A outputs after cuttig.


src/frontends/onnx/tests/models/model_editor/reference/subgraph__input_edge_from_tensor_with_multiple_consumers_2.prototxt
```diff
@@ -147,17 +147,11 @@ graph {
     }
   }
   output {
-    name: "mul2"
+    name: "add2"
```
mbencer, line 150: After the change outputs list on a model matches order of outputs passed by a user.

artkuli (approved):
LGTM 🚀

tomdol (approved):

PR #17072, opened 2023-04-19
Handle serialization of std::vector<double>

mbencer:
Ticket: 98641


mbencer:
@ilyachur @rkazants Do you maybe know if lack of supported of vector<double> during serialization is expected?
Attribute visitor [handles such a case](https://github.com/openvinotoolkit/openvino/blob/master/src/core/include/openvino/core/attribute_visitor.hpp#L90).

mbencer:
TODO: Handling deserialization in `openvino/src/frontends/ir/src/ir_deserializer.cpp`

PR #17074, opened 2023-04-19
Update run_timetest.py

AlbertAnglesIntel:
User story: As a validation engineer I would like to Change method for calculating final KPI result for FIL, FEIL. For now the KPI is calculated via AVG from 5 executions. This method might not be authoritative as the attached report shows. Better solution might be to calculate median instead of AVG and 5 iterations might be not enough for authoritative result.

### Details:
 - Change timetest FIL/FEIL use media instead of mean for averaging the fil/feil timetest results. In attached ticket it was observed that the mean is not be good method for averaging 

### Tickets:
 - https://jira.devtools.intel.com/browse/EISW-70394


vurusovs:
At initial stage of tests development we determined metrics that will be more useful for comparison from run to run. Because time values are volatile, we need to determine range values - `[AVG - STDEV, AVG + STDEV]`. And idea is to have `STDEV` as smaller as possible (by using locked CPU frequency, doing warm up, manage caches etc.) to bring values from range to some approximate value. So I see no reason to replace `AVG` with `MEAN`. If there is a real mathematical proof for the need to add `MEAN`, we may do it in addition to other metrics

AlbertAnglesIntel:
> At initial stage of tests development we determined metrics that will be most useful for comparison from run to run. Because time values are volatile, we need to determine range values - `[AVG - STDEV, AVG + STDEV]`. And idea is to have `STDEV` as smaller as possible (by using locked CPU frequency, doing warm up, manage caches etc.) to bring values from range to some approximate value. So I see no reason to replace `AVG` with `MEAN`. If there is a real mathematical proof for the need to add `MEAN`, we may do it in addition to other metrics

its not about MEAN but MEDIAN. 

MEDIAN is better when the results are highly sparse

PR #17076, opened 2023-04-19
[LPT] AddTransformation fix

v-Golubev:
### Details:
AddTransformation branch selection fix: the logic corrected for the cases when bias after Conv/GrConv/MatMul has constant on zero's input.

### Tickets:
 - *CVS-108848*

### DIFF
--- a/src/common/low_precision_transformations/tests/add_transformation.cpp
+++ b/src/common/low_precision_transformations/tests/add_transformation.cpp
@@ -69,6 +69,7 @@ public:
     Actual actual;
     Expected expected;
     std::string additionalLayer;
+    std::string postops_configuration;
 };
 
 typedef std::tuple<ngraph::element::Type,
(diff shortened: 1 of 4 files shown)


src/common/low_precision_transformations/tests/add_transformation.cpp
```diff
@@ -439,6 +443,23 @@ const std::vector<AddTransformationTestValues> testValuesWithoutConstantBranches
...
+      {{ngraph::element::f32}, {3.f}, {5.f}},
+      {}},
+     {ngraph::element::u8,
+      {{}, {}, {}},
+      ngraph::element::u8,
+      {{ngraph::element::f32}, {17.f}, {0.5f}},
+      {{}, {}, {10.f}},
+      {}},
+     "convolution",
+     "bias_on_zero_input"},
```
eshoguli, line 462: I'm just interesting: is clang formatting more understandable for you : )))

eshoguli (approved):
LGTM, thanks!

src/common/low_precision_transformations/tests/add_transformation.cpp
v-Golubev: In test instances definitely no) But I wanted to avoid additional changes in this PR (I mean changing formatting for all test instances in this file)

PR #17078, opened 2023-04-20
[GPU] Fix a bug of fusing eltwise sum post-op.

jade-cho:
+ When input of eltwise is full-tensor constant layer, use binary add instead of sum as post-op on oneDNN.

### Tickets:
 - *104840*, 105495

### DIFF
--- a/src/plugins/intel_gpu/src/graph/program_helpers.cpp
+++ b/src/plugins/intel_gpu/src/graph/program_helpers.cpp
@@ -114,6 +114,7 @@ add_fusing_type onednn_add_fusing_helpers::get_add_fusing_type(
             && p_layout.format == d_layout.format && p_layout.get_tensor() == d_layout.get_tensor()
             && p_layout.data_padding == d_layout.data_padding
             && dep_node.get_users().size() == 1
+            && !dep_node.is_constant()
             && !p_node.is_type<pooling>()) {
             return add_fusing_type::sum;
         } else if (p_layout.get_tensor() == d_layout.get_tensor()) {
--- a/src/plugins/intel_gpu/tests/fusions/convolution_fusion_test.cpp
+++ b/src/plugins/intel_gpu/tests/fusions/convolution_fusion_test.cpp
@@ -4378,6 +4378,64 @@ INSTANTIATE_TEST_SUITE_P(fusings_gpu, conv_before_permute_optimizing, ::testing:
     convolution_test_params{ CASE_CONV_FP16_PERMUTE_2, 3, 2, 4 },
 }));
 
+class EltwiseSumWithConstantFullTensorFusingTestOneDNN : public BaseFusingTest<convolution_eltw_sum_test_params> {
+public:
+    void execute(convolution_eltw_sum_test_params& p) {
+        if (!engine.get_device_info().supports_immad)
+            return;
+        auto input_prim = get_mem(get_weights_layout(p));
 
+        network network_not_fused(this->engine, this->topology_non_fused, cfg_not_fused);
+        network network_fused(this->engine, this->topology_fused, cfg_fused);
+        network_fused.set_input_data("input", input_prim);
+        network_not_fused.set_input_data("input", input_prim);
+
+        // Multiple executions of network to increase error of result.
+        // The output of constant layer will be changed through this iterations bigger.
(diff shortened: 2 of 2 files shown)


src/plugins/intel_gpu/src/graph/program_helpers.cpp
```diff
@@ -114,6 +114,7 @@ add_fusing_type onednn_add_fusing_helpers::get_add_fusing_type(
             && p_layout.format == d_layout.format && p_layout.get_tensor() == d_layout.get_tensor()
             && p_layout.data_padding == d_layout.data_padding
             && dep_node.get_users().size() == 1
+            && !dep_node.is_constant()
```
isanghao, line 117: What if the input node is not a constant, but will be a constant as the result of constant propagation? In that case, first execution of this function(before constant propagation) will return "not a constant". isn't it?

src/plugins/intel_gpu/tests/fusions/convolution_fusion_test.cpp
```diff
@@ -4378,6 +4378,74 @@ INSTANTIATE_TEST_SUITE_P(fusings_gpu, conv_before_permute_optimizing, ::testing:
...
+        auto info_fused = std::find_if(pi_fused.begin(), pi_fused.end(), [](primitive_info& p) -> bool {
+            if (p.original_id == "conv_prim")
+                return true;
+            return false;
+        });
+
+        if (info_fused != pi_fused.end()) {
+            std::cout << "kernel: " << info_fused->kernel_id << std::endl;
+            EXPECT_TRUE(info_fused->kernel_id.find("jit:ir") != std::string::npos);
+        }
```
isanghao, line 4410: what does this code test? I tried this code(without your patch) and it just succeeds when I increased tolerance..

src/plugins/intel_gpu/tests/fusions/convolution_fusion_test.cpp
```diff
@@ -4378,6 +4378,74 @@ INSTANTIATE_TEST_SUITE_P(fusings_gpu, conv_before_permute_optimizing, ::testing:
...
+class onednn_replace_full_tensor_sum_to_binary_add : public EltwiseSumWithConstantFullTensorFusingTestOneDNN {};
+TEST_P(onednn_replace_full_tensor_sum_to_binary_add, basic) {
+    auto p = GetParam();
+    if (engine.get_device_info().supports_immad)
+        p.expected_fused_primitives = p.expected_fused_primitives_onednn;
+
+    create_topologies(
+        data("src0", get_mem(get_input_layout(p))),
+        input_layout("input", get_weights_layout(p)),  // Input is weights.
+        data("eltwise_data", get_mem(layout{ p.eltw_type, p.eltw_format, p.out_shape })),
```
isanghao, line 4434: as mentioned above, what if we have additional layer between this `eltwise_data` and `sum`? For example, eltwise operation... In that case, is it recognized as constant well?

isanghao (changes requested):

src/plugins/intel_gpu/tests/fusions/convolution_fusion_test.cpp
jade-cho: Got it. It will be removed.

src/plugins/intel_gpu/src/graph/program_helpers.cpp
jade-cho: This function is used after constant propagation pass. So, there is no intermediate node between eltwise and constant.

src/plugins/intel_gpu/tests/fusions/convolution_fusion_test.cpp
jade-cho: At this line, network::execute() function is called multiple times in execute function of test class. Please see line 4393.

isanghao (approved):

hyunback (approved):

isanghao:
Performance was good: --this ../log/pr/2023-04-21--03-50.report --ref ../log/2023-04-19--13-00.report

Could you check graph diff on dgpu?

isanghao:
Performance was good
--this ../log/pr/2023-04-21--16-48.report --ref ../log/2023-04-20--09-00.report

PR #17081, opened 2023-04-20
[PyOV] Check for glibc version in python test

akuporos:
### Details:
 - *item1*
 - *...*

### Tickets:
 - 108465

### DIFF
--- a/src/bindings/python/tests_compatibility/test_inference_engine/test_IECore.py
+++ b/src/bindings/python/tests_compatibility/test_inference_engine/test_IECore.py
@@ -294,9 +294,10 @@ def test_load_network_release_gil(device):
 
 def test_nogil_safe(device):
     libc_name, libc_version = platform.libc_ver()
-    version = tuple(int(x) for x in libc_version.split('.'))
-    if libc_name == 'glibc' and version < (2, 34):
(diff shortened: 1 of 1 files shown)


ilya-lavrenov (approved):

src/bindings/python/tests_compatibility/test_inference_engine/test_IECore.py
```diff
@@ -293,6 +293,11 @@ def detect_long_gil_holds():
 
 
 def test_nogil_safe(device):
+    libc_name, libc_version = platform.libc_ver()
+    version = tuple(int(x) for x in libc_version.split('.'))
+    if libc_name == 'glibc' and version < (2, 34):
+        pytest.skip("There is an issue in glibc for elder version.")
```
p-wysocki, line 299:
```suggestion
        pytest.skip("There is an issue in glibc for older version.")
```
Just a nit, I don't really mind if it's changed.

p-wysocki (approved):

src/bindings/python/tests_compatibility/test_inference_engine/test_IECore.py
akuporos: updated

PR #17095, opened 2023-04-20
[ARM CPU] Fix smoke_if tests

allnes:
### Details:
 - *item1*
 - *...*

### Tickets:
 - CVS-108787

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/executors/acl/acl_eltwise.cpp
+++ b/src/plugins/intel_cpu/src/nodes/executors/acl/acl_eltwise.cpp
@@ -62,7 +62,7 @@ bool AclEltwiseExecutor::init(const EltwiseAttrs &eltwiseAttrs, const std::vecto
         auto dim_size = srcDescs[0]->getShape().getDims().size();
         auto mover = [&dim_size](TensorShape &_shape) {
             if (dim_size == 5) { std::swap(_shape[2], _shape[3]); }
-            std::swap(_shape[1], _shape[2]);
+            if (dim_size == 4) { std::swap(_shape[1], _shape[2]); }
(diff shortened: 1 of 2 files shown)


src/plugins/intel_cpu/src/nodes/executors/acl/acl_eltwise.cpp
```diff
@@ -62,7 +62,7 @@ bool AclEltwiseExecutor::init(const EltwiseAttrs &eltwiseAttrs, const std::vecto
         auto dim_size = srcDescs[0]->getShape().getDims().size();
         auto mover = [&dim_size](TensorShape &_shape) {
             if (dim_size == 5) { std::swap(_shape[2], _shape[3]); }
-            std::swap(_shape[1], _shape[2]);
+            if (dim_size == 4) { std::swap(_shape[1], _shape[2]); }
             std::swap(_shape[0], _shape[1]);
```
dmitry-gorokhov, line 66: This is not correct fix as long as breaks 5D case. Should be smt like that: 
```
if (dim_size > 4) { std::swap(_shape[2], _shape[3]); }
if (dim_size > 3) { std::swap(_shape[1], _shape[2]); }
if (dim_size > 2) { std::swap(_shape[0], _shape[1]); }
```

dmitry-gorokhov (changes requested):

src/plugins/intel_cpu/src/nodes/executors/acl/acl_eltwise.cpp
allnes: @dmitry-gorokhov done

dmitry-gorokhov (approved):

PR #17096, opened 2023-04-20
[LPT] FQ reference implementation reused in foldFakeQuantize function

v-Golubev:
### Tickets:
 - *CVS-108861*

### DIFF
--- a/src/common/low_precision_transformations/src/network_helper.cpp
+++ b/src/common/low_precision_transformations/src/network_helper.cpp
@@ -732,16 +732,20 @@ std::shared_ptr<Node> NetworkHelper::foldFakeQuantize(
         }
     }
 
-    const auto& original_et = fq->get_output_element_type(0);
-    OPENVINO_SUPPRESS_DEPRECATED_START
-    std::shared_ptr<Node> result = ov::get_constant_from_source(std::make_shared<ov::op::TypeRelaxed<opset1::FakeQuantize>>(*fq, element::f32));
-    OPENVINO_SUPPRESS_DEPRECATED_END
-    if (result != nullptr) {
+    if (ov::is_type<opset1::Constant>(fq->get_input_node_shared_ptr(0))) {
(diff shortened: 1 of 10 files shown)


src/common/low_precision_transformations/src/network_helper.cpp
```diff
@@ -762,99 +732,20 @@ std::shared_ptr<Node> NetworkHelper::foldFakeQuantize(
...
-        assert(constPShape.is_static());
-        const Shape constShape = constPShape.to_shape();
-
-        if (constShape.size() > 5lu) {
-            THROW_IE_LPT_EXCEPTION(*fq) << "Unexpected dimensions count " << constShape.size();
-        }
-        if (outChannelsShapeIndex != 0 && outChannelsShapeIndex != 1) {
-            THROW_IE_LPT_EXCEPTION(*fq) << "Unexpected outChannelsShapeIndex " << outChannelsShapeIndex;
+    if (ov::is_type<opset1::Constant>(fq->get_input_node_shared_ptr(0))) {
+        std::shared_ptr<Node> subgraph = std::make_shared<ov::op::TypeRelaxed<opset1::FakeQuantize>>(*fq, element::f32);
```
eshoguli, line 736: do we really need `TypeRelaxed` here if output precision is `fp32`?

eshoguli (approved):
LGTM, one minor comment

src/common/low_precision_transformations/src/network_helper.cpp
v-Golubev: TypeRelaxed::evaluate method uses fields `m_input_data_types`, `m_output_data_types` in internal logic. At the same time, TypeRelaxed constructor allow to create node with empty fields (we don't meet such nodes in real scenario but might meet them in tests). And evaluation for such node fails if the fields are empty.
So I create new node to be avoid potential issues.

PR #17098, opened 2023-04-20
[Pt FE]: aten::embedding_bag

eaidova:
### Details:
 - *support aten::embedding_bag*

### Tickets:
 - *ticket-id*


src/frontends/pytorch/src/op_table.cpp
```diff
@@ -28,6 +28,7 @@ OP_CONVERTER(translate_batch_norm);
 OP_CONVERTER(translate_bitwise_not);
 OP_CONVERTER(translate_cat);
 OP_CONVERTER(translate_clamp);
+OP_CONVERTER(translate_col2im);
```
mvafin, line 31:
```suggestion
```

src/frontends/pytorch/src/op/embedding_bag.cpp
```diff
@@ -0,0 +1,69 @@
...
+            auto per_sample_weight = context.get_input(6);
+            per_sample_weight = context.mark_node(std::make_shared<ov::op::v1::ConvertLike>(per_sample_weight, weight));
+            result = context.mark_node(std::make_shared<ov::op::v3::EmbeddingBagOffsetsSum>(weight,
+                                                                                            indices,
+                                                                                            offsets,
+                                                                                            zero,
+                                                                                            per_sample_weight));
+        }
+        // aten::embedding_bag returns a tuple of 4 elements: output, offset2bag, bag_size, max_indices.
+        // But the last three outputs are not used in torch.nn.EmbeddingBag or torch.nn.functional.embedding_bag.
```
mvafin, line 61: That is concerning, does it unpacks with `TupleUnpack` later? Maybe better way would be creating `TuplePack` fw node here?

eaidova: There is no TupleConstruct and TupleUnpack in model graph. operation returns 4 outputs, so it is aligned with pt.
I believe TupleConstruct used only in case if number of outputs can be unknown and depends on data  (e.g. for split op), but here number of outputs is constant, so no need in additional packaging

mvafin (approved):

PR #17102, opened 2023-04-21
extend PaddlePaddle elementwise broadcast type support

liubo-intel:
### Details:
 - *Extend PaddlePaddle elementwise broadcast type support: Add support ov broadcast for 'BroadcastType::PDPD' type. More tests for PDPD broadcast added (static and dynamic shapes), The documentation is modified for this kind of broadcast type testcase description*

### Tickets:
 - *93618*


liubo-intel:
Hi, @mitruska @jane-intel @ceciliapeng2011 : could you please help review this PR when you are free? Thx.

docs/ops/broadcast_rules.md
mitruska, line 1: Not only B to A, but also A to B broadcasting is allowed by current `broadcast_merge_into` implementation of PDPD mode, based on the test:
https://github.com/openvinotoolkit/openvino/blob/ede43f473de768c5428138f138754345b44039c4/src/core/tests/type_prop/arithmetic_ops.hpp#L253-L261

But reference implementation for binary autobroadcast seems to be not able to handle such case:
https://github.com/openvinotoolkit/openvino/blob/793bbb6ee2106edcbacf27ac50865be43c071d7d/src/core/reference/include/ngraph/runtime/reference/autobroadcast_binop.hpp#L210-L213

Would be great to have such case described in the spec as for numpy, whether it should be supported or not.
For example:
* ``A: Shape(8, 1, 6, 1)`` 
  ``B: Shape(   7, 1, 5) with axis = 1`` 
 
  ``Result: Shape(8, 7, 6, 5)``
  or
    ``Result: broadcast not possible due to dimensions mismatch, only B to A broadcast is supported for PDPD``

src/core/include/openvino/op/util/attr_types.hpp
```diff
@@ -155,8 +155,8 @@ std::ostream& operator<<(std::ostream& s, const TopKMode& type);
 
 /// \brief Implicit broadcast specification
 struct OPENVINO_API AutoBroadcastSpec {
-    AutoBroadcastSpec() : m_type(AutoBroadcastType::NONE), m_axis(0) {}
-    AutoBroadcastSpec(AutoBroadcastType type) : m_type(type), m_axis(0) {}
+    AutoBroadcastSpec() : m_type(AutoBroadcastType::NONE), m_axis(-1) {}
+    AutoBroadcastSpec(AutoBroadcastType type) : m_type(type), m_axis(-1) {}
```
mitruska, line 159: Is there any chance to break backward compatibility because of the default value change?

src/core/src/partial_shape.cpp
```diff
@@ -325,35 +325,39 @@ bool ov::PartialShape::broadcast_merge_into(PartialShape& dst,
...
             auto src_rank = src.rank().get_length();
-            // source rank can't be bigger than destination rank according to PDPD broadcast rule.
-            if (src_rank > dst_rank)
-                return false;
-            if (dst_rank == src_rank && dst.compatible(src))
-                return true;
 
             int64_t axis = autob.m_axis;
-            if (axis < -1) {
+            if (src_rank > dst_rank || axis < -1)
```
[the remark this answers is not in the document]
liubo-intel: done

docs/ops/broadcast_rules.md
liubo-intel: thanks for pointing this out. based on spec and reference implementation, looks like this time better to support  'only B to A' for PDPD(and can be extend to support A to B broadcast like numpy if requested in the future.)

src/core/include/openvino/op/util/attr_types.hpp
liubo-intel: yes, we indeed found some previous CI tests(some are in other components and plugins) failed if change this default value. But I still think using `-1` as default value making more sense for `BroadcastType::PDPD` based on its specs. what about using `-1` for PDPD, and keep `0` for another `BroadcastType`?
For the CI failed issues, I will try to fix them in this PR(maybe in next commit, so currently will convert this PR to draft) if no need to modify kernel implement of other components and plugins.

docs/ops/broadcast_rules.md
mitruska: Thank you for clarification. Then it should be tested anyway with ASSERT_THROW (or even better [OV_EXPECT_THROW](https://github.com/mitruska/openvino/blob/a0a6e1c141d1246607c79dfc32d5c49d87677d56/src/core/tests/type_prop/binary_elementwise.cpp#L163-L165)) to ensure that such case is handled.

mitruska (approved):

src/core/include/openvino/op/util/attr_types.hpp
mitruska: Actually the `axis` attribute is used only for `BroadcastType::PDPD`, so it shouldn't affect other broadcast types.
Anyway current solution lgtm.

jane-intel (approved):
Could you also check if transformations which are working with broadcasting are not confused with the same logic that we had before?

PR #17103, opened 2023-04-21
Review detectron generate proposal for shape inference aspects

praasz:
### Details:
- Review interval shape propagation.
- Review label propagation.
- Review template implementation of `shape_infer`.
- Add more unit tests for dynamic and static shapes.

### Tickets:
 - 97244
 - 97314
 - 97363
 - 97412


src/core/src/op/experimental_detectron_generate_proposals.cpp
```diff
@@ -45,17 +44,36 @@ bool op::v6::ExperimentalDetectronGenerateProposalsSingleImage::visit_attributes
...
+
+    for (size_t i = 0; i < input_size; ++i) {
+        NODE_VALIDATION_CHECK(this,
+                              element::Type::merge(out_et, out_et, get_input_element_type(i)) &&
+                                  (out_et.is_dynamic() || out_et.is_real()),
+                              "Input[",
+                              i,
+                              "] type '",
+                              get_input_element_type(i),
+                              "' is not floating point or not same as others inputs.");
```
mitruska, line 64: As the input element types were not validated before, it's always worth to ensure that it's backward compatible change.

mitruska (approved):

src/core/src/op/experimental_detectron_generate_proposals.cpp
praasz: Right now looks there is no issue, this can check can be easily set to less strict or removed if there will be any issue.

PR #17111, opened 2023-04-21
[CPU] Enable Python test test_infer_request.test_infer_mixed_values with bool for ARM

alvoron:
bool tests were skipped because legacy ARM plugin does not support bool: CVS-101726


jiwaszki (approved):
👌

ilya-lavrenov (approved):

PR #17115, opened 2023-04-21
Specify nonempty begin/end_mask for StridedSlice in translate_conv_2d…

mateusztabaka:
…_backprop_input_op

Fixes 'AssertionError: begin_mask is not defined for the node StridedSlice'

### DIFF
--- a/src/frontends/tensorflow_common/src/op/conv_2d_backprop.cpp
+++ b/src/frontends/tensorflow_common/src/op/conv_2d_backprop.cpp
@@ -90,8 +90,8 @@ OutputVector translate_conv_2d_backprop_input_op(const NodeContext& node) {
                                                    ss_begin,
                                                    ss_end,
                                                    ss_strides,
-                                                   std::vector<int64_t>{},
-                                                   std::vector<int64_t>{});
+                                                   std::vector<int64_t>{0},
+                                                   std::vector<int64_t>{0});
 
     auto conv_backprop = make_shared<ConvolutionBackpropData>(out_backprop,
                                                               filter,


src/frontends/tensorflow_common/src/op/conv_2d_backprop.cpp
```diff
@@ -90,8 +90,8 @@ OutputVector translate_conv_2d_backprop_input_op(const NodeContext& node) {
                                                    ss_begin,
                                                    ss_end,
                                                    ss_strides,
-                                                   std::vector<int64_t>{},
-                                                   std::vector<int64_t>{});
+                                                   std::vector<int64_t>{0},
+                                                   std::vector<int64_t>{0});
```
rkazants, line 94: why are you changing it? The specification allows to have empty attributes. May be the problem in the Core part that erases these attributes.

rkazants (changes requested):
Not needed change

src/frontends/tensorflow_common/src/op/conv_2d_backprop.cpp
mateusztabaka: I can see that spec says that begin/end_mask are required. I couldn't find it mentions that empty arrays are allowed.

rkazants: But there is also no constraints to have empty masks:) It does make sense to have non-empty mask, if there is at least one non-zero element. Otherwise, it means no dimensions are ignored.

Am I correct that the problem is that for some reason these attributes are absent in IR? What component removes it, if `begin_mask = {}`? I think you need to figure this out first.

Also, please share the ticket number and what kind of problem you are solving.

mateusztabaka: > Also, please share the ticket number and what kind of problem you are solving.

There is no ticket number, but I found this issue during calibrating a model with ConvolutionBackpropData - StridedSlice_extender complains about "AssertionError: begin_mask is not defined for the node StridedSlice'".

> Am I correct that the problem is that for some reason these attributes are absent in IR? What component removes it, if begin_mask = {}? I think you need to figure this out first.

No, the attribute is present in the IR, but it's empty. This is why StridedSlice_extender fires this assert. So no component removes this attribute.

rkazants: Well, the core part allows to create SS with empty attribute but the extender complains. The core part is aligned with the specification. So let us update the extender to make it to be aligned with the specification.
I see no issue in TF FE in this perspective.

mateusztabaka: Oh, I can see it's being done here https://github.com/openvinotoolkit/openvino/pull/17019.

PR #17118, opened 2023-04-21
Fix of groupconv_to_conv() in MO IR reader

popovaan:
Root cause analysis: 
groupconv_to_conv() uses incorrect methods to compare shapes which do not work for numpy.masked_array. 
groupconv_to_conv() disconnects reshape node while transforming GroupConvolution to Convolution. The disconnected reshape breaks Op.normalize_outputs() pass of MO IR reader which results in adding wrong Result node after reshape.

Solution: 
Use strict_compare_tensors() for shapes comparison.
Add graph.clean_up() after groupconv_to_conv().

Ticket: 89007

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Transformation tests

Documentation:
* [x]  User guide update

### DIFF
--- a/tools/mo/openvino/tools/mo/utils/ir_reader/layer_to_class.py
+++ b/tools/mo/openvino/tools/mo/utils/ir_reader/layer_to_class.py
@@ -8,7 +8,7 @@ import numpy as np
 
 from openvino.tools.mo.back.MaxPool import MaxPool
 from openvino.tools.mo.back.TopKNormalizer import TopKNormalizer
-from openvino.tools.mo.front.common.partial_infer.utils import int64_array
+from openvino.tools.mo.front.common.partial_infer.utils import int64_array, strict_compare_tensors
 from openvino.tools.mo.graph.graph import Graph, Node
 from openvino.tools.mo.ops.Cast import Cast
 from openvino.tools.mo.ops.GRU import GRU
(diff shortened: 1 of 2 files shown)


rkazants (approved):
Also, please add a test similar to https://github.com/openvinotoolkit/openvino/pull/17019 in the next PR.

tools/mo/openvino/tools/mo/utils/ir_reader/layer_to_class.py
```diff
@@ -195,12 +195,13 @@ def groupconv_to_conv(op: Node):
...
     elif weights_node.type == 'Reshape':
         # We remove reshape node added in ConvolutionWithGroupsResolver pass
-        assert weights_node.in_port(0).get_source().data.get_shape() == new_shape, \
+        assert strict_compare_tensors(weights_node.in_port(0).get_source().data.get_shape(), new_shape), \
             'Weight shape and calculated shape mismatch in GroupConv node {}.'.format(op.name)
         op.in_port(1).disconnect()
         # We use add_destination method here to support case with multiple destinations of source port
         weights_node.in_port(0).get_source().get_connection().add_destination(op.in_port(1))
         weights_node.in_port(0).disconnect()
+        op.graph.remove_node(weights_node.id)
```
mvafin, line 204: Is this really needed? Why can't this node be removed by graph.clean_up() at later stage?

tools/mo/unit_tests/mo/utils/ir_reader/layer_to_class_test.py
```diff
@@ -6,85 +6,66 @@
...
-                                 ('weights_data', 'conv'),
-                                 ('conv', 'conv_data'),
-                                 ('conv_data', 'add'),
-                                 ('add_const', 'add_const_data'),
-                                 ('add_const_data', 'add'),
-                                 ('add', 'add_data'),
-                                 ('add_data', 'result'),
-                                 ], nodes_with_edges_only=True)
+        for op in graph.get_op_nodes(type='GroupConvolution'):
+            groupconv_to_conv(op)
```
mvafin, line 60: How this is different from lines 69-70?

popovaan: This pass is not needed. Removed it, thanks.

mvafin (approved):

PR #17125, opened 2023-04-21
Fix of tf.GenericFunction conversion in convert_model()

popovaan:
Root cause analysis: 
tf.GenericFunction test was initially broken which led to improper testing of the conversion of tf.GenericFunction.

Solution:
Fixed tf.GenericFunction test, fixed conversion of tf.GenericFunction.
Added support of conversion of tf.Graph.

Ticket: 106247

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Transformation tests

Documentation:
* [x]  User guide update


tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
```diff
@@ -269,21 +269,14 @@ def create_tf_checkpoint(tmp_dir):
 def create_tf_function(temp_dir):
```
rkazants, line 269: I think it makes sense to have functions and tests for both `keras.Model` and `ConcreteFunction`

tools/mo/openvino/tools/mo/front/tf/loader.py
```diff
@@ -226,6 +226,14 @@ def tf_function(x):
 
         conc_func = tf_function.get_concrete_function(model_inputs)
         return freeze_tf2_concrete_function(model, conc_func, env_setup)
+    if isinstance(model, tf.types.experimental.GenericFunction):
+        env_setup = get_environment_setup("tf")
+
+        assert hasattr(model, "input_signature") and model.input_signature is not None, \
+            "'input_signature' needs to be set for model conversion."
```
rkazants, line 233: can we have a test for this case with `input_signature` when obtaining of `ConcreteFunction` from `GenericFunction` is not possible

tools/mo/openvino/tools/mo/front/tf/loader.py
```diff
@@ -226,6 +226,14 @@ def tf_function(x):
 
         conc_func = tf_function.get_concrete_function(model_inputs)
         return freeze_tf2_concrete_function(model, conc_func, env_setup)
+    if isinstance(model, tf.types.experimental.GenericFunction):
```
rkazants, line 229: It looks that `tf.types.experimental.GenericFunction` is introduced recently in `2.4`. Can we leave MO compatible with early TF versions please?

tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
popovaan: We have keras.Model test here: https://github.com/openvinotoolkit/openvino/blob/656d7fe380e8a5d8c245a47c2c0e41b56baddeb7/tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py#L40

tools/mo/openvino/tools/mo/front/tf/loader.py
popovaan: Done.

tools/mo/openvino/tools/mo/front/tf/loader.py
popovaan: I double checked it, `GenericFunction `and `ConcreteFunction `was introduced in 2.6. I added checks of TF version for these types.

rkazants: this is experimental type. May be it makes sense to check stable API. Is it possible? Please check

rkazants (approved):

PR #17127, opened 2023-04-21
Error info telemetry events in MO.

popovaan:
Description:

Added new telemetry events with following information:

- Error message, where any lines that contain filepath are removed.
- Failed transformation name and stage for legacy MO.

Ticket: 98739, 98737

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update


tools/mo/openvino/tools/mo/utils/telemetry_utils.py
```diff
@@ -98,6 +99,22 @@ def send_framework_info(framework: str):
     t.send_event('mo', 'framework', framework)
 
 
+def remove_path_lines(message):
```
rkazants, line 102: I see that this function is not used in MO

tools/mo/openvino/tools/mo/utils/class_registration.py
```diff
@@ -302,6 +307,8 @@ def apply_transform(graph: Graph, replacer_cls, **kwargs):
             for_graph_and_each_sub_graph_recursively(graph, lambda _: graph.check_shapes_consistency())
 
     except Error as err:
+        t = tm.Telemetry()
+        t.send_event('mo', 'error_info', "stage:{},transformation:{}".format(graph.stage, replacer_cls.__name__))
```
rkazants, line 311: why do we need to send it? Don't we duplicate the same one in `convert_impl.py`?

tools/mo/openvino/tools/mo/convert_impl.py
```diff
@@ -972,6 +972,10 @@ def _convert(cli_parser: argparse.ArgumentParser, framework, args, python_api_us
                 if not argv.use_legacy_frontend and framework_is_tf(args, argv) and not is_fallback:
                     print(get_try_legacy_fe_message())
 
+        msg = remove_path_lines(str(e.with_traceback(None))).strip()
+        if len(msg):
+            telemetry.send_event('mo', 'error_info', "error_message:'{}'".format(msg))
```
rkazants, line 977: I am afraid that we will duplicate telemetry messages sent in FE itself and here. For example, it can happen for unsupported translators.

akladiev:

akladiev:

PR #17132, opened 2023-04-22
Torchscript backend

suryasidd:
### Details:
 - Added torchscript backend to OpenVINO
 - Added a parameter use_ts_backend to enable layer tests to run on torch script backend
### Tickets:
 - *ticket-id*


tests/layer_tests/pytorch_tests/test_einsum.py
```diff
@@ -51,8 +51,8 @@ def forward(self, x):
     @pytest.mark.nightly
     @pytest.mark.precommit
     @pytest.mark.xfail(reason='OpenVINO CPU plugin does not support einsum diagonal')
-    def test_einsum_batch_diagonal(self, ie_device, precision, ir_version):
-        self._test(*self.create_model(), ie_device, precision, ir_version, dynamic_shapes=False)
+    def test_einsum_batch_diagonal(self, ie_device, precision, ir_version, use_ts_backend):
+        self._test(*self.create_model(), ie_device, precision, ir_version, dynamic_shapes=False, use_ts_backend=use_ts_backend)
```
slyalin, line 55: @mvafin, can you recommend a better way to extend signature to avoid such a massive change and to make it more future proofed solution, like kwargs or something like that. With the currently extended approach we will add another parameter, like `use_fx_decoder` in the future and it will be again a bulky change in all the tests.

It is also a bit surprising to see such parameters like `ir_version` and others. Can we avoid explicitly trampolining data that are not required for a test?

mvafin: Agree, we need to process such parameters as device, precision, version, etc as kwargs

mvafin:
Also it doesn't look like you added to precommit any of the tests. I don't think we need all of them in precommit, but several tests would be good to add I think.

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
```diff
@@ -0,0 +1,67 @@
...
+
+@register_backend
+@fake_tensor_unsupported
+def openvino(subgraph, example_inputs):
+    return ts_openvino(subgraph, example_inputs)
+
+
+def ts_openvino(subgraph, example_inputs):
+    try:
+        model = torch.jit.script(subgraph)
```
slyalin, line 23: If we always have example_inputs, why cannot we just default to `torch.jit.trace` all the time? `torch.jit.script` is legitimate in the standalone application because user may not provide example input, but torchdynamo looks to be always using tracing as an essential part of the infrastructure. I can guess that we may have issues with dynamic shapes if always rely on tracing, but even now when we are using scripting, later in the code we always setting static shapes based on example_inputs so from this perspective we don't utilize any opportunities related to dynamic shapes.

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
```diff
@@ -0,0 +1,67 @@
...
+        }
+        decoder = TorchScriptPythonDecoder(fr_model)
+
+        im = fe.load(decoder)
+        om = fe.convert(im)
+
+        for idx, input_data in enumerate(example_inputs):
+            om.inputs[idx].get_node().set_element_type(dtype_mapping[input_data.dtype])
+            om.inputs[idx].get_node().set_partial_shape(PartialShape(list(input_data.shape)))
+        om.validate_nodes_and_infer_types()
```
slyalin, line 47: Please bookmark for the future modification when `mo` core part -- `convert_model` -- becomes a part of OV runtime, just use a single call to `convert_model` instead this fragment. Now it is not possible because `convert_model` is a part of `openvino-dev`.

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
```diff
@@ -0,0 +1,67 @@
...
+
+        device = 'CPU'
+        if (os.getenv("OPENVINO_DEVICE") is not None):
```
slyalin, line 50: Can we consider more "specialized" name for that? I mean `OPENVINO_DEVICE` looks too global that can be mixed with other part of APIs, like default device in `compile_model`. Not a big deal, but if you have other options like `OPENVINO_TORCH_BACKEND_DEVICE` it is worth considering.

Another question is can we really avoid using such env variables and provide the device as a parameter of torch.compile with our backend? I don't know about technical possibility and not sure about convenience in all the scenarios, but it is more "local" solution that doesn't allow any wrong associations.

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
```diff
@@ -0,0 +1,67 @@
...
+            device = os.getenv("OPENVINO_DEVICE")
+            assert device in core.available_devices, "Specified device " + device + " is not in the list of OpenVINO Available Devices"
+
+        compiled_model = core.compile_model(om, device)
+
+        def _call(*args):
+            ov_inputs = [a.detach().cpu().numpy() for a in args]
+            try:
+                res = compiled_model(ov_inputs)
```
slyalin, line 59: Is it a correct assumption that it always use memory sharing with the original numpy object? @mvafin could you comment on that?

tests/layer_tests/pytorch_tests/test_einsum.py
slyalin: @suryasidd, you mentioned that it would be fixed. If you have issues with that, please share your experience.

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
suryasidd: As of now, we don't have the option to add any config parameters to torch.compile. It's only available for inductor backend

suryasidd: But we can use your suggestion for changing the device name

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
```diff
@@ -0,0 +1,67 @@
...
+            except Exception as e:
+                return subgraph.model.forward(*args)
+            result = [torch.from_numpy(res[out]) for out in compiled_model.outputs]
+            return result
+        return _call
+    except Exception as e:
+        print(f"Exception is {str(e)}")
+        return compile_fx(subgraph, example_inputs)
```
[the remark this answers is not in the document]
suryasidd: Yes, sergey if we use the default backend for torch.compile which is torch inductor, compile_fx is used underneath torch.compile

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
suryasidd: Yeah we can use jit.trace for now but we will have to switch it back once we enable dynamic shapes and control flow ops

tests/layer_tests/pytorch_tests/test_einsum.py
suryasidd: Added env variable for USE_TS_BACKEND

suryasidd:
> Also it doesn't look like you added to precommit any of the tests. I don't think we need all of them in precommit, but several tests would be good to add.

Added a few tests to precommit_ts_backend

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
suryasidd: Added a TODO with this comment

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
```diff
@@ -0,0 +1,67 @@
...
+            assert device in core.available_devices, "Specified device " + device + " is not in the list of OpenVINO Available Devices"
+
+        compiled_model = core.compile_model(om, device)
+
+        def _call(*args):
+            ov_inputs = [a.detach().cpu().numpy() for a in args]
+            try:
+                res = compiled_model(ov_inputs)
+            except Exception as e:
+                return subgraph.model.forward(*args)
```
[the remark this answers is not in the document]
suryasidd: Changed this

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
```diff
@@ -0,0 +1,80 @@
...
+'''
+    This is a preview feature in OpenVINO. Torchscript backend
+    enables users to compile PyTorch models using torch.compile
+    with OpenVINO as a target backend.
+
+    Sample usage:
+    This sample code loads resnet50 torchvision model and compiles it using torch dynamo.
+    We can then use this model for inference
+
+    model = torchvision.models.resnet50()
```
ynimmaga, line 23: @suryasidd, it will be better to mention that only two additional lines of code are needed and also add the import statement.

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
```diff
@@ -0,0 +1,80 @@
...
+from openvino.frontend import FrontEndManager
+from openvino.runtime import Core, Type, PartialShape
+from openvino.frontend.pytorch.decoder import TorchScriptPythonDecoder
+
+log = logging.getLogger(__name__)
+
+'''
+    This is a preview feature in OpenVINO. Torchscript backend
+    enables users to compile PyTorch models using torch.compile
+    with OpenVINO as a target backend.
```
ynimmaga, line 17: @suryasidd, it is not clear that this can be used in applications with PyTorch APIs directly. Perhaps adding "in PyTorch applications" or "in applications with PyTorch APIs" at the end of this sentence will make it more clear

slyalin:
@ynimmaga, @suryasidd, we need to move forward quickly. If we really want to merge before the release, then please remove Draft mark, retarget to the release branch and address the remaining issues and prepare a message for the committee to be able to merge to the release branch.

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
mvafin: Yes it does by default when it is possible (when tensor.numpy(force=True) return shared tensor)

mvafin (approved):

slyalin (approved):

mvafin:
build_jenkins

PR #17136, opened 2023-04-23
[GPU] Fix levit-128s accuracy issue

hyunback:
Fix wrong batch dims for fused eltwise of gemm, it make different batch dims between fused eltwise and gemm.

levit-128s ACC has 2 issues.
1. Wrong batch dims for fused eltwise of gemm.
    -> The issue is getting incorrect batch size of fused eltwise used by gemm, it causes to reduce eltwise dims. It is only reproduce in batch 1 and full tensor.
         The batch size in here means all of non spatial dims, but previous implementation was default batch dim role.
2. Reuse sum post operation buffer issue in case constant input.
    -> Fixed by https://github.com/openvinotoolkit/openvino/pull/17078
    
Add non broadcast gemm eltwise fusion unit test.

### Tickets:
 - *105495*


isanghao:
no perf issue --this ../log/pr/2023-04-24--12-49.report --ref ../log/2023-04-19--13-00.report

PR #17148, opened 2023-04-24
Fix C API test ie_core_register_plugins error

zhaixuejun1993:
### Details:
 - Triger plugin load after register in test

### Tickets:
 - 109148


src/bindings/c/tests/test_model_repo.hpp
```diff
@@ -12,6 +12,7 @@
 #include "openvino/core/visibility.hpp"
 #include "openvino/pass/manager.hpp"
 #include "openvino/util/file_util.hpp"
+#include "common_test_utils/file_utils.hpp"
```
ilyachur, line 15: Please fix code style

src/bindings/c/tests/test_model_repo.hpp
```diff
@@ -80,12 +68,21 @@ inline std::string generate_test_xml_file() {
 
     plugin_xml_file << "<ie>\n";
     plugin_xml_file << "    <plugins>\n";
-    plugin_xml_file << "        <plugin location=\"" << libraryname << "\" name=\"CUSTOM\">\n";
+    plugin_xml_file << "        <plugin location=\"";
+    plugin_xml_file << CommonTestUtils::getExecutableDirectory();
+    plugin_xml_file << ov::util::FileTraits<char>::file_separator;
+    plugin_xml_file << ov::util::FileTraits<char>::library_prefix();
+    plugin_xml_file << "mock_engine";
```
ilyachur, line 75: Can we add dependency on mock_engine for c tests in CMake?

src/bindings/c/tests/CMakeLists.txt
```diff
@@ -3,6 +3,8 @@
 #
 
 # OpenVINO Legacy C API test sample
+set(DEPENDENCIES mock_engine)
```
ilyachur, line 6: I think it is not correct command. You need to use `add_dependencies()` command

ilyachur: And do we still need to have dependencies on hw plugins below?

zhaixuejun1993: > add_dependencies()
Will correct this.


> And do we still need to have dependencies on hw plugins below?
I think u are right, the hw plugins dependencies will be removed

zhaixuejun1993:
Closed! Using the new PR replace. https://github.com/openvinotoolkit/openvino/pull/17800

PR #17155, opened 2023-04-24
[Spec] Update Pad-1 spec with negative pads info

mitruska:
### Details:
 - Update Pad-1 spec with negative pads info (already supported by reference implementation and CPU plugin)
 
### Tickets:
 - 105055


slyalin (changes requested):
Please make sure that such a change doesn't implicitly involve extending of the operation that is not backward compatible with older versions of all existing OV plugins. Please follow discussion in the ticket.

rkazants:
> Please make sure that such a change doesn't implicitly involve extending of the operation that is not backward compatible with older versions of all existing OV plugins. Please follow discussion in the ticket.

@mitruska, do we really have support of negative pads on all platforms GPU? Now we have some conformance pass-rate for these platforms and we change the requirements for the current version. This way we decrease pass-rate for these platforms.

mitruska:
Closing in favor of introducing new version of Pad:
- https://github.com/openvinotoolkit/openvino/pull/17510

PR #17157, opened 2023-04-24
[CPU] Fixed enforcebf16 condition for transformation pipeline

a-sidorova:
### Details:
 - *Added check for config parameter `inference_precision`. Before there was only `PluginConfigParams::KEY_ENFORCE_BF16`. It's not enough to know that `enforcebf16` will be really enabled. So some transformations which depend on `enforcebf16` flag, can work incorrectly.*

### Tickets:
 - *N/A*

### DIFF
--- a/src/tests/functional/plugin/shared/src/snippets/mha.cpp
+++ b/src/tests/functional/plugin/shared/src/snippets/mha.cpp
@@ -16,33 +16,50 @@ namespace snippets {
 std::string MHA::getTestCaseName(testing::TestParamInfo<ov::test::snippets::MHAParams> obj) {
     std::vector<ov::PartialShape> inputShapes;
     bool withMul;
+    ov::element::Type prc;
     std::string targetDevice;
     size_t num_nodes, num_subgraphs;
-    std::tie(inputShapes, withMul, num_nodes, num_subgraphs, targetDevice) = obj.param;
+    std::map<std::string, std::string> additionalConfig;
+    std::tie(inputShapes, withMul, prc, num_nodes, num_subgraphs, targetDevice, additionalConfig) = obj.param;
 
     std::ostringstream result;
     for (size_t i = 0; i < inputShapes.size(); ++i)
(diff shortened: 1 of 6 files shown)


a-sidorova:
@dmitry-gorokhov take a look please. This is about Snippets MHA tokenization when the execution is in `bf16`

dmitry-gorokhov:
@a-sidorova Is it possible to add some tests? At least snippets based

dmitry-gorokhov:
@EgorDuplensky Could you please take a look as well? As a fast fix looks good to me, but in future we migh need to refactor this. Maybe it would be a better solution to update plugin config with compile_model properties before transformation pipeline. What is your thoughts on that?

EgorDuplensky:
> @EgorDuplensky Could you please take a look as well? As a fast fix looks good to me, but in future we might need to refactor this. Maybe it would be a better solution to update plugin config with compile_model properties before transformation pipeline. What is your thoughts on that?

Yes, the idea was to merge engine and model configs before doing everything else.
Thus, we will avoid all these extra checks and variables web currently have before executing transformation pipeline.

src/tests/functional/plugin/shared/src/snippets/mha.cpp
```diff
@@ -92,16 +115,41 @@ void MHASelect::generate_inputs(const std::vector<ngraph::Shape>& targetInputSta
...
+
+    auto f = ov::test::snippets::MHAWOTransposeFunction(inputDynamicShapes);
+    function = f.getOriginal();
+
+    configuration.insert(additionalConfig.begin(), additionalConfig.end());
+
+    setInferenceType(prc);
+    inType = outType = prc;
+    if (prc == ov::element::bf16)
+        abs_threshold = 0.3;
```
maxnick, line 152: Please use a relative threshold. The value should be `0.01` and at least no more than `0.05`

maxnick: Agreed to adjust the input data generation in a separate PR.

maxnick (approved):
LGTM

PR #17166, opened 2023-04-24
[Dont_merge] Revert "Fixed compilation errors on Linux arm64 (#17138)"

peterchen-intel:
This reverts commit 11c3623ebbe698bd60575bc1806dece738e11de3.

### Details:
 - *Try if reverting this work through it is unlikely, but I have tried this commit 11c3623e again, it reproduced the issue and the one before didn't reproduce CVS-109159*

### Tickets:
 - *CVS-109159*


peterchen-intel:
build-linux-ubuntu18/144844/ reproduce the same.

PR #17167, opened 2023-04-24
DOCS remove deprecated options

pavel-esir:
### Details:
- remove from DOCs: --mean_file, mean_file_offsets, --data_type. They are already removed from cli_parser, and are left only in FAQ

### Tickets:
 - xxx-106415


docs/MO_DG/prepare_model/Model_Optimizer_FAQ.md
```diff
@@ -80,10 +80,10 @@ Q4. What does the message "Input image of shape is larger than mean image from f
 
 .. _question-5:
 
-Q5. What does the message "Mean file is empty" mean?
+Q5.
```
rkazants, line 83: Propose to say: `Deprecated question`. Also, if posssible, let us delete `**A:**` and `**Q5**` - don't think it is needed

pavel-esir: Agreed, i think we can even write: `This MO argument/option was deprecated and removed, question is not actual anymore`

pavel-esir: Removed **A** **Q**. Added `This MO argument/option was deprecated and removed, question is not actual anymore`.

Also removed questions 4 and 6 which also refer to deprecated mean_file and this questions are never referred in mo.

pavel-esir: generated HTML doesn't look nice
image

i will remove completely the item keeping numbers for other questions the same

pavel-esir: Done. Now the generated DOCs look the following 
image

docs/MO_DG/prepare_model/Model_Optimizer_FAQ.md
```diff
@@ -263,17 +234,15 @@ Q16. What does the message "Input shape is required to convert MXNet model. Plea
 
 .. _question-17:
 
-Q17. What does the message "Both --mean_file and mean_values are specified. Specify either mean file or mean values" mean?
 #####################################################################################################################################################
 
-**A:** The ``--mean_file`` and ``--mean_values`` options are two ways of specifying preprocessing for the input. However, they cannot be used together, as it would mean double subtraction and lead to ambiguity. Choose one of these options and pass it with the corresponding CLI option.
+This MO argument/option was deprecated and removed, question is not actual anymore.
```
tsavina, line 239: Are Q17-18 planned to be removed too? If not, let's maybe change the word actual to something like - This MO option/argument has been deprecated, the question is no longer relevant.

pavel-esir: Q17-18 should also be removed. this green lines are left by mistake, i will remove them as well

pavel-esir: removed these lines

rkazants (approved):

PR #17183, opened 2023-04-25
[GPU] Fix i8/u8 representation error for clamp due to overflow

andrew-k-park:
### Details:
 - jitter saves all compile time constants as `float`. So if the activation function is clamp and the output data type size is 1, values for min/max are converted to output type(`i8`or`u8`). At this time, `i8` or `u8` representation error occurs due to overflow.
 - Fix jitter to clamp original min/max to output type range and adjust them to input value

### Tickets:
 - 100877

### DIFF
--- a/src/plugins/intel_gpu/src/kernel_selector/jitter.cpp
+++ b/src/plugins/intel_gpu/src/kernel_selector/jitter.cpp
@@ -1520,8 +1520,23 @@ JitConstants MakeActivationJitConstants(std::vector<kernel_selector::base_activa
     std::string activation_params = "";
     for (size_t i = 0; i < params.size(); i++) {
         std::string activation_suffix = suffix + "_" + toCodeString(i);
-        auto jitConstants = JitConstants{MakeJitConstant("NL_M" + activation_suffix, params[i].m),
-                                         MakeJitConstant("NL_N" + activation_suffix, params[i].n)};
+        std::string nl_m = toCodeString(params[i].m);
+        std::string nl_n = toCodeString(params[i].n);
+        if (params[i].function == ActivationFunction::CLAMP) {
+            if (out_dt == Datatype::INT8) {
+                float max_i8 = 127.f;
+                float min_i8 = -128.f;
+                nl_m = "fmax(" + nl_m + ", " + toCodeString(min_i8) + ")";
+                nl_n = "fmin(" + nl_n + ", " + toCodeString(max_i8) + ")";
+            } else if (out_dt == Datatype::UINT8) {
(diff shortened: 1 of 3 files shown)


e-ddykim (approved):
Looks good to me

src/plugins/intel_gpu/src/kernel_selector/jitter.cpp
```diff
@@ -1520,8 +1520,23 @@ JitConstants MakeActivationJitConstants(std::vector<kernel_selector::base_activa
...
     for (size_t i = 0; i < params.size(); i++) {
         std::string activation_suffix = suffix + "_" + toCodeString(i);
-        auto jitConstants = JitConstants{MakeJitConstant("NL_M" + activation_suffix, params[i].m),
-                                         MakeJitConstant("NL_N" + activation_suffix, params[i].n)};
+        std::string nl_m = toCodeString(params[i].m);
+        std::string nl_n = toCodeString(params[i].n);
+        if (params[i].function == ActivationFunction::CLAMP) {
+            if (out_dt == Datatype::INT8) {
+                float max_i8 = 127.f;
+                float min_i8 = -128.f;
```
yeonbok, line 1528: INT_VAL_MAX, UINT_VAL_MAX, INT_VAL_MIN, UINT_VAL_MIN will work for this?

src/plugins/intel_gpu/src/kernel_selector/jitter.cpp
```diff
@@ -1520,8 +1520,23 @@ JitConstants MakeActivationJitConstants(std::vector<kernel_selector::base_activa
...
+                nl_m = "fmax(" + nl_m + ", " + toCodeString(min_i8) + ")";
```
vladimir-paramuzov, line 1529: What's the reason to include that in ocl code?
Why not just
```cpp
nl_m = toCodeString(std::max(params[i].m, max_i8));
```
?

src/plugins/intel_gpu/src/kernel_selector/jitter.cpp
andrew-k-park: Replaced them to `SCHAR_MIN`, `SCHAR_MAX`, `UCHAR_MAX` macros

src/plugins/intel_gpu/src/kernel_selector/jitter.cpp
andrew-k-park: Fixed to apply clamp with output type range in advance without being included in ocl code as your suggestion. Thanks!

yeonbok (approved):

PR #17191, opened 2023-04-25
[CPU][coverity] fix coverity issue in  concat and graph_optimizer

tiger100256-hu:
### Details:
 - *fix coverity issue in  concat*
 - *fix coverity issue in graph_optimizer*

### Tickets:
 - *108975*
 - *108979*


src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2259,6 +2259,9 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
             // transposeNode support blocked input & non-blocked output, in the case, the reorder
             // cannot be optimized
             auto* transposeNode = dynamic_cast<Transpose*>(parentNode.get());
+            if (transposeNode == nullptr) {
+                IE_THROW() << "[CPU] parent node is not transpose node";
```
maxnick, line 2263:
```suggestion
                IE_THROW() << "[CPU] parent node of type:" << parentNode->getTypeStr() << " with name: " << parentNode->getName() << " is not a transpose node";
```

maxnick: Now it looks like double throw call, the first one should be removed I guess.

maxnick (approved):

PR #17192, opened 2023-04-25
[Gna][coverity] fixes for issue type AUTO_CAUSES_COPY

tadamowicz:
### Details:
 - CID: 1491505, 1491595, 1502494, 1502500, 1504698, 1504769, 1507058

### Tickets:
 - 108942

### DIFF
--- a/src/plugins/intel_gna/legacy/include/legacy/graph_tools.hpp
+++ b/src/plugins/intel_gna/legacy/include/legacy/graph_tools.hpp
@@ -612,7 +612,7 @@ inline CNNNetwork CNNNetCopy(const CNNNetwork& input, const Copier& cp) {
     InputsDataMap inputsInfo = input.getInputsInfo();
     std::set<DataPtr> insDatas;
     for (auto&& info : inputsInfo) {
-        for (auto& secondLayer : getInputTo(info.second->getInputData())) {
+        for (const auto& secondLayer : getInputTo(info.second->getInputData())) {
             auto secondLayerNew = oldToNewLayers[secondLayer.second.get()];
             InputInfo::Ptr infoNew = std::make_shared<InputInfo>();
             infoNew->setInputData(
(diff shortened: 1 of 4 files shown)


mkwapul (approved):

marcinkusm (approved):

src/plugins/intel_gna/legacy/include/legacy/graph_tools.hpp
```diff
@@ -192,7 +192,7 @@ inline void UnorderedDFS(std::unordered_set<CNNLayer*>& visited,
 
         // visit children
         for (auto& od : cnnLayer->outData) {
-            for (auto nl : getInputTo(od)) {
+            for (auto& nl : getInputTo(od)) {
```
dorloff, line 195: Isn't it better to use const auto& in all occurences as we are only reading the values, not modifying them?

tadamowicz: Good catch

src/plugins/intel_gna/legacy/include/legacy/graph_tools.hpp
```diff
@@ -612,7 +612,7 @@ inline CNNNetwork CNNNetCopy(const CNNNetwork& input, const Copier& cp) {
     InputsDataMap inputsInfo = input.getInputsInfo();
     std::set<DataPtr> insDatas;
     for (auto&& info : inputsInfo) {
-        for (auto secondLayer : getInputTo(info.second->getInputData())) {
+        for (auto& secondLayer : getInputTo(info.second->getInputData())) {
```
dorloff, line 615: Shouldn't it be const auto, too?

src/plugins/intel_gna/src/optimizer/gna_pass_manager.cpp
```diff
@@ -1024,7 +1024,7 @@ void InsertCopyLayerPass::run() {
             LayerInfo(l).isSplit()) {
             std::vector<FuncChildrenInfo> copy_insertion_tuples;
             std::vector<FuncChildrenInfo> delayed_copy_insertion_tuples;
-            for (auto output : l->outData) {
+            for (auto& output : l->outData) {
```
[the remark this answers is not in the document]
tadamowicz: yes, it should. I am using github web interface for this change but I do not know I missed those places. Thanks Denis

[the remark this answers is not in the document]
mkwapul: wouldn't const be deduced if needed in this case?

PR #17197, opened 2023-04-25
[Shape inference] Pooling: Dimension div fix

v-Golubev:
### Details:
 - *Dimension div produces an incorrect result in the case of dynamic shapes with range values when dimension max value is  `dim::inf_bound.` This PR fixes the issue*

### Tickets:
 - *CVS-109316*


jane-intel (commented):
I approve the change

v-Golubev:
@jane-intel Could you please review my [last commit](https://github.com/openvinotoolkit/openvino/pull/17197/commits/478c2e4323b0b20b89ff096b899d233d8eba2a9c): I had to change 2 convolution test instances related to label propagation. To my mind, the current behavior is correct and label mustn't be propagated. But I might have missed something. Thanks

jane-intel (approved):
@praasz could you review too? from my perspective -- all good

PR #17199, opened 2023-04-25
[GPU] Fix layout propagation logic

vladimir-paramuzov:
### Details:
- Adjust layout optimization logic to avoid unwanted rank changes

### Tickets:
 - *108740*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
@@ -512,7 +512,9 @@ void minimize_local_reorders(program& p, std::map<program_node*, format::type>& 
             continue;
 
         for (auto new_fmt : local_formats) {
-            if (fmt_map.at(node) != format::any && format::dimension(fmt_map.at(node)) != format::dimension(new_fmt))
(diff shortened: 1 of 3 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_inputs.cpp
```diff
@@ -512,7 +512,9 @@ void minimize_local_reorders(program& p, std::map<program_node*, format::type>&
             continue;
 
         for (auto new_fmt : local_formats) {
-            if (fmt_map.at(node) != format::any && format::dimension(fmt_map.at(node)) != format::dimension(new_fmt))
+            // Avoid setting of formats which will require transform from higher rank to smaller one which requires dimension squeeze
+            // TODO: Needs to be updated once we improve layout assignment logic
+            if (fmt_map.at(node) != format::any && format::dimension(fmt_map.at(node)) > format::dimension(new_fmt))
```
yeonbok, line 517: Won't this be problematic for dynamic shape? Shouldn't we add condition for (allow_new_shape_infer == false)?

vladimir-paramuzov: I don't think that we should disable it for new shape infer path as format selection logic is pretty much same

sshlyapn (approved):

PR #17207, opened 2023-04-25
use input parameter for building example_inputs

eaidova:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*

### DIFF
--- a/tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
+++ b/tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
@@ -101,7 +101,6 @@ def prepare_torch_inputs(example_inputs, input_shape, input_info=None, allow_non
     elif input_info is not None or input_shape is not None:
         input_info = input_to_input_cut_info(input_info) or []
         input_shape_to_input_cut_info(input_shape, input_info)
-        print(len(input_info))
         inputs = []
         for inp in input_info:
             shape = inp.shape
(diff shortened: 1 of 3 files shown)


tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
```diff
@@ -71,24 +98,35 @@ def prepare_torch_inputs(example_inputs, input_shape, allow_none=False):
                 inputs[name] = to_torch_tensor(tensor)
         else:
             inputs = to_torch_tensor(inputs)
-    elif input_shape is not None:
+    elif input_info is not None or input_shape is not None:
+        input_info = input_to_input_cut_info(input_info) or []
+        input_shape_to_input_cut_info(input_shape, input_info)
+        print(len(input_info))
```
eaidova, line 104:
```suggestion
```

tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
```diff
@@ -39,18 +41,43 @@ def to_torch_tensor(tensor):
...
+        "f32": torch.float32,
+        "f64": torch.float64,
+        "i8": torch.int8,
+        "i16":torch.int16,
+        "i32": torch.int32,
+        "i64": torch.int64,
+        "u8": torch.uint8,
+    }
+    if dtype is None:
+        return torch.float
```
slyalin, line 65: Convenient enough? For example if an LM is being converted what kind of error I get if dtype is not specified?

eaidova: it depends from model and ability to script it (in case if scripting works, we just ignore example inputs), if not for bert it will be: 
RuntimeError: Expected tensor for argument #1 'indices' to have one of the following scalar types: Long, Int; but got torch.FloatTensor instead (while checking arguments for embedding)

Do you mean we should provide some warning that the default type selected is torch.float32, please specify directly if different type required?

mvafin (approved):

PR #17210, opened 2023-04-25
[CPU] Deconv dummy shape generation fix

v-Golubev:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/deconv.cpp
+++ b/src/plugins/intel_cpu/src/nodes/deconv.cpp
@@ -353,13 +353,20 @@ std::pair<VectorDims, VectorDims> Deconvolution::makeDummyInOutShape() {
             ov::CoordinateDiff pe = autoPad ? ov::CoordinateDiff(paddingR.size(), 0) : paddingR;
 
             const auto& origInDims = getInputShapeAtPort(0).getDims();
+            const auto& origInMinDims = getInputShapeAtPort(0).getMinDims();
+            const auto& origInMaxDims = getInputShapeAtPort(0).getMaxDims();
             const auto& weightDims = getWeightDims();
             const size_t wghOffset = getAlgorithm() == Algorithm::DeconvolutionGrouped ? 1 : 0;
             for (size_t i = 0; i < inputDims.size() - 2; i++) {
                 if (origInDims[2 + i] == Shape::UNDEFINED_DIM) {
-                    inputDims[2 + i] = ((lastOutputSpatialDims[i] - (dilation[i] + 1) *
-                                        (weightDims[wghOffset + 2 + i] - 1) - 1 + pb[i] + pe[i] - outputPadding[i])) /
+                    inputDims[2 + i] = (lastOutputSpatialDims[i] - (dilation[i] + 1) *
+                                        (weightDims[wghOffset + 2 + i] - 1) - 1 + pb[i] + pe[i] - outputPadding[i]) /
                                         stride[i] + 1;
+                    // WA: if deconv output shape is provided in a separate input and autoPad is true,
+                    // paddings must be calculated based on input and output shapes (per specification).
+                    // However, the inputDims are unknown so the paddings cannot be calculated.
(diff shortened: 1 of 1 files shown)


v-Golubev:
@maxnick could you please take a look? Thanks

src/plugins/intel_cpu/src/nodes/deconv.cpp
```diff
@@ -349,16 +349,46 @@ std::pair<VectorDims, VectorDims> Deconvolution::makeDummyInOutShape() {
...
+                VectorDims paddings(paddingL.size());
+                if (!autoPad) {
+                    for (size_t i = 0; i < paddings.size(); ++i) {
+                        paddings[i] = paddingL[i] + paddingR[i];
+                    }
+                } else {
+                    for (size_t i = 0; i < origInDims.size() - 2; i++) {
+                        if (origInDims[i + 2] == Shape::UNDEFINED_DIM &&
+                            (origInMinDims[i + 2] != 0 || origInMaxDims[i + 2] != Shape::UNDEFINED_DIM)) {
+                            // if input shape is dynamic and boundable, paddings should be computed basing on the following limitations:
```
maxnick, line 369:
```suggestion
                            // if input shape is dynamic and bounded, paddings should be computed basing on the following limitations:
```

src/plugins/intel_cpu/src/nodes/deconv.cpp
```diff
@@ -349,16 +349,46 @@ std::pair<VectorDims, VectorDims> Deconvolution::makeDummyInOutShape() {
...
+                            // 1. paddings must be non negative
```
maxnick, line 370:
```suggestion
                            // 1. paddings must not be negative
```

src/plugins/intel_cpu/src/nodes/deconv.cpp
```diff
@@ -349,16 +349,46 @@ std::pair<VectorDims, VectorDims> Deconvolution::makeDummyInOutShape() {
...
+                            // 2. paddings must be between min and max posible values of the corresponding input dimension
```
maxnick, line 371:
```suggestion
                            // 2. the result padding must have such a value to keep the dummy dimensions inside the predefined interval 
```

src/plugins/intel_cpu/src/nodes/deconv.cpp
```diff
@@ -349,16 +349,46 @@ std::pair<VectorDims, VectorDims> Deconvolution::makeDummyInOutShape() {
...
+                            }
+                            auto lower_bound = stride[i] * static_cast<int32_t>(origInMinDims[i + 2] - 1) - c1;
+                            if (lower_bound > 0) {
+                                paddings[i] = lower_bound;
+                            }
+                        }
+                    }
+                }
+                return paddings;
+            }();
```
maxnick, line 386: It seems that it does not make sense to create this lambda. Since it is called immediately and is not used anywhere else. Suggest inline the lambda body right here.

maxnick (changes requested):

src/plugins/intel_cpu/src/nodes/deconv.cpp
v-Golubev: Suggestion applied

src/plugins/intel_cpu/src/nodes/deconv.cpp
v-Golubev: Suggestion applied

src/plugins/intel_cpu/src/nodes/deconv.cpp
v-Golubev: Done, thanks

src/plugins/intel_cpu/src/nodes/deconv.cpp
v-Golubev: Suggestion applied

maxnick (approved):

PR #17219, opened 2023-04-26
[CPU][NGraph]Enable node scatter_nd_update support negative indices.

xipingyan:
### Details:
 - *Enable node `scatter_nd_update` support negative indices.*
 - *`scatter_elements_update` and `scatter_update` are not updated.*
 - *[Doc](https://docs.openvino.ai/latest/openvino_docs_ops_movement_ScatterNDUpdate_3.html) is not updated.*
 - *ONNX, Pytorch support negative indices, tf doesn't support.*

### Tickets:
 - *106939*


src/plugins/intel_cpu/src/nodes/scatter_update.cpp
```diff
@@ -418,7 +418,11 @@ void ScatterUpdate::scatterNDUpdate(uint8_t *indices, uint8_t *update, uint8_t *
         size_t indicesOffset = tupleIdx * k;
         size_t dstOffset = 0;
         for (int i = 0; i < k; i++) {
-            size_t idxValue = getIndicesValue(indices, indicesOffset + i);
+            int64_t idxValue = getIndicesValue(indices, indicesOffset + i);
+            if (idxValue < 0) {
+                // Negative value for indices means counting backwards from the end.
+                idxValue += srcDataDim[0];
```
chenhu-wang, line 424:
```suggestion
                idxValue += srcDataDim[i];
```
negative value conversion should be applied to current axis.

chenhu-wang (changes requested):

src/core/reference/include/ngraph/runtime/reference/scatter_nd_update.hpp
```diff
@@ -35,9 +35,17 @@ void scatterNdUpdate(const dataType* const inputData,
         return padding;
     }();
 
+    // Negative value for indices means counting backwards from the end.
+    std::vector<indicesType> indicesCopy(indices, indices + shape_size(indicesShape));
+    for (size_t i = 0; i < indicesCopy.size(); i++) {
+        if (indicesCopy[i] < 0) {
+            indicesCopy[i] += static_cast<indicesType>(dataShape[0]);
```
chenhu-wang, line 42:
```suggestion
            indicesCopy[i] += static_cast<indicesType>(dataShape[i]);
```

chenhu-wang (changes requested):

src/plugins/intel_cpu/tests/functional/single_layer_tests/scatter_ND_update.cpp
```diff
@@ -152,6 +152,14 @@ const std::vector<ScatterNDUpdateLayerParams> scatterParams = {
         },
         IndicesValues{ 0, 1, 1, 2, 2, 2 }
     },
+    ScatterNDUpdateLayerParams{
+        ScatterNDUpdateShapes{
+            {{{3, 10}, {4, 11}, {3, 9}, {8, 15}}, {{ 10, 9, 9, 11 }, { 7, 5, 3, 12 }, { 3, 4, 9, 8 }}},
+            {{2, 3}, {{2, 3}, {2, 3}, {2, 3}}},
+            {{{2, 4}, -1}, {{2, 11}, {2, 12}, {2, 8}}}
+        },
+        IndicesValues{ 0, 1, 1, 2, -2, -1 }
```
chenhu-wang, line 161:
```suggestion
        IndicesValues{ -1, -1, -1, -2, -2, -2 }
```
let's fully cover negative index, as positive values already tested in previous case.

src/core/reference/include/ngraph/runtime/reference/scatter_nd_update.hpp
```diff
@@ -35,10 +35,22 @@ void scatterNdUpdate(const dataType* const inputData,
...
         const auto coord = span(indices_coord, indicesShape.back());
+
+        // Negative value for indices means counting backwards from the end.
+        int j = 0;
+        for (auto& c : coord) {
+            if (c < 0) {
+                c += static_cast<indicesType>(dataShape[j]);
+                j++;
+            }
+        }
```
chenhu-wang, line 52:
```suggestion
        for (auto& c : coord) {
            if (c < 0) {
                c += static_cast<indicesType>(dataShape[j]);
            }
            j++;
        }
```

chenhu-wang (changes requested):

chenhu-wang (approved):
LGTM for CPU plugin.

akladiev:

xipingyan:
@yuxu42 @usstq do you have new comments? if no, could you help merge ?

yuxu42:
@dmitry-gorokhov Could you please help review? Thanks!

dmitry-gorokhov:
@tomdol @mitruska Current specification clearly says only positive indexes are supported. However based on investigation from the ticket such cases can appear after conversion from ONNX. So 2 questions:
1. Should we update the specification (new op or just spec fix)? 
2. Should we update other "Scatter" operations?

dmitry-gorokhov:
@xipingyan Could you please clarify the performance impact from that fix? Like numbers on some syntetic test showing negative indices handling performance penalty (if any).

tomdol:
> @tomdol @mitruska Current specification clearly says only positive indices are supported. However based on investigation from the ticket such cases can appear after conversion from ONNX. So 2 questions:
> 
> 1. Should we update the specification (new op or just spec fix)?
> 2. Should we update other "Scatter" operations?

@dmitry-gorokhov the new version of ScatterNDUpdate is planned for the upcoming release and it will cover the negative values.

dmitry-gorokhov:
> > @tomdol @mitruska Current specification clearly says only positive indices are supported. However based on investigation from the ticket such cases can appear after conversion from ONNX. So 2 questions:
> > 
> > 1. Should we update the specification (new op or just spec fix)?
> > 2. Should we update other "Scatter" operations?
> 
> @dmitry-gorokhov the new version of ScatterNDUpdate is planned for the upcoming release and it will cover the negative values.

Thanks, @tomdol.
@xipingyan Lets clarify on performance questions and merge CPU part only with disabled tests (which can be enabled after ov opset part is fully implemented).

nshchego:
ScatterNDUpdate_12 also includes `reduction` attribute. Is it going to be supported in plugin or via transformation?

nshchego:
Why tests were completely removed? They should be disabled in file [skip_tests_config.cpp](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp)

xipingyan:
> ScatterNDUpdate_12 also includes `reduction` attribute. Is it going to be supported in plugin or via transformation?

My understanding is that there is no `reduction` attribute, our scatterndupdate is different with https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_.html#torch.Tensor.scatter_.
Maybe the key word "update" mean only replace value. Don't include other operation.

xipingyan:
Add performance data:
**CPU:** Intel(R) Xeon(R) Gold 6346 CPU @ 3.10GHz, (processor number: 64)
**OS:** Ubuntu 20.04.2 LTS
Test **ScatterNDUpdateLayerCPUTest** 60 iters, only statistic min value of ScatterNDUpdate.
Original version: 0.004534 ms
New version:     0.004661 ms

xipingyan:
hi @maxnick @chenhu-wang 
All CI test cases passed finally, could you help merge it?

maxnick:
@ilyachur , could you please review changes to the ngraph reference?

xipingyan:
> @ilyachur , could you please review changes to the ngraph reference?

hi @maxnick @ilyachur , @dmitry-gorokhov had advised disable the corresponding functional test. In this PR, we just fix customer's problem. If you think ngraph changes is not necessary, I can remove this part codes. BTW, new implementation is no effect on old implementation. 

New Spec is working in progress, please see: https://github.com/openvinotoolkit/openvino/pull/18074.
I hope we can merge it 2023.1, because customer is waiting this fix.

src/core/reference/include/ngraph/runtime/reference/scatter_nd_update.hpp
```diff
@@ -35,10 +35,22 @@ void scatterNdUpdate(const dataType* const inputData,
         return padding;
     }();
 
+    std::vector<indicesType> indicesCopy(indices, indices + shape_size(indicesShape));
```
ilyachur, line 38: Why do you need a copy of indices?
I don't see that indices are changing...

dmitry-gorokhov (approved):
CPU part LGTM. Lets get approval for reference impl as well

src/core/reference/include/ngraph/runtime/reference/scatter_nd_update.hpp
xipingyan: hi @ilyachur , here I need to change negative indices to positive.
I have to make copy, because I can't update originial value of indices, it will araise crash.
https://github.com/openvinotoolkit/openvino/pull/17219/files/ffc8fed0b0b4e80ff0a0d6243aa74d6f63860376#diff-7c7092e635ee31af64501f9e1aaef6ddd49adafe3f40669c2288428be0835ebbR49 **Line49** will update value of indice, it pass pointer, from `indicesCopy` to `coord`

mitruska (approved):
From the opset point of view, new version of ScatterNDUpdate op is no longer planned for 2023.1 (Opset12), but it still makes sense to merge negative indices support in CPU, and simply resolve the issue.

The changes added to reference implementation are backward compatible,
and will be needed for new version in the future anyway, so I'm okay with the reference part as well. 

But to rely on comparison with the reference implementation, corresponding tests with negative indices and expected output values should be added here: [src/plugins/template/tests/functional/op_reference/scatter_nd_update.cpp#L78](https://github.com/openvinotoolkit/openvino/blob/cd02804d5b8d65473a247dd82b130d0c1bf69f75/src/plugins/template/tests/functional/op_reference/scatter_nd_update.cpp#L78)

ilyachur (approved):
Ok, this LGTM

dmitry-gorokhov (approved):

PR #17220, opened 2023-04-26
Reduce the binary size of transformation lib

itikhono:
### Details:
each transformation adds ~ 20mb to the inference_engine_transformations_obj.lib. Z7 -> Zi flag decreases the lib size by~40mb only.

In total we have over 200 transformations. Size of the lib ~ sum of each transformation .obj (200 * 20mb = 4gb+, that exceeds the limit). We didn't find the way how we can increase this limit.
Another finding is if we delete the transformation body and even class declaration and definition and leave the empty file with opset and graph_rewrite includes, e.g.

#include <openvino/pass/graph_rewrite.hpp> (+11mb)

#include <openvino/opsets/opset5.hpp> (+ 6mb)

the .obj file takes 17mb+. 

So the current solution is to use the concreate op versions (ov::op::v1:: ) instead of opset5:: and delete the opset5 (6, 7, n) includes.
In this PR the approach (op versions ov::op::vN instead of opsetN:: ) was applied only for the part of transformations to reduce the risks for this Release. 

### Tickets:
 - *91401*

### DIFF
--- a/src/common/transformations/src/transformations/transpose_sinking/ts_reduction.cpp
+++ b/src/common/transformations/src/transformations/transpose_sinking/ts_reduction.cpp
@@ -9,10 +9,10 @@
 
 #include "itt.hpp"
 #include "openvino/core/validation_util.hpp"
+#include "openvino/op/constant.hpp"
+#include "openvino/op/transpose.hpp"
 #include "openvino/op/util/arithmetic_reductions_keep_dims.hpp"
 #include "openvino/op/util/logical_reduction_keep_dims.hpp"
-#include "openvino/op/transpose.hpp"
-#include "openvino/op/constant.hpp"
 #include "openvino/pass/pattern/op/wrap_type.hpp"
 #include "transformations/rt_info/transpose_sinking_attr.hpp"
 #include "transformations/transpose_sinking/ts_utils.hpp"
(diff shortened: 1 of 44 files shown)


ilya-lavrenov (approved):

src/common/transformations/src/transformations/transpose_sinking/ts_reduction.cpp
```diff
@@ -42,21 +42,21 @@ bool get_keep_dims(const std::shared_ptr<Node>& main_node) {
 TSReductionForward::TSReductionForward() {
     MATCHER_SCOPE(TSReductionForward);
 
-    auto transpose_label = wrap_type<Transpose>({any_input(), wrap_type<Constant>()});
+    auto transpose_label = wrap_type<ov::op::v1::Transpose>({any_input(), wrap_type<ov::op::v0::Constant>()});
```
t-jankowski, line 45: _A general comment which applies elsewhere_
Adding below piece of code at the beginning of file or function would avoid such changes in this file. Just an idea for further consideration.
```
using ov::op::v0::Constant;
using ov::op::v1::Transpose;
```

t-jankowski (approved):

PR #11605, opened 2022-04-29
Proofreading MO Guide

sgolebiewski-intel:
Stylistic and grammar corrections. Fixing broken links.

PR #17222, opened 2023-04-26
Fix faq numbers

pavel-esir:
### Details:
 - Most of the time when `refer_to_faq_msg` is called in MO it was referred to the wrong FAQ questions: for most of them they are shifter by 1, and for some are intermingled. This numbering confusion was introduced in https://github.com/openvinotoolkit/openvino/pull/11605 
 - We sent to Telemetry the old wrong numbers, because telemetry relies on a error messages not on a FAQ text. Changing Telemetry numbers will cause significant confusion, therefore faq numbers in python code should be unchanged.
 -  Corrected FAQ numbering in docs. There are blank questions, they are necessary to keep numbering aligned. In the code blank questions were never referred with `refer_to_faq_msg`.

### Tickets:
 - xxx-99850


tools/mo/openvino/tools/mo/utils/utils.py
```diff
@@ -25,6 +25,22 @@ def refer_to_faq_msg(question_num: int):
...
+        94: 95,  # but actual info in FAQ should be updated
+        97: 93,
+        100: 96,
+        101: 97,
+    }
+
+    if 22 >= question_num <= 93:  # From 22 to 93 they are shifted by one
+        question_num -= 1
+    elif question_num in map_to_correct_faq_numbers:
+        question_num = map_to_correct_faq_numbers[question_num]
```
rkazants, line 42: may be it makes sense to update numbers in FAQ?

pavel-esir: Reverted code changes and updated docs. No committee approval needed for merge into 2023.0

There are blank questions, they are necessary to keep numbering aligned. In the code blank questions were never referred with `refer_to_faq_msg`. Numbering was broken here
https://github.com/openvinotoolkit/openvino/pull/11605 mainly current PR restores that broken numbering.

popovaan (approved):

docs/MO_DG/prepare_model/Model_Optimizer_FAQ.md
```diff
@@ -873,70 +885,45 @@ For all frameworks:
...
-Q100. What does the message "SyntaxError: 'yield' inside list comprehension" during MxNet model conversion mean?
-#####################################################################################################################################################
-
-**A:** The issue "SyntaxError: ``yield`` inside list comprehension" might occur during converting MXNet models (``mobilefacedet-v1-mxnet``, ``brain-tumor-segmentation-0001``) on Windows platform with Python 3.8 environment. This issue is caused by the API changes for ``yield expression`` in Python 3.8.
-The following workarounds are suggested to resolve this issue:
-1. Use Python 3.7 to convert MXNet models on Windows
-2. Update Apache MXNet by using ``pip install mxnet==1.7.0.post2``
-Note that it might have conflicts with previously installed PyPI dependencies.
-
-.. _question-101:
```
rkazants, line 937: why do you remove it? Looks still valid.

pavel-esir: We are no longer referring to this message in our code. I suspect this is because we always force user to use mxnet with 1.7.0.post2 or greated. But let me double check this

rkazants (approved):

PR #17228, opened 2023-04-26
[PT FE] Fix aten::conv2d conversion issues related to bias

mmikolajcz:
### Details:
 - *Fix issue with conversion due to missing bias shape in constant*
 - *...*

### Tickets:
 - *101077*


src/frontends/pytorch/src/op/convnd.cpp
```diff
@@ -50,7 +50,7 @@ OutputVector translate_convnd(const NodeContext& context) {
                                                       pad_type);
     }
     if (!context.input_is_none(2)) {
-        auto bias = context.get_input(2);
+        auto bias = context.get_input_from_visible_context(2);
```
mvafin, line 53: We can do this only when bias is constant. Please add such verification.

mmikolajcz: Added cast to Constant as condition to check whether bias can be acquired from visible context

src/frontends/pytorch/src/op/convnd.cpp
```diff
@@ -51,6 +51,9 @@ OutputVector translate_convnd(const NodeContext& context) {
     }
     if (!context.input_is_none(2)) {
         auto bias = context.get_input(2);
+        if (std::dynamic_pointer_cast<v0::Constant>(context.get_input_from_visible_context(2).get_node_shared_ptr())) {
```
mvafin, line 54: Would be better to call `get_input_from_visible_context` once also please add at least one test

mmikolajcz: Done

PR #17232, opened 2023-04-26
[CPU][Tests] Fixed MHA input generation

a-sidorova:
### Details:
 - *Fixed input generation for MHA tests: changed to relative*
 - *It's leftover from [PR#17157](https://github.com/openvinotoolkit/openvino/pull/17157)*

### DIFF
--- a/src/tests/functional/plugin/shared/src/snippets/mha.cpp
+++ b/src/tests/functional/plugin/shared/src/snippets/mha.cpp
@@ -68,7 +68,7 @@ void MHA::generate_inputs(const std::vector<ngraph::Shape>& targetInputStaticSha
     for (int i = 0; i < model_inputs.size(); ++i) {
         const auto& model_input = model_inputs[i];
         ov::Tensor tensor;
-        tensor = ov::test::utils::create_and_fill_tensor_normal_distribution(model_input.get_element_type(), targetInputStaticShapes[i], 1.0f, 0.5f);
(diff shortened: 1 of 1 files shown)


a-sidorova:
@maxnick take a look please

src/tests/functional/plugin/shared/src/snippets/mha.cpp
```diff
@@ -68,7 +68,7 @@ void MHA::generate_inputs(const std::vector<ngraph::Shape>& targetInputStaticSha
     for (int i = 0; i < model_inputs.size(); ++i) {
         const auto& model_input = model_inputs[i];
         ov::Tensor tensor;
-        tensor = ov::test::utils::create_and_fill_tensor_normal_distribution(model_input.get_element_type(), targetInputStaticShapes[i], 1.0f, 0.5f);
+        tensor = ov::test::utils::create_and_fill_tensor(model_input.get_element_type(), targetInputStaticShapes[i], 11, 0, 10);
```
maxnick, line 71: Suggest considering https://github.com/openvinotoolkit/openvino/blob/2255bb25fd18e901da2a7af915c17d44ba537dc0/src/tests/functional/shared_test_classes/src/base/utils/generate_inputs.cpp#L22-L43
The main idea is that the input data are divided by modulo 256, which provides setting only first 8 mantissa bits (which is bf16 numbers resolution).

a-sidorova:
Closed since it's fixed in bounds of PR https://github.com/openvinotoolkit/openvino/pull/15647

PR #17238, opened 2023-04-26
[Shape_infer] Use axes param shape to deduce Squeeze output rank

mitruska:
### Details:
 - [Shape_infer] Use axes param shape (if scalar or 1D) to deduce Squeeze output rank
 
### Tickets:
 - 92973


src/bindings/python/tests_compatibility/test_onnx/test_backend.py
```diff
@@ -313,7 +307,8 @@ def expect_fail(test_case_path, xfail):  # type: (str) -> None
     ),
     (
         skip_dynamic_model,
-        "OnnxBackendNodeModelTest.test_triu_one_row_cpu"
+        "OnnxBackendNodeModelTest.test_triu_one_row_cpu",
+        "OnnxBackendNodeModelTest.test_squeeze_negative_axes_cpu",
```
mitruska, line 311: This test is skipped only for compatibility (legacy) api, because unsupported dynamic shapes:
https://github.com/openvinotoolkit/openvino/blob/4f2a041a4de2edc70fef920ab68b99fd968230fc/src/bindings/python/tests_compatibility/__init__.py#L167

src/core/shape_inference/include/squeeze_shape_inference.hpp
```diff
@@ -50,11 +50,34 @@ void shape_infer(const Squeeze* op,
...
+                    std::count_if(arg_shape.begin(), arg_shape.end(), [&](const DimType& dim) {
+                        if (dim.compatible(1)) {
+                            return true;
+                        } else {
+                            // Copy not squeezable dimensions to the output shape
+                            output_shape.push_back(dim);
+                            return false;
+                        }
+                    });
+                if (squeezable_dims_count > axes_shape[0].get_min_length()) {
```
jane-intel, line 72: Unfortunately, I cannot agree on this solution.

The following case output rank must be dynamic:
```

Parameter([-1, -1, 2, 3, -1, 5])---\
                                             Squeeze 
Parameter([2]) ------------------/

```

TensorFlow example:

image

I propose to exclusively work with scalars and 1D tensors with 1 element. We are sure that they are unique so we can build our solution on top of that.

jane-intel (changes requested):

src/core/shape_inference/include/squeeze_shape_inference.hpp
```diff
@@ -50,11 +50,34 @@ void shape_infer(const Squeeze* op,
...
+                    // If the number of squeezable (1 or with 1 in range) dimensions is bigger than the number of axes,
+                    // only rank can be deduced (assuming uniqueness of axes)
+                    output_shape = PartialShape::dynamic(arg_shape.size() - shape_size(axes_shape.to_shape()));
```
praasz, line 75: This block tries to get squeeze, try move output shape creation to block below e.g. somewhere between line 123-130

src/core/shape_inference/include/squeeze_shape_inference.hpp
```diff
@@ -50,11 +50,34 @@ void shape_infer(const Squeeze* op,
...
-            normalize_axes(op, arg_rank.get_length(), axes);
-            OPENVINO_SUPPRESS_DEPRECATED_END
-            unique_axes.reset(new std::set<int64_t>(axes.cbegin(), axes.cend()));
+        if (arg_rank.is_static() && axes_shape.is_static()) {
+            if (get_data_as_int64<T>(1, op, axes, constant_data)) {
+                OPENVINO_SUPPRESS_DEPRECATED_START
+                normalize_axes(op, arg_rank.get_length(), axes);
+                OPENVINO_SUPPRESS_DEPRECATED_END
+                unique_axes.reset(new std::set<int64_t>(axes.cbegin(), axes.cend()));
+            } else if (axes_shape.size() == 1 && axes_shape[0].get_min_length() > 0) {
```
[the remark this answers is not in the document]
mitruska: Actually we can, updated.

src/core/shape_inference/include/squeeze_shape_inference.hpp
mitruska: The logic for constant duplicated `axes` stays the same (as in Tensorflow example), 
the proposed change affects only `axes` as Parameter. 
Enablement of (non single element) Parameter as input is possible only by assuming uniqueness of axes, 
otherwise plugins are not able to infer dynamic rank anyway. 
So the hard choice is between keeping unsupported dynamic rank or accepting the risk of output shape mismatch (if axes are duplicated) in favour of extending Parameter input support.

For the agreement, I have updated the changes to support only Parameter with single element.

src/core/tests/type_prop/squeeze.cpp
```diff
@@ -41,6 +41,60 @@ TEST(type_prop, squeeze_incorrect_negative_axes) {
                     HasSubstr("Parameter axis -10 out of the tensor rank range"));
 }
 
+TEST(type_prop, data_static_param_axes_static_shape_squeezable_dims_equal) {
```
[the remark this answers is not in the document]
mitruska: Done

src/core/shape_inference/include/squeeze_shape_inference.hpp
mitruska: I have tried (https://github.com/openvinotoolkit/openvino/pull/17238/commits/34d742885e01a3d98be15c189b24fde92591e6e4), but then some checks like number of inputs and axes shape need to be repeated.
Also with the latest commits the solution is not so complex.

src/core/tests/type_prop/squeeze.cpp
```diff
@@ -145,7 +280,12 @@ TEST_P(SqueezeTest, partial_shape_dimension_propagation_parameter_axes_no_data)
     const auto squeeze = std::make_shared<op::v0::Squeeze>(param, axes_node);
 
     EXPECT_EQ(squeeze->get_element_type(), element::f32);
-    EXPECT_EQ(squeeze->get_output_partial_shape(0), PartialShape::dynamic());
+
+    if (axes.size() == 0 || param->get_output_partial_shape(0).rank().is_dynamic()) {
+        EXPECT_EQ(squeeze->get_output_partial_shape(0), PartialShape::dynamic());
```
praasz, line 285: I think is not OK the the correct shape should be set in test parameter and stored in variable `exp_shape`
It looks like bypass expected result.
Update the `static_partial_shapes_test_values ` and/or `empty_axes_test_values `.

src/core/tests/type_prop/squeeze.cpp
```diff
@@ -41,6 +41,141 @@ TEST(type_prop, squeeze_incorrect_negative_axes) {
...
+    auto param = std::make_shared<ov::op::v0::Parameter>(ov::element::f32, PartialShape{2, 1, 4});
+    const auto axes_node = std::make_shared<op::Parameter>(element::u64, PartialShape{2});
+    const auto squeeze = std::make_shared<op::v0::Squeeze>(param, axes_node);
+
+    EXPECT_EQ(squeeze->get_element_type(), element::f32);
+    EXPECT_EQ(squeeze->get_output_partial_shape(0), PartialShape::dynamic());
+}
+
+TEST(type_prop, squeeze_data_static_param_axes_1D_single_elem_static_shape_squeezable_dims_one) {
+    auto param = std::make_shared<ov::op::v0::Parameter>(ov::element::f32, PartialShape{2, 1, 4});
```
praasz, line 63: What about case where the input shape will be [2,2,4]?
Then reduce the output shape rank I think is not correct (maybe validation error should be thrown as there is no squeezable axis in input shape).

src/core/shape_inference/include/squeeze_shape_inference.hpp
```diff
@@ -50,11 +50,19 @@ void shape_infer(const Squeeze* op,
...
+            if (get_data_as_int64<T>(1, op, axes, constant_data)) {
+                // The values of `axes` input are known
+                OPENVINO_SUPPRESS_DEPRECATED_START
+                normalize_axes(op, arg_rank.get_length(), axes);
+                OPENVINO_SUPPRESS_DEPRECATED_END
+                unique_axes.reset(new std::set<int64_t>(axes.cbegin(), axes.cend()));
+            } else if (arg_rank.get_length() > 0 && shape_size(axes_shape.to_shape()) == 1) {
+                // The `axes` input must be a Parameter with single element to ensure uniqueness of axes
+                // only rank is deduced
+                output_shape = PartialShape::dynamic(arg_rank.get_length() - 1);
```
praasz, line 63: If axes is parameter without any data for input parameter then you squeeze rank but what if input shape has no axis which is compatible with dimension `1`?

src/core/shape_inference/include/squeeze_shape_inference.hpp
praasz: Sure, but added another comment I think some check has to be done.

jane-intel (changes requested):
Please address @praasz comments

src/core/tests/type_prop/squeeze.cpp
mitruska: > This line should be restored to EXPECT_EQ(squeeze->get_output_partial_shape(0), exp_shape);

This line can't be "restored" to such check as it wasn't originally like that. 
The original check expected for all cases fully dynamic shape instead of "exp_shape":
`EXPECT_EQ(squeeze->get_output_partial_shape(0), PartialShape::dynamic());`

As the goal of the improvement is to make the output rank not always dynamic, the original check is no longer universal as sometimes the rank can be static.
So I have updated the check to check compatibility of output shape with "exp_shape" (expected for constant input). 
`EXPECT_TRUE(squeeze->get_output_partial_shape(0).compatible(exp_shape));`

Tests expected specific shape/rank have been added separately.

src/core/tests/type_prop/squeeze.cpp
mitruska: The check for number of squeezable dimensions has been added. But previously output shape rank was dynamic in such case.

src/core/shape_inference/include/squeeze_shape_inference.hpp
mitruska: The check for number of squeezable dimensions has been added. But previously output shape rank was dynamic in such case.

praasz (approved):

src/core/shape_inference/include/squeeze_shape_inference.hpp
mitruska: Right, thanks, done:)

jane-intel (approved):

PR #17243, opened 2023-04-26
Add FakeQuantize op support in TS transformations

itikhono:
Added support for FQ op in TS transformation
Tests

Before
image

After
image

### Tickets:
 - *105091*

### DIFF
--- a/src/common/transformations/src/transformations/transpose_sinking/ts_binary.cpp
+++ b/src/common/transformations/src/transformations/transpose_sinking/ts_binary.cpp
@@ -6,6 +6,7 @@
 
 #include "itt.hpp"
 #include "openvino/op/constant.hpp"
+#include "openvino/op/fake_quantize.hpp"
 #include "openvino/op/prelu.hpp"
 #include "openvino/op/transpose.hpp"
 #include "openvino/op/util/op_types.hpp"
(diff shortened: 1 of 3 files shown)


jane-intel:
FQ TS looks legit. I approve this particular change.
However, the "After" graph that you show in the PR description brings more questions. I assume that this is an IR since it is visualized via netron.

- Why quantized weights are kept in a non-compressed form?
- Why Reshape and Transpose are still present on the weights sub-graph and are not optimized out?

t-jankowski (approved):

evkotov (commented):
I don't see how we add FakeQuantize to the list of supported TS operations

itikhono:
> I don't see how we add FakeQuantize to the list of supported TS operations

image

andrei-kochin (approved):

evkotov (commented):
I mean src/common/transformations/src/transformations/transpose_sinking/ts_utils.cpp
bool CanPropagateForwardThrough(Node* node) {

itikhono:
> I mean src/common/transformations/src/transformations/transpose_sinking/ts_utils.cpp bool CanPropagateForwardThrough(Node* node) Without that you could have problems with back propagation

Agree

evkotov (approved):

jane-intel (approved):

PR #14691, opened 2022-12-16
add input range for convolution/deformableconvolution

wilson-seok:
### Details:
 - Input range -1 to 1 for Convolution/DeformableConvolution

### Tickets:
 - 82001

PR #17244, opened 2023-04-27
[IE_TESTS] Correct random data generation

EgorDuplensky:
### Details:
 - In current implementation k (resolution) parameter of generation function affect the range of the generated data. Change the behavior so k parameter only affects granularity.
 Example:
 - Before: start_from -1, range 2, resolution 100 generates numbers in range: [-1, 0.01] with resolution 100
 - After: start_from -1, range 2, resolution 100 generates numbers in range: [-1, 1] with resolution 100
 - [x] Many tests failed, should be fixed / adopted to the corrected behavior.
 - [x] Create tickets for skipped tests
 
 ### Ticket:
 - 110384

Multiple test instances started to fail after correction of random data generation.
Failing instances were skipped for now and follow-up tickets were created for future investigation.
### GPU:
- 111440 - GatherElements
- 111437 - Deconv, GroupDeconv
### Template:
- 111443 - Eltwise Mod
### CPU:
- 111418 - Snippets Convert
- 111412 - Proposal
- 111406 - Interpolate
- 111405 - GridSample
- 111404 - GatherElements

### DIFF
--- a/src/tests/functional/shared_test_classes/src/base/utils/generate_inputs.cpp
+++ b/src/tests/functional/shared_test_classes/src/base/utils/generate_inputs.cpp
@@ -26,7 +26,7 @@ ov::runtime::Tensor generate(const std::shared_ptr<ov::Node>& node,
     size_t inNodeCnt = node->get_input_size();
     InputGenerateData inGenData;
     if (elemType.is_real()) {
-        inGenData.range = 2560;
+        inGenData.range = 10;
         inGenData.resolution = 256;
     }
     auto it = inputRanges.find(node->get_type_info());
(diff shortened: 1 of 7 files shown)


iefode:
> @iefode Could you please take a look? The contexts is the following: **fill_data_random()** is used in scope of CPU plugin func tests (and other plugins as well) I have noticed that this function fills blobs with unexpected range of values. From my understanding the generated range should not depend on "k" parameter (so-called resolution). The "k" parameter should only affect granularity of the range. For example the **fill_random_unique_sequence** does almost the same job but handles the "range" and "k" correctly.
> 
> The problem with the fix is that the plugin tests are started to fail, most likely because they rely on this unexpected behavior fill_random_data().

@EgorDuplensky I have reviewed the changes. In general, I agree with the fix. But we need to check input data to generate inputs per layer in [https://github.com/openvinotoolkit/openvino/blob/master/src/tests/functional/shared_test_classes/src/base/utils/generate_inputs.cpp](genarate_input) and [https://github.com/openvinotoolkit/openvino/blob/master/src/tests/functional/shared_test_classes/include/shared_test_classes/base/utils/ranges.hpp](ranges)

EgorDuplensky:
@vladimir-paramuzov FYI. Few GPU tests were skipped and expected to be fixed by GPU team.

src/tests/functional/shared_test_classes/src/base/utils/generate_inputs.cpp
```diff
@@ -26,8 +26,8 @@ ov::runtime::Tensor generate(const std::shared_ptr<ov::Node>& node,
     size_t inNodeCnt = node->get_input_size();
     InputGenerateData inGenData;
     if (elemType.is_real()) {
-        inGenData.range = 2560;
```
iefode, line 29: As I remember, we used such ranges for bf16/f16... @maxnick What do you think?

iefode (commented):
In general, LGTM except the comment

src/tests/functional/shared_test_classes/src/base/utils/generate_inputs.cpp
maxnick: The whole idea of using the range of 2560 and the resolution 256 is to generate such input data where all the floating point values have mantissa with bits set in the first 7 binary digits, so that to avoid input and output data rounding when fp32 is converted to bf16. In other words, with such parameters of the old data generation routine, all the input fp32 values were directly converted to bf16 as is, thus helping to avoid results difference between runs on the CPU plugin with bf16 precision and fp32 reference.  Also this problem was explained in: https://github.com/openvinotoolkit/openvino/pull/14691#discussion_r1104391322

EgorDuplensky: @maxnick resolution changed back to 256

maxnick (approved):
Agreed to adjust the resolution parameter to 16 in a follow up PR in order to have exactly 7 significant digits in the mantissa part of generated values.

PR #17250, opened 2023-04-27
[GPU] Use BFS processing order for out_of_order queue

sshlyapn:
### Details:
 - This PR fixes performance issue caused by changing processing order from DFS to BFS during remove redundant reorder pass, which leads to less efficient kernels execution in case of out_of_order queue and higher host overheads in case of barriers usage for kernels synchronization 

### Tickets:
 - 108347


vladimir-paramuzov (approved):
LGTM if perf is ok

isanghao:
Let me check perf impact from dGPU

yeonbok (approved):

p-durandin:
wait for DLB results

PR #17254, opened 2023-04-27
[GPU] Added missing second input to splited reshape in handle_reshape pass

vladimir-paramuzov:
### Details:
 - When reshape was copied in handle_reshape pass, second input wasn't connected to the new reshape node, so with 1 dependency shape infer worked incorrectly in dynamic case.

### Tickets:
 - *109479*


src/plugins/intel_gpu/src/graph/graph_optimizer/handle_reshape.cpp
```diff
@@ -181,6 +181,10 @@ void handle_reshape::run(program& p) {
                                        *reorder_reshape_node,
                                        0,
                                        reshape_input_node.get_dependencies().empty());
+
+                    if (reorder_reshape_node->get_primitive()->input_size() == 2) {
```
sshlyapn, line 185: Wouldn't it be better to move this node modification to [node creation part](https://github.com/openvinotoolkit/openvino/blob/22bb3af7df965dc85f7b4d506f9649ecedaff775/src/plugins/intel_gpu/src/graph/graph_optimizer/handle_reshape.cpp#L159) rather than apply it in reorders adding stage?

vladimir-paramuzov: updated

sshlyapn (approved):

PR #17275, opened 2023-04-27
[GPU] Improve dump naming rule in debug feature.

hyunback:
Now, we can support below dump naming rule.
- Exec_graph name
- Regular express
- Case-insensitive name searching

New introduces OV_GPU_xxx
- OV_GPU_ListLayers = 1 (Show layer names and exit)
- OV_GPU_VerboseColor = 1 (Show color verbose)

Add file, line, function in log prefix

### Tickets:
 - *106742*

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
@@ -42,6 +42,7 @@ enum class LogLevel : int8_t {
 }  // namespace ov
 
 #ifdef GPU_DEBUG_CONFIG
+#define __FILENAME__ (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
 #define GPU_DEBUG_IF(cond) if (cond)
 #define GPU_DEBUG_CODE(...) __VA_ARGS__
 #define GPU_DEBUG_DEFINE_MEM_LOGGER(stage) \
@@ -52,8 +53,16 @@ enum class LogLevel : int8_t {
 #define GPU_DEBUG_PROFILED_STAGE_CACHE_HIT(val) stage_prof.set_cache_hit(val)
 
 #define GPU_DEBUG_LOG_RAW_INT(min_verbose_level) if (cldnn::debug_configuration::get_instance()->verbose >= min_verbose_level) \
-                                                std::cout << cldnn::debug_configuration::prefix
+    ((cldnn::debug_configuration::get_instance()->verbose_color == 0) ? GPU_DEBUG_LOG_PREFIX : GPU_DEBUG_LOG_COLOR_PREFIX)
 #define GPU_DEBUG_LOG_RAW(min_verbose_level) GPU_DEBUG_LOG_RAW_INT(static_cast<std::underlying_type<ov::intel_gpu::LogLevel>::type>(min_verbose_level))
+#define GPU_DEBUG_LOG_PREFIX    std::cout << cldnn::debug_configuration::prefix << __FILENAME__ << ":" <<__LINE__ << ":" << __func__ << ": "
+#define GPU_DEBUG_LOG_COLOR_PREFIX  std::cout << DARK_GRAY << cldnn::debug_configuration::prefix << \
+    BLUE << __FILENAME__ << ":" << PURPLE <<  __LINE__ << ":" << CYAN << __func__ << ": " << RESET
+#define DARK_GRAY   "\033[1;30m"
+#define BLUE        "\033[1;34m"
+#define PURPLE      "\033[1;35m"
+#define CYAN        "\033[1;36m"
+#define RESET       "\033[0m"
 #else
 #define GPU_DEBUG_IF(cond) if (0)
 #define GPU_DEBUG_CODE(...)
@@ -66,6 +75,7 @@ enum class LogLevel : int8_t {
 // Macro below is inserted to avoid unused variable warning when GPU_DEBUG_CONFIG is OFF
 #define GPU_DEBUG_GET_INSTANCE(name) auto name = cldnn::debug_configuration::get_instance(); (void)(name);
 
+#define GPU_DEBUG_COUT              GPU_DEBUG_LOG_RAW(ov::intel_gpu::LogLevel::DISABLED)
 #define GPU_DEBUG_INFO              GPU_DEBUG_LOG_RAW(ov::intel_gpu::LogLevel::INFO)
 #define GPU_DEBUG_LOG               GPU_DEBUG_LOG_RAW(ov::intel_gpu::LogLevel::LOG)
 #define GPU_DEBUG_TRACE             GPU_DEBUG_LOG_RAW(ov::intel_gpu::LogLevel::TRACE)
@@ -81,6 +91,8 @@ public:
     static const char *prefix;
     int help;                                   // Print help messages
     int verbose;                                // Verbose execution
+    int verbose_color;
+    int list_layers;                            // Print list layers
     int print_multi_kernel_perf;                // Print execution time of each kernel in multi-kernel primitimive
     int disable_usm;                            // Disable usm usage
     int disable_onednn;                         // Disable onednn for discrete GPU (no effect for integrated GPU)
--- a/src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
@@ -42,7 +42,12 @@ enum class LogLevel : int8_t {
 }  // namespace ov
 
 #ifdef GPU_DEBUG_CONFIG
-#define __FILENAME__ (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
+#if defined(_WIN32) || defined(_WIN64)
+#define SEPARATE '\\'
+#else
+#define SEPARATE '/'
+#endif
+#define __FILENAME__ (strrchr(__FILE__, SEPARATE) ? strrchr(__FILE__, SEPARATE) + 1 : __FILE__)
 #define GPU_DEBUG_IF(cond) if (cond)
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
```diff
@@ -41,6 +41,7 @@ enum class LogLevel : int8_t {
 }  // namespace ov
 
 #ifdef GPU_DEBUG_CONFIG
+#define __FILENAME__ (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
```
isanghao, line 44: does it work for windows as well?

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
```diff
@@ -104,6 +105,7 @@ static void print_help_messages() {
     std::vector<std::pair<std::string, std::string>> message_list;
     message_list.emplace_back("OV_GPU_Help", "Print help messages");
     message_list.emplace_back("OV_GPU_Verbose", "Verbose execution");
+    message_list.emplace_back("OV_GPU_ListLayers", "Print layers names");
```
isanghao, line 108: Please add `OV_GPU_VerboseColor` too

isanghao (changes requested):

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
```diff
@@ -238,16 +244,50 @@ const debug_configuration *debug_configuration::get_instance() {
...
+                return elem - ('a'-'A');
+            }
+            return elem;
+        };
+        for (size_t j = 1; j <= m; ++j)
+            if (pattern[j - 1] == '*')
+                lookup_tbl[0][j] = lookup_tbl[0][j - 1];
+        for (size_t i = 1; i <= n; ++i) {
+            for (size_t j = 1; j <= m; ++j) {
+                if (pattern[j - 1] == '*') {
```
yeonbok, line 278: how about 
1) replace start to .* 
2) replace ? to .
3) use std::regex_match ?
instead of custom matcher.

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
hyunback: Oh, I my missed, applied

src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
hyunback: Windows has an issue for different separator issue, I updated to use different separator in Windows.

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
hyunback: Good suggestion, I updated to use std::regex_match instead custom matcher. it looks much simpler.

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
```diff
@@ -238,16 +246,31 @@ const debug_configuration *debug_configuration::get_instance() {
...
+        auto exec_graph_name = layer_name.substr(pos + 1, layer_name.size());
+        auto upper_exec_graph_name = std::string(exec_graph_name.length(), '\0');
+        std::transform(exec_graph_name.begin(), exec_graph_name.end(), upper_exec_graph_name.begin(), ::toupper);
+        auto upper_pattern = std::string(pattern.length(), '\0');
+        std::transform(pattern.begin(), pattern.end(), upper_pattern.begin(), ::toupper);
+        if (upper_exec_graph_name.compare(upper_pattern) == 0) {
+            return true;
+        }
+        // Check pattern with regular expression
+        std::regex re(pattern, std::regex::icase);
```
yeonbok, line 269: 1) Oh it seems that icase enables to match regardless cases
2) But seems that in some version of gcc, icase has a bug 
3) So anyway it seems that transform to upper looks better
4) But then, we don't need icase?

hyunback: Okay, I see, I updated to use upper with transformer.
BTW what means "we don't need icase"?  Our expectation is insensitive case and std::regex with std::regex::icase define it runwithout regard to caseignore.

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -1150,6 +1150,13 @@ void network::execute_impl(const std::vector<event::ptr>& events) {
 
     set_arguments();
     GPU_DEBUG_GET_INSTANCE(debug_config);
+    GPU_DEBUG_IF(debug_config->list_layers == 1) {
```
yeonbok, line 1153: Is there a reason doing this in network? Could we do this in program?

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
```diff
@@ -145,6 +149,8 @@ static void print_help_messages() {
 debug_configuration::debug_configuration()
         : help(0)
         , verbose(0)
+        , verbose_color(0)
```
yeonbok, line 152: How about set this as true by default?

src/plugins/intel_gpu/src/graph/network.cpp
hyunback: No special reason, I don't care for it. Could you let me know where's your recommendation in program? then I'll update it.

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
hyunback: As you know, we can use it by environment, so color setting can be default by personal environment.
I think non color as default is better in personal.

src/plugins/intel_gpu/src/graph/network.cpp
yeonbok: Hmm not important to me too, sorry for the confusion :) 
BTW can this cover body networks in loop?

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
yeonbok: Hmm I am not sure "default non color is better" (I prefer colored default but yeah its nit...)

yeonbok:
LGTM if my last question about body network is resolved :)

isanghao (approved):
LGTM

src/plugins/intel_gpu/src/graph/network.cpp
hyunback: I tried to test it with loop model with unrolling, but I have an issue loop in model.
I changed some code because I heard some condition run loop with unrolling.
But there's some issue run loop in my side. (I'm not familiar loop model)
I found only 3 models with loop type( grep type="Loop" in IR) from all models in 1 team covered.
yolo_v3 -> always unrolling
deepspeech2 -> running failed
ocr-perpetuuiti -> hang issue
**Code change for force enable unrolling loop**
image

So my question, 
1. Code change is correct for force running loop?
2. If true, do we have an known issue for loop? or If not could you share the model which is working well with loop?

yeonbok: You can try with text-recognition-0011 for running a network w/ loops. 
With that, you don't need any change in transformations_pipeline.cpp to prevent unrolling.

hyunback: Update to show loop body primitives.

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
```diff
@@ -18,11 +18,11 @@ const char *debug_configuration::prefix = "GPU_Debug: ";
 
 #ifdef GPU_DEBUG_CONFIG
 
-#define GPU_DEBUG_COUT std::cout << cldnn::debug_configuration::prefix
+#define GPU_DEBUG_COUT_ std::cout << cldnn::debug_configuration::prefix
```
ahnyoung-paul, line 21: nit: just asking out of curiosity, why do you change the name of define to GPU_DEBUG_COUT_ ?

hyunback: I newly added GPU_DEBUG_COUT definition in debug_configuraton.hpp, it needs debug_configuration instant (from GPU_DEBUG_LOG_RAW_INT).
But GPU_DEBUG_COUT in cpp was used during instatnt creation, so there's no debug_configuration instant at that time. For the simple, I rename GPU_DEBUG_COUT in cpp.

yeonbok (approved):

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
ahnyoung-paul: I see. thanks.

ahnyoung-paul (approved):

isanghao (approved):

isanghao (approved):

src/plugins/intel_gpu/src/runtime/debug_configuration.cpp
```diff
@@ -257,16 +264,31 @@ const debug_configuration *debug_configuration::get_instance() {
...
+        (layer_name.find("constant:") == std::string::npos))
         return true;
     if (dump_layers.empty() && dump_layers_result == 0)
         return true;
 
+    auto is_match = [](const std::string& layer_name, const std::string& pattern) -> bool {
+        auto upper_layer_name = std::string(layer_name.length(), '\0');
+        std::transform(layer_name.begin(), layer_name.end(), upper_layer_name.begin(), ::toupper);
+        auto upper_pattern = std::string(pattern.length(), '\0');
+        std::transform(pattern.begin(), pattern.end(), upper_pattern.begin(), ::toupper);
```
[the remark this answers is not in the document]
hyunback: Looks nice, I'll update it next time.

PR #17283, opened 2023-04-28
Fix the names copying in TransposeSinking backward transformations

itikhono:
### Details:

- Fix copying of `tensor names` in TS sinking transformations in multi consumers cases. The new logic:
image

- Fix `friendly names` copying in TS transformation in multi consumers cases.
The new logic: We detect Transpose directly connected to Result operation and copy its name.
If there are multiple such Transposes, TS will work anyway and copy one of available names.
 image

- Fix `sinking is available` check in TS transformations.

- Added `sorting by tensor names` option to GraphComparator 

Need to run performance check before merge.

### Tickets:
 - *108928*


src/common/transformations/src/transformations/transpose_sinking/ts_utils.cpp
```diff
@@ -430,15 +428,47 @@ bool HasSameOutputTransposeNodes(const Output<Node>& output) {
     return HasSameOutputTransposeNodes(output.get_node_shared_ptr());
 }
 
-void RemoveSingleOutputConsumers(const NodePtr& node) {
+bool RemoveTransposeConsumers(const NodePtr& node) {
+    std::map<size_t, std::vector<ov::op::v1::Transpose*>> out_idx_to_redundant_transposes;
```
evkotov, line 432: consider to use std::unordered_map since it has constant complexity instead of std::map

src/common/transformations/src/transformations/transpose_sinking/ts_utils.cpp
```diff
@@ -112,11 +112,6 @@ void SwapFriendlyNames(const NodePtr& node1, const NodePtr& node2) {
     node1->set_friendly_name(node2_name);
 }
 
-void SwapNames(const NodePtr& node1, const NodePtr& node2) {
-    SwapFriendlyNames(node1, node2);
-    SwapOutputNames(node1->output(0), node2->output(0));
```
evkotov, line 117: SwapFriendlyNames and SwapOutputNames both removed.
RemoveTransposeConsumers now sets friendly names, but where and how set_names() called for tensor names?

src/common/transformations/src/transformations/transpose_sinking/ts_utils.cpp
```diff
@@ -418,6 +413,9 @@ bool HasSameOutputTransposeNodes(const NodePtr& main_node) {
                             transpose_axis_order.end(),
                             first_transpose_axis_order.begin()))
                 return false;
+            if (!is_sinking_node(input.get_node())) {
+                return false;
+            }
```
evkotov, line 418: It's a good idea to move is_sinking_node() call to one place. But now HasSameOutputTransposeNodes() does 2 roles:
- checking if a node has the same output transpose nodes
- checking if each transpose is sinking
Name of function is left the same. Later it will hard to understand why some node is not moving backward since there is no is_sinking_node() call in pattern check now. It is inside HasSameOutputTransposeNodes

evkotov: Consider to rename HasSameOutputTransposeNodes

src/common/transformations/src/transformations/transpose_sinking/ts_utils.cpp
itikhono: Good point, `replace` method do it 
image
image

src/common/transformations/src/transformations/transpose_sinking/ts_utils.cpp
itikhono: not so sure, maybe CheckTransposeConsumers?

src/common/transformations/src/transformations/transpose_sinking/ts_utils.cpp
itikhono: done

evkotov (approved):

PR #17294, opened 2023-04-28
Fix enable_plugins_xml

Maxim-Doronin:
### Details:
`cmake_dependent_option(<option> "<help_text>" <value> <depends> <force>)`
Makes `<option>` available to the user if the [semicolon-separated list](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-lists) of conditions in `<depends>` are all true. Otherwise, a local variable named `<option>` is set to `<force>`.

https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

### Tickets:
 - E78176
 - E78327

### Release cherry-pick
https://github.com/openvinotoolkit/openvino/pull/17293

### DIFF
--- a/cmake/features.cmake
+++ b/cmake/features.cmake
@@ -96,7 +96,7 @@ ie_option (ENABLE_HETERO "Enables Hetero Device Plugin" ON)
 
 ie_option (ENABLE_TEMPLATE "Enable template plugin" ON)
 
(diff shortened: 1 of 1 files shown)


DariaMityagina:
@ilya-lavrenov, @ilyachur hi!
Could you please take a look at this PR?

ilya-lavrenov (approved):

PR #17295, opened 2023-04-28
[TF FE] Use original input types for SavedModel

rkazants:
**Details:** The problem is that if user tries to convert a model with specified shapes `"--input input1[2,3],input2[4,5]"`, TF FE creates a new tensor place with dynamic type. Though, it should use the original type for `input1` and `input2`. That is a reason to get inconsistent type during conversion of SM format.

Also, refactor TF FE unit-tests. 

**Ticket:** 109544

### DIFF
--- a/src/frontends/tensorflow/src/input_model.cpp
+++ b/src/frontends/tensorflow/src/input_model.cpp
@@ -175,7 +175,15 @@ void InputModel::InputModelTFImpl::load_places() {
             }
             std::vector<std::string> names = {op_name};
             auto tensor_place = std::make_shared<TensorPlace>(m_input_model, pshape, type, names);
+
+            // In Model Optimizer user can refer to model inputs by a name of Placeholder
+            // and its output port, for example, using `input_name` and `input_name:0`
+            // Also, SavedModel format contains a signature that maps external input names to internal ones with port
+            // index like `input_name` maps to `serving_default_input_name:0`.
+            // So we have to store with `:0` as well to get its tensor by `get_place_by_tensor_name` method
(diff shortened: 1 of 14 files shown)


src/frontends/tensorflow/src/input_model.cpp
```diff
@@ -175,7 +175,15 @@ void InputModel::InputModelTFImpl::load_places() {
...
             std::vector<std::string> names = {op_name};
             auto tensor_place = std::make_shared<TensorPlace>(m_input_model, pshape, type, names);
+
+            // In Model Optimizer user can refer to model inputs by a name of Placeholder
+            // and its output port, for example, using `input_name` and `input_name:0`
+            // Also, SavedModel format contains a signature that maps external input names to internal ones with port
+            // index like `input_name` maps to `serving_default_input_name:0`.
+            // So we have to store with `:0` as well to get its tensor by `get_place_by_tensor_name` method
             m_tensor_places[op_name] = tensor_place;
+            m_tensor_places[op_name + ":0"] = tensor_place;
```
gkrivor, line 185: I didn't see such models. I'm not sure it is possible. Just FYI, something strange may be in case op_name already contains :0

gkrivor (approved):
Manual verification has been finished successfully.

src/frontends/tensorflow/src/input_model.cpp
rkazants: It is added because internal name for original inputs `serving_default_input:0`. Namely, it goes with suffix `:0`.
So for them we create new tensor places always, though, tensor places were created for them in `load_places` method.

PR #17303, opened 2023-05-01
[Core] Conv reference optimization.

nshchego:
### Details:
 - *Convolution reference time execution was improved in scope of task "Optimize CPU tests time".*

### Tickets:
 - *98677*

### DIFF
--- a/src/core/include/openvino/op/proposal.hpp
+++ b/src/core/include/openvino/op/proposal.hpp
@@ -78,7 +78,7 @@ namespace v4 {
 /// \ingroup ov_ops_cpp_api
 class OPENVINO_API Proposal : public op::v0::Proposal {
 public:
-    OPENVINO_OP("Proposal", "opset4", op::Op);
+    OPENVINO_OP("Proposal", "opset4", op::v0::Proposal);
     Proposal() = default;
     /// \brief Constructs a Proposal operation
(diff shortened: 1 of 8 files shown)


src/tests/ie_test_utils/common_test_utils/ov_tensor_utils.cpp
```diff
@@ -252,12 +252,14 @@ void compare(const ov::Tensor& expected,
             }
         }
     }
+#ifndef NDEBUG
```
EgorDuplensky, line 255: I would propose to perform this change in scope of separate PR just in case (since it is not related to the original change)

iefode:
Conformance jobs to check potential regression with results and validate speed up:
* [PR17303](https://openvino-ci.toolbox.iotg.sclab.intel.com/job/private-ci/job/developer-services/job/conformance/307/)
* [master](https://openvino-ci.toolbox.iotg.sclab.intel.com/job/private-ci/job/developer-services/job/conformance/308/)

src/tests/ie_test_utils/common_test_utils/ov_tensor_utils.cpp
nshchego: Done

PR #17328, opened 2023-05-02
Added Torchscript backend

suryasidd:
### Details:
 - Added torchscript backend to OpenVINO
 - Added an environment variable to enable torchscript backend USE_TS_BACKEND
### Tickets:
 - *ticket-id*

### DIFF
--- a/src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
+++ b/src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # flake8: noqa
+# mypy: ignore-errors
 
 import logging
 import os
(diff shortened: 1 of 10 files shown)


ilya-lavrenov (changes requested):
too late for 2023.0
CF has passed

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/backend.py
```diff
@@ -0,0 +1,90 @@
...
+
+        compiled_model = core.compile_model(om, device)
+
+        def _call(*args):
+            ov_inputs = [a.detach().cpu().numpy() for a in args]
+            try:
+                res = compiled_model(ov_inputs)
+            except Exception as e:
+                log.debug(f"Failed in OpenVINO execution: {e}")
+                return compile_fx(subgraph, *args)
```
slyalin, line 84: `compile_fx` returns a callable, but `_call` should return inference results, they should be tensors. You need run `to compile_fx` once and cache it for further usages. I think this like of code is not tested or I just misinterpreted this part completely.

suryasidd: Hey sergey, made the change according to our discussion.

mvafin (approved):

slyalin (approved):

PR #17331, opened 2023-05-03
[CPU][ARM] Enable oneDNN acl fullyconnected (inner_product)

EgorDuplensky:
### Details:
 - The reason why at the moment a oneDNN acl implementation is picked for fullyconnected node is the fallback logic we have for selecting optimal primitive descriptor. So acl implementation was not in the list and we picked the first supported primitive descriptor (which is acl implementation).
With the fix the node will actually find acl implementation in the priority list.
There is no impact on overall logic, but the fallback should be avoided in general.

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
+++ b/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
@@ -508,6 +508,7 @@ bool FullyConnected::created() const {
 const std::vector<impl_desc_type>& FullyConnected::getPrimitivesPriority() {
     std::vector<impl_desc_type> priorities = {
             impl_desc_type::unknown,
+            impl_desc_type::acl,
             impl_desc_type::brgemm_sparse_avx512_amx,
             impl_desc_type::brgemm_avx512_amx,
             impl_desc_type::brgemm_avx512,


EgorDuplensky:
@alvoron I think if we have not observed any issues with matmul primitive via benchdnn and all the shared tests are passed, we could try to merge this

alvoron:
> @alvoron I think if we have not observed any issues with matmul primitive via benchdnn and all the shared tests are passed, we could try to merge this

We aren't going to push this change to the release, are we?

EgorDuplensky:
> > @alvoron I think if we have not observed any issues with matmul primitive via benchdnn and all the shared tests are passed, we could try to merge this
> 
> We aren't going to push this change to the release, are we?

No, this change is not so important

EgorDuplensky:
@dmitry-gorokhov Could you please take a look?

alvoron (approved):

EgorDuplensky:
I propose to not merge it yet, want to double check with @dmitry-gorokhov

EgorDuplensky:
@dmitry-gorokhov Could you please take a look

dmitry-gorokhov (approved):

PR #17313, opened 2023-05-02
NMS-5 calculate upper-bound

jane-intel:
### Details:
Fix regression for legacy IRs with NMS-5 after: https://github.com/openvinotoolkit/openvino/pull/15853

PR to release branch: https://github.com/openvinotoolkit/openvino/pull/17332

PR #17332, opened 2023-05-03
NMS-5 calculate upper-bound

jane-intel:
### Details:
Fix regression for legacy IRs with NMS-5 after: https://github.com/openvinotoolkit/openvino/pull/15853

PR to master: https://github.com/openvinotoolkit/openvino/pull/17313


olpipi (approved):
LGTM

PR #17337, opened 2023-05-03
[PT FE] Support default strides for avg and max pooling

mvafin:
### Details:
 - *Support default strides for avg and max pooling*

### Tickets:
 - *107347*

### DIFF
--- a/src/frontends/pytorch/src/op/avg_poolnd.cpp
+++ b/src/frontends/pytorch/src/op/avg_poolnd.cpp
@@ -22,7 +22,14 @@ OutputVector translate_avg_poolnd(const NodeContext& context) {
     num_inputs_check(context, 6, 7);
     auto input = context.get_input(0);
     auto kernel = context.const_input<Shape>(1);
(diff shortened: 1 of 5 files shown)


src/frontends/pytorch/src/op/avg_poolnd.cpp
```diff
@@ -22,7 +22,14 @@ OutputVector translate_avg_poolnd(const NodeContext& context) {
     num_inputs_check(context, 6, 7);
     auto input = context.get_input(0);
     auto kernel = context.const_input<Shape>(1);
-    auto strides = context.const_input<Strides>(2);
```
rkazants, line 25: is `strides` always constant if provided? Can it be output of some operation?

rkazants (approved):

src/frontends/pytorch/src/op/avg_poolnd.cpp
mvafin: It can, but we didn't met such models.

PR #17351, opened 2023-05-04
[GPU] convolution primitive updates

vladimir-paramuzov:
### Details:
 - Removed most of `convolution` constructors. Only 3 c-tors left: regular, asymmetric and deformable
 - Added auto_pad parameter
 - Removed `pad` parameter, so `padding_begin` + `padding_end` are supposed to be used
 - Removed `output_size` parameter and `with_output_size` mode
 - Weights/bias ids have string type now instead of vector<string> as we don't support split anymore
 - Removed handle_input_padding pass
 - Use ngraph shape infer for legacy shape infer which enables asymmetric pad support + autopad
 - Related tests updates


src/plugins/intel_gpu/src/graph/graph_optimizer/handle_input_padding.cpp
```diff
@@ -1,129 +0,0 @@
-// Copyright (C) 2018-2023 Intel Corporation
```
yeonbok, line 1: Maybe I missed something, but is async padding is supported by many kernels? (AFAIK it isn't..is it?)
If only ref kernel support async padding, is it better to use ref kernel than opt kernel with borders?

vladimir-paramuzov: I think pad_end is not really needed in the kernels if we know output shape. Anyway, kernel selector parameters for convolution don't even have second parameter for pad, it means that none of the kernels use pad_end explicitly (even ref)

src/plugins/intel_gpu/src/graph/convolution.cpp
```diff
@@ -169,137 +91,70 @@ layout convolution_inst::calc_output_layout(convolution_node const& node, kernel
...
+        op.set_auto_pad(desc->auto_pad);
+        input_shapes.insert(std::next(input_shapes.begin()), impl_param.get_input_layout(1).get_partial_shape());
+        output_shapes = ov::op::v8::shape_infer(&op, input_shapes, pads_begin, pads_end);
+    } else if (desc->grouped_weights_shape || desc->groups > 1) {
+        ov::op::v1::GroupConvolution op;
+        op.set_dilations(dilation);
+        op.set_strides(strides);
+        op.set_auto_pad(desc->auto_pad);
+        auto& weights_shape = input_shapes[1];
+        if (input_shapes[1].size() == 4 && input_shapes[0].size() == 4) {
```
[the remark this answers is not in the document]
vladimir-paramuzov: That's actually WA for unit tests which create grouped weights w/o explicit group dimension in some cases (which needs to be fixed in the future). Added comment that explains motivation. So it may be applicable for 5d too, but as long as we don't have any failures caused by current version of WA, I'd prefer not to extend it.

src/plugins/intel_gpu/src/graph/convolution.cpp
```diff
@@ -169,137 +91,70 @@ layout convolution_inst::calc_output_layout(convolution_node const& node, kernel
...
+        op.set_auto_pad(desc->auto_pad);
+        output_shapes = ov::op::v1::shape_infer(&op, input_shapes, pads_begin, pads_end);
     }
-    return {output_type, out_fmt, output_size};
+
+    return {output_shapes[0], output_type, out_fmt};
 }
 
 template<typename ShapeType>
 std::vector<layout> convolution_inst::calc_output_layouts(convolution_node const& node, kernel_impl_params const& impl_param) {
```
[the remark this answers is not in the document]
vladimir-paramuzov: Extracted code to common function

src/plugins/intel_gpu/src/graph/graph_optimizer/handle_input_padding.cpp
yeonbok: I have thought that the reason why the handle_input_padding exists is because the kernels are not supporting asymmetric padding. But you mean that it is not true, right? If so the original pass was redundant.. I think it would be good to check the benchmark test.

sshlyapn (approved):

PR #17357, opened 2023-05-04
[CPU] gemm convolution: fixed bias offset

antonvor:
We added the group offset twice to the biases, as a result of which we could get accuracy errors or segfaults. It was noticed on the arm64 platform, because under x64 other primitives are mainly used.

**TODO**
- [x] need to add tests for gemm_convolution primitive with biases

PR to onednn fork: https://github.com/openvinotoolkit/oneDNN/pull/194

### DIFF
--- a/src/plugins/intel_cpu/tests/functional/single_layer_tests/group_convolution.cpp
+++ b/src/plugins/intel_cpu/tests/functional/single_layer_tests/group_convolution.cpp
@@ -383,6 +383,20 @@ INSTANTIATE_TEST_SUITE_P(smoke_GroupConv_1D_Gemm_FP32, GroupConvolutionLayerCPUT
                                 ::testing::Values(cpuEmptyPluginConfig)),
                         GroupConvolutionLayerCPUTest::getTestCaseName);
 
+INSTANTIATE_TEST_SUITE_P(smoke_GroupConv_1D_Gemm_with_bias_FP32, GroupConvolutionLayerCPUTest,
+                        ::testing::Combine(
+                                ::testing::Combine(
+                                        groupConvParams_ExplicitPadding_Gemm_1D,
+                                        ::testing::Values(ElementType::f32),
+                                        ::testing::Values(ElementType::undefined),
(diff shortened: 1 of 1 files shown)


maxnick:
@EgorDuplensky , could you please review?

EgorDuplensky (commented):
In general LGTM.
@antonvor As discussed, please double check single_layer_tests with valgrind if we still have unresolved issues after this fix.

src/plugins/intel_cpu/tests/functional/single_layer_tests/group_convolution.cpp
```diff
@@ -383,6 +383,20 @@ INSTANTIATE_TEST_SUITE_P(smoke_GroupConv_1D_Gemm_FP32, GroupConvolutionLayerCPUT
...
+                        ::testing::Combine(
+                                ::testing::Combine(
+                                        groupConvParams_ExplicitPadding_Gemm_1D,
+                                        ::testing::Values(ElementType::f32),
+                                        ::testing::Values(ElementType::undefined),
+                                        ::testing::Values(ElementType::undefined),
+                                        ::testing::ValuesIn(inShapesGemm1D),
+                                        ::testing::Values(CommonTestUtils::DEVICE_CPU)),
+                                ::testing::ValuesIn(filterCPUInfoForDevice(CPUParams_Gemm_1D)),
+                                ::testing::Values(fusingAddPerChannel),
```
EgorDuplensky, line 396: Do we really need to test both gemm_1D and gemm_1D_nspc?
Does it matter from the perspective of the issue we fix?
Applicable to all added instances.

EgorDuplensky (approved):

src/plugins/intel_cpu/tests/functional/single_layer_tests/group_convolution.cpp
antonvor: As part of this task, we only need tests for the nspc layout, but I would leave bias tests for the ncsp layout for testing coverage

dmitry-gorokhov (approved):

PR #17366, opened 2023-05-04
Check model file isn't empty before reading via FEs

vurusovs:
### Details:
 - TF FE `GraphIteratorProtoTxt::is_supported(model_path)` check may be passed if `model_path` is empty file or contains space character

### Tickets:
 - CVS-109938


vurusovs:
Closing due issue should be fixed on TF FE side

PR #17371, opened 2023-05-05
[CPU] Fix kernel precision mismatch in Reduce node

xuchen-intel:
### Details:
 - *The bug is introduced in by the last commit 44b7e448c3037a599fd321cdb6cb0936f9055796 of https://github.com/openvinotoolkit/openvino/pull/16829. The precision constraints imposed on the usage of auxiliary kernel (use_aux_kernel), should also be imposed on the optimization entrance (ReduceAll_opt).*

### Tickets:
 - *[CVS-110021](https://jira.devtools.intel.com/browse/CVS-110021)*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/reduce.cpp
+++ b/src/plugins/intel_cpu/src/nodes/reduce.cpp
@@ -1820,8 +1820,9 @@ void Reduce::initSupportedPrimitiveDescriptors() {
     }
 
     support_split = algorithm != Algorithm::ReduceL2 && algorithm != Algorithm::ReduceLogSumExp &&
-                    algorithm != Algorithm::ReduceSumSquare;
-    precision_change = input_prec == Precision::BF16 && output_prec == Precision::FP32;
+                    algorithm != Algorithm::ReduceSumSquare &&
+                    (input_prec == output_prec || (input_prec == Precision::BF16 && output_prec == Precision::FP32));
+    precision_change = input_prec != output_prec;
 
     src_data_size = input_prec.size();
     dst_data_size = output_prec.size();
--- a/src/plugins/intel_cpu/src/nodes/reduce.cpp
+++ b/src/plugins/intel_cpu/src/nodes/reduce.cpp
@@ -1819,10 +1819,10 @@ void Reduce::initSupportedPrimitiveDescriptors() {
         }
     }
 
+    precision_change = input_prec != output_prec;
     support_split = algorithm != Algorithm::ReduceL2 && algorithm != Algorithm::ReduceLogSumExp &&
                     algorithm != Algorithm::ReduceSumSquare &&
-                    (input_prec == output_prec || (input_prec == Precision::BF16 && output_prec == Precision::FP32));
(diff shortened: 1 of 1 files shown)


xuchen-intel:
@maxnick Hi Maksim, could you please take a look? Thanks!

src/plugins/intel_cpu/src/nodes/reduce.cpp
```diff
@@ -1820,8 +1820,9 @@ void Reduce::initSupportedPrimitiveDescriptors() {
     }
 
     support_split = algorithm != Algorithm::ReduceL2 && algorithm != Algorithm::ReduceLogSumExp &&
-                    algorithm != Algorithm::ReduceSumSquare;
-    precision_change = input_prec == Precision::BF16 && output_prec == Precision::FP32;
+                    algorithm != Algorithm::ReduceSumSquare &&
+                    (input_prec == output_prec || (input_prec == Precision::BF16 && output_prec == Precision::FP32));
+    precision_change = input_prec != output_prec;
```
maxnick, line 1826: Would be more logical to have the following:
```cpp
   precision_change = input_prec != output_prec;
   support_split = algorithm != Algorithm::ReduceL2 && algorithm != Algorithm::ReduceLogSumExp &&
                    algorithm != Algorithm::ReduceSumSquare &&
                    (!precision_change  || (input_prec == Precision::BF16 && output_prec == Precision::FP32));
```

maxnick:
@xuchen-intel , could you please add a SL test which covers the original issue?

src/plugins/intel_cpu/src/nodes/reduce.cpp
xuchen-intel: Good point! Applied. Thanks Maksim!

xuchen-intel:
> @xuchen-intel , could you please add a SL test which covers the original issue?

Thanks Maksim for comment! The cases of bf16-in-fp32-out precision have been indirectly covered by the cases of bf16-in-bf16-out in SL tests. Because as far as I know SLT infrastructure don't cover different input/output precisions directly now. It has also been discussed here https://github.com/openvinotoolkit/openvino/pull/16829#discussion_r1169430124. Further discussion is always welcomed.

maxnick (approved):

xuchen-intel:
> > @xuchen-intel , could you please add a SL test which covers the original issue?
> 
> Thanks Maksim for comment! The cases of bf16-in-fp32-out precision have been indirectly covered by the cases of bf16-in-bf16-out in SL tests. Because as far as I know SLT infrastructure don't cover different input/output precisions directly now. It has also been discussed here [#16829 (comment)](https://github.com/openvinotoolkit/openvino/pull/16829#discussion_r1169430124). Further discussion is always welcomed.

@maxnick According to our discussion, though the original PR is on bf16-in-fp32-out precision, the bug here is on int8-in-fp32-out precision. But, the current SL tests are not able to cover the latter. And to properly test int8->fp32 case we need a separate subgraph test with FQ inserted. That should be covered in a separate PR.

PR #17373, opened 2023-05-05
[TF FE] Provide single tensor names for inputs and outputs in SavedModel

rkazants:
**Details:** A lot of confusion is around multiple names we use for input and output tensor names in generated IRs for SavedModel format. Users suffered from multiple names that can happen after fusion multiple Identity nodes. Also, OVMS complains on it. So we made decision to clean-up any tensor name for inputs and outputs except the main names, i.e. user specific ones.

**Ticket:** 109671


tests/layer_tests/tensorflow2_keras_tests/test_tf2_keras_activation.py
```diff
@@ -50,8 +50,8 @@ def create_keras_activation_net(self, activation_func, input_names, input_shapes
              input_type=tf.float32),
         dict(activation_func="swish", input_names=["x1"], input_shapes=[[5, 4, 8]],
              input_type=tf.float32),
-        dict(activation_func="tanh", input_names=["x1"], input_shapes=[[5, 4, 8, 3]],
-             input_type=tf.float32),
+        pytest.param(dict(activation_func="tanh", input_names=["x1"], input_shapes=[[5, 4, 8, 3]],
+                          input_type=tf.float32), marks=pytest.mark.skip(reason="110006")),
```
jane-intel, line 54: what's wrong with this test as well as conv lstm one?

rkazants: Looks internal problem in TF itself. I will investigate this in the ticket.

jane-intel: So, this isn't connected to the changes you are making, right? Why bother including them in this PR then. A bit confusing.

src/frontends/tensorflow/tests/convert_saved_model.cpp
```diff
@@ -59,3 +88,29 @@ TEST_F(TransformationTestsF, SavedModelWithInputIntegerType) {
...
+
+TEST_F(TransformationTestsF, SavedModelMultipleTensorNames) {
+    // The test aims to check tensor names of input and output tensors
+    // it checks that TF FE preserved user specific names for input and output tensor
+    // and exclude internal names
+    {
+        model = convert_model("saved_model_parameter_result");
+
+        // check tensor names in the resulted model
+        unordered_set<string> tensor_names = {"params", "test_output_name"};
```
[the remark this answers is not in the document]
rkazants: Please pay attention to the generating script:

```
    @tf.function(input_signature=[tf.TensorSpec([20, 5], tf.float32)])
    def __call__(self, params):
        identity1 = tf.raw_ops.Identity(input=params)
        identity2 = tf.raw_ops.Identity(input=identity1)
        identity3 = tf.raw_ops.Identity(input=identity2)
        return {'test_output_name': identity3}
```

params - input, and `test_output_name` - output

tests/layer_tests/tensorflow2_keras_tests/test_tf2_keras_activation.py
rkazants: yes, right. We just changed interface of TF inference with named inputs and it looks broken there.

src/frontends/tensorflow/tests/convert_saved_model.cpp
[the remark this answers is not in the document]
jane-intel: I don't see where you provide params as string

dtrawins (approved):
Tested in OVMS: input and output names are as expected - compatible with TFS

src/frontends/tensorflow/tests/convert_saved_model.cpp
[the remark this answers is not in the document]
rkazants: Yes, it is listed in the signature of `__call__`:

```
def __call__(self, params):
```

jane-intel (approved):

PR #17379, opened 2023-05-05
Install OpenVINO™ Runtime fix for master

msmykx-intel:
Fix for sphinx directives in Install OpenVINO™ Runtime article


msmykx-intel:
Fixed in https://github.com/openvinotoolkit/openvino/pull/17353

PR #17380, opened 2023-05-05
Install OpenVINO™ Runtime fix for 23.0

msmykx-intel:
Fix for sphinx directives in Install OpenVINO™ Runtime article

Port from https://github.com/openvinotoolkit/openvino/pull/17379


msmykx-intel:
Fixed in https://github.com/openvinotoolkit/openvino/pull/17353

PR #17381, opened 2023-05-05
[TF frontend] use InterpolateMode::LINEAR_ONNX if input rank is 4

mateusztabaka:
This change mimicks LinearToLinearONNXReplacer transformation in legacy frontend, where linear interpolate mode is replaced with linear_onnx due to performance reasons.

Ticket: CVS-108343

### DIFF
--- a/src/frontends/tensorflow_common/src/op/interpolate.cpp
+++ b/src/frontends/tensorflow_common/src/op/interpolate.cpp
@@ -36,7 +36,12 @@ OutputVector translate_interpolate_op(const NodeContext& node) {
         interpolate_attrs.mode = Interpolate::InterpolateMode::NEAREST;
         interpolate_attrs.nearest_mode = Interpolate::NearestMode::FLOOR;
     } else if (op_type == "ResizeBilinear") {
-        interpolate_attrs.mode = Interpolate::InterpolateMode::LINEAR;
+        auto input_rank = images.get_partial_shape().rank();
+        if (input_rank.is_static() && input_rank.get_length() == 4) {
+            interpolate_attrs.mode = Interpolate::InterpolateMode::LINEAR_ONNX;
+        } else {
+            interpolate_attrs.mode = Interpolate::InterpolateMode::LINEAR;
+        }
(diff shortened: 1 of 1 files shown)


rkazants (changes requested):
what is the real reason of this change? Currently, it is functionally correct. Just to be aligned with the legacy this is not good argument because it did not have the layer tests before. So I don't think the legacy FE is ideal and we need to orient based on it.

mateusztabaka:
> what is the real reason of this change? Currently, it is functionally correct. Just to be aligned with the legacy this is not good argument because it did not have the layer tests before. So I don't think the legacy FE is ideal and we need to orient based on it.

As I mentioned in the description - the reason is performance. Take a look at https://github.com/openvinotoolkit/openvino/blob/master/tools/mo/openvino/tools/mo/back/LinearToLinearONNXReplacer.py#L11-L13, unfortunately that's still the case.

src/frontends/tensorflow_common/src/op/interpolate.cpp
```diff
@@ -36,7 +36,12 @@ OutputVector translate_interpolate_op(const NodeContext& node) {
         interpolate_attrs.mode = Interpolate::InterpolateMode::NEAREST;
         interpolate_attrs.nearest_mode = Interpolate::NearestMode::FLOOR;
     } else if (op_type == "ResizeBilinear") {
-        interpolate_attrs.mode = Interpolate::InterpolateMode::LINEAR;
+        auto input_rank = images.get_partial_shape().rank();
+        if (input_rank.is_static() && input_rank.get_length() == 4) {
+            interpolate_attrs.mode = Interpolate::InterpolateMode::LINEAR_ONNX;
+        } else {
+            interpolate_attrs.mode = Interpolate::InterpolateMode::LINEAR;
+        }
```
rkazants, line 44: Please create ticket for CPU and GPU to provide optimized kernel for it. And refer this ticket with explanation here in the code.

rkazants (approved):

PR #17399, opened 2023-05-08
Revert "[GPU] Fix proposal sort condition (#16981)"

hyunback:
Daily Perf regression was occurred because of changing sort condition.
mask_rcnn_resnet101_atrous_coco,int8
mask_rcnn_resnet50_atrous_coco,int8
faster_rcnn_inception_v2_coco,int8

This reverts commit bd8ca523b9661549bdeb1f6a6390feecdd1ab85c.

### Tickets:
 - *109855, 109680*


vladimir-paramuzov:
@hyunback Why we need to revert it? As I remember, this change was needed to fix accuracy

hyunback:
> @hyunback Why we need to revert it? As I remember, this change was needed to fix accuracy

Hi @vladimir-paramuzov 
I checked RFCN_ResNet101_coco with accuracy-checker, it showed almost same result between sort(revert PR) and partial sort. 
image
Considering Perf degradation, we need to change e2e test, so I made revert-PR first, and I'll discuss e2e validation team to this issue.
Note that,
image

vladimir-paramuzov:
@hyunback Which version is required by spec? I think we should align with it.

isanghao (approved):

PR #17409, opened 2023-05-08
Remove `funcTestUtils` from `ov_conditional_compilation_tests`

vurusovs:
### Details:
 - Because `ov_conditional_compilation_tests` is enabled via `ENABLE_TESTS`, so with `ENABLE_FUNCTIONAL_TESTS=OFF` build fails with "fatal error: openvino/itt.hpp: No such file or directory" due `funcTestUtils` dependency

### DIFF
--- a/src/common/conditional_compilation/tests/CMakeLists.txt
+++ b/src/common/conditional_compilation/tests/CMakeLists.txt
@@ -9,8 +9,10 @@ ov_add_test_target(
         ROOT ${CMAKE_CURRENT_SOURCE_DIR}
         DEPENDENCIES
         LINK_LIBRARIES
+            gtest
+            gtest_main
             gmock
-            funcTestUtils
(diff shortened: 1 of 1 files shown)


src/common/conditional_compilation/tests/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ ov_add_test_target(
         DEPENDENCIES
         LINK_LIBRARIES
             gmock
-            funcTestUtils
+            commonTestUtils
```
ilyachur, line 13: I don't see any dependency in code on these libraries. Can we remove this dependency?

vurusovs: Yes, it depends only from `openvino::conditional_compilation`

ilyachur (approved):

PR #17411, opened 2023-05-08
[MO] fix IR graph comparator

pavel-esir:
### Details:
 - masks "[0]" and "" should be treated as same for StridedSlice during graph comparrison

### Tickets:
 - xxx-98053

### DIFF
--- a/tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
+++ b/tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
@@ -127,7 +127,8 @@ def compare_graphs(graph: Graph, graph_ref: Graph, last_node: str, last_node_ref
                             num_insertions = rank - len(curr_node[mask_name])
                             curr_node[mask_name] = np.append(curr_node[mask_name], [0] * num_insertions).astype(int)
 
-                    if cur_node_type == 'StridedSlice' and node.in_node(1).shape is not None:
+                    if cur_node_type == 'StridedSlice' and node.in_node(1).has('shape') \
+                            and node.in_node(1).shape is not None:
                         from openvino.tools.mo.ops.strided_slice import StridedSlice
                         slice_rank = node.in_node(1).shape.item()
                         align_strided_slice_masks(node, slice_rank)
--- a/tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
+++ b/tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
@@ -116,6 +116,23 @@ def compare_graphs(graph: Graph, graph_ref: Graph, last_node: str, last_node_ref
                                       ''.format(node.id, cur_node_type, attr))
                         continue
 
+                    def align_strided_slice_masks(curr_node: Node, rank: int):
+                        for mask_name in StridedSlice.get_mask_names():
+                            if type(curr_node[mask_name]) == int:
+                                curr_node[mask_name] = [curr_node[mask_name]]
+                            elif type(curr_node[mask_name]) == str:  # if mask is an empty string ''
+                                assert len(curr_node[mask_name]) == 0
+                                curr_node[mask_name] = []
+
+                            num_insertions = rank - len(curr_node[mask_name])
+                            curr_node[mask_name] = np.append(curr_node[mask_name], [0] * num_insertions).astype(int)
+
+                    if cur_node_type == 'StridedSlice':
+                        from openvino.tools.mo.ops.strided_slice import StridedSlice
+                        slice_rank = node.in_node(1).shape.item()
+                        align_strided_slice_masks(node, slice_rank)
+                        align_strided_slice_masks(node_ref, slice_rank)
+
                     if attr == 'value':
                         if not values_are_equal(node.value, node_ref.value):
                             stderr.append('Current node "{}" with type {} and reference node "{}" with type "{}" have '
--- a/tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
+++ b/tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
@@ -127,12 +127,15 @@ def compare_graphs(graph: Graph, graph_ref: Graph, last_node: str, last_node_ref
                             num_insertions = rank - len(curr_node[mask_name])
                             curr_node[mask_name] = np.append(curr_node[mask_name], [0] * num_insertions).astype(int)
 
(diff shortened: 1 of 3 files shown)


tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
```diff
@@ -116,6 +116,24 @@ def compare_graphs(graph: Graph, graph_ref: Graph, last_node: str, last_node_ref
                                       ''.format(node.id, cur_node_type, attr))
                         continue
 
+                    def align_strided_slice_masks(curr_node: Node, rank: int):
```
mvafin, line 119: Nice to have comment about what this alignment means and why we do it for strided slice

tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
```diff
@@ -116,6 +116,24 @@ def compare_graphs(graph: Graph, graph_ref: Graph, last_node: str, last_node_ref
...
+                                curr_node[mask_name] = [curr_node[mask_name]]
+                            elif type(curr_node[mask_name]) == str:  # if mask is an empty string ''
+                                assert len(curr_node[mask_name]) == 0
+                                curr_node[mask_name] = []
+
+                            num_insertions = rank - len(curr_node[mask_name])
+                            curr_node[mask_name] = np.append(curr_node[mask_name], [0] * num_insertions).astype(int)
+
+                    if cur_node_type == 'StridedSlice' and node.in_node(1).has('shape') \
+                            and node.in_node(1).shape is not None:
```
mvafin, line 131: Do we really need to align masks for rank? I was thinking that we can only have [0] if we have empty "" mask

pavel-esir: >Do we really need to align masks for rank?

We could've also aligned one mask to another, e.g. 'begin_mask' of `node` to the same mask of `node_ref`; but aligning into rank seemed as an already known practice, we already use such thing for SS https://github.com/openvinotoolkit/openvino/blob/master/tools/mo/openvino/tools/mo/ops/strided_slice.py#L105-L111

>I was thinking that we can only have [0] if we have empty "" mask

Not necessary. Even if we have one empty mask "" other node could have `[0,0,0,0]` and they will behave the same. E.g. if slice rank is e.g. 4 and `begin_mask` for `node` is empty `""` and for `node_ref` is `[0,0,0,0]` they work exactly the same and they should be treated as equal during attrs comparison.

tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
pavel-esir: Done.

pavel-esir:
Added explanation comment. Also renamed `slice_rank` -> `input_rank` since `slice_rank` was wrong. @mvafin @rkazants please take a look.

note: i didn't add unit-tests because this comparison function is used in almost all of our replacer unit-tests and therefore we can be sure i didn't break anything. And also there wasn't separate unit-test for graph_compare.py, so it's not clear where and what should be add

mvafin (approved):

tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
```diff
@@ -116,6 +116,27 @@ def compare_graphs(graph: Graph, graph_ref: Graph, last_node: str, last_node_ref
                                       ''.format(node.id, cur_node_type, attr))
                         continue
 
+                    def align_strided_slice_masks(curr_node: Node, rank: int):
+                        for mask_name in StridedSlice.get_mask_names():
+                            if type(curr_node[mask_name]) == int:
```
rkazants, line 121:
```suggestion
                            if isinstance(curr_node[mask_name], int):
```

tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
```diff
@@ -116,6 +116,27 @@ def compare_graphs(graph: Graph, graph_ref: Graph, last_node: str, last_node_ref
...
+
+                            num_insertions = rank - len(curr_node[mask_name])
+                            curr_node[mask_name] = np.append(curr_node[mask_name], [0] * num_insertions).astype(int)
+
+                    # Need to align StridedSlice masks since such masks as [] and [0]; [] and [0,0]; [] and [0,0,0]
+                    # or [1] and [1,0]; [1] and [1,0,0] and so on for the input with rank 4 do exactly the same slicing and
+                    # should be treated as equal. Therefore, before attr comparison we align all masks to the input rank
+                    if cur_node_type == 'StridedSlice' and node.in_node(1).has('shape') \
+                            and node.in_node(1).shape is not None:
+                        from openvino.tools.mo.ops.strided_slice import StridedSlice
```
rkazants, line 135: Looks not needed import. Please remove it

tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
```diff
@@ -116,6 +116,27 @@ def compare_graphs(graph: Graph, graph_ref: Graph, last_node: str, last_node_ref
...
+                        input_rank = node.in_node(1).shape.item()
```
rkazants, line 136: Is it always valid? Can it be less than the real rank?

tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
pavel-esir: It is needed in the function `align_strided_slice_masks`. Moved to the place where it is used

tools/mo/openvino/tools/mo/utils/ir_engine/compare_graphs.py
pavel-esir: BTW, here i called it `input_rank` but indeed is `slice_rank`. I fixed by renaming this variable to `slice_rank`
https://github.com/pavel-esir/openvino/blob/master/tools/mo/openvino/tools/mo/ops/strided_slice.py#L67C8-L68


If you mean by the real rank the `input_rank` (rank of the main tensor on port(0)), yes it can be less that it. But we should rely on `slice_rank` not to `input_rank`. 

if `begin=[]` then slice_rank will be 0. 
image

I will add unit-tests and will ping when they are ready

pavel-esir: Unit-tests are ready. Indeed, `compare_graphs` is also called in almost every unit-test in python MO therefore we are sure that we didn't break anything

rkazants (approved):

PR #17421, opened 2023-05-08
DOCS Change optimization images

tsavina:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*

### DIFF
--- a/docs/_static/images/DEVELOPMENT_FLOW_V3_crunch.svg
+++ b/docs/_static/images/DEVELOPMENT_FLOW_V3_crunch.svg
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:40dfcea78a329e99d4bf2ead79678df96f1d392dcc4f278cf0b30ad3e8c7c795
-size 207081
+oid sha256:7a48c68fc6b2c138dac7a86cb2d1d3f7ed9fef5093435008b27aafcc91cf05f1
+size 251168
--- a/docs/_static/images/WHAT_TO_USE.svg
+++ b/docs/_static/images/WHAT_TO_USE.svg
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:285a60ebb9d6b7b4c5a8394e7c6eeed66ecf009f36b36a74c51bb28e6ee699ca
-size 271583
+oid sha256:9fb6c374add1b92d5d581b3f70de5caad51d63a748cde8d20645a71f4e78df69
+size 288421
--- a/docs/_static/images/workflow_simple.svg
+++ b/docs/_static/images/workflow_simple.svg
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:12af46d0d211361d9f195720329ee76f9d7c50d4ec8378cd33ef47595345795b
-size 32625
+oid sha256:84fc7114eef9ad310d72abc5d8f59b076d30031e0a42f18d518acc02e19bcc8d
+size 59755


MaximProshin:
@AlexKoff88 , please check updates in the diagrams

docs/_static/images/DEVELOPMENT_FLOW_V3_crunch.svg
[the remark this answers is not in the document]
AlexKoff88: `(val and train)` -> `(train and val)`. FYI, sometimes it's only train.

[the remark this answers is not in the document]
AlexKoff88: "Get full precision Model" -> "Framework Model"

docs/_static/images/workflow_simple.svg
AlexKoff88, line 1: I think NNCF should be mentioned instead of POT.

docs/_static/images/WHAT_TO_USE.svg
AlexKoff88, line 1: It is not obvious here when users should use Quantization and when Pruning and Sparsity.
First, I would remove the word Sparsity since we are pretty limited in running a sparse model, and replace it with quantization. So, the right block scheme should be named as "Pruning and Quantization".

tsavina: done

MaximProshin (approved):

docs/_static/images/workflow_simple.svg
AlexKoff88, line 1: I cannot find where this diagram is used. Is this a POT-specific picture? The reason why I am asking is that if it is a common picture that POT should be replaced with NNCF.

tsavina: yes. sorry I might have accidentally closed the comment with explanation
it fixes POT description page  https://docs.openvino.ai/latest/pot_introduction.html

yury-gorbachev:
Within this document (tmo_introduction.html) there is diagram that has model optimizer. We probably want to change it to convert modes or something similar. Not sure if this is covered by separate PR. cc: @slyalin

docs/_static/images/DEVELOPMENT_FLOW_V3_crunch.svg
slyalin, line 1: IR is just a serialized representation for OpenVINO model. As we are becoming more Pythonic, and moving away from CLI tool, here is a better option for names:
- `Convert to OpenVINO IR` -> `Convert to OpenVINO Model`
- `Full precision IR` -> `Full precision OpenVINO Model`
- `Optimized IR` -> `Optimized OpenVINO Model`

And, probably in this particular diagram, to emphasize separation between `Development` and `Deployment` the latest step in `Development` should be serialization of `Optimized OpenVINO Model` to `Optimized OpenVINO IR`. Some additional tuning for names is still required.

docs/_static/images/workflow_simple.svg
slyalin, line 1: If you change `Run Model Optimizer` to `Convert to OpenVINO IR` in one of the pictures, then apply similar changes to this one.

slyalin:
> Within this document (tmo_introduction.html) there is diagram that has model optimizer. We probably want to change it to convert modes or something similar. Not sure if this is covered by separate PR. cc: @slyalin

Don't know about another PR. Adding @kblaszczak-intel and @andrei-kochin. The changes haven't been prepared yet. In other comments in this PR, I recommended updating pictures immediately as it has been already partially done.

docs/_static/images/DEVELOPMENT_FLOW_V3_crunch.svg
helena-intel: This image says you need to convert your model to IR before you "optimize with NNCF" (in the PTQ part). This is not true for PTQ with NNCF (it was for POT).

docs/_static/images/WHAT_TO_USE.svg
helena-intel: From my perspective this diagram now adds more confusion than that it resolves it. And it doesn't answer much. The flowcharts also only have yes. What if my answer is no? Also I usually have val & train datasets, a training pipeline and a PyTorch model (everything from the "pruning and quantization" column). But I rarely use QAT because for my models that tends to require 12 hours or more on a machine with specific hardware.

akladiev:

kblaszczak-intel (approved):

PR #17422, opened 2023-05-08
[ARM CPU] Add ACL Normalize L2 kernel

allnes:
- separate executors
- add acl normalize l2 kernel


akladiev:

PR #17434, opened 2023-05-09
Align version of urllib3 for all components

ilyachur:
### Details:
 - Use version 1.26.5

### Tickets:
 - CVS-110138

### DIFF
--- a/tools/constraints.txt
+++ b/tools/constraints.txt
@@ -17,6 +17,6 @@ pylint>=2.7.0
 pyenchant>=3.0.0
 test-generator==0.1.1
 py>=1.9.0
-urllib3>=1.26.4
+urllib3==1.26.5
 openvino-telemetry>=2022.1.0
 opencv-python>=4.5


tools/constraints.txt
```diff
@@ -17,6 +17,6 @@ pylint>=2.7.0
 pyenchant>=3.0.0
 test-generator==0.1.1
 py>=1.9.0
-urllib3>=1.26.4
+urllib3==1.26.5
```
rkazants, line 20: I don't think to set exact version for the module is good. For some platform, this exact version can be absent and such change is not good for the future.
May be it is better to use markers to differentiate the exact platform.

ilyachur: We already use such approach with version 1.26.5 in other components (for example docs).
 now we have a problem with the newest version of urllinb3

rkazants: docs are not a reference for us because no one user needs docs building.
Which tool is using this module? May be we can remove this at all? Alternatively, we can check `ssl` module that seems outdated installed in CI and should be installed explicitly then.

rkazants: I see that the issue is raised on Python 3.8. So let us make this change for this particular version of Python, for others use the newest one. It seems that there is no compiled ssl module on Python 3.8

PR #17445, opened 2023-05-09
[ARM CPU] ACL TBB scheduler

allnes:


src/plugins/intel_cpu/src/nodes/executors/acl/acl_ie_scheduler.cpp
```diff
@@ -0,0 +1,62 @@
...
+        run_workloads(workloads);
+    }
+}
+
+void ACLScheduler::run_workloads(std::vector<arm_compute::IScheduler::Workload> &workloads) {
+    InferenceEngine::parallel_for(workloads.size(), [&] (int wid) {
+        ThreadInfo info;
+        info.cpu_info    = &cpu_info();
+        info.num_threads = parallel_get_num_threads();
+        info.thread_id   = parallel_get_thread_num();
```
xufang-lisa, line 56:
```suggestion
        info.thread_id   = wid;
```
I think thread_id here should be guaranteed to be unique, am I right? I tested hello_classification with resnet-50 model and found that using wid as thread_id returned the same output.

allnes: @xufang-lisa it's work, thank you

wangleis (approved):

src/plugins/intel_cpu/src/nodes/executors/acl/acl_ie_scheduler.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+void ACLScheduler::set_num_threads(unsigned int num_threads) {}
+
+void ACLScheduler::schedule(ICPPKernel *kernel, const Hints &hints) {
+    ITensorPack tensors;
+    schedule_common(kernel, hints, kernel->window(), tensors);
+}
+
+void ACLScheduler::schedule_op(ICPPKernel *kernel, const Hints &hints, const Window &window, ITensorPack &tensors) {
+    schedule_common(kernel, hints, kernel->window(), tensors);
```
xufang-lisa, line 29:
```suggestion
    schedule_common(kernel, hints, window, tensors);
```
@allnes I think here input parameter "window" should be used instead of kernel->window(). After adding this modification, I tested "hg-s8-b1-mpii" using hello_classification and there is no accuracy issue.

allnes: thank you, launched accuracy checker pipeline

alvoron:
Should we keep scheduler in executors folder?

dmitry-gorokhov:
> Should we keep scheduler in executors folder?

I think current place is good enough until we will not have dedicated place with per-target global configuration logic

dmitry-gorokhov:
@allnes Please rebase on latest master to get fresh CI.

PR #17449, opened 2023-05-10
[DOCS] operation support changes

kblaszczak-intel:


sgolebiewski-intel (approved):

msmykx-intel (approved):

msmykx-intel (approved):

rkazants (changes requested):
MO team needs to review this

docs/resources/Supported_Devices.md
```diff
@@ -0,0 +1,173 @@
...
+Supported Precision
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+==================  ===========================  ============================  ==========================
+ Plugin              FP32                         FP16                          I8                              
+==================  ===========================  ============================  ========================== 
+ CPU plugin          Supported and preferred      Supported                     Supported                                                        
+ GPU plugin          Supported                    Supported and preferred       Supported                                                        
+ GNA plugin          Supported                    Supported                     Not supported                                                    
+ Arm® CPU plugin     Supported and preferred      Supported                     Supported (partially)
```
rkazants, line 99: we no longer have ARM plugin after merging it into CPU plugin, should we have it here? @dmitry-gorokhov, please advice

docs/resources/supported_operations.md
```diff
@@ -0,0 +1,1072 @@
...
+   See the full conformance report table
+
+
+.. tab-set::
+
+   .. tab-item:: Operations Alphabetically
+
+      =================================  ==============  ===============  ================  ==================
+       Operations                         GPU             CPU              GNA               Arm® CPU    
+      =================================  ==============  ===============  ================  ==================
```
rkazants, line 22: May be the order should be CPU, GPU, GNA? @dmitry-gorokhov, @vladimir-paramuzov, please advice. Also, shoul we have ARM here?

docs/resources/supported_operations.md
```diff
@@ -0,0 +1,1072 @@
...
+       Upsampling                         Supported       Not Supported    Not Supported     Not Supported   
+      =================================  ==============  ===============  ================  ==================
+      
+      |   `*` - support is limited to the specific parameters. Refer to "Known Layer Limitations" section for the device :doc:`from the list of supported <openvino_docs_OV_UG_supported_plugins_Supported_Devices>`.
+      |   `**` - support is implemented via :doc:`Extensibility mechanism <openvino_docs_Extensibility_UG_Intro>`.
+      |   `***` - supports NCDHW layout.
+      |   `****` - support is implemented via runtime reference.
+
+
+   .. tab-item:: Operations by Frameworks
```
rkazants, line 175: let us decouple `operations by framework` into a separate document `supported_framework_operations.md` just to avoid mixing our opset operations and frameworks' ones.

docs/resources/supported_operations.md
```diff
@@ -0,0 +1,1072 @@
...
+
+      .. tab-set::
+      
+         .. tab-item:: Caffe
```
rkazants, line 179: Also, I would like to change the order of frameworks here and put deprecated frameworks in the last order (Caffe, Kaldi, MXNet). So the final order should be: TensorFlow, TensorFlow 2 Keras, TensorFlow Lite, ONNX, PaddlePaddle, Caffe, Kaldi, MXNet.

@mvafin, will you put PyTorch supported operations into this doc?

rkazants (commented):
please do not change thir-party deps: onednn

docs/resources/supported_operations.md
kblaszczak-intel: The idea here is to aggregate information on operations in general in one place, possibly combining it with the IR and OpSets section in the future.
Currently, operations by framework are in a separate article but it is still not clear what the article presents exactly - are these ops supported by frameworks, or by these frontends in OV, or are these OV-supported layers pertaining to specific frameworks? 
Do you think we could keep these two tables in one article and provide a clear distinction between the two by proper descriptions? This way the information would be all in one place and clearly explained.

docs/resources/supported_operations.md
mvafin: Yes, seems like a good idea

docs/resources/Supported_Devices.md
```diff
@@ -0,0 +1,173 @@
...
+
+.. note::
+
+   With OpenVINO™ 2023.0 release, support has been cancelled for all VPU accelerators based on Intel® Movidius™.
+
+
++---------------------------------------------------------------------+----------------------------------------------------------------------------------------------------+
+| OpenVINO Device                                                     | Supported Hardware                                                                                 |
++=====================================================================+====================================================================================================+
+|| :doc:`GPU <openvino_docs_OV_UG_supported_plugins_GPU>`             | Intel® Processor Graphics, including Intel® HD Graphics and Intel® Iris® Graphics                  |
```
vladimir-paramuzov, line 18: Could you add [ARC](https://www.intel.com/content/www/us/en/products/details/discrete-gpus/arc/arc-a-series/products.html), [FLEX](https://www.intel.com/content/www/us/en/products/details/discrete-gpus/data-center-gpu/flex-series.html) and [MAX](https://www.intel.com/content/www/us/en/products/details/discrete-gpus/data-center-gpu/max-series.html) series to supported HW list?

docs/resources/supported_operations.md
vladimir-paramuzov: I think the devices order is not that important. What really matters is that operation names and supported flags are outdated. To my mind we need to update this table based on the latest conformance report.

docs/resources/Supported_Devices.md
kblaszczak-intel: I believe that we should not list it if it does not function as an independent plugin. But we should add information to the CPU plugin that it supports Arm. What do you think?

docs/resources/supported_operations.md
kblaszczak-intel: I think it is a good idea. The order will reflect the ops coverage - most covered first, least covered last... I don't know about Arm, but since it differs from the standard CPU I would leave it. Just change the name to "CPU (Arm®)"

docs/resources/supported_operations.md
rkazants: I am leaning not to have long document and decouple framework operations support by FEs and opset operations support by the plugins.

docs/resources/Supported_Devices.md
```diff
@@ -0,0 +1,174 @@
...
+=================  ===========  ===============  ==============  ===============  ==============  =================
+ Plugin             FP32         FP16             U8              U16              I8              I16            
+=================  ===========  ===============  ==============  ===============  ==============  =================
+ CPU plugin         Supported    Supported        Supported       Supported        Supported       Supported      
+ GPU plugin         Supported    Supported*       Supported*      Supported*       Not supported   Supported*    
+ GNA plugin         Supported    Not supported    Supported       Not supported    Supported       Supported      
+ Arm® CPU plugin    Supported    Supported        Supported       Supported        Not supported   Not supported  
+=================  ===========  ===============  ==============  ===============  ==============  =================
+
+'*' - Supported via ``SetBlob`` only, ``GetBlob`` returns FP32
```
vladimir-paramuzov, line 120: I think this line is outdated.
1. It refers to the old API
2. 2.0 API expects that FP32 precision to be used for all floating-point inputs/outputs. Usage of another data type requires explicit preprocessing pipeline setup and after that both get_tensor() and set_tensor() works with same data type.

docs/resources/supported_operations.md
```diff
@@ -0,0 +1,1316 @@
...
+      |   `*` - support is limited to the specific parameters. Refer to "Known Layer Limitations" section for the device :doc:`from the list of supported <openvino_docs_OV_UG_supported_plugins_Supported_Devices>`.
+      |   `**` - support is implemented via :doc:`Extensibility mechanism <openvino_docs_Extensibility_UG_Intro>`.
+      |   `***` - supports NCDHW layout.
+      |   `****` - support is implemented via runtime reference.
+
+
+
+
+
+   .. tab-item:: Operations Supported by Other Frameworks
```
rkazants, line 178: It is not correct to call it as `Operations Supported by Other Frameworks`.
Please separate into two documents: `Operations supported by OpenVINO plugins` and `Operations supported by OpenVINO Framework Frontend`

docs/resources/supported_operations.md
```diff
@@ -0,0 +1,1316 @@
...
+            torchvision::deform_conv2d
+            torchvision::nms
+            torchvision::roi_align
+            ==========================================  ==========================================================================================
+
+         .. tab-item:: TensorFlow
+
+            Some of TensorFlow operations do not match any OpenVINO operations. Yet, they are still supported by 
+            the conversion API (previously Model Optimizer) and can be used on the constant propagation path. These layers are labeled 
+            with ``Constant propagation`` in the table below:
```
rkazants, line 426: Please remove these lines. This is not longer valid since TF FE release. All limitations are described per operation.

rkazants (changes requested):

docs/resources/supported_operations_inference devices.md
```diff
@@ -0,0 +1,171 @@
+# Supported Operations - by Inference Devices {#openvino_resources_supported_operations}
```
rkazants, line 1: you seem to have a white-space in the document name

rkazants (commented):
Please recover third-party dependencies. Other stuff looks good to me

rkazants (approved):
From MO/Frontend perspective looks good

PR #17459, opened 2023-05-10
HOT FIX: disable set_cpu_used in master branch

wangleis:
### Details:
 - *disable set_cpu_used in master branch*

### Tickets:
 - *110077*


dmitry-gorokhov (approved):

dmitry-gorokhov:
@wangleis Is there any way how CPU pinninh behavior can be tests. Like we create behavior tests that run pipeline of two different models. Is very any way to extract information about pinning from StreamExecutor or from the process itself?

wangleis:
@dmitry-gorokhov So far e2e CPU pinning behavior validation is relied on e2e performance test.

Now we only enabled smoke test for CPU mapping detection and streams scheduling during multi-threading 2.0 improvement. Will investigate smoke test for e2e CPU pinning behavior validation in following release.

PR #17474, opened 2023-05-11
[ShapeInfer] Review experimental detectron prior grid generator class for shape inference aspects

praasz:
### Details:
- Review interval shapes and label propagation.
- Review template implementation of shape infer.
- Add update or correct unit test for static and dynamic shapes.
- Remove `ngraph` namespace from operator.

### Tickets:
 - 97225
 - 97313
 - 97362
 - 97411


src/core/shape_inference/include/experimental_detectron_prior_grid_generator_shape_inference.hpp
```diff
@@ -41,22 +42,21 @@ void shape_infer(const ExperimentalDetectronPriorGridGenerator* op,
...
 
-        NODE_VALIDATION_CHECK(op,
-                              num_batches_featmap.compatible(num_batches_im_data),
-                              "The first dimension of both 'feature_map' and 'im_data' must match. "
-                              "Feature_map: ",
-                              num_batches_featmap,
-                              "; Im_data: ",
-                              num_batches_im_data);
+            NODE_VALIDATION_CHECK(op,
+                                  num_batches.compatible(1),
```
mitruska, line 50: This check is aligned with spec, but can break backward compatibility.
(Previously possible equality to 1 wasn't validated).

praasz: Yes it looks like current tests have got different assumption and check against dimension `1` is too strict. 
Validation updated.

mitruska (approved):

jane-intel (approved):

PR #17482, opened 2023-05-11
[GPU] enables ImportNetwork with remote_context API

e-ddykim:
### Details:
 - This PR enables the `ImportNetwork` with remote_context API.

### Tickets:
 - 110532

### DIFF
--- a/src/plugins/intel_gpu/src/plugin/plugin.cpp
+++ b/src/plugins/intel_gpu/src/plugin/plugin.cpp
@@ -352,16 +352,6 @@ QueryNetworkResult Plugin::QueryNetwork(const CNNNetwork& network,
     return res;
 }
 
-InferenceEngine::IExecutableNetworkInternal::Ptr Plugin::ImportNetwork(const std::string& modelFileName,
-                                                                       const std::map<std::string, std::string>& config) {
-    if (FileUtils::fileExist(modelFileName)) {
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/src/plugin/plugin.cpp
```diff
@@ -352,19 +352,38 @@ QueryNetworkResult Plugin::QueryNetwork(const CNNNetwork& network,
     return res;
 }
 
+InferenceEngine::IExecutableNetworkInternal::Ptr Plugin::ImportNetwork(const std::string& modelFileName,
```
vladimir-paramuzov, line 355: Why is it needed? Doesn't base impl work?
```
std::shared_ptr<IExecutableNetworkInternal> IInferencePlugin::ImportNetwork(
    const std::string& modelFileName,
    const std::map<std::string, std::string>& config) {
    std::ifstream blobFile(modelFileName, std::ios::binary);

    if (!blobFile.is_open()) {
        IE_THROW(NetworkNotRead);
    }

    return ImportNetwork(blobFile, config);
}
```

e-ddykim: Oh, I'll remove it. Thank you so much!

yeonbok (approved):

PR #17478, opened 2023-05-11
Add transformation to convert adaptive pool to reduce

mvafin:
### Details:
 - *Add transformation to convert adaptive pool to reduce. That fixes performance problems in efficentnet on older HW*

### Tickets:
 - *108400*

PR #17488, opened 2023-05-11
Add transformation to convert adaptive pool to reduce

mvafin:
### Details:
 - *Add #17478 to master.*
 - *Add transformation to common pipeline*

### Tickets:
 - *108400*


itikhono (approved):

src/plugins/intel_cpu/tests/functional/single_layer_tests/adaptive_pooling.cpp
```diff
@@ -248,7 +248,7 @@ const std::vector<ElementType> netPrecisions = {
 };
 
 const std::vector<std::vector<int>> pooled3DVector = {
-        { 1 },
+        //{ 1 }, // AdaptivePool will be replace with reduce with AdaptivePoolToReduce transformation
```
mvafin, line 251: @dmitry-gorokhov please look on this change

src/common/transformations/include/transformations/common_optimizations/adaptive_pool_to_reduce.hpp
```diff
@@ -0,0 +1,29 @@
...
+namespace pass {
+
+class TRANSFORMATIONS_API AdaptivePoolToReduce;
+
+}  // namespace pass
+}  // namespace ov
+
+/**
+ * @ingroup ie_transformation_common_api
+ * @brief AdaptivePoolToReduce transformation replaces AdaptiveXXXPool with ReduceXXX when possible
```
rkazants, line 22: I think it makes sense to add use cases for this transformation here.

src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
```diff
@@ -205,6 +206,7 @@ bool ov::pass::MOCTransformations::run_on_model(const std::shared_ptr<ngraph::Fu
     ADD_MATCHER(common_fusions, DepthToSpaceFusion)
     ADD_MATCHER(common_fusions, ShuffleChannelsFusion, !m_use_shapes)
     ADD_MATCHER(common_fusions, NonZeroHorizontalFusion)
+    ADD_MATCHER(common_fusions, AdaptivePoolToReduce)
```
rkazants, line 209: I would recommend to add one integration test case for sub-graph with Adaptive pooling to ensure that IR will contain Reduce. It will prevent performance degradation in the future for this case.

rkazants: Also, is NNCF tool fine with this change and about other plugins? May be we need to switch this transformation precisely only for CPU?

dmitry-gorokhov (approved):
Approved CPU plugin changes

PR #17489, opened 2023-05-11
[PT FE]: support aten::argmax and aten::argmin

eaidova:
### Details:
 - *support aten::argmax and aten::argmin operators*

### Tickets:
 - *TBD*


src/frontends/pytorch/src/op/argmax_argmin.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto input = context.get_input(0);
+    bool keep_dims = false;
+    auto k = context.mark_node(v0::Constant::create(element::i32, Shape{}, {1}));
+    Output<Node> indices;
+    if (!context.input_is_none(2)) {
+        keep_dims = context.const_input<bool>(2);
+    }
+    if (!context.input_is_none(1)) {
+        auto axis = context.const_input<int64_t>(1);
+        auto topk = context.mark_node(std::make_shared<v3::TopK>(input, k, axis, mode, TopKSortType::NONE));
```
rkazants, line 34: Please pay attention to the case with repeating elements. If the same max element is repeating several times. what index should be reported? You need to properly select `sort` and `stable` attributes for this. Also please add explicit tests case for this for which you need to use special data generator

mvafin (changes requested):
Please add test for stable/unstable case

src/frontends/pytorch/src/op/argmax_argmin.cpp
eaidova: first index should be reported, my parameters aligned with behavior for PyTorch onnx export, test for repeated values added

src/frontends/pytorch/src/op/argmax_argmin.cpp
```diff
@@ -0,0 +1,70 @@
...
+        indices = context.mark_node(std::make_shared<v0::Convert>(topk->output(1), element::i64));
+        if (!keep_dims) {
+            auto axis_to_remove = context.mark_node(v0::Constant::create(element::i32, Shape{}, {axis}));
+            indices = context.mark_node(std::make_shared<v0::Squeeze>(indices, axis_to_remove));
+        }
+    } else {
+        int64_t axis = 0;
+        auto munus_one = context.mark_node(v0::Constant::create(element::i32, Shape{1}, {-1}));
```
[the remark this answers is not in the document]
eaidova: fixed

tests/layer_tests/pytorch_tests/test_argmax_argmin.py
```diff
@@ -0,0 +1,80 @@
...
+        (3, False),
+        (3, True),
+        (-3, False),
+        (-3, True),
+        (-2, False),
+        (-2, True),
+        (-1, False),
+        (-1, True)])
+    @pytest.mark.parametrize("op_type", ['max', 'min'])
+    @pytest.mark.parametrize("repeat_values", [True, False])
```
mvafin, line 73: I don't think we need tests for both stable and unstable cases, especially if both work fine. Non-repeat tests just increase precommit time without real need.

eaidova: removed

mvafin (approved):

PR #17490, opened 2023-05-11
[CPU] Fix bug with reshape on empty tensor CPU Plugin

mmikolajcz:
### Details:
 - *Fix bug when reshape is called with empty tensor containing 0 in shape*
 - *https://github.com/openvinotoolkit/openvino/pull/17014*

### Tickets:
 - *ticket-id*


tiger100256-hu:
@mmikolajcz  thanks very much for your help to fix this bug, I also tried to add zero shape test case like this PR in reshape, but it threw error that didn't support 0 dims, so at last I added a test case (nonzero + reshape) , nonzero created 0 shape as the input of reshape op. it's strange.

tiger100256-hu (approved):

PR #17505, opened 2023-05-12
[Core] StridedSlice improvements of bound evaluation and constant folding

praasz:
### Details:
- Perform bound evaluation when begin, end inputs got partial value but mask for corresponding input is set and this value is ignored.
- Implement custom fold if default not work to check `StridedSlice` special case where begin and end values are ignored but mask but they are not constants.

### Tickets:
 -  101315

### DIFF
--- a/src/core/src/op/strided_slice.cpp
+++ b/src/core/src/op/strided_slice.cpp
@@ -66,6 +66,17 @@ shared_ptr<Node> calculate_default_strides(const Output<Node>& begin, const Outp
 
     return op::Constant::create(element::i64, ov::Shape{strides_length}, vector<int64_t>(strides_length, 1));
 }
+
+/**
+ * @brief Check if all indicies in 1-D input shape are ignored by masks.
+ *
+ * @param shape        Indicies shape (assume compatible 1-D shape).
+ * @param ignored_mask Axis set of ignored indicies.
+ * @return True if all ignored other wise false.
+ */
+bool all_indicies_ignored(const ov::PartialShape& shape, const AxisSet& ignore_mask) {
+    return shape.rank().is_static() && ov::cmp::le(shape[0].get_interval().get_max_val(), ignore_mask.size());
+}
 }  // namespace
 
 op::v1::StridedSlice::StridedSlice(const Output<Node>& data,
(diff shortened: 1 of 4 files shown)


src/core/src/op/strided_slice.cpp
```diff
@@ -237,27 +238,54 @@ bool op::v1::StridedSlice::has_evaluate() const {
...
+
+    if (lb_t && ub_t) {
+        using TCast = int64_t;
+        constexpr auto i64_cast = ov::util::Cast<TCast>();
+        const auto lb = ov::get_tensor_data_as<TCast>(lb_t, i64_cast);
+        const auto ub = ov::get_tensor_data_as<TCast>(ub_t, i64_cast);
+        const auto mask_set = convert_mask_to_axis_set(masks);
+
+        size_t axis = 0;
+        return std::equal(lb.cbegin(), lb.cend(), ub.cbegin(), [&axis, &mask_set](TCast lhs, TCast rhs) -> bool {
```
jane-intel, line 254: should it be const references instead?
```suggestion
        return std::equal(lb.cbegin(), lb.cend(), ub.cbegin(), [&axis, &mask_set](const TCast& lhs, const TCast& rhs) -> bool {
```

itikhono: it's ok for int64_t type, ref is not needed

itikhono (approved):

src/core/src/op/strided_slice.cpp
```diff
@@ -237,27 +238,54 @@ bool op::v1::StridedSlice::has_evaluate() const {
...
-namespace {
-bool strided_slice_input_check(const ov::Node* node) {
-    if (!node->get_input_tensor(1).has_and_set_bound() || !node->get_input_tensor(2).has_and_set_bound() ||
-        !node->get_input_tensor(3).has_and_set_bound())
+bool op::v1::StridedSlice::indicies_input_has_and_set_bounds(const size_t port,
+                                                             const std::vector<int64_t>& masks) const {
+    const auto& lb_t = get_input_tensor(port).get_lower_value();
+    const auto& ub_t = get_input_tensor(port).get_upper_value();
+
+    if (lb_t && ub_t) {
```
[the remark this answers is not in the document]
praasz: Good catch, just missed that case. Updated with solution and unit test for that case.

However I had to also modify the `constant_fold` function in case when there is Parameter{3} and bit mask are defined for all dimension it looks like this input must be replaced by dummy constant to correctly process bound evaluation. Wihtout it the `could_propagate` in the evaluation will return false and evaluates of `StridedSlice` will be not executed.

If required I making clone of op with replaced inputs and then try get it as constant.

src/core/src/op/strided_slice.cpp
praasz: Using reference for fundamental types can cause indirect access which can be slower than direct (by value) access. And pass fundamental type by value shouldn't be issue.

jane-intel (approved):

PR #17510, opened 2023-05-12
[Spec][Pad-12] Pad-12 op specification

mitruska:
### Details:
 - Specification of Pad-12 operation
 - The only change comparing to Pad-1 is official support for negative pads (negative values in pad_begin and pad_end inputs),
 it allows to remove elements from the corresponding dimension.
 
- *[TBD] new pad_mode "wrap"

### Tickets:
 - 110239


docs/ops/movement/Pad_12.md
```diff
@@ -0,0 +1,261 @@
...
+
+**Inputs**
+
+* **1**: `data` tensor of arbitrary shape and type *T*. **Required.**
+
+* **2**: `pads_begin` 1D tensor of type *T_INT*. Number of elements matches the shape rank of *data* input. Specifies the number of padding elements to add at the beginning of each axis. Negative value means removal of the corresponding dimensions. **Required.**
+
+* **3**: `pads_end` 1D tensor of type *T_INT*. Number of elements matches the shape rank of *data* input. Specifies the number of padding elements to add at the ending of each axis. Negative value means removal of the corresponding dimensions. **Required.**
+
+* **4**: `pad_value` scalar tensor of type *T*. Used with the `pad_mode = "constant"` only. All new elements are populated with this value or with 0 if input not provided. Shouldn't be set for other `pad_mode` values. **Optional.**
```
tomdol, line 151:
```suggestion
* **4**: `pad_value` scalar tensor of type *T*. Takes effect only if `pad_mode == "constant"` only. All padding elements are populated with this value or with 0 if the input is not provided. This input should not be set with other values of `pad_mode`. **Optional.**
```

docs/ops/movement/Pad_12.md
```diff
@@ -0,0 +1,261 @@
...
+
+
+**Outputs**
+
+* **1**: Output padded tensor of type *T* with dimensions `pads_begin[D] + data.shape[D] + pads_end[D]` for each `D` from `0` to `len(data.shape) - 1`.
```
tomdol, line 156:
```suggestion
* **1**: Padded output tensor of type *T* with dimensions `pads_begin[D] + data.shape[D] + pads_end[D]` for each `D` from `0` to `len(data.shape) - 1`.
```

a-sidorova, line 156: I'm not sure but maybe it should smth like this:
```suggestion
* **1**: Output padded tensor of type *T* with dimensions `max(pads_begin[D] + data.shape[D] + pads_end[D], 0)` for each `D` from `0` to `len(data.shape) - 1`.
```

I think the case with `shape[5, 7], pads_begin[-3,1], pads_end[-3,1]` might be in dynamic scenario at least. What output shape should be here? Seems like `shape[0, 9]` not `shape[-1, 9]`. Could you please clarify this moment? Thanks!

tomdol (changes requested):

docs/ops/movement/Pad_12.md
mitruska: Thank you for this comment. Good point, I've updated the spec with `max` according to the suggestion to avoid negative dimensions.

docs/ops/movement/Pad_12.md
```diff
@@ -0,0 +1,261 @@
...
+  * **Type**: `string`
+  * **Required**: *yes*
+
+**Inputs**
+
+* **1**: `data` tensor of arbitrary shape and type *T*. **Required.**
+
+* **2**: `pads_begin` 1D tensor of type *T_INT*. Number of elements matches the shape rank of *data* input. Specifies the number of padding elements to add at the beginning of each axis. Negative value means removal of the corresponding dimensions. **Required.**
+
+* **3**: `pads_end` 1D tensor of type *T_INT*. Number of elements matches the shape rank of *data* input. Specifies the number of padding elements to add at the ending of each axis. Negative value means removal of the corresponding dimensions. **Required.**
```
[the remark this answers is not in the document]
mitruska: Sure, updated.

a-sidorova (approved):

docs/ops/movement/Pad_12.md
```diff
@@ -0,0 +1,261 @@
...
+    * `reflect` - padded values are a reflection of the input `data` tensor. Values on the edges are not duplicated, `pads_begin[D]` and `pads_end[D]` must be not greater than `data.shape[D] – 1` for any valid `D`.
+    * `symmetric` - padded values are symmetrically added from the input `data` tensor. This method is similar to the `reflect`, but values on edges are duplicated. Refer to the examples above for more details. `pads_begin[D]` and `pads_end[D]` must be not greater than `data.shape[D]` for any valid `D`.
+  * **Type**: `string`
+  * **Required**: *yes*
+
+**Inputs**
+
+* **1**: `data` tensor of arbitrary shape and type *T*. **Required.**
+
+* **2**: `pads_begin` 1D tensor of type *T_INT*. Number of elements matches the shape rank of *data* input. Specifies the number of padding elements to add at the beginning of each axis. Negative value means shrinking of the corresponding dimension's value. **Required.**
```
rkazants, line 147: may be instead of `shrinking` it is better to use `cropping`?

docs/ops/movement/Pad_12.md
```diff
@@ -0,0 +1,261 @@
...
+
+**Outputs**
+
+* **1**: Padded output tensor of type *T* with dimensions `max(pads_begin[D] + data.shape[D] + pads_end[D], 0)` for each `D` from `0` to `len(data.shape) - 1`.
+
+**Types**
+
+* *T*: any numeric type.
+
+* *T_INT*: any integer type.
```
rkazants, line 162: since `pads` can be negative, will it be safe to use `unsigned` types? We need to be very careful with it

rkazants (approved):

docs/ops/movement/Pad_12.md
mitruska: Sounds good, updated

docs/ops/movement/Pad_12.md
mitruska: It seems to me that in favor of compatibility with previous Pad version, `unsigned` types can be allowed (but used at own risk).

mitruska:
Merged current version with negative pads description. 
Proposal of new mode will be reviewed separately.

PR #17514, opened 2023-05-12
[CPU][ARM] Dynamic shapes support in ARM transformations 

alvoron:
Ticket: [CVS-101331](https://jira.devtools.intel.com/browse/CVS-101331)


alvoron:
Duplication of https://github.com/openvinotoolkit/openvino/pull/17517

PR #17519, opened 2023-05-12
[PyOV] Add objects representations

akuporos:
### Details:
 - *item1*
 - *...*

### Tickets:
 - 110622


src/bindings/python/src/pyopenvino/graph/passes/transformations.cpp
```diff
@@ -141,4 +157,7 @@ void regclass_transformations(py::module m) {
         convert(m, "ConvertFP32ToFP16");
     convert.doc() = "openvino.runtime.passes.ConvertFP32ToFP16 transformation";
     convert.def(py::init<>());
+    convert.def("__repr__", [](const ov::pass::ConvertFP32ToFP16& self) {
+        return "<ConvertFP32ToFP16>";
+    });
```
jiwaszki, line 162:
```cpp
       auto pyclass_name = py::detail::c_str(std::string("ConvertFP32ToFP16"));

       py::class_<ov::pass::ConvertFP32ToFP16,
               std::shared_ptr<ov::pass::ConvertFP32ToFP16>,
               ov::pass::ModelPass,
               ov::pass::PassBase> convert(m, pyclass_name); // reuse same name
       convert.doc() = "openvino.runtime.passes." + pyclass_name + " transformation";
       convert.def(py::init<>());
       convert.def("__repr__", [](const ov::pass::ConvertFP32ToFP16& self) {
           return "<" + pyclass_name  + ">";   // reuse same name
       });
```
If the name is extracted to some variable, it has less chances of being misspelled in other places (or at least it is misspelled everywhere in the same way).
If some class has very different name concept and/or standards of `__repr__`, it still has potential benefits. May this name also be also concatenated into doc strings? etc.
What do you think about something like this? 

Another idea where `pyclass_name` helps:
```cpp
// probably common.cpp/hpp
std::string get_repr_string(std::string name) {
    return "<" + name + ">";
}
```

akuporos: added common function
but I found other way of extracting class name

jiwaszki (commented):
Are test cases planned? Ideally it would be nice to have validation of object names and their stability (like having correct "<>" closing).

src/bindings/python/src/pyopenvino/core/common.hpp
```diff
@@ -96,6 +96,12 @@ py::dict outputs_to_dict(InferRequestWrapper& request);
 
 ov::pass::Serialize::Version convert_to_version(const std::string& version);
 
+template <typename T>
+std::string get_simple_repr(const T& obj) {
+    std::string class_name = py::str(py::cast(obj).get_type().attr("__name__"));
```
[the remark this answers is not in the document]
akuporos: added `get_class_name` function and refactoring the rest of repr functions

PR #17522, opened 2023-05-14
[CPU] Use switch case to ensure to_string methods are complete

EgorDuplensky:
Usage of do-while wrapper is unnecessary here since the macro is local

### Details:
 - Add missing cases
 - Compiler can warn about missed enum entry in switch case


EgorDuplensky:
@dmitry-gorokhov Could you please take a look

dmitry-gorokhov (approved):

PR #17548, opened 2023-05-15
[CPU][ARM] Dynamic shapes support in ARM transformations

alvoron:
Ticket: [CVS-101331](https://jira.devtools.intel.com/browse/CVS-101331)
Propagation of https://github.com/openvinotoolkit/openvino/pull/17517


dmitry-gorokhov (approved):

EgorDuplensky (changes requested):
I think we should add tests for the new arm transformations at least for master branch

alvoron:
> I think we should add tests for the new arm transformations at least for master branch

Tests are added, could you please review?

dmitry-gorokhov:
I am just wondering why do we need separate subgraph tests there? It seems like we can add ARM instances for Convoltuion, GroupConvolution and Reduce [CPU single layer tests](https://github.com/openvinotoolkit/openvino/tree/master/src/plugins/intel_cpu/tests/functional/single_layer_tests) to cover dynamic shapes for these operations and additionally correntess of decomposition in the whole pipeline.
At the same time we usually implement [simple unit test](https://github.com/openvinotoolkit/openvino/tree/master/src/plugins/intel_cpu/tests/unit/ngraph_transformations)to cover transformation itself.

src/plugins/intel_cpu/tests/unit/ngraph_transformations/arm/convert_group_conv.cpp
```diff
@@ -0,0 +1,94 @@
...
+            auto conv = std::make_shared<ngraph::opset1::Convolution>(out,
+                                                                      filter,
+                                                                      ngraph::Strides{1},
+                                                                      ngraph::CoordinateDiff{0},
+                                                                      ngraph::CoordinateDiff{0},
+                                                                      ngraph::Strides{1});
+            concat_inputs.push_back(conv);
+        }
+        auto concat = std::make_shared<ov::opset8::Concat>(concat_inputs, 1);
+        function = std::make_shared<ngraph::Function>(ngraph::NodeVector{ concat }, ngraph::ParameterVector{ param });
```
EgorDuplensky, line 68: I guess it should be "function_ref", otherwise we do not actually test anything here (and some other test instances as well).
TBH, I don't really like the idea of performing the test check inside TearDown, since from the test instance perspective it is not clear, what is being tested here.
I think we should explicitly compare the functions, i.e. how it is done in convert_to_interaction.cpp tests?

src/plugins/intel_cpu/tests/unit/ngraph_transformations/arm/convert_reduce_multi_axis.cpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+    {
+        auto param = std::make_shared<ngraph::opset1::Parameter>(ngraph::element::f32, dynamic_param_shape);
+        function_ref = createInitGraph<ngraph::opset1::ReduceSum>(param);
+    }
+    auto res = compare_functions(function, function_ref);
+    ASSERT_TRUE(res.first) << res.second;
+}
+
+TEST(TransformationTests, CheckConvertReduceProdTransformationIsNotAppliedForDynaimcShapes) {
```
EgorDuplensky, line 167: It seems all this duplication can be avoided using
https://github.com/google/googletest/blob/main/docs/advanced.md#typed-tests
Which makes sense, since the transformation itself and the functions you are using to create the models in scope of this test are also type templates.

EgorDuplensky (approved):

src/plugins/intel_cpu/tests/unit/ngraph_transformations/arm/convert_reduce_multi_axis.cpp
alvoron: Applied typed tests approach

src/plugins/intel_cpu/tests/unit/ngraph_transformations/arm/convert_group_conv.cpp
alvoron: fixed

alvoron:
@dmitry-gorokhov could we merge it?

dmitry-gorokhov (approved):

PR #17565, opened 2023-05-17
[GPU] Minor fix for shape inference of dynamic reshape

andrew-k-park:
### Details:
 - Reshape does not support 0D tensor(scalar) for `shape` input
 - If the preceding reshape node which converts from 0D tensor to 1D tensor is removed during `handle_reshape` opt pass, the next reshape node should handle 0D tensor with 1D tensor manually

### Tickets:
 - 110534

### DIFF
--- a/src/plugins/intel_gpu/src/graph/reshape.cpp
+++ b/src/plugins/intel_gpu/src/graph/reshape.cpp
@@ -74,6 +74,11 @@ std::vector<layout> reshape_inst::calc_output_layouts(reshape_node const& /*node
 
     ShapeType pattern_shape = impl_param.input_layouts.size() == 2 ? impl_param.get_input_layout(1).get<ShapeType>()
                                                                    : ShapeType(ov::Shape{ prim->output_pattern.size() });
+    // Since reshape does not support 0D tensor(scalar) for shape input
+    // the case propagated to 0D tensor should be handled manually with 1D tensor
+    if (pattern_shape.size() == 0) {
+        pattern_shape = ShapeType{1};
+    }
     std::vector<ShapeType> output_shapes = {ShapeType()};
     std::vector<ShapeType> input_shapes = {
         input_layout.get<ShapeType>(),
--- a/src/plugins/intel_gpu/tests/unit/shape_infer/reshape_si_test.cpp
+++ b/src/plugins/intel_gpu/tests/unit/shape_infer/reshape_si_test.cpp
@@ -82,6 +82,11 @@ INSTANTIATE_TEST_SUITE_P(smoke, reshape_test_two_inputs,
             layout{ov::PartialShape{1, 384, 16, 64}, data_types::f32, format::bfyx}
         },
         {
+            layout{ov::PartialShape{1, 128, 1024}, data_types::f32, format::bfyx},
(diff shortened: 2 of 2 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/handle_reshape.cpp
```diff
@@ -59,7 +59,8 @@ void handle_reshape::run(program& p) {
     while (node_itr != p.get_processing_order().end()) {
         auto& node = (*node_itr++);
         program_helpers::do_for_types<reshape>(*node, [&p](reshape_node& node) {
-            if (node.is_output() || node.get_users().size() > 1 || node.has_fused_primitives() || node.is_dynamic())
+            if (node.is_output() || node.get_users().size() > 1 || node.has_fused_primitives() ||
+                node.is_dynamic() || node.is_unsqueeze_0d_to_1d())
```
vladimir-paramuzov, line 63: Didn't get what's the issue with shape inference of second reshape if we have scalar input. Could you explain a bit more?

andrew-k-park: - In the case of `reshape`, it has shape infer dependency for the 2nd input, and it creates and passes HostTensor from memory dependency for output pattern from it during shape inference. At this time, If the layout of the 2nd input is propagated to 0D, it is an empty tensor without dimension information. so it's empty buffer for output pattern causing runtime error regarding validation check during reshape shape inference
- By the way, the first reshape(Unsqueeze from 0D tensor to 1D tensor) is removed during consecutive reshape nodes optimization on `handle_reshape` opt pass. So when I try to skip to remove this reshape case(Unsqueeze from 0D tensor to 1D tensor), 1D tensor is propagated to the following reshape and shape inference of second reshape works normally

vladimir-paramuzov: 1. 0d tensor is not empty, it's scalar (single element), so I suspect that we need to fix handling of scalars in shape infer impl if it causes an issue
2. Looks like it's applicable only for dynamic shapes and new shape infer
3. `is_unsqueeze_0d_to_1d` method in reshape node api doesn't look good. So if we really need this code, it should be a part of handle_reshape pass

yeonbok: @andrew-k-park 
>>  At this time, If the layout of the 2nd input is propagated to 0D, it is an empty tensor without dimension information. so it's empty buffer for output pattern causing runtime error regarding validation check during reshape shape inference

Can we fix this issue by either of 
1) passing layout of PartialShape{1} to make_host_tensor 
2) If the output pattern in reshape is scalar, not to call ngraph shape infer but just set manually from calc_outptu_layout ? 

I think 2) would be better for the performance

andrew-k-park: I checked that the first option is enough. Because creating HostTensor with 0D tensor is not issue and shape inference could work normally just by chainging input layout with 1D tensor. and if we calculate output layout manually without using ngraph shape infer, implementation becomes more complicated because we have to consider the case where the value of shape input is 0 or -1.

yeonbok (approved):

PR #17569, opened 2023-05-17
[AUTO] Support setting log_level from config

WeldonWangwang:
### Details:
 - *Support setting log_level from config*


yangwang201911 (approved):

src/plugins/auto/src/plugin.cpp
```diff
@@ -360,6 +360,9 @@ IExecutableNetworkInternal::Ptr MultiDeviceInferencePlugin::LoadNetworkImpl(cons
     // updateFromMap will check config valid
     loadConfig.set_user_property(PreProcessConfig(config));
     loadConfig.apply_user_properties();
+    if (!setLogLevel(loadConfig.get_property(ov::log::level.name()).as<std::string>())) {
+        IE_THROW() << "Unsupported log level";
+    }
```
songbell, line 365: then log from other calls except than loadmodel will still not getting outputs.

WeldonWangwang: Added judgment on whether log_level comes from user settings.

src/plugins/auto/src/plugin_config.cpp
```diff
@@ -51,6 +51,12 @@ void PluginConfig::set_property(const ov::AnyMap& properties) {
             internal_properties[name] = val;
             // when user call set_property to set some config to plugin, we also respect this and pass through the config in this case
             user_properties[name] = val;
+            if (kv.first == ov::log::level.name()) {
```
peterchen-intel, line 54: Any test cases cover setting log level successfully and failed?

wangleis (approved):

src/plugins/auto/src/plugin_config.cpp
WeldonWangwang: Added test case.

songbell (approved):

wgzintel (approved):

PR #17572, opened 2023-05-17
[MO] disabling inserting redundant Converts for int in legacy IR serialization

pavel-esir:
**Problem statement:**
ir reader test were failing for TF_Cyberlink_MFaceNet. It's because during serialization in legacy MO there were additional Converts from int32->int64 inserted.
image

In serializer there was already a code which skipped inserting additional converts for integers, but it is activated only for constants. Made it to be activated for all data types and the test got green.

**Tickets:**
 - xxx-103814


tools/mo/openvino/tools/mo/pipeline/common.py
```diff
@@ -169,15 +169,17 @@ def convert_inputs_of_specific_ops(graph: Graph):
...
                         np_type = data_type_str_to_np(precision)
+                        in_node = node.in_port(port_id).get_source().node
+                        in_node_type = in_node.out_port(0).get_data_type()
+                        if np.issubdtype(in_node_type, np.integer) and np.issubdtype(np_type, np.integer):
+                            # do not convert if both source and destination types are of integer types
+                            # otherwise, it affects compatibility of MO IR Engine and TF FE
+                            # TF FE intents to use original model type for layers if it is possible.
+                            # Also, this skip is needed to get the same graphs from original IR and
+                            # restored with ir_reader_utils
+                            continue
```
rkazants, line 180: Actually, it can be not safe change. I remember Evgenya stopped to do it and proposed to apply only for `Const`.
@jane-intel, can you please advice if this is safe change? What pitfalls can be here?

pavel-esir: Discussed with @jane-intel. Skipping for non-const nodes should be done selectively only for conversions within int32, int64, uint32, uint64. Corrected current PR to do so.

But if we leave `int8`, `int4` (or similar) types on `Reshape, Squeeze, etc.` inputs potentially there can be problems, therefore for such cases current serializer behavior is left unchanged.

jane-intel (approved):

rkazants (changes requested):
Unit tests are failing. Please fix them

pavel-esir:
> Unit tests are failing. Please fix them

sure, will do that

rkazants (approved):

PR #17574, opened 2023-05-17
improve log check in PDFE

xczhai:
### Details:
 - *improve the log accuracy to detect the model file, weight file*
 - *recover PD old api, such as __model__ and separate weight*
 - *split the old and new api for loadConst. It is better for maintenance.*

### Tickets:
 - *72006*


src/frontends/paddle/src/input_model.cpp
```diff
@@ -259,21 +274,68 @@ std::vector<std::shared_ptr<OpPlace>> InputModel::InputModelImpl::determine_cut_
...
         const auto& name = item.first;
         if (ov::util::ends_with(name, std::string{"feed"}) || ov::util::ends_with(name, std::string{"fetch"}))
             continue;
+
+        // var_desc.persistable() is used to mark node const value or not.
         if (!var_desc.persistable())
             continue;
 
         FRONT_END_GENERAL_CHECK(var_desc.type().type() == ::paddle::framework::proto::VarType::LOD_TENSOR);
+        FRONT_END_GENERAL_CHECK(weight_stream->peek() != EOF,
```
liubo-intel, line 334: Hi, @xczhai : just one question here, can't  this 'weight_stream' be nullptr? from 'get_model_path' [description](https://github.com/openvinotoolkit/openvino/blob/639235d09944f0a86503b0dda8b17e65a28214da/src/frontends/paddle/src/input_model.cpp#L202) looks like it can be nullptr(although this looks Strange and uncommon in real model case)

xczhai: Hi @liubo-intel,
- If loading model from disk, `weight_stream` can't be `nullptr` because its origin is from `std::ifstream` type in the method `InputModelImpl`.
- But your concern makes sense when loading from memory. So, I will add a `weight_stream == nullptr` check condition for this scenario.

meiyang-intel (approved):

src/frontends/paddle/src/input_model.cpp
xczhai: @liubo-intel Fixed.

liubo-intel (approved):

PR #17578, opened 2023-05-17
[ShapeInfer] Review Proposal op for shape inference aspects

praasz:
### Details:
- Check partial shape and label propagation.
- Check template implementation of shape_infer.
- Add/update unit test for static and dynamic shapes.
- Drop using `ngraph` namespace from reviewed operator.

### Tickets:
 - 97243
 - 97295
 - 97344
 - 97383


src/core/shape_inference/include/proposal_shape_inference.hpp
```diff
@@ -30,78 +32,70 @@ void infer_prop_shape(const OpType* op,
...
+                                  TDim::merge(proposed_boxes_shape[0], proposed_boxes_shape[0], bbox_deltas_ps[0]),
+                                  "Batch size inconsistent between class_probs (",
+                                  class_probs_ps[0],
+                                  ") and bbox deltas (",
+                                  bbox_deltas_ps[0],
+                                  ").");
+        }
     } else {
-        out_dim = Dimension::dynamic();
+        proposed_boxes_shape.emplace_back(ov::util::dim::inf_bound);
```
mitruska, line 69: Looks like `batch_size` dimension can be lost if `class_probs_ps.rank()` is dynamic, 
but `bbox_deltas_ps.rank()` static.

src/core/shape_inference/include/proposal_shape_inference.hpp
```diff
@@ -30,78 +32,70 @@ void infer_prop_shape(const OpType* op,
...
+    proposed_boxes_shape.emplace_back(5);
+    return proposed_boxes_shape;
 }
+}  // namespace proposal
 
+namespace v0 {
+template <class TShape>
+std::vector<TShape> shape_infer(const Proposal* op,
+                                const std::vector<TShape>& input_shapes,
+                                const ITensorAccessor& tensor_accessor = make_tensor_accessor()) {
```
mitruska, line 82: Is there a need to create `tensor_accessor` even if it's not used?

src/core/shape_inference/include/proposal_shape_inference.hpp
praasz: Good catch will fix.

src/core/shape_inference/include/proposal_shape_inference.hpp
praasz: It was not possible with adding some classes to shape inference.
I agree if parameter not used should not be defined at all and I updated it.

Just for information the `make_tensor_accessor` is cheap and I've done improvement  to tensor accessor that it will even simpler is just return reference to existing single object (created once per program life), so using this empty accessor as default parameter is similar to `int=0`;

mitruska (approved):

jane-intel (approved):

PR #17581, opened 2023-05-17
[ARM CPU] Fix MVN6 decompose drop accuracy

allnes:
CSV-111880


src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/arm/eltwise.cpp
```diff
@@ -0,0 +1,59 @@
...
+const std::vector<float> epsilonF = {
+        0.0001f
+};
+
+const std::vector<std::string> epsMode = {
+        "inside_sqrt",
+        "outside_sqrt"
+};
+
+INSTANTIATE_TEST_SUITE_P(smoke_power_mvn_decomposition_3D, Mvn6LayerTest,
```
EgorDuplensky, line 32: Mvn6LayerTest is seemed to be the class from the shared_tests.
Here we should work only with the plugin's single_layer_tests
Also, I do not think we need to deal with MVN here at all.
I understand that this eltwise-power node is the result of MVN decomposition.
But for accuracy verification we can just create an instance for the eltwise-power test (even the file itself is called eltwise.cpp)

EgorDuplensky (changes requested):

src/plugins/intel_cpu/tests/unit/ngraph_transformations/arm/mnv6_power_decomposition.cpp
```diff
@@ -0,0 +1,57 @@
...
+    {
+        ov::pass::Manager manager;
+        manager.register_pass<MVN6PowerDecomposition>();
+        manager.run_passes(m);
+    }
+
+    auto list_of_ops = m->get_ordered_ops();
+    ASSERT_STRCASEEQ(list_of_ops[0]->get_type_info().name, "Parameter");
+    ASSERT_STRCASEEQ(list_of_ops[1]->get_type_info().name, "Multiply");
+    ASSERT_STRCASEEQ(list_of_ops[2]->get_type_info().name, "Result");
```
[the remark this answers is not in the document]
allnes: @EgorDuplensky done

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/arm/eltwise.cpp
allnes: @EgorDuplensky done

EgorDuplensky (approved):

dmitry-gorokhov (changes requested):
ACL doesn't support negative base for power operation. Given that current solution will still work incorrectly for cases with power != 2.
The proposal is not to enable ACL for power operation, but use optimized C++ instead.

PR #17582, opened 2023-05-17
[GNA] Transpose fusing 

mryzhov:
### Details:
 - Detects 3d/4d unsupported transposes and fuses it's dimentions to reduce the shape rank and make such transposes eligible  

### Tickets:
 - *N/A*


src/plugins/intel_gna/src/transformations/transpose_2d.hpp
```diff
@@ -0,0 +1,39 @@
...
+#pragma once
+
+#include "openvino/pass/graph_rewrite.hpp"
+
+namespace ov {
+namespace intel_gna {
+namespace pass {
+
+/**
+ * @brief Reduce the rank of Transpose shape by fusing by fusing dimentions
```
sirzabek, line 14:
```suggestion
 * @brief Reduce the rank of Transpose shape by fusing dimensions
```

src/plugins/intel_gna/src/transformations/transpose_2d.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    ov::matcher_pass_callback callback = [=](pattern::Matcher& m) {
+        const auto& pattern_map = m.get_pattern_value_map();
+        const auto transpose_order = as_type_ptr<Constant>(pattern_map.at(transpose_const).get_node_shared_ptr());
+        const auto transpose_node = pattern_map.at(transpose).get_node_shared_ptr();
+        const ov::Shape& shape = transpose_node->get_input_shape(0);
+        const ov::Shape& shape_out = transpose_node->get_output_shape(0);
+        ov::AxisVector axises = transpose_order->get_axis_vector_val();
+        ov::AxisVector axises_fused = {};
+        ov::Shape shape_fused_out = {};
```
sirzabek, line 33: I think we should not use word "fuse" in the context of this transformation, it may be confusing (pun intended ;-) ). We use fusing for whole layers. I suggest squeeze, although maybe it's not perfect.

sirzabek (changes requested):
General remarks/questions:
- there is no ticket connected
- are we sure we want to treat all 3d/4d transposes in this way at this particular moment in transformation pipeline? Maybe I can't see the whole picture of the layout transformation...

src/plugins/intel_gna/src/transformations/transpose_2d.hpp
```diff
@@ -0,0 +1,39 @@
...
+ *         |           ->          |
+ *         |           <-     [A, D, B*C]
+ *         |                       |
+ *         |                    Reshape
+ *         |                       |
+ *         |                  [A, D, B, C]
+ *         |                       |
+ *     Any Layer                Any Layer
+ */
+class Transpose2D : public ov::pass::MatcherPass {
```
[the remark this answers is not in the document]
mryzhov: renamed to TransposeCompress )

PR #17584, opened 2023-05-17
[PT FE] Fix aten::squeeze translation + add tests for aten::slice

mmikolajcz:
### Details:
 - *Improve squeeze translation to fix mismatch between OV and PyTorch*
 - *Fix type mismatch in remainder*
 - *Add missing tests for squeeze and slice*

### Explanation of mismatch in squeeze:
If one of provided dimensions couldn't be squeezed, OV raises exception, while PyTorch returns dimension unchanged. For example:
Tensor `input` of shape (1, 1, 32),
- PyTorch `squeeze(input, [-2, -1])` returns tensor of shape (1, 32),
- OV `squeeze(input, [-2, -1])` raises exception because last dim cannot be squeezed.

### Tickets:
 - *101087*


tomdol (approved):

src/frontends/pytorch/src/op/remainder.cpp
```diff
@@ -20,6 +20,7 @@ OutputVector translate_remainder(const NodeContext& context) {
     num_inputs_check(context, 2, 2);
     auto x = context.get_input(0);
     auto y = context.get_input(1);
+    align_eltwise_input_types(context, x, y, true);
```
mvafin, line 23: This is a good change, but please add tests for it

mlukasze:
@mvafin can we merge it, or you need more time for next iteration of review? :)

src/frontends/pytorch/src/op/squeeze.cpp
```diff
@@ -5,20 +5,67 @@
...
     }
-    return {context.mark_node(std::make_shared<ov::op::v0::Squeeze>(x, context.get_input(1)))};
+    // Cannot provide dimensions to ov v0::Squeeze directly due to mismatch in behavior between OV and PyTorch:
+    // If provided dimension cannot be squeezed, OV raises exception, PyTorch returns dimension unmodified.
+    auto dim_input = context.get_input(1);
+    auto const_0 = context.mark_node(v0::Constant::create(element::i64, Shape{}, {0}));
+    auto const_1 = context.mark_node(v0::Constant::create(element::i64, Shape{}, {1}));
+    auto input_shape = context.mark_node(std::make_shared<v3::ShapeOf>(input));
+    auto input_rank = context.mark_node(std::make_shared<v3::ShapeOf>(input_shape));
+    auto input_rank_s = context.mark_node(std::make_shared<v0::Squeeze>(input_rank));
```
mvafin, line 40: Please use `get_shape_rank` and i32 type where possible

src/frontends/pytorch/src/op/squeeze.cpp
```diff
@@ -5,20 +5,67 @@
...
+    // Create boolean mask containing True on 1, else 0.
+    auto squeezable_dimensions_mask = context.mark_node(std::make_shared<v1::Equal>(input_shape, const_1));
+    auto input_shape_indices =
+        context.mark_node(std::make_shared<v4::Range>(const_0, input_rank_s, const_1, element::i64));
+    // Translate negative dimension indices into positive only.
+    auto dim_input_pos_only = context.mark_node(std::make_shared<v8::Gather>(input_shape_indices, dim_input, const_0));
```
mvafin, line 46: This is a good way to normalize negative axes, but we have a function in `utils.hpp` now that can do the same called `normalize_axis`, it uses sligtly different approach, I am not sure which way is better, but please use that helper function, you can update it to use your approach if you think it is more efficient.

tests/layer_tests/pytorch_tests/test_squeeze.py
```diff
@@ -28,8 +29,8 @@ def forward(self, x):
...
         return aten_squeeze(dim), ref_net, "aten::squeeze"
 
-    @pytest.mark.parametrize("dim,dynamic_shapes", [(-2, True), (0, True), (None, False)])
+    @pytest.mark.parametrize("dim", [-2, -1, 0, None, [-1, -2], 1, 2])
     @pytest.mark.nightly
     @pytest.mark.precommit
-    def test_squeeze(self, dim, ie_device, precision, ir_version, dynamic_shapes):
-        self._test(*self.create_model(dim), ie_device, precision, ir_version, dynamic_shapes=dynamic_shapes)
+    def test_squeeze(self, dim, ie_device, precision, ir_version):
+        self._test(*self.create_model(dim), ie_device, precision, ir_version, dynamic_shapes=False)
```
mvafin, line 36: Why you switch off dynamic shapes? I think your approach shouldn't depend on knowing shapes, am I wrong?

mvafin (changes requested):

src/frontends/pytorch/src/op/squeeze.cpp
mmikolajcz: Thanks, done

src/frontends/pytorch/src/op/squeeze.cpp
mmikolajcz: Thanks, done

tests/layer_tests/pytorch_tests/test_squeeze.py
mvafin: I was thinking that dynamic shapes flag just changes the shapes, not rank. Please leave a comment that Squeeze will fail with dynamic shapes

mmikolajcz: Thanks, added comment to test. Should translation also contain this comment?

mvafin (approved):

PR #17586, opened 2023-05-17
Temporary remove cron trigger from stale action

akazakov-github:
to prevent unwanted closing of already stalled issues

### DIFF
--- a/.github/workflows/stale_prs_and_issues.yml
+++ b/.github/workflows/stale_prs_and_issues.yml
@@ -1,8 +1,6 @@
 name: 'Close stale issues and PRs'
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


akladiev (approved):

.github/workflows/stale_prs_and_issues.yml
```diff
@@ -1,8 +1,6 @@
 name: 'Close stale issues and PRs'
 on:
   workflow_dispatch:
```
akladiev, line 3: Probably need to remove the on: workflow_dispatch: as well

akladiev (changes requested):

akladiev (approved):

PR #17591, opened 2023-05-18
Reverse instead of sort in NMS ref

p-durandin:
### Details:
 - Implemented comparator in reference implementation for some cases can be true in A < B and B <A which leads to uncertain behaviour
 - NMS creates sorted result, so just reverse is enough

### Tickets:
 - 93735


tomdol:
Is this change already covered by any existing test of this op? If not, please add at least one test demonstrating the use case that this change fixes. I believe we should cover that with a test for reference implementation only and additionally by a layer test comparing plugins against ref.

p-durandin:
Yes, smoke_Nms9LayerTest/Nms9LayerTest.CompareWithRefs/numBatches=2_numBoxes=50_numClasses=50_paramsPrec=FP32_maxBoxPrec=I32_thrPrec=FP32_maxOutBoxesPerClass=5_iouThr=0.3_scoreThr=0.3_softNmsSigma=0.5_boxEncoding=corner_sortResDescend=1_outType=i32_TargetDevice=GPU
Fails in debug with "invalid comparator" assert

jane-intel (commented):
The change makes sense. 

I see that test has smoke_ prefix -- meaning it should go to the pre-commit. Means it is disabled somehow -- let's enable it then

jane-intel (approved):

PR #17613, opened 2023-05-19
[GNA] Add preliminary DWSC support

sirzabek:
### Details:
 - add basic support for DWSC HW operation
 - add IR group convolutions translation into DWSC by graph compiler.

### Tickets:
 - 100001


src/plugins/intel_gna/src/backend/dnn_types.hpp
```diff
@@ -96,6 +96,7 @@ typedef enum {
     kDnnDiagonalOp,
     kDnnConvolutional1dOp,
     kDnnConvolutional2dOp,
+    kDnnDWSCOp,
```
marcinkusm, line 99: do we need to add this to FP32 runtime ?

src/plugins/intel_gna/src/backend/gna_limitations.cpp
```diff
@@ -248,6 +248,19 @@ class Validator_30 : public AbstractValidator {
                        const uint32_t dilationW,
                        OvGnaType inPrecision,
                        bool exception = true) const override;
+
+    bool ValidateDwsc(const std::string& name,
+                      const uint32_t inHeight,
```
marcinkusm, line 253: `in_height`
`in_width`
and so on

src/plugins/intel_gna/src/backend/gna_limitations.cpp
```diff
@@ -617,16 +673,83 @@ bool Validator_35::ShouldUseOnlyConv2DGnaIface() const {
...
+                                const uint32_t kernelN,
+                                const uint32_t strideH,
+                                const uint32_t strideW,
+                                const uint32_t dilationH,
+                                const uint32_t dilationW,
+                                const bool throwOnError) const {
+    auto error = kDwscLimits.kInputHWLimit.GetErrorOrEmpty(inHeight, inWidth);
+    const IsEqualToLimit kernels_count{inChannels, "number of kernels (must be equal to input channels count)"};
+    error += kernels_count.GetErrorOrEmpty(kernelN);
+    auto& inputChannelsNumberLimit = kDwscLimits.kInputChannelsNumberLimit;
```
marcinkusm, line 732:
```suggestion
    const auto& inputChannelsNumberLimit = kDwscLimits.kInputChannelsNumberLimit;
```

src/plugins/intel_gna/src/backend/gna_limitations.cpp
```diff
@@ -617,16 +673,83 @@ bool Validator_35::ShouldUseOnlyConv2DGnaIface() const {
...
+}
+
 std::shared_ptr<AbstractValidator> AbstractValidator::Create(const DeviceVersion& target) {
     switch (target) {
     case DeviceVersion::GNA3_0:
     case DeviceVersion::GNA3_1:
         return std::make_shared<Validator_30>();
     case DeviceVersion::GNA3_5:
     case DeviceVersion::GNAEmbedded3_5:
+        return tools::make_unique<Validator_35>();
```
marcinkusm, line 749:
```suggestion
        return std::make_shared<Validator_35>();
```

src/plugins/intel_gna/src/backend/gna_limitations.cpp
```diff
@@ -617,16 +673,83 @@ bool Validator_35::ShouldUseOnlyConv2DGnaIface() const {
...
     case DeviceVersion::GNA3_6:
     case DeviceVersion::GNA4_0:
-        return std::make_shared<Validator_35>();
+        return tools::make_unique<Validator_36>();
```
marcinkusm, line 752:
```suggestion
        return std::make_shared<Validator_36>();
```

src/plugins/intel_gna/src/gna_device.cpp
```diff
@@ -389,6 +389,7 @@ const std::map<Gna2ErrorType, const std::string> GNADeviceHelper::errorReasons =
 const std::map<Gna2OperationType, const std::string> GNADeviceHelper::operationTypes = {
     {Gna2OperationTypeNone, "None"},
     {Gna2OperationTypeConvolution, "Convolution"},
+    {Gna2OperationTypeConvolutionDWSC, "Dwsc"},
```
marcinkusm, line 392: Is it proper name?
I saw in few places:
"Dwsc"
"DWSC"
"dwsc"

why not 
`ConvolutionDWSC`

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -736,6 +764,8 @@ void GNAGraphCompiler::finalizeConvolution2DPrimitive(InferenceEngine::CNNLayerP
     const auto weightPrec = OvGnaTypeIntFromBytes(convolution._weights->getTensorDesc().getPrecision().size());
     const auto biasPrec = OvGnaTypeIntFromBytes(biasPrecision.size());
 
+    const auto is_dwsc = (convolution._group > 1);
```
marcinkusm, line 767: maybe it would be good to create common method for checking if convolution is dwsc? 
this condition is used at least in two places.

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -361,14 +388,15 @@ void GNAGraphCompiler::ConvolutionPrimitive(InferenceEngine::CNNLayerPtr layer)
...
-    // Map 2d convolution to 1d if it's possible.
-    if (!ShouldUseOnlyConv2DGnaIface() && gna_convolution_layer::isMappableFrom2DTo1D(in_height,
-                                                                                      in_width,
-                                                                                      in_channels,
-                                                                                      convolution._kernel_y,
-                                                                                      convolution._kernel_x,
-                                                                                      convolution._stride_y,
-                                                                                      convolution._stride_x)) {
+    // Map 2d convolution to 1d if it's possible
+    if (!is_dwsc && !ShouldUseOnlyConv2DGnaIface() &&
```
marcinkusm, line 392: Can we have situation that ShouldUseOnlyConv2DGnaIface() will return false but is_dwsc will be true?

marcinkusm (changes requested):
Please add const for reference to consts  and change back unique_ptr to shared_ptr .

marcinkusm (changes requested):
Please add const for reference to consts  and change back unique_ptr to shared_ptr .

src/plugins/intel_gna/src/backend/gna_limitations.cpp
sirzabek: It's in line with the whole file, I don't see a point in changing naming convention for a single piece of code in a whole file.

src/plugins/intel_gna/src/gna_device.cpp
sirzabek: DWSC = depth-wise separable convolution, so I think it's enough

src/plugins/intel_gna/src/gna_graph_compiler.cpp
sirzabek: is_dwsc check can be removed without changing the logic, because it's broke anyway for group convolutions.

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -805,14 +843,17 @@ void GNAGraphCompiler::finalizeConvolution2DPrimitive(InferenceEngine::CNNLayerP
     const auto single_kernel_size = in_channels * kernelHW * convolution_precision;
 
     const auto effective_kernel_h_w = convolution._kernel_y * effective_kernel_width;
-    const auto effective_single_kernel_size = in_channels * effective_kernel_h_w * convolution_precision;
+    const auto effective_single_kernel_size =
+        in_channels / convolution._group * effective_kernel_h_w * convolution_precision;
```
marcinkacprzak, line 847: convolution._group is not checked against 0 value.

sirzabek: It is, during graph validation.

src/plugins/intel_gna/src/transformations/convert_dwsc_to_scaleshifts.cpp
```diff
@@ -15,8 +15,10 @@
 
 #include "utils/transformation_helper.hpp"
 
-using namespace ov::intel_gna::pass;
-using namespace ov::intel_gna::pass::helper;
+namespace ov {
+namespace intel_gna {
+namespace pass {
+using namespace helper;
```
marcinkusm, line 22: Shouldn't we disable ConvertDWSCToScaleShifts transformation for for targets >= 3.6

sirzabek: No.

src/plugins/intel_gna/src/gna_device.cpp
marcinkusm: I just wonder that we want the same convetion as for other operation types
which uses name after "Gna2OperationType"

src/plugins/intel_gna/tests/unit/gna_dwsc_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    void SetUp() override {
+        const std::vector<size_t> c_strides{1, 1};
+        const std::vector<size_t> c_dilations{1, 1};
+        constexpr size_t c_num_out_channels = 8;
+        const auto& model = GetParam().model;
+
+        using ngraph::element::f32;
+        auto parameter = std::make_shared<ngraph::opset9::Parameter>(f32, ngraph::Shape{model.input_size});
```
marcinkacprzak, line 65: Newest opset should be used (11).

src/plugins/intel_gna/src/backend/gna_limitations.cpp
mkwapul: not agreed. deduction from assignment above will bring const type already, so extra const is not needed here. please re-set to resolved after reading.

src/plugins/intel_gna/src/common/gna_target.cpp
```diff
@@ -44,8 +44,8 @@ static const std::unordered_map<Gna2DeviceVersion, DeviceVersion> GnaDeviceMap{
     {Gna2DeviceVersionEmbedded3_1, DeviceVersion::GNA3_1},
     {Gna2DeviceVersion3_5, DeviceVersion::GNA3_5},
     {Gna2DeviceVersionEmbedded3_5, DeviceVersion::GNAEmbedded3_5},
-    {Gna2DeviceVersionEmbedded3_5, DeviceVersion::GNA3_6},
-    {Gna2DeviceVersionEmbedded3_5, DeviceVersion::GNA4_0},
+    {Gna2DeviceVersionEmbedded3_6, DeviceVersion::GNA3_6},
+    {Gna2DeviceVersionEmbedded3_6, DeviceVersion::GNA4_0},
```
mkwapul, line 48: please decide on either

src/plugins/intel_gna/src/gna_device.cpp
sirzabek: Notice this isn't a convention, it's not kept in every case.

src/plugins/intel_gna/src/common/gna_target.cpp
sirzabek: Could you elaborate? Decide on either what?

mkwapul: due to the same key, the second entry is dead right?

sirzabek: It's not dead, it means mapping 4.0 <-> 3.6. Please see how this map is used.

mkwapul: when you iterate over GnaDeviceMap do you see Gna2DeviceVersionEmbedded3_6 -> DeviceVersion::GNA4_0 mapping present?
additionally, does the following code throw or not?:

          DeviceToGna(DeviceVersion::GNA4_0);

src/plugins/intel_gna/src/backend/dnn_types.hpp
sirzabek: I think so.

src/plugins/intel_gna/src/common/gna_target.cpp
mkwapul: do you want the second dead entry for documentation purposes? IF yes, then for not confuse the readers, please say it out explicitly - prepend the line with following comment as minimum:

             // for documentation purposes:

other, more meaningful comments would be welcome of course

PR #17615, opened 2023-05-19
[CPU] Reduce node improve parallelism

xuchen-intel:
### Details:
 - *Improve parallelism of cases with small batch size in nspc layout*
 - *Extend bf16 optimization in https://github.com/openvinotoolkit/openvino/pull/16829 to support int8*

### Tickets:
 - *CVS-110961*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/reduce.cpp
+++ b/src/plugins/intel_cpu/src/nodes/reduce.cpp
@@ -2527,14 +2527,29 @@ inline void Reduce::reduce_kernel_process(const uint8_t *in_p, uint8_t *out_p, s
 inline void Reduce::reduce_kernel_post_process(uint8_t *out_ptr) {
     const size_t integerDivisor = IB * IC * ID * IH * IW / (OB * OC * OD * OH * OW);
     const float divisor = static_cast<float>(integerDivisor);
-    if (layout == ReduceLayoutType::reduce_ncsp || layout == ReduceLayoutType::reduce_nspc) {
+    if (layout == ReduceLayoutType::reduce_ncsp) {
         parallel_for2d(OB, OC, [&](size_t ob, size_t oc) {
             uint8_t *out_p = out_ptr + (ob * OC + oc) * OD * OH * OW * dst_data_size;
             auto arg = jit_reduce_post_call_args();
             arg.dst = static_cast<void *>(out_p);
-            arg.oc_off = layout == ReduceLayoutType::reduce_nspc ? 0 : oc * sizeof(float);
-            arg.channel_size = layout == ReduceLayoutType::reduce_nspc ? OW : OC; // OW is related to nspc-ncsp dimension reinterpret
+            arg.oc_off = oc * sizeof(float);
+            arg.channel_size = OC;
             arg.work_amount = OD * OH * OW;
+            arg.divisor = &divisor;
+            arg.post_op_data = static_cast<const void **>(postOpsDataPtrs.data());
+            (*reduce_post_kernel)(&arg);
+        });
+    } else if (layout == ReduceLayoutType::reduce_nspc) {
+        size_t num_threads = static_cast<size_t>(parallel_get_max_threads());
+        size_t OP = OB * OC >= num_threads ? OB * OC : OB * OC * OD;
+        size_t work_amount = OB * OC * OD * OH * OW / OP;
+        parallel_for(OP, [&](size_t op) {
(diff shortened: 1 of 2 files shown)


xuchen-intel:
@dmitry-gorokhov Hi Dmitry, could you please take a look?

dmitry-gorokhov:
@chenhu-wang May I ask you to take a look?

src/plugins/intel_cpu/src/nodes/reduce.cpp
```diff
@@ -2503,18 +2502,33 @@ inline void Reduce::reduce_kernel_process(const uint8_t *in_p, uint8_t *out_p, s
...
+            arg.oc_off = oc * sizeof(float);
+            arg.channel_size = OC;
             arg.work_amount = OD * OH * OW;
             arg.divisor = &divisor;
             arg.post_op_data = static_cast<const void **>(postOpsDataPtrs.data());
             (*reduce_post_kernel)(&arg);
         });
+    } else if (layout == ReduceLayoutType::reduce_nspc) {
+        size_t num_threads = static_cast<size_t>(parallel_get_max_threads());
+        size_t OP = OB * OC >= num_threads ? OB * OC : OB * OC * OD;
```
chenhu-wang, line 2519: extend: As I can imagine OD could be 1 in some cases, such as 4d nchw input to normalized ncdhw dimension. So parallelism in OD does not help. In this case OH could be considered parallel if threads number is large enough and OW is also big(at least bigger than a cache line).

src/plugins/intel_cpu/tests/functional/single_layer_tests/reduce_ops.cpp
```diff
@@ -303,6 +303,11 @@ std::vector<std::vector<ov::test::InputShape>> inputShapes_SmallChannel = {
     {{{{1, 5}, 3, {1, 5}, {1, 10}}, {{2, 3, 2, 2}, {2, 3, 2, 9}}}},
 };
 
+std::vector<std::vector<ov::test::InputShape>> inputShapes_SingleBatch = {
+    {{{}, {{1, 19, 2, 9}}}},
+    {{{{1, 5}, 19, {1, 5}, {1, 10}}, {{1, 19, 2, 2}, {1, 19, 2, 9}}}},
+};
```
chenhu-wang, line 309: 1. I am not sure how many threads is used in single layer test, but some machine do not have 19 cores in total, that never trigger the code path. At least we can reduce Channel size(19) to cover code path with higher possibility.
2. use 5D shape, and set OD not to 1, to test?

chenhu-wang: and why not put these test shapes in inputShapes_5D in line 286, to avoid duplication.

src/plugins/intel_cpu/src/nodes/reduce.cpp
xuchen-intel: Thanks Chenhu for the comments! As layout nspc reuses the kernel of ncsp layout. Dimensions have been reinterpreted here https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/nodes/reduce.cpp#L2829. During the reinterpretation, B,C,D,H,W always refer to the 1st to 5th physical dimensions respectively in both nspc and ncsp. So OD is exactly the original OH you are mentioning. Further discussions are always welcomed. Thanks!

chenhu-wang: Ok, I see the mapping. But my major point is that this schedule optimization is generic (to extend parallel to next level dimension if current parallel num is small to not fully use all cores . For example, given a tensor with logical shape of {1, 128, 1, 2, 64}. Blocked shape for nspc layout is {1, 1, 2, 64, 128}. In this case, we can parallel for (1*1*2*64) instead of (1*1*2) in if there are 32 or 64 or some larger number threads in the stream. Current extend to 1*1*2, but not to 1*1*2*64. What do you think?

xuchen-intel: Good point! Applied. Thanks Chenhu!

chenhu-wang (approved):

dmitry-gorokhov (approved):

dmitry-gorokhov:
@xuchen-intel Please reabase the PR on latest master. We have another [Reduce changes](https://github.com/openvinotoolkit/openvino/pull/17828) merged already so need to verify together.

PR #17629, opened 2023-05-19
[PT FE] Add aten::LogSoftmax

PiotrKrzem:
### Details:
 - Add aten::LogSoftmax implementation & tests

### Tickets:
 - 110804


tests/layer_tests/pytorch_tests/test_log_softmax.py
```diff
@@ -0,0 +1,40 @@
...
+        [np.random.randn(10, 13, 11), torch.float32]
+    ])
+    @pytest.mark.parametrize("dim", [
+        0,
+        1, 
+        -1
+    ])
+    @pytest.mark.nightly
+    @pytest.mark.precommit
+    def test_log_softmax(self, input_tensor, dim, dtype, ie_device, precision, ir_version):
```
mvafin, line 37: Please add comment that dtype can't be int and input type cant be int by pytorch limitation

src/frontends/pytorch/src/op/log_softmax.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+OutputVector translate_log_softmax(const NodeContext& context) {
+    /*
+    aten::log_softmax(
+        Tensor input,
+        int64 dim,
+        dtype dtype = None
+    )
+    */
+    num_inputs_check(context, 1, 3);
```
mvafin, line 24:
```suggestion
    num_inputs_check(context, 2, 3);
```
You access `dim` on line 26 without checking if it exist

tests/layer_tests/pytorch_tests/test_log_softmax.py
```diff
@@ -0,0 +1,40 @@
...
+        self.dtype = dtype
+
+    def forward(self, input_tensor):
+        return F.log_softmax(input_tensor, dim = self.dim, dtype = self.dtype)
+
+class TestLogSoftmax(PytorchLayerTest):
+    def _prepare_input(self):
+        return (self.input_tensor,)
+
+    @pytest.mark.parametrize(["input_tensor", "dtype"], [
```
mvafin, line 24: Please reimplement to create tensor in `_prepare_input` otherwise it makes test name too long and makes it much harder to debug such test fails

mvafin (changes requested):

tests/layer_tests/pytorch_tests/test_log_softmax.py
PiotrKrzem: Done!

tests/layer_tests/pytorch_tests/test_log_softmax.py
PiotrKrzem: Done!

tests/layer_tests/pytorch_tests/test_log_softmax.py
```diff
@@ -0,0 +1,47 @@
...
+class TestLogSoftmax(PytorchLayerTest):
+    def _prepare_input(self):
+        if self.input_dtype == torch.float:
+            self.input_tensor = np.random.randn(5, 9, 7)
+        else:
+            self.input_tensor = np.random.randint(-100, 100, (5, 9, 7))
+        return (self.input_tensor,)
+
+    @pytest.mark.parametrize(["input_dtype", "convert_dtype"], [
+        # For a tensor with element dtype int, it is necessary to
```
mvafin, line 29: Please take a look on #17418 PR where same issue was fixed on operation level. Basically you need to align input type to fp32

mvafin (changes requested):

tests/layer_tests/pytorch_tests/test_log_softmax.py
PiotrKrzem: Done, although I couldn't add a test for it since pytorch crashes when input dtype == int & convert dtype == None

mvafin: Maybe I misunderstood your comment then. What if input is int, and output type is int, does it work? If pytorch doesn't do automatic int to float conversion then this change to align types is not really needed

PiotrKrzem: Convert_dtype happens before log_softmax, it's not the output type - in your example, pytorch raises `RuntimeError: "log_softmax_lastdim_kernel_impl" not implemented for 'Int'`

mvafin: Please remove align dtype if pytorch itself do not support such case. We may return it in future if pytorch will be supporting this, but now it is just an extra step that does nothing

mvafin (approved):

PR #17630, opened 2023-05-19
GroupNormalization op specification

tomdol:
### Details:
 - Specification of the new normalization operator
 - Specified according to the original paper https://arxiv.org/pdf/1803.08494.pdf and the ONNX version of this op https://github.com/onnx/onnx/blob/main/docs/Operators.md#GroupNormalization

### Tickets:
 - 101515

### DIFF
--- a/docs/ops/normalization/GroupNormalization_12.md
+++ b/docs/ops/normalization/GroupNormalization_12.md
@@ -1,0 +1,79 @@
+# GroupNormalization {#openvino_docs_ops_normalization_GroupNormalization_12}
+
+@sphinxdirective
+
+**Versioned name**: *GroupNormalization-12*
+
+**Category**: *Normalization*
+
+**Short description**: Performs normalization of the input tensor according to the method described in https://arxiv.org/abs/1803.08494
+
+**Detailed description**
+
+The GroupNormalization operation performs the following transformation of the input tensor:
+
+  `y = scale * (x - mean) / sqrt(variance + epsilon) + bias`
+
+The operation is applied per batch, per group of channels. This means that the example input with `N x C x H x W` layout is transformed to the `N x G x C/G x H x W` form. The `scale` and `bias` coefficients are the inputs to the model and need to be specified separately for each group. The `mean` and `variance` are calculated for each group and applied accordingly.
+
+**Attributes**
+
+* *num_groups*
+
+  * **Description**: Specifies the number of groups `G` that the channel dimesion will be divided into.
+  * **Range of values**: between `1` and the number of channels `C` in the input tensor
+  * **Type**: `int`
+  * **Required**: *yes*
+
+* *epsilon*
+
+  * **Description**: A very small value added to the variance for numerical stability. Ensures that division by zero does not occur for any normalized element.
+  * **Range of values**: a positive floating-point number
+  * **Type**: `float`
+  * **Required**: *yes*
+
+**Inputs**
+
+* **1**: `data` - The input tensor to be normalized. The type of this tensor is *T*. The tensor's shape is arbitrary but the first two dimensions are interpreted as `batch` and `channels` respectively. **Required.**
+
+* **2**: `scale` - 1D tensor of type *T* containing the scale values for each group. The expected shape of this tensor is `[G]` where `G` is equal to the value of the `num_groups` attribute. **Required.**
+
+* **3**: `bias` - 1D tensor of type *T* containing the bias values for each group. The expected shape of this tensor is `[G]` where `G` is equal to the value of the `num_groups` attribute. **Required.**
+
+**Outputs**
+
+* **1**: Output tensor of the same shape and type as the `data` input tensor.
+
+**Types**
+
+* *T*: any supported floating point type.
+
+**Example**
+
+```xml
+<layer ... type="GroupNormalization">
+    <data epsilon="1e-5" num_groups="4"/>
+    <input>
+        <port id="0">
+            <dim>3</dim>
+            <dim>12</dim>
+            <dim>100</dim>
+            <dim>100</dim>
+        </port>
+        <port id="1">
+            <dim>4</dim> <!-- 4 scale values, 1 for each group -->
+        </port>
+        <port id="2">
+            <dim>4</dim> <!-- 4 bias values, 1 for each group -->
+        </port>
+    </input>
+    <output>
+        <port id="3">
+            <dim>3</dim>
+            <dim>12</dim>
+            <dim>100</dim>
+            <dim>100</dim>
+        </port>
+    </output>
+</layer>
+```
--- a/docs/OV_Runtime_UG/Operations_specifications.md
+++ b/docs/OV_Runtime_UG/Operations_specifications.md
@@ -82,6 +82,7 @@
    Greater-1 <openvino_docs_ops_comparison_Greater_1>
    GroupConvolutionBackpropData-1 <openvino_docs_ops_convolution_GroupConvolutionBackpropData_1>
    GroupConvolution-1 <openvino_docs_ops_convolution_GroupConvolution_1>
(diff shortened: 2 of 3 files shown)


docs/ops/normalization/GroupNormalization_12.md
```diff
@@ -0,0 +1,79 @@
...
+  * **Type**: `float`
+  * **Required**: *yes*
+
+**Inputs**
+
+* **1**: `data` - The input tensor to be normalized. The type of this tensor is *T*. The tensor's shape is arbitrary but the first two dimensions are interpreted as `batch` and `channels` respectively. **Required.**
+
+* **2**: `scale` - 1D tensor of type *T* containing the scale values for each group. The expected shape of this tensor is `[G]` where `G` is equal to the value of the `num_groups` attribute. **Required.**
+
+* **3**: `bias` - 1D tensor of type *T* containing the bias values for each group. The expected shape of this tensor is `[G]` where `G` is equal to the value of the `num_groups` attribute. **Required.**
```
mitruska, line 41: In the paper or Pytorch docs the `scale`(gamma) and `bias`(beta) are described as learnable per-channel.
Based on that I would expect that the shape of those inputs will be related to the channels [C] instead of number of groups [G]. 
I understand that [G] is aligned with ONNX, but is there any reason of such inconsistency?

mitruska (commented):
Correctness of `scale` and `bias` shape needs confirmation.

docs/ops/normalization/GroupNormalization_12.md
```diff
@@ -0,0 +1,79 @@
...
+    <output>
+        <port id="3">
+            <dim>3</dim>
+            <dim>12</dim>
+            <dim>100</dim>
+            <dim>100</dim>
+        </port>
+    </output>
+</layer>
+```
```
mitruska, line 79: Try to add at the end of the file to fix the CI:

@endsphinxdirective

docs/ops/normalization/GroupNormalization_12.md
```diff
@@ -0,0 +1,79 @@
...
+
+* **1**: Output tensor of the same shape and type as the `data` input tensor.
+
+**Types**
+
+* *T*: any supported floating point type.
+
+**Example**
+
+```xml
```
mitruska, line 53: Looks like all of the spec documents were refactored to use 
`.. code-block:: cpp` instead of xml for those examples

Reference: https://github.com/openvinotoolkit/openvino/pull/17363

docs/ops/normalization/GroupNormalization_12.md
```diff
@@ -0,0 +1,79 @@
...
+
+**Short description**: Performs normalization of the input tensor according to the method described in https://arxiv.org/abs/1803.08494
+
+**Detailed description**
+
+The GroupNormalization operation performs the following transformation of the input tensor:
+
+  `y = scale * (x - mean) / sqrt(variance + epsilon) + bias`
+
+The operation is applied per batch, per group of channels. This means that the example input with `N x C x H x W` layout is transformed to the `N x G x C/G x H x W` form. The `scale` and `bias` coefficients are the inputs to the model and need to be specified separately for each group. The `mean` and `variance` are calculated for each group and applied accordingly.
```
mitruska, line 17: Currently double ticks are used in the docs.
Reference: https://github.com/openvinotoolkit/openvino/pull/17158

```suggestion
The operation is applied per batch, per group of channels. This means that the example input with ``N x C x H x W`` layout is transformed to the ``N x G x C/G x H x W`` form. The ``scale`` and ``bias`` coefficients are the inputs to the model and need to be specified separately for each group. The ``mean`` and ``variance`` are calculated for each group and applied accordingly.
```

docs/ops/normalization/GroupNormalization_12.md
```diff
@@ -0,0 +1,79 @@
...
+
+**Category**: *Normalization*
+
+**Short description**: Performs normalization of the input tensor according to the method described in https://arxiv.org/abs/1803.08494
+
+**Detailed description**
+
+The GroupNormalization operation performs the following transformation of the input tensor:
+
+  `y = scale * (x - mean) / sqrt(variance + epsilon) + bias`
```
mitruska, line 15: Consider usage of `.. math::` for this formula
Reference: https://github.com/openvinotoolkit/openvino/pull/17158

mitruska (commented):
Few additional comments regarding the latest update of the documentation format (and CI).
Please follow the rules in https://github.com/openvinotoolkit/openvino/pull/17363, https://github.com/openvinotoolkit/openvino/pull/17158 (as example).

docs/ops/normalization/GroupNormalization_12.md
tomdol: I've done more research and I think ONNX's documentation is incorrect, it should expect the equality with channel dimension(like in the paper) and not with `G`. I've changed it in the spec.

docs/ops/normalization/GroupNormalization_12.md
mitruska: Okay, thanks!

docs/ops/normalization/GroupNormalization_12.md
```diff
@@ -0,0 +1,85 @@
...
+            <dim>12</dim>
+            <dim>100</dim>
+            <dim>100</dim>
+        </port>
+        <port id="1">
+            <dim>4</dim> <!-- 4 scale values, 1 for each group -->
+        </port>
+        <port id="2">
+            <dim>4</dim> <!-- 4 bias values, 1 for each group -->
+        </port>
```
mitruska, line 71: To align with the latest version:
```suggestion
        <port id="1">
            <dim>12</dim> <!-- 12 scale values, for each channel -->
        </port>
        <port id="2">
            <dim>12</dim> <!-- 12 bias values, for each channel -->
        </port>
```

docs/ops/normalization/GroupNormalization_12.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Detailed description**
+
+The GroupNormalization operation performs the following transformation of the input tensor:
+
+.. math::
+
+   y = scale * (x - mean) / sqrt(variance + epsilon) + bias
+
+The operation is applied per batch, per group of channels. This means that the example input with ``N x C x H x W`` layout is transformed to the ``N x G x C/G x H x W`` form. The ``scale`` and ``bias`` coefficients are the inputs to the model and need to be specified separately for each group. The ``mean`` and ``variance`` are calculated for each group and applied accordingly.
```
mitruska, line 19:
```suggestion
The operation is applied per batch, per group of channels. This means that the example input with ``N x C x H x W`` layout is transformed to the ``N x G x C/G x H x W`` form. The ``scale`` and ``bias`` coefficients are the inputs to the model and need to be specified separately for each channel. The ``mean`` and ``variance`` are calculated for each group and applied accordingly.
```

mitruska (approved):

docs/ops/normalization/GroupNormalization_12.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Detailed description**
+
+The GroupNormalization operation performs the following transformation of the input tensor:
+
+.. math::
+
+   y = scale * (x - mean) / sqrt(variance + epsilon) + bias
+
+The operation is applied per batch, per group of channels. This means that the example input with ``N x C x H x W`` layout is transformed to the ``N x G x C/G x H x W`` form. The ``scale`` and ``bias`` coefficients are the inputs to the model and need to be specified separately for each channel. The ``mean`` and ``variance`` are calculated for each group.
```
vladimir-paramuzov, line 19: @tomdol How it's supposed to work for other ranks (5d or 3d for example)? Does it always split input_shape[1]? Don't we have cases where other axis is needed?

PR #17634, opened 2023-05-19
[CPU] fix assert in oneDNN dw conv

EgorDuplensky:


EgorDuplensky:
Removed "need tests" since the issue exists only with Debug build

PR #17650, opened 2023-05-22
enable 1-D axis in cumsum

xczhai:
### Details:
 - *enable 1-D axis in cumsum, which is aligned with ort*

### Tickets:
 - *111483*

### DIFF
--- a/src/frontends/onnx/frontend/src/op/cum_sum.cpp
+++ b/src/frontends/onnx/frontend/src/op/cum_sum.cpp
@@ -7,6 +7,7 @@
 #include <memory>
 
 #include "default_opset.hpp"
+#include "utils/reshape.hpp"
 
 namespace ngraph {
 namespace onnx_import {
@@ -22,10 +23,7 @@ OutputVector cum_sum(const Node& node) {
     if (inputs.size() > 1) {
         // optional input, 0-D or 1-D tensor
         const auto& axis_shape = inputs.at(1).get_partial_shape();
-        NGRAPH_CHECK(axis_shape.rank().get_length() == 0 || axis_shape.rank().get_length() == 1,
-                     "axis shape should be 0-D or 1-D.");
-        axis =
-            axis_shape.rank().get_length() == 0 ? inputs.at(1) : std::make_shared<default_opset::Squeeze>(inputs.at(1));
+        axis = axis_shape.is_dynamic() ? inputs.at(1) : ngraph::onnx_import::reshape::interpret_as_scalar(inputs.at(1));
     } else {
         axis = default_opset::Constant::create(element::i64, Shape{}, {0});  // default
     }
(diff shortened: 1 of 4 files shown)


src/frontends/onnx/frontend/src/op/cum_sum.cpp
```diff
@@ -20,7 +20,12 @@ OutputVector cum_sum(const Node& node) {
     Output<ngraph::Node> axis;
 
     if (inputs.size() > 1) {
-        axis = inputs.at(1);  // optional input, 0-D tensor
+        // optional input, 0-D or 1-D tensor
+        const auto& axis_shape = inputs.at(1).get_partial_shape();
+        NGRAPH_CHECK(axis_shape.rank().get_length() == 0 || axis_shape.rank().get_length() == 1,
```
tomdol, line 25: There's no axis input check in the core operator intentionally so I'd rather not add it here either https://github.com/openvinotoolkit/openvino/blob/master/src/core/src/op/cum_sum.cpp#L47. @mitruska any thoughts?

src/frontends/onnx/frontend/src/op/cum_sum.cpp
```diff
@@ -20,7 +20,12 @@ OutputVector cum_sum(const Node& node) {
...
+                     "axis shape should be 0-D or 1-D.");
+        axis =
```
tomdol, line 27: Although ONNX only allows scalar axis input I think this is ok. I'd like to see it demonstrated with even a single test though. Please add it to this file https://github.com/openvinotoolkit/openvino/blob/master/src/frontends/onnx/tests/onnx_import.in.cpp#L621 and use any of the existing CumSum tests as a template.

tomdol (changes requested):

src/frontends/onnx/frontend/src/op/cum_sum.cpp
```diff
@@ -20,7 +20,12 @@ OutputVector cum_sum(const Node& node) {
...
+            axis_shape.rank().get_length() == 0 ? inputs.at(1) : std::make_shared<default_opset::Squeeze>(inputs.at(1));
```
mitruska, line 28: To avoid issues `get_legth()` shouldn't be called without dynamism check, 
my suggestion is to check the shape and call the helper from `#include "utils/reshape.hpp"`:

```suggestion
            axis_shape.is_dynamic() ? inputs.at(1) : reshape::interpret_as_scalar(inputs.at(1));
```

wenjiew:
@xczhai Can we look into the review feedback and address these? Thanks!

xczhai:
fix according to the comments.
new implement can support the pattern below. The axis can be 1-D tensor.
image

mitruska (approved):

mbencer (approved):

PR #17657, opened 2023-05-22
[GNA]Fix message in case of error when generating PWL segments for Id…

marcinkusm:
### Details:
 - Fixed message to be more descriptive for the user in case of error during PWL segments generation.

### Tickets:
 - *111357*

### DIFF
--- a/src/plugins/intel_gna/src/backend/pwl_segments_creator_identity.cpp
+++ b/src/plugins/intel_gna/src/backend/pwl_segments_creator_identity.cpp
@@ -114,7 +114,8 @@ void PWLSegmentsCreatorIdentity::UpdateSegmentOnTheLeftOf0_0(const gna_pwl_segme
     auto segment_1_in_int64 = ConvertSegmentTo64(segment_1);
 
     if (segment_1_in_int64.slope == 0) {
-        THROW_GNA_EXCEPTION << "Slope is 0 possible division by 0 when updating left segment!.";
+        THROW_GNA_EXCEPTION << "Slope is 0 possible division by 0 when updating left segment!. Check if your input "
+                               "data for POT were valid.";
     }
 
     int64_t new_segment_1_x_base =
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gna/src/backend/pwl_segments_creator_identity.cpp
```diff
@@ -114,7 +114,8 @@ void PWLSegmentsCreatorIdentity::UpdateSegmentOnTheLeftOf0_0(const gna_pwl_segme
     auto segment_1_in_int64 = ConvertSegmentTo64(segment_1);
 
     if (segment_1_in_int64.slope == 0) {
-        THROW_GNA_EXCEPTION << "Slope is 0 possible division by 0 when updating left segment!.";
+        THROW_GNA_EXCEPTION << "Slope is 0 possible division by 0 when updating left segment!. Check if your input "
+                               "data for POT were valid.";
```
sirzabek, line 118:
```suggestion
        THROW_GNA_EXCEPTION << "PWL segment slope is 0 - possible division by 0 when updating left segment!. Check if your input "
                               "data for POT are valid.";
```

src/plugins/intel_gna/src/backend/pwl_segments_creator_identity.cpp
```diff
@@ -139,7 +140,8 @@ gna_pwl_segment_t PWLSegmentsCreatorIdentity::CreateSegmentOnTheRight(const gna_
     int64_t right_y_base_64 = static_cast<int64_t>(border_values.y_upper);
 
     if (segment_0_0_in_int64.slope == 0) {
-        THROW_GNA_EXCEPTION << "Slope is 0 possible division by 0 when calculating right segment!.";
+        THROW_GNA_EXCEPTION << "Slope is 0 possible division by 0 when calculating right segment!. Check if your input "
+                               "data for POT were valid.";
```
sirzabek, line 144:
```suggestion
        THROW_GNA_EXCEPTION << "PWL segment slope is 0 - possible division by 0 when calculating right segment!. Check if your input "
                               "data for POT are valid.";
```

src/plugins/intel_gna/src/backend/pwl_segments_creator_identity.cpp
sirzabek: I suggest to add some more information about the particular activation function or other layer which caused the issue, if possible.

sirzabek (approved):
The ticket is closed, why is the PR not closed then?

PR #17660, opened 2023-05-22
[GPU] Mvn skipconfig update and typos fix

p-durandin:
### Details:
 - MVN skipconfig update
 - Tests for 8d tensors enables
 - Typos fixes

### Tickets:
 - no (sporadic fails)

### DIFF
--- a/src/plugins/intel_gpu/tests/functional/single_layer_tests/dynamic/mvn.cpp
+++ b/src/plugins/intel_gpu/tests/functional/single_layer_tests/dynamic/mvn.cpp
@@ -96,17 +96,6 @@ namespace {
 const std::vector<InputShape> inputShapes_1D = {
        {
            // dynamic
-           {-1},
-           // target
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/tests/functional/single_layer_tests/dynamic/mvn.cpp
```diff
@@ -283,29 +247,5 @@ const auto Mvn1D = ::testing::Combine(
```
vladimir-paramuzov, line 247: What about test case above? I think all test cases which use `empty_reduction_axes` and `empty_reduction_axes` itself should be removed.

p-durandin: Updated

vladimir-paramuzov (approved):

PR #17668, opened 2023-05-23
[GPU] Enable runtime buffer fusing for dynamic shape

yeonbok:
### Details:
 - Apply in place concat for dynamic shape at runtime

### Tickets:
 - 91381


e-ddykim (approved):
Looks good to me

PR #17670, opened 2023-05-23
[GPU] primitive serialization

e-ddykim:
### Details:
 - This PR enables serialization of all primitives.

### Tickets:
 - 110986


src/plugins/intel_gpu/include/intel_gpu/graph/serialization/activation_serializer.hpp
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2018-2023 Intel Corporation
```
andrew-k-park, line 1: nit: copyright notices for the new source code
```suggestion
// Copyright (C) 2023 Intel Corporation
```

src/plugins/intel_gpu/include/intel_gpu/graph/serialization/tensor_serializer.hpp
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2018-2023 Intel Corporation
```
andrew-k-park, line 1: nit: same as above

andrew-k-park (approved):

src/plugins/intel_gpu/include/intel_gpu/graph/serialization/activation_serializer.hpp
e-ddykim: I updated the copyright year. Thank you.

src/plugins/intel_gpu/include/intel_gpu/graph/serialization/tensor_serializer.hpp
e-ddykim: I updated the copyright year here too. Thank you.

ahnyoung-paul (approved):

yeonbok:
As far as I understand, the primitive serilization is only needed for dynamic shape. Then, can we do it only for dynamic program to reduce the cached bin size for static model?

e-ddykim:
> As far as I understand, the primitive serilization is only needed for dynamic shape. Then, can we do it only for dynamic program to reduce the cached bin size for static model?

Some primitives (such as quantize) are also required for static models too. And the increased blob size is not much. So, if possible, I'd like to use the same logic for static and dynamic models.

PR #17675, opened 2023-05-23
Remove some preprocessing API from GPU and CPU plugins

ilyachur:
### Details:
 - A part of PR: https://github.com/openvinotoolkit/openvino/pull/17571

### Tickets:
 - *ticket-id*


ilya-lavrenov (approved):

iefode (approved):

src/inference/src/dev/core_impl.cpp
```diff
@@ -595,19 +595,7 @@ ov::SoPtr<ov::ICompiledModel> ov::CoreImpl::compile_model_with_preprocess(ov::Pl
...
-
-        auto cloned_model = model->clone();
-        manager.run_passes(cloned_model);
-        preprocessed_model = cloned_model;
-    }
-
-    return context._impl ? plugin.compile_model(preprocessed_model, context, config)
-                         : plugin.compile_model(preprocessed_model, config);
+    return context._impl ? plugin.compile_model(model, context, config) : plugin.compile_model(model, config);
 }
```
songbell, line 599: ilya, where is PPP handled in the case of old api call + cpu 2.0 plugin?

ilyachur: This PR move all plugins to use 2.0 pre-processing. No difference between old and new cases

PR #17676, opened 2023-05-23
Multiple fixes to get list of ops from torchvision detection models

mvafin:
### Details:
 - *These changes allow to see list of unconverted operations in detection models from `torchvision`*
 - *This PR shouldn't be merged as is*

### Tickets:
 - *None*

### DIFF
--- a/src/core/src/op/util/multi_subgraph_base.cpp
+++ b/src/core/src/op/util/multi_subgraph_base.cpp
@@ -6,6 +6,7 @@
 
 #include "ngraph/graph_util.hpp"
 #include "ngraph/opsets/opset5.hpp"
+#include "openvino/util/log.hpp"
 
 ov::op::util::MultiSubGraphOp::InputDescription::InputDescription(uint64_t input_index, uint64_t body_parameter_index)
     : m_input_index(input_index),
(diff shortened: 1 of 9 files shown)


src/frontends/pytorch/src/translate_session.cpp
```diff
@@ -220,8 +220,15 @@ OutputVector TranslateSession::convert_node(const NodeContext& context) {
         OPENVINO_DEBUG << "Some exception happened during conversion of node of type: " << context.get_op_type()
                        << '\n';
     }
-    // Create PtFrameworkNode for everything that wasn't able to be converted normally
-    return make_framework_node(context);
+    try {
+        // Create PtFrameworkNode for everything that wasn't able to be converted normally
+        return make_framework_node(context);
```
rkazants, line 225: Really interesting that creation of FW node can throw an exception. How can it happen?:)

src/core/src/op/util/multi_subgraph_base.cpp
```diff
@@ -160,7 +161,17 @@ void ov::op::util::MultiSubGraphOp::validate_and_infer_type_body(
...
+    try {
+        body->validate_nodes_and_infer_types();
+    } catch (std::exception& e) {
+        if (m_bodies.size() > 1) {
+            // If one body cannot be validated it may mean that it is not needed and will be removed on future stages
+            OPENVINO_DEBUG << "Exception happened during validation of one of the body of MultiSubGraphOp: " << e.what()
+                           << '\n';
+        } else {
+            throw e;
+        }
```
rkazants, line 173: Suspicious change because I think we won't have the situation that some body graph for TF FE will be skipped. Can we do this change only for PT FE?

src/frontends/pytorch/src/translate_session.cpp
mvafin: If body of Loop lets say fails for some reason on validation, but it is a good point, maybe better solve that during fw node creation

mvafin:
Closing due to no longer required.

PR #17678, opened 2023-05-23
[Snippets] Documentation update

IvanNovoselov:
### Details:
 - *Update Snippets' documentation to reflect the present state of the pipeline*

### Tickets:
 - *75569*

### DIFF
--- a/src/common/snippets/docs/snippets_design_guide.md
+++ b/src/common/snippets/docs/snippets_design_guide.md
@@ -1,301 +1,465 @@
-# SnippetS Design Guide
-This document describes the design and rationale for a snippets code generator. Implementation of code functionality is located [here](https://github.com/openvinotoolkit/openvino/tree/master/src/common/snippets). A proposal for CPU backend integration is [here](https://github.com/openvinotoolkit/openvino/pull/2824).
+# Snippets Design Guide
+## Introduction
 
-## Rationale
+Any artificial neural network topology can be represented as a sequence of mathematical operations (or `Ops`) on some data. The first thing we need to do in order to evaluate an `Op` is to load its input data to the processing unit (typically CPU or GPU, we consider CPU below for definiteness). Then the unit can perform all the necessary manipulations with the data. When this stage is finished, the computation results should be stored back to memory to make space for the next portion of input data, and also to make the results available to subsequent `Ops`. So basically an `Op` execution time is defined by two factors: how fast we can move data from the memory to the processor (and back), and how fast the processor can perform the computations. These factors are called memory and computation bandwidth, respectively.
 
-Core **CNN operators (convolution, gemm, fully connected) are limited by compute, the rest is memory bound**. Math approximations (like transcendental functions) are rare in emerging workloads and could be treated with the same machinery. **Snippets are designed to optimize topology for memory**, while leaving compute intensive kernels for backend developers.
+Due to significant advancements in microprocessor technologies, computational bandwidth has grown faster than the memory bandwidth over the past decades. As a result, the majority of `Ops` are memory-bound, so their execution time is controlled by a memory bandwidth. The rare exceptions include convolutions and matrix multiplications. These exceptions could be very important for some workloads, so a lot of work is done to increase their computational bandwidth (Advanced Matrix Registers for example). However, this topic is covered elsewhere and here we will focus on more common memory-bound `Ops`.
 
-The **potential speedup is proportional to shrink in memory-walked bytes**. Therefore, you can transform the problem to a task to optimize for memory walks, whatever pattern snippet has and operations it contains. The number of memory walks should be less or equal to handcrafted optimizations. This guarantees performance improvements over the previous approach (excluding corner cases caused by cache effects). *Shrinkage factor might be encoded to some cost function in future evolution of code generator*. Snippets generator provides diagnostics to estimate this shrinkage factor with `ov::snippets::op::Subgraph::print_statistics(bool verbose)` member.
-
-The SnippetS generator is designed for back-end developers. The main purpose of inventing the snippets code generator is an **operator fusion**, **register allocation** and **target kernel generation** decomposition. This allows modifications (like new fusion support) and feature extensions (like new operation support) to be done in a single point of modification and avoid combinatorial explosion for fusions/types/architectures etc.
-
-Creating a full-fledged compiler or usage of existing compiler infrastructure (like LLVM & MLIR) is superfluous at this point of evolution. The aim is to provide a **flexible and performant framework for operation fusions**, leaving micro optimizations (for example, instruction scheduling) to the backend H/W.
-
-There are no plans to invent a DSL for SnippetS. DSL gives users more flexibility to express uncommon operations. However, the shift towards an approach to encode topologies with elementary operations followed by smart enough fusions is already expressive and performant enough.
-
-**Snippet** is a compiled compute **kernel** generated from a subgraph using the SnippetS code generator for a specific architecture with a **scheduling domain**. Using this scheduling domain and calling convention backend can execute generated compute kernels. For the first generation, snippets are **statically scheduled towards the output domain**. Multi-output snippets are supported if all outputs are broadcast-compatible in a sense that domains for all outputs can be broadcasted from one root domain that defines snippet schedule. It is a subject of extension for future generations.
-
-nGraph is used as the highest level IR for subgraph representation and lowering transformations. **Opset1** is a base operation set for code generation. The aim is to **keep the minimal possible and sufficient operation set** (or ISA) and keep it **RISC-like** (memory and compute decomposed).
-
-**One subgraph corresponds to one snippet**. Operations which cannot be scheduled by a single schedule should not be placed in the same subgraph. A snippet is somewhat conceptually close to OpenCL kernel without a restriction to express only embarrassingly parallel tasks.
-**Subgraph** once extracted from full topology IR is **treated as an operation and data flow descriptor in scalar notation** (similar to OpenCL/CUDA). Tensor sizes are used for defining scheduling domain and detecting broadcasts/reductions.
-
-Operations are split into 3 groups: **layout-oblivious (LOO), layout-aware(-tolerant) and layout-dependent(-specific)**. **Layout-oblivious** operation semantics and implementation are completely agnostic to a specific layout in which tensors are placed in memory. For example, like elements-wise math and ReLU in this category. Implementation of **layout-aware** operation depends on the layout of input/output tensors. For example, convolutions and other block-wise kernels or layout repacks. **Layout-specific** operation semantics and implementation depend on the layout. For example, the Yolo region. Patterns to fuse are constructed in terms of taxonomy above.
-
-## Design
-
-Code generation is split into 2 phases, **tokenization** and **lowering**.
-
-### Tokenization
-
-Tokenization runs on full topology nGraph function inside a specific plugin in a stage of common transformations. Input of tokenization is a topology graph. Output is a modified topology graph with `ov::snippets::op::Subgraph` operations installed. Each subgraph contains nGraph function (called **body**) which holds a part of original topology legal for snippet generation (can be scheduled with a single schedule).
-
-A procedure of finding subgraphs suitable for code generation is called **tokenization**. During tokenization the topology tree is split into subgraphs in the same greedy approach which is used for parsing input stream of characters into the tokens. It may also be seen as and modified into a basic block construction problem, since there is a leader and potentially terminators. See the example of implementation [here](https://github.com/openvinotoolkit/openvino/blob/master/src/common/snippets/src/pass/collapse_subgraph.cpp).
-
-Tokenization has an advantage over the pattern matching approach (used in traditional and MLIR-based compilers) since it can handle arbitrary patterns of operations. Pattern matching deduces specific configuration of operations to translate to another one, more suitable for target machine or further lowering. This means that relations between operations are fixed. Tokenization, on the other hand, has the only limitation on specific operation types which are **suitable and profitable** to fuse, respecting original topology correctness (keeping it as a direct acyclic graph).
-
-The extracted body comes to a plug-in wrapped as a composite `Subgraph` operation which is seen as a block box from a plugin standpoint and can participate in any plugin specific subroutines (for example, layout assignment, memory allocation, etc.).
-
-### Supported subgraph patterns
-
-Subgraph accepts arbitrary numbers of inputs and outputs. There is 1:1 mapping for external (subgraph node’s) and internal (body) parameters indexes.
-
-Pattern here is an exact subgraph configuration (nodes and edges between them). **The first generation of snippets supports only layout-oblivious operations which may have broadcast on inputs and broadcast-compatible outputs**. For example Shapes `<1, 42, 17, 31>`, `<1, 42, 17, 1>` and `<1, 42, 1, 31>` are considered as broadcast-compatible. Layout-oblivious operation with multiple outputs serves as a snippet leader and forms a new subgraph. The most beneficial patterns are subgraphs with complex control flow but minimal number of inputs/and outputs. For example, GeLU has a 5x shrinkage factor from original unfused subgraph in number of bytes walked. Subgraph below could be considered as an example of such a subgraph. Leader detection procedure aims to find such subgraphs.
-
+A typical approach to accelerate execution of memory-bound `Ops` is to perform vertical fusings (or simply fusings below). It is important to understand the concept of fusings, so let's cover it in more detail. Imagine we need to execute two operations as a part of our workflow: `Op 1` and `Op 2`, as shown on the picture below.
 ```mermaid
- flowchart LR
-    nodeA1(...) --> nodeA2(Add)
-    nodeA2(Add) --> nodeA3(Add)
-    nodeA2(Add) --> nodeA5(Multiply)
-    nodeA3(Add) --> nodeA4(Clamp)
-    nodeA4(Clamp) --> nodeA5(Multiply)
-    nodeA5(Multiply) --> nodeA6(...)
+ graph LR
+    Tensor1[Tensor1] 
+    op1(Op 1)
+    Tensor2[Tensor2]
+    Tensor1 --> op1 --> Tensor2
+    op2(Op 2)
+    Tensor3[Tensor3]
+    Tensor2 --> op2 --> Tensor3
 classDef no-bg-color fill:none,stroke-width:0px
 classDef steel1 fill:#B9D6E5, stroke: #86B3CA, color: #262626
 classDef daisy1 fill:#FFE17A, stroke: #FEC91B, color: #262626
-class nodeA1,nodeA6 no-bg-color
-class nodeA2,nodeA3 daisy1
-class nodeA4,nodeA5 steel1
-class nodeA3 steel1
+%%class op1,op2 steel1
+class Tensor1,Tensor2,Tensor3 daisy1
+```
+`Op 1` takes `Tensor 1` as a single input and produces `Tensor 2` as an output, this tensor is then read by `Op 2` that in turn produces `Tensor 3`. 
+
+A typical execution pipeline of this minimalistic example is depicted on the next figure. Note that the pipeline is usually deployed on multicore processing units which can run several execution threads in parallel.
+```mermaid
+ graph LR
+    Tensor1[Tensor1]
+    subgraph ParallelSection1 [Parallel Section]
+        op1_0(Op 1 Kernel)
+        op1_1(...)
+        op1_2(...)
+        op1_3(Op 1 Kernel)
+    end
+    Tensor2[Tensor2]
+    Tensor1 --> op1_0 --> Tensor2
+    Tensor1 --> op1_1 --> Tensor2
+    Tensor1 --> op1_2 --> Tensor2
+    Tensor1 --> op1_3 --> Tensor2
+   subgraph ParallelSection2 [Parallel Section]
+      op2_0(Op 2 Kernel)
+      op2_1(...)
+      op2_2(...)
+      op2_3(Op 2 Kernel)
+   end
+   Tensor3[Tensor3]
+   Tensor2 --> op2_0 --> Tensor3
+   Tensor2 --> op2_1 --> Tensor3
+   Tensor2 --> op2_2 --> Tensor3
+   Tensor2 --> op2_3 --> Tensor3
+   
+classDef no-bg-color fill:none,stroke-width:0px
+classDef steel1 fill:#B9D6E5, stroke: #86B3CA, color: #262626
+classDef daisy1 fill:#FFE17A, stroke: #FEC91B, color: #262626
+class op1_0,op1_1,op1_2,op1_3 steel1
+class op2_0,op2_1,op2_2,op2_3 steel1
+class Tensor1,Tensor2,Tensor3 daisy1
+```
+ In order to take advantage of the multithreading, the `Op's` input tensor is divided into several blocks that could be processed independently, we will call them `subtensors`. Then each of the threads executes its own copy of `Op's` kernel that reads only a part of the input tensor and writes a corresponding part of the output tensor. As one can see from the picture, tensors serve as synchronisation points between parallel regions. It means that a thread that finished earlier has to wait for all other threads to complete the task, which could impact overall performance of the pipeline.
+
+To understand another potential performance bottleneck, let's have a look at `Op 1 Kernel` implementation scheme which is shown on the graph below.
+```mermaid
+ graph LR
+    InputMemory[In Memory\n Pointer]
+    OutputMemory[Out Memory\n Pointer]
+    subgraph OpKernel [Op Kernel]
+        subgraph Loop [Loop]
+            direction RL
+            Load(Load)
+            op(Op \n instructions)
+            Store(Store)
+            Load --> |VR| op
+            op -->|VR| Store
+            Store -..->|Increment GPRs\n And repeat while needed| Load
+        end
+    end
+   InputMemory -- GPR --> Load
+   Store -- GPR --> OutputMemory
+   
+classDef no-bg-color fill:none,stroke-width:0px
+classDef steel1 fill:#B9D6E5, stroke: #86B3CA, color: #262626
+classDef daisy1 fill:#FFE17A, stroke: #FEC91B, color: #262626
+class OpKernel steel1
+class InputMemory,OutputMemory daisy1
+```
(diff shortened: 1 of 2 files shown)


src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
+# Snippets Design Guide
+## Introduction
 
-## Rationale
+Any artificial neural network topology can be represented as a sequence of mathematical operations (or `Ops`) on some data. The first thing we need to do in order to evaluate an `Op` is to load its input data to the processing unit (typically CPU or GPU, we consider CPU below for definiteness). Then the unit can perform all the necessary manipulations with the data. When this stage is finished, the computation results should be stored back to memory to make space for the next portion of input data, and also to make the results available to subsequent `Ops`. So basically an `Op` execution time is defined by two factors: how fast we can move data from the memory to the processor (and back), and how fast the processor can perform the computations. These factors are called memory and computation bandwidth, respectively.
```
a-sidorova, line 4: It'd be really great if you split the sentences by lines. Markdown thinks that it's new line only if there is empty line between paragraph. It's easier to review and suggest changes

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
 
-The goal for this step is to transform a subgraph (body function) into a form possible for code generation. Input for this step is a subgraph in a canonical form. Output is a subgraph in snippets dialect.
+The `Snippets` is a code generation framework that provides a scalable approach to operations. fusings and  enablement. One of the most distinctive feature of `Snippets` is a flexible yet universal design. They are not bound to a specific backend or concrete kernel implementations, but provide a jit kernel interface instead. This interface allows `Snippets` to perform register assignment as well as a wide range of other backend-independent optimizations. `Snippets` are also highly configurable and allow a backend to provide a custom optimizations that can be executed at different stages of the pipeline. This unique combination of flexibility and scalability is driven by `Snippets` design, so let's discuss it in more detail.
```
a-sidorova, line 156: Seems like there is a mark
`The Snippets is a code generation framework that provides a scalable approach to operation fusings and enablement.`

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
...e processed by `Snippets`. If such a part is found, `Tokenizer` converts it to an `nGraph IR` and stores inside a `Subgraph` node. `nGraph IR` - is one of the two `IR` types used by `Snippets`, it is simply a small `nGraph model` that can contain `Snippets`-specific operations. `nGraph IR` is then passed to the `Optimizer` unit that in turn consists of two subunits. The purpose of the first subunit is to perform data flow optimizations, whereas the second subunit is focused on control flow optimizations. Note that the control flow optimizations are performed on `Linear IR` which was especially designed for this purpose. Finally, the optimized `Linear IR` is used by the `Generator` (which is `Snippets` `Backend`) to produce executable code, which we will refer to as `Kernel`. As discussed in the Introduction, the purpose of the `Kernel` is to process a part of the initial tensor, and several `Kernels` are usually executed in parallel to process the whole tensor. Note that a `Kernel` usually applies several operations (`Op instructions` blocks from the Introduction), the exact number depends on the source model topology: sometimes it's just a couple, but could more than twenty `Ops`.
```
a-sidorova, line 204: I'm not sure but maybe there can be questions how it's stored
`If such a part is found, `Tokenizer` converts it to an nGraph IR and stores inside a Subgraph node.` -> `If such a part is found, `Tokenizer` converts it to an nGraph IR and stores inside a Subgraph node as small nGraph model which is called body.`

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
-
-#### Target-specific optimizations
+The algorithm described above reflects a general idea behind the tokenization process, but there are of course other details that should be considered in a real-life implementation. For instance, the `Tokenizer` should take additional care to avoid creation of cyclic dependencies when attaching a Node to `Subgraph`. Similarly, `Subgraph` is not allowed to connect with more than one `Result` node (no node can do that). In both of the cases, a new `Subgraph` is started instead of attaching to the existing one to avoid an illegal configuration. Another interesting case is the tokenization of constant nodes. If a `Constant` represents a single value (scalar) and its child was attached to a `Subgraph`, then the `Constant` also attached to the same `Subgraph`. If a `Constant` is not scalar, then it can't be tokenized since storing `Constant's` values requires memory management, and memory management is a plugin responsibility. Please refer to the [collapse_subgraph.cpp](https://github.com/openvinotoolkit/openvino/blob/master/src/common/snippets/src/pass/collapse_subgraph.cpp) to gain more insights on the tokenization process.
```
a-sidorova, line 319: `Subgraph is not allowed to connect with more than one Result node (no node can do that)`
I'm not sure that it's good explanation for `Subgraph` :smile:
Because some operations can have several outputs (`MaxPool-8`, for example). For snippets, it's a problem because origin model expects `Results` with special names. But `Subgraph` cannot have several names. So if we fuse ops before `Result` into `Subgraph`, we cannot find the `Result` by name because `Subgraph` has only one name. Moreover, as I remember, outputs are described by tensor names. I think because of that simple `nGraph` nodes can have several outputs

a-sidorova: Also I'd suggest to create the new paragraph with name `limitations` where you can describe limitations of tokenization (count of `Result` nodes, cycle dependencies, the plugin limitation with gpr count - need to notice that it's plugin specific limitation and we're going to remove it - ticket is in code)

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
+To summarize, `Snippets` workflow consists of three major blocks: `Tokenizer`, `Optimizer` and `Generator`. Let's discuss each of them in detail below.
+
+### Tokenizer
+
+`Tokenizer` is run on an `nGraph model` and its main purpose is to identify subgraphs that are suitable for code generation. These subgraphs are then replaced with the `ov::op::Subgraph` node. This stage is called tokenization because the `Tokenizer` employs a greedy  algorithm similar to the ones used for parsing input stream of characters into tokens. One of the distinctive features of this algorithm is its flexibility, so it can seamlessly handle arbitrary operations' patterns. This differentiates the tokenization from the pattern matching approach used in traditional and MLIR-based compilers. Pattern matching can indeed process only a limited set of predefined operations' configurations, so the relations between the operations are fixed in this case. Thus, the tokenization's flexibility becomes a significant advantage when the number of new ML topologies increases rapidly, so it becomes more and more expensive to support and extend a large set of patterns.
```
a-sidorova, line 258: Should we add some words about `Tokenization` in general? `Tokenization` has the pipeline of tokenization: `Enumeration` (to enumerate nodes for cycle dependency verification), `MHATokenization` (the pass which tokenizes MHA patterns), `CommonTokenizations` (...) and `CommonOptimizations` (some optimizations before conversion to snippet dialect)

What do you think? I'm not sure that we should explain how `MHATokenization` works but maybe several words and maybe link to article with this pattern and mermaid with skeleton of pattern.

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
+3. `InsertBuffers` - analyzes `LoopInfo` and `Expression` semantics, inserts `snippets::op::Buffer`. `Buffer` is an operation that represents a memory buffer needed to save some intermediate results.
+4. `InsertLoadStore` - inserts explicit memory access operations like `Load` and `Store` (both from `snippets::op`). These are needed, so appropriate instructions will be emitted during the code generation stage to move data between memory and vector registers.
+5. `InitLoops` - inserts explicit `LoopBeegin` and `LoopEnd` (`snippets::op`) operations based on the acquired `LoopInfo`. Again, the explicit operations are needed to emit appropriate instructions later.
+6. `LoadMoveBroadcastToBroadcastLoad` fuse `Load->MoveBroadcast` sequences into a single `BroadcastLoad` expressions.
 
-* nGraph operation (for example, `class FMA : public ov::op::Op`)
-* Emitter for the operation (for example, `class FmaEmitter : public Emitter` )
-* register the pair in `jitters` map
+The buffer pipeline is focused on managing `op::Buffer` operations:
+1. `IdentifyBuffers` - analyzes buffers using graph coloring algorithm to enable memory reuse.
```
a-sidorova, line 436: It's not for memory reuse unfortunately :(
It's just to resolve problem with non-proportional shifts of data pointer after each loop iteration

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
+2. `FuseLoops` - analyzes the assigned `loop_ids` and `LoopInfo`, fuses some loops if possible. This pass can move some `Expressions` up or down the graph, so the `Expressions` with the same `loop_ids` are grouped together.
+3. `InsertBuffers` - analyzes `LoopInfo` and `Expression` semantics, inserts `snippets::op::Buffer`. `Buffer` is an operation that represents a memory buffer needed to save some intermediate results.
+4. `InsertLoadStore` - inserts explicit memory access operations like `Load` and `Store` (both from `snippets::op`). These are needed, so appropriate instructions will be emitted during the code generation stage to move data between memory and vector registers.
+5. `InitLoops` - inserts explicit `LoopBeegin` and `LoopEnd` (`snippets::op`) operations based on the acquired `LoopInfo`. Again, the explicit operations are needed to emit appropriate instructions later.
```
a-sidorova, line 432:
```suggestion
5. `InitLoops` - inserts explicit `LoopBegin` and `LoopEnd` (`snippets::op`) operations based on the acquired `LoopInfo`. Again, the explicit operations are needed to emit appropriate instructions later.
```

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
+6. `LoadMoveBroadcastToBroadcastLoad` fuse `Load->MoveBroadcast` sequences into a single `BroadcastLoad` expressions.
 
-* nGraph operation (for example, `class FMA : public ov::op::Op`)
-* Emitter for the operation (for example, `class FmaEmitter : public Emitter` )
-* register the pair in `jitters` map
+The buffer pipeline is focused on managing `op::Buffer` operations:
+1. `IdentifyBuffers` - analyzes buffers using graph coloring algorithm to enable memory reuse.
+2. `AllocateBuffers` - calculates minimal memory size required to handle all the buffers. The memory will be allocated during the execution stage by the plugin `Snippet` node.
```
a-sidorova, line 437: Maybe we should notice that in the near future we are going to write whole `BufferManager` which calculates memory size for allocation using `MemorySolver` and marks `Buffers` which can be inplace instead of `AllocateBuffers` pass. What do you think?

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
+    
+The final pipeline applies finishing touches before the `IR` will be passed to the `Generator` module:
+1. `PropagateLayout` - propagates data layouts to `Parameters` and `Results` (actually to corresponding `Expressions`), so `Kernel` will be able to calculate appropriate data offsets for every iteration of an external parallel loop.
+2. `CleanupLoopOffsets` - eliminates redundant pointer increments from the loops.
 
-### Calling convention
+More details on control flow optimization passes could be found in the `control_flow_transformations(...)` method inside [subgraph.cpp](https://github.com/openvinotoolkit/openvino/blob/master/src/common/snippets/src/op/subgraph.cpp). When all the passes are applied, the `LinearIR` is handled further to the `Generator` to emit executable code.
```
a-sidorova, line 443: BTW you can use relative links in markdown to avoid the absolute paths in network.
Also you can use reference links (to reuse them, for example)

For [more](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links) information.

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
+2. `InsertTailLoop` injects tail-processing section after a loop body if needed. Note that every loop has two parameters that specify how its body is evaluated: `work_amount` and `increment` The `work_amount` indicates how much of the data needs to be processed, it often equals to the dimension's size the loop is working on. The `increment` defines how many data entries are processed on every loop iteration (it usually equals to vector size for the innermost loops of elementwise subgraph). So if a loop's `work_amount` is not evenly divisible by its `increment`, it means that a tail processing is required. `InsertTailLoop` duplicates the body of such a loop, rescales pointer increments and load/store masks appropriately, and injects these `Ops` immediately after the processed loop.
```
a-sidorova, line 450: Maybe we can describe `Loop` work before as the main part of opset? I mean to explain that to emulate loops there are two ops `LoopBegin` and `LoopEnd`. The loop has the parameters for cycle execution: `work_amount` (loop trip count), `increment` (step), etc. And add the diagram from `Linear IR`

What do you think?
I hope it would be really helpful for new developers to understand how to work with loops

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
+As briefly discussed in the ***Architecture*** section, `Optimizer` consists of two major units: the first one performs data flow optimization, and the second one is focused on control flow. Note however that some data-flow-related passes can be performed only after the control flow optimizations, so the second unit modifies the dataflow as well. Nevertheless, we will refer to the units as `Data flow optimizer` and `Control flow optimizer` to reflect their main purpose. Keep mind that, as discussed above, the `Data flow optimizer` operates exclusively on the `nGraph IR`, while the `Control flow optimizer` works with the `Linear IR`. We will discuss these units in more detail below.
 
-#### Schedule generation
+#### Data flow optimizer
```
a-sidorova, line 329: Should we add a couple of words about precision propagation?

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,465 @@
...
 
 ## See also
 
  * [OpenVINO™ README](../../../../README.md)
- * [OpenVINO SnippetS](../README.md)
+ * [OpenVINO Snippets](../README.md)
  * [OpenVINO Core Components](../../../README.md)
  * [Developer documentation](../../../../docs/dev/index.md)
+
```
a-sidorova, line 465: Maybe it'd be really great if we clarify some limitations (for example, we support only these precisions, support elementwise operations and some operation only from MHA pattern (and links to the code with this list)), future plans.

What do you think? Seems like we should always update this document but I think that we really should do it before each Release.

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: right, thanks

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: I've intentionally tried to minimize the amount of details here, since this section is dedicated to an overall architecture, and too many details could be overwhelming. In fact, I'm not even sure that it's worth to mention two IR types and data/control flow optimization blocks at this point.
So I think it's better to leave the `body` for the optimizations section

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: I thought of us better than we actually are 😁 . Will fix the description
BTW it's a shame we don't rely on buffer ids to optimize memory usage. It looks like we have all the necessary information at hand, let's discuss it offline

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: The purpose of this documentation is to make it easier for the developers to navigate the Snippets pipeline, so the docs should reflect the actual state of the pipeline.
That said, I think that our plans should be in presentations and jira tickets rather than in the docs.

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: The ticket for a more detailed Tokenization description: 111891

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: Thanks for the suggestion! will do

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: Totally agree. Moreover, all Snippets-specific operations should be covered, especially Kernel, LoopBegin, LoopEnd, MemoryAcess class, etc. 
I've created a ticket for this: 111895

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: Agreed. Moreover, we need to create a dedicated section, where precision handling is explained in general, not only the precision propagation pass. 
A ticket for this: 111899

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: Yes, I think it's a good idea. We should create a `limitations` subsection in every section (Tokenization, Precision propagation, Generation), and summarize it (just one sentence + link) in a high-level "Present Limitations" section.

Totally agree, we should dedicate more time to documentation development, and update the docs on a regular basis.

a-sidorova (approved):

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,642 @@
...
+For instance, `PortDescriptor` stores a tensor shape and layout it expects (or utilizes) on the port. 
+It also describes the port's subtensor, which is a part of the tensor that will be processed by a dedicated emitter on a single iteration. 
+This information will be used by the control flow optimization pipeline to determine whether two `Expressions` are compatible (can be executed inside one loop for example), or to calculate proper pointer increments between consecutive loop iterations. 
+An `Expression` internally stores two separate vectors of input and output `PortDescriptors` which could be accessed by calling `get_input_port_descriptors()` or `get_input_port_descriptor(i)` (and similar for outputs).
+
+Finally, `PortConnectors` specify how the `Expression's` ports are connected. 
+Note that an `Expression` output can be connected to several inputs (like with nGraph nodes), So every `PortConnector` stores one source `ExpressionPort` and a set of consumer `ExpressionPorts` that can be accessed by the `get_cource()` or `get_consumers()` methods, respectively.
```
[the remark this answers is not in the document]
IvanNovoselov: fixed, thanks

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,642 @@
...
+`Tokenizer` replaces parts of the `nGraph model` that can be executed by `Snippets` with `ov::op::Subgraph` nGraph nodes. 
+Each of the nodes stores a piece of the initial `nGraph model` that was replaced by the node. 
+This piece is stored as an nGraph model itself, which we refer to  as `nGraph IR` to distinguish from the original `nGraph model`. 
+Note that sometimes the exact type of `IR` is not important in our discussion. 
+In such cases, we will refer to the `IR` (`nGraph` or `Linear`) as `body function`, or simply `body`. 
+
+When the plugin finalizes all `nGraph model` transformations, the model is converted to an internal plugin graph representation. 
+At this point `ov::op::Subgraph` is converted to `ov::intel_cpu::node::Snippet` which still retains the `nGraph IR`. 
+This IR is then optimized and an executable `Kernel` is produced during the `CreatePrimitive()` stage.
```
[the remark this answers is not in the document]
IvanNovoselov: Right, lets change for smth more general. There is initialization already, so let it be `CreateComputePrimitive` maybe?

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,642 @@
...
+It is a plugin responsibility to mark all nodes that (can be tokenized, but) should be ignored by `Snippets` with a `SkippedByPlugin` tag (defined in [tokenization.hpp](../include/snippets/pass/tokenization.hpp)). 
+This is usually implemented as a separate markup transformation that is applied immediately before the tokenization. 
+Please, refer to the [snippets_mark_skipped.cpp](../../../plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp) pass for a detailed example.
+
+### Optimizer
+
+As briefly discussed in the ***Architecture*** section, `Optimizer` consists of two major units: the first one performs data flow optimization, and the second one is focused on control flow. 
+Note however that some data-flow-related passes can be performed only after the control flow optimizations, so the second unit modifies the dataflow as well. 
+Nevertheless, we will refer to the units as `Data flow optimizer` and `Control flow optimizer` to reflect their main purpose. 
+Keep mind that, as discussed above, the `Data flow optimizer` operates exclusively on the `nGraph IR`, while the `Control flow optimizer` works with the `Linear IR`.
```
v-Golubev, line 424: Looks like "Keep in mind" is meant here.
```suggestion
Keep in mind that, as discussed above, the `Data flow optimizer` operates exclusively on the `nGraph IR`, while the `Control flow optimizer` works with the `Linear IR`. 
```

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,642 @@
...
 
-A subgraph can be executed with nGraph references if no generator is present.
+`LinearIR` is our graph representation, it's an analog to an nGraph model. 
+It is simply a container for `Expressions`, the order of `Expressions` represents control flow. 
+`LIR` also incorporates a range of useful methods to manage the `Expressions`, for example `create_expression(...)` to build `Expressions` from nGraph nodes, or `replace_input(...)` to modify data dependencies between `Expressions`. 
+Please refer to the implementation in [linear_ir.cpp](../src/lowered/linear_ir.cpp) for more details.
+
+`Expression` is the main building block of a `Linear IR`. 
+It contains a pointer to the nGraph node it was created from and a pointer to the emitter it will be mapped to (which is null until `Expression::init_emitter(...)` is called). 
+An `Expressions` can have an arbitrary number of inputs and outputs, we will refer to them simply as ports.
```
v-Golubev, line 543:
```suggestion
An `Expression` can have an arbitrary number of inputs and outputs, we will refer to them simply as ports. 
```
or 
```suggestion
`Expressions` can have an arbitrary number of inputs and outputs, we will refer to them simply as ports. 
```

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,642 @@
...
+
+Finally, `PortConnectors` specify how the `Expression's` ports are connected. 
+Note that an `Expression` output can be connected to several inputs (like with nGraph nodes), So every `PortConnector` stores one source `ExpressionPort` and a set of consumer `ExpressionPorts` that can be accessed by the `get_source()` or `get_consumers()` methods, respectively. 
+Like with `PortDescriptors`, an `Expression` stores input and output `PortConnectors` in two separate vectors accessed via `get_input_port_connector(i)` (or its output twin).
+
+An example on how `PortConnectors` can be used to move between `Expressions` is given on the right side of the above picture. 
+So if we need to get a child expression connected to the i-th expression's output, we need to call `expression->get_output_port_connector(i)` first to obtain a corresponding connector, then to call `connecotr->get_consumers()` to obtain a set of connected `ExpressionPorts`, and finally `ExpressionPort->get_expr()` to access a child expression.
```
v-Golubev, line 560:
```suggestion
So if we need to get a child expression connected to the i-th expression's output, we need to call `expression->get_output_port_connector(i)` first to obtain a corresponding connector, then to call `connector->get_consumers()` to obtain a set of connected `ExpressionPorts`, and finally `ExpressionPort->get_expr()` to access a child expression. 
```

v-Golubev (commented):
A clear and well-readable guide for the first acquaintance with the current Snippets pipeline. Great work, thank you!

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,642 @@
...
-
-###### Explicit broadcast
-
-For each operation in body function inputs are checked against broadcasting. When Parameters are to be broadcasted, an explicit broadcast operation is generated. For example, with `<1, 42, 17, 31>` and `<1, 42, 17, 1>` for the subgraph above, the resulting subgraph will be:
-
+In order to make any manipulations with data, the kernel needs to know where the input data is located and where it should place the output data. 
+So the typical kernel takes input and output memory pointers as runtime arguments. 
+In our example, input memory pointers are passed to the kernel in general-purpose registers (note the `GPR` label on the arrow to `Load`). 
+At first, a small portion of data is loaded from this location into a vector register (or several registers), which is designated by the `Load` box and `VR` label on the picture. 
+Then the kernel performs all the necessary manipulations with the registers (the `Op1 instructions` box) and places the result in another vector register.
```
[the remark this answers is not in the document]
IvanNovoselov: Corrected, thanks

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,642 @@
...
-
-#### Schedule generation
-
-The goal of this step is to transform subgraphs in a scalar notation into kernel functions callable from user code. The `Kernel` and `Tile` operations are introduced for this purpose. Each of these operations has a constructor from code region described as a collection of operation and operand pairs `Kernel(const std::vector<std::pair<std::shared_ptr<ov::snippets::Emitter>, ov::snippets::RegInfo>>& region);`.
+## Design
+Before we consider any specific transformations and implementation features, it is very useful to keep in mind a clear coarse-grained picture of the whole pipeline. 
+So we start or discussion with `Snippets` architecture.
```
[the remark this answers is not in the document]
IvanNovoselov: You're right, fixed

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: Fixed, thanks

src/common/snippets/docs/snippets_design_guide.md
IvanNovoselov: Good catch, fixed

src/common/snippets/docs/snippets_design_guide.md
```diff
@@ -1,301 +1,642 @@
...
+As shown on the figure below, `Snippets` are organized in a very similar way.
+```mermaid
+ graph LR
+   Source[nGraph \n model]
+   subgraph Snippets
+      direction LR
+      subgraph Optimizer[Optimizer]
+        direction LR
+        Data[Data flow \n optimizations]
+        Control[Control flow \n optimizations]
```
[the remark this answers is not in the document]
IvanNovoselov: Agree, I also thought about adding this stage after the presentation. Thanks a lot!
Please check the updated graph and the corresponding text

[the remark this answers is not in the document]
v-Golubev: Now the graph as obvious as possible, the changes are corresponded to my expectations. Thanks

dmitry-gorokhov (approved):

PR #17679, opened 2023-05-23
Fix StridedSlice constant folding when disabled attribute set.

praasz:
### Details:
The `StridedSlice` folded to constant even the `ShapeOf` on `StridedSlice` inputs (begin or end) has got set disable constant folding. Add additional check if constant folding is disabled on `StridedSlice`.

### Tickets:
 - 103696

### DIFF
--- a/src/core/src/op/strided_slice.cpp
+++ b/src/core/src/op/strided_slice.cpp
@@ -20,6 +20,7 @@
 #include "ngraph/util.hpp"
 #include "openvino/core/validation_util.hpp"
 #include "openvino/op/util/precision_sensitive_attribute.hpp"
+#include "openvino/pass/constant_folding.hpp"
 #include "strided_slice_shape_inference.hpp"
 
 using namespace std;
@@ -296,7 +297,7 @@ bool op::v1::StridedSlice::evaluate_label(TensorLabelVector& output_labels) cons
 
 bool op::v1::StridedSlice::constant_fold(OutputVector& output_values, const OutputVector& inputs_values) {
     auto is_folded = Node::constant_fold(output_values, inputs_values);
-    if (!is_folded) {
+    if (!is_const_fold_disabled() && !is_folded) {
         // If all ignored mask are set for all begin or end then replace this input by dummy constant
         // to avoid return false from `could_propagate` during bound evaluation (value of const will be ignored).
         auto get_indices_input = [&inputs_values](size_t port, const std::vector<int64_t>& mask) -> Output<Node> {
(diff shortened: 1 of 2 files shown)


src/core/src/op/strided_slice.cpp
```diff
@@ -325,11 +326,17 @@ bool op::v1::StridedSlice::constant_fold(OutputVector& output_values, const Outp
...
-            is_folded = true;
+        std::vector<Node*> nodes;
+        // Check if bounds can be evaluated and none of output nodes have disabled constant folding.
+        if (ov::could_propagate(output, nodes) && std::none_of(nodes.begin(), nodes.end(), [](const Node* n) {
+                return ov::pass::constant_folding_is_disabled(n);
+            })) {
+            OPENVINO_SUPPRESS_DEPRECATED_START
+            if (const auto c = ov::get_constant_from_source(output)) {
+                OPENVINO_SUPPRESS_DEPRECATED_END
+                output_values[0] = c;
```
itikhono, line 337: I guess that we need to copy runtime info here at least from StridedSlice node to this constant, but maybe from input nodes also.
Please refer `Node::constant_fold` implementation

src/core/tests/constant_folding.cpp
```diff
@@ -2424,6 +2425,84 @@ TEST(constant_folding, strided_slice_not_ignored_dynamic_begin_from_shape_of) {
...
+                                                         stride,
+                                                         std::vector<int64_t>{0, 1, 0, 0, 0},
+                                                         std::vector<int64_t>{1, 1, 0, 0, 1});
+    slice->set_friendly_name("test");
+
+    auto model = make_shared<ov::Model>(slice, ParameterVector{p_begin, p_end});
+
+    pass::Manager pass_manager;
+    pass_manager.register_pass<ov::pass::DisableShapeOfConstantFolding>();
+    pass_manager.register_pass<ov::pass::InitNodeInfo>();
```
jane-intel, line 2460: Init Node Info should go before all the passes

jane-intel (approved):
LGTM
Lets trigger nightly-like run on bunch of models

src/core/src/op/strided_slice.cpp
praasz: I follow the `Node` implementation which use all node's inputs to copy `rt_info`. In case of Strides copy from nodes which will be folded into constant.

src/core/tests/constant_folding.cpp
praasz: Corrected.

PR #17681, opened 2023-05-23
Updated list of GPU drivers that used for internal validation

artanokhov:
### Details:
 - Removed gpu-setup-guide.md
 - Updated list of GPU drivers that used for internal validation


kblaszczak-intel (approved):

artanokhov:
Changes will be ported with following PR openvinotoolkit/openvino/pull/17672

PR #17685, opened 2023-05-23
[GNA] convertFunctionToICNNNetwork operation performance improovement

mkwapul:
### Tickets:
 - 93627

### DIFF
--- a/src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
+++ b/src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
@@ -355,7 +355,7 @@ class CNNLayerCreator : public ::ngraph::AttributeVisitor {
 public:
     explicit CNNLayerCreator();
 
-    CNNLayerPtr create(const std::shared_ptr<::ngraph::Node>& node_);
+    CNNLayerPtr create(const std::shared_ptr<::ngraph::Node>& origin);
 
 protected:
     using CreatorFor = std::function<CNNLayerPtr(const std::shared_ptr<::ngraph::Node>& node,
@@ -1970,20 +1970,23 @@ CNNLayerCreator::CNNLayerCreator() {
                        });
 }
 
-CNNLayerPtr CNNLayerCreator::create(const std::shared_ptr<::ngraph::Node>& node_) {
-    node = node_;  // node used by node->visit_attributes(..) > AttributeVisitor::on_attribute(..)
+CNNLayerPtr CNNLayerCreator::create(const std::shared_ptr<::ngraph::Node>& origin) {
+    node = origin;  // node used by node->visit_attributes(..) > AttributeVisitor::on_attribute(..)
 
-    if (!node_->visit_attributes(*this))
+    if (!node->visit_attributes(*this))
         return nullptr;
 
-    LayerParams attrs = {node_->get_friendly_name(),
-                         node_->description(),
-                         details::convertPrecision(node_->get_output_element_type(0))};
+    LayerParams attrs = {node->get_friendly_name(),
+                         node->description(),
+                         details::convertPrecision(node->get_output_element_type(0))};
 
     CNNLayerPtr res;
 
-    if (creators.find(node_->description()) != creators.end())
-        res = creators[node_->description()](node_, params);
+    auto creator = creators.find(node->description());
+    ;
+
+    if (creator != creators.end())
+        res = creator->second(node, params);
     else {
         res = std::make_shared<CNNLayer>(attrs);
         res->params = params;
--- a/src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
+++ b/src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
@@ -353,11 +353,13 @@ CNNLayer::Ptr createSubGraphLayer(const std::shared_ptr<ngraph::Node>& layer) {
  */
 class CNNLayerCreator : public ::ngraph::AttributeVisitor {
 public:
+    explicit CNNLayerCreator();
+
+    CNNLayerPtr create(const std::shared_ptr<::ngraph::Node>& node_);
+
+protected:
     using CreatorFor = std::function<CNNLayerPtr(const std::shared_ptr<::ngraph::Node>& node,
                                                  const std::map<std::string, std::string>& param)>;
-    explicit CNNLayerCreator(const std::shared_ptr<::ngraph::Node>& node);
-
-    CNNLayerPtr create();
 
     void on_adapter(const std::string& name, ::ngraph::ValueAccessor<bool>& value) override {
         params[name] = value.get() ? "true" : "false";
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
```diff
@@ -481,7 +483,7 @@ void CNNLayerCreator::on_adapter(const std::string& name, ::ngraph::ValueAccesso
     }
 }
 
-CNNLayerCreator::CNNLayerCreator(const std::shared_ptr<::ngraph::Node>& node) : node(node) {
+CNNLayerCreator::CNNLayerCreator() {
     addSpecificCreator({"Parameter"},
```
marcinkusm, line 487: I think this addSpecificCreator invoked multiple times in constructor of CNNLayerCreator.
As far as I remember this is invoked few times during LoadNetwork.
addSpecificCreator add data to the map which is filed of CNNLayerCreator. 
Just wonder if this could be improved by using the same creator instance for all conversion or using this field( creators) as static member.

mkwapul: >As far as I remember this is invoked few times during LoadNetwork.
thanks to the pull request, it's no longer

>Just wonder if this could be improved by using the same creator instance for all conversion
it is for now

mkwapul: for real world model, (having multiple layers inside), we are going from about 3200 calls to CNNLayerCreator::CNNLayerCreator() down to 1.

src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
```diff
@@ -1968,15 +1970,26 @@ CNNLayerCreator::CNNLayerCreator(const std::shared_ptr<::ngraph::Node>& node) :
                        });
 }
 
-CNNLayerPtr CNNLayerCreator::create() {
-    LayerParams attrs = {node->get_friendly_name(),
-                         node->description(),
-                         details::convertPrecision(node->get_output_element_type(0))};
-    if (creators.find(node->description()) != creators.end())
-        return creators[node->description()](node, params);
+CNNLayerPtr CNNLayerCreator::create(const std::shared_ptr<::ngraph::Node>& node_) {
```
marcinkusm, line 1973: rename node_ to e.g. node_to_visit or remove node.

src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
```diff
@@ -1968,15 +1970,26 @@ CNNLayerCreator::CNNLayerCreator(const std::shared_ptr<::ngraph::Node>& node) :
...
+    node = node_;  // node used by node->visit_attributes(..) > AttributeVisitor::on_attribute(..)
```
marcinkusm, line 1974: if you pass and use node_ everywhere field `node` of creator seems to be redundant and could be removed.
Thanks that you could make probably make `create` const due the fact no field would be modified.

src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
```diff
@@ -1968,15 +1970,26 @@ CNNLayerCreator::CNNLayerCreator(const std::shared_ptr<::ngraph::Node>& node) :
...
+    if (!node_->visit_attributes(*this))
+        return nullptr;
+
+    LayerParams attrs = {node_->get_friendly_name(),
+                         node_->description(),
+                         details::convertPrecision(node_->get_output_element_type(0))};
+
+    CNNLayerPtr res;
+
+    if (creators.find(node_->description()) != creators.end())
```
marcinkusm, line 1985: I guess one search loop instead of two would be faster if we want to improve performance.

```suggestion
    auto creator_iterator = creators.find(node_->description());
    if ( creator_iterator != creators.end()){
        res = *creator_iterator(node_, params);
    }
```

marcinkusm (changes requested):
Nice work.

It looks like field node is no longer needed.

src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
mkwapul: >if you pass and use node_ everywhere field node of creator seems to be redundant and could be removed.

it is needed by callbacks. that's why comment added to the code

src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
mkwapul: indeed. changed.

mkwapul: thank you for all comments

src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
```diff
@@ -1968,15 +1970,29 @@ CNNLayerCreator::CNNLayerCreator(const std::shared_ptr<::ngraph::Node>& node) :
                        });
 }
 
-CNNLayerPtr CNNLayerCreator::create() {
+CNNLayerPtr CNNLayerCreator::create(const std::shared_ptr<::ngraph::Node>& origin) {
+    node = origin;  // node used by node->visit_attributes(..) > AttributeVisitor::on_attribute(..)
```
marcinkusm, line 1974: Why cannot we remove node field of the class and operate only on origin?

mkwapul: used by CNNLayerCreator::on_adapter(..), which in turn is used by visit_attributes(..) > on_attribute(...) generic mechanism.

src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
```diff
@@ -353,11 +353,13 @@ CNNLayer::Ptr createSubGraphLayer(const std::shared_ptr<ngraph::Node>& layer) {
  */
 class CNNLayerCreator : public ::ngraph::AttributeVisitor {
 public:
+    explicit CNNLayerCreator();
+
+    CNNLayerPtr create(const std::shared_ptr<::ngraph::Node>& origin);
```
marcinkusm, line 358: origin -> current_node?

marcinkusm (approved):

src/plugins/intel_gna/legacy/src/convert_function_to_cnn_network.cpp
mkwapul: I don't see previous or next here, so current looks to be redundant info here. having 'node' in field name while you do have access to type in function definition (I mean it's not "auto origin") also seemed to be a bit redundant to me. Additionally when you look at call stack, You'll see following:

        CNNLayerPtr cnnLayer = createCNNLayer(layer);

that's why I finally decided that 'origin' is enough.

tadamowicz (approved):

mryzhov (approved):

PR #17688, opened 2023-05-23
TFLite layer tests second part

rnugmanx:


tests/layer_tests/common/layer_utils.py
```diff
@@ -119,6 +119,9 @@ def get_inputs_info(self, precision) -> dict:
         net = core.read_model(self.model, self.weights)
         inputs_info = {}
         for item in net.inputs:
-            inputs_info[item.get_any_name()] = list(item.shape)
+            if item.partial_shape.is_dynamic:
+                inputs_info[item.get_any_name()] = item.partial_shape
+            else:
+                inputs_info[item.get_any_name()] = item.partial_shape.to_shape()
```
rkazants, line 125: Looks your changes affect TF FE layer tests for `GatherND`. Please fix

rkazants (changes requested):

tests/layer_tests/common/layer_utils.py
rnugmanx: Done

tests/layer_tests/tensorflow_lite_tests/test_tfl_Unary.py
```diff
@@ -41,9 +41,6 @@
 
     # This op could not be converted standalone -- tries to become FlexOp (offload from tfl to tf)
     # {'op_name': 'SIGN', 'op_func': tf.math.sign},
-
-    # TF has no such standalone operation
-    # {'op_name': 'HARD_SWISH'}
```
jane-intel, line 46: you may also delete Shape and try to test Rank (It may only appear in the dynamic rank models which is not supported by OV)

jane-intel (approved):

PR #17695, opened 2023-05-23
MultiSubGraphOp ops optimization for common Constants

mbencer:
Ticket: 109208

The current state:
We can have a big Constant in the main `ov::Model` scope which is consumed inside `else` and `then` subgraphs of If. In the current state we have two copies of such `Costant` (inside subgraphs) in IR path.
The proposed solution: skip `PushConstantToSubgraph` for such a case, changing ONNX FE to use one `set_input` call for single `Constant` used in both branches.
Option 1 to consider: Changing deserialization (+ probably serialization) to avoid duplication of Constants allocation from the same memory of *.bin
Option 2 to consider: Marking such common Constants with some `RtInfo` during `set_input`  (not sure how to deal with subgraphs that can be const-folded in such a case)

### DIFF
--- a/src/common/transformations/src/transformations/common_optimizations/push_constant_to_subgraph.cpp
+++ b/src/common/transformations/src/transformations/common_optimizations/push_constant_to_subgraph.cpp
@@ -11,36 +11,47 @@
 
 using MultiSubGraphOp = ov::op::util::MultiSubGraphOp;
 
-static std::shared_ptr<ov::op::v0::Constant> try_constantfold_input(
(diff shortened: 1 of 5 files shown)


itikhono:
we discussed this PR with Mateusz B. and decided to implement this approach for PushConstantToSubgraph transformation:
image
We don't need to copy OuterConstant at all if, this is a represented as ov::op::Constant, we just need to directly connect it to the body nodes and delete Input, InputDescription and body Parameter.
If OuterConstant is represented as const-folded subgraph, we need to run "const fold" function only once to create ov::op::Constant and apply steps above.

PR #17701, opened 2023-05-24
[PT FE]: support frobenius norm and fix aten::norm

eaidova:
### Details:
 - *support aten::frobenius_norm*

### Tickets:
 - *101102*
tested on clip model

### DIFF
--- a/tests/layer_tests/pytorch_tests/test_norm.py
+++ b/tests/layer_tests/pytorch_tests/test_norm.py
@@ -8,9 +8,6 @@ import torch
 from pytorch_layer_test_class import PytorchLayerTest
 
 
-@pytest.mark.parametrize('p', [-2, -1, 0, 1, 2, 2.5, float('inf'), float('-inf')])
-@pytest.mark.parametrize('dim', [[0], [0, 1], [0, 1, 2]])
-@pytest.mark.parametrize('keepdim', [True, False])
 class TestNorm(PytorchLayerTest):
 
     def _prepare_input(self):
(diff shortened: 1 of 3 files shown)


tests/layer_tests/pytorch_tests/test_norm.py
```diff
@@ -32,12 +29,87 @@ def forward(self, input_data):
...
+    def test_norm_tensor(self, ie_device, precision, ir_version, p, dim, keepdim):
+        self._test(*self.create_model_tensor_norm(p, dim, keepdim),
+                   ie_device, precision, ir_version)
+
+
+class TestFrobeniusNorm(PytorchLayerTest):
+    def _prepare_input(self, out=False):
+        if not out:
+            return (np.random.randn(1, 2, 3).astype(np.float32),)
+        x = np.random.randn(1, 2, 3).astype(np.float32)
```
mvafin, line 79: Tensor is too small to verify the correctness of norm calculation, could you increase it to at least [10,12,14].

tests/layer_tests/pytorch_tests/test_norm.py
```diff
@@ -32,12 +29,87 @@ def forward(self, input_data):
...
+        ref_net = None
+
+        return aten_frobenius_norm(dim, keepdim, out), ref_net, "aten::frobenius_norm"
+
+    @pytest.mark.nightly
+    @pytest.mark.precommit
+    @pytest.mark.parametrize('dim', [(1, ), (0, ), (-1, ), (0, 1), (1, 0)])
+    @pytest.mark.parametrize('keepdim', [True, False])
+    @pytest.mark.parametrize("out", [False, True])
+    def test_frobenius_norm(self, ie_device, precision, ir_version, dim, keepdim, out):
```
mvafin, line 108: What if the input to this operation is integer type? Does pytorch support such case?

mvafin (changes requested):
In general looks good, please fix code style and small comments

tests/layer_tests/pytorch_tests/test_norm.py
eaidova: PyTorch does not support such case:
RuntimeError: forbenius_norm(): input dtype should be either floating point or complex. Got Int instead

tests/layer_tests/pytorch_tests/test_norm.py
eaidova: updated

mvafin (approved):

PR #17702, opened 2023-05-24
[CPU] Graph::DropNode fix

v-Golubev:
### Tickets:
 - *CVS-111649*


dmitry-gorokhov:
@nshchego please take a look

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1414,7 +1414,7 @@ void Graph::DropNode(const NodePtr &node) {
             if (!child)
                 continue;
 
-            EdgePtr &remEdge = p_edge;
+            EdgePtr remEdge = p_edge;
```
nshchego, line 1417: The code L1417-L1423 is not needed here from now and can be moved to the line L1409. Actually just `inNum` is required in this loop.
The same is applicable for fn `DropDWConvNode`

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1414,7 +1414,7 @@ void Graph::DropNode(const NodePtr &node) {
             if (!child)
```
nshchego, line 1414: May I ask you to fix code on the lines L1411-L1413 as well?
weak_ptr must be locked and validated just once, like on L1405, otherwise it can be invalidated in other thread.
The same is applicable for fn `DropDWConvNode`.

src/plugins/intel_cpu/src/graph.cpp
v-Golubev: Good point, thanks. Fixed

src/plugins/intel_cpu/src/graph.cpp
v-Golubev: Corrected, please double check. Thanks

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1407,21 +1407,21 @@ void Graph::DropNode(const NodePtr &node) {
         auto parent = p_edge->getParent();
         if (!parent) continue;
 
+        EdgePtr remEdge = p_edge;
+        int inNum = 0;
+        if (remEdge) {
```
nshchego, line 1412: Minor. Variable `remEdge` and this check are not needed as well. `p_edge` already validated on line L1406. The same for `c_edge`.

nshchego (approved):

src/plugins/intel_cpu/src/graph.cpp
v-Golubev: Yes indeed. `remEdge` and unnecessary check were removed, thank you.

PR #17705, opened 2023-05-24
[PT FE]: support flip operation

eaidova:
### Details:
 - *supported aten::flip operation*

### Tickets:
 - *110802*

### DIFF
--- a/tests/layer_tests/pytorch_tests/test_flip.py
+++ b/tests/layer_tests/pytorch_tests/test_flip.py
@@ -1,0 +1,42 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+from pytorch_layer_test_class import PytorchLayerTest
+
+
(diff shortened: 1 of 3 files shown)


tests/layer_tests/pytorch_tests/test_flip.py
```diff
@@ -0,0 +1,42 @@
...
+
+import pytest
+
+from pytorch_layer_test_class import PytorchLayerTest
+
+
+class TestFlip(PytorchLayerTest):
+    def _prepare_input(self, out=False):
+        import numpy as np
+        x = np.random.randn(2, 3, 4, 5).astype(np.float32)
```
mvafin, line 12: Please also add integer tests

mvafin (changes requested):

tests/layer_tests/pytorch_tests/test_flip.py
eaidova: added

mvafin (approved):

PR #17707, opened 2023-05-24
[Benchmark_app] - Update comments and help text

DariaMityagina:
The issue is for the benchmark_app.exe, the -ip and -op params used to take the value “FP16”.  Now you must enter “F16”.  According to the help text, it still says ”FP16”

image

PR #17710, opened 2023-05-24
[Benchmark_app] - Update comments and help text

DariaMityagina:
The issue is for the benchmark_app.exe, the -ip and -op params used to take the value “FP16”.  Now you must enter “F16”.  According to the help text, it still says ”FP16”

image

Copy of: https://github.com/openvinotoolkit/openvino/pull/17707


Wovchena:
You should probably apply F->f here as well

DariaMityagina:
> You should probably apply F->f here as well

Done, thanks!

ilya-lavrenov (approved):

PR #17714, opened 2023-05-25
nGraph PassTracker POC

GlebKazantaev:
## Description

This PR adds pass profiling which can be visualised by PassViz.
For more details see: https://github.com/GlebKazantaev/pass-viz

### DIFF
--- a/src/core/src/pass/graph_rewrite.cpp
+++ b/src/core/src/pass/graph_rewrite.cpp
@@ -18,6 +18,7 @@
 #include "ngraph/log.hpp"
 #include "ngraph/op/util/sub_graph_base.hpp"
 #include "perf_counters.hpp"
+#include "openvino/pass/pass_tracker.hpp"
 
 /* GraphRewrite algorithm:
  * GraphRewrite processes an input graph in an topological order(i.e. args before users)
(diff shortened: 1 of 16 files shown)


src/core/src/pass/graph_rewrite.cpp
```diff
@@ -151,7 +152,16 @@ bool ov::pass::GraphRewrite::apply_matcher_passes(std::shared_ptr<Model> f,
...
         // to this node
-        bool status = m_pass->apply(node);
+
+        bool status = false;
+        {
+            if (m_pass->skip_profiling()) {
+                status = m_pass->apply(node.lock());
+            } else {
+                PassTracker tracker(f, m_pass->get_name());
+                status = m_pass->apply(node.lock());
```
slyalin, line 162: I would suggest to move real profiling code under some preprocessor macro to be able to cut such code completely off to avoid any binary size increase doubts. It should be controlled by a new pre-processing frag. I think this particular place in the code is a starting point where we should apply this conditional compilation and include all the code solely dedicated to such profiling.

I think that a customer of this service is a transformation developer who really needs to rebuild openvino, so providing a cmake flag and rebuild is not a big deal.

sysicvvpux:
Can one of the admins verify this patch?

PR #17718, opened 2023-05-25
[PT FE]: support aten::cdist and aten::pairwise_distance

eaidova:
### Details:
 - *support translation aten::cdist operator*
 - *support translation aten::pairwise_distance*

### Tickets:
 - *111693*


mvafin (approved):
I think `p` and `eps` can be dynamic, but in general looks good

PR #17724, opened 2023-05-25
[CPU] RNN: Shape checks are relaxed

v-Golubev:
### Details:
 - *At the node creation stage, It is not guaranteed that first input shape is equal to the first output shape: in case of some complicated shapeOf subgraphs first input shape can be dynamic whereas output shape is deduced basing on other cell input shapes. So these checks are valid only in case of static shapes*

### Tickets:
 - *CVS-111649*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/rnn.cpp
+++ b/src/plugins/intel_cpu/src/nodes/rnn.cpp
@@ -479,26 +479,29 @@ void RNN::initCell() {
     else
         DC = getInputShapeAtPort(2).getDims()[1];
 
-    // Expected shapes.
-    const Shape shapeD{{N.minVal, DC}, {N.maxVal, DC}}, shapeS{{N.minVal, SC}, {N.maxVal, SC}};
+    if (N.isStatic()) {
+        // Expected shapes.
+        const Shape shapeD{{N.minVal, DC}, {N.maxVal, DC}}, shapeS{{N.minVal, SC}, {N.maxVal, SC}};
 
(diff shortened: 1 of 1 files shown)


dmitry-gorokhov:
@EgorDuplensky could you please take a look?

src/plugins/intel_cpu/src/nodes/rnn.cpp
```diff
@@ -479,26 +479,29 @@ void RNN::initCell() {
...
-
-    if ((getInputShapeAtPort(0).isStatic() && getInputShapeAtPort(0) != shapeD) ||
-            (getInputShapeAtPort(1).isStatic() && getInputShapeAtPort(1) != shapeS) ||
-            (getOutputShapeAtPort(0) != shapeS)) {
-        THROW_ERROR << "has incorrect input/output shapes. Data shape: " << getInputShapeAtPort(0).toString() <<
-                "; Hidden state input: " << getInputShapeAtPort(1).toString() << "; Hidden state output: " << getOutputShapeAtPort(0).toString();
-    }
+    if (N.isStatic()) {
+        // Expected shapes.
+        const Shape shapeD{{N.minVal, DC}, {N.maxVal, DC}}, shapeS{{N.minVal, SC}, {N.maxVal, SC}};
```
EgorDuplensky, line 484: [Mi]
The code within the **if (N.isStatic())** brackets still uses N.minVal N.maxVal values even it actually does not matter.
Can we use like
auto B = N.minVal
To avoid any confusion

EgorDuplensky (approved):

PR #17725, opened 2023-05-25
[HETERO]: Avoid duplicated results

nkogteva:
### Details:
 - *HETERO: avoid duplicated results on the same output port, it will give performance improvement on such devices as NVIDIA GPU*

### Tickets:
 - *111877*


vladimir-paramuzov (approved):
Overall LGTM. Please add a test if possible.

PR #17729, opened 2023-05-25
Calculate TensorView strides on demand not in ctor

praasz:
### Details:
 - Improve the strides calculation performance by reduce number of loops (2 ->1). 
 - Move strides calculation from `TensorView` constructor to `get_strides` and `set_shape` member functions. Strides are calculated only once until (on demand) or when new shape is set. This change improve speed of tensor creation from `TensorView` as this type  of tensor is wrapper for existing memory and use default strides which not always is used.

### Tickets:
 - 111776


jane-intel:
Can we provide unit-test level testing for it?

praasz:
> Can we provide unit-test level testing for it?

The `get_strides` is already tested  by `OVTensorTest`, and is also used in other test e.g. CPU functional test.
This change is transparent from usage of tensor API if `get_strides()` provide wrong value there will be some failure test.

Regarding internal stuff like check that `update_strides` is called once is hard to test as `ViewTensor` is not exported and cannot be mocked by gtest to check number of calls on this member function.

ilyachur (approved):

PR #17732, opened 2023-05-25
[CPU] Skip PriorBox-8 tests.

nshchego:
### Details:
 - *Shared tests for PriorBox are based on the opset8 version which is not supported by CPU plugin now. Actually they are executed on the ref implementation.*

### Tickets:
 - *98677*

### DIFF
--- a/src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
+++ b/src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
@@ -120,6 +120,8 @@ std::vector<std::string> disabledTestPatterns() {
         R"(.*smoke_FakeQuantizeLayerCPUTest.*bin.*)",
         // Issue: 69222
         R"(.*smoke_PriorBoxClustered.*PriorBoxClusteredLayerCPUTest.*_netPRC=f16_.*)",
+        // Skip the PriorBox from opset8 due to this is not supported by plugin and executed on reference.
+        R"(.*smoke_PriorBox_Basic/PriorBoxLayerTest.*)",
         // Issue: 72005
         // there are some inconsistency between cpu plugin and ng ref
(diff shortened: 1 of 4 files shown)


src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
```diff
@@ -120,6 +120,8 @@ std::vector<std::string> disabledTestPatterns() {
         R"(.*smoke_FakeQuantizeLayerCPUTest.*bin.*)",
         // Issue: 69222
         R"(.*smoke_PriorBoxClustered.*PriorBoxClusteredLayerCPUTest.*_netPRC=f16_.*)",
+        // Skip the PriorBox from opset8 due to this is not supported by plugin and executed on reference.
+        R"(.*smoke_PriorBox_Basic/PriorBoxLayerTest.*)",
```
dmitry-gorokhov, line 124: Does it really save any sufficient amount of time in bounds of func test run? If no I would still prefer to keep the test given it covers fallback on reference impl path.

nshchego: They last about 1 second.

dmitry-gorokhov: Doesn't looks like significant value. I still would prefer to test fallback on referenec codepath. 
We can descrease number of instances if required.

dmitry-gorokhov: Doesn't looks like significant value. I still would prefer to test fallback on reference codepath. 
We can descrease number of instances if required.

dmitry-gorokhov:
Closing as not relevant. Feel free to reopen

PR #17774, opened 2023-05-29
[Snippets] Fixed copy runtime info which contains PortDescriptors

a-sidorova:
### Details:
 - *Since `PortDescriptors` are contained in runtime info, need to be careful to call `copy_runtime_info`. Because the new node may have incorrect `PortDescriptors` after copying. Because of that `PortDescriptorAttribute` is not copyable now*

### Tickets:
 - *111954*

### DIFF
--- a/src/common/snippets/include/snippets/utils.hpp
+++ b/src/common/snippets/include/snippets/utils.hpp
@@ -29,7 +29,9 @@ ov::PartialShape get_port_planar_shape(const Output<Node>& out);
 ov::PartialShape get_reordered_planar_shape(const ov::PartialShape& shape, const std::vector<size_t>& layout);
 
 // Copy runtime info using default ngraph method but delete PortDescriptors which may be transferred after copying
-void safe_copy_runtime_info(const std::shared_ptr<ov::Node>&, const std::shared_ptr<ov::Node>& to);
+void safe_copy_runtime_info(const std::shared_ptr<ov::Node>& from, const std::shared_ptr<ov::Node>& to);
+void safe_copy_runtime_info(const ov::NodeVector& from, const std::shared_ptr<ov::Node>& to);
+void safe_copy_runtime_info(const std::shared_ptr<ov::Node>& from, const ov::NodeVector& to);
 
 inline auto normalize_rank(int32_t allocation_rank, const size_t shape_rank) -> int32_t {
     return allocation_rank < 0 ? allocation_rank + static_cast<int32_t>(shape_rank) + 1 : allocation_rank;
--- a/src/common/snippets/include/snippets/utils.hpp
+++ b/src/common/snippets/include/snippets/utils.hpp
@@ -28,11 +28,6 @@ ov::PartialShape get_port_planar_shape(const Input<Node>& out);
 ov::PartialShape get_port_planar_shape(const Output<Node>& out);
 ov::PartialShape get_reordered_planar_shape(const ov::PartialShape& shape, const std::vector<size_t>& layout);
 
-// Copy runtime info using default ngraph method but delete PortDescriptors which may be transferred after copying
-void safe_copy_runtime_info(const std::shared_ptr<ov::Node>& from, const std::shared_ptr<ov::Node>& to);
-void safe_copy_runtime_info(const ov::NodeVector& from, const std::shared_ptr<ov::Node>& to);
-void safe_copy_runtime_info(const std::shared_ptr<ov::Node>& from, const ov::NodeVector& to);
-
 inline auto normalize_rank(int32_t allocation_rank, const size_t shape_rank) -> int32_t {
     return allocation_rank < 0 ? allocation_rank + static_cast<int32_t>(shape_rank) + 1 : allocation_rank;
 }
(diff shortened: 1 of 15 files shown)


src/common/snippets/include/snippets/utils.hpp
```diff
@@ -29,7 +29,9 @@ ov::PartialShape get_port_planar_shape(const Output<Node>& out);
 ov::PartialShape get_reordered_planar_shape(const ov::PartialShape& shape, const std::vector<size_t>& layout);
 
 // Copy runtime info using default ngraph method but delete PortDescriptors which may be transferred after copying
-void safe_copy_runtime_info(const std::shared_ptr<ov::Node>&, const std::shared_ptr<ov::Node>& to);
+void safe_copy_runtime_info(const std::shared_ptr<ov::Node>& from, const std::shared_ptr<ov::Node>& to);
```
IvanNovoselov, line 32: Do I get it right that `safe` means `disregard PortDescriptors`?
So the idea here is here is that we copy everything except for `PortDescriptors`?

a-sidorova: Yeah, you're right. It was previous idea

src/common/snippets/include/snippets/utils.hpp
```diff
@@ -29,7 +29,9 @@ ov::PartialShape get_port_planar_shape(const Output<Node>& out);
...
+void safe_copy_runtime_info(const ov::NodeVector& from, const std::shared_ptr<ov::Node>& to);
+void safe_copy_runtime_info(const std::shared_ptr<ov::Node>& from, const ov::NodeVector& to);
```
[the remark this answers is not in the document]
a-sidorova: Honestly, I don't like the name `copy_rt_info_except_port_descriptors` (my eyes... :smile:)

So I'd suggest to make `PortDescriptorAttribute` as not copyable. I believe, it's safer. Look:
The copy of attributes ihappens only in `copy_runtime_info` as I understand. We call this method only in transformations when we want to move attributes from old nodes to new nodes. It's invalid for `PortDescriptorAttribute` in general (in more cases). If you want to build the same new node, you should manually move the `PortDescriptorAttribute`.
Also `clone_model` doesn't use `copy_runtime_info`, it explicitly copies the class field using `new->get_rt_info() = from->get_rt_info();`. In this case we safely copy `PortDescriptorAttribute`.

I added the note in class description that we should manually copy `PortDescriptorAttribute` if needed.

src/common/snippets/src/utils.cpp
```diff
@@ -101,6 +101,16 @@ void safe_copy_runtime_info(const std::shared_ptr<ov::Node>& from, const std::sh
     ov::copy_runtime_info(from, to);
     lowered::PortDescriptorUtils::clean(to);
```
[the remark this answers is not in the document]
a-sidorova: You're totally right! Good catch!
I decided to remove these unsafe methods. The reason is in the previous comment, Thanks a lot

src/plugins/intel_cpu/thirdparty/onednn
IvanNovoselov, line 1: Do we really need to change onednn sources in this PR?
Looks like an unintended modification.

src/plugins/intel_cpu/src/transformations/snippets/x64/pass/mul_add_to_fma.cpp
```diff
@@ -2,10 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-#include <snippets/itt.hpp>
-
 #include "mul_add_to_fma.hpp"
-#include "snippets/snippets_isa.hpp"
+
+#include "snippets/itt.hpp"
+#include "snippets/utils.hpp"
```
IvanNovoselov, line 8: Same: not sure we need this header here after fallback to `copy_runtime_info`

IvanNovoselov (approved):

src/plugins/intel_cpu/thirdparty/onednn
a-sidorova: Really good catch! Thank you very much! I didn't notice it after rebasing. Fixed

src/plugins/intel_cpu/src/transformations/snippets/x64/pass/mul_add_to_fma.cpp
a-sidorova: It's outdated change. Thanks

src/plugins/intel_cpu/src/transformations/snippets/x64/pass/enforce_precision.cpp
```diff
@@ -4,12 +4,12 @@
 
 #include "enforce_precision.hpp"
 
-#include <memory>
-
-#include "ov_ops/type_relaxed.hpp"
 #include "snippets/itt.hpp"
-#include "openvino/core/rt_info.hpp"
+#include "snippets/utils.hpp"
```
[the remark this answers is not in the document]
a-sidorova: Yeap, thanks

a-sidorova:
@IvanNovoselov thank you very much for the review!

@dmitry-gorokhov take a look too please

dmitry-gorokhov (approved):

dmitry-gorokhov (approved):

PR #17781, opened 2023-05-29
GroupNormalization core op

tomdol:
### Details:
 - Introduction of opset12
 - GroupNormalization core operator and necessary tests

### Tickets:
 - 101517
 - 101522


src/core/src/op/group_normalization.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+bool op::v12::GroupNormalization::visit_attributes(AttributeVisitor& visitor) {
+    OV_OP_SCOPE(v12_GroupNormalization_visit_attributes);
+    visitor.on_attribute("num_groups", m_num_groups);
+    visitor.on_attribute("epsilon", m_epsilon);
+    return true;
+}
+
+void op::v12::GroupNormalization::validate_and_infer_types() {
```
mitruska, line 31: Consider moving the shape validation to the common shape_infer function in [shape_inference](https://github.com/openvinotoolkit/openvino/tree/master/src/core/shape_inference/include).

src/core/src/op/group_normalization.cpp
```diff
@@ -0,0 +1,66 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "openvino/op/group_normalization.hpp"
+
+#include "itt.hpp"
+#include "ngraph/validation_util.hpp"
```
mitruska, line 8: Looks like this header is redundant (and all oft the functions there are deprecated).

src/core/src/op/group_normalization.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    NODE_VALIDATION_CHECK(this,
+                          scale_partial_shape.rank().compatible(Dimension{1}),
+                          "The scale input is required to be 1D");
+    NODE_VALIDATION_CHECK(this,
+                          bias_partial_shape.rank().compatible(Dimension{1}),
+                          "The bias input is required to be 1D");
+
+    if (data_rank.is_static() && data_rank.get_length() >= 2) {
+        const auto channels_dim = data_partial_shape[1];
```
mitruska, line 46: - The value of `num_groups` is not validated, but according to the spec it can't be lower than `1`.
- Also (not described by spec), is the result of static `channels_dim % num_groups` expected to be zero 
or rounded down like `channels_dim // num_groups`? 
- What if the `num_groups` is bigger than `channels_dim`?

src/core/src/op/group_normalization.cpp
praasz: Just for information what should be check regarding shape inference.
1. Dimension and labels propagation (do check on dynamic rank input, static rank and interval dimension with labels)
2. Add template implementation for shape inference in future it can be re used with CPU or GPU plugin and check if possible to call `shape_infer` function where operator created with default ctor (best option no set inputs to simylate GPU usage)
These point will not apply to this op as it just copy input shape to output.
3. If output shape depends on input values then there should be check if this input (e.g as result of ShapeOf) which can have interval values and labels are propageted correctly to output.
4. If operator is used in SsapeOf sub-graph then it should have evaluate upper/lower and labels to propagate the values and labels.

src/core/src/op/group_normalization.cpp
tomdol: I've done some research and some experiments and it seems that we should assert that the channels dimension can be evenly divided by the number of groups. Additionally we should limit the upper bound value of `num_groups` to the value of `channels_dim`.

src/core/src/op/group_normalization.cpp
```diff
@@ -0,0 +1,66 @@
...
+    const auto bias_partial_shape = get_input_partial_shape(2);
+
+    NODE_VALIDATION_CHECK(this,
+                          scale_partial_shape.rank().compatible(Dimension{1}),
+                          "The scale input is required to be 1D");
+    NODE_VALIDATION_CHECK(this,
+                          bias_partial_shape.rank().compatible(Dimension{1}),
+                          "The bias input is required to be 1D");
+
+    if (data_rank.is_static() && data_rank.get_length() >= 2) {
```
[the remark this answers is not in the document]
tomdol: In fact GN doesn't really make sense for tensors ranked lower than 2D. Torch fails for 1D and ONNXRT expects 4D tensors for this op.

src/core/src/op/group_normalization.cpp
mitruska: I confirm, also points 3 and 4 seem to be not applicable for GroupNormalization.

src/core/tests/type_prop/group_normalization.cpp
```diff
@@ -0,0 +1,193 @@
...
+    const auto scale = std::make_shared<opset12::Parameter>(element::f16, PartialShape::dynamic());
+    const auto bias = std::make_shared<opset12::Parameter>(element::f16, PartialShape::dynamic());
+
+    const auto gn = std::make_shared<opset12::GroupNormalization>(data, scale, bias, 12, 0.00001f);
+    EXPECT_EQ(gn->get_element_type(), element::f16);
+    EXPECT_EQ(gn->get_output_partial_shape(0), (PartialShape::dynamic()));
+}
+
+TEST(type_prop, group_normalization_dynamic_intervals) {
+    const auto data = std::make_shared<opset12::Parameter>(element::f32, PartialShape{2, Dimension{10, 20}, 6, 6});
```
praasz, line 96: Maybe add here labels to all shapes on input and check that only labels from  data inputs are propagated.

src/core/tests/type_prop/group_normalization.cpp
```diff
@@ -0,0 +1,193 @@
...
+    const auto scale = std::make_shared<opset12::Parameter>(element::f32, Shape{12});
+    const auto bias = std::make_shared<opset12::Parameter>(element::f32, Shape{12});
+
+    const auto gn = std::make_shared<opset12::GroupNormalization>(data, scale, bias, 4, 0.00001f);
+    EXPECT_EQ(gn->get_element_type(), element::f32);
+    EXPECT_EQ(gn->get_shape(), (Shape{1, 12, 6, 6}));
+}
+
+TEST(type_prop, group_normalization_labels) {
+    auto data_shape = PartialShape{1, 12, 6, 6};
```
praasz, line 25: Set also labels on other inputs to check that others not propagated (e.g. when some dimension are merged etc).

src/core/shape_inference/include/group_normalization_shape_inference.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include <openvino/core/validation_util.hpp>
+#include <openvino/op/group_normalization.hpp>
+
+namespace ov {
+namespace op {
+namespace v12 {
+template <class TShape>
+std::vector<TShape> shape_infer(const GroupNormalization* op,
+                                const std::vector<TShape>& input_shapes,
+                                const std::map<size_t, HostTensorPtr>& constant_data = {}) {
```
praasz, line 15: Remove this parameter is not used at all to avoid empty map creation on each call.

src/core/shape_inference/include/group_normalization_shape_inference.hpp
```diff
@@ -0,0 +1,73 @@
...
+    const auto& inputs_count = input_shapes.size();
+    NODE_VALIDATION_CHECK(op, (inputs_count == 3));
+    TShape output_shape;
+
+    const auto data_partial_shape = input_shapes[0];
+    const auto data_rank = data_partial_shape.rank();
+    const auto scale_partial_shape = input_shapes[1];
+    const auto bias_partial_shape = input_shapes[2];
```
praasz, line 23:
```suggestion
    const auto& data_partial_shape = input_shapes[0];
    const auto data_rank = data_partial_shape.rank();
    const auto& scale_partial_shape = input_shapes[1];
    const auto& bias_partial_shape = input_shapes[2];
```
to avoid copies

src/core/shape_inference/include/group_normalization_shape_inference.hpp
```diff
@@ -0,0 +1,73 @@
...
+namespace v12 {
+template <class TShape>
+std::vector<TShape> shape_infer(const GroupNormalization* op,
+                                const std::vector<TShape>& input_shapes,
+                                const std::map<size_t, HostTensorPtr>& constant_data = {}) {
+    const auto& inputs_count = input_shapes.size();
+    NODE_VALIDATION_CHECK(op, (inputs_count == 3));
+    TShape output_shape;
+
+    const auto data_partial_shape = input_shapes[0];
```
praasz, line 20: Consider rename ` data_partial_shape` -> `data_shape` as shapes are template type can be different than `PartialShape`

src/core/shape_inference/include/group_normalization_shape_inference.hpp
```diff
@@ -0,0 +1,73 @@
...
+namespace ov {
+namespace op {
+namespace v12 {
+template <class TShape>
+std::vector<TShape> shape_infer(const GroupNormalization* op,
+                                const std::vector<TShape>& input_shapes,
+                                const std::map<size_t, HostTensorPtr>& constant_data = {}) {
+    const auto& inputs_count = input_shapes.size();
+    NODE_VALIDATION_CHECK(op, (inputs_count == 3));
+    TShape output_shape;
```
praasz, line 18: remove it.

src/core/shape_inference/include/group_normalization_shape_inference.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+    const auto data_partial_shape = input_shapes[0];
+    const auto data_rank = data_partial_shape.rank();
+    const auto scale_partial_shape = input_shapes[1];
+    const auto bias_partial_shape = input_shapes[2];
+
+    NODE_VALIDATION_CHECK(op, op->get_num_groups() > 0, "The number of groups needs to be a positive integer value");
+
+    NODE_VALIDATION_CHECK(op,
+                          scale_partial_shape.rank().compatible(Dimension{1}),
```
praasz, line 28:
```suggestion
                          scale_partial_shape.rank().compatible(1),
```

Dimension is not use values for static shapes. Correct others.

src/core/shape_inference/include/group_normalization_shape_inference.hpp
```diff
@@ -0,0 +1,73 @@
...
+                          "The scale input is required to be 1D");
+    NODE_VALIDATION_CHECK(op,
+                          bias_partial_shape.rank().compatible(Dimension{1}),
+                          "The bias input is required to be 1D");
+
+    NODE_VALIDATION_CHECK(op,
```
praasz, line 34: Move this check inside `if` statment below and check rank length only.

src/core/shape_inference/include/group_normalization_shape_inference.hpp
```diff
@@ -0,0 +1,73 @@
...
+                          data_rank.is_dynamic() || data_rank.get_length() >= 2,
+                          "The input tensor is required to be at least 2D");
+
+    if (data_rank.is_static()) {
+        const auto channels_dim = data_partial_shape[1];
```
praasz, line 39:
```suggestion
        const auto& channels_dim = data_partial_shape[1];
```

src/core/shape_inference/include/group_normalization_shape_inference.hpp
```diff
@@ -0,0 +1,73 @@
...
+                          "The shapes of both scale and bias inputs need to match");
+
+    return {input_shapes[0]};
+}
+
+template <class TShape>
+void shape_infer(const GroupNormalization* op,
+                 const std::vector<TShape>& input_shapes,
+                 std::vector<TShape>& output_shapes,
+                 const std::map<size_t, HostTensorPtr>& constant_data = {}) {
```
praasz, line 68: Remove constant_data.

praasz (approved):

src/core/src/op/group_normalization.cpp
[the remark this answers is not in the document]
mitruska: Sure, then such restriction should be added to the specification as well.

src/core/src/op/group_normalization.cpp
mitruska: Okay, it should be reflected in the ov GroupNormalization-12 specification.

mitruska (approved):
Just to highlight, this PR introduce shape_infer for GroupNormalization without adding it to the CPU [shape_inference map](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp), but it can be done separately with corresponding tests.

PR #17796, opened 2023-05-30
[Pad-12] Introduce PadBase and Pad-12 operation 

mitruska:
### Details:
 - Introduce common PadBase and Pad-12 operation 
 - Pad-12 operation officially support negative pads_begin/end 
 (it was allowed by core Pad-1 against the spec and plugins impl)
- Reuse Pad-1 tests and make them common for both versions (type_prop, attributes, reference)
- Add op impl check test
- Update Pad-1 and Pad-12 evaluate to use ov::TensorVector
- Register Pad-12 in shape inference map for CPU (Pad-1 is already registered)

-----
- Introduction of opset12 (changes from PR https://github.com/openvinotoolkit/openvino/pull/17781)
 
### Tickets:
 - 110240, 110245
 
** Possible extension of Pad-12 with new mode is considered separately


src/tests/functional/plugin/shared/src/single_layer_tests/op_impl_check/single_op_graph.cpp
```diff
@@ -741,6 +741,15 @@ std::shared_ptr<ov::Model> generate(const std::shared_ptr<ov::op::v1::Pad> &node
...
 
+std::shared_ptr<ov::Model> generate(const std::shared_ptr<ov::op::v12::Pad> &node) {
+    const auto params = ngraph::builder::makeDynamicParams(ov::element::f32, {{6, 10}});
+    const auto pad_begin = ngraph::builder::makeConstant<int64_t>(ov::element::i64, {2}, {4, -2});
+    const auto pad_end = ngraph::builder::makeConstant<int64_t>(ov::element::i64, {2}, {5, -1});
+    const auto pad = std::make_shared<ov::op::v12::Pad>(params[0], pad_begin, pad_end, ov::op::PadMode::CONSTANT);
+    ov::ResultVector results{std::make_shared<ov::op::v0::Result>(pad)};
+    return std::make_shared<ov::Model>(results, params, "Pad-12");
+}
+
```
rkazants, line 752: May be it makes sense to check a case with negative `pad_begin` and positive `pad_end` value for the same dimension, i.e. on the one corner it is cropped and on the opposite the image is padded.

src/core/shape_inference/include/pad_shape_inference.hpp
```diff
@@ -152,5 +161,24 @@ void shape_infer(const Pad* op,
...
 
 }  // namespace v1
+
+namespace v12 {
+template <class TShape>
+std::vector<TShape> shape_infer(const Pad* op,
+                                const std::vector<TShape>& input_shapes,
+                                const std::map<size_t, HostTensorPtr>& constant_data = {}) {
+    return util::shape_infer(op, input_shapes, constant_data);
+}
```
rkazants, line 171: is this function needed? I thought only `util::shape_infer` is only needed for implementation of `PadBase`.

rkazants (approved):
Left a couple of comments

src/core/include/openvino/op/pad.hpp
```diff
@@ -47,39 +48,52 @@ class OPENVINO_API Pad : public Op {
     /// \param pad_mode The padding mode: CONSTANT, EDGE, REFLECT or SYMMETRIC.
     Pad(const Output<Node>& arg, const Output<Node>& pads_begin, const Output<Node>& pads_end, PadMode pad_mode);
 
-    /// \brief Constructs a generic padding operation.
+    std::shared_ptr<Node> clone_with_new_inputs(const OutputVector& new_args) const override;
+    bool has_evaluate() const override;
```
praasz, line 52: Can be `has_evaluate` and `evaluate` part of base class?
It looks like they are same and could be shared as they call common protected function.

src/core/src/op/pad.cpp
```diff
@@ -2,118 +2,33 @@
...
-#include "ngraph/op/constant.hpp"
-#include "ngraph/op/util/op_types.hpp"
-#include "ngraph/runtime/reference/pad.hpp"
-#include "openvino/op/util/precision_sensitive_attribute.hpp"
+#include "openvino/op/constant.hpp"
 #include "pad_shape_inference.hpp"
 
 using namespace std;
-using namespace ngraph;
+using namespace ov;
```
praasz, line 13:
```suggestion
namespace ov{
```
instead and wrap all class implementation in this namesapce

src/core/include/openvino/op/pad.hpp
mitruska: This is a special case because there is no difference in the evaluate implementation between Pad-1 and Pad-12, 
but in general evaluate is related to the specific operation not with the Base class.
Overall `evaluate` can be a part of the PadBase class, and it was in the initial commits, 
but has been moved after discussion with @tomdol.

src/core/shape_inference/include/pad_shape_inference.hpp
mitruska: Actually `std::vector<TShape> shape_infer(const Pad*,...)` was redundant, so I have removed this function. 
But `void op::v1::shape_infer(const Pad*,...)` is currently used by GPU, so at this point I would like to keep it for compatibility.

tomdol (approved):

PR #17797, opened 2023-05-30
[DOCS] Add ability to build notebooks from local files

bstankix:


ilya-lavrenov:
Hi,

Could you please provide more details how is it supposed to work?
Screenshot 2023-05-30 at 8 17 22 PM
I've tried to click on `001-hello-world` and to reaction. At the left bottom corner we can see the link refers to `notebooks` folder which is absent in downloaded documentation.

PR #17799, opened 2023-05-30
[DOCS] Added initial version of selector tool

as-suvorov:
### Details:
 - Integration is done by including static html, js, css files into docs file structure and using raw html iframe tag
 - Static files are build automatically by ci


kblaszczak-intel (approved):
From the non-technical point of view, that's great!

yatarkan (approved):
LGTM

PR #17802, opened 2023-05-31
[AUTO] Limit selected device compilation threads when accelerator device is enable in AUTO

yangwang201911:
### Details:
 - AUTO will set ov::compilation_num_threads to selected device that shared the CPU (accelerator) threads to compile model.

### Tickets:
 - 110826


src/plugins/auto/src/auto_schedule.cpp
```diff
@@ -446,6 +447,11 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
                 _loadContext[CPU].isEnabled = true;
                 _loadContext[CPU].deviceInfo = *CPUIter;
                 _loadContext[CPU].deviceInfo.config[CONFIG_KEY(PERFORMANCE_HINT)] = IE::PluginConfigParams::LATENCY;
+                // limit the compilation threads for the actual selected device
+                const int num_logic_cores = InferenceEngine::getNumberOfLogicalCPUCores();
+                auto max_num_threads = (num_logic_cores / 2) == 0 ? 1 : (num_logic_cores / 2);
+                _loadContext[ACTUALDEVICE].deviceInfo.config[ov::compilation_num_threads.name()] =
+                    std::to_string(max_num_threads);
```
peterchen-intel, line 454: Need to make sure device support this property, if not, should not pass to device and avoid the exception.

src/plugins/auto/src/auto_schedule.cpp
```diff
@@ -446,6 +447,18 @@ void AutoSchedule::init(const ScheduleContext::Ptr& sContext) {
                 _loadContext[CPU].isEnabled = true;
                 _loadContext[CPU].deviceInfo = *CPUIter;
                 _loadContext[CPU].deviceInfo.config[CONFIG_KEY(PERFORMANCE_HINT)] = IE::PluginConfigParams::LATENCY;
+                // limit the compilation threads for the actual selected device
+                const int num_logic_cores = InferenceEngine::getNumberOfLogicalCPUCores();
```
peterchen-intel, line 451: Use new API ov::get_number_of_logical_cpu_cores instead of InferenceEngine::getNumberOfLogicalCPUCores

PR #17819, opened 2023-05-31
[Snippets] Added support of MatMuls with transposed inputs

a-sidorova:
### Details:
 - *Updated pass `ExplicitTransposeMatMulInputs`. Now the pass can extract `Transpose` from transposed inputs from MatMul if there isn't another `Transpose` even*
 - *Added function that extracts unsupported `Transpose` from body*
 - *Failed onnx tests are fixed by https://github.com/openvinotoolkit/openvino/pull/17774*

### Tickets:
 - *111980*

### DIFF
--- a/src/common/snippets/src/pass/explicit_transpose_matmul_inputs.cpp
+++ b/src/common/snippets/src/pass/explicit_transpose_matmul_inputs.cpp
@@ -11,18 +11,20 @@
 #include "openvino/pass/pattern/op/wrap_type.hpp"
 #include "openvino/core/rt_info.hpp"
 
+bool ov::snippets::pass::ExplicitTransposeMatMulInputs::are_weights_scalar(const std::shared_ptr<ov::Node>& node) {
+    const auto inputs = node->inputs();
+    return std::all_of(inputs.begin() + 1, inputs.end(),
+                       [](const ov::Input<ov::Node>& in) {
+                           return in.get_partial_shape().is_static() && ov::shape_size(in.get_shape()) == 1;
+                       });
+}
 
 void ov::snippets::pass::ExplicitTransposeMatMulInputs::extract(const ov::Input<ov::Node>& input) {
     auto parent = input.get_source_output().get_node_shared_ptr();
     auto transpose = ov::as_type_ptr<ov::op::v1::Transpose>(parent);
     while (!transpose && !ov::is_type<ov::op::v0::Parameter>(parent)) {
         // We can set supported order and transposed_<a|b>=false only if ops have scalar shapes to avoid shape mismatching
-        const auto parent_count = parent->inputs().size();
-        bool are_weights_scalar = true;
-        for (size_t j = 1; j < parent_count; ++j) {
-            are_weights_scalar = are_weights_scalar && ov::shape_size(parent->get_input_shape(j)) == 1;
-        }
-        if (!are_weights_scalar)
+        if (!are_weights_scalar(parent))
             break;
 
         parent = parent->get_input_node_shared_ptr(0);
--- a/src/common/snippets/src/pass/explicit_transpose_matmul_inputs.cpp
+++ b/src/common/snippets/src/pass/explicit_transpose_matmul_inputs.cpp
@@ -2,79 +2,98 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+#include "snippets/pass/explicit_transpose_matmul_inputs.hpp"
+
+#include "snippets/op/subgraph.hpp"
 #include "snippets/itt.hpp"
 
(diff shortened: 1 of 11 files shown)


src/common/snippets/src/pass/explicit_transpose_matmul_inputs.cpp
```diff
@@ -2,79 +2,98 @@
...
+        const auto parent_count = parent->inputs().size();
+        bool are_weights_scalar = true;
+        for (size_t j = 1; j < parent_count; ++j) {
+            are_weights_scalar = are_weights_scalar && ngraph::shape_size(parent->get_input_shape(j)) == 1;
+        }
+        if (!are_weights_scalar)
+            break;
+
+        parent = parent->get_input_node_shared_ptr(0);
+        transpose = ngraph::as_type_ptr<ov::op::v1::Transpose>(parent);
```
rkazants, line 29: please use `ov::` if possible

src/common/snippets/src/pass/explicit_transpose_matmul_inputs.cpp
```diff
@@ -2,79 +2,98 @@
...
+    }
+
+    // Create new Transpose before Parameter
+    OPENVINO_ASSERT(ov::is_type<opset1::Parameter>(parent),
+                    "ExplicitTransposeMatMulInputs expects Parameter in cases when there isn't existing Transpose on input");
+    const auto& consumers = parent->get_output_target_inputs(0);
+    OPENVINO_ASSERT(consumers.size() == 1,
+                    "ExplicitTransposeMatMulInputs expects Parameter with one consumer in cases when there isn't existing Transpose on input");
+    // Extract Transpose from MatMul
+    const auto rank = input.get_shape().size();
```
rkazants, line 58: Please double-check that shape is static in order to use `get_shape()`

src/common/snippets/src/pass/explicit_transpose_matmul_inputs.cpp
a-sidorova: Yeap, good point. Thanks

src/common/snippets/src/pass/explicit_transpose_matmul_inputs.cpp
a-sidorova: Done, thank you for the review!

src/common/snippets/src/pass/mha_tokenization.cpp
```diff
@@ -288,57 +288,61 @@ ov::snippets::pass::TokenizeMHASnippets::TokenizeMHASnippets() {
 
         // First input branch of MatMul0 should be executed before second input branch of MatMul0,
         // so firstly we insert Transpose1 on the beginning of ordered_ops and then Transpose1
```
IvanNovoselov, line 290: so firstly we insert **Transpose1** on the beginning of ordered_ops and then **Transpose1**
I assume one of them must be Transpose0 🙃 

BTW It's clear what Matmul0 and Matmul1 are, but it's much less obvious where Transpose0 and Transpose1 are located. One may think that Transpose1 is connected to Matmul1 for example. So my suggestion here is either to reflect transposes on the scheme on L216-227, or at least to state somewhere smth like: Transpose0 and Transpose1 are connected to the 0th and 1st inputs of Matmul0.

src/common/snippets/src/pass/explicit_transpose_matmul_inputs.cpp
```diff
@@ -2,79 +2,99 @@
...
+    auto transpose = ov::as_type_ptr<ov::op::v1::Transpose>(parent);
+    while (!transpose && !ov::is_type<ov::op::v0::Parameter>(parent)) {
+        // We can set supported order and transposed_<a|b>=false only if ops have scalar shapes to avoid shape mismatching
+        const auto parent_count = parent->inputs().size();
+        bool are_weights_scalar = true;
+        for (size_t j = 1; j < parent_count; ++j) {
+            are_weights_scalar = are_weights_scalar && ov::shape_size(parent->get_input_shape(j)) == 1;
+        }
+        if (!are_weights_scalar)
+            break;
```
IvanNovoselov, line 26: We have exactly the same code in mha_tokenization.cpp on L308-315. Do we need this logic duplication? If we do, then can we avoid code duplication at least?

a-sidorova: It's needed to avoid the possible conflicts and bugs in the future. Just in case 🙂 
But you're right. We can avoid code duplication. Done, thanks for the idea!

IvanNovoselov (approved):
Followup: discuss whether Brgemm-Transpose fusings should be handled in the main pipeline

dmitry-gorokhov (approved):

PR #17841, opened 2023-06-01
[GNA] Remove warning suppression, fix all warnings

rjeziers:
### Details:
Removed warning suppression from the plugin's CMake.txt. Fixed all warnings.
Most of the warnings were related to implicit type conversion, specifically `size_t` -> `uint32_t`, because of inconsistent usage of these types for sizes and indices (OpenVino core and new code in GNA plugin use mostly `size_t`; legacy GNA code and GNA api use mostly `uint32_t`).
I refactored some of the plugin code to use `size_t` for sizes where it wasn't problematic. In the remaining places, static casts were added.

### Tickets:
 - 93715

### DIFF
--- a/src/plugins/intel_gna/src/backend/am_intel_dnn.cpp
+++ b/src/plugins/intel_gna/src/backend/am_intel_dnn.cpp
@@ -574,7 +574,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
         return l;
     };
 
-    for (int k = 0; k < components.size(); ++k) {
+    for (int k = 0; k < static_cast<int>(components.size()); ++k) {
         std::string l = generate_layer_name(k);
         layersNames.insert(l);
         int lidx = static_cast<int>(std::distance(layersNames.begin(), layersNames.find(l)));
@@ -602,7 +602,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
 
         bool inputConnected = false;
 
-        for (int k2 = 0; k2 < components.size(); ++k2) {
+        for (int k2 = 0; k2 < static_cast<int>(components.size()); ++k2) {
             if (k2 == k)
                 continue;
 
@@ -658,7 +658,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
         }
         if (!inputConnected) {
             // searching for TMP connection
-            size_t tidx = -1;
+            size_t tidx = std::numeric_limits<size_t>::max();
             for (auto&& en : outputs) {
                 if (intersected(en.first, en.second.size, INPUTS(k))) {
                     tidx = en.second.idx;
@@ -671,7 +671,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
                 }
             }
 
-            if (tidx == -1) {
+            if (tidx == std::numeric_limits<size_t>::max()) {
                 outputs[components[k].ptr_inputs] = InputEndPoint(static_cast<int>(outputs.size()),
                                                                   sizeofTensor(INPUTS(k)),
                                                                   components[k].num_bytes_per_input);
@@ -681,7 +681,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
         }
     }
 
-    for (int k = 0; k < components.size(); ++k) {
+    for (int k = 0; k < static_cast<int>(components.size()); ++k) {
         std::string l = generate_layer_name(k);
 
         int tidx = 0;
@@ -1386,7 +1386,7 @@ void AMIntelDNN::InitGNAStruct(Gna2Model* gnaModel) {
         THROW_GNA_EXCEPTION << "out of memory in AMIntelDNN::InitGNAStruct()";
     memset(gnaModel->Operations, 0, gnaModel->NumberOfOperations * sizeof(Gna2Operation));
     gnaOperation = gnaModel->Operations;
-    for (int i = 0; i < component.size(); i++) {
+    for (int i = 0; i < static_cast<int>(component.size()); i++) {
         log::debug() << "Component + " << i << "=GNA_" << std::distance(gnaModel->Operations, gnaOperation) << "\n";
 
         auto& comp = component[i];
--- a/src/plugins/intel_gna/src/backend/am_intel_dnn.cpp
+++ b/src/plugins/intel_gna/src/backend/am_intel_dnn.cpp
@@ -265,7 +265,7 @@ void AMIntelDNN::updateNumberOfOutputsIfPoolingEnabled(Gna2Model& gnaModel, bool
                                                          poolStride.Dimensions[0],
                                                          useLegacyFormula || isOperationCnnLegacySpecific(gnaOp));
             auto& outputTensor = *gnaOp.Operands[OutOpIdx];
-            const_cast<uint32_t&>(outputTensor.Shape.Dimensions[1]) = numberOfOutputs;
+            const_cast<uint32_t&>(outputTensor.Shape.Dimensions[1]) = static_cast<uint32_t>(numberOfOutputs);
         }
     }
 }
@@ -509,7 +509,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
     std::map<void*, InputEndPoint> outputs;
     std::set<std::string> layersNames;
 
-    auto generate_layer_name = [&](int k) {
+    auto generate_layer_name = [&](size_t k) {
         std::string l;
         if (components[k].operation == kDnnPiecewiselinearOp) {
             l += intel_dnn_activation_name[components[k].op.pwl.func_id];
@@ -574,7 +574,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
         return l;
     };
 
-    for (int k = 0; k < components.size(); ++k) {
+    for (size_t k = 0; k < components.size(); ++k) {
         std::string l = generate_layer_name(k);
         layersNames.insert(l);
         int lidx = static_cast<int>(std::distance(layersNames.begin(), layersNames.find(l)));
@@ -602,7 +602,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
 
         bool inputConnected = false;
 
-        for (int k2 = 0; k2 < components.size(); ++k2) {
+        for (size_t k2 = 0; k2 < components.size(); ++k2) {
             if (k2 == k)
                 continue;
 
@@ -658,7 +658,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
         }
         if (!inputConnected) {
             // searching for TMP connection
-            size_t tidx = -1;
+            size_t tidx = std::numeric_limits<size_t>::max();
             for (auto&& en : outputs) {
                 if (intersected(en.first, en.second.size, INPUTS(k))) {
                     tidx = en.second.idx;
@@ -671,7 +671,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
                 }
             }
 
-            if (tidx == -1) {
+            if (tidx == std::numeric_limits<size_t>::max()) {
                 outputs[components[k].ptr_inputs] = InputEndPoint(static_cast<int>(outputs.size()),
                                                                   sizeofTensor(INPUTS(k)),
                                                                   components[k].num_bytes_per_input);
@@ -681,7 +681,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
         }
     }
 
-    for (int k = 0; k < components.size(); ++k) {
+    for (size_t k = 0; k < components.size(); ++k) {
         std::string l = generate_layer_name(k);
 
         int tidx = 0;
@@ -1386,7 +1386,7 @@ void AMIntelDNN::InitGNAStruct(Gna2Model* gnaModel) {
         THROW_GNA_EXCEPTION << "out of memory in AMIntelDNN::InitGNAStruct()";
     memset(gnaModel->Operations, 0, gnaModel->NumberOfOperations * sizeof(Gna2Operation));
     gnaOperation = gnaModel->Operations;
-    for (int i = 0; i < component.size(); i++) {
+    for (size_t i = 0; i < component.size(); i++) {
         log::debug() << "Component + " << i << "=GNA_" << std::distance(gnaModel->Operations, gnaOperation) << "\n";
 
         auto& comp = component[i];
@@ -1565,9 +1565,9 @@ void AMIntelDNN::InitGNAStruct(Gna2Model* gnaModel) {
                             const auto fltStride = fltStrideShape->Dimensions[beginOfHInHW + dimHW];
                             const auto outFromConv = outputFromConv(inputPadded, nFltSize, fltStride);
                             outputTensor.Shape.Dimensions[beginOfHInNHWC + dimHW] =
-                                outputFromPooling(outFromConv,
-                                                  poolWindow->Dimensions[beginOfHInHW + dimHW],
-                                                  poolStride->Dimensions[beginOfHInHW + dimHW]);
+                                static_cast<uint32_t>(outputFromPooling(outFromConv,
+                                                                        poolWindow->Dimensions[beginOfHInHW + dimHW],
+                                                                        poolStride->Dimensions[beginOfHInHW + dimHW]));
                         }
                     }
                     AdvanceOperationIfAllApplied(component, i, gnaOperation);
--- a/src/plugins/intel_gna/src/backend/am_intel_dnn.cpp
+++ b/src/plugins/intel_gna/src/backend/am_intel_dnn.cpp
@@ -265,7 +265,7 @@ void AMIntelDNN::updateNumberOfOutputsIfPoolingEnabled(Gna2Model& gnaModel, bool
                                                          poolStride.Dimensions[0],
                                                          useLegacyFormula || isOperationCnnLegacySpecific(gnaOp));
             auto& outputTensor = *gnaOp.Operands[OutOpIdx];
-            const_cast<uint32_t&>(outputTensor.Shape.Dimensions[1]) = static_cast<uint32_t>(numberOfOutputs);
+            const_cast<uint32_t&>(outputTensor.Shape.Dimensions[1]) = numberOfOutputs;
         }
     }
 }
@@ -509,7 +509,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
     std::map<void*, InputEndPoint> outputs;
     std::set<std::string> layersNames;
 
-    auto generate_layer_name = [&](size_t k) {
+    auto generate_layer_name = [&](int k) {
         std::string l;
         if (components[k].operation == kDnnPiecewiselinearOp) {
             l += intel_dnn_activation_name[components[k].op.pwl.func_id];
@@ -574,7 +574,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
         return l;
     };
 
-    for (size_t k = 0; k < components.size(); ++k) {
+    for (int k = 0; k < components.size(); ++k) {
         std::string l = generate_layer_name(k);
         layersNames.insert(l);
         int lidx = static_cast<int>(std::distance(layersNames.begin(), layersNames.find(l)));
@@ -602,7 +602,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
 
         bool inputConnected = false;
 
-        for (size_t k2 = 0; k2 < components.size(); ++k2) {
+        for (int k2 = 0; k2 < components.size(); ++k2) {
             if (k2 == k)
                 continue;
 
@@ -658,7 +658,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
         }
         if (!inputConnected) {
             // searching for TMP connection
-            size_t tidx = std::numeric_limits<size_t>::max();
+            size_t tidx = -1;
             for (auto&& en : outputs) {
                 if (intersected(en.first, en.second.size, INPUTS(k))) {
                     tidx = en.second.idx;
@@ -671,7 +671,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
                 }
             }
 
-            if (tidx == std::numeric_limits<size_t>::max()) {
+            if (tidx == -1) {
                 outputs[components[k].ptr_inputs] = InputEndPoint(static_cast<int>(outputs.size()),
                                                                   sizeofTensor(INPUTS(k)),
                                                                   components[k].num_bytes_per_input);
@@ -681,7 +681,7 @@ void AMIntelDNN::WriteGraphWizModel(const char* filename) {
         }
     }
 
-    for (size_t k = 0; k < components.size(); ++k) {
+    for (int k = 0; k < components.size(); ++k) {
         std::string l = generate_layer_name(k);
 
         int tidx = 0;
@@ -1386,7 +1386,7 @@ void AMIntelDNN::InitGNAStruct(Gna2Model* gnaModel) {
         THROW_GNA_EXCEPTION << "out of memory in AMIntelDNN::InitGNAStruct()";
     memset(gnaModel->Operations, 0, gnaModel->NumberOfOperations * sizeof(Gna2Operation));
     gnaOperation = gnaModel->Operations;
-    for (size_t i = 0; i < component.size(); i++) {
+    for (int i = 0; i < component.size(); i++) {
         log::debug() << "Component + " << i << "=GNA_" << std::distance(gnaModel->Operations, gnaOperation) << "\n";
 
         auto& comp = component[i];
@@ -1565,9 +1565,9 @@ void AMIntelDNN::InitGNAStruct(Gna2Model* gnaModel) {
                             const auto fltStride = fltStrideShape->Dimensions[beginOfHInHW + dimHW];
                             const auto outFromConv = outputFromConv(inputPadded, nFltSize, fltStride);
                             outputTensor.Shape.Dimensions[beginOfHInNHWC + dimHW] =
-                                static_cast<uint32_t>(outputFromPooling(outFromConv,
-                                                                        poolWindow->Dimensions[beginOfHInHW + dimHW],
-                                                                        poolStride->Dimensions[beginOfHInHW + dimHW]));
+                                outputFromPooling(outFromConv,
+                                                  poolWindow->Dimensions[beginOfHInHW + dimHW],
+                                                  poolStride->Dimensions[beginOfHInHW + dimHW]);
                         }
                     }
                     AdvanceOperationIfAllApplied(component, i, gnaOperation);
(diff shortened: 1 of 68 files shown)


src/plugins/intel_gna/src/frontend/scale_factor_calc.cpp
```diff
@@ -306,8 +306,8 @@ bool ScaleFactorCalculator::requantizeInput(InferenceEngine::CNNLayerPtr input,
         return false;
     }
 
-    auto prevLayer = InferenceEngine::CNNNetHasPrevLayer(layer.get(), prevInputIdx)
-                         ? InferenceEngine::CNNNetPrevLayer(layer, prevInputIdx)
+    auto prevLayer = InferenceEngine::CNNNetHasPrevLayer(layer.get(), static_cast<int>(prevInputIdx))
```
marcinkusm, line 309: you set second argument to size_t and not you cast to int? 
inline bool CNNNetHasPrevLayer(const InferenceEngine::CNNLayer* layer, size_t idx = 0) {

shouldn't this produce warning?

https://github.com/openvinotoolkit/openvino/pull/17841/files#diff-637b2179f395ecedb9fb41cb1ee70aa666b3b7c2ce3d8fb435b9c3c8bbc248ffR355

src/plugins/intel_gna/src/common/numerical_utils.hpp
```diff
@@ -28,6 +28,18 @@ inline int32_t FloatToInt32(float a) {
 inline int64_t FloatToInt64(float a) {
     return FloatToInteger<int64_t>(a);
 }
+inline int8_t FloatToInt8(double a) {
```
marcinkusm, line 31:
```suggestion
//Option 1. Add method for doubles.
template <typename T>
inline T DoubleToInteger(Double a) {
    return static_cast<T>((a < 0.0f) ? (a - 0.5f) : (a + 0.5f));
}
//Option 2. or improve template methods.
template <typename T, Typename U,  typename std::enable_if<std::is_floating_point<U>::value>::type* = nullptr> 
inline T DoubleToInteger(U a) {
    return static_cast<T>((a < 0.0f) ? (a - static_cast<U>(0.5)) : (a + static_cast<U>(0.5)));
}
```

src/plugins/intel_gna/src/common/numerical_utils.hpp
```diff
@@ -28,6 +28,18 @@ inline int32_t FloatToInt32(float a) {
...
+    return FloatToInteger<int8_t>(static_cast<float>(a));
```
marcinkusm, line 32: Rename to DoubleToInt16
and remove casting.

src/plugins/intel_gna/src/frontend/model_quantizer.hpp
```diff
@@ -92,10 +92,10 @@ class ModelQuantizer {
             // We are looking for infinite loop by using algorithm of compute prefix function, complexity O(N)
             // (a part of the Knuth–Morris–Pratt algorithm).
             std::map<int, int> prefix_function;
-            int32_t k = inf_loop_history.size();
-            for (int32_t i = inf_loop_history.size() - 2; i >= 0; i--) {
+            size_t k = inf_loop_history.size();
```
marcinkusm, line 95:
```suggestion
            auto k = inf_loop_history.size();
```

src/plugins/intel_gna/src/frontend/model_quantizer.hpp
```diff
@@ -127,9 +127,9 @@ class ModelQuantizer {
                         log::debug() << "\t " << s << '\n';
                     }
                     inf_loop_pattern.clear();
-                    int pattern_len = (inf_loop_history.size() - i) / 2;
+                    size_t pattern_len = (inf_loop_history.size() - i) / 2;
```
marcinkusm, line 130:
```suggestion
                    auto pattern_len = (inf_loop_history.size() - i) / 2;
```

src/plugins/intel_gna/src/frontend/model_quantizer.hpp
```diff
@@ -141,7 +141,7 @@ class ModelQuantizer {
                     break;
                 }
 
-                prefix_function.emplace(i, k);
+                prefix_function.emplace(i, static_cast<int>(k));
```
marcinkusm, line 144: can prefix_function have negative values? if not maybe it would be good to change map parameters types to size_t/uint32_t?

src/plugins/intel_gna/src/frontend/quantization.hpp
```diff
@@ -81,7 +81,7 @@ inline T SaturationCast(float value, uint32_t* saturation_counter = nullptr) {
 /**
  * @brief Apply FQ levels onto a value
  */
-float ApplyFQ(float value, float input_low, float input_high, float output_low, float output_high, uint32_t levels);
+float ApplyFQ(float value, float input_low, float input_high, float output_low, float output_high, size_t levels);
```
marcinkusm, line 84: do you need size_t to uint32_t because of some specific reason or because you changed argument type used for invocation of the method from auto to size_t?

src/plugins/intel_gna/src/frontend/scale_factor_calc.cpp
```diff
@@ -261,7 +261,7 @@ bool ScaleFactorCalculator::requantizeInput(InferenceEngine::CNNLayerPtr input,
     if (info.isWeightableIdentity() && !AreFpEq(quantDataForInputLayer->_weights_quant.GetScale(), 1.0f)) {
         auto reducer = std::max(1.0f, quantDataForInputLayer->_dst_quant.GetScale() / newOutputScale);
         auto newWeightsScale = std::max(1.0f, quantDataForInputLayer->_weights_quant.GetScale() / reducer);
-        quantDataForInputLayer->_weights_quant.SetScale(static_cast<int32_t>(newWeightsScale));
+        quantDataForInputLayer->_weights_quant.SetScale(trunc(newWeightsScale));
```
marcinkusm, line 264: don't you have warning here?
It looks like newWeightsScale is double and trunc returns double, and later you pass it to SetScale which takes float.

src/plugins/intel_gna/src/frontend/scale_factor_calc.cpp
```diff
@@ -410,7 +410,8 @@ float ScaleFactorCalculator::getActivationScale(InferenceEngine::CNNLayer const*
 #endif
     } else if (layer.isRelu()) {
         // if activation is one from relu family, we need to apply heuristic to avoid activation output overflow
-        auto limit = (inputsSize == 1 ? std::numeric_limits<int8_t>::max() : std::numeric_limits<int32_t>::max()) - 1;
+        auto limit = static_cast<unsigned int>(
```
marcinkusm, line 413: it is later used to comapred with uint64_t why don't use this type for casting?

src/plugins/intel_gna/src/frontend/scale_factor_calc.cpp
```diff
@@ -800,7 +801,7 @@ bool ScaleFactorCalculator::ScaleFactorPerLayerCNN(InferenceEngine::CNNLayer* cn
         }
         return true;
     } else if (layerInfo.isCropAffined()) {
-        auto weightsScaleFactor = 1;
+        auto weightsScaleFactor = 1.0f;
```
marcinkusm, line 804:
```suggestion
        const auto weightsScaleFactor = 1.0f;
```

src/plugins/intel_gna/src/frontend/scale_factor_calc.cpp
```diff
@@ -921,7 +922,7 @@ bool ScaleFactorCalculator::ScaleFactorPerLayerEltwise(InferenceEngine::EltwiseL
             return true;
         }
 
-        auto weightsReducer = calculateWeightsReducerFromDstStats(quantData->_dst_quant);
+        float weightsReducer = static_cast<float>(calculateWeightsReducerFromDstStats(quantData->_dst_quant));
         if (weightsReducer > initial_weights_reducer_val) {
```
marcinkusm, line 926: initial_weights_reducer_val is double and it was compared to double , but you changed it to float.
This is quite risky, due the fact precision is very important here. I would not change type from double to float.

If this is only to fix warning in line 927 I would change it like that.
```
            float newOutputScale = quantParams1->_dst_quant.GetScale() / weightsReducer;
//to
            float newOutputScale = quantParams1->_dst_quant.GetScale() / static_cast<float>(weightsReducer);

```

src/plugins/intel_gna/src/gna_device_allocation.hpp
```diff
@@ -84,7 +84,7 @@ struct GnaAllocation {
     }
 
     uint32_t sizeForExport() const {
-        return ALIGN64(sizeRequested);
+        return ALIGN64(static_cast<uint32_t>(sizeRequested));
```
marcinkusm, line 87: why just not change type of sizeRequested in header?

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -337,15 +337,23 @@ void GNAGraphCompiler::ConvolutionPrimitive(InferenceEngine::CNNLayerPtr layer)
...
-    const auto in_channels = InferenceEngine::GetDataDimByName(inputs, InferenceEngine::DataDimName::C);
-    auto in_height = InferenceEngine::GetDataDimByName(inputs, InferenceEngine::DataDimName::H);
-    auto in_width = InferenceEngine::GetDataDimByName(inputs, InferenceEngine::DataDimName::W);
-
-    const auto out_batch = InferenceEngine::GetDataDimByName(outputs, InferenceEngine::DataDimName::N);
-    const auto out_channels = InferenceEngine::GetDataDimByName(outputs, InferenceEngine::DataDimName::C);
-    auto out_height = InferenceEngine::GetDataDimByName(outputs, InferenceEngine::DataDimName::H);
-    auto out_width = InferenceEngine::GetDataDimByName(outputs, InferenceEngine::DataDimName::W);
+    const uint32_t in_batch =
+        static_cast<uint32_t>(InferenceEngine::GetDataDimByName(inputs, InferenceEngine::DataDimName::N));
```
marcinkusm, line 341: are this changes needed?

GetDataDimByName returns uint32_t
so you don't need change auto to uint32_t and you don't need to use static_cast.

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -543,10 +551,10 @@ void GNAGraphCompiler::finalizeConvolution1DPrimitive(InferenceEngine::CNNLayerP
     auto biasPrecision =
         convolution._biases ? convolution._biases->getTensorDesc().getPrecision() : outputs->getPrecision();
 
-    uint32_t num_bytes_per_input = inputs->getPrecision().size();
-    uint32_t num_bytes_per_output = outputs->getPrecision().size();
-    uint32_t num_bytes_per_weight = convolution._weights->getTensorDesc().getPrecision().size();
-    uint32_t num_bytes_per_bias = biasPrecision.size();
+    uint32_t num_bytes_per_input = static_cast<uint32_t>(inputs->getPrecision().size());
```
marcinkusm, line 554: you have to change all this types because you changes size_t to uint32_t.

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -1417,23 +1434,24 @@ void GNAGraphCompiler::EltwisePrimitive(InferenceEngine::CNNLayerPtr layer) {
...
-        quantized == nullptr ? inputs2Bytes->getPrecision().size() : (gna_config.gnaFlags.input_low_precision ? 1 : 2),
-        quantized == nullptr ? inputs4Bytes->getPrecision().size() : (gna_config.gnaFlags.input_low_precision ? 1 : 4),
-        GetScaleFactor(layer, QuantizedDataType::weights),
-        GetScaleFactor(layer, QuantizedDataType::output),
-        ptr_inputs,
-        ptr_outputs,
-        ptr_weights,
-        ptr_biases,
-        true);
+    dnn->InitAffineComponent(currentComponent,
```
marcinkusm, line 1437: was this not formatted properly earlier?

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -1520,12 +1538,12 @@ void GNAGraphCompiler::GemmPrimitive(InferenceEngine::CNNLayerPtr layer) {
...
     auto in_dims = input_1->getDims();
-    auto batch_size = (in_dims.size() == 1) ? 1 : in_dims.front();
-    uint32_t num_rows_in = InferenceEngine::details::product(in_dims) / batch_size;
+    uint32_t batch_size = (in_dims.size() == 1) ? 1 : static_cast<uint32_t>(in_dims.front());
+    uint32_t num_rows_in = static_cast<uint32_t>(InferenceEngine::details::product(in_dims)) / batch_size;
     uint32_t num_columns_in = batch_size;
     const auto out_dims = outputs->getDims();
     const auto out_dims_size = ngraph::shape_size(out_dims);
-    uint32_t num_rows_out = InferenceEngine::GetDimFromBack(out_dims, 1);
+    uint32_t num_rows_out = static_cast<uint32_t>(InferenceEngine::GetDimFromBack(out_dims, 1));
```
marcinkusm, line 1546: does not GetDimFromBack return uint32_t by default?

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -2041,10 +2060,10 @@ void GNAGraphCompiler::PWLPrimitive(InferenceEngine::CNNLayerPtr layer) {
 
     auto orientation = kDnnInterleavedOrientation;
 
-    uint32_t w_dim_in = GetDataDimByName(inputs, DataDimName::W);
-    uint32_t h_dim_in = GetDataDimByName(inputs, DataDimName::H);
-    uint32_t c_dim_in = GetDataDimByName(inputs, DataDimName::C);
-    uint32_t n_dim_in = GetDataDimByName(inputs, DataDimName::N);
+    uint32_t w_dim_in = static_cast<uint32_t>(GetDataDimByName(inputs, DataDimName::W));
```
marcinkusm, line 2063: GetDataDimByName returns uint32_t by default

src/plugins/intel_gna/src/gna_graph_compiler.cpp
```diff
@@ -2588,8 +2607,8 @@ void GNAGraphCompiler::connectOutput(InferenceEngine::CNNLayerPtr layer, void* p
 ConnectionDetails GNAGraphCompiler::connectInput(CNNLayerPtr layer,
                                                  void* ptr,
                                                  size_t num_data_bytes_in,
-                                                 int32_t offset,
-                                                 int idx,
+                                                 size_t offset,
+                                                 size_t idx,
```
marcinkusm, line 2611: when you change `int` to `size_t` ensure that it is safe in many places in code negative values is used to find issue or special case - in such case you have unexpected behavior if you lost the sign.

src/plugins/intel_gna/src/gna_graph_tools.hpp
```diff
@@ -86,7 +86,7 @@ inline InferenceEngine::CNNLayerPtr CNNNetPrevLayer(const InferenceEngine::CNNLa
  * @param idx - index in previous layer collection
  * @param layer
  */
-inline InferenceEngine::CNNLayerPtr CNNNetPrevLayer(const InferenceEngine::CNNLayer* layer, int idx = 0) {
+inline InferenceEngine::CNNLayerPtr CNNNetPrevLayer(const InferenceEngine::CNNLayer* layer, size_t idx = 0) {
```
marcinkusm, line 89: ensure that negative value is not used as condition before you change int to size_t

marcinkusm (changes requested):

src/plugins/intel_gna/src/frontend/scale_factor_calc.cpp
```diff
@@ -284,7 +284,7 @@ bool ScaleFactorCalculator::requantizeInput(InferenceEngine::CNNLayerPtr input,
                 break;
             }
         }
-        auto prevLayer = InferenceEngine::CNNNetPrevLayer(layer, prevInputIdx);
+        auto prevLayer = InferenceEngine::CNNNetPrevLayer(layer, static_cast<int>(prevInputIdx));
```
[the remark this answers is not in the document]
rjeziers: A leftover from before I changed the function signatures - I'll check all places

src/plugins/intel_gna/src/gna_graph_compiler.cpp
rjeziers: I think formatting changed due to additional casting of the arguments

src/plugins/intel_gna/src/frontend/scale_factor_calc.cpp
rjeziers: newWeightsScale is float, so trunc also returns a float.

src/plugins/intel_gna/src/frontend/model_quantizer.hpp
rjeziers: `prefix_function` can't have negative values, but `i` can, in fact `i >= 0` is a condition for searching an infinite loop; i've changed map parameter types to <int, size_t>

src/plugins/intel_gna/src/frontend/model_quantizer.hpp
```diff
@@ -57,7 +57,7 @@ class ModelQuantizer {
             auto input_layer = getCreatorLayer(inputData.second->getInputData()).lock();
             auto quant_data = InferenceEngine::getInjectedData<frontend::QuantizedLayerParams>(input_layer);
             IE_ASSERT(quant_data != nullptr);
-            quant_data->_src_quant.SetScale(inputs.at(input_layer->name).scale_factor);
+            quant_data->_src_quant.SetScale(static_cast<float>(inputs.at(input_layer->name).scale_factor));
```
tadamowicz, line 60: I would suggest change type of the scale_factor from double to float

src/plugins/intel_gna/src/optimizer/gna_pass_manager.cpp
```diff
@@ -486,22 +486,23 @@ void ReorderMaxPoolPass::run() {
...
+     * power |                  power |
+     *  \   /                    \   /
+     *    mul                      mul
+     */
 
     auto hasNChildren = [](CNNLayerPtr l, int N) {
         if (l->outData.size() != 1)
             return false;
-        if (getInputTo(l->outData.front()).size() != N)
+        if (static_cast<int>(getInputTo(l->outData.front()).size()) != N)
```
marcinkusm, line 505: just wonder if it wouldn't be better to make N as size_t

src/plugins/intel_gna/src/gna_plugin.cpp
```diff
@@ -1347,11 +1348,11 @@ RequestStatus GNAPlugin::WaitFor(uint32_t request_idx, int64_t millisTimeout) {
         ExportScores(output_blob->buffer(),
                      gna_output_blob->cbuffer(),
                      gna_output_desc.orientation,
-                     batchSize,
-                     batchSize,
-                     elementsPerBatch,
-                     elementsPerBatch,
-                     elementsPerBatch,
+                     static_cast<uint32_t>(batchSize),
```
marcinkusm, line 1351: what is a type of batchSize and elementsPerBatch?
cannot they be declared as uint32_t to avoid casting in many places?

src/plugins/intel_gna/src/runtime/floatmath.cpp
```diff
@@ -9,10 +9,6 @@
 #include <cstdint>
 #include <cstdio>
 
-#ifdef __cplusplus
```
marcinkusm, line 12: are you sure about this removal ?

src/plugins/intel_gna/src/serial/gna_model_serial.cpp
```diff
@@ -63,7 +63,7 @@ inline void write_pre_processing_model(const std::shared_ptr<ov::Model>& model,
     writeString(xml_buf.str(), os);
 
     // write BIN
-    size_t ir_bin_size = bin_buf.str().size();
+    uint32_t ir_bin_size = static_cast<uint32_t>(bin_buf.str().size());
```
marcinkusm, line 66: please check writeBits it is template method and it uses size of template type for writing a file of model.
you cannot change a type passed there.

src/plugins/intel_gna/src/serial/gna_model_serial.cpp
```diff
@@ -44,8 +44,8 @@ inline void writeBits(const T& obj, std::ostream& os) {
 
 inline void writeString(const std::string& str, std::ostream& os) {
     const char* c_str = str.c_str();
-    const size_t str_len = strlen(c_str) + 1;
-    writeBits(static_cast<uint32_t>(str_len), os);
+    const uint32_t str_len = static_cast<uint32_t>(strlen(c_str)) + 1;
+    writeBits(str_len, os);
```
marcinkusm, line 48: don't change a type of variable passes to writeBits or writeNBytes

src/plugins/intel_gna/src/backend/am_intel_dnn.cpp
```diff
@@ -1386,7 +1386,7 @@ void AMIntelDNN::InitGNAStruct(Gna2Model* gnaModel) {
         THROW_GNA_EXCEPTION << "out of memory in AMIntelDNN::InitGNAStruct()";
     memset(gnaModel->Operations, 0, gnaModel->NumberOfOperations * sizeof(Gna2Operation));
     gnaOperation = gnaModel->Operations;
-    for (int i = 0; i < component.size(); i++) {
+    for (int i = 0; i < static_cast<int>(component.size()); i++) {
```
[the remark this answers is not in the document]
rjeziers: `i` is used as an argument in 7 different places - if it was `size_t` we'd need to cast it to `int` 7 times, this is a better option imho

src/plugins/intel_gna/src/backend/gna_limitations.cpp
```diff
@@ -944,7 +945,7 @@ bool Limitations::validate_concat_axis(const InferenceEngine::CNNLayerPtr layer,
                        ptr->GetParamAsInts("order") == std::vector<int32_t>{0, 2, 1} /* NCW to NWC */))));
         };
 
-        for (auto input_idx = 0; input_idx != concat_layer->insData.size(); input_idx++) {
+        for (auto input_idx = 0; input_idx != static_cast<int>(concat_layer->insData.size()); input_idx++) {
             prev_layer = InferenceEngine::CNNNetPrevLayerSkipCertain(layer, input_idx, isFusableWithConv);
```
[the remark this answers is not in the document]
rjeziers: With option 1, I'd need to:
1. cast this argument inside CNNNetPrevLayerSkipCertain again to int, or
2. refactor whole graph tools - something which I've tried to do before.

If we don't want to introduce a lot of refactoring in this PR, option 2 is better.

src/plugins/intel_gna/src/runtime/floatmath.cpp
rjeziers: For me, it looks like this file was copy-pasted from some library in the early stages of this project and nobody bothered to modify it afterwards. I don't think these functions are actually used anywhere else.

src/plugins/intel_gna/src/serial/gna_model_serial.cpp
rjeziers: I didn't change the type - before, str_len was size_t but it was casted to uint32_t as an argument to writeBits. Now it's declared as uint32_t to avoid unnecessary cast to uint32_t in the next line.

writeBits was and is called with uint32_t; writeNBytes accepts only uint32_t, so it was called with size_t with implicit conversion to uint32_t; now it's called explicitly with uint32_t.

Nothing changed here.

src/plugins/intel_gna/src/runtime/floatmath.cpp
marcinkusm: this is ifdef and extern C are used to provide C like api( to prevent name mangling).
If you are sure that it is not needed we can remove it.

src/plugins/intel_gna/src/backend/am_intel_dnn.hpp
```diff
@@ -272,22 +272,22 @@ class AMIntelDNN {
     }
 
     template <class T>
-    void AdvanceOperationIfAllApplied(const std::vector<intel_dnn_component_t>& cmp, int i, T*& operation) {
+    void AdvanceOperationIfAllApplied(const std::vector<intel_dnn_component_t>& cmp, size_t i, T*& operation) {
         if (i == cmp.size() - 1 || cmp[i + 1].operation != kDnnPiecewiselinearOp) {
```
marcinkusm, line 276: I suggested change to `i + 1 = cma.size()` becayse I wanted to avoid situation that cmp.size() will be 0 so substracting 1 would be undefined behaviour, but I understand it is not possible here.

src/plugins/intel_gna/src/gna_graph_tools.hpp
```diff
@@ -173,12 +173,12 @@ inline std::pair<InferenceEngine::CNNLayerPtr, std::vector<int>> CNNNetCheckNext
     int iidx,
```
marcinkusm, line 173: oidx, iidx, new_oidx  and new_iidx are compared everywhere to size_t  from  .size() of containers.
it they cannot have negavative value maybe it would be good to change their type to size_t?

src/plugins/intel_gna/src/serial/gna_model_serial.cpp
```diff
@@ -149,7 +149,7 @@ header_latest::ModelHeader GNAModelSerial::ReadHeader(std::istream& is) {
     auto size_of_headers_header = sizeof(header_latest::ModelHeader::gnam) +
                                   sizeof(header_latest::ModelHeader::headerSize) +
                                   sizeof(header_latest::ModelHeader::Version);
-    if (stream_len > size_of_headers_header) {
+    if (static_cast<uint32_t>(stream_len) > size_of_headers_header) {
```
marcinkusm, line 152: are you sure that stream_len at this point will be >= 0?

rjeziers: yes, in order for it to be <0 the stream starting position would have to be greater than stream length which is impossible

src/plugins/intel_gna/src/backend/am_intel_dnn.hpp
rjeziers: If `cmp.size() == 0` we'd have an out-of-bounds error anyway, in the next condition. I could add some check for correctness of the data, but:
- it has worked before
- it's legacy code anyway

do you think it's in scope of this task?

src/plugins/intel_gna/src/gna_graph_tools.hpp
rjeziers: oidx, iidx are arguments to the function, so I'd need to refactor a lot of other functions from gna_graph_tools.hpp, which is what we wanted to avoid in this PR - maybe something for the future

marcinkusm (approved):
👍

PR #17855, opened 2023-06-02
Cecilia/poc/zerocopy

ceciliapeng2011:
### Details:
 - *Implement zero-copy output between cpu plugin Graph and infer_request*
 - *...*

### To Do list:

- [ ] Perform cases when users call set_tensor for output.
- [ ] Perform cases when actual TensorDesc is not the same as expected tensordesc.
- [ ] Handle cases when there are more than one infer_request performing over a Graph
- [ ] Perform cases when PreProcessing
- [ ] Perform cases when Split is the last output node.
- [ ] Perform cases when output is a ROI.
- [ ] Perform accuracy and performance benchmarking across model zoo

### Tickets:
 - *109294*


src/plugins/intel_cpu/tests/unit/shape_inference_test/transpose_shape_infernece_test.cpp
ceciliapeng2011, line 1: It is a mistake I committed this file change. I will remove it later.

src/frontends/paddle/CMakeLists.txt
ceciliapeng2011, line 1: I will remove this change later.. a mistake unintentionally

src/plugins/intel_cpu/src/node.cpp
```diff
@@ -612,7 +612,10 @@ void Node::redefineOutputMemory(const std::vector<VectorDims> &newOutputShapes)
         const bool hasZeroDims = std::count(std::begin(newOutputShape), std::end(newOutputShape), 0) > 0;
         const auto memDesc = getBaseMemDescAtOutputPort(i)->cloneWithNewDims(newOutputShape, hasZeroDims);
         for (size_t j = 0; j < edges.size(); j++) {
+            auto old_mem_ptr = edges[j]->getMemoryPtr()->GetData();
             edges[j]->getMemoryPtr()->redefineDesc(memDesc);
+            auto new_mem_ptr = edges[j]->getMemoryPtr()->GetData();
+            DEBUG_LOG(getName(), " edge", j, " ", old_mem_ptr, " -> ", new_mem_ptr);
```
ceciliapeng2011, line 618: Just for debugging, I will revert this change later.

src/plugins/intel_cpu/src/infer_request.h
```diff
@@ -56,7 +56,9 @@ class InferRequestBase : public InferenceEngine::IInferRequestInternal {
     virtual void PushInputData() = 0;
 
     Graph* graph = nullptr;
-    std::unordered_map<std::string, void*> externalPtr;
+    std::vector<std::string> inplacedOutPorts;
+    std::vector<std::string> externalInPorts;
+    std::vector<std::string> externalOutPorts;
```
ceciliapeng2011, line 62: There is no need to use externalPtr, use externalInPorts and externalOutPorts to keep port names instead, as we can access blob from _inputs and _outputs of InferRequestBase class.
externalPtr would make syncing between InferRequest and Graph more troublesome. So I replaced it.

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -950,7 +952,21 @@ void Graph::PullOutputData(BlobMap &out) {
             if (expectedDesc.getLayout() == InferenceEngine::Layout::BLOCKED) {
                 expectedDesc = TensorDesc(expectedDesc.getPrecision(), expectedDesc.getLayout());
             }
-            out[name]->setShape(outDims);
+
+            if (std::count(inplacedOutPorts.begin(), inplacedOutPorts.end(), name) &&
```
ceciliapeng2011, line 956: Sync up InferRequest output with resized Graph output.

The old value is destroyed as it is unreferenced. Graph output is switched to InferRequest.

Probably we need recover Graph output to its previous memory (and memory manager)... in case it serves another infer_request which cannot changeDefaultPtr.

maxnick:
@ceciliapeng2011 is this PR still relevant? If not, I suggest closing.

PR #17856, opened 2023-06-02
[PT FE] Support inplace operations on aliases of tensors

mvafin:
### Details:
 - *Support inplace operations on aliases of tensors*

### Tickets:
 - *109778*

### DIFF
--- a/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
+++ b/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
@@ -91,7 +91,7 @@ pt_to_ov_type_map = {
 
 
 class TorchScriptPythonDecoder (Decoder):
-    def __init__(self, pt_module, graph_element=None, example_input=None, freeze=True):
+    def __init__(self, pt_module, graph_element=None, example_input=None, freeze=True, alias_db=None):
         Decoder.__init__(self)
         # We store every decoder created by this decoder so that all them are not deleted until the first decoder is deleted
         self.m_decoders = []
@@ -113,8 +113,10 @@ class TorchScriptPythonDecoder (Decoder):
                     " yourself, please refer to PyTorch documentation: "
                     "https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html.")
             self.graph_element = pt_module.inlined_graph
+            self.alias_db = self.graph_element.alias_db()
         else:
             self.graph_element = graph_element
+            self.alias_db = alias_db
         self.pt_module = pt_module
         self.raw_inputs = list(self.graph_element.inputs())
         self.raw_outputs = list(self.graph_element.outputs())
@@ -273,7 +275,7 @@ class TorchScriptPythonDecoder (Decoder):
     def visit_subgraph(self, node_visitor) -> None:
         # make sure topological order is satisfied
         for node in self.graph_element.nodes():
-            decoder = TorchScriptPythonDecoder(self.pt_module, node)
+            decoder = TorchScriptPythonDecoder(self.pt_module, node, alias_db=self.alias_db)
             self.m_decoders.append(decoder)
             node_visitor(decoder)
 
@@ -289,7 +291,7 @@ class TorchScriptPythonDecoder (Decoder):
         return list(self.graph_element.blocks())
 
     def get_subgraph_decoder(self, index: int):
-        decoder = TorchScriptPythonDecoder(self.pt_module, self.get_subgraphs()[index])
+        decoder = TorchScriptPythonDecoder(self.pt_module, self.get_subgraphs()[index], alias_db=self.alias_db)
         self.m_decoders.append(decoder)
         return decoder
 
@@ -388,6 +390,13 @@ class TorchScriptPythonDecoder (Decoder):
                     pt_value = get_value_from_getattr(in_node, self.pt_module)
                     return pt_value is None
         return False
+
+    def may_produce_alias(self, in_index: int, out_index: int) -> bool:
+        try:
+            return self.alias_db.may_contain_alias(self._raw_input(in_index), self._raw_output(out_index))
+        except:
+            # Sometimes pytorch fails to get result with IndexError exception while these indexes exist in node
+            return False
 
     @staticmethod
     def _transform_tensor_list_constants_to_listconstruct(graph: torch.Graph):
--- a/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
+++ b/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
@@ -392,6 +392,9 @@ class TorchScriptPythonDecoder (Decoder):
         return False
 
     def may_produce_alias(self, in_index: int, out_index: int) -> bool:
+        if self.get_op_type() in ["aten::conv1d", "aten::conv2d", "aten::conv3d"]:
+            # AliasDB::may_contain_alias sometimes return True for tensors produced by convnd, we have to workaround that
+            return False
         try:
             return self.alias_db.may_contain_alias(self._raw_input(in_index), self._raw_output(out_index))
         except:
--- a/src/frontends/pytorch/src/frontend.cpp
+++ b/src/frontends/pytorch/src/frontend.cpp
@@ -42,8 +42,10 @@ std::set<std::string> get_unconverted_types_from_model(const std::shared_ptr<Mod
     std::set<std::string> unconverted_ops_types;
     for (const auto& node : model->get_ordered_ops()) {
         if (const auto& fw_node = ov::as_type_ptr<PtFrameworkNode>(node)) {
-            auto op_type = fw_node->get_decoder()->get_op_type();
-            unconverted_ops_types.insert(op_type);
+            auto attrs = fw_node->get_attrs();
+            FRONT_END_GENERAL_CHECK(attrs.find("PtTypeName") != attrs.end(),
+                                    "FrameworkNode attributes do not contain PtTypeName.");
+            unconverted_ops_types.insert(attrs.at("PtTypeName"));
         }
         if (const auto& fw_node = ov::as_type_ptr<ov::op::util::MultiSubGraphOp>(node)) {
             for (size_t i = 0; i < fw_node->get_internal_subgraphs_size(); i++) {
--- a/src/frontends/pytorch/src/pt_framework_node.hpp
+++ b/src/frontends/pytorch/src/pt_framework_node.hpp
@@ -14,13 +14,18 @@ class PtFrameworkNode : public ov::op::util::FrameworkNode {
 public:
     OPENVINO_OP("PtFrameworkNode", "util", ::ov::op::util::FrameworkNode);
 
-    PtFrameworkNode(const std::shared_ptr<TorchDecoder>& decoder, const OutputVector& inputs, size_t output_size)
+    PtFrameworkNode(const std::shared_ptr<TorchDecoder>& decoder, const OutputVector& inputs, size_t output_size, bool is_backprop=false)
         : ov::op::util::FrameworkNode(inputs, output_size, decoder->get_subgraph_size()),
           m_decoder(decoder) {
         ov::op::util::FrameworkNodeAttrs attrs;
         attrs.set_type_name("PTFrameworkNode");
-        attrs["PtTypeName"] = m_decoder->get_op_type();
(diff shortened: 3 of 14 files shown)


src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -113,8 +113,10 @@ def __init__(self, pt_module, graph_element=None, example_input=None, freeze=Tru
                     " yourself, please refer to PyTorch documentation: "
                     "https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html.")
             self.graph_element = pt_module.inlined_graph
+            self.alias_db = self.graph_element.alias_db()
```
rkazants, line 116: does it always have `alias_db()` method for both `trace` and `script` cases?

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -389,6 +391,16 @@ def input_is_none(self, index: int) -> bool:
                     return pt_value is None
         return False
 
+    def may_produce_alias(self, in_index: int, out_index: int) -> bool:
+        if self.get_op_type() in ["aten::conv1d", "aten::conv2d", "aten::conv3d"]:
+            # AliasDB::may_contain_alias sometimes return True for tensors produced by convnd, we have to workaround that
```
rkazants, line 396: strange moment. Why is it valid for these particular operations? It can be a case for other operations then

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -42,8 +42,10 @@ std::set<std::string> get_unconverted_types_from_model(const std::shared_ptr<Mod
     std::set<std::string> unconverted_ops_types;
     for (const auto& node : model->get_ordered_ops()) {
         if (const auto& fw_node = ov::as_type_ptr<PtFrameworkNode>(node)) {
-            auto op_type = fw_node->get_decoder()->get_op_type();
-            unconverted_ops_types.insert(op_type);
+            auto attrs = fw_node->get_attrs();
+            FRONT_END_GENERAL_CHECK(attrs.find("PtTypeName") != attrs.end(),
+                                    "FrameworkNode attributes do not contain PtTypeName.");
```
rkazants, line 47: user unfriendly message error, please correct if possible. `PtTypeName` is confusing

src/frontends/pytorch/src/translate_session.cpp
```diff
@@ -143,10 +145,29 @@ std::shared_ptr<Model> TranslateSession::convert_pytorch_model(
...
+                        std::tie(recorded_in_tensor_id, recorded_node, std::ignore) = m_may_be_alias.at(fw_tensor_id);
+                        FRONT_END_GENERAL_CHECK(recorded_in_tensor_id == in_tensor_id,
+                                                "Operation ",
+                                                context.get_op_type(),
+                                                " creates alias to tensor already added to may_produce_alias list by ",
+                                                recorded_node->get_op_type(),
+                                                ", but from different tensor: ",
+                                                in_tensor_id,
+                                                " vs ",
+                                                recorded_in_tensor_id);
```
rkazants, line 165: also, maybe it makes sense to consider more user friendly message?

src/frontends/pytorch/src/translate_session.cpp
```diff
@@ -257,6 +281,111 @@ size_t TranslateSession::decode_tensor_name(const Output<Node>& output) {
...
+using BackpropCreatorFunction =
+    std::function<ov::Output<ov::Node>(std::shared_ptr<TorchDecoder>, Output<Node>, Output<Node>)>;
+
+Output<Node> TranslateSession::get_backprop_op(std::shared_ptr<TorchDecoder> node,
+                                               Output<Node> direct_op_output,
+                                               Output<Node> value) {
+    std::map<std::string, BackpropCreatorFunction> backprop_map = {
+        {"aten::slice", slice_backprop},
+        {"aten::select", select_backprop},
+    };
```
rkazants, line 372: not clear why is this map needed? Should we create for other operations? Also, does it make sense to place them into operation specific files?

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
mvafin: Yes, it does

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
mvafin: It can be and this is really confusing, but currently I saw this happen on conv2d only

src/frontends/pytorch/src/transforms/index_loop_getitem_replacer.cpp
```diff
@@ -0,0 +1,154 @@
...
+        inp_descs.push_back(std::make_shared<ov::op::util::MultiSubGraphOp::InvariantInputDescription>(
+            loop_op->get_input_size() - 1,
+            body->get_parameters().size() - 1));
+        loop_op->set_input_descriptions(0, inp_descs);
+
+        auto start = std::make_shared<v1::Multiply>(chunk_counter, chunks_size_body);
+        auto stop = std::make_shared<v1::Add>(start, chunks_size_body);
+        auto curr_chunk = std::make_shared<v8::Slice>(chunk_param, start, stop, one_1d, dim_body);
+        replace_node(getitem, curr_chunk);
+        copy_runtime_info({chunk_op, getitem},
```
itikhono, line 131: I would recommend to use `pass::NodeRegistry` and simplify the transformation a bit:
`pass::NodeRegistry rg;
 // use rg.make<...>() instead of std::make_shared<...>() e.g.
auto start = rg.make<v1::Multiply>(...);
copy_runtime_info({chunk_op, getitem}, rg.get());

src/frontends/pytorch/src/transforms/index_loop_getitem_replacer.cpp
```diff
@@ -0,0 +1,154 @@
...
+        if (!chunk_op)
+            return false;
+
+        auto body = loop_op->get_function();
+        size_t chunk_param_idx;
+        bool found_param_idx = false;
+        for (auto input_desc : loop_op->get_input_descriptions()) {
+            if (input_desc->m_input_index == chunk_idx) {
+                chunk_param_idx = input_desc->m_body_parameter_index;
+                found_param_idx = true;
```
itikhono, line 70: `break`; ?

src/frontends/pytorch/src/transforms/index_loop_getitem_replacer.cpp
```diff
@@ -0,0 +1,154 @@
...
+                chunk_op = loop_inputs.at(i).get_node_shared_ptr();
+                chunk_idx = i;
+            }
+        }
+        if (!chunk_op)
+            return false;
+
+        auto body = loop_op->get_function();
+        size_t chunk_param_idx;
+        bool found_param_idx = false;
```
itikhono, line 65: It looks like we can use 
```
std::shared_ptr<Node> chunk_param;
for (auto input_desc : loop_op->get_input_descriptions()) {
    if (input_desc->m_input_index == chunk_idx) {
         chunk_param = body->get_parameters().at(input_desc->m_body_parameter_index);
         break;
     }
}
if (!chunk_param)
    return false;
```

src/frontends/pytorch/src/transforms/index_loop_getitem_replacer.cpp
```diff
@@ -0,0 +1,154 @@
...
+        }
+        if (!found_param_idx)
+            return false;
+
+        auto chunk_param = body->get_parameters().at(chunk_param_idx);
+        auto param_targets = chunk_param->get_output_target_inputs(0);
+        if (param_targets.size() != 1)
+            return false;
+
+        auto getitem = param_targets.begin()->get_node()->shared_from_this();
```
itikhono, line 81: it's not necessary to use `shared_from_this()` here, you can change the check to:
`
if (!ov::as_type<v8::Gather>(getitem))
`

src/frontends/pytorch/src/transforms/index_loop_getitem_replacer.cpp
```diff
@@ -0,0 +1,154 @@
...
+        auto chunk_counter = getitem->input_value(1);
+
+        auto tensor_0 = v0::Constant::create(element::i32, Shape{1}, {0});
+        auto one_1d = v0::Constant::create(element::i32, Shape{1}, {1});
+
+        auto input_shape = std::make_shared<v3::ShapeOf>(chunk_input, element::i32);
+        auto input_dimension = std::make_shared<v8::Gather>(input_shape, dim, tensor_0);
+        auto init_chunk_size = std::make_shared<v1::Divide>(input_dimension, chunks_outside, true);
+
+        // Add 1 if input is not evenly divisible by chunks
```
itikhono, line 106: I believe it's better to use ov Ceil operation here instead of the code below
or you can use Divide with `m_pythondiv="true"`, but in this case you need to use Negative operator twice, e.g.: 
auto out = Divide(input_dimension, Negative(chunks_outside), **true**)
auto res = Negative(out);

itikhono: disccussed offline, Ceiling can cause problems here.

src/frontends/pytorch/src/translate_session.cpp
```diff
@@ -257,6 +294,111 @@ size_t TranslateSession::decode_tensor_name(const Output<Node>& output) {
...
+    auto data_indices_1d = std::make_shared<v1::Reshape>(data_indices, scattering_shape, false);
+    auto to_be_inserted_data_1d = std::make_shared<v1::Reshape>(value, neg_one_1d, false);
+    auto updated_data_1d =
+        std::make_shared<v3::ScatterNDUpdate>(to_insert_data_1d, data_indices_1d, to_be_inserted_data_1d);
+
+    // Reshape to initial shape
+    return std::make_shared<v1::Reshape>(updated_data_1d, input_shape, false);
+}
+
+Output<Node> select_backprop(std::shared_ptr<TorchDecoder> node, Output<Node> select_output, Output<Node> value) {
```
rkazants, line 343: I see that `std::shared_ptr<TorchDecoder> node` is not used at all. Can we remove it?

rkazants: Also, I would recommend to use `const Output<Node>& value` to avoid copies

rkazants: The same comment for other function

src/frontends/pytorch/src/translate_session.cpp
```diff
@@ -257,6 +294,111 @@ size_t TranslateSession::decode_tensor_name(const Output<Node>& output) {
...
+        if (it != backprop_map.end()) {
+            return it->second(node, direct_op_output, value);
+        }
+
+    } catch (std::exception& e) {
+        OPENVINO_DEBUG << "Exception happened during conversion of backprop op: " << node->get_op_type()
+                       << " with schema: " << node->get_schema() << ": " << e.what();
+    }
+    // Create PtFrameworkNode representing unconverted backprop operation
+    return std::make_shared<PtFrameworkNode>(node, OutputVector{value}, 1, true);
```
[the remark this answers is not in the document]
mvafin: These fw node ops are different from regular ones, they have `is_backprop` flag as true, which makes them have different type then regular ops.

src/frontends/pytorch/src/pt_framework_node.hpp
```diff
@@ -14,13 +14,21 @@ class PtFrameworkNode : public ov::op::util::FrameworkNode {
...
         ov::op::util::FrameworkNodeAttrs attrs;
         attrs.set_type_name("PTFrameworkNode");
-        attrs["PtTypeName"] = m_decoder->get_op_type();
-        attrs["PtSchema"] = m_decoder->get_schema();
+        if (is_backprop) {
+            attrs["PtTypeName"] = m_decoder->get_op_type() + "_backprop";
+            attrs["PtSchema"] = "None";
+        } else {
+            attrs["PtTypeName"] = m_decoder->get_op_type();
+            attrs["PtSchema"] = m_decoder->get_schema();
```
rkazants, line 30: may be better to add a comment what is the schema?

rkazants (approved):

eaidova (approved):

PR #17859, opened 2023-06-02
[Snippets] Fixed set for Windows

a-sidorova:
### Details:
 - *Fixed SIGSEGV for Win OS*

### DIFF
--- a/src/common/snippets/src/lowered/loop_manager.cpp
+++ b/src/common/snippets/src/lowered/loop_manager.cpp
@@ -106,8 +106,7 @@ void LinearIR::LoopManager::get_io_loop_ports(LinearIR::constExprIt loop_begin_p
         const auto& expr = *expr_it;
         for (size_t i = 0; i < expr->get_input_count(); ++i) {
             const auto in_port = expr->get_input_port(i);
-            const auto source_port = *in_port.get_connected_ports().begin();
-            const auto& parent_expr = source_port.get_expr();
+            const auto parent_expr = in_port.get_connected_ports().begin()->get_expr();
             if (!ov::is_type<ov::op::v0::Constant>(parent_expr->get_node()) &&
                 std::find(loop_begin_pos, expr_it, parent_expr) == expr_it) {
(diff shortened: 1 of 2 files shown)


src/common/snippets/src/lowered/loop_manager.cpp
```diff
@@ -106,7 +106,8 @@ void LinearIR::LoopManager::get_io_loop_ports(LinearIR::constExprIt loop_begin_p
         const auto& expr = *expr_it;
         for (size_t i = 0; i < expr->get_input_count(); ++i) {
             const auto in_port = expr->get_input_port(i);
-            const auto& parent_expr = in_port.get_connected_ports().begin()->get_expr();
+            const auto source_port = *in_port.get_connected_ports().begin();
+            const auto& parent_expr = source_port.get_expr();
```
dmitry-gorokhov, line 110: Please describe in details whu original code led to segfault on Windows?

a-sidorova: Honestly, I'm faced with this problem several times on ngraph (`get_output_target_inputs` has the same behavior) :smile:

Let's take a look at previous 109 line:
`get_connected_ports()` returns sets of `ExpressionPort` instances (the method **creates new set** and returns it).
Then we take iterator on first element using `begin()` (this element is definitely in the set).
And we call `get_expr` from this first element and takes **reference** of shared pointer. But the set of connected ports is destroyed since it's local variable and isn't saved in this code frame. Thus,  **reference** of shared pointer refs to destroyed memory as I understand. And we get corrupted data:
image
But as I said, this element is definitely in the set:
image


Sure, there is another solution. Save just the copy of shared pointer, not reference:
```suggestion
            const auto parent_expr = in_port.get_connected_ports().begin()->get_expr();
```
for every taste 🙂

a-sidorova: As you say :grin:
Done

dmitry-gorokhov (approved):

PR #17862, opened 2023-06-02
Add static shape adapters to reduce dimension conversion in shape_infer for CPU

praasz:
### Details:
 - Add static shape adapters which hold reference or container of CPU `VectorDims`. This allow to move result shape inference or use them directly without conversion from `StaticShape`.
 - Add new infer function member which use adapter to reduce shape translation cost.
 - Add new return type as `ov::optional`.
 
 Proposal solution of new shape inference on Tile operator. Other ops can be migrated when solution will be accepted.

### Tickets:
 - 111112


src/plugins/intel_cpu/src/utils/shape_inference/static_shape.hpp
```diff
@@ -57,13 +63,310 @@ class StaticShape : public std::vector<StaticDimension>  {
...
+
+public:
+    using ShapeContainer = StaticShapeCon;
+
+    using value_type = StaticDimension;
+    using iterator = typename TDims::iterator;
+    using const_iterator = typename TDims::const_iterator;
+
+    static_assert(std::is_same<dim_type, typename StaticDimension::value_type>::value,
+                  "Static dimension must same type as CPU dimension.");
```
maxnick, line 106:
```suggestion
                  "Static dimension must be of the same type as the CPU dimension.");
```

src/plugins/intel_cpu/src/utils/shape_inference/static_shape.hpp
```diff
@@ -57,13 +63,310 @@ class StaticShape : public std::vector<StaticDimension>  {
...
+    static_assert(std::is_standard_layout<StaticDimension>::value,
+                  "StaticShape must be standard layout to cast on CPU dimension type.");
+    static_assert(sizeof(dim_type) == sizeof(StaticDimension),
+                  "StaticDimension must have same number of bytes as CPU dimension type.");
```
maxnick, line 110:
```suggestion
                  "StaticDimension must have the same number of bytes as the CPU dimension type.");
```

src/plugins/intel_cpu/src/utils/shape_inference/static_shape.hpp
```diff
@@ -57,13 +63,310 @@ class StaticShape : public std::vector<StaticDimension>  {
...
+    size_t size() const noexcept {
+        return m_dims ? m_dims->size() : 0;
+    }
+
+    bool empty() const {
+        return m_dims ? m_dims->empty() : true;
+    }
+
+private:
+    const TDims* m_dims;
```
maxnick, line 349:
```suggestion
    const TDims* m_dims = nullptr;
```

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
```diff
@@ -357,6 +362,45 @@ class ShapeInferTA : public ShapeInferBase {
...
+                           std::back_inserter(result.shapes),
+                           [](StaticShapeCon& s) {
+                               return std::move(reinterpret_cast<StaticShape&&>(*s));
+                           });
+        }
+
+        return result;
+    }
+
+    ov::optional<std::vector<StaticShapeCon>> infer(const std::vector<StaticShapeRef>& input_shapes,
```
mitruska, line 394: Could you elaborate more about advantages of `optional` usage here?
Is there a case when the returned output shape is not needed, like inputs validation only?

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
```diff
@@ -357,6 +362,45 @@ class ShapeInferTA : public ShapeInferBase {
...
+public:
+    using ShapeInferBase::ShapeInferBase;
+
+    IShapeInferCommon::Result infer(const std::vector<StaticShape>& input_shapes,
+                                    const ov::ITensorAccessor& tensor_accessor) override {
+        // Temporary support of StaticShape.
+        auto in_shapes = std::vector<StaticShapeRef>(input_shapes.size());
+        std::transform(input_shapes.begin(), input_shapes.end(), in_shapes.begin(), [](const StaticShape& s) {
+            return StaticShapeRef(reinterpret_cast<const VectorDims&>(s));
+        });
```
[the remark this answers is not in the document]
praasz: Refactored, but as StaticShapeRef holds only pointer there will be no much benefit using the back inserter.

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
praasz: As the `infer` interface function is modified (different input types and return type), I introduced to use optional to replace the `IShapeInferCommon::Result` which do similar thing but using different approach.

This result struct  returned from `infer` in most cases is {output_shapes, success}, the optional do same thing return valid data. But there are some cases like in fallback path or in CPU specific implementation when this status can be `skip` and initialize vector with invalid data (no data).

The benefit of this optional is that `output_shapes` there is dummy vector initialization in case of invalid data and can be done later if required. In case of using `IShapeInferCommon::Result` or std::pair<T, bool> this vector will be always initialized.

mitruska (approved):

maxnick (approved):

jane-intel (approved):

PR #17864, opened 2023-06-02
[TF FE] Workaround for Broadcast/Concat issue with empty tensors

gkrivor:
### Details:
 - Added transformation which removes Concat branches with empty tensors

### Tickets:
 - 107554

### DIFF
--- a/src/frontends/tensorflow/tests/convert_saved_model.cpp
+++ b/src/frontends/tensorflow/tests/convert_saved_model.cpp
@@ -114,3 +114,13 @@ TEST_F(FrontEndConversionWithReferenceTestsF, SavedModelMultipleTensorNames) {
         model_ref = make_shared<Model>(OutputVector{result}, ParameterVector{x});
     }
 }
+
+TEST_F(TransformationTestsF, SavedModelBroadcastIssue) {
+    { model = convert_model("saved_model_broadcast_issue"); }
+    {
+        // create a reference graph
+        auto x = make_shared<Constant>(element::i64, Shape{2,2}, vector<int64_t>{0,0,0,0});
+
+        model_ref = make_shared<Model>(OutputVector{x}, ParameterVector{});
+    }
+}
--- a/src/frontends/tensorflow/tests/convert_saved_model.cpp
+++ b/src/frontends/tensorflow/tests/convert_saved_model.cpp
@@ -115,11 +115,11 @@ TEST_F(FrontEndConversionWithReferenceTestsF, SavedModelMultipleTensorNames) {
     }
 }
 
-TEST_F(TransformationTestsF, SavedModelBroadcastIssue) {
+TEST_F(FrontEndConversionWithReferenceTestsF, SavedModelBroadcastIssue) {
     { model = convert_model("saved_model_broadcast_issue"); }
     {
         // create a reference graph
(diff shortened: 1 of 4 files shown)


src/frontends/tensorflow/tests/convert_saved_model.cpp
```diff
@@ -114,3 +114,13 @@ TEST_F(TransformationTestsF, SavedModelMultipleTensorNames) {
...
     }
 }
+
+TEST_F(TransformationTestsF, SavedModelBroadcastIssue) {
+    { model = convert_model("saved_model_broadcast_issue"); }
+    {
+        // create a reference graph
+        auto x = make_shared<Constant>(element::i64, Shape{2,2}, vector<int64_t>{0,0,0,0});
+
+        model_ref = make_shared<Model>(OutputVector{x}, ParameterVector{});
```
rkazants, line 124: `ConstToResultRemover` does not remove `Const->Result` sub-graph?

gkrivor: Fortunately (?) - no. Because at this point graph isn't Const->Result (it'll be Const->Result after TransposeSinking). Otherwise, I think, it produces wrong model without nodes.

rkazants (approved):

src/frontends/tensorflow/tests/test_models/gen_scripts/generate_saved_model_broadcast_issue.py
```diff
@@ -0,0 +1,18 @@
...
+import tensorflow as tf
+import numpy as np
+
+# Create the graph and model
+tf.compat.v1.reset_default_graph()
+with tf.compat.v1.Session() as sess:
+    tf_x = tf.constant(np.array([[1, 2]], dtype=np.dtype('i8')))
+    tf_cast = tf.raw_ops.Cast(x=-1, DstT=tf.int64)
+    tf_fill = tf.fill([0, 2], tf_cast, name="FillOperation")
+    tf_fill2 = tf.fill([1, 2], tf_cast, name="FillOperation2")
```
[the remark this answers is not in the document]
gkrivor: I suggest to keep it as is because it is important to do a constant folding of exactly defined graph.

Otherwise I may need more time to lookup for another graph which will have exactly reproduce same behavior as I want to verify.

Moreover, it is good test to check correctness of Const->Result too, as you mentioned before :)

src/frontends/tensorflow/tests/convert_saved_model.cpp
```diff
@@ -114,3 +114,13 @@ TEST_F(TransformationTestsF, SavedModelMultipleTensorNames) {
         model_ref = make_shared<Model>(OutputVector{result}, ParameterVector{x});
     }
 }
+
+TEST_F(TransformationTestsF, SavedModelBroadcastIssue) {
```
rkazants, line 118: please use `FrontEndConversionWithReferenceTestsF` and update sources. we no longer use ngraph utils

rkazants:
Need to investigate failing test for TensorList with empty tensor. We intentionally use empty constants for tensor list, need to be careful with the transformation:(

andrei-kochin (approved):

PR #17867, opened 2023-06-02
[CPU][TESTS] Exclude cpplint check for custom test targets

EgorDuplensky:


iefode (dismissed):
Why?

EgorDuplensky:
> Why?

These targets are for development purpose only (under specific compilation flags) and not used in CI or anywhere besides local machines.
From development perspective we do not really need to run cpplint for each of these targets (and dozens of them are actually created).

dmitry-gorokhov (approved):

iefode:
> @iefode Could you please a look one more time?

Honestly, i did not see any real arguments to merge these changes and switch off CPP_lint. Anyway CPU tests are repronsibity area of @dmitry-gorokhov

dmitry-gorokhov:
@EgorDuplensky Please make CI green

PR #17873, opened 2023-06-05
update cpu export/import model behaviors

xufang-lisa:
### Details:
 - *update cpu export/import models behaviors for MT 2.0*

### Tickets:
 - *112149*


src/plugins/intel_cpu/src/plugin.h
```diff
@@ -53,6 +53,10 @@ class Engine : public InferenceEngine::IInferencePlugin {
                             int stream_mode,
                             const bool enable_hyper_thread = true) const;
 
+    void SaveMultiThreadingConfig(std::map<std::string, std::string>& config,
+                                  const std::shared_ptr<ngraph::Function>& ngraphFunc);
+    void LoadMultiThreadingConfig(Config& config, const std::shared_ptr<ov::Model>& function);
```
riverlijunjie, line 58: Why not use snake_case naming style?

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -483,6 +556,11 @@ Engine::LoadExeNetworkImpl(const InferenceEngine::CNNNetwork &network, const std
     // TODO: Clarify the behavior of SetConfig method. Skip eng_config or not?
     Config conf = engConfig;
 
+    // save hints parameters to model rt_info
+    if(is_cpu_map_available()) {
```
riverlijunjie, line 560: I supposed we don't need this condition, do we?

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -272,13 +272,92 @@ void Engine::ApplyPerformanceHints(std::map<std::string, std::string> &config, c
...
+        ngraphFunc->set_rt_info(hints_props, "intel_cpu_hints_config");
+    }
+}
+
+void Engine::LoadMultiThreadingConfig(Config& config, const std::shared_ptr<ov::Model>& function) {
+    if (function->has_rt_info("intel_cpu_hints_config") && !config.perfHintsConfig.ovPerfHint.empty()) {
+        const auto& hints_config = function->get_rt_info<ov::AnyMap>("intel_cpu_hints_config");
+
+        const auto perf_name = std::string(ov::hint::performance_mode.name());
+        const auto it_perf = hints_config.find(perf_name);
```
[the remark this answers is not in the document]
xufang-lisa: thanks, I have updated.

src/plugins/intel_cpu/src/plugin.cpp
xufang-lisa: for threading 1.0, saving hints parameters is performed in function "ApplyPerformanceHints", so I think here this condition is needed and I will add condition for loading caching configs.

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -779,31 +898,32 @@ InferenceEngine::IExecutableNetworkInternal::Ptr Engine::ImportNetwork(std::istr
...
+    const int latency_streams = get_num_numa_nodes();
+    int streams;
+    if (conf.streamExecutorConfig._streams_changed) {
+        streams = conf.streamExecutorConfig._streams;
+    } else if (conf.perfHintsConfig.ovPerfHint == CONFIG_VALUE(LATENCY)) {
+        streams = latency_streams;
+    } else if (conf.perfHintsConfig.ovPerfHint == CONFIG_VALUE(THROUGHPUT)) {
+        streams = 0;
+    } else {
+        streams = conf.streamExecutorConfig._streams == 1 ? 0 : conf.streamExecutorConfig._streams;
```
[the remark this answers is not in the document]
xufang-lisa: thanks, done

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -784,22 +898,29 @@ InferenceEngine::IExecutableNetworkInternal::Ptr Engine::ImportNetwork(std::istr
...
-        if (mode_name == CONFIG_VALUE(LATENCY) || mode_name == CONFIG_VALUE(THROUGHPUT)) {
-            const auto& hints_config = function->get_rt_info<ov::AnyMap>("intel_cpu_hints_config");
-            const auto hints_param_name = mode_name + "_" + std::string(ov::num_streams.name());
-            const auto it = hints_config.find(hints_param_name);
-            if (it != hints_config.end()) {
-                conf.readProperties({{std::string(ov::num_streams.name()), it->second.as<std::string>()}});
-            } else {
-                IE_THROW() << "Cache file doesn't contain precalculated number of streams for mode " << mode_name;
+    if (is_cpu_map_available()) {
+        LoadMultiThreadingConfig(conf, function);
```
dmitry-gorokhov, line 907: What is the technical justification for necessety to store/load all multithreading config during export/import? In other words why should LoadNetwork and ImportNetwork routines have different behavior?

src/plugins/intel_cpu/src/plugin.cpp
wangleis: Will remove "is_cpu_map_available()" after Xiaoxia merge all PRs for MT 2.0 flow.

src/plugins/intel_cpu/src/plugin.cpp
wangleis: The different routines between LoadNetwork and ImportNetwrok is legacy behavior. 

Since we need initial CPU map and schedule streams in both LoadNetwork and ImportNetwork, would check if we can merge them in single routine. TODO JIRA CVS-112767 is created.

wangleis (approved):

src/plugins/intel_cpu/src/plugin.cpp
dmitry-gorokhov: Why do we need this WA then? Are we fixing smt urgent?
Again my point is ther shouldn't be any differences between `LoadNetwork` and `ImportNetwrok` pipelines in terms of multithreading set up.
The only exception is storage of model prefered num streams al long as we are using common MemBandwidthPressureTolerance function which recognize only common ngraph opset, while model loaded from cache contains cpu specific operations. So the solution is to store already computated value within the model. However other multithreading parameters don't have such dependency and I would prefer to avoid its introduction.

peterchen-intel:
Replaced by https://github.com/openvinotoolkit/openvino/pull/18024

PR #17878, opened 2023-06-05
[Snippets] Added tailing processing support for splitted Loops

a-sidorova:
### Details:
 - *Added `ptr_increments` and etc. storage in `LoopInfo`* 
 - *Split `InitLoops` into 2 passes: `InitLoops` (inits `ptr_increments` and etc. storage in `LoopInfo`) and `InsertLoops` (explicitly insert in Linear IR)*
 - *Added `loop_ids` support for `Loops`, `Buffers` and passes that are called after `InitLoops` to have correct `LoopManager` during the whole pipeline*
 - *Updated `InsertTailLoop` logic: added tailing processing support for splitted Loops*
 - *Fixed `clone_with_new_inputs` for `Brgemm` ops - before `count` weren't copied*

### TODO:
- [x] *Added test*

### DIFF
--- a/src/common/snippets/src/lowered/pass/insert_tail_loop.cpp
+++ b/src/common/snippets/src/lowered/pass/insert_tail_loop.cpp
@@ -29,14 +29,6 @@ std::shared_ptr<op::LoopEnd> InsertTailLoop::create_tail_loop(LinearIR& linear_i
     // finalization offsets which are supported by LoopEnd.
     if (need_vector_loop) {
         auto vector_loop_deep_copy = LinearIR::deep_copy_range(vector_begin, vector_end);
-        auto is_par_or_res = [](const ExpressionPtr& expr) {
-            return is_type<ov::op::v0::Parameter>(expr->get_node()) ||
-                   is_type<ov::op::v0::Result>(expr->get_node());
-        };
-        // Note: It's illegal to insert Parameter or Result to the IR, but they can appear inside vector loop
-        //  So we have to remo them before injecting tail loop into linear_ir
-        auto to_erase = std::remove_if(vector_loop_deep_copy.begin(), vector_loop_deep_copy.end(), is_par_or_res);
-        vector_loop_deep_copy.erase(to_erase, vector_loop_deep_copy.end());
         tail_begin = linear_ir.insert(vector_end, vector_loop_deep_copy.begin(), vector_loop_deep_copy.end());
         tail_end = vector_end;
     } else {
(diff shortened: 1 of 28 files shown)


IvanNovoselov (commented):
It's great that we now store tail work amounts in loop info. Nice job!
However, in order to unblock blocking 🙃 we also need to support generic tail insertion for multi-level vector tiles (remember the "brackets" task?). This logic is already implemented on the matmul perf branch. Do you think we can add it to this PR?
We also need to create a couple of representative tests to cover the blocking scenario

src/common/snippets/src/lowered/pass/insert_tail_loop.cpp
```diff
@@ -13,9 +15,74 @@ namespace snippets {
...
+    // finalization offsets which are supported by LoopEnd.
+    if (need_vector_loop) {
+        auto vector_loop_deep_copy = LinearIR::deep_copy_range(vector_begin, vector_end);
+        auto is_par_or_res = [](const ExpressionPtr& expr) {
+            return is_type<ov::op::v0::Parameter>(expr->get_node()) ||
+                   is_type<ov::op::v0::Result>(expr->get_node());
+        };
+        // Note: It's illegal to insert Parameter or Result to the IR, but they can appear inside vector loop
+        //  So we have to remo them before injecting tail loop into linear_ir
+        auto to_erase = std::remove_if(vector_loop_deep_copy.begin(), vector_loop_deep_copy.end(), is_par_or_res);
```
IvanNovoselov, line 38: Why do we need to erase them explicitly? `Results` should be outside of the loop, because there is a special pass for that (`lowered::pass::MoveResultOutOfLoop`), and `Parameters` must be outside by construction (they are always the 1st N nodes in LIR), so `LoopBegin` should be simply inserted after them.
Could you please clarify this point?

src/common/snippets/src/lowered/pass/insert_tail_loop.cpp
```diff
@@ -13,9 +15,74 @@ namespace snippets {
...
+        tail_begin = linear_ir.insert(vector_end, vector_loop_deep_copy.begin(), vector_loop_deep_copy.end());
+        tail_end = vector_end;
+    } else {
+        tail_begin = vector_begin;
+        tail_end = vector_end;
+    }
+
+    const auto& loop_manager = linear_ir.get_loop_manager();
+    const auto current_dim_idx = loop_manager->get_loop_info(vector_loop_end->get_id())->dim_idx;
+    for (auto it = std::next(tail_begin); it != std::prev(tail_end); ++it) {
```
IvanNovoselov, line 49: The logic of this pass is not trivial, so it's worth to add some comments here. Smth like: we need to check the loop body for any enclosed loops that work on the same dimension and rescale their work_amount and increment accordingly

src/common/snippets/src/lowered/pass/insert_tail_loop.cpp
```diff
@@ -13,9 +15,74 @@ namespace snippets {
...
+        const auto& expr = *it;
+        const auto inner_loop_end = ov::as_type_ptr<op::LoopEnd>(expr->get_node());
+        if (!inner_loop_end)
+            continue;
+        const auto loop_info = loop_manager->get_loop_info(inner_loop_end->get_id());
+        if (loop_info->dim_idx != current_dim_idx)
```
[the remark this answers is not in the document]
a-sidorova: As I remember before I had some concerns, but now I found some solution and implemented the logic with parameter in `LoopInfo`. Take a look one more time please. Thanks!

IvanNovoselov (approved):
Nice job! Hopefully it'll work great in the blocking activity

dmitry-gorokhov (approved):

PR #17881, opened 2023-06-05
[GPU] Fuse type conversion only reorders to the prev nodes

andrew-k-park:
### Details:
 - Update reorders that only do type conversion to be fused to previous nodes. i.e. `mvn` and `concatenation`

### Tickets:
 - 110530

### DIFF
--- a/src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
+++ b/src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
@@ -418,6 +418,7 @@ bool layout_optimizer::can_fuse_reorder_to_prev(program_node& prev, reorder_node
     // fusing reorder to the previous node can be done even if it is a dynamic shape case
     if ((prev.is_type<mvn>() || prev.is_type<concatenation>()) &&
         node.as<reorder>().get_primitive()->truncate &&
+        format::is_simple_data_format(fmt_prev) &&
         fmt_prev == fmt_next)
         return true;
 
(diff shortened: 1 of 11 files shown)


andrew-k-park:
There is no performance and accuracy regression for static shape case

e-ddykim (approved):
Looks good to me

src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
```diff
@@ -414,6 +414,14 @@ bool layout_optimizer::can_fuse_reorder(program_node& prev, program_node& next,
 }
 
 bool layout_optimizer::can_fuse_reorder_to_prev(program_node& prev, reorder_node& node, format fmt_prev, format fmt_next) {
+    // Because mvn and concatenation kernel can work cross-layout, if reorder only performs type conversion,
+    // fusing reorder to the previous node can be done even if it is a dynamic shape case
+    if ((prev.is_type<mvn>() || prev.is_type<concatenation>()) &&
+        node.as<reorder>().get_primitive()->truncate &&
```
yeonbok, line 420: Even though the reorder is truncate, it might be also fused with other reorder or reshape. 
How about add a new flag to reorder: is_type_conversion_only ? 
When the reorder is created with truncate, => is_type_conversion_only is to be true. 
And once the reorder is fused with other reshape => set is_type_conversion_only false.
Also we need to check such a behavior with a unittest.

andrew-k-park: Before checking `can_fuse_reorder_to_prev`, in the corresponding logic of `remove_redundant_reorders`, if reorder has a  fused primitive, it will be skipped. Therefore, it seems that there is no need to consider whether reorder is fused with other reorder or reshape in this part.

andrew-k-park: AFAIK, I think the case is different in `shrink reorders chains`. The part related to `shrink reorders chains` in `remove_redundant_reorders` determines whether to optimize out dep reorder or the current reorder. So that part only cares for reorders.

andrew-k-park: As discussed, I added common function to check whether reorder is only type conversion and reinforced condition of this part.

PR #17882, opened 2023-06-05
[PyOV] Restrict `numpy` for unsupported Python versions

p-wysocki:
### Details:
 - pip resolver is sporadically choosing `numpy==1.25.0rc1` for unsupported Python versions, causing random CI fails
 - the PR is currently a draft because of a **discussion in the ticket** - I recommend waiting to see if there are more reproductions before merging, but it's a subject to change and discuss

### Tickets:
 - 112557


rkazants (changes requested):
We wouldn't have these changes. Is pip tool updated on the validation machine?

PR #17885, opened 2023-06-05
[CPU] Enable mlas for FC

zhangYiIntel:
### Details:
 - *Enable MLAS for FullyConnected layer*
-  PR to OV MLAS repo: https://github.com/openvinotoolkit/mlas/pull/1

### Tickets:
 - *112462*


zhangYiIntel:
@ilya-lavrenov Hi this is the formal PR ready for review, you could review here, thanks!

ilya-lavrenov:
> @ilya-lavrenov Hi this is the formal PR ready for review, you could review here, thanks!

Thanks, the comments are left [there](https://github.com/openvinotoolkit/openvino/pull/18076#event-9537225537) can be also be applied for this PR (at least partially)

alvoron:
@zhangYiIntel could you please restore arm related code that was removed in the last commit?
CPU plugin supports ARM platform and we're going to use MLAS on ARM as well

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -294,7 +316,62 @@ void FullyConnected::prepareParams() {
...
+            std::string format = "gemm_mlas_" + std::to_string(N) + "_" + std::to_string(K);
+            if (!edgeMem)
+                IE_THROW() << "Cannot get const weights edgeMem for node " << getName() << ".";
+            auto packedBsize = ov_sgemm_pack_get_size(N, K);
+            MemoryPtr ptr;
+            auto create = [&] () {
+                float* weightPtr = reinterpret_cast<float*>(getParentEdgeAt(1)->getMemoryPtr()->GetPtr());
+                size_t ldb = K;
+                MemoryPtr _ptr = std::make_shared<Memory>(getEngine());
+                _ptr->Create(intel_cpu::DnnlBlockedMemoryDesc(Precision::FP32, intel_cpu::Shape{1, packedBsize / sizeof(float)}));
```
maxnick, line 334: Two questions here:
1. What is the unit of the packedBsize (bytes, elements ?). I assume that elements number is implied. Thus, suggest changing the `ov_sgemm_pack_get_size` call semantics so that it returns the size in bytes.
2. Why tensor of rank 2 with the leading dimension 1 is used?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -294,7 +316,62 @@ void FullyConnected::prepareParams() {
...
+                    ptr = *weightCache->findOrCreate(string_hash, create);
+                } else {
+                    ptr = create();
+                }
+                mlasPackedPtr = ptr;
+            }
+            return ptr;
+        };
+        auto& wgtDims = getParentEdgeAt(WEIGHTS_ID)->getMemoryPtr()->getStaticDims();
+        auto& outDims = getChildEdgeAt(0)->getMemoryPtr()->getStaticDims();
```
maxnick, line 357:
```suggestion
        auto& outDims = dstMemPtr->getStaticDims();
```

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -635,7 +751,23 @@ void FullyConnected::createDescriptor(const std::vector<MemoryDescPtr> &inputDes
 void FullyConnected::initSupportedPrimitiveDescriptors() {
     if (!supportedPrimitiveDescriptors.empty())
         return;
-
+    if (useMlas) {
+        auto dataPrecision = getOriginalInputPrecisionAtPort(0);
+        if (withBiases) {
+            //To Do withBias is not supported for mlas now
```
maxnick, line 757: Is this comment still relevant?

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
```diff
@@ -325,6 +345,14 @@ const std::vector<ShapeRelatedParams> IS2D_nightly = {
     },
 };
 
+std::vector<fusingSpecificParams> filterFusingParams(const std::vector<fusingSpecificParams>& orig) {
+#ifndef OV_CPU_WITH_MLAS
```
maxnick, line 349: It disables fusing for all the matmul tests when compiled with MLAS, doesn't it? May be it does make sense to move MLAS specific instances to a separate folder?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
zhangYiIntel: No, will remove it

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
zhangYiIntel: 1. ov_sgemm_pack_get_size return the bytes of needed memory, let me comment it there
2. I think the memory shape [1, packedBsize ] could be reduced to [packedBsize] 
Thanks!

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
zhangYiIntel: I agree that we should setup specific tests for MLAS. However current MLAS enabled-FC could intercept the FC tests for oneDNN. We don't have a switch to turn off MLAS in original FC tests. Therefore, I need to filter out those tests designed for oneDNN.

zhangYiIntel:
> @zhangYiIntel could you please restore arm related code that was removed in the last commit? CPU plugin supports ARM platform and we're going to use MLAS on ARM as well

I can revert that commit, but current `CMakeLists.txt` doesn't compile those ARM codes. I think it's better to enbale ARM build in another PR. also cc @dmitry-gorokhov

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -294,7 +316,62 @@ void FullyConnected::prepareParams() {
...
+                float* weightPtr = reinterpret_cast<float*>(getParentEdgeAt(1)->getMemoryPtr()->GetPtr());
+                size_t ldb = K;
+                MemoryPtr _ptr = std::make_shared<Memory>(getEngine());
+                _ptr->Create(intel_cpu::DnnlBlockedMemoryDesc(Precision::FP32, intel_cpu::Shape{1, packedBsize / sizeof(float)}));
+                float* prepackedDst = reinterpret_cast<float*>(_ptr->GetPtr());
+                ov_sgemm_pack("T", N, K, ldb, weightPtr, prepackedDst);
+                return _ptr;
+            };
+            if (mlasPackedPtr != nullptr) {
+                ptr = mlasPackedPtr;
```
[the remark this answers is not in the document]
zhangYiIntel: Agree that we move the prepack stage to `createPrimitive` . Thanks

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -261,6 +265,18 @@ void FullyConnected::getSupportedDescriptors() {
 
     inDims = isDynamicNode() ? makeDummyInputDims() : getInputShapeAtPort(DATA_ID).getStaticDims();
     outDims = isDynamicNode() ? makeDummyOutputDims(inDims) : getOutputShapeAtPort(0).getStaticDims();
+#ifdef OV_CPU_WITH_MLAS
```
EgorDuplensky, line 268: Could you please isolate all the new ifdef entries with the separate functions.
Otherwise the code becomes hard to read.

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
maxnick: Agree to leave it as it is as an intermediate solution. But, when we introduce heuristics to select oneDNN/MLAS at runtime (which seems to be inevitable), we will have to turn to the separate test instances solution anyway.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
zhangYiIntel: Great suggestion, will add 2 methods to compute GEMM.
1. WeightPrepack
2. GemmCompute

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
```diff
@@ -182,9 +182,29 @@ TEST_P(MatMulLayerCPUTest, CompareWithRefs) {
...
+                auto primType = getExecValue(ExecGraphInfoSerialization::IMPL_TYPE);
+                if (primType.find("gemm_mlas") != std::string::npos) {
+                    useMlas = true;
+                }
+            }
+        }
+        if (!useMlas) {
+            CheckPluginRelatedResults(compiledModel, cpuNodeType);
+        }
+    }
```
maxnick, line 207: Is not it better to simply set "gemm_mlas" as an expected primitive type in some version of `filterFusingParams` but used only for fullyConnected test instances? The same approach as the special postops filtering you've introduced.

src/plugins/intel_cpu/src/gemm/ov_cpu_gemm.h
```diff
@@ -0,0 +1,98 @@
...
+ */
+void ov_sgemm_pack(const char* transb,
+                   const int64_t N,
+                   const int64_t K,
+                   const int64_t ldb,
+                   const float* src,
+                   float* dst);
+
+/**
+ * @brief  Signle threaded SGEMM
```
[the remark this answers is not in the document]
zhangYiIntel: it's used for mha custom node to calculate the gemm

src/plugins/intel_cpu/src/gemm/ov_cpu_gemm.cpp
```diff
@@ -0,0 +1,99 @@
...
+                      const float alpha,
+                      const float* A,
+                      const int64_t lda,
+                      const float* B,
+                      const int64_t ldb,
+                      const float beta,
+                      float* C,
+                      const int64_t ldc) {
+    // C = alpha*op( A )op( B ) + beta * C
+    std::vector<MLAS_SGEMM_DATA_PARAMS> data(1);
```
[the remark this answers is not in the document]
zhangYiIntel: Allocate on stack is enough, the code here is copied from mlas. let me revise it.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -294,7 +317,20 @@ void FullyConnected::prepareParams() {
...
+    // M, N should be normalized and updated
+    if (useMlas) {
+        outDims = dstMemPtr->getStaticDims();
+        if (outDims.size() == 3) {
+            M = outDims[0] * outDims[1];
+            N = outDims[2];
+        } else {
+            M = outDims[0];
+            N = outDims[1];
+        }
```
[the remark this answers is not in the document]
zhangYiIntel: Great suggestion, previously I did what exactly you say but  later I find the code in `normalizeDims` processes the `dims == 3` only.

src/plugins/intel_cpu/src/gemm/ov_cpu_gemm.h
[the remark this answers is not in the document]
maxnick: Got it, but what if we want to introduce the most unformal call for sgemm, it will be called `ov_sgemm`?
Why `compute` postfix is used? This does make sense in the case of prepacked matrix, since there are two stages: prepack and compute. But this call signature looks more like a typical BLAS `sgemm`. I'm not against keeping this `compute` postfix, I simply want to understand the motivation behind this naming convention.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -308,7 +332,18 @@ void FullyConnected::prepareParams() {
     NodeDesc *selected_pd = getSelectedPrimitiveDescriptor();
     if (selected_pd == nullptr)
         IE_THROW() << "Preferable primitive descriptor is not set for node " << getName() << ".";
-
+#ifdef OV_CPU_WITH_MLAS
+    // M, N should be normalized and updated
+    if (useMlas) {
+        outDims = dstMemPtr->getStaticDims();
+        if (outDims.size() >= 3) {
```
maxnick, line 339:
```suggestion
        if (outDims.size() > 2) {
```

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -308,7 +332,18 @@ void FullyConnected::prepareParams() {
     NodeDesc *selected_pd = getSelectedPrimitiveDescriptor();
     if (selected_pd == nullptr)
         IE_THROW() << "Preferable primitive descriptor is not set for node " << getName() << ".";
-
+#ifdef OV_CPU_WITH_MLAS
+    // M, N should be normalized and updated
```
maxnick, line 336: What about N dimension? It is not updated below. Should we edit the comment?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -922,6 +992,75 @@ bool FullyConnected::useSparseWeightsDecompression() {
...
+void FullyConnected::prepackMLASWeight() {
+    auto prepareMLASWeight = [&](const int64_t N, const int64_t K) {
+        if (!getParentEdgeAt(WEIGHTS_ID)->getParent()->isConstant())
+            IE_THROW() << "Weight input is not const for node " << getName() << ".";
+        auto weightsMem = getParentEdgeAt(WEIGHTS_ID)->getMemoryPtr();
+        std::string format = "gemm_mlas_" + std::to_string(N) + "_" + std::to_string(K);
+        if (!weightsMem)
+            IE_THROW() << "Cannot get const weights edgeMem for node " << getName() << ".";
+        auto packedBsize = ov_sgemm_pack_get_size(N, K);
+        MemoryPtr ptr;
```
maxnick, line 1006: Better move it closer to the place it's actually used.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -922,6 +992,75 @@ bool FullyConnected::useSparseWeightsDecompression() {
...
     return true;
 }
+
+#ifdef OV_CPU_WITH_MLAS
+void FullyConnected::prepackMLASWeight() {
+    auto prepareMLASWeight = [&](const int64_t N, const int64_t K) {
+        if (!getParentEdgeAt(WEIGHTS_ID)->getParent()->isConstant())
+            IE_THROW() << "Weight input is not const for node " << getName() << ".";
+        auto weightsMem = getParentEdgeAt(WEIGHTS_ID)->getMemoryPtr();
+        std::string format = "gemm_mlas_" + std::to_string(N) + "_" + std::to_string(K);
```
maxnick, line 1002: Look like this string is necessary only when the cache is available.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -922,6 +992,75 @@ bool FullyConnected::useSparseWeightsDecompression() {
...
+            ptr = *weightCache->findOrCreate(string_hash, create);
+        } else {
+            ptr = create();
+        }
+        return ptr;
+    };
+    const auto& wgtDims = getParentEdgeAt(WEIGHTS_ID)->getMemoryPtr()->getStaticDims();
+    // Weight is transpoed by MatMulConstTransposesExtraction
+    // K is the IC of weight
+    K = wgtDims[1];
```
maxnick, line 1032: Are we sure that weights always have second rank?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -922,6 +992,75 @@ bool FullyConnected::useSparseWeightsDecompression() {
...
+    N = wgtDims[0];
+
+    if (mlasPackedPtr == nullptr) {
+        mlasPackedPtr = prepareMLASWeight(N, K);
+    }
```
maxnick, line 1037: Suggest removing this check. The method `prepackMLASWeight` is already called once on the create primitive stage, and, it is better to avoid the silent skip.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
zhangYiIntel: N dimension is fixed in this fc case, let me remove this

src/plugins/intel_cpu/src/gemm/ov_cpu_gemm.h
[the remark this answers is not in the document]
zhangYiIntel: I just want to align the postfix the name between FC and Matmul. You are right, it's better to simplify both to `ov_sgemm` and `ov_packed_sgemm`

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
zhangYiIntel: I think so. 
The  `convert_to_matmul_to_fc` transformation has reshaped the weight to `[-1, K]` in Line 138

src/plugins/intel_cpu/src/gemm/ov_cpu_gemm.h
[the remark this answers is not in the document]
maxnick: I think we can use the same name convention for BLAS extensions as in MKL https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-fortran/2023-0/gemm-compute-002.html
That is `ov_sgemm_pack` `ov_sgemm_compute` and `ov_sgemm` when there is no prepacking. However, we probably also need `ov_sgemm` which utilize all available threads, and, though we do not need it right now, but we need to understand how we should distinguish single threaded `ov_sgemm` and multithreaded `ov_sgemm`.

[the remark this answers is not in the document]
zhangYiIntel: Maybe we could introduce an extra API like `set_sgemm_threads_num()` to control the threads before `ov_sgemm`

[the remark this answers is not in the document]
maxnick: Can we simply pass the threads number as a call parameter? With default value 0, which means all the available threads. And the MHA implementation will simply call it with parameter 1. What do you think?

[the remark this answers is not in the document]
zhangYiIntel: That's a good idea, the MLAS internally uses the pointer of ThreadPool as an indicator whether to use single-thread or the threadpool, I think we could use a `bool` to indicate single thread/multi-thread

zhangYiIntel:
@maxnick The mlas library is changed to submodule, but it faces compilation issues now, because the `mlas` library is not public yet.

maxnick:
@EgorDuplensky , do you have any additional comments?

maxnick (approved):

src/plugins/intel_cpu/tests/unit/gemm_api_test.cpp
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <utility>
+#include <gtest/gtest.h>
+#include "gemm/ov_cpu_gemm.h"
+
+
+TEST(GemmTests, getPackedSize) {
```
EgorDuplensky, line 10: Could you please name this test (and maybe add some comment)
To be honest it is hard to understand why we need this test at all

EgorDuplensky (approved):

src/plugins/intel_cpu/tests/unit/gemm_api_test.cpp
zhangYiIntel: Hi, this test is used to check whether the mlas is successfully compiled and integrated into OV

licensing/runtime-third-party-programs.txt
```diff
@@ -1399,3 +1399,11 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 
+-------------------------------------------------------------
```
zhangYiIntel, line 1402: @ilya-lavrenov Could you help review the license statement ?

yssemaev (commented):
Please insert the full text of MIT license as is from https://github.com/microsoft/onnxruntime/blob/main/LICENSE inside the 3rd party file

ilya-lavrenov (approved):

src/plugins/intel_cpu/src/gemm/ov_cpu_gemm.cpp
dmitry-gorokhov, line 1: I would say we need to rearrange file structure:
1.  Extract thread pool into separate file
2. We can add src/mlas folder and put all mlas related stuff there:
2.1.  src/mlas/gemm; src/mlas/thread_pool
2.2. src/mlas should be compiled only in case `OV_CPU_WITH_MLAS` is `ON` (see how its done for ACL inside CMakeLists)
3. Don't see any reason why file should be named with `ov_cpu` prefix - there is no such practice for plugin files naming. Let it be smt like that:
3.1. src/gemm/sgemm.* ; src/mlas/gemm/sgemm.*

src/plugins/intel_cpu/src/gemm/ov_cpu_gemm.cpp
```diff
@@ -0,0 +1,122 @@
...
+void ov_sgemm_pack(const char* transb,
+                   const int64_t N,
+                   const int64_t K,
+                   const int64_t ldb,
+                   const float* src,
+                   float* dst) {
+    MlasGemmPackB(*transb == 'T' ? CblasTrans : CblasNoTrans, N, K, src, ldb, dst);
+}
+
+void ov_sgemm(const char* transa,
```
dmitry-gorokhov, line 61: This is not ov_sgemm. This is exactly mlas_sgemm. Please see my comment above. From my understanding we should place mlas impl inside corresponding folder. 
src/sgemm implementation should serve as dispatcher between different sgemm providers. Basic impl can look like that:
```
void sgemm(...) {
#if defined (OV_CPU_WITH_MLAS)
  mlas_sgemm(...);
#else 
  THROW(); // or fallback on oneDNN/ngraph reference
#endif
}
```

However given the real dispatching is done on FC level we might even avoid introduction of this common interface for now

src/plugins/intel_cpu/CMakeLists.txt
```diff
@@ -64,6 +64,10 @@ if(NOT (AARCH64 OR ARM))
     list(APPEND EXCLUDE_PATHS ${CMAKE_CURRENT_SOURCE_DIR}/src/transformations/cpu_opset/arm/*)
 endif()
 
+if (NOT ENABLE_MLAS)
```
dmitry-gorokhov, line 67: `ENABLE_MLAS` should be explicitly enabled by default, isn't it?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -275,6 +279,21 @@ void FullyConnected::getSupportedDescriptors() {
 
     inDims = isDynamicNode() ? makeDummyInputDims() : getInputShapeAtPort(DATA_ID).getStaticDims();
     outDims = isDynamicNode() ? makeDummyOutputDims(inDims) : getOutputShapeAtPort(0).getStaticDims();
+#ifdef OV_CPU_WITH_MLAS
+    bool isINT8 = one_of(inputDataType, memory::data_type::u8, memory::data_type::s8) && weightsDataType == memory::data_type::s8;
+    // MLAS doesn't support post-ops fusing and BF16. INT8 is not enabled yet
```
dmitry-gorokhov, line 284: Have you tried to measure performance with disabled FC + post_ops fusion? For example common fusing from transformers is FC + Gelu activation. Given Gelu is quite compute intensive, fusing itself might be not benefitial at all. At the same time using mlas for the FC might bring much more perf improvements.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -275,6 +279,21 @@ void FullyConnected::getSupportedDescriptors() {
...
+    if (withBiases) {
+        const auto& biasDims = getInputShapeAtPort(BIAS_ID).getStaticDims();
+        bool isByChannel = biasDims.back() == outDims.back();
+        for (size_t i = 0; i < biasDims.size() - 1; i++) {
+            isByChannel = isByChannel && biasDims[i] == 1;
+        }
+        useMlas = useMlas && isByChannel;
+    }
+#endif
+    if (useMlas) return;
```
dmitry-gorokhov, line 296: Seems like we need to extend CPU debug capabilites to allow change gemm provider via env variable. Please create corresponding if not already.

src/plugins/intel_cpu/CMakeLists.txt
zhangYiIntel: Current, it's not enabled by default, we could open it now

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
zhangYiIntel: Hi Dmitry, previous static shape/dynamic shape perf tests are testing with disabled FC + post_ops fusion. We see regression in some models. The reason here for disable MLAS when post-ops existing is to avoid regression. I think the tradeoff here is to balance normal models and LLM models

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
zhangYiIntel: Should this debug capability be enabled only when `ifdef CPU_DEBUG_CAPS`, or this capability always exists and MLAS is selected by default.

dmitry-gorokhov: Yeah, I expect it should be under CPU_DEBUG_CAPS.
I don't insist to implement the feature in bounds if this PR. Follow-up ticket will be enough.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
dmitry-gorokhov: Ok, thanks for sharing

src/plugins/intel_cpu/src/mlas/sgemm.hpp
```diff
@@ -0,0 +1,103 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+/**
```
dmitry-gorokhov, line 10: Please don't forget to put new functionality under `ov::intel_cpu` namespace.
```
namespace ov {
namespace intel_cpu {
```

src/plugins/intel_cpu/src/mlas/sgemm.hpp
```diff
@@ -0,0 +1,103 @@
...
+ */
+void mlas_sgemm_pack(const char* transb,
+                     const int64_t N,
+                     const int64_t K,
+                     const int64_t ldb,
+                     const float* src,
+                     float* dst);
+
+/**
+ * @brief  SGEMM
```
dmitry-gorokhov, line 38: 1. Do I correctly understand that difference between `mlas_sgemm` and `mlas_sgemm_compute` is B matrix layout? If yes lets directly state in the description that the first one works with planar B matrix, while the second one with prepacked via `mlas_sgemm_pack` function. 
2.  Lets add `thread_num` param handling to `mlas_sgemm_compute` to have aligned API.

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
```diff
@@ -758,7 +782,7 @@ const auto fullyConnectedParams2D_Brgemm_nightly = ::testing::Combine(::testing:
 
 const auto testParams2D_Brgemm_nightly = ::testing::Combine(fullyConnectedParams2D_Brgemm_nightly,
                                              ::testing::Values(MatMulNodeType::FullyConnected),
-                                             ::testing::ValuesIn(fusingParamsSet2D_nightly),
+                                             ::testing::ValuesIn(filterFusingParams(fusingParamsSet2D_nightly)),
```
dmitry-gorokhov, line 785: Have you tried this `nightly` test locally? From my understanding it should pass since it expects brgemm impl to be chosen, but actually mlas will be launched. Is my understadning correct?

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
```diff
@@ -193,13 +192,19 @@ namespace {
...
 
 std::vector<std::map<std::string, std::string>> additionalConfig {
+#ifndef OV_CPU_WITH_MLAS
+    // empty config runs in FP32 and mlas is used
     std::map<std::string, std::string>{/* empty config */},
+#endif
     {{PluginConfigParams::KEY_ENFORCE_BF16, PluginConfigParams::YES}}
 };
 
 std::vector<std::map<std::string, std::string>> filterAdditionalConfig_Brgemm() {
```
dmitry-gorokhov, line 202: General ask for test changes. Can we avoid modification of current test instances and add additional ones to directly cover mlas?  Current changes assumes we throw out most of the brgemm test cases is case the plugin is built with mlas which results in worse coverage. 
So what we need to do is to exclude `empty` and 'bias' fusing configs from `brgemm` instances since such cases are now mapped to mlas impl and add new instances for them.
@zhangYiIntel  what do you think?

src/plugins/intel_cpu/src/mlas/thread_pool.hpp
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <functional>
+#include "mlas.h"
+
+namespace ov {
+namespace cpu {
+class OVThreadPool : public ThreadPool {
```
dmitry-gorokhov, line 14: 1. Please see this comment https://github.com/openvinotoolkit/mlas/pull/1#discussion_r1271873450
2. Maybe better to rename to MlasThreadPool or OVMlasThreadPool

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -275,6 +279,38 @@ void FullyConnected::getSupportedDescriptors() {
...
+        for (size_t i = 0; i < biasDims.size() - 1; i++) {
+            isByChannel = isByChannel && biasDims[i] == 1;
+        }
+        useMlas = useMlas && isByChannel;
+    }
+#endif
+    std::string sgemmType = "MLAS";
+#ifdef CPU_DEBUG_CAPS
+    // Select Sgemm type by ENV MLAS/ONEDNN, MLAS is used by default
+    if (getenv("OV_CPU_GEMM")) {
```
dmitry-gorokhov, line 309: I think we will need to refactor this debug capability using generic primitives priority mechanism, but this indees should be done in separate PR. For now I would ask to do the following:
1. Please rename the variable to make it aligned with other debug options: "OV_CPU_FC_EXEC_TYPE"
2. Lets just set useMlas=false in case the variable not eqaul MLAS instead of usage of additional string outside debug define.

src/bindings/python/tests/test_onnx/test_zoo_models.py
```diff
@@ -130,6 +131,7 @@ def tolerance_map_key_in_model_path(path):
     file_name = path.name
     if path.is_file() and not file_name.startswith("."):
         model = {"model_name": path, "model_file": file_name, "dir": mdir}
+        print(model)
```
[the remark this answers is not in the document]
zhangYiIntel: sorry, let me remove this

src/plugins/intel_cpu/CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ elseif(OV_COMPILER_IS_CLANG)
     ie_add_compiler_flags(-Wno-delete-non-abstract-non-virtual-dtor)
 endif()
 
+# enbale mlas for X86 cpus only
+if (X86_64 OR X86)
```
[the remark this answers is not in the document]
dmitry-gorokhov: UPD: please put this option declaration into directly `src/plugins/intel_cpu/CmakeLists.txt` instead of feature.cmake

src/plugins/intel_cpu/src/mlas/sgemm.hpp
zhangYiIntel: Yes the difference between `mlas_sgemm` and `mlas_sgemm_compute` is whether B matrix is repakced. Let me comment it clearly.

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
zhangYiIntel: Yes I have try the nightly test locally, I have separate that MLAS tests from original tests.

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
zhangYiIntel: Thanks for your proposal, it's more reasonable to separate different tests. I have implement this in latest commit

zhangYiIntel:
@dmitry-gorokhov CI is green now

dmitry-gorokhov (approved):

PR #17906, opened 2023-06-06
[CPU]  MHA extension

eshoguli:
### Details:
 - *Support SmoothQuant method on CPU*

### Tickets:
 - *CVS-110694*

### DIFF
--- a/src/plugins/intel_cpu/tests/functional/subgraph_tests/src/mha.cpp
+++ b/src/plugins/intel_cpu/tests/functional/subgraph_tests/src/mha.cpp
@@ -440,28 +440,17 @@ static std::shared_ptr<ov::Model> initMHAQuantSubgraph1(const std::vector<ov::Pa
     return std::make_shared<ngraph::Function>(results, ngraphParam, "mha");
 }
 
-typedef std::tuple<
-    std::vector<InputShape>,   // Input shapes
-    std::vector<ElementType>,  // Input precisions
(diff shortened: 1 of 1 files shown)


src/plugins/intel_cpu/tests/functional/subgraph_tests/src/mha.cpp
```diff
@@ -562,12 +581,17 @@ std::vector<size_t> patternTypesQuant = {
...
+    false, true
+};
+
 INSTANTIATE_TEST_SUITE_P(smoke_MHAQuant, MHAQuantTest,
                         ::testing::Combine(
                                 ::testing::ValuesIn(static_shapes_to_test_representation(inputShapesQuant)),
                                 ::testing::ValuesIn(inputPrecisionsQuant),
                                 ::testing::ValuesIn(matMulIn0PrecisionsQuant),
                                 ::testing::ValuesIn(patternTypesQuant),
+                                ::testing::ValuesIn(fakeQuantize3Exists),
```
dmitry-gorokhov, line 594: We have tests duplication for case of  `quantPatternType=0` since fakeQuantize3Exists is not actually used in that case.
I would propose either cover new that case via new quant pattern type or separate the instances.

eshoguli: done, thanks

dmitry-gorokhov (approved):

PR #17912, opened 2023-06-07
Allow new shape infer of ShapeOf

steve-y:
### Details:
 - *Fixed to use input shape rank when calculating output layout*
 - *In the case of ShapeOfOp, enable allow_new_shape_infer (deprecated)*
 
### Tickets:
 - *110900*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/shape_of.cpp
+++ b/src/plugins/intel_gpu/src/graph/shape_of.cpp
@@ -24,7 +24,8 @@ layout shape_of_inst::calc_output_layout(shape_of_node const& node, kernel_impl_
         dt = impl_param.get_fused_output_layout().data_type;
     }
 
-    cldnn::tensor out_size{static_cast<tensor::value_type>(prim->output_rank), 1, 1, 1};
+    auto in_shape = impl_param.get_input_layout();
+    cldnn::tensor out_size{static_cast<tensor::value_type>(in_shape.get_rank()), 1, 1, 1};
 
     return layout{dt, format::bfyx, out_size};
 }
--- a/src/plugins/intel_gpu/src/graph/shape_of.cpp
+++ b/src/plugins/intel_gpu/src/graph/shape_of.cpp
@@ -24,8 +24,7 @@ layout shape_of_inst::calc_output_layout(shape_of_node const& node, kernel_impl_
         dt = impl_param.get_fused_output_layout().data_type;
     }
 
-    auto in_shape = impl_param.get_input_layout();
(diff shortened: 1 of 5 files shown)


src/plugins/intel_gpu/src/plugin/program.cpp
```diff
@@ -550,6 +550,11 @@ bool Program::requires_new_shape_infer(const ngraph::Node& op) const {
         return true;
     }
 
+    if (ov::is_type<ov::op::v0::ShapeOf>(&op) ||
+        ov::is_type<ov::op::v3::ShapeOf>(&op)) {
+        return true;
+    }
+
```
yeonbok, line 557: do we still need this?

steve-y: Nope, typo so it is removed.

src/plugins/intel_gpu/src/graph/shape_of.cpp
```diff
@@ -24,7 +24,8 @@ layout shape_of_inst::calc_output_layout(shape_of_node const& node, kernel_impl_
         dt = impl_param.get_fused_output_layout().data_type;
     }
 
-    cldnn::tensor out_size{static_cast<tensor::value_type>(prim->output_rank), 1, 1, 1};
+    auto in_shape = impl_param.get_input_layout();
+    cldnn::tensor out_size{static_cast<tensor::value_type>(in_shape.get_rank()), 1, 1, 1};
```
vladimir-paramuzov, line 28: Why? With old shape infer input rank may be extended, thus produced output shape may be invalid with such change.

steve-y: With the definition of [ShapeOf](https://docs.openvino.ai/2022.3/openvino_docs_ops_shape_ShapeOf_3.html), the outpu number of elements is equal to input tensor rank, but it was output_rank, and it was 1 in this case. So it is fixed with input rank. But as you say that input rank may be extended, then should I use other approach?

vladimir-paramuzov: > then should I use other approach?

Yes, I think you should replace out rank with in rank here: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_gpu/src/plugin/ops/shape_of.cpp#L22
and maybe rename parameter from output_rank to input_rank in the primitive

steve-y: I tried to update as you suggested. And could you check it please? Thanks for your comments.

vladimir-paramuzov (approved):

PR #17917, opened 2023-06-07
Don't constantfold weights in MatMulConstTransposesExtraction transfo…

mateusztabaka:
…rmation

get_constant_from_source for Transpose node calls evaluate method twice which is unnecessary in this case.

Ticket: CVS-105967


src/common/transformations/src/transformations/common_optimizations/matmul_const_transposes_extraction.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2022 Intel Corporation
```
itikhono, line 1: 2022-2023?

mateusztabaka: done

src/common/transformations/src/transformations/common_optimizations/matmul_const_transposes_extraction.cpp
```diff
@@ -33,13 +33,6 @@ ov::pass::MatMulConstTransposesExtraction::MatMulConstTransposesExtraction() {
         std::shared_ptr<Node> transpose = std::make_shared<opset8::Transpose>(
             weights,
             opset8::Constant::create(element::i32, {transpose_order.size()}, transpose_order));
-        if (ov::is_type<opset8::Constant>(weights.get_node())) {
```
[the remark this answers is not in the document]
mateusztabaka: > Aren't we doing the same thing via get_constant_from_source as we do inside ConstFolding transformation

No, get_constant_from_source can call evaluate twice (evalute_lower and evaluate_upper) and this is the case for Transpose. ConstantFolding just calls node->constant_fold(...).

itikhono (approved):

itikhono (approved):

PR #17932, opened 2023-06-07
[TFLite] Custom attribute reading and While operation support

jane-intel:
### Tickets:
 - *CVS-106156*


src/frontends/tensorflow_lite/src/op/while.cpp
```diff
@@ -0,0 +1,105 @@
...
+namespace ov {
+namespace frontend {
+namespace tensorflow_lite {
+namespace op {
+
+OutputVector while_op(const ov::frontend::tensorflow_lite::NodeContext& node) {
+    const auto& decoder = get_decoder(node);
+    int32_t cond_idx = decoder->get_attribute(&tflite::WhileOptions::cond_subgraph_index);
+    int32_t body_idx = decoder->get_attribute(&tflite::WhileOptions::body_subgraph_index);
+
```
rkazants, line 23: the code is quite similar to TF While translator. Can we create `translate_session` class for storing cached previously converted body graphs. I think it makes sense to do because in this case we can use `Frontend` class inplementation for `convert` method that contain logic for telemetry.

I am leaning to align this code with TF implementation.

jane-intel: I have considered it. However, upon the closer look into the TF FE translation session I have encountered lots of logic which is TF specific -- there is no need for it from TFL FE perspective. Remaining code is quite simple. This is why I kept translation logic in InputModel and overloaded [public frontend::NodeContext API|https://github.com/openvinotoolkit/openvino/blob/master/src/frontends/common/include/openvino/frontend/node_context.hpp#L101:L111] that we have right now.

With all that `frontend::tensorflow_lite::NodeContext` is able to trigger translation on demand.

With regards to the alignment -- It would be great to pull all the similar code into common space. Perhaps passing condition / body as `ov::Model` would be a good start. None the less, there are several places where TF implementation has shape / type setting for better TensorList and DT_VARIANT support. These parts are not needed for TFL.

I had an attempt to unify this code, but it broke into several pieces which made it less elegant. So, I left them separate for now. I believe we will return to this unification question when unification along all the frontends would come. But for now I don't see the urgency in it.

src/frontends/tensorflow_lite/include/openvino/frontend/tensorflow_lite/node_context.hpp
```diff
@@ -12,6 +12,9 @@ namespace ov {
 namespace frontend {
 namespace tensorflow_lite {
 
+using SubGraphFuncs = std::vector<std::function<std::shared_ptr<ov::Model>()>>;
+using SubGraphFuncsPtr = std::shared_ptr<SubGraphFuncs>;
```
ilyachur, line 16: Sorry maybe for stupid question, but why do you need to have a shared pointer on vector?
I think we can pass vector by reference...

src/frontends/tensorflow_lite/src/decoder_flatbuffer.cpp
```diff
@@ -87,9 +92,38 @@ std::shared_ptr<ov::frontend::tensorflow_lite::TensorLitePlace> DecoderFlatBuffe
...
+}
+
+ov::Any DecoderFlatBuffer::get_attribute(const std::string& name) const {
+    const auto opts = m_node_def->custom_options();
+    if (opts == nullptr)
+        return {};
+    const flexbuffers::Map& m = flexbuffers::GetRoot(opts->Data(), opts->size()).AsMap();
+    try {
+        return get_value_as_ov_any(m[name]);
+    } catch (...) {
```
ilyachur, line 124: I think you can catch only some specific types of exceptions....

src/frontends/tensorflow_lite/src/frontend.cpp
```diff
@@ -168,6 +168,22 @@ void FrontEnd::translate_graph(const InputModel::Ptr& model,
...
+    auto input_to_ov_model = [&](const std::shared_ptr<ov::frontend::tensorflow_lite::InputModel>& in_model) {
+        auto simple_lambda = [&]() -> std::shared_ptr<ov::Model> {
+            std::shared_ptr<ov::Model> model;
+            if (in_model)
+                translate_graph(in_model, fail_fast, no_conversion, model);
+            return model;
+        };
+        return simple_lambda;
+    };
+    auto submodel_translation_functions = std::make_shared<std::vector<std::function<std::shared_ptr<ov::Model>()>>>();
```
ilyachur, line 181: It can be a vector...

src/frontends/tensorflow_lite/src/op/while.cpp
```diff
@@ -0,0 +1,105 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <openvino/opsets/opset11.hpp>
```
ilyachur, line 5:
```suggestion
#include "openvino/opsets/opset11.hpp"
```

src/frontends/tensorflow_lite/src/decoder_flatbuffer.cpp
jane-intel: It appears the `flexbuffers::Map::operator[]` doesn't throw if it couldn't find the key with requested name -- it returns `flexbuffers::Reference` to a `nullptr` of `FBT_NULL` type which is being interpreted by `get_value_as_ov_any` with no problem.

src/frontends/tensorflow_lite/src/op/while.cpp
rkazants: Yes, there are some TF specific stuff in `TranslateSession` but you can easily skip them and design the common `TranslateSession` class and inherit TFL `TranslateSession` class with your specific methods.

My point is to have the common TranslateSession with common `get_converted_model` in the future. Now you can overload this function if we don't have FW node support in TFL FE.

Now if we decline use of `TranlateSession`, we create additional gap from moving to the common Frontend where we can place telemetry stuff, for example.

src/frontends/tensorflow_lite/src/graph_iterator_flatbuffer.hpp
```diff
@@ -58,6 +61,14 @@ class GraphIteratorFlatBuffer {
...
     /// Return Decoder for the current node that iterator points to
     std::shared_ptr<ov::frontend::tensorflow_lite::DecoderFlatBuffer> get_decoder() const;
+
+    /// \brief Returns the number of sub-graphs that can be enumerated with get_subgraph
+    size_t get_subgraph_size() const;
+
+    /// \brief Returns iterator for a subgraph created on demand
+    /// If there is no query for specific sub-graph iterator shouldn't be created
+    /// idx should be in range 0..get_subgraph_size()-1
+    std::shared_ptr<GraphIteratorFlatBuffer> get_subgraph(const size_t& idx) const;
```
rkazants, line 71: do you have the common cache of already converted sub-graph accessible simultaneously to all sub-graphs?

src/frontends/tensorflow_lite/src/input_model.cpp
```diff
@@ -162,6 +170,17 @@ void InputModel::InputModelTFLiteImpl::load_model() {
...
+    size_t subgraph_size = m_graph_iterator->get_subgraph_size();
+    if (subgraph_size > 1) {
+        m_subgraphs.reserve(subgraph_size);
+        m_subgraphs.push_back(nullptr);  // no main graph
+        for (size_t i = 1; i < subgraph_size; ++i) {
+            m_subgraphs.push_back(
+                std::make_shared<ov::frontend::tensorflow_lite::InputModel>(m_graph_iterator->get_subgraph(i),
+                                                                            m_telemetry));
+        }
+    }
```
rkazants, line 183: I didn't understand. On the on hand, above it is said that sub-graph is created on demand. On the other hand, you create `InputModel` for all sub-graphs

src/frontends/tensorflow_lite/src/input_model.hpp
```diff
@@ -24,6 +24,9 @@ class InputModel : public ov::frontend::InputModel {
     std::map<std::string, std::shared_ptr<ov::frontend::tensorflow_lite::TensorLitePlace>> get_tensor_places() const;
     std::map<std::string, Output<Node>> get_tensor_values() const;
 
+    ////// Subgraph Handling /////
```
rkazants, line 27: strange comment:)

src/frontends/tensorflow_lite/src/frontend.cpp
```diff
@@ -168,6 +168,22 @@ void FrontEnd::translate_graph(const InputModel::Ptr& model,
...
+                translate_graph(in_model, fail_fast, no_conversion, model);
+            return model;
+        };
+        return simple_lambda;
+    };
+    std::vector<std::function<std::shared_ptr<ov::Model>()>> submodel_translation_functions;
+    submodel_translation_functions.reserve(subgraphs_as_input_models.size());
+    for (const auto& subgraph : subgraphs_as_input_models) {
+        submodel_translation_functions.push_back(input_to_ov_model(subgraph));
+    }
```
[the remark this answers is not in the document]
jane-intel: It is not a full translation. It is just a creation of `InputModel`. Here for each sub-graph I make creation functions from `InputModel` to `ov::Model` which are not triggered here -- they are triggered in the While op translator. So sub-graph is only getting translated during `NodeContext::get_subgraph(int)` call.

src/frontends/tensorflow_lite/src/graph_iterator_flatbuffer.cpp
```diff
@@ -25,63 +25,91 @@ GraphIteratorFlatBuffer::GraphIteratorFlatBuffer(const std::string& path) {
...
+    if (is_op) {
+        auto node = m_nodes[node_index].as<const tflite::Operator*>();
+        auto buffers = m_model->buffers();
+
+        std::map<size_t, TensorInfo> input_info = {}, output_info = {};
+        size_t i = 0;
+        for (auto input : *node->inputs()) {
+            if (input == -1)
+                continue;
+            auto buffer = (*buffers)[(*tensors)[input]->buffer()];
```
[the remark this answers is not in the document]
jane-intel: All the asserts are in place by `flatbuffers`.

src/frontends/tensorflow_lite/src/graph_iterator_flatbuffer.cpp
```diff
@@ -25,63 +25,91 @@ GraphIteratorFlatBuffer::GraphIteratorFlatBuffer(const std::string& path) {
...
-    // If we have any m_nodes[node_index]->intermediates() than trigger internal smth? no
-    // put all the info in Decoder as a sub-graph!
+        i = 0;
+        for (auto output : *node->outputs()) {
+            auto buffer = (*buffers)[(*tensors)[output]->buffer()];
+            auto tensor = (*tensors)[output];
+            output_info[i++] = TensorInfo{tensor, buffer};
+        }
+        auto op_codes = m_model->operator_codes();
+        auto operator_code = (*op_codes)[node->opcode_index()];
```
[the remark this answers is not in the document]
jane-intel: All the asserts are in place by flatbuffers.

src/frontends/tensorflow_lite/src/graph_iterator_flatbuffer.hpp
jane-intel: no I don't. there is no such use case in known TFL models

src/frontends/tensorflow_lite/src/input_model.cpp
jane-intel: Commented in previous comment about the logic of this PR.

src/frontends/tensorflow_lite/src/input_model.hpp
jane-intel: I struggle to understand the action which is required here. Could you make your comments structural to make this review productive?

src/frontends/tensorflow_lite/src/decoder_flatbuffer.h
```diff
@@ -68,15 +67,32 @@ class DecoderFlatBuffer : public ov::frontend::DecoderBase {
...
     std::shared_ptr<ov::frontend::tensorflow_lite::TensorLitePlace> decode_tensor(
-            const ov::frontend::tensorflow_lite::TensorInfo& tensor_info, const InputModel& model) const;
+            const ov::frontend::tensorflow_lite::TensorInfo& tensor_info, const ov::frontend::InputModel& model) const;
 
     const tflite::Operator* m_node_def;
     std::string m_type, m_name;
     std::map<size_t, ov::frontend::tensorflow_lite::TensorInfo> m_input_info, m_output_info;
 };
 
+class DecoderFlatBufferTensors : public DecoderFlatBuffer {
```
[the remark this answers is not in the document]
jane-intel: I could handle this in the next PR.

src/frontends/tensorflow_lite/src/graph_iterator_flatbuffer.cpp
```diff
@@ -25,63 +25,91 @@ GraphIteratorFlatBuffer::GraphIteratorFlatBuffer(const std::string& path) {
...
+    auto any_item = m_nodes[node_index];
+    bool is_op = any_item.is<const tflite::Operator*>();
+    FRONT_END_GENERAL_CHECK(is_op || any_item.is<int32_t>());
+    auto tensors = m_graph->tensors();
+
+    if (is_op) {
+        auto node = m_nodes[node_index].as<const tflite::Operator*>();
+        auto buffers = m_model->buffers();
+
+        std::map<size_t, TensorInfo> input_info = {}, output_info = {};
```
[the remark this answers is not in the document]
jane-intel: This isn't a change of current PR. Getting decoder for operations was supported before. So I will change it in the next PR as it doesn't break the behavior

src/frontends/tensorflow_lite/src/graph_iterator_flatbuffer.cpp
```diff
@@ -25,63 +25,91 @@ GraphIteratorFlatBuffer::GraphIteratorFlatBuffer(const std::string& path) {
...
+            auto tensor = (*tensors)[output];
+            output_info[i++] = TensorInfo{tensor, buffer};
+        }
+        auto op_codes = m_model->operator_codes();
+        auto operator_code = (*op_codes)[node->opcode_index()];
+        std::string type;
+        if (operator_code->deprecated_builtin_code() <
+            tflite::BuiltinOperator::BuiltinOperator_PLACEHOLDER_FOR_GREATER_OP_CODES) {
+            type = tflite::EnumNamesBuiltinOperator()[operator_code->deprecated_builtin_code()];
+        } else {
```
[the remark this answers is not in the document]
jane-intel: This code was committed some time ago. Since this isn't critical for the actual code outcome -- I will add comment in the next PR

src/frontends/tensorflow_lite/src/graph_iterator_flatbuffer.cpp
```diff
@@ -25,63 +25,91 @@ GraphIteratorFlatBuffer::GraphIteratorFlatBuffer(const std::string& path) {
...
+            type = tflite::EnumNamesBuiltinOperator()[operator_code->builtin_code()];
+        }
+        if (type == "CUSTOM") {
```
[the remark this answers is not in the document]
jane-intel: I find it quite self-explanatory, however if there is a need to explain it even more -- I will add comment in the next PR

src/frontends/tensorflow_lite/src/graph_iterator_flatbuffer.cpp
```diff
@@ -25,63 +25,91 @@ GraphIteratorFlatBuffer::GraphIteratorFlatBuffer(const std::string& path) {
...
-    auto op_codes = m_model->operator_codes();
-    auto operator_code = (*op_codes)[m_nodes[node_index]->opcode_index()];
-    std::string type;
-    if (operator_code->deprecated_builtin_code() <
-        tflite::BuiltinOperator::BuiltinOperator_PLACEHOLDER_FOR_GREATER_OP_CODES) {
-        type = tflite::EnumNamesBuiltinOperator()[operator_code->deprecated_builtin_code()];
-    } else {
-        type = tflite::EnumNamesBuiltinOperator()[operator_code->builtin_code()];
+            output_it == outputs->end() ? -1 : static_cast<int64_t>(std::distance(outputs->begin(), output_it));
+        return std::make_shared<DecoderFlatBufferTensors>(info, input_idx, output_idx);
```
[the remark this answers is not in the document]
jane-intel: here we handle input and/or output tensors. we explicitly trigger them only for these types of tensors. so having -1 is okay

rkazants (approved):

ilyachur (approved):

PR #17936, opened 2023-06-07
do not change names while replacing FQ nodes

pavel-esir:
### Details:
When we try to get outputs from the intermediate node FakeQuantize_6 we end up having error on CPU plugin
```
File ".../site-packages/openvino/runtime/ie_api.py", line 394, in compile_model
    super().compile_model(model, {} if config is None else config),
RuntimeError: Check 'false' failed at src/inference/src/core.cpp:114:
Cannot find blob with name: FakeQuantize_6
```
This happens because original node was replaced with a node with different name.
Before ngraph transformations
image

After ngraph transformations right before passing it to the plugin
image

### Solution:
Copy name of the old node to the new without any additions.

### Tickets:
 - xxx-110363


src/common/low_precision_transformations/src/mat_mul.cpp
```diff
@@ -168,7 +168,7 @@ bool MatMulTransformation::transform(TransformationContext &context, ngraph::pat
         ov::op::TemporaryReplaceOutputType(parent, deqPrecision).get(),
         ov::op::TemporaryReplaceOutputType(newMulConst, deqPrecision).get());
 
-    newMultiply->set_friendly_name(newMatMul->get_friendly_name() + "/DequantizationMultiply");
+    newMultiply->set_friendly_name(newMatMul->get_friendly_name());
```
itikhono, line 171: I think this is an incorrect change.
We should set the friendly_name from old matmul node  to this new multiply
But as I can see `updateOutput(context, newMultiply, newMatMul);` on line 176 already do it
maybe for [network_helper.cpp](https://github.com/openvinotoolkit/openvino/pull/17936/files#diff-030828c31ef1590e501e4fd5c4e58cf49e3520cd5a18de79e88fa1cadf46bfb9) we should also use updateOutput function

pavel-esir: Agreed. Change on this line is redundant because `updateOutput` already does such renaming. 

But with reusing `updateOutputs` in network_helper.cpp there are problems, we don't have variable `TransformationContext &context` to call it from `network_helper.cpp`. Therefore have added manual renaming to `network_helper.cpp` if there are `Results`. Please take a look.
 
Also @eshoguli can you confirm that the current solution is ok?

pavel-esir:
> Please, check and fix LayerTransformation::updateOutput, thanks!

Done. Since i reverted changes in `low_precision_transformations/src/mat_mul.cpp` now tests are green

eshoguli (approved):
LGTM

src/common/low_precision_transformations/src/mat_mul.cpp
eshoguli: `decomposeFakeQuantize` is used by some transformation. In the transformation: can you check if `updateOutput` is used after FQ decomposition?

src/common/low_precision_transformations/src/network_helper.cpp
```diff
@@ -1022,6 +1022,16 @@ std::tuple<std::shared_ptr<Node>, std::shared_ptr<Node>> NetworkHelper::decompos
             ov::op::TemporaryReplaceOutputType(sub == nullptr ? (convert2 == nullptr ? newFQ : convert2) : sub, element::f32).get(),
             ov::op::TemporaryReplaceOutputType(scale, element::f32).get());
     dequantize->set_friendly_name(newFQ->get_friendly_name() + "/DequantizationMultiply");
+
+    // return original name if there is a Result on the output
+    for (const auto& input : dequantize->output(0).get_target_inputs()) {
+        if (ov::is_type<ov::opset1::Result>(input.get_node())) {
+            const std::string originalName = newFQ->get_friendly_name();
+            newFQ->set_friendly_name(originalName + "original");
```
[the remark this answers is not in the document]
pavel-esir: applied some comments from Edward, this code has been removed and renaming suffix now matches other naming

src/common/low_precision_transformations/src/mat_mul.cpp
pavel-esir: rewritten with `updateOutput`

eshoguli (approved):
much better, thanks! Let's merge

PR #17944, opened 2023-06-08
[DOCS] Adding Datumaro document into OV Ecosystems

wonjuleee:
Adding Datumaro document into OV Ecosystems.


kblaszczak-intel (approved):
That's a nice addition!

PR #17973, opened 2023-06-09
[CPU] [ARM] Enable MVN SLT tests on ARM

alvoron:


src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/mvn.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+std::vector<ElementType> inpPrc = {
+        ElementType::bf16
+};
+std::vector<ElementType> outPrc = {
+        ElementType::bf16
+};
+
+std::vector<CPUSpecificParams> cpuParams_4D = {
+        CPUSpecificParams({nhwc}, {nhwc}, {}, {}),
```
EgorDuplensky, line 34: It seems unclear why nhwc is not supported by arm / acl
Is it so? If yes, is it by design or we just haven't finished the executor implementation?

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/common/mvn.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+const std::vector<ov::Shape> inputShapesStatic_4D = {
+        {1, 7, 3, 5},
+        {1, 15, 9, 5},
+        {4, 41, 6, 9},
+        // cover channel case 4*16*2+16+3=147
+        {1, 147, 2, 2}
+};
+
+const std::vector<ov::Shape> inputShapesStatic_5D = {
```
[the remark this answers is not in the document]
alvoron: Fixed

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/common/mvn.cpp
```diff
@@ -0,0 +1,204 @@
...
+       ::testing::ValuesIn(outPrc));
+
+INSTANTIATE_TEST_SUITE_P(smoke_CompareWithRefs_Mvn3D, MvnLayerCPUTest, Mvn3D, MvnLayerCPUTest::getTestCaseName);
+
+const auto Mvn4D = ::testing::Combine(
+       ::testing::Combine(
+               ::testing::ValuesIn(inputShapes_4D()),
+               ::testing::Values(ElementType::f32),
+               ::testing::ValuesIn(emptyReductionAxes()),
+               ::testing::Values(true),
```
EgorDuplensky, line 68: could you please add a comment here, like:
::testing::Values(true), // "initAcrossChannels = false" is not supported (or fails by accuracy?) for NHWC tensor.
Please also add the similar comment to the code where the check happens.
Also, it seems the check in isSupported method only cares about NHWC layout.
Here we are not testing the combination HCHW + initAcrossChannels = false for some reason.

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/common/mvn.cpp
```diff
@@ -0,0 +1,204 @@
...
+       ::testing::ValuesIn(outPrc));
+
+INSTANTIATE_TEST_SUITE_P(smoke_CompareWithRefs_Mvn4D, MvnLayerCPUTest, Mvn4D, MvnLayerCPUTest::getTestCaseName);
+
+const auto Mvn5D = ::testing::Combine(
+       ::testing::Combine(
+               ::testing::ValuesIn(inputShapes_5D()),
+               ::testing::Values(ElementType::f32),
+               ::testing::ValuesIn(emptyReductionAxes()),
+               ::testing::Values(true),
```
EgorDuplensky, line 83: Is the limitation applicable for the 5D as well?
The check in the source code seems to be applicable only for 4D tensor.

alvoron: ACL really limits MVN to handle up to 4 rank.
Dmitry suggested to implement WA: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/nodes/executors/acl/acl_mvn.cpp#L22-L47

ACL has 2D input shape [Y, X] and actual normalization is done over X axis, while Y serves as "batch". We can translate any 2+ dimension MVN shape into ACL representation using the following rules (example for 4D):

NCHW layout, across_channels == false; : Y = NC, X = HW
NCHW layout, across_channels == true; : Y = N, X = CHW
NHWC layout, across_channels == false; : unsupported
NHWC layout, across_channels == true; : Y = N, X = HWC

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/common/mvn.cpp
alvoron: Added missed case "HCHW + initAcrossChannels = false" for 4D and 5D

EgorDuplensky: As discussed, proposed to leave this case in common and write test-local check to skip the test for arm in case this combination appeared.

EgorDuplensky (approved):

EgorDuplensky:
@maxnick Could you please proceed with the review

src/plugins/intel_cpu/tests/functional/single_layer_tests/classes/mvn.cpp
```diff
@@ -0,0 +1,337 @@
...
+    InputShape inputShapes;
+    ElementType netPrecision;
+    ngraph::AxisSet axes;
+    bool normalizeVariance;
+    double eps;
+    std::tie(inputShapes, netPrecision, axes, acrossChanels, normalizeVariance, eps) = basicParamsSet;
+
+    if (!isSupportedTestCase()) {
+        GTEST_SKIP() << "Skip MVN test since such combination of parameters is not supported." << std::endl;
+    }
```
maxnick, line 89: It looks like we have some cases not supported on ARM. But, isn't it enough to simply not instance them in the dedicated tests set? Why do we need to check the condition here?
Also, if ARM have to support these configurations, don't we have to make a dedicated ticket?

alvoron: It makes common tests logic more complex, since we need to split one common test case into common test case and x64 test case (to exclude case not supported by arm).
The issue is ACL has quite a lot of limitations, so some limitations are combined in some test cases. It makes "split" logic even more complex and its maintenance would be complicated.

maxnick: Ok. Then we need a follow up ticket to refactor this solution later.

maxnick (approved):

PR #17975, opened 2023-06-09
[Snippets] Refactored LoopManager

a-sidorova:
### Details:
 - *Hided all methods with explicit `Loop ID` work inside private section of `LoopManager`*
 - *Added methods `update_loop_ports`*
 - *Added methods `fuse_loops` that unite two `LoopInfo` into one and mark expression by new `Loop ID`*
 - *Added method `expression_replacement`. The method should be removed after `Softmax Decomposition` refactoring (moving to data flow)*
 - *Fixed `FuseLoops` pass. Inner Loops can contain ports which are ports of outer Loops as well. When we fuse and move these inner loops, we can corrupt the sort of LoopPorts of outer Loops. Because of that we have to sort outer Loop ports if we moved loops*

### Tickets:
 - *112195*

### DIFF
--- a/src/common/snippets/src/lowered/loop_manager.cpp
+++ b/src/common/snippets/src/lowered/loop_manager.cpp
@@ -62,6 +62,13 @@ LoopInfoPtr LinearIR::LoopManager::get_loop_info(size_t index) const {
     return it->second;
 }
 
+std::vector<size_t> LinearIR::LoopManager::get_outer_expr_loops(const ExpressionPtr& expr, size_t loop_id) {
+    const auto loop_ids = expr->get_loop_ids();
+    const auto it = std::find(loop_ids.cbegin(), loop_ids.cend(), loop_id);
+    OPENVINO_ASSERT(it != loop_ids.cend(), "Loop ID hasn't been found");
+    return std::vector<size_t>(loop_ids.cbegin(), it);
+}
+
 void LinearIR::LoopManager::get_loop_bounds(const LinearIR &linear_ir,
                                             size_t loop_id,
                                             LinearIR::constExprIt &loop_begin_pos,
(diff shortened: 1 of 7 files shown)


src/common/snippets/src/lowered/loop_manager.cpp
```diff
@@ -207,7 +214,169 @@ void LinearIR::LoopManager::mark_loop(LinearIR::constExprIt loop_begin_pos,
...
+    // In some cases actual ExpressionPort may not be LoopPort. We shouldn't throw exception here since ExpressionPort is not strong condition as LoopPort
+    // For example, not all inner loop ports are ports of outer loops
+    if (port_it == ports.end())
+        return;
+
+    // to save other parameters except expression port
+    std::vector<LoopPort> target_loop_ports(target_ports.size(), *port_it);
+    std::transform(target_loop_ports.begin(), target_loop_ports.end(), target_ports.begin(), target_loop_ports.begin(),
+                   [](const LoopPort& loop_port, const ExpressionPort& expr_port) {
+                       LoopPort copy = loop_port;  // to save loop port parameters
```
IvanNovoselov, line 318: Minor: we can pass the first arg by value, so it would be copied automatically. This will save you a couple code lines.

src/common/snippets/src/lowered/loop_manager.cpp
```diff
@@ -207,7 +214,169 @@ void LinearIR::LoopManager::mark_loop(LinearIR::constExprIt loop_begin_pos,
...
+    };
+    auto loop_info = get_loop_info(loop_id);
+    const auto& loop_entries = loop_info->entry_points;
+    const auto& loop_exits = loop_info->exit_points;
+    std::vector<LoopPort> entries, exits;
+    entries.reserve(loop_entries.size());
+    exits.reserve(loop_exits.size());
+    for (auto it = loop_begin_pos; it != loop_end_pos; ++it) {
+        const auto& expr = *it;
+        push(loop_entries, entries, expr);
```
IvanNovoselov, line 374: Minor: It looks like this sort complexity is O(n2) in the worst case, where n is the number of expressions in the loop. This could be relaxed to O(n*log(n)) if we had float expression ids.

IvanNovoselov (approved):

dmitry-gorokhov (approved):

PR #17976, opened 2023-06-09
[GPU] Fix in-order queue synchronization issue related to OCL/OneDNN impls interaction with CPU impls

sshlyapn:
### Details:
 - Aligned behavior of `ocl_stream::enqueue_marker()` method with OpenCL specification (marker with empty dependencies vector waits for all previous tasks in the queue)
 - Fixed events collection in `network::execute_primitive()` method. Previously we stored events only for OOO queue and profiling mode, despite the fact that we need events for synchronization between GPU and CPU impls in in-order queue
 - Added dependency events preparation in `primitive_inst::execute()` method for CPU implementations and optimized out implementation if it has CPU users (previously it was done only for OOO queue)
 - Return marker from `onednn_primitive::execute()` method if the next operation require synchronization
 - Related tests added
 
### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/common/loop.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/common/loop.cpp
@@ -110,6 +110,7 @@ struct loop_impl : typed_primitive_impl<loop> {
             }
         }
 
+        std::vector<event::ptr> all_events;
         std::vector<event::ptr> loop_carried_dep(events.begin(), events.end());
         int64_t current_iteration_idx = 0;
         while (current_iteration_idx < trip_count && execution_condition) {
@@ -145,6 +146,15 @@ struct loop_impl : typed_primitive_impl<loop> {
                 loop_carried_dep.emplace_back(body_event);
             }
 
+            // Collect output events for waiting for all iterations finishing
+            for (auto& out : body_network->get_outputs()) {
+                auto output_id = out->id();
+                if (body_network->has_event(output_id)) {
+                    auto output_event = body_network->get_primitive_event(output_id);
+                    all_events.push_back(output_event);
+                }
+            }
+
             //TODO: execution_condition is prepared as they are presented in the
             //      ngraph opset document for loop operation.
             // However they are not being used yet and only TensorIterator which
@@ -157,7 +167,9 @@ struct loop_impl : typed_primitive_impl<loop> {
             ++current_iteration_idx;
         }
 
-        body_network->reset_execution();
+        // Reset network and wait for all collected events
+        body_network->reset_execution(false);
+        stream.wait_for_events(all_events);
 
         // Concatenate sliced output to the outer network
         for (size_t i = 0; i < concatenated_output_mem_mappings.size(); ++i) {
--- a/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
+++ b/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
@@ -460,16 +460,6 @@ protected:
 
     void init_kernels(const kernels_cache&, const kernel_impl_params&) override { }
 
-    event::ptr aggregate_events(const std::vector<event::ptr>& events, stream& stream, bool group = false, bool is_output = false) const {
-        if (events.size() == 1 && !is_output)
-            return events[0];
-
-        if (group && !is_output)
-            return stream.group_events(events);
-
-        return stream.enqueue_marker(events, is_output);
-    }
-
     void set_arguments_impl(typed_primitive_inst<PType>& instance) override {
         if (instance.can_be_optimized())
             return;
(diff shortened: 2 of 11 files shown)


src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
```diff
@@ -499,6 +489,11 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
                 }
                 throw;    // rethrowing dnnl::error if not out_of_memory
             }
+
+            // Enqueue marker with empty events wait list (which will trigger wait for all previously enqueued tasks) and
+            // return it as oneDNN primitive's event as it is a single option for proper synchronization
+            if (instance.is_output_event())
+                event = stream.enqueue_marker({});
```
isanghao, line 496: Is it to ensure that all onednn kernels are done before CPU layer execution, not for network output? If so, could you describe that on comment? I think it is unclear why it is necessary..

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -803,7 +805,8 @@ primitive_inst::primitive_inst(network& network, program_node const& node, bool
     , _fused_mem_offset((_fused_mem_count > 0 && node.has_fused_dep()) ? node.get_first_fused_dep_idx() : 0)
     , _can_be_optimized(node.can_be_optimized())
     , _can_share_buffer(node.can_share_buffer())
-    , _is_constant(node.is_constant()) {
+    , _is_constant(node.is_constant())
+    , _is_output_event(is_any_user_cpu(node.get_users()) || node.is_output()) {
```
isanghao, line 809: I'm afraid this `_is_output_event` might mislead the meaning. What about more descriptive name, such as `needs_to_wait`? (not sure whether this is a proper name)

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
sshlyapn: Yes, in most cases we need it to ensure that onednn kernel is finished before CPU layer execution, but also event with marker will be created if onednn primitive will be network's output (it may be used in loop primitive for networks synchronization, and maybe in some other cases I'm not aware of...)
And I think we can try to use it for get_memory() or reset_network() methods for in_order queue - in such a case we will avoid clFinish() call at all which theoretically may improve performance a bit (if I'm not mistaken, clFinish() call flushes caches, and maybe has some other side-effects)

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
sshlyapn: Yeah, agree, what do you think about `_needs_event` or maybe even `_needs_completion_event`?

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
sshlyapn: @isanghao, I have updated comment and mentioned cases when this event is used

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
isanghao: looks good to me.

src/plugins/intel_gpu/src/graph/impls/common/loop.cpp
```diff
@@ -157,7 +167,9 @@ struct loop_impl : typed_primitive_impl<loop> {
             ++current_iteration_idx;
         }
 
-        body_network->reset_execution();
+        // Reset network and wait for all collected events
+        body_network->reset_execution(false);
```
yeonbok, line 171: shouldn't we set this true or do clfinish here for in order queue case?

yeonbok: ah.. seems that you created an event for the output primitives.

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
yeonbok: Hmm basically I agree with removing clfinish as much as possible. 
However I have a curiosity about "flush cache" you mentioned, becuase 
previously I experienced cache inconsistency b/w CPU and dGPU on host memory even though I did clFinish.

yeonbok (approved):

isanghao:
Performance looks good: 
--this /home/mingyuki/log/pr/2023-06-13--21-31-PR17976.report --ref /home/mingyuki/log/pr/2023-06-13--18-06-PR18003.report

isanghao (approved):

PR #17978, opened 2023-06-09
[GPU] Fix dynamic padding processing of static dimension

sshlyapn:
### Details:
 - This PR fixes "[GPU] Can't choose implementation for ..." error at compile graph stage caused by calling `Pad::Total()` of dynamic padding (affected models: chatGlm, dolly quantized version)

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_gpu/src/kernel_selector/jitter.cpp
+++ b/src/plugins/intel_gpu/src/kernel_selector/jitter.cpp
@@ -190,14 +190,14 @@ std::string toCodeString(size_t val) {
 std::string toCodeString(const Tensor::Dim& dim, size_t offset, bool padded, bool pad_is_dynamic, size_t pad_offset) {
     std::string pad_str = "";
     if (padded) {
-        if (dim.pad.is_dynamic) {
+        if (pad_is_dynamic) {
             pad_str = " + (shape_info[" + std::to_string(pad_offset) + "] + shape_info[" +
                       std::to_string(pad_offset + 1) + "])";
         } else {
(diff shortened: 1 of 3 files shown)


sshlyapn:
Hi @yeonbok , could you please take a look on these changes?

src/plugins/intel_gpu/src/kernel_selector/jitter.cpp
```diff
@@ -190,14 +190,14 @@ std::string toCodeString(size_t val) {
 std::string toCodeString(const Tensor::Dim& dim, size_t offset, bool padded, bool pad_is_dynamic, size_t pad_offset) {
     std::string pad_str = "";
     if (padded) {
-        if (dim.pad.is_dynamic) {
+        if (pad_is_dynamic) {
```
yeonbok, line 193: Could you tell me what was the situaton? 
dim.pad.is_dynamic was not set as dynamic but the dim.pad was actually dynamic?

sshlyapn: @yeonbok hmm, the issue happens for concat's dependency when dimension is static (the dimension on which we apply concat) and in the same time it has assigned dynamic padding (because of concat optimization). In that case during jitter code generation at line 200 we will chose else branch and try to obtain `dim.pad.Total()` for dynamic padding and get error
You can try to reproduce this situation with test from my PR

yeonbok: Thanks! I will check :)

yeonbok (approved):

PR #17988, opened 2023-06-12
[CPU]Shape agnostic jit kernel for MVN

chenhu-wang:
### Details:
 - *shape agnostic jit kernel for MVN*

### Tickets:
 - *112476*


chenhu-wang:
> @chenhu-wang Is it ready for review?

@dmitry-gorokhov Will ping you after fully updated all code path, maybe tomorrow.

chenhu-wang:
> @chenhu-wang Is it ready for review

No, maybe tomorrow.

> > @chenhu-wang Is it ready for review?
> 
> @dmitry-gorokhov Will ping you after fully updated all code path, maybe tomorrow.

@dmitry-gorokhov , could you please start review or assign. There are 3 post-ops fusion failure in test, but do not block review, I am fixing it at meanwhile.

chenhu-wang:
> > @chenhu-wang Is it ready for review
> 
> No, maybe tomorrow.
> 
> > > @chenhu-wang Is it ready for review?
> > 
> > 
> > @dmitry-gorokhov Will ping you after fully updated all code path, maybe tomorrow.
> 
> @dmitry-gorokhov , could you please start review or assign. There are 3 post-ops fusion failure in test, but do not block review, I am fixing it at meanwhile.

@dmitry-gorokhov All the failure is fixed. Could you please review?

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -108,14 +102,15 @@ struct jit_uni_mvn_mean_variance_kernel_f32 : public jit_uni_mvn_mean_variance_k
...
     void generate() override {
-        tail_step = jcp_.layout == MVNLayoutType::mvn_planar ? (jcp_.D * jcp_.H * jcp_.W) - ((jcp_.D * jcp_.H * jcp_.W) / vector_step) * vector_step :
-                   jcp_.C - (jcp_.C / vector_step) * vector_step;
-
         Precision dst_prc = isFloatCompatible(jcp_.src_prc) ? Precision::FP32 : Precision::I32;
         load_vector_emitter.reset(new jit_load_emitter(this, isa, jcp_.src_prc, dst_prc, vector_step));
-        load_tail_emitter.reset(new jit_load_emitter(this, isa, jcp_.src_prc, dst_prc, tail_step));
-        load_tail_with_fill_emitter.reset(new jit_load_emitter(this, isa, jcp_.src_prc, dst_prc, tail_step, Precision::FP32, true, "zero"));
-        load_scalar_with_fill_emitter.reset(new jit_load_emitter(this, isa, jcp_.src_prc, dst_prc, scalar_step, Precision::FP32, true, "zero"));
+        load_tail8_emitter.reset(new jit_load_emitter(this, isa, jcp_.src_prc, dst_prc, 8));
```
dmitry-gorokhov, line 107: I believe you can use static array to store all required emitters of the same type.

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -333,97 +244,617 @@ struct jit_uni_mvn_mean_variance_kernel_f32 : public jit_uni_mvn_mean_variance_k
...
+
+            Label tail_blk8_label;
+            Label tail_blk8_exit_label;
+            Label tail_blk4_label;
+            Label tail_blk4_exit_label;
+            Label tail_blk2_label;
+            Label tail_blk2_exit_label;
+            Label tail_blk1_label;
+            Label tail_blk1_exit_label;
+            L(tail_blk8_label);
```
dmitry-gorokhov, line 564: General ask for the whole PR: need to get rid of code duplication. Most of it produced due to necessety to handle tiles with different granularity. Possible solution is to create loop over array with this tail size {8, 4, 2, 1} and generate code for its tile size by loop body in unified way.

dmitry-gorokhov (changes requested):

dmitry-gorokhov:
@xuchen-intel Would appreciate if you can take a look as well.

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -333,97 +244,617 @@ struct jit_uni_mvn_mean_variance_kernel_f32 : public jit_uni_mvn_mean_variance_k
...
-            size_t elt_num = std::min(jcp_.C - ur_num * unroll_size * vector_step, unroll_size * vector_step);
-            for (size_t ur_size = 0; ur_size < unroll_size_rt; ur_size++) {
-                uni_vpxor(Vmm(ur_base + 4 + ur_size), Vmm(ur_base + 4 + ur_size), Vmm(ur_base + 4 + ur_size));
-            }
-            if (jcp_.normalize_variance) {
-                for (size_t ur_size = 0; ur_size < unroll_size_rt; ur_size++) {
-                    uni_vmovups(Vmm(ur_base + 8 + ur_size), ptr[reg_mean + ur_offset + ur_size * vector_step * sizeof(float)]);
-                }
-            }
+            Xbyak::Label label_init_end;
```
xuchen-intel, line 319: For initializing vector registers here, AFAIK we can use 'for' loops to avoid code duplication, which will be unrolled during JIT compilation stage.

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -815,179 +1281,755 @@ struct jit_uni_mvn_kernel_f32 : public jit_uni_mvn_kernel, public jit_generator
...
+        mov(addr_unroll_size, rax);
+        L(label_reset_unroll_size_end);
+
+        // last unroll_size: vector_num % unroll_size
+        Xbyak::Label label_reset_last_unroll_size;
+        Xbyak::Label label_reset_last_unroll_size_end;
+        mov(rax, addr_vector_num);
+        xor_(rdx, rdx);
+        div(addr_unroll_size);  // rdx
+        cmp(rdx, 0);
```
xuchen-intel, line 1466: As I understand it, `rdx = rax / addr_unroll_size = addr_vector_num / addr_unroll_size`. And code block here https://github.com/chenhu-wang/openvino/blob/chenhu/shape_agnostic_jit_kernel_for_MVN/src/plugins/intel_cpu/src/nodes/mvn.cpp#L1450-L1458 assures `addr_vector_num >= addr_unroll_size`. So it seems `rdx == 0` will never be true. Did I miss anything?

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -333,97 +244,617 @@ struct jit_uni_mvn_mean_variance_kernel_f32 : public jit_uni_mvn_mean_variance_k
...
+        Xbyak::Label label_unroll_num_end;
+        L(label_unroll_num);
+        {
+            cmp(reg_unroll_num, 0);
+            jle(label_unroll_num_end, T_NEAR);
+
+            Xbyak::Label label_not_last;
+            cmp(reg_unroll_num, 1);
+            jne(label_not_last, T_NEAR);
+            mov(reg_unroll_size, last_unroll_size);
```
xuchen-intel, line 315: As I understatnd, code block here https://github.com/chenhu-wang/openvino/blob/chenhu/shape_agnostic_jit_kernel_for_MVN/src/plugins/intel_cpu/src/nodes/mvn.cpp#L279-L294 shows` last_unroll_size = reg_vector_num / reg_unroll_size`. And here we assign last_unroll_size back to reg_unroll_size again. Could you please help me understand the meaning of last_unroll_size for different cases?

xuchen-intel (changes requested):

wenjiew:
@chenhu-wang Did we address all the required updates? Thanks!

src/plugins/intel_cpu/src/nodes/mvn.cpp
chenhu-wang: applied, thanks Dmitry!

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -1505,8 +2662,8 @@ void MVN::MVNJitExecutor::mvn_pln(const uint8_t* src_data, uint8_t* dst_data, co
                 auto arg = jit_mvn_call_args();
                 arg.src = src_data + cc * src_data_size;
                 arg.sum = static_cast<float*>(&mean_internal);
-                arg.src_stride = src_stride_size;
                 arg.work_amount = static_cast<size_t>(C2 / blk_size); // for vector part
+                arg.rt_shape_size = static_cast<size_t>(C2 % blk_size);
```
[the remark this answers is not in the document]
chenhu-wang: It's a good practice to explicit cast if not the same type, means we are aware of this, implicit convert may have potential danger.

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -1860,15 +3031,16 @@ void MVN::MVNJitExecutor::mvn_blk(const uint8_t* src_data, uint8_t* dst_data, co
...
                 for (size_t i = 0; i < blk_size; i++)
                     mean_buffer_ptr[i] = 0.f;
 
                 auto arg = jit_mvn_call_args();
                 arg.src = src_data + src_offset * src_data_size;
                 arg.sum = mean_buffer_ptr;
-                arg.src_stride = src_stride_size;
                 arg.work_amount = static_cast<size_t>(W);
+                // real tail number or tail is 0(for full vector block).
+                arg.rt_shape_size = (C - cb * blk_size) < blk_size ? static_cast<size_t>(C % blk_size) : 0;
```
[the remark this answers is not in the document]
chenhu-wang: `arg.rt_shape_size = C % blk_size` is not correct. We set the value for each `cb`. If not tail `cb`, value should always 0, even if C % blk_size !=0. Thanks.

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -333,97 +244,617 @@ struct jit_uni_mvn_mean_variance_kernel_f32 : public jit_uni_mvn_mean_variance_k
...
+        Xbyak::Reg64 reg_unroll_num = rax;
+        Xbyak::Reg64 reg_vector_num = rbp;
+        Xbyak::Reg64 reg_tail_num = r13;
+        // size_t unroll_size = 4;
+        mov(reg_unroll_size, 4);
+        // size_t vec_num = C / vector_step
+        mov(rax, reg_rt_shape);
+        mov(reg_vector_num, vector_step);
+        xor_(rdx, rdx);
+        div(reg_vector_num);    // reg_rt_shape / vector_step, rax is result, rdx is tails(yushu)
```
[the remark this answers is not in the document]
chenhu-wang: fixed, thanks Chen.

src/plugins/intel_cpu/src/nodes/mvn.cpp
chenhu-wang: "for loop" is available only we know the iteration number, which is ok for static. For dynamic, we do not know the iteration number in compilation stage. we need dynamically decide if go to next iteration based on runtime shape.

src/plugins/intel_cpu/src/nodes/mvn.cpp
chenhu-wang: assume total shape is 16 *4 + 16 *4 + 16*2 +1. unroll with 4 vector. 16*2 is the last unroll, 1 is the tail

src/plugins/intel_cpu/src/nodes/mvn.cpp
chenhu-wang: rdx is the remainder(rax % addr_unroll_size), not rax / addr_unroll_size.

src/plugins/intel_cpu/src/nodes/mvn.cpp
chenhu-wang: @dmitry-gorokhov In case you have potential concern there are perf penalty with more compare, I had experiment shows that the cmp and jmp have nearly zero impact as all hidden in ILP in mvn code context, they are on different execution port.
Instead we can create static 1/2/3/4 unroll code block and compare only once on the top. But this will increase compilation time and compiled kernel size.

xuchen-intel: > "for loop" is available only we know the iteration number, which is ok for static. For dynamic, we do not know the iteration number in compilation stage. we need dynamically decide if go to next iteration based on runtime shape.

I mean, since the unroll factor is 4 for this code block https://github.com/chenhu-wang/openvino/blob/chenhu/shape_agnostic_jit_kernel_for_MVN/src/plugins/intel_cpu/src/nodes/mvn.cpp#L308-L326, why not just use "for loop" to generator the loop body. It seems can get the same code block you write, and avoid duplication.

chenhu-wang: @xuchen-intel , As I mentioned, runtime reg_unroll_size is not always 4 for dynamic case. We can create loops for each possible values. but again, it increase load-time, and kernel footprint size. Experiment shows kernel footprint size have slight perf negative impact.

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -333,97 +244,617 @@ struct jit_uni_mvn_mean_variance_kernel_f32 : public jit_uni_mvn_mean_variance_k
...
+            L(tail_blk8_label);
+            {
+                cmp(reg_tails_num_active, 8);
+                jl(tail_blk8_exit_label, T_NEAR);
+
+                worker_block(8);
+                add(reg_src_aux, 8 * jcp_.src_data_size);
+
+                sub(reg_tails_num_active, 8);
+                jmp(tail_blk8_label, T_NEAR);
```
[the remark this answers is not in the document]
chenhu-wang: This is included in [06cf822](https://github.com/openvinotoolkit/openvino/pull/17988/commits/06cf822d7405beebca294f1f60d8ec817973fead), thanks Chen

src/plugins/intel_cpu/src/nodes/mvn.cpp
chenhu-wang: Refactored in commit:https://github.com/openvinotoolkit/openvino/pull/17988/commits/621014e23bb2096fe8cc7aa1238698779223f4e4
Due to performance consideration, 8+4+2+1 tails should be processed continues, otherwise loop for each tile will go through the whole tensor, as this is nspc layout. I extract the process logic of tiles and vectors, accepting different worker type. This could help clean tails part as well as vector part. Thanks Dmitry!

xuchen-intel (approved):

chenhu-wang:
> @chenhu-wang Did we address all the required updates? Thanks!
@wenjiew yes, all necessary changes have applied.

src/plugins/intel_cpu/src/nodes/mvn.cpp
```diff
@@ -333,97 +233,460 @@ struct jit_uni_mvn_mean_variance_kernel_f32 : public jit_uni_mvn_mean_variance_k
...
+                }
+                L(loop_end_label);
+            };
+            Label tail_blk8_label;
+            Label tail_blk8_exit_label;
+            Label tail_blk4_label;
+            Label tail_blk4_exit_label;
+            Label tail_blk1_label;
+            Label tail_blk1_exit_label;
+            L(tail_blk8_label);
```
dmitry-gorokhov, line 601: The idea is to avoid code duplication by wrapping tiles processing into `for` loop. Look smt like that: 
```
Label tail_start[3]
Label tail_exit[3]
int[3] tails_sizes = {8, 4, 1}

for (int i = 0; i < 3; i++)
    L(tail_start[i]);
    {
        cmp(reg_rt_shape, tails_sizes[i]);
        jl(tail_exit[i], T_NEAR);

        unroll_w(tails_sizes[i]);
        save_result();
        reset_with_offset(tails_sizes[i]);

        sub(reg_rt_shape, tails_sizes[i]);
        jmp(tail_start[i], T_NEAR);
    }
    L(tail_start[i]);
```
Applying such approach for the whole MVN you can achive great adjustability of the tile processing. Like you can have one array of desired tile sizes which will be automatiacally applied the each part of the impl.

src/plugins/intel_cpu/src/nodes/mvn.cpp
dmitry-gorokhov: By loops I mentioned code generation level loops, not execution time. See my comment with example: https://github.com/openvinotoolkit/openvino/pull/17988#discussion_r1262211384
What do you think?

chenhu-wang: I see, with loop of tile number, we have clean code, and no more compilation time and compiled kernel size increased. Thanks a lot Dmitry! done in [d464ca6](https://github.com/openvinotoolkit/openvino/pull/17988/commits/d464ca6a2cb92f858670f01553892b975c0c20f5)

src/plugins/intel_cpu/src/nodes/mvn.cpp
chenhu-wang: Applied anywhere applicable, the single tail function with any work_type as parameters(std::function). (except nspc&&per_channel case, which is a dedicated process with unroll similar vector part, unify it seem don't make more clear), Welcome comment if you have more concerns. Thanks Dmitry. [d464ca6](https://github.com/openvinotoolkit/openvino/pull/17988/commits/d464ca6a2cb92f858670f01553892b975c0c20f5)

dmitry-gorokhov (commented):
LGTM in general. Can be merged after internal perforamnce validation

dmitry-gorokhov (approved):

PR #17996, opened 2023-06-12
[GPU] Support FC 6d output by compressing as 4d in kernel

kelvinchoi-intel:
### Details:
 - *FC kernel supports only 4d input/output for dynamic case, so compress the shape 6d/5d to 4d for fc kernel execution*

### Tickets:
 - *105930*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/fully_connected.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/fully_connected.cpp
@@ -37,7 +37,7 @@ public:
     static kernel_params_t get_kernel_params(const kernel_impl_params& impl_param, bool is_shape_agnostic = false) {
         const auto& primitive = impl_param.typed_desc<fully_connected>();
 
-        auto get_fc_input_layouts = [primitive](const std::vector<layout>& input_layouts) {
+        auto get_fc_input_layouts = [primitive](const std::vector<layout>& input_layouts, bool allow_new_shape_infer) {
             auto reshape_to_2d = [](const ov::PartialShape& shape, const ov::Dimension& feature, size_t rank) {
                 if (shape.is_static()) {
                     auto static_shape = shape.to_shape();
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/src/graph/impls/ocl/fully_connected.cpp
```diff
@@ -56,15 +56,23 @@ struct fully_connected_impl : typed_primitive_impl_ocl<fully_connected> {
...
 
             ov::Dimension feature = input0_pshape[std::min(primitive->input_size, static_cast<size_t>(4)) - 1ul];
+            if (allow_new_shape_infer) {
+                feature = input0_pshape[primitive->input_size - 1ul];
+            }
 
             if (primitive->input_size > 3) {
                 input0_layout.set_partial_shape(reshape_to_2d(input0_pshape, feature, primitive->input_size));
+                if (allow_new_shape_infer)
+                    input0_layout.format = format::bfyx;
```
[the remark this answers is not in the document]
kelvinchoi-intel: I wanted for my code not to affect static model.
Now, I've just deleted allow_new_shape_infer condition and successfully done with twins static model. Let's see if this code pass all tests.

src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
```diff
@@ -1727,6 +1727,15 @@ format layout_optimizer::get_preferred_format(program_node& node) {
         if (use_onednn_impls) {
             expected = node.get_preferred_output_fmt();
         }
+
+        if (!allow_new_shape_infer && node.is_type<fully_connected>()) {
+            auto& fc_node = node.as<fully_connected>();
+            auto input_layout = fc_node.input().get_output_layout();
+            if (input_layout.format.dimension() > 4) {
```
yeonbok, line 1734: Do we need this condition? Seems that it is always setting it bfyx

kelvinchoi-intel: For static case, 4d input could be others like yxfb. In this case, some tests are failed if it forces to set to bfyx.

yeonbok (approved):

PR #18006, opened 2023-06-12
[CPU] [ARM] Enable Reduce SLT tests on ARM

alvoron:


src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
```diff
@@ -1,217 +1,52 @@
...
-
-            inputs.insert({funcInput.get_node_shared_ptr(), tensor});
-        }
-    }
-
-private:
-    ngraph::helpers::ReductionType reductionType;
-    ElementType netPrecision;
+std::vector<std::vector<ov::test::InputShape>> inputShapes_dyn = {
+    {{{{1, 5}, 19, {1, 5}, {1, 10}}, {{2, 19, 2, 2}, {2, 19, 2, 9}}}},
```
EgorDuplensky, line 21: Just wondering, why we don't support dynamic shapes for reduce in case of ACL?

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
```diff
@@ -1,217 +1,52 @@
...
-        {0, 2, 3},
-        {1, 2, 3},
-        {0, 1, 2, 3}
+std::vector<std::vector<ov::test::InputShape>> inputShapes_SingleBatch_dyn = {
+    {{{{1, 5}, 19, {1, 5}, {1, 10}}, {{1, 19, 2, 2}, {1, 19, 2, 9}}}},
 };
 
+std::vector<CPUSpecificParams> cpuParams_4D = {
+        CPUSpecificParams({nChw16c}, {nChw16c}, {}, {}),
+        CPUSpecificParams({nhwc}, {nhwc}, {}, {}),
```
EgorDuplensky, line 46: Here we explicitly test nhwc only for x86.
Don't we plan to fix / address the issue with nhwc for ACL?

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
alvoron: Dynamic shapes are not supported by ConvertReduceMultiAxisBase: https://github.com/openvinotoolkit/openvino/blob/0541a12730b31fe8811811a26dc016bfa4e79316/src/plugins/intel_cpu/src/transformations/cpu_opset/arm/pass/convert_reduce_multi_axis.cpp#L24

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
alvoron: NHWC was disabled for ACL Executor explicitly due to accuracy issues:
https://github.com/openvinotoolkit/openvino/blob/0541a12730b31fe8811811a26dc016bfa4e79316/src/plugins/intel_cpu/src/nodes/reduce.cpp#L1875
Issue is raised in ACL repo: https://github.com/ARM-software/ComputeLibrary/issues/1044

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
EgorDuplensky: Isn't it enough to just check the second input (reduce axes)?
I think the transformation should work if only the first input is dynamic

alvoron: Dynamic shapes tests are added to common section (smoke_Reduce_OneAxis_dynamic_CPU and params_MultiAxis_4D_dynamic)

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
EgorDuplensky: Agreed to move ifdef into SetUp() (using some util method) to skip the tests based on NHWC layout

EgorDuplensky (approved):

EgorDuplensky:
@maxnick Could you please proceed with the review

maxnick (approved):
In general LGTM. The nitpicking comments may be resolved later.

alvoron:
> In general LGTM. The nitpicking comments may be resolved later.

CI is green. Could you merge it? I'll fix RTTI comments in the next PR.

PR #18008, opened 2023-06-12
[GNA] Fix for concat requantization problem

marcinkacprzak:
### Details:
 Problem: if between one of Concat inputs and the Parameter layer there is no layer, which could get requantized according to scale factors from the second Concat input, plugin cannot properly propagate scale factors

Solution: in such a case additional layer has been added - Identity, which is used as a holder for scale factors data

### Tickets:
 - 111356

### DIFF
--- a/src/plugins/intel_gna/src/common/graph_utils.hpp
+++ b/src/plugins/intel_gna/src/common/graph_utils.hpp
@@ -202,6 +202,18 @@ inline bool is_pooling(const std::shared_ptr<ngraph::Node>& node) {
             std::dynamic_pointer_cast<ov::intel_gna::op::GNAMaxPool>(node) != nullptr);
 }
 
+inline bool is_concat(const std::shared_ptr<ngraph::Node>& node) {
+    return (std::dynamic_pointer_cast<ov::opset12::Concat>(node) != nullptr);
+}
+
+inline bool is_fake_quantize(const std::shared_ptr<ngraph::Node>& node) {
+    return (std::dynamic_pointer_cast<ov::opset12::FakeQuantize>(node) != nullptr);
+}
+
+inline bool is_read_value(const std::shared_ptr<ngraph::Node>& node) {
+    return (std::dynamic_pointer_cast<ov::opset12::ReadValue>(node) != nullptr);
+}
+
 template <typename T>
 inline bool is_Tbit_fq(const std::shared_ptr<ngraph::Node>& node) {
     auto fq_node = std::dynamic_pointer_cast<ngraph::opset9::FakeQuantize>(node);
@@ -275,8 +287,7 @@ inline bool is_interleaved(const std::shared_ptr<ov::Node>& node) {
 inline bool is_gna_precision_agnostic(std::shared_ptr<ngraph::Node> node) {
     return ((std::dynamic_pointer_cast<ngraph::opset9::VariadicSplit>(node) != nullptr) ||
             (std::dynamic_pointer_cast<ngraph::opset9::Split>(node) != nullptr) ||
-            (std::dynamic_pointer_cast<ngraph::opset9::Slice>(node) != nullptr) ||
-            (std::dynamic_pointer_cast<ngraph::opset9::Concat>(node) != nullptr) ||
+            (std::dynamic_pointer_cast<ngraph::opset9::Slice>(node) != nullptr) || is_concat(node) ||
             (std::dynamic_pointer_cast<ngraph::opset9::Reshape>(node) != nullptr) ||
             (std::dynamic_pointer_cast<ngraph::opset9::Squeeze>(node) != nullptr) ||
             (std::dynamic_pointer_cast<ngraph::opset9::Unsqueeze>(node) != nullptr) ||
--- a/src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
+++ b/src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
@@ -157,8 +157,7 @@ InsertCopyBeforeAssignLayer::InsertCopyBeforeAssignLayer() {
 
             // Crop -> Memory, Input -> Split -> Memory, Concat -> Memory
             if ((std::dynamic_pointer_cast<ngraph::op::CropIE>(current_node) && !is_crop_affined(current_node)) ||
-                std::dynamic_pointer_cast<ngraph::opset8::Concat>(current_node) ||
-                std::dynamic_pointer_cast<ngraph::opset8::Split>(current_node) ||
+                is_concat(current_node) || std::dynamic_pointer_cast<ngraph::opset8::Split>(current_node) ||
                 std::dynamic_pointer_cast<ngraph::opset8::VariadicSplit>(current_node)) {
                 insert_copy_layer_between(matched_node_input, node, i);
             }
@@ -281,7 +280,7 @@ bool HandleMultiConnectedLayerToConcatAndMemory::run_on_model(const std::shared_
                 for (const auto& child_info : children_info) {
                     auto child = std::get<1>(child_info);
 
-                    if (std::dynamic_pointer_cast<ngraph::opset8::Concat>(child)) {
+                    if (is_concat(child)) {
                         concat_nodes.push_back(child_info);
                     } else if (std::dynamic_pointer_cast<ngraph::op::ReadValueBase>(child) ||
                                std::dynamic_pointer_cast<ngraph::op::AssignBase>(child)) {
--- a/src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
+++ b/src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
@@ -24,18 +24,16 @@ using namespace ov::intel_gna::graph_utils;
 using namespace ov::pass::pattern;
 using namespace ov::op::util;
 
-using namespace std;
-
 namespace {
-void mark_for_identity_insertion(shared_ptr<ov::Node> node, size_t input_index) {
+void mark_for_identity_insertion(std::shared_ptr<ov::Node> node, size_t input_index) {
     log::debug() << "Mark input as candidate for identity insertion " << input_index << ":" << node->get_friendly_name()
-                 << endl;
+                 << std::endl;
     auto input = node->input(input_index);
     add_precision_change_flag(input, ov::element::i32, ov::element::i16);
 }
 
-shared_ptr<ov::intel_gna::op::Identity> create_indentity(shared_ptr<ov::Node>& input_op) {
-    auto identity_op = make_shared<ov::intel_gna::op::Identity>(input_op);
+std::shared_ptr<ov::intel_gna::op::Identity> create_indentity(std::shared_ptr<ov::Node>& input_op) {
+    auto identity_op = std::make_shared<ov::intel_gna::op::Identity>(input_op);
     // Keep name of previous operation
     identity_op->set_friendly_name(input_op->get_friendly_name());
     input_op->set_friendly_name(input_op->get_friendly_name() + "/previous");
@@ -43,11 +41,11 @@ shared_ptr<ov::intel_gna::op::Identity> create_indentity(shared_ptr<ov::Node>& i
     return identity_op;
 }
 
-void insert_identity_layer_after(shared_ptr<ov::Node>& input_op, size_t index) {
+void insert_identity_layer_after(std::shared_ptr<ov::Node>& input_op, size_t index) {
     NGRAPH_CHECK(input_op);
 
     log::debug() << "Insert identity layer after " << input_op->get_friendly_name() << " (" << input_op->get_type_name()
-                 << "):" << index << endl;
+                 << "):" << index << std::endl;
 
     auto consumers = input_op->output(index).get_target_inputs();
     auto identity_op = create_indentity(input_op);
@@ -56,27 +54,29 @@ void insert_identity_layer_after(shared_ptr<ov::Node>& input_op, size_t index) {
     }
 }
 
-void insert_identity_layer_between(shared_ptr<ov::Node>& input_op, shared_ptr<ov::Node>& output_op, size_t index) {
+void insert_identity_layer_between(std::shared_ptr<ov::Node>& input_op,
+                                   std::shared_ptr<ov::Node>& output_op,
+                                   size_t index) {
     NGRAPH_CHECK(input_op);
     NGRAPH_CHECK(output_op);
 
     log::debug() << "Insert identity layer after " << input_op->get_friendly_name() << " (" << input_op->get_type_name()
                  << ") and before " << index << ":" << output_op->get_friendly_name() << " ("
-                 << output_op->get_type_name() << ")" << endl;
+                 << output_op->get_type_name() << ")" << std::endl;
 
     auto identity_op = create_indentity(input_op);
     output_op->input(index).replace_source_output(identity_op);
 }
 
 // forward declaration
-bool walk_through_the_outputs(shared_ptr<ov::Node>& prev_node,
+bool walk_through_the_outputs(std::shared_ptr<ov::Node>& prev_node,
                               size_t& prev_node_output_index,
-                              const shared_ptr<ov::Node>& node,
+                              const std::shared_ptr<ov::Node>& node,
                               bool first_iteration = false);
 
-bool process_next_node(shared_ptr<ov::Node>& prev_node,
+bool process_next_node(std::shared_ptr<ov::Node>& prev_node,
                        size_t& prev_node_output_index,
-                       const shared_ptr<ov::Node>& node,
+                       const std::shared_ptr<ov::Node>& node,
                        const size_t input_index) {
     // Check whether node is going to be skipped
     bool to_be_skipped = (is_gna_precision_agnostic(node) && !is_concat(node)) || is_pooling(node);
@@ -104,9 +104,9 @@ bool process_next_node(shared_ptr<ov::Node>& prev_node,
     return false;
 }
 
-bool walk_through_the_outputs(shared_ptr<ov::Node>& prev_node,
+bool walk_through_the_outputs(std::shared_ptr<ov::Node>& prev_node,
                               size_t& prev_node_output_index,
-                              const shared_ptr<ov::Node>& node,
+                              const std::shared_ptr<ov::Node>& node,
                               bool first_iteration) {
     bool is_identity_inserted = false;
     // walk over all outputs
@@ -133,7 +133,7 @@ bool walk_through_the_outputs(shared_ptr<ov::Node>& prev_node,
 }
 }  // namespace
 
-bool MarkIdentityCandidates::run_on_model(const shared_ptr<ov::Model>& m) {
+bool MarkIdentityCandidates::run_on_model(const std::shared_ptr<ov::Model>& m) {
     RUN_ON_FUNCTION_SCOPE(MarkIdentityCandidates);
     for (auto& node : m->get_ordered_ops()) {
         auto check_previous_node_and_mark = [&node]() {
@@ -145,7 +145,7 @@ bool MarkIdentityCandidates::run_on_model(const shared_ptr<ov::Model>& m) {
                 }
             }
         };
-        if (dynamic_pointer_cast<ngraph::op::Eltwise>(node)) {
+        if (std::dynamic_pointer_cast<ngraph::op::Eltwise>(node)) {
             auto input0_node = node->get_input_node_shared_ptr(0);
             auto input1_node = node->get_input_node_shared_ptr(1);
             auto func_input0_node = get_prev_node_skipping_certain(input0_node, is_gna_precision_agnostic);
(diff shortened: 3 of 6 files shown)


src/plugins/intel_gna/src/common/graph_utils.hpp
```diff
@@ -20,6 +20,7 @@
 #include "ngraph/opsets/opset7.hpp"
 #include "ngraph/opsets/opset8.hpp"
 #include "ngraph/opsets/opset9.hpp"
+#include "openvino/opsets/opset11.hpp"
```
marcinkusm, line 23: 12?

src/plugins/intel_gna/src/common/graph_utils.hpp
```diff
@@ -31,6 +32,8 @@ namespace ov {
 namespace intel_gna {
 namespace graph_utils {
 
+using namespace ov::opset11;
```
marcinkusm, line 35: it is better to not use using in header

src/plugins/intel_gna/src/common/graph_utils.hpp
```diff
@@ -252,6 +267,10 @@ inline bool is_gna_precision_agnostic(std::shared_ptr<ngraph::Node> node) {
             ((std::dynamic_pointer_cast<ngraph::op::CropIE>(node) != nullptr) && !is_crop_affined(node)));
 }
 
+inline bool is_gna_not_quantizable(std::shared_ptr<ngraph::Node> node) {
```
marcinkusm, line 270: is_node_not_quantizable_by_gna()?

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -231,3 +231,29 @@ bool IdentityCandidatesCleanup::run_on_model(const std::shared_ptr<ov::Model>& f
...
 }
+
+bool InsertIdentityForNonQuantizableConcatInput::insert_identity_for_non_quantizable_concat_inputs(
+    const std::shared_ptr<ov::Node>& node) {
+    bool is_graph_modified = false;
+    for (size_t i = 0; i < node->get_input_size(); i++) {
+        auto concat_input_node = node->get_input_node_shared_ptr(i);
+        auto prev_node = get_prev_node_skipping_certain(concat_input_node, is_gna_not_quantizable);
+        if (is_parameter(concat_input_node) || is_parameter(prev_node)) {
+            insert_identity_layer_after(concat_input_node, 0);
```
marcinkusm, line 242: is 0 proper value?

marcinkusm (changes requested):
Great work!
but please remove using from header

sirzabek (changes requested):

src/plugins/intel_gna/src/common/graph_utils.hpp
marcinkacprzak: Changed to 12.

src/plugins/intel_gna/src/common/graph_utils.hpp
marcinkacprzak: Function moved to CPP according to Kuba's comment.

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
marcinkacprzak: Good question. Added method for calculation of proper value.

src/plugins/intel_gna/src/common/graph_utils.hpp
```diff
@@ -197,6 +200,18 @@ inline bool is_pooling(const std::shared_ptr<ngraph::Node>& node) {
     return (std::dynamic_pointer_cast<ngraph::opset7::MaxPool>(node) != nullptr);
 }
 
+inline bool is_parameter(const std::shared_ptr<ngraph::Node>& node) {
+    return (std::dynamic_pointer_cast<Parameter>(node) != nullptr);
+}
+
+inline bool is_concat(const std::shared_ptr<ngraph::Node>& node) {
```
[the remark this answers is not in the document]
marcinkacprzak: Added use of this function to other places in the code.

src/plugins/intel_gna/src/transformations/insert_identity_layer.hpp
```diff
@@ -68,6 +68,44 @@ class IdentityCandidatesCleanup : public ov::pass::ModelPass {
...
+ *       |
+ *   Convolution        Parameter2
+ *       |                  |
+ *   Activation      Non-Quantizable or FQ
+ *              \       /
+ *               Concat
+ *                  |
+ *               Result
+ *
+ * For the above model, during scale factors propagation, when left Concat input
```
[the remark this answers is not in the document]
marcinkacprzak: Changed

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -231,3 +231,29 @@ bool IdentityCandidatesCleanup::run_on_model(const std::shared_ptr<ov::Model>& f
     }
     return false;
 }
+
+bool InsertIdentityForNonQuantizableConcatInput::insert_identity_for_non_quantizable_concat_inputs(
```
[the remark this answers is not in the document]
marcinkacprzak: Doing so made code complicated, and unreadable.

src/plugins/intel_gna/tests/unit/transformations/gna_insert_identity_layer.cpp
```diff
@@ -7,49 +7,60 @@
...
 #include <ngraph/function.hpp>
-#include <ngraph/opsets/opset7.hpp>
-#include <ngraph/opsets/opset9.hpp>
 #include <ngraph/pass/manager.hpp>
 #include <ngraph_functions/builders.hpp>
 #include <transformations/init_node_info.hpp>
 #include <transformations/utils/utils.hpp>
 
+#include "openvino/opsets/opset12.hpp"
+#include "openvino/opsets/opset7.hpp"
```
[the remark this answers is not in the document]
marcinkacprzak: Yes.

src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
```diff
@@ -157,8 +157,7 @@ InsertCopyBeforeAssignLayer::InsertCopyBeforeAssignLayer() {
 
             // Crop -> Memory, Input -> Split -> Memory, Concat -> Memory
             if ((std::dynamic_pointer_cast<ngraph::op::CropIE>(current_node) && !is_crop_affined(current_node)) ||
-                std::dynamic_pointer_cast<ngraph::opset8::Concat>(current_node) ||
-                std::dynamic_pointer_cast<ngraph::opset8::Split>(current_node) ||
+                is_concat(current_node) || std::dynamic_pointer_cast<ngraph::opset8::Split>(current_node) ||
```
marcinkusm, line 160: don't you want to add is_split()?

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -20,29 +21,33 @@ using namespace ov::intel_gna;
 using namespace ov::intel_gna::pass;
 using namespace ov::intel_gna::rt_info;
 using namespace ov::intel_gna::graph_utils;
+using namespace ov::pass::pattern;
+using namespace ov::op::util;
+
+using namespace std;
```
marcinkusm, line 27: I would remove this using to be consistent with rest sources, but it is not mandatory.
I think we don't use using for std namespace in any place. Are you sure you want to do this?
I don't understand why using namespace std would be beneficial, and at the same time not  adding using namespace ov( when all other nemespaces ov like namespaces are added)

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -144,7 +145,7 @@ bool MarkIdentityCandidates::run_on_model(const std::shared_ptr<ov::Model>& m) {
                 }
             }
         };
-        if (std::dynamic_pointer_cast<ngraph::op::Eltwise>(node)) {
+        if (dynamic_pointer_cast<ngraph::op::Eltwise>(node)) {
```
marcinkusm, line 148: `is_eltwise()`?

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -231,3 +232,98 @@ bool IdentityCandidatesCleanup::run_on_model(const std::shared_ptr<ov::Model>& f
     }
     return false;
 }
+
+bool InsertIdentityForPrecAgnosticConcatInput::all_inputs_point_the_same_node(const shared_ptr<ov::Node>& node) {
```
marcinkusm, line 236: usually first work in function/method name is a verbe 
```suggestion
bool InsertIdentityForPrecAgnosticConcatInput::are_all_inputs_point_the_same_node(const shared_ptr<ov::Node>& node) {
```

src/plugins/intel_gna/src/transformations/insert_identity_layer.hpp
```diff
@@ -68,6 +68,73 @@ class IdentityCandidatesCleanup : public ov::pass::ModelPass {
...
+ *                  Prec-Agnostic or FQ
+ *                        |
+ *      Functional     Identity
+ *              \       /
+ *               Concat
+ *                  |
+ *               Result
+ */
+class InsertIdentityForPrecAgnosticConcatInput : public ov::pass::ModelPass {
+    /**
```
marcinkusm, line 100: 1. as it was agreed in coding convention private members and methods declaration should be done after public.
2. Are you sure you want to expose this methods in pass declaration. Almost all methods in passes which are not constructors are declared as free functions in source file(static or inline to avoid missing forward declaration warning).

src/plugins/intel_gna/src/transformations/insert_identity_layer.hpp
```diff
@@ -68,6 +68,73 @@ class IdentityCandidatesCleanup : public ov::pass::ModelPass {
...
+     * @brief Invoke Identity layer insertion in case the Concat input is unable
+     * to store scale factors
+     */
+    bool insert_identity_for_prec_agnostic_concat_inputs(const std::shared_ptr<ov::Node>& node);
+
+    /**
+     * @brief Find the output index of 'prev' layer, on which it is connected to 'next' layer
+     */
+    size_t find_prev_layer_output_index(std::shared_ptr<ov::Node> prev, std::shared_ptr<ov::Node> next);
```
marcinkusm, line 131: in previous declarations const & are used for shared_ptr-s, but in this method not. Is it done on purpose?

src/plugins/intel_gna/src/transformations/insert_identity_layer.hpp
```diff
@@ -68,6 +68,73 @@ class IdentityCandidatesCleanup : public ov::pass::ModelPass {
...
+ * will be selected as a scale factors source, then algorithm will not be able to
+ * apply the scale factors to the right Concat input and will throw exception.
+ * InsertIdentityForPrecAgnosticConcatInput pass adds Identity layer, which
+ * is capable of storing scale factors, so the scale factors propagation can proceed:
+ *
+ *                    Parameter
+ *                        |
+ *                  Prec-Agnostic or FQ
+ *                        |
+ *      Functional     Identity
```
marcinkusm, line 93: 1.what in case if there will be more than two inputs to concat?
2. what in case second input will be functional and first Prec-agnostic?

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -231,3 +232,98 @@ bool IdentityCandidatesCleanup::run_on_model(const std::shared_ptr<ov::Model>& f
...
+size_t InsertIdentityForPrecAgnosticConcatInput::find_prev_layer_output_index(shared_ptr<ov::Node> prev,
+                                                                              shared_ptr<ov::Node> next) {
+    for (const auto& output : prev->outputs()) {
+        for (const auto& input : output.get_target_inputs()) {
+            if (input.get_node() == next.get()) {
+                return output.get_index();
+            }
+        }
+    }
+    return 0;
```
marcinkusm, line 253: is 0 correct or fault result?
in this approach if there will be no matching inputs 0 will be returned and used as proper value
shouldn't be here any exception?

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -231,3 +232,98 @@ bool IdentityCandidatesCleanup::run_on_model(const std::shared_ptr<ov::Model>& f
...
+
+bool InsertIdentityForPrecAgnosticConcatInput::insert_identity_after_nodes(const vector<shared_ptr<ov::Node>>& nodes,
+                                                                           const shared_ptr<ov::Node>& next) {
+    bool is_graph_modified = false;
+    for (auto node : nodes) {
+        size_t index = find_prev_layer_output_index(node, next);
+        insert_identity_layer_after(node, index);
+        is_graph_modified = true;
+    }
+    return is_graph_modified;
```
marcinkusm, line 264: this value will be always true if there will be at least one node on nodes list.

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -231,3 +232,98 @@ bool IdentityCandidatesCleanup::run_on_model(const std::shared_ptr<ov::Model>& f
...
+    }
+
+    if (!has_fq_on_any_input(concat_node)) {
+        return false;
+    }
+
+    const auto& nodes = get_nodes_for_identity_insertion(concat_node);
+
+    // Skip adding Identity if none of inputs can hold scale factors
+    bool no_input_can_hold_sf = concat_node->get_input_size() == nodes.size();
```
marcinkusm, line 311: based on comment shouldn't condition look like `if( nodes.empty())` ?

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -231,3 +232,98 @@ bool IdentityCandidatesCleanup::run_on_model(const std::shared_ptr<ov::Model>& f
...
+    return insert_identity_after_nodes(nodes, concat_node);
+}
+
+bool InsertIdentityForPrecAgnosticConcatInput::run_on_model(const shared_ptr<ov::Model>& m) {
+    RUN_ON_FUNCTION_SCOPE(InsertIdentityForPrecAgnosticConcatInput);
+    bool is_graph_modified = false;
+    for (auto& node : m->get_ordered_ops()) {
+        if (is_concat(node)) {
+            is_graph_modified |= insert_identity_for_prec_agnostic_concat_inputs(node);
+            continue;
```
marcinkusm, line 325: what is a sense of continue here if there is no more operations after this condition?
```
{
    if (!is_concat(node)) {
        continue;
    }

    if(!insert_identity_for_prec_agnostic_concat_inputs(node)) {
        continue;
    }

    is_graph_modified  = true;
}
```
or 
```
{
    if (is_concat(node) && insert_identity_for_prec_agnostic_concat_inputs(node)) {
        is_graph_modified  = true;
    }
}
```

marcinkusm (commented):
Nice work, I have just few question for clarification and I see possible improvements.

src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
marcinkacprzak: No, there are tons of this kind of check. I have added is_concat, because I was using it in my code, then changed all occurrences of std::dynamic_pointer_cast<ngraph::opset8::Split>() in the rest of the code. 
Changing all other checks for other nodes is outside of scope of this task.

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
marcinkacprzak: The obvious benefit to me is shorten code. I was trying to add "using namespace ov", but there were some ambiguity problems and I didn't have time to investigate it. So left as is for now.
Is "using namespace std" inconsistent with our coding rules?

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
marcinkacprzak: As above.

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
marcinkacprzak: No, because `nodes` contain all inputs, which are not able to hold scale factors.

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
marcinkacprzak: Yes, `continue` here was a nonsense.

src/plugins/intel_gna/src/transformations/insert_identity_layer.hpp
marcinkacprzak: 1. Agreed / changed
2. I'm sure ;)

src/plugins/intel_gna/src/transformations/insert_identity_layer.hpp
marcinkacprzak: Updated description.

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
marcinkacprzak: Added exception.

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
marcinkacprzak: Changed to `return nodes.size() > 0`

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
marcinkusm: It is not inconsistent, but introducing this blurred a bit the PR, because dozes of lines were changes only because of that.

Problem I see that in most part of the code we don't use this using so it would be inconsistent with rest of the code.

marcinkusm: you can resolve this comment if you want stay with `using namespace std;`

mryzhov: Yes, the code will be shorten but imho the most important for development to leave the code uniform, plus using std:: helps to identify stl functions in the code. But it doesn't block the PR

mryzhov (approved):

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
marcinkacprzak: Ok, reverted, however I don't see any advantage of keeping this in code as shared_ptr, vector, map, make_shared and so on are pretty clear.

src/plugins/intel_gna/src/transformations/insert_identity_layer.hpp
marcinkacprzak: When I have implemented the needed functionality for InsertIdentityForPrecAgnosticConcatInput pass, the code was far too long to be easily readable and understandable for others. So I divided it into shorter, functional parts. And because still those parts functionally are inseparably related to InsertIdentityForPrecAgnosticConcatInput, I just created in-class methods. I didn't want to create static functions in cpp file, which contains implementation for many classes, because these functions are not universal and are used just by one of these classes for specific purpose.

src/plugins/intel_gna/src/transformations/insert_identity_layer.hpp
marcinkacprzak: No. Changed.

marcinkusm (approved):
👍

src/plugins/intel_gna/src/transformations/insert_identity_layer.hpp
```diff
@@ -68,6 +68,77 @@ class IdentityCandidatesCleanup : public ov::pass::ModelPass {
...
+     * @brief Invokes Identity layer insertion after each node in vector
+     */
+    bool insert_identity_after_nodes(const std::vector<std::shared_ptr<ov::Node>>& nodes,
+                                     const std::shared_ptr<ov::Node>& next);
+
+    /**
+     * @brief Invoke Identity layer insertion in case the Concat input is unable
+     * to store scale factors
+     */
+    bool insert_identity_for_prec_agnostic_concat_inputs(const std::shared_ptr<ov::Node>& node);
```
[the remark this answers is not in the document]
marcinkacprzak: Ok, done.

src/plugins/intel_gna/src/transformations/insert_identity_layer.cpp
```diff
@@ -231,3 +232,95 @@ bool IdentityCandidatesCleanup::run_on_model(const std::shared_ptr<ov::Model>& f
...
+    }
+
+    return insert_identity_after_nodes(nodes, concat_node);
+}
+
+bool InsertIdentityForPrecAgnosticConcatInput::run_on_model(const shared_ptr<ov::Model>& m) {
+    RUN_ON_FUNCTION_SCOPE(InsertIdentityForPrecAgnosticConcatInput);
+    bool is_graph_modified = false;
+    for (auto& node : m->get_ordered_ops()) {
+        if (is_concat(node)) {
```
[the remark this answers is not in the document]
marcinkacprzak: Suggestion rejected - a matter of taste.

sirzabek (approved):

PR #18015, opened 2023-06-12
[PT FE] Print better exception message

mvafin:
### Details:
 - *Print better exception message*

### Tickets:
 - *107879*


src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -65,8 +65,9 @@ def get_value_from_getattr(getattr_node, self_module):
     module = self_module
     while len(stack) > 0:
         node = stack.pop()
-        assert (hasattr(module, node.s("name")))
-        module = getattr(module, node.s("name"))
+        attr_name = node.s("name")
+        assert hasattr(module, attr_name), f"No attribute with name \"{attr_name}\" found in module."
```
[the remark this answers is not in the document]
rkazants: Just have the same style

src/frontends/common/src/exception.cpp
```diff
@@ -8,33 +8,33 @@ void ov::frontend::GeneralFailure::create(const CheckLocInfo& check_loc_info,
...
     throw ov::frontend::OpValidationFailure(
-        make_what(check_loc_info, "FrontEnd API failed with OpValidationFailure: " + context_info, explanation));
+        make_what(check_loc_info, "FrontEnd API failed with OpValidationFailure" + context_info, explanation));
 }
 
 void ov::frontend::OpConversionFailure::create(const CheckLocInfo& check_loc_info,
                                                const std::string& context_info,
                                                const std::string& explanation) {
     throw ov::frontend::OpConversionFailure(
-        make_what(check_loc_info, "FrontEnd API failed with OpConversionFailure: " + context_info, explanation));
```
rkazants, line 32: I am not sure that we should remove `:`
Logically, below it is going explanation of some failure

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -38,16 +38,24 @@ namespace frontend {
...
+                unconverted_ops_types[attrs.at(PtFrameworkNode::op_type_key)] = exception_msg;
+            }
         }
         if (const auto& fw_node = ov::as_type_ptr<ov::op::util::MultiSubGraphOp>(node)) {
             for (size_t i = 0; i < fw_node->get_internal_subgraphs_size(); i++) {
-                auto internal_types = get_unconverted_types_from_model(fw_node->get_function(i));
+                const auto& internal_types = get_unconverted_types_from_model(fw_node->get_function(i));
                 unconverted_ops_types.insert(internal_types.begin(), internal_types.end());
             }
         }
```
rkazants, line 61: can we re-use the same function from TF FE? And move it to the common part?

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -38,16 +38,24 @@ namespace frontend {
...
+            const auto& attrs = fw_node->get_attrs();
+            FRONT_END_GENERAL_CHECK(attrs.find(PtFrameworkNode::op_type_key) != attrs.end(),
+                                    "FrameworkNode attributes do not contain operation type.");
+            std::string exception_msg;
+            if (attrs.find(PtFrameworkNode::failed_conversion_key) != attrs.end()) {
+                exception_msg = attrs.at(PtFrameworkNode::failed_conversion_key);
+            }
+            if (!unconverted_ops_types.count(attrs.at(PtFrameworkNode::op_type_key))) {
+                unconverted_ops_types[attrs.at(PtFrameworkNode::op_type_key)] = exception_msg;
+            }
```
rkazants, line 54: How do you differentiate unsupported operation and failures in existing translator? May be it is better to have two separate containers for better handling

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -60,17 +68,51 @@ FrontEnd::FrontEnd() : m_op_translators(get_supported_ops()) {}
...
+        normalize(converted_model);
+    } catch (const std::exception& e) {
+        norm_err = "\n-- normalize step failed with: " + std::string(e.what());
+    }
+
+    const auto& unconverted_ops = get_unconverted_types_from_model(converted_model);
+    std::stringstream unconverted_ops_msg;
+    std::stringstream failed_ops_msg;
+    std::stringstream failed_ops_short;
+    unconverted_ops_msg << "-- No conversion rule found for operations: ";
```
rkazants, line 85: I think we definitely need to align outputs among FEs. To have same output format. Now it is not quite good.

src/frontends/pytorch/src/translate_session.cpp
```diff
@@ -207,21 +207,28 @@ std::shared_ptr<Model> TranslateSession::convert_pytorch_model(
...
                        << '\n';
+        exception = "Unknown exception type.";
+    }
+    try {
+        // Create PtFrameworkNode for everything that wasn't able to be converted normally
+        return make_framework_node(context, exception);
+    } catch (std::exception& e) {
+        exception += " Failed to create FrameworkNode with subgraphs with exception: " + std::string(e.what());
+        return make_framework_node_ignore_bodies(context, exception);
     }
```
rkazants, line 231: Also, we need to handle unknown exception types. For example, someone implemented extension and throws smth bad there

src/frontends/pytorch/src/utils.hpp
```diff
@@ -52,7 +52,8 @@ op::PadType convert_pad(const std::string& pt_pad);
 
 Output<Node> concat_list_construct(const Output<Node>& input);
 
-OutputVector make_framework_node(const NodeContext& context);
+OutputVector make_framework_node_ignore_bodies(const NodeContext& context, const std::string& exception);
+OutputVector make_framework_node(const NodeContext& context, const std::string& exception);
```
rkazants, line 57: Please provide tests for the report to cover cases with different exceptions, etc.

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -60,17 +68,51 @@ FrontEnd::FrontEnd() : m_op_translators(get_supported_ops()) {}
...
-    std::stringstream ops_str;
-    for (auto&& op_type : unconverted_ops_types) {
+    std::stringstream error_msg("Model wasn't fully converted.");
+    error_msg << "Model wasn't fully converted.";
+
+    std::string norm_err;
+    try {
+        normalize(converted_model);
+    } catch (const std::exception& e) {
+        norm_err = "\n-- normalize step failed with: " + std::string(e.what());
```
slyalin, line 78: It doesn't contain information about which particular transformation fails -- it's hard to follow -- I'm looking at an example of error log that you sent me earlier. Can we capture the name of transformation that failed?

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -60,17 +68,51 @@ FrontEnd::FrontEnd() : m_op_translators(get_supported_ops()) {}
...
+        if (op.second.empty()) {
+            if (at_least_one)
+                unconverted_ops_msg << ", ";
+            unconverted_ops_msg << op.first;
+            at_least_one = true;
+        } else {
+            if (at_least_one_except)
+                failed_ops_short << ", ";
+            failed_ops_short << op.first;
+            failed_ops_msg << "\n-- " << op.first << " with a message:\n" << op.second;
```
slyalin, line 103: Also print a name of a node. When scoped name is available it gives understanding where in the graph an issue appeared.

slyalin, line 103: Also recommend to make an indentation for paragraphs. As you are printing basically a list of issues, it is hard to find where one ends and the next starts. `--` is a good try, but as I see this can be better.

Also please use more blank lines as a separator (in addition to indentation or alone). Output may include multiple python stack traces (as I see in your output example) which itself contains black lines as a separators, and provide another level of separation of items.

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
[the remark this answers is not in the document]
mvafin: Same style with what? This file always use f"{}" form of string formatting

src/frontends/common/src/exception.cpp
mvafin: Yes we should. ":" are added in `make_what` function, if we add them here we have double `: :`

src/frontends/pytorch/src/utils.hpp
mvafin: What kind of tests is expected here? Testing exact wording of error I think is useless

src/frontends/pytorch/src/frontend.cpp
mvafin: I personally think that less newlines is better, otherwise it just occupy more space with same information

src/frontends/pytorch/src/frontend.cpp
mvafin: Both functions have slightly different approach, unification can be done as next stage, ad will require changing TF FE

src/frontends/pytorch/src/frontend.cpp
mvafin: TF FE doesn't suffer from Python stack trace in the error so it can layout error messages differently then PT FE

eaidova (approved):

src/frontends/pytorch/src/utils.hpp
rkazants: Not useless, you just validate that all unsupported operations and exceptions are printed. And nothing is missed.
We validate this in TF FE and quite useful test

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -40,41 +40,88 @@ namespace frontend {
...
+    error_msg << "Model wasn't fully converted.";
+    unconverted_ops_msg << "-- No conversion rule found for operations: ";
+    failed_ops_msg << " Failed operations detailed log:";
+    failed_ops_short << "-- Conversion is failed for: ";
+    bool at_least_one = false;
+    bool at_least_one_except = false;
+    for (auto&& op : unconverted_ops) {
+        if (op.second.empty()) {
+            if (at_least_one)
+                unconverted_ops_msg << ", ";
```
rkazants, line 83: Will we have sorted list of unsupported ops?

rkazants (approved):
Please implement test(s) for this logic

src/frontends/pytorch/src/frontend.cpp
mvafin: Yes

mvafin:
> Please implement test(s) for this logic

Added test for `convert_model` because there was concern that it might suppress some exceptions, but it works as expected.

src/frontends/pytorch/src/utils.hpp
mvafin: Added test

src/frontends/pytorch/src/frontend.cpp
slyalin: Exactly, please create a ticket against core to add such information?

src/frontends/pytorch/src/frontend.cpp
slyalin: Indentation was not a good idea indeed without knowing desirted "screen width", so OK, let's leave it as-is.

src/frontends/pytorch/src/frontend.cpp
mvafin: > Exactly, please create a ticket against core to add such information?

Done, ticket 113306

slyalin (approved):

PR #18018, opened 2023-06-12
[GPU] Several fixes for format traits

Lyamin-Roman:


sshlyapn:
In general looks good to me
But we can also remove related code from `layout.cpp`:
https://github.com/openvinotoolkit/openvino/blob/6043bcb5c0518fd814691d771fab99f75f5add7a/src/plugins/intel_gpu/src/runtime/layout.cpp#L18-L20
https://github.com/openvinotoolkit/openvino/blob/6043bcb5c0518fd814691d771fab99f75f5add7a/src/plugins/intel_gpu/src/runtime/layout.cpp#L576-L578

vladimir-paramuzov (approved):

vladimir-paramuzov:
> But we can also remove related code from layout.cpp:

I think it's still needed for grouped formats. So let's merge this PR as is for now, and check if we can remove that handling separately

PR #18021, opened 2023-06-13
Disable mmap to check POT

ilyachur:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


p-durandin:
It would be good to add assert into load_mmap_object to be sure that it is not used

PR #18023, opened 2023-06-13
[Snippets] Fixed insertion position iterators

a-sidorova:
### Details:
 - *Fixed Buffer insertion position: `potential_consumers` is unsorted by `linear IR` set. We have to find first expr in `Linear IR` from the set to insert `Buffer` before **all** consumers*
 - *Added `OPENVINO_ASSERT` for all `std::find` if needed*

### Tickets:
 - *N/A*

### DIFF
--- a/src/common/snippets/src/lowered/pass/insert_buffers.cpp
+++ b/src/common/snippets/src/lowered/pass/insert_buffers.cpp
@@ -225,12 +225,26 @@ void InsertBuffers::insertion(LinearIR& linear_ir, const LinearIR::LoopManagerPt
                 }
             }
 
+            // potential_consumers is unsorted by linear IR set.
+            // We have to find first expr in Linear IR from the set to insert Buffer before *all* consumers
+            OPENVINO_ASSERT(!potential_consumers.empty(), "Buffer should have one consumer at least");
+            auto consumer_expr = potential_consumers.begin()->get_expr();
+            if (potential_consumers.size() > 1) {
+                std::set<ExpressionPtr> consumers;
+                for (const auto& port : potential_consumers)
+                    consumers.insert(port.get_expr());
+                const auto it = std::find_if(linear_ir.cbegin(), linear_ir.cend(),
+                                             [&consumers](const ExpressionPtr& expr) { return consumers.count(expr) > 0; });
+                OPENVINO_ASSERT(it != linear_ir.cend(), "Consumer of Buffer has not been found in Linear IR");
+                consumer_expr = *it;
+            }
+
             // We should insert Buffer between first different Loops.
             // Example: Current expr Loop identifies: 3, 2, 1
             //          Target consumers Loop identifies:  3, 4, 6
             //          Need to insert after 2nd Loops
             // Note: All potential consumers must have the same count of first equal Loop identifies and the same count of different last identifies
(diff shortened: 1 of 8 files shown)


src/common/snippets/src/lowered/pass/insert_load_store.cpp
```diff
@@ -67,7 +67,9 @@ bool InsertLoadStore::insert_load(LinearIR& linear_ir, const LinearIR::constExpr
         const auto load = std::make_shared<op::Load>(data_node->output(0), get_count(data_expr->get_output_port_descriptor(0)));
         PortDescriptorUtils::set_port_descriptor_ptr(load->output(0), consumer_input.get_descriptor_ptr()->clone());
         const auto load_expr = linear_ir.create_expression(load, {output_connector});
-        linear_ir.insert(std::find(data_expr_it, linear_ir.cend(), consumer_expr), load_expr);
+        const auto insertion_pos = std::find(data_expr_it, linear_ir.cend(), consumer_expr);
+        OPENVINO_ASSERT(insertion_pos != linear_ir.cend(), "Consumer should be after data producer in Linear IR");
```
IvanNovoselov, line 71: It looks like we should think strategically and implement `find` helpers that will encapsulate these checks. Otherwise this custom-checks approach could lead to considerable amount of code duplication in future. Probably the easiest way is to create `linear_ir.find_before(expression_it, target_expression)` and a similar `linear_ir.find_after(...)`. 

An alternative approach is to perform all-graph validation passes and rely on graph validity afterwards. This validation requires two passes: we enumerate nodes first, and then check that sourse has lower index than any of the consumers for every PortConnector.

What do you think?

src/common/snippets/src/lowered/pass/insert_buffers.cpp
```diff
@@ -163,17 +163,33 @@ void InsertBuffers::insertion(LinearIR& linear_ir, const LinearIR::LoopManagerPt
...
 
+            // potential_consumers is unsorted by linear IR set.
+            // We have to find first expr in Linear IR from the set to insert Buffer before *all* consumers
+            OPENVINO_ASSERT(!potential_consumers.empty(), "Buffer should have one consumer at least");
+            auto consumer_expr = potential_consumers.begin()->get_expr();
+            if (potential_consumers.size() > 1) {
+                std::set<ExpressionPtr> consumers;
+                for (const auto& port : potential_consumers)
+                    consumers.insert(port.get_expr());
+                const auto it = std::find_if(linear_ir.cbegin(), linear_ir.cend(),
```
IvanNovoselov, line 180: It looks really expensive to go through the entire IR just to find the first consumer. Especially considering the fact that it would be repeated for every loop exit (BTW what about loop inputs? don't we have a similar problem there?).
Can we at least start from the source expression itself to avoid passing through all the expr before the source?

src/common/snippets/src/lowered/pass/insert_load_store.cpp
a-sidorova: I fully agree with the both points. Thank you very much for the ideas! I like it.
As for first suggestion, I added these methods.
As for second, I've created the task 113536 for expression numeration. I hope we will implement it in the near future.

a-sidorova:
@IvanNovoselov could you take a look one more time please? Thank you!

IvanNovoselov (approved):

PR #18024, opened 2023-06-13
merge loadnetwork and importnetwork for cpu plugin

xufang-lisa:
### Details:
 - *merge loadnetwork and importnetwork for cpu plugin*

### Tickets:
 - *112767*


dmitry-gorokhov (changes requested):
Please read my initial comment again: https://github.com/openvinotoolkit/openvino/pull/17873#discussion_r1226448791. We shouldn't save all multithreading config within cached model. That's not correct.
The only difference between LoadNetwork and ImportNetwork for now should be logic that takes value of optimal `num_streams`

wangleis (approved):

xipingyan:
@yuxu42 @wenjiew @dmitry-gorokhov This PR is very important for LLM.
1: For 2 nodes CPU, when we set nstream to 1, if we enable cache_dir, it will not work. (This PR can fix this problem)
2: For LLM, we have to set nstream to 1, because it can reduce memory consume.

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -289,22 +287,43 @@ void Engine::GetPerformanceStreams(Config& config, const std::shared_ptr<ngraph:
...
+Config Engine::ReadConfigAndGetPerformanceStreams(const std::map<std::string, std::string>& config,
+                                                  const std::shared_ptr<ngraph::Function>& function,
+                                                  bool imported) {
+    // TODO: Clarify the behavior of SetConfig method. Skip eng_config or not?
+    Config conf = engConfig;
+
+    conf.readProperties(config);
+
+    // import config props from caching model
+    if (imported && !is_cpu_map_available()) {
```
dmitry-gorokhov, line 304: Didn't get why we read optimal number of streams from imported model only in case legacy threading? I would expect we should do it for both scenarious.

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -289,22 +287,43 @@ void Engine::GetPerformanceStreams(Config& config, const std::shared_ptr<ngraph:
...
-
     get_num_streams(streams, ngraphFunc, config);
 
-    hints_props.insert({latency_name, std::to_string(latency_streams)});
-    hints_props.insert({tput_name, std::to_string(config.streamExecutorConfig._streams)});
-    ngraphFunc->set_rt_info(hints_props, "intel_cpu_hints_config");
     config._config[CONFIG_KEY(CPU_THROUGHPUT_STREAMS)] = std::to_string(config.streamExecutorConfig._streams);
 }
 
+Config Engine::ReadConfigAndGetPerformanceStreams(const std::map<std::string, std::string>& config,
```
dmitry-gorokhov, line 295: I would advise to split functions into two separate. 
We are doing this code to merge core config with compiled model config and there is no any logical dependency with streams calculation.
```
 Config conf = engConfig;
 conf.readProperties(config); 
```

src/plugins/intel_cpu/tests/functional/behavior/export_import.cpp
```diff
@@ -35,34 +37,91 @@ std::shared_ptr<ov::Model> MakeMatMulModel() {
...
+        EXPECT_EQ(new_properties_output[3], imported_properties_output[3]);
+        EXPECT_EQ(new_properties_output[4], imported_properties_output[4]);
+        EXPECT_EQ(new_properties_output[5], imported_properties_output[5]);
+        EXPECT_EQ(new_properties_output[6], imported_properties_output[6]);
     }
 }
 
-INSTANTIATE_TEST_CASE_P(smoke_ExportImportTest, ExportOptimalNumStreams, ::testing::Values(std::string("CPU")));
+const std::vector<std::vector<ov::AnyMap>> testing_properties = {
+    {{ov::num_streams(1)}, {ov::num_streams(2)}},
```
dmitry-gorokhov, line 100: Maybe I missed smt, but I don't see coverage of the followig case:  num_streams/num_threads is not provided inside the compilation config -> plugin automatically computes optimal number of streams based on the model -> import model also not specifying num streams via config -> num streams is equal with original model.

xufang-lisa: I have modified test cases, only test one property at a time.

src/plugins/intel_cpu/src/plugin.cpp
xufang-lisa: Reading optimal number of streams from imported model is needed by the legacy threading, for threading 2.0, streams is recaculated each time and reading streams from imported model  is not needed. And this logic will be removed after legacy threading is removed.

src/plugins/intel_cpu/src/plugin.cpp
xufang-lisa: Updated.

src/plugins/intel_cpu/src/plugin.cpp
dmitry-gorokhov: Yes, it is recalculated each time. The problem is: `GetPerformanceStreams` uses  `get_model_prefer_threads` to calculate prefered number of stream based on model itself. This function utilizes common `MemBandwidthPressureTolerance` which works only with common OV opset while model imported from cache is already in CPU opset. In other words might return different results for `compile_model` and `import_model` calls. Not sure if this tests reproduces such behavior.

wangleis: In the test case where only performance hint is set, different results of `get_model_prefer_threads` will cause test failure. So far we do not observe different result in CI checks and local test.

Since `model_perfer_threads` is internal parameters which cannot be changed by user application, we can keep it in cache file and reuse it in `import_model` to avoid this risk. Will update.

xufang-lisa: @dmitry-gorokhov model_prefer_threads which is computed by MemBandwidthPressureTolerance has been saved to cache for compile_model and loaded from cache for import_model.

src/plugins/intel_cpu/src/plugin.h
```diff
@@ -49,6 +49,9 @@ class Engine : public InferenceEngine::IInferencePlugin {
 
     void GetPerformanceStreams(Config &config, const std::shared_ptr<ngraph::Function>& ngraphFunc);
 
+    Config MergeConfig(const std::map<std::string, std::string>& config);
```
dmitry-gorokhov, line 52: I would say this method introduce more confusion than benefit.
I would propose to remove it at all and leave previous logic untouched inside the code:
```
    Config conf = engConfig;
    conf.readProperties(config);
```

dmitry-gorokhov (approved):
Please fix remaining comment

src/plugins/intel_cpu/src/plugin.h
xufang-lisa: ok, updated

PR #18033, opened 2023-06-13
Fixed SpaceToBatch and BatchToSpace for 3d case and add testcases

steve-y:
### Details:
 - *Fixed SpaceToBatch and BatchToSpace ops to use format order not rank*
 - *Added functional testcases*

### Tickets:
 - *111074*

### DIFF
--- a/src/core/src/op/space_to_batch.cpp
+++ b/src/core/src/op/space_to_batch.cpp
@@ -102,7 +102,7 @@ bool ngraph::op::v1::SpaceToBatch::evaluate_space_to_batch(const HostTensorVecto
 
     auto data_shape = data->get_shape();
 
-    if (!(data->get_shape().size() == 4 || data->get_shape().size() == 5)) {
+    if (!(data->get_shape().size() == 3 || data->get_shape().size() == 4 || data->get_shape().size() == 5)) {
         return false;
     }
 
--- a/src/plugins/intel_gpu/src/plugin/ops/batch_to_space.cpp
+++ b/src/plugins/intel_gpu/src/plugin/ops/batch_to_space.cpp
@@ -31,7 +31,7 @@ static void CreateBatchToSpaceOp(Program& p, const std::shared_ptr<ngraph::op::v
 
         std::vector<int32_t> sizes = inConst->cast_vector<int32_t>();
         int32_t default_size = i == 1 ? 1 : 0;
-        for (size_t s = sizes.size(); s < rank; s++) {
+        for (size_t s = sizes.size(); s < format.dimension(); s++) {
             sizes.push_back(default_size);
         }
         tensor_inputs.emplace_back(format, sizes, default_size);
--- a/src/plugins/intel_gpu/src/plugin/ops/space_to_batch.cpp
+++ b/src/plugins/intel_gpu/src/plugin/ops/space_to_batch.cpp
@@ -31,7 +31,7 @@ static void CreateSpaceToBatchOp(Program& p, const std::shared_ptr<ngraph::op::v
 
         std::vector<int32_t> sizes = inConst->cast_vector<int32_t>();
         int32_t default_size = i == 1 ? 1 : 0;
-        for (size_t s = sizes.size(); s < rank; s++) {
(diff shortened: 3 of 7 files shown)


src/core/src/op/space_to_batch.cpp
```diff
@@ -102,7 +102,7 @@ bool ngraph::op::v1::SpaceToBatch::evaluate_space_to_batch(const HostTensorVecto
 
     auto data_shape = data->get_shape();
 
-    if (!(data->get_shape().size() == 4 || data->get_shape().size() == 5)) {
```
wilson-seok, line 105: Can we touch source code in core module directly? I think you should bring reviewer from OpenVINO core team and get confirmation.

steve-y: I invited @ilyachur as a reviewer. Thank you.

ilyachur: Could you please add new tests for batch_to_space/space_to_batch to template plugin? To cover reference implementation: https://github.com/openvinotoolkit/openvino/tree/master/src/plugins/template/tests/functional/op_reference

steve-y: Added functional tests for batch_to_space/space_to_batch to template plugin. Could you please review it?

src/core/src/op/space_to_batch.cpp
```diff
@@ -102,7 +102,7 @@ bool ngraph::op::v1::SpaceToBatch::evaluate_space_to_batch(const HostTensorVecto
...
+    if (!(data->get_shape().size() == 3 || data->get_shape().size() == 4 || data->get_shape().size() == 5)) {
```
ilyachur, line 105: @jane-intel Do we really need this check? Can we use it for other ranks?

src/plugins/intel_gpu/src/plugin/ops/space_to_batch.cpp
```diff
@@ -31,7 +31,7 @@ static void CreateSpaceToBatchOp(Program& p, const std::shared_ptr<ngraph::op::v
 
         std::vector<int32_t> sizes = inConst->cast_vector<int32_t>();
         int32_t default_size = i == 1 ? 1 : 0;
-        for (size_t s = sizes.size(); s < rank; s++) {
+        for (size_t s = sizes.size(); s < format.order().size(); s++) {
```
vladimir-paramuzov, line 34:
```suggestion
        for (size_t s = sizes.size(); s < format.dimension(); s++) {
```

src/plugins/intel_gpu/src/plugin/ops/batch_to_space.cpp
```diff
@@ -31,7 +31,7 @@ static void CreateBatchToSpaceOp(Program& p, const std::shared_ptr<ngraph::op::v
 
         std::vector<int32_t> sizes = inConst->cast_vector<int32_t>();
         int32_t default_size = i == 1 ? 1 : 0;
-        for (size_t s = sizes.size(); s < rank; s++) {
+        for (size_t s = sizes.size(); s < format.order().size(); s++) {
```
vladimir-paramuzov, line 34:
```suggestion
        for (size_t s = sizes.size(); s < format.dimension(); s++) {
```

src/plugins/intel_gpu/src/plugin/ops/space_to_batch.cpp
steve-y: Applied your comments, and thanks.

src/plugins/intel_gpu/src/plugin/ops/batch_to_space.cpp
steve-y: Applied your comments, and thanks.

ilyachur (approved):

src/core/src/op/space_to_batch.cpp
jane-intel: I don't think so. As you can see [here](https://docs.openvino.ai/2023.0/openvino_docs_ops_movement_SpaceToBatch_2.html) -- the input could be nD, n>=2

PR #18034, opened 2023-06-13
[GPU] Add condition check for dynamic shape and onednn_impl in concat_in_place_optimization::match()

wilson-seok:
### Details:
 - Add condition check for dynamic shape and onednn_impl in concat_in_place_optimization::match()
 
### Tickets:
 - 112919

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/primitive_base.hpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/primitive_base.hpp
@@ -88,7 +88,7 @@ struct typed_primitive_impl_ocl : public typed_primitive_impl<PType> {
     template<typename ImplType>
     static std::unique_ptr<primitive_impl> create(const typed_program_node<PType>& arg, const kernel_impl_params& impl_param) {
         // concat buffer fusing for dynamic shape is adaptively applied at runtime. So we need to build dynamic impl at build time.
-        if (impl_param.can_be_optimized() && !(impl_param.typed_desc<concatenation>() && impl_param.is_dynamic())) {
+        if (impl_param.can_be_optimized() && !(impl_param.is_type<concatenation>() && impl_param.is_dynamic())) {
             return make_unique<ImplType>(kernel_selector::kernel_data{});
         }
         auto kernel_params = ImplType::get_kernel_params(ImplType::static_canonicalize_shapes(impl_param));
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/primitive_base.hpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/primitive_base.hpp
@@ -87,6 +87,7 @@ struct typed_primitive_impl_ocl : public typed_primitive_impl<PType> {
 
     template<typename ImplType>
     static std::unique_ptr<primitive_impl> create(const typed_program_node<PType>& arg, const kernel_impl_params& impl_param) {
+        // concat buffer fusing for dynamic shape is adaptively applied at runtime. So we need to build dynamic impl at build time.
         if (impl_param.can_be_optimized() && !(impl_param.typed_desc<concatenation>() && impl_param.is_dynamic())) {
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/src/graph/impls/ocl/primitive_base.hpp
```diff
@@ -86,7 +87,7 @@ struct typed_primitive_impl_ocl : public typed_primitive_impl<PType> {
 
     template<typename ImplType>
     static std::unique_ptr<primitive_impl> create(const typed_program_node<PType>& arg, const kernel_impl_params& impl_param) {
-        if (impl_param.can_be_optimized()) {
+        if (impl_param.can_be_optimized() && !(impl_param.typed_desc<concatenation>() && impl_param.is_dynamic())) {
```
yeonbok, line 90: Could you please add note here? 
// concat buffer fusing for dynamic shape is adaptively applied at runtime. So we need to build dynamic impl at build time.

src/plugins/intel_gpu/tests/unit/passes/prepare_buffer_fusing_test.cpp
```diff
@@ -258,6 +259,157 @@ TEST(prepare_buffer_fusing, in_place_concat_dynamic) {
...
+    cldnn::mem_lock<FLOAT16> input2_ptr(input_memory2, get_test_stream());
+
+    const auto& concat_node_n = net.get_primitive("concat")->get_node();
+    auto concat_mem = net.get_primitive("concat")->output_memory_ptr();
+    auto reorder1_mem = net.get_primitive("reorder1")->output_memory_ptr();
+    auto reorder2_mem = net.get_primitive("reorder2")->output_memory_ptr();
+
+    ASSERT_NE(concat_mem.get(), reorder1_mem.get());
+    ASSERT_NE(concat_mem.get(), reorder2_mem.get());
+    ASSERT_TRUE(concat_node_n.can_be_optimized());
```
yeonbok, line 406: Could you add one more check for concat_inst (not node) with can_be_optimized = false

src/plugins/intel_gpu/tests/unit/passes/prepare_buffer_fusing_test.cpp
```diff
@@ -258,6 +259,157 @@ TEST(prepare_buffer_fusing, in_place_concat_dynamic) {
...
+    cldnn::mem_lock<FLOAT16> input2_ptr(input_memory2, get_test_stream());
+
+    const auto& concat_node_n = net.get_primitive("concat")->get_node();
+    auto concat_mem = net.get_primitive("concat")->output_memory_ptr();
+    auto reorder1_mem = net.get_primitive("reorder1")->output_memory_ptr();
+    auto reorder2_mem = net.get_primitive("reorder2")->output_memory_ptr();
+
+    ASSERT_EQ(concat_mem.get(), reorder1_mem.get());
+    ASSERT_EQ(concat_mem.get(), reorder2_mem.get());
+    ASSERT_TRUE(concat_node_n.can_be_optimized());
```
yeonbok, line 330: Could you check concat_inst (not node) is also can_be_optimized = true

wilson-seok: Done. Thanks!

src/plugins/intel_gpu/tests/unit/passes/prepare_buffer_fusing_test.cpp
wilson-seok: Done. Thanks!

src/plugins/intel_gpu/src/graph/impls/ocl/primitive_base.hpp
wilson-seok: Done. Thanks!

yeonbok (approved):

PR #18048, opened 2023-06-14
[GPU] graceful exit

sungeunk:
### Issue.
- preparation: revert commit (fef04e468a / WA code to avoid hang issue)
- 'Segmentation fault' when the error occurs from OpenCL api. Benchmark_app could not catch the exception because the exception ptr is released before the app catches it.

### Reproduce.
- Model link: https://jira.devtools.intel.com/secure/attachment/3407787/east_resnet_v1_50.fp16.tf.zip
- cmd
```shell
/// need to try increasing the batch number to occur 'CL_OUT_OF_RESOURCES' erorr from the OpenCL api.
benchmark_app -d GPU -m $MODEL -nireq 4 -nstreams 2 --hint none -t 10 -ip u8 --infer_precision f32 -b 32
``` 

### Avoid this issue.
- It needs to move the compiled_model variable out of try-catch block at main.cpp.
- or it copies the exception and throws it at infer_request_wrap.hpp.

### Tickets:
 - 112952

### DIFF
--- a/samples/cpp/benchmark_app/infer_request_wrap.hpp
+++ b/samples/cpp/benchmark_app/infer_request_wrap.hpp
@@ -168,7 +168,12 @@ public:
         std::unique_lock<std::mutex> lock(_mutex);
         _cv.wait(lock, [this] {
             if (inferenceException) {
-                std::rethrow_exception(inferenceException);
+                try {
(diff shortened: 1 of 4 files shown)


isanghao:
This is blocked due to driver stability issue. Once driver gets stabilized, we will merge this PR.

PR #18051, opened 2023-06-14
[GPU] Impl cldnn::condition to support dynamic shape

ahnyoung-paul:
### Details:
 - *Impl CreateIfOp*
 - *Update calc_output_layouts and execute_impl*
 - *Enable gpu unit / functional tests*

### Tickets:
 - *89336*

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/graph/program.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/graph/program.hpp
@@ -146,7 +146,6 @@ public:
     std::vector<program_node*>& get_outputs() {
         return outputs;
     }  // ToDo: redesign reorder-inputs pass to make it const as_well as get_engine and get options
-    std::string get_output_layout_str();
     bool is_loop_body() const { return is_body_program; }
     const nodes_ordering& get_processing_order() const;
     nodes_ordering& get_processing_order();
--- a/src/plugins/intel_gpu/include/intel_gpu/graph/program.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/graph/program.hpp
@@ -146,6 +146,7 @@ public:
     std::vector<program_node*>& get_outputs() {
         return outputs;
     }  // ToDo: redesign reorder-inputs pass to make it const as_well as get_engine and get options
+    std::string get_output_layout_str();
     bool is_loop_body() const { return is_body_program; }
     const nodes_ordering& get_processing_order() const;
     nodes_ordering& get_processing_order();
--- a/src/plugins/intel_gpu/include/intel_gpu/graph/program.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/graph/program.hpp
@@ -125,6 +125,7 @@ public:
     program(engine& engine_ref,
             topology const& topology,
             const ExecutionConfig& config,
+            InferenceEngine::CPUStreamsExecutor::Ptr task_executor,
             bool is_internal = false,
             bool no_optimizations = false,
             bool is_body_program = false);
@@ -237,6 +238,13 @@ public:
                              bool no_optimizations = false,
                              bool is_body_program = false);
     static ptr build_program(engine& engine,
+                             const topology& topology,
+                             const ExecutionConfig& config,
+                             InferenceEngine::CPUStreamsExecutor::Ptr task_executor,
+                             bool is_internal = false,
+                             bool no_optimizations = false,
+                             bool is_body_program = false);
+    static ptr build_program(engine& engine,
                              const std::set<std::shared_ptr<program_node>>& nodes,
                              const ExecutionConfig& config,
                              std::shared_ptr<InferenceEngine::CPUStreamsExecutor> task_executor,
@@ -252,6 +260,8 @@ public:
     ImplementationsCache& get_implementations_cache() const { return *_impls_cache; }
     ICompilationContext& get_compilation_context() const { return *_compilation_context; }
     void cancel_compilation_context();
+
+    static std::shared_ptr<InferenceEngine::CPUStreamsExecutor> make_task_executor(const ExecutionConfig& config);
 
 private:
     uint32_t prog_id = 0;
@@ -306,9 +316,6 @@ private:
     void pre_optimize_graph(bool is_internal);
     void post_optimize_graph(bool is_internal);
     void transfer_memory_to_device();
-
-    InferenceEngine::CPUStreamsExecutor::Config make_task_executor_config(const ExecutionConfig& config, std::string tags = "") const;
-    std::shared_ptr<InferenceEngine::CPUStreamsExecutor> make_task_executor(const ExecutionConfig& config) const;
 
     /*
     ** Analysis functions
--- a/src/plugins/intel_gpu/src/graph/condition.cpp
+++ b/src/plugins/intel_gpu/src/graph/condition.cpp
@@ -139,6 +139,24 @@ static ov::PartialShape resolve_shape(const ov::PartialShape& true_pshape, const
     return ov::PartialShape(new_dims);
 }
 
+static std::vector<layout> resolve_shape(std::vector<layout>& target_list, std::vector<layout>& other_list) {
+    std::vector<layout> resolved_layout;
+    for (size_t i = 0; i < target_list.size(); i++) {
+        auto target = target_list[i];
+        auto other = other_list[i];
+        auto target_pshape  = target.get_partial_shape();
+        auto other_pshape   = other.get_partial_shape();
+        auto target_rank    = target_pshape.rank();
+        auto other_rank     = other_pshape.rank();
+        if (target_rank.get_length() == 0 && other_rank.get_length() == 1) {
+            resolved_layout.push_back({ov::PartialShape{1}, target.data_type, target.format});
+        } else {
+            resolved_layout.push_back(target);
+        }
+    }
+    return resolved_layout;
+}
+
 template<typename ShapeType>
 std::vector<layout> condition_inst::calc_output_layouts(condition_node const& /* node */, kernel_impl_params const& impl_param) {
     if (impl_param.inner_nets.empty()) {
@@ -164,16 +182,20 @@ std::vector<layout> condition_inst::calc_output_layouts(condition_node const& /*
 
         return output_layouts;
     } else {
+        auto layouts_true  = get_output_layouts(get_out_layout_map(impl_param.inner_nets[idx_branch_true]),  impl_param.io_output_maps[idx_branch_true]);
+        auto layouts_false = get_output_layouts(get_out_layout_map(impl_param.inner_nets[idx_branch_false]), impl_param.io_output_maps[idx_branch_false]);
+        const size_t num_outputs = impl_param.output_layouts.size();
+        OPENVINO_ASSERT((num_outputs == layouts_true.size() && num_outputs == layouts_false.size()),
+                            "The number of outputs for each branch should be same!");
+
         auto& memory_deps = impl_param.memory_deps;
         OPENVINO_ASSERT(memory_deps.count(0) > 0, "The count of memory deps should not be zero");
         auto mem_ptr = memory_deps.at(0);
         auto pred = condition_inst::get_pred_from_memory(mem_ptr, impl_param.get_stream());
         if (pred) {
-            auto layouts_true  = get_output_layouts(get_out_layout_map(impl_param.inner_nets[idx_branch_true]),  impl_param.io_output_maps[idx_branch_true]);
-            return layouts_true;
+            return resolve_shape(layouts_true, layouts_false);
         } else {
-            auto layouts_false = get_output_layouts(get_out_layout_map(impl_param.inner_nets[idx_branch_false]), impl_param.io_output_maps[idx_branch_false]);
-            return layouts_false;
+            return resolve_shape(layouts_false, layouts_true);
         }
     }
 }
@@ -197,8 +219,8 @@ Condition primitive is resuing memory with the input.
 */
 condition_inst::typed_primitive_inst(network& network, condition_node const& node)
     : parent(network, node),
-      _net_true(network::allocate_network(node.get_program().get_engine(), node.get_branch_true().inner_program, true)),
-      _net_false(network::allocate_network(node.get_program().get_engine(), node.get_branch_false().inner_program, true)) {
+      _net_true(network::allocate_network(node.get_program().get_engine(), node.get_branch_true().inner_program)),
+      _net_false(network::allocate_network(node.get_program().get_engine(), node.get_branch_false().inner_program)) {
     this->set_inner_networks({_net_true, _net_false});
 }
 
--- a/src/plugins/intel_gpu/src/graph/condition.cpp
+++ b/src/plugins/intel_gpu/src/graph/condition.cpp
@@ -100,6 +100,45 @@ bool condition_inst::get_pred_from_memory(memory::ptr mem, stream& stream) {
     }
 }
 
+static ov::PartialShape resolve_shape(const ov::PartialShape& true_pshape, const ov::PartialShape& false_pshape) {
+    // true_pshape - shape of output from then_body
+    // false_pshape - shape of output from else_body
+    auto then_rank = true_pshape.rank();
+    auto else_rank = false_pshape.rank();
+
+    // if rangs of shapes are not equal or rang of one of them is dynamic function
+    // return shape with dynamic rank
+    if (then_rank.is_dynamic() || else_rank.is_dynamic()) {
+        return ov::PartialShape::dynamic();
+    }
+    if (then_rank.get_length() != else_rank.get_length()) {
+        // Union of scalar and 1D case
+        if (then_rank.get_length() <= 1 && else_rank.get_length() <= 1) {
+            return ov::PartialShape::dynamic(1);
+        } else {
+            return ov::PartialShape::dynamic();
+        }
+    }
+    std::vector<ov::Dimension> new_dims;
+
+    // If rangs are equal each dimesion of then_body output is union with each dimension of
+    // else_body
+    for (auto then_it = true_pshape.cbegin(), else_it = false_pshape.cbegin(); then_it != true_pshape.cend();
+         then_it++, else_it++) {
+        if ((*then_it).is_dynamic() || (*else_it).is_dynamic()) {
+            new_dims.push_back(ov::Dimension::dynamic());
+        } else if (*then_it == *else_it) {
+            new_dims.emplace_back(*then_it);
+        } else {
+            auto dim_min = std::min((*then_it).get_min_length(), (*else_it).get_min_length());
+            auto dim_max = std::max((*then_it).get_min_length(), (*else_it).get_min_length());
+            new_dims.emplace_back(dim_min, dim_max);
+        }
+    }
+
+    return ov::PartialShape(new_dims);
+}
+
 template<typename ShapeType>
 std::vector<layout> condition_inst::calc_output_layouts(condition_node const& node, kernel_impl_params const& impl_param) {
     if (impl_param.inner_nets.empty()) {
@@ -107,12 +146,23 @@ std::vector<layout> condition_inst::calc_output_layouts(condition_node const& no
         auto layouts_true  = get_output_layouts(get_out_layout_map(impl_param.inner_progs[idx_branch_true]),  impl_param.io_output_maps[idx_branch_true]);
         auto layouts_false = get_output_layouts(get_out_layout_map(impl_param.inner_progs[idx_branch_false]), impl_param.io_output_maps[idx_branch_false]);
 
-        if (layouts_true[0].is_static() && layouts_true[0] == layouts_false[0]) {
-            return {layouts_true[0]};
-        } else {
-            OPENVINO_ASSERT(layouts_true[0].get_rank() == layouts_false[0].get_rank(), "dynamic rank is not supported");
-            return {layout{ov::PartialShape::dynamic(layouts_true[0].get_rank()), layouts_true[0].data_type, layouts_true[0].format }};
+        const size_t num_outputs = impl_param.output_layouts.size();
+        OPENVINO_ASSERT((num_outputs == layouts_true.size() && num_outputs == layouts_false.size()),
+                            "The number of outputs for each branch should be same!");
+        std::vector<layout> output_layouts;
+
+        for (size_t i = 0; i < num_outputs; i++) {
+            if (layouts_true[i] == layouts_false[i]) {
+                output_layouts.push_back(layouts_true[i]);
+            } else {
+                OPENVINO_ASSERT(layouts_true[i].data_type == layouts_false[i].data_type, "data type of each branches should be same");
+                OPENVINO_ASSERT(layouts_true[i].format == layouts_false[i].format, "output format of each branches should be same");
+                auto out_layout = resolve_shape(layouts_true[i].get_partial_shape(), layouts_false[i].get_partial_shape());
+                output_layouts.push_back(layout{out_layout, layouts_true[i].data_type, layouts_true[i].format });
+            }
         }
+
+        return output_layouts;
     } else {
         auto& memory_deps = impl_param.memory_deps;
         OPENVINO_ASSERT(memory_deps.count(0) > 0, "The count of memory deps should not be zero");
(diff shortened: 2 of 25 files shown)


src/plugins/intel_gpu/include/intel_gpu/graph/program.hpp
```diff
@@ -146,6 +146,7 @@ struct program {
     std::vector<program_node*>& get_outputs() {
         return outputs;
     }  // ToDo: redesign reorder-inputs pass to make it const as_well as get_engine and get options
+    std::string get_output_layout_str();
```
yeonbok, line 149: Seems redundant, do we need this?

src/plugins/intel_gpu/src/graph/condition.cpp
```diff
@@ -12,6 +12,48 @@
...
+    std::string branch_name = (branch_idx == 0) ? "true" : "false";
+    auto& outputs  = impl_param.inner_progs[branch_idx]->get_outputs();
+    auto& io_output_map  = impl_param.io_output_maps[branch_idx];
+
+    CLDNN_ERROR_NOT_EQUAL(impl_param.desc->id,
+                        "Count of branch_" + branch_name + " outputs",
+                        io_output_map.size(),
+                        "expected outputs size",
+                        1,
+                        "Branch_" + branch_name + " should have one output.");
```
yeonbok, line 25: Is this temporal limitation ?

src/plugins/intel_gpu/src/graph/condition.cpp
```diff
@@ -12,6 +12,48 @@
...
+
+    auto inner_prim_id = io_output_map.at(0);
+    for (size_t idx = 0; idx < outputs.size(); idx++) {
+        if (outputs[idx]->id() == inner_prim_id) {
+            return outputs.at(idx)->get_output_layout();
+        }
+    }
+    OPENVINO_THROW("Not found output with prim_id: ", inner_prim_id);
+}
```
yeonbok, line 55: Seems that we can merge the above two util functions.

src/plugins/intel_gpu/src/graph/graph_optimizer/update_loop_primitive_map.cpp
```diff
@@ -14,18 +15,27 @@ using namespace cldnn;
 
 void update_loop_primitive_map::run(program& p) {
```
yeonbok, line 16: Now that it is not only updating loop, how about modify the function name to update_inner_program_io_map?

src/plugins/intel_gpu/src/graph/impls/common/condition.cpp
```diff
@@ -36,16 +36,24 @@ struct condition_impl : typed_primitive_impl<condition> {
...
-            memory_to_copy = execute_branch(instance.get_net_false(), instance.result_id(), instance.input_memory_ptr());
-        // just copy memory
-        mem_lock<float, mem_lock_type::read> inp_ptr{memory_to_copy, instance.get_network().get_stream()};
-        mem_lock<float, mem_lock_type::write> out_ptr{instance.output_memory_ptr(), instance.get_network().get_stream()};
-        std::copy(inp_ptr.begin(), inp_ptr.end(), out_ptr.begin());
+        auto pred = condition_inst::get_pred_frem_memory(instance.pred_memory_ptr(), instance.get_network().get_stream());
+        network::ptr executed_net = instance.get_inner_networks(pred);
+
+        executed_net->execute({});
+
```
yeonbok, line 43: We'll need to handle sync for the completion of inner program.
Please refer to the new change in this PR: https://github.com/openvinotoolkit/openvino/pull/17976/files

src/plugins/intel_gpu/src/graph/include/primitive_inst.h
```diff
@@ -265,6 +268,8 @@ class primitive_inst {
 
     virtual void update_output_memory() {}
 
+    bool has_inner_body();
```
yeonbok, line 271: just more generally, inner_network

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -719,6 +722,13 @@ event::ptr primitive_inst::execute(const std::vector<event::ptr>& events) {
             }
         }
 
+        if (has_inner_body()) {
```
yeonbok, line 725: Can't we do this in update_shape?
Like:
`in update_shape() {
   if _node.is_type<condition>()
       execute inner netowkr 
       wait_for_completion of inner networks 
      ...

}`

src/plugins/intel_gpu/tests/functional/subgraph_tests/condition.cpp
```diff
@@ -0,0 +1,591 @@
...
+
+INSTANTIATE_TEST_SUITE_P(smoke_ConditionGPUTest_dynamic02, ConditionLayerGPUTest,
+                testing::Combine(
+                    testing::ValuesIn(dynamicInputShapes),                          // input shapes
+                    testing::ValuesIn(innerBodyTypes_for_constant),                 // inner body type
+                    testing::ValuesIn(netPrecisions),                               // network precision
+                    testing::ValuesIn({TestModelGenerator::CondTypes::CONSTANT}),   // cond type
+                    testing::ValuesIn(condExecutionValues),                         // cond execution value
+                    testing::Values<std::string>(CommonTestUtils::DEVICE_GPU)),     // device type
+                ConditionLayerGPUTest::getTestCaseName);
```
yeonbok, line 590: Could we have a nested branches' teest?
also please add other dimensions' tests  (D=3 && D>=5) examples too.

src/plugins/intel_gpu/src/graph/impls/common/condition.cpp
yeonbok: Or using get_output().get_memory() will enable sync.

src/plugins/intel_gpu/include/intel_gpu/graph/program.hpp
ahnyoung-paul: No, it is just for debug. removed it. thanks.

src/plugins/intel_gpu/src/graph/condition.cpp
ahnyoung-paul: Remove the limitation

src/plugins/intel_gpu/src/graph/condition.cpp
ahnyoung-paul: Refactoring the function. please review it again. thanks.

src/plugins/intel_gpu/src/graph/impls/common/condition.cpp
ahnyoung-paul: Fixed these codes as I commented offline. please review it again. thanks

src/plugins/intel_gpu/src/graph/include/primitive_inst.h
ahnyoung-paul: Remove the function.

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
ahnyoung-paul: Override execute() function for condition_inst.

src/plugins/intel_gpu/src/plugin/ops/condition.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <cpp/ie_cnn_network.h>
+
+#include "ngraph/op/if.hpp"
+#include "ngraph/op/constant.hpp"
+#include "ngraph/op/util/sub_graph_base.hpp"
+#include "transformations/utils/utils.hpp"
+#include "ie_ngraph_utils.hpp"
+
+#include "intel_gpu/primitives/condition.hpp"
+#include "intel_gpu/primitives/mutable_data.hpp"
```
vladimir-paramuzov, line 17: It looks unused

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -623,33 +625,33 @@ event::ptr primitive_inst::execute(const std::vector<event::ptr>& events) {
...
-
-            for (auto& d : _deps) {
-                if (!d.first->get_node().is_type<data>()) {
-                    auto allocated_mem = d.first->output_memory_ptr();
-                    auto actual_input_layout = d.first->get_output_layout();
-                    auto& engine = _network.get_engine();
-                    // Need to use actual layout, not the fake aligned memory layout
-                    auto actual_mem = engine.reinterpret_buffer(*allocated_mem, actual_input_layout);
-                    subgraph->set_input_data(d.first->id(), actual_mem);
+            if (!is_valid_fusion()) {
```
vladimir-paramuzov, line 628: Looks like unexpected indentation change

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -1222,6 +1222,15 @@ void network::execute_impl(const std::vector<event::ptr>& events) {
         }
 
         execute_primitive(inst, events);
+        // {
+        //     get_stream().finish();
+        //     const std::string layer_name = inst->id();
+        //     for (size_t i = 0; i < get_primitive(inst->id())->outputs_memory_count(); i++) {
+        //         check_mem(get_primitive(inst->id())->output_memory_ptr(i),
```
vladimir-paramuzov, line 1229: Dead code

src/plugins/intel_gpu/src/graph/condition.cpp
```diff
@@ -69,23 +147,67 @@ Condition primitive is resuing memory with the input.
...
-                                          compare_with_offster_tensor,
-                                          "input tensor",
-                                          input_tensor,
-                                          "Offset is too big.");
+      _net_true(network::allocate_network(node.get_program().get_engine(), node.get_branch_true().inner_program, true)),
+      _net_false(network::allocate_network(node.get_program().get_engine(), node.get_branch_false().inner_program, true)) {
+    this->set_inner_networks({_net_true, _net_false});
+}
+
+event::ptr condition_inst::execute(const std::vector<event::ptr>& events) {
```
vladimir-paramuzov, line 155: Why is it needed? What's wrong with the common primitive_inst::execute() ?

src/plugins/intel_gpu/src/graph/condition.cpp
```diff
@@ -20,38 +53,83 @@ GPU_DEFINE_PRIMITIVE_TYPE_ID(condition)
...
+
+        if (layouts_true[0].is_static() && layouts_true[0] == layouts_false[0]) {
+            return {layouts_true[0]};
+        } else {
+            OPENVINO_ASSERT(layouts_true[0].get_rank() == layouts_false[0].get_rank(), "dynamic rank is not supported");
+            return {layout{ov::PartialShape::dynamic(layouts_true[0].get_rank()), layouts_true[0].data_type, layouts_true[0].format }};
+        }
+    } else {
+        auto& memory_deps = impl_param.memory_deps;
+        OPENVINO_ASSERT(memory_deps.count(0) > 0, "");
```
vladimir-paramuzov, line 118: Add some meaningful error message here

src/plugins/intel_gpu/src/graph/condition.cpp
```diff
@@ -20,38 +53,83 @@ GPU_DEFINE_PRIMITIVE_TYPE_ID(condition)
...
+    if (impl_param.inner_nets.empty()) {
+        OPENVINO_ASSERT(impl_param.inner_progs.empty() == false, "The count of inner programs should not be zero");
+        auto layouts_true  = get_output_layouts(get_out_layout_map(impl_param.inner_progs[idx_branch_true]),  impl_param.io_output_maps[idx_branch_true]);
+        auto layouts_false = get_output_layouts(get_out_layout_map(impl_param.inner_progs[idx_branch_false]), impl_param.io_output_maps[idx_branch_false]);
+
+        if (layouts_true[0].is_static() && layouts_true[0] == layouts_false[0]) {
+            return {layouts_true[0]};
+        } else {
+            OPENVINO_ASSERT(layouts_true[0].get_rank() == layouts_false[0].get_rank(), "dynamic rank is not supported");
+            return {layout{ov::PartialShape::dynamic(layouts_true[0].get_rank()), layouts_true[0].data_type, layouts_true[0].format }};
```
vladimir-paramuzov, line 114: As I can see, ngraph shape infer has a bit more complex logic for shape inference

src/plugins/intel_gpu/src/graph/graph_optimizer/update_inner_program_io_map.cpp
```diff
@@ -0,0 +1,41 @@
+// Copyright (C) 2018-2023 Intel Corporation
```
yeonbok, line 1: nit: 2023

src/plugins/intel_gpu/src/graph/graph_optimizer/update_loop_primitive_map.cpp
ahnyoung-paul: Fixed it.

src/plugins/intel_gpu/src/graph/condition.cpp
ahnyoung-paul: updated it.

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
ahnyoung-paul: Yes, you're right. I remove this changes now. thanks.

src/plugins/intel_gpu/src/graph/graph_optimizer/update_inner_program_io_map.cpp
ahnyoung-paul: I remained 2018 because it's just renamed from updated_loop_primitive_map. Should I remove 2018?

src/plugins/intel_gpu/src/graph/network.cpp
ahnyoung-paul: Yes, you're right. it is the code for debug. I removed it. thanks.

src/plugins/intel_gpu/tests/functional/subgraph_tests/condition.cpp
ahnyoung-paul: I added more test cases including other dimension (D=3 && D=5) and nested if case

src/plugins/intel_gpu/src/plugin/ops/condition.cpp
ahnyoung-paul: Remove redundant codes. thanks.

src/plugins/intel_gpu/src/graph/condition.cpp
ahnyoung-paul: I tried to find shape_infer for ngraph::If operator but I can't find it. So I understood that it calculated output layout size in ov::op::v8::If::validate_and_infer_types() without shape_infer function. 
But, If you let me know what I'm missing where ngraph::If's shape_infer function is, I will update it. thanks.

vladimir-paramuzov: https://github.com/openvinotoolkit/openvino/blob/master/src/core/src/op/if.cpp#L59

src/plugins/intel_gpu/src/graph/condition.cpp
ahnyoung-paul: I create this custom execute() function to run cldnn::condition without update shape and update impl because the output layout of  condition will be got from inner network's result layout after inner network execution.

At first, I add condition to avoid update shape and impl when current primitive_inst is condition_inst. but the condition code makes complicated code for primitive_inst::execute(). So I created new custom execute() function to run it without update impl/shape.

vladimir-paramuzov: I just don't like that we have a partial duplication of the logic here as the code itself is not that trivial, so ideally we should reuse it as much as possible.
Also, I'm curious if dynamic loop primitive will require similar trick (updating outer primitive shape after completion of body network. If yes, then we definitely need to come up with better solution. 

Overall, I don't fully understand why condition_impl::execute() method can't do update of condition_inst object passed as a parameter. It already updates output_memory, so isn't it just a mater of updating `_impl_params` as well?

src/plugins/intel_gpu/src/graph/condition.cpp
ahnyoung-paul: @vladimir-paramuzov , Do you means that calculate output using resolve_shape like shape_infer?

vladimir-paramuzov: as I can see, resolve_shape is internal method and not exposed in headers, so you can't use it. My point is that resolve_shape does a bit more carful shapes handling, for example it keeps static dimensions intact if shape is partially static and tries to merge ranges for each dynamic dimension which may be useful for us too

src/plugins/intel_gpu/src/graph/condition.cpp
ahnyoung-paul: Okay, I will find better solution for this issue. thanks.

Currently, In condition_impl::execute(), condition_inst update output memory after inner network execution. at latest update, I move the code to update output memory of condition_inst into the condition_impl::execute() because output memory update issue.

yeonbok: How about add a pre-stage like "execute_inner_networks()" before update_shape() in common primitive_inst;:execute()?

src/plugins/intel_gpu/src/graph/condition.cpp
ahnyoung-paul: update it according to your comments. please review it again. thanks.

src/plugins/intel_gpu/src/graph/condition.cpp
yeonbok: Or I think we can do execute() like 
if (is_dynamic() || !has_inner_networks()) {
    update_shape()
    update_impl()
    update_weight()
} 
...
execute() // update_shape of current inst will be done inside the impl execution of condition & loop.

ahnyoung-paul: Update codes to remove custom condition_inst::execute(). please review it again. thanks.

src/plugins/intel_gpu/src/graph/include/primitive_inst.h
```diff
@@ -190,7 +190,7 @@ class primitive_inst {
 
     memory::ptr shape_info_memory_ptr() const { return _shape_info_memory; }
 
-    event::ptr execute(const std::vector<event::ptr>& events);
+    virtual event::ptr execute(const std::vector<event::ptr>& events);
```
vladimir-paramuzov, line 193: This is not needed anymore

src/plugins/intel_gpu/src/plugin/ops/condition.cpp
```diff
@@ -0,0 +1,89 @@
...
+        for (auto& out : body_network.getOutputsInfo()) {
+            const auto& model_output = internal_body->get_output_op(tidx++);
+            auto output_data_type = InferenceEngine::details::convertPrecision(model_output->get_output_tensor(0).get_element_type());
+            if (out.second->getPrecision() != output_data_type)
+                out.second->setPrecision(output_data_type);
+        }
+    }
+
+    auto config = p.get_config();
+    config.set_property(ov::intel_gpu::max_dynamic_batch(1));
```
vladimir-paramuzov, line 42: Don't we need to change `compilation_num_threads` too? Also, I think we should consider sharing task executor of inner and outer network (may be done separately)

ahnyoung-paul: the case to share task executor and impl_cache will be updated at next step.

src/plugins/intel_gpu/src/graph/include/primitive_inst.h
ahnyoung-paul: Remove it. thanks.

src/plugins/intel_gpu/src/plugin/ops/condition.cpp
ahnyoung-paul: I think the config from Program::get_config() already has the property for **compilation_num_threads**. 
So I just to add the property for max_dynamic_batch to disable dynamic batch for inner program build.
Is it necessary additional setting for **compilation_num_threads** ?

vladimir-paramuzov, line 42: I've tried your branch, and it seems that there's an issue with allow_new_shape_infer() property for inner programs when one of the branches have single `Constant` node. The value for the flag is calculated in `Program::BuildProgram` so it doesn't depend on the outer program anyhow and may introduce mix of old and new shape infers

vladimir-paramuzov: > I think the config from Program::get_config() already has the property for compilation_num_threads.

It has, but I assume that the value is same as for outer network which means that each inner program will create task executor with `compilation_num_threads` threads, so we'll have `(num_inner_networks + 1) * compilation_num_threads` threads created which is probably not what we want.

vladimir-paramuzov: And it seems that there's an issue with `propagate_constants` pass. I've got
```
[ ERROR ] Check '(mem.get_layout() == layout) || layout.is_dynamic()' failed at src/plugins/intel_gpu/src/graph/primitive_inst.cpp:144:
[GPU] Unexpected layout of input memory for parameter:Parameter_3879 node!
Node layout: f16:bfyx:1x4:nopad
Memory layout: f16:bfyx:2125x4:nopad
```
error on `propagate_constants` call for some mask-rcnn version with If primitive

ahnyoung-paul: Fixed codes to share single task_executor between all programs including inner programs. thanks.

ahnyoung-paul: I succeed to run the model by fixing calc_output_layouts for single const inner network. thanks

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -345,6 +345,13 @@ network::network(engine& engine,
                  bool is_internal)
     : network(program::build_program(engine, topo, config, is_internal), config, engine.create_stream(config), is_internal) {}
 
+network::network(engine& engine,
+                 const topology& topo,
+                 const ExecutionConfig& config,
+                 std::shared_ptr<InferenceEngine::CPUStreamsExecutor> task_executor,
+                 bool is_internal)
```
yeonbok, line 352: Do we need this additional constructor? Can't we merge this to the previous one

src/plugins/intel_gpu/src/plugin/program.cpp
```diff
@@ -301,12 +304,66 @@ Program::Program(cldnn::engine& engine, const ExecutionConfig& config,
...
         , m_engine(engine)
         , queryMode(false) {
+    m_task_executor = cldnn::program::make_task_executor(m_config);
     if (inputs != nullptr)
         m_networkInputs = *inputs;
     if (outputs != nullptr)
         m_networkOutputs = *outputs;
 }
 
+Program::Program(InferenceEngine::CNNNetwork& network,
```
yeonbok, line 314: Can't we merge this constructor too?

yeonbok (approved):

src/plugins/intel_gpu/src/graph/network.cpp
ahnyoung-paul: Fixed it. thanks.

src/plugins/intel_gpu/src/plugin/program.cpp
ahnyoung-paul: Fixed it. thanks.

src/plugins/intel_gpu/src/graph/network.cpp
ahnyoung-paul: I will remove it at next step. thanks.

src/plugins/intel_gpu/src/plugin/program.cpp
ahnyoung-paul: I will remove it at next step. thanks

ahnyoung-paul: Fixed it.

src/plugins/intel_gpu/src/graph/network.cpp
ahnyoung-paul: Fixed it.

PR #18063, opened 2023-06-14
[DO NOT MERGE] Trace ov::enable_mmap

vurusovs:
### Tickets:
 - 112712


vurusovs:
Close due issue is already debugged

PR #18069, opened 2023-06-14
ScatterElementsUpdate-12 core op

tomdol:
### Tickets:
 - 111101


src/core/tests/type_prop/scatter_elements_update.cpp
```diff
@@ -97,83 +101,100 @@ TEST(type_prop, scatter_elements_update_default_ctor) {
...
-    const auto axis = make_shared<op::Constant>(element::i16, Shape{}, 0);
+    const auto axis = make_shared<op::v0::Constant>(element::i16, Shape{}, 0);
 
-    const auto shape_of_u = std::make_shared<op::ShapeOf>(std::make_shared<op::Parameter>(element::i64, updates_shape));
+    const auto shape_of_u =
+        std::make_shared<op::v0::ShapeOf>(std::make_shared<op::v0::Parameter>(element::i64, updates_shape));
     const auto scatter = make_shared<op::v3::ScatterElementsUpdate>(data, indices, shape_of_u, axis);
 
-    auto param = std::make_shared<op::Parameter>(element::f32, PartialShape{1});
+    auto param = std::make_shared<op::v0::Parameter>(element::f32, PartialShape{1});
```
praasz, line 116: As this scatter operator is used in shape-of sup-graphs (in models) the bound and label evaluation should be implemented to accurate calculate shapes. As new operator  not implements these functions this test should fail. I think this evaluate  bounds and label should be provided for v12 also, as usage of this op can be similar as for previous version.

src/core/shape_inference/include/scatter_elements_update_shape_inference.hpp
```diff
@@ -9,13 +9,13 @@
 #include "utils.hpp"
 
 namespace ov {
-namespace op {
-namespace v3 {
 
+namespace op {
+namespace {
```
praasz, line 14: Cosmetic make add name for namespace like `scatter_el_update` and then `shape_infer_impl`->`shape_infer`.
Or maybe introduce the Scatter base class and convert this common shape infer to 

```
namespace ov{
namespace op{
shape_infer(const util::ScatterElementUpdateBase* op...
```

Then the template wrappers for v3,v12 I think could be removed and Scatters can call common function as they will be children classes.

Maybe it will help with implementation of evaluates functions also (for reduction mode NONE I think it works as v3).

src/core/shape_inference/include/scatter_elements_update_shape_inference.hpp
praasz, line 1: check file 'shape_inference.cpp' in CPU plugin and update factory registry with bot version of scatters.
`IShapeInferCommonFactory::registry`

Update tests in file `scatter_elements_update_shape_inference_test.cpp` for v12 version.

src/core/src/op/scatter_elements_update.cpp
```diff
@@ -273,3 +280,73 @@ bool op::v3::ScatterElementsUpdate::evaluate_label(TensorLabelVector& output_lab
...
+}
+
+void op::v12::ScatterElementsUpdate::validate_and_infer_types() {
+    OV_OP_SCOPE(v12_ScatterElementsUpdate_validate_and_infer_types);
+    validate_scatter_element_types(this);
+
+    OPENVINO_SUPPRESS_DEPRECATED_START
+    const auto output_shape = shape_infer(this, get_node_input_partial_shapes(*this)).front();
+    OPENVINO_SUPPRESS_DEPRECATED_END
+    set_output_type(0, get_input_element_type(0), output_shape);
```
praasz, line 311: Consider set the merge result instead input_0 type (maybe is possible to change it for v3 also). In case when data is dynamic and updates are e.g. i32 the merge result of element type will be i32 and will give more accurate output.

The dynamic type can be only set when both are dynamic.

src/core/include/openvino/op/scatter_elements_update.hpp
```diff
@@ -44,5 +44,65 @@ class OPENVINO_API ScatterElementsUpdate : public Op {
...
+
+    bool visit_attributes(AttributeVisitor& visitor) override;
+
+    Reduction get_reduction() const {
+        return m_reduction;
+    }
+
+    void set_reduction(const Reduction reduction) {
+        m_reduction = reduction;
+    }
```
rkazants, line 78: I don't think that we need public `set` method , not good to change it on fly. How do you think?

src/core/src/op/scatter_elements_update.cpp
```diff
@@ -16,6 +16,31 @@
...
+    const element::Type& updates_et = op->get_input_element_type(2);
+    const element::Type& axis_et = op->get_input_element_type(3);
+
+    NODE_VALIDATION_CHECK(op,
+                          indices_et.is_integral(),
+                          "Indices element type must be integral_number, but is: ",
+                          indices_et);
+
+    NODE_VALIDATION_CHECK(op, axis_et.is_integral(), "Axis element type must be integral_number, but is: ", axis_et);
+
```
rkazants, line 32: One thing is that not all data types are supported in case of reduction mode. For example, what does it mean `sum` reduction for boolean type

src/core/include/openvino/op/scatter_elements_update.hpp
praasz: It can be required e.g for GPU as they create operator using default ctor and then set attributes if required.

tomdol: As Paweł said, the setters and getters are used to manipulate the attribute values after the op has been constructed with the default ctor. You can have a look at the whole attribute visitor machinery as well as `NodeFactory` class, this is where those methods are used (and possibly in many places in the plugins). Each op in each opset follows this approach so I don't think we can skip the setters just for the op which is the subject of this particular PR.

src/core/include/openvino/op/scatter_elements_update.hpp
```diff
@@ -44,5 +44,65 @@ class OPENVINO_API ScatterElementsUpdate : public Op {
     bool evaluate_scatter_element_update(const HostTensorVector& outputs, const HostTensorVector& inputs) const;
 };
 }  // namespace v3
+namespace v12 {
+class OPENVINO_API ScatterElementsUpdate : public Op {
```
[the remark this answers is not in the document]
tomdol: At this point I don't see the benefit of inheriting from v3. Perhaps when the reference implementation is added and the v12 gets the evaluate() method this will be changed. In the past inheritance from the previous versions of the op caused problems while working on the opset and sometimes led to a major refactor with base classes creation so I'd rather keep v3 and v12 separate.

src/core/src/op/scatter_elements_update.cpp
mitruska: In my opinion, boolean type shouldn't be a problem for ``sum``, ``prod``, ``min``, ``max`` and it is supported case by frameworks, additional info can be added to the spec:
* For ``boolean`` type of ``data`` input, *reduction* ``sum``, ``prod`` behaves like logical ``OR``, ``AND`` accordingly, but there is no implementation for ``mean``.

rkazants: `mean` mode is not clear for all integer types.:)
@tomdol, @mitruska, can you please investigate behaviors of different frameworks for these conner cases?

src/core/tests/type_prop/scatter_elements_update.cpp
```diff
@@ -8,37 +8,41 @@
 #include "util/type_prop.hpp"
 
 using namespace std;
-using namespace ngraph;
+using namespace ov;
 using namespace testing;
 
-TEST(type_prop, scatter_elements_update_output_shape) {
+template <class T>
+class ScatterElementsUpdateTest : public ::testing::Test {};
```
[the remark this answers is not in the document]
tomdol: `make_op` didn't work for me so I left it as is. More tests of this type inherit from the `Test` class directly.

src/core/tests/type_prop/scatter_elements_update.cpp
tomdol: Fixed by moving the evaluate call to the base class.

src/core/shape_inference/include/scatter_elements_update_shape_inference.hpp
tomdol: Shouldn't it be added along with the official support of this op in the plugin later?

src/core/src/op/scatter_elements_update.cpp
tomdol: The `mean` reduction type is only present in pytorch and it always rounds down the calculated value to the nearest integer. We've added this reduction type to OV to cover this case in PT FE and I think it should be implemented accordingly.

src/core/shape_inference/include/scatter_elements_update_shape_inference.hpp
praasz: I think it can be added now, as it will use shape_infer function instead using fallback path which use op cloning etc., and is slower.

src/core/src/op/util/scatter_elements_update_base.cpp
```diff
@@ -0,0 +1,262 @@
...
+                          "Data type and updates type are required to be the same. ",
+                          "Got: ",
+                          data_et,
+                          " and: ",
+                          updates_et);
+    const auto output_shape = shape_infer(this, get_node_input_partial_shapes(*this)).front();
+    OPENVINO_SUPPRESS_DEPRECATED_END
+    element::Type out_et = get_input_element_type(0);
+    std::ignore = element::Type::merge(out_et, get_input_element_type(0), get_input_element_type(2));
+    set_output_type(0, out_et, output_shape);
```
praasz, line 52:
```suggestion
    set_output_type(0, merged_type, output_shape);
```

praasz (approved):

PR #18070, opened 2023-06-14
[CPU] [ARM] Enable Deconvolution SLT tests on ARM 

alvoron:


EgorDuplensky (approved):
@dmitry-gorokhov 
I would propose to rename the files to deconvolution.cpp
To align with the source doe.

dmitry-gorokhov:
> @dmitry-gorokhov I would propose to rename the files to deconvolution.cpp To align with the source doe.

I agree. Lets rename

dmitry-gorokhov:
I still see blocked layout instances are placed as common. Obvioulsy that should be enabled for x64 only.

PR #18083, opened 2023-06-15
[PT FE]: use example input for dtype and rank detection

eaidova:
### Details:
 - *use example_input for rank and input dtype detection*
 - *support unordered dict inputs*

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
+++ b/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
@@ -9,6 +9,7 @@ from openvino.frontend.pytorch.py_pytorch_frontend import _Type as DecoderType
 from openvino.runtime import op, PartialShape, Type as OVType, OVAny, Shape
 
 import typing
+from packaging.version import parse
 import torch
 import numpy as np
 
@@ -133,16 +134,24 @@ class TorchScriptPythonDecoder (Decoder):
         import inspect
 
         def prepare_example_inputs(inputs, input_signature):
+            is_torch_2 = parse(torch.__version__) >= parse("2.0.0")
             if inputs is not None:
                 if isinstance(inputs, dict):
+                    ordered_inputs = []
                     if input_signature is not None:
                         used_sign = []
                         for key in input_signature:
                             if key not in inputs:
                                 continue
+                            ordered_inputs.append(inputs[key])
                             used_sign.append(key)
                         input_signature = used_sign
-                    return {"example_kwarg_inputs": inputs}, input_signature
+                    else:
+                        ordered_inputs = list(inputs.values())
+                    if is_torch_2:
+                        return {"example_kwarg_inputs": inputs}, input_signature
+                    else:
+                        inputs = ordered_inputs
                 if isinstance(inputs, torch.Tensor):
                     inputs = [inputs]
                 
--- a/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
+++ b/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
@@ -133,24 +133,21 @@ class TorchScriptPythonDecoder (Decoder):
         import inspect
 
         def prepare_example_inputs(inputs, input_signature):
+            trace_args = {}
             if inputs is not None:
                 if isinstance(inputs, dict):
                     if input_signature is not None:
-                        ordered_inputs = []
                         used_sign = []
                         for key in input_signature:
                             if key not in inputs:
                                 continue
-                            ordered_inputs.append(inputs[key])
                             used_sign.append(key)
-                        inputs = ordered_inputs
                         input_signature = used_sign
-                    else:
-                        inputs = list(inputs.values())
-                        input_signature = input_signature[:len(inputs)]
+                    return {"example_kwarg_inputs": inputs}, input_signature
                 if isinstance(inputs, torch.Tensor):
                     inputs = [inputs]
-            return inputs, input_signature
+                
+            return {"example_inputs": inputs}, input_signature
 
         if isinstance(pt_module, torch.nn.Module):
             pt_module.eval()
@@ -160,14 +157,14 @@ class TorchScriptPythonDecoder (Decoder):
             if example_inputs is None:
                 scripted = torch.jit.script(pt_module)
             else:
-                inputs, input_signature = prepare_example_inputs(example_inputs, input_signature)
+                input_paremeters, input_signature = prepare_example_inputs(example_inputs, input_signature)
                 try:
-                    scripted = torch.jit.trace(pt_module, inputs)
+                    scripted = torch.jit.trace(pt_module, **input_paremeters)
                 except Exception:
                     try:
                         scripted = torch.jit.script(pt_module)
                     except Exception:
-                        scripted = torch.jit.trace(pt_module, inputs, strict=False)
+                        scripted = torch.jit.trace(pt_module, **input_paremeters, strict=False)
             skip_freeze = False
             for n in scripted.inlined_graph.nodes():
                 # TODO: switch off freezing for all traced models
--- a/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
+++ b/src/bindings/python/src/openvino/frontend/pytorch/decoder.py
@@ -133,7 +133,6 @@ class TorchScriptPythonDecoder (Decoder):
         import inspect
 
         def prepare_example_inputs(inputs, input_signature):
-            trace_args = {}
             if inputs is not None:
                 if isinstance(inputs, dict):
                     if input_signature is not None:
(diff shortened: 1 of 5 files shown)


src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -160,14 +157,14 @@ def prepare_example_inputs(inputs, input_signature):
             if example_inputs is None:
                 scripted = torch.jit.script(pt_module)
             else:
-                inputs, input_signature = prepare_example_inputs(example_inputs, input_signature)
+                input_paremeters, input_signature = prepare_example_inputs(example_inputs, input_signature)
```
eaidova, line 160:
```suggestion
                input_parameters, input_signature = prepare_example_inputs(example_inputs, input_signature)
```

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -160,14 +157,14 @@ def prepare_example_inputs(inputs, input_signature):
...
                 try:
-                    scripted = torch.jit.trace(pt_module, inputs)
+                    scripted = torch.jit.trace(pt_module, **input_paremeters)
```
eaidova, line 162:
```suggestion
                    scripted = torch.jit.trace(pt_module, **input_parameters)
```

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -160,14 +157,14 @@ def prepare_example_inputs(inputs, input_signature):
...
                 except Exception:
                     try:
                         scripted = torch.jit.script(pt_module)
                     except Exception:
-                        scripted = torch.jit.trace(pt_module, inputs, strict=False)
+                        scripted = torch.jit.trace(pt_module, **input_paremeters, strict=False)
```
eaidova, line 167:
```suggestion
                        scripted = torch.jit.trace(pt_module, **input_parameters, strict=False)
```

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -136,21 +136,17 @@ def prepare_example_inputs(inputs, input_signature):
...
                             if key not in inputs:
                                 continue
-                            ordered_inputs.append(inputs[key])
                             used_sign.append(key)
-                        inputs = ordered_inputs
                         input_signature = used_sign
-                    else:
-                        inputs = list(inputs.values())
-                        input_signature = input_signature[:len(inputs)]
+                    return {"example_kwarg_inputs": inputs}, input_signature
```
slyalin, line 145: I tried this for dolly with 3 inputs (ids, mask, past_keys_values) and it doesn't work: `jit.trace` complains about too many arguments passed (just ignoring past_key_values input for some reason in the original model signature). The only thing that works is to wrap model.forward to a separate function, define all inputs as positional arguments, and pass it in `example_inputs` instead of `example_kwarg_inputs`. I don't know why it happens, but I cannot reproduce the same issue with your test (where three inputs exist: x, y, z).

slyalin: Looks like I was able to reproduce: it works only for tensor inputs, and doesn't work for tuples. For some reason, they just cannot be passed in this way. Modified test model (`class NeuralNetwork` defined below) to expect tuple as one of the inputs (z) and pass it in `example_kwarg_inputs` to `jit.trace` and it failed with error: `Exception:
	forward() expected at most 3 argument(s) but received 4 argument(s). Declaration: forward(__torch__.___torch_mangle_1447.NeuralNetwork self, Tensor x, Tensor y) -> Tensor`.

eaidova: which pytorch version do you use?
I see in optimum code that they uses example_kwarg_input of pkv only if version 2.0.1 and on previous versions it does not work

slyalin: `torch.__version__` gives `2.0.1+cu117`

tests/layer_tests/mo_python_api_tests/test_mo_convert_pytorch.py
```diff
@@ -595,6 +637,39 @@ def forward(self, x):
...
+    example_input = {"x": torch.zeros((1,3,10,10)), "z": torch.ones((1,3,10,10))}
+    ref_model = make_ref_pt_model_with_optional_inputs([-1, -1, -1, -1], z_exist=True)
+    return net, ref_model, {"example_input": example_input}
+
+
+def create_pytorch_module_with_optional_inputs_case3(tmp_dir):
+    net = make_pt_model_with_optional_input()
+    example_input = {"x": torch.zeros((1,3,10,10)), "z": torch.ones((1,3,10,10))}
+    ref_model = make_ref_pt_model_with_optional_inputs([3, 3, 3, 3], z_exist=True)
+    return net, ref_model, {"example_input": example_input, "input_shape": [[3, 3, 3, 3], [3, 3, 3, 3]]}
```
rkazants, line 658: Is it a normal case that shapes of examples and `input_shape` to be different? What does it mean and is it described clearly in `help`?

tests/layer_tests/mo_python_api_tests/test_mo_convert_pytorch.py
```diff
@@ -595,6 +637,39 @@ def forward(self, x):
     return net, ref_model, {}
 
 
+def create_pytorch_module_with_optional_inputs_case1(tmp_dir):
+    net = make_pt_model_with_optional_input()
+    example_input = {"x": torch.zeros((1,3,10,10)), "y": torch.ones((1,3,10,10))}
+    ref_model = make_ref_pt_model_with_optional_inputs([-1, -1, -1, -1])
```
rkazants, line 643: Very interesting that `example_input` does not set input shapes of the generated model.
Is it planned so? May we have confusion from users? I think it is better to describe it in `help`.
Now it says `Sample of model input in original framework.`

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -133,24 +134,28 @@ def _get_scripted_model(self, pt_module, example_inputs=None):
...
                             ordered_inputs.append(inputs[key])
                             used_sign.append(key)
-                        inputs = ordered_inputs
                         input_signature = used_sign
                     else:
-                        inputs = list(inputs.values())
-                        input_signature = input_signature[:len(inputs)]
+                        ordered_inputs = list(inputs.values())
+                    if is_torch_2:
+                        return {"example_kwarg_inputs": inputs}, input_signature
```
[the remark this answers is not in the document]
eaidova: example_kwarg_inputs ([dict](https://docs.python.org/3/library/stdtypes.html#dict), optional) – This parameter is a pack of keyword arguments of example inputs that will be passed to the function while tracing. Default: None. Either this argument or example_inputs should be specified. The dict will be unpacking by the arguments name of the traced function. If the keys of the dict don’t not match with the traced function’s arguments name, a runtime exception will be raised.

it is used in trorch.jit.trace
https://pytorch.org/docs/stable/generated/torch.jit.trace.html

tests/layer_tests/mo_python_api_tests/test_mo_convert_pytorch.py
eaidova: no, we do not plan restrict input shapes by example_inputs, we use it only for rank prediction. it is needed only for case when reverse infer can not properly setup it for improving user experience

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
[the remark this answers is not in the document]
eaidova: this dict represents input_parameters in code where we call trace and passed as kwargs to torch.jit.trace. I used such format for minimization difference of call between regular inputs case and kwargs

tests/layer_tests/mo_python_api_tests/test_mo_convert_pytorch.py
rkazants: I think we need to describe this better in `help` because now we may have some misalignment among FEs.
Just to set rules for all FEs.

@popovaan, do we set `example_input` shapes for model conversion in TF FE or resulted input shapes can be different from `example_input` shapes?

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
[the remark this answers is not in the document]
rkazants: can we use `example_inputs` instead?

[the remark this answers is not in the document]
eaidova: what do you mean? Just different name of variable or use another parameter for tracing.
If second, no it may lead to incorrect results in case when forward has optional arguments ad provided not all inputs with skipping in middle.
I provided test for this case (when we have input x, y=None, z=None and provided dict with x and z only. using example_nputs will lead that model got for tracing x and y as it accept list of unnamed inputs

src/bindings/python/src/openvino/frontend/pytorch/decoder.py
slyalin: 2.0.1 looks to be the latest official version in PyPI, as you mentioned probably nightly works. OK, let's put tuple issues aside, I will cover it somehow separately.

tools/mo/openvino/tools/mo/moc_frontend/extractor.py
```diff
@@ -399,7 +399,7 @@ def convert_params_lists_to_dicts(input_model,
 
         # this cycle adds each unnamed type to dictionary using name from model_inputs
         for idx, node_type in enumerate(input_user_data_types):
-            assert isinstance(node_type, (type, Type)), "Got incorrect format of input types. " \
+            assert isinstance(node_type, (type, np.dtype, Type)), "Got incorrect format of input types. " \
```
eaidova, line 402: @rkazants @popovaan please help me to understand, it is bug in the code or wrong error message here?

error says that numpy dtype is expected, but it is not listed in isinstance check, as the result when I use numpy dtype, this function report error. If it is not expected anymore and desired ov.Type, I can switch on it.

rkazants: Great catch! @popovaan, please advice.

PR #18084, opened 2023-06-15
[TF FE] Extended support of tagged MetaGraphs and variables handling in SavedModel

gkrivor:
### Details:
- Added support of alternate MetaGraph tags
- Added support of variables without explicit RestoreV2

### Tickets:
 - 111260, 111262

### DIFF
--- a/src/frontends/tensorflow/src/graph_iterator_saved_model.cpp
+++ b/src/frontends/tensorflow/src/graph_iterator_saved_model.cpp
@@ -72,6 +72,31 @@ std::basic_string<wchar_t> get_variables_index_name<wchar_t>() {
 }
 #endif
 
+std::vector<std::string> GraphIteratorSavedModel::split_tags(const std::string tags) const {
+    std::vector<std::string> tag_list = {};
+    std::size_t len = tags.length();
+    if (len == 0) {
+        return tag_list;
+    }
+    std::string tag = "";
+    std::size_t last_delimeter_pos = 0;
+    std::size_t delimeter_pos = std::string::npos;
+    while ((delimeter_pos = tags.find_first_of(",", last_delimeter_pos)) != std::string::npos) {
+        tag = tags.substr(last_delimeter_pos, delimeter_pos - last_delimeter_pos);
+        tag_list.push_back(tag);
+        last_delimeter_pos = delimeter_pos + 1;
+    }
+    if (last_delimeter_pos != std::string::npos) {
+        if (last_delimeter_pos < len) {
+            tag = tags.substr(last_delimeter_pos);
+        } else {
+            tag = "";
+        }
+        tag_list.push_back(tag);
+    }
+    return tag_list;
+}
+
 }  // namespace tensorflow
 }  // namespace frontend
 }  // namespace ov
--- a/src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
+++ b/src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
@@ -85,68 +85,97 @@ private:
         bool res = m_saved_model->ParseFromIstream(&sm_stream);
         FRONT_END_GENERAL_CHECK(res && m_saved_model->meta_graphs_size(), "Saved Model cannot be parsed");
 
+        const ::tensorflow::MetaGraphDef* meta_graph_with_no_tags = nullptr;
+
         for (const auto& meta_graph : m_saved_model->meta_graphs()) {
+            if (meta_graph.meta_info_def().tags_size() == 0) {
+                meta_graph_with_no_tags = &meta_graph;
+            }
+
             if (!meta_graph.has_graph_def()) {
                 continue;
             }
 
-            if (m_saved_model->meta_graphs_size() > 1) {
-                bool tag_found = false;
-                for (const auto& tag : meta_graph.meta_info_def().tags()) {
-                    if (tags.find(tag) != std::string::npos) {
-                        tag_found = true;
-                        break;
-                    }
+            bool tag_found = false;
+            for (const auto& tag : meta_graph.meta_info_def().tags()) {
+                if (tags.find(tag) != std::string::npos) {
+                    tag_found = true;
+                    break;
                 }
-                if (!tag_found) {
+            }
+            if (tag_found) {
+                return load_meta_graph(meta_graph);
+            }
+        }
+
+        // Alternate behavior for working with "default tag" to support additional cases for read_model
+        if (tags == "serve") {
+            // If we have only one MetaGraph - try to use it
+            if (m_saved_model->meta_graphs_size() == 1 && m_saved_model->meta_graphs(0).has_graph_def()) {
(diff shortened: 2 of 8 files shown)


src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
```diff
@@ -85,68 +85,97 @@ class GraphIteratorSavedModel : public GraphIteratorProto {
         bool res = m_saved_model->ParseFromIstream(&sm_stream);
         FRONT_END_GENERAL_CHECK(res && m_saved_model->meta_graphs_size(), "Saved Model cannot be parsed");
 
+        const ::tensorflow::MetaGraphDef* meta_graph_with_no_tags = nullptr;
+
```
rkazants, line 89: It looks that `meta_graph_with_no_tags` is not used. Please check

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
```diff
@@ -85,68 +85,97 @@ class GraphIteratorSavedModel : public GraphIteratorProto {
...
         for (const auto& meta_graph : m_saved_model->meta_graphs()) {
```
rkazants, line 90: minor:
please add a comment above `for`:
`SavedModel can contain several MetaGraph with different tags. Look for MetaGraph with the required tag`

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
```diff
@@ -85,68 +85,97 @@ class GraphIteratorSavedModel : public GraphIteratorProto {
...
                 }
-                if (!tag_found) {
+            }
+            if (tag_found) {
+                return load_meta_graph(meta_graph);
+            }
+        }
+
+        // Alternate behavior for working with "default tag" to support additional cases for read_model
+        if (tags == "serve") {
```
rkazants, line 112: May be it is better to move "serve" into `checkpoint utils` and declare with `#define`?

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
```diff
@@ -85,68 +85,97 @@ class GraphIteratorSavedModel : public GraphIteratorProto {
...
+            bool tag_found = false;
+            for (const auto& tag : meta_graph.meta_info_def().tags()) {
+                if (tags.find(tag) != std::string::npos) {
+                    tag_found = true;
+                    break;
                 }
-                if (!tag_found) {
+            }
+            if (tag_found) {
+                return load_meta_graph(meta_graph);
```
rkazants, line 107: Should not we check that this `MetaGraph` has at least one `GraphDef`?

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
```diff
@@ -85,68 +85,97 @@ class GraphIteratorSavedModel : public GraphIteratorProto {
...
-            if (m_saved_model->meta_graphs_size() > 1) {
-                bool tag_found = false;
-                for (const auto& tag : meta_graph.meta_info_def().tags()) {
-                    if (tags.find(tag) != std::string::npos) {
-                        tag_found = true;
-                        break;
-                    }
+            bool tag_found = false;
+            for (const auto& tag : meta_graph.meta_info_def().tags()) {
+                if (tags.find(tag) != std::string::npos) {
```
rkazants, line 101: Please double check why `tags` is `std::string`. I though it is a vector of strings

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
```diff
@@ -85,68 +85,97 @@ class GraphIteratorSavedModel : public GraphIteratorProto {
...
+                if (meta_graph.meta_info_def().tags_size() == 0) {
+                    return load_meta_graph(meta_graph);
                 }
             }
 
-            // MetaGraph may have a list of signatures, but at this moment we need information only about
-            // "serving_default" signature which contains information about inputs/outputs names for the
-            // model. Situation when it is missing in a file also could be.
-            auto serving_default = validSignatures.find("serving_default");
+            FRONT_END_GENERAL_CHECK(false, "Saved Model doesn't contain any applicable MetaGraph");
```
rkazants, line 130: please add `[TensorFlow Frontend]`. It will help to identify TF FE error

rkazants (commented):
Great test!
Left comments regarding the code

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
gkrivor: ok

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
gkrivor: Yep, it is an artifact, removed

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
gkrivor: Regarding information I found, I'll change this part. We should parse "tags" with ","-delimeter and compare resulting vectors.

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
gkrivor: Here - no, we skip all MetaGraphs without GraphDef (has_graph_def() == false) at line 89/95 above

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
gkrivor: done

src/frontends/tensorflow/src/graph_iterator_saved_model.hpp
gkrivor: Np. I strongly suggest to define this part somewhere else and use it as a constant. We have nonzero-overhead adding it each time into all strings.

rkazants: We have a ticket for this activity. Agree, I also don't like it to add it everytime

src/frontends/tensorflow/src/graph_iterator_saved_model.cpp
```diff
@@ -72,6 +72,31 @@ std::basic_string<wchar_t> get_variables_index_name<wchar_t>() {
 }
 #endif
 
+std::vector<std::string> GraphIteratorSavedModel::split_tags(const std::string tags) const {
```
rkazants, line 75:
```suggestion
std::vector<std::string> GraphIteratorSavedModel::split_tags(const std::string& tags) const {
```

rkazants: Minor, can be fixed in the next PR

rkazants (approved):

PR #17749, opened 2023-05-26
Update MakeStateful transformation to support dynamic shapes

itikhono:
Update MakeStateful transformation and ReadValue op to support dynamic shapes

### Tickets:
 - *111978*

PR #18086, opened 2023-06-15
[GPU] Dynamism support for ReadValue and Assign ops

andrew-k-park:
### Details:
 - The initial implementation to support dynamic variable states for shape inference and variable state memories reallocation

### Tickets:
 - 112620

### DIFF
--- a/src/plugins/intel_gpu/src/graph/assign.cpp
+++ b/src/plugins/intel_gpu/src/graph/assign.cpp
@@ -7,6 +7,8 @@
 #include <sstream>
 #include <json_object.h>
 
+#include "assign_shape_inference.hpp"
+
 namespace cldnn {
 GPU_DEFINE_PRIMITIVE_TYPE_ID(assign)
 
@@ -17,6 +19,23 @@ assign_inst::typed_primitive_inst(network& network, const assign_node& node) :
 
 layout assign_inst::calc_output_layout(const assign_node& node, kernel_impl_params const& impl_param) {
     return impl_param.typed_desc<assign>()->output_layout;
+}
+
+template<typename ShapeType>
+std::vector<layout> assign_inst::calc_output_layouts(const assign_node& /*node*/, const kernel_impl_params& impl_param) {
+    auto desc = impl_param.typed_desc<assign>();
+    auto input_layout = impl_param.get_input_layout(0);
+    auto dt = desc->output_data_types[0].value_or(input_layout.data_type);
+
+    ov::op::v6::Assign op;
+
+    std::vector<ShapeType> output_shapes = { ShapeType{} };
+    std::vector<ShapeType> input_shapes = {
+        input_layout.get<ShapeType>()
+    };
+
+    ov::op::v6::shape_infer(&op, input_shapes, output_shapes);
+    return {{output_shapes[0], dt, format::get_default_format(output_shapes[0].size())}};
 }
 
 std::string assign_inst::to_string(const assign_node& node) {
--- a/src/plugins/intel_gpu/src/graph/assign.cpp
+++ b/src/plugins/intel_gpu/src/graph/assign.cpp
@@ -7,8 +7,6 @@
 #include <sstream>
 #include <json_object.h>
 
-#include "assign_shape_inference.hpp"
-
 namespace cldnn {
 GPU_DEFINE_PRIMITIVE_TYPE_ID(assign)
 
@@ -19,23 +17,6 @@ assign_inst::typed_primitive_inst(network& network, const assign_node& node) :
 
 layout assign_inst::calc_output_layout(const assign_node& node, kernel_impl_params const& impl_param) {
     return impl_param.typed_desc<assign>()->output_layout;
-}
-
-template<typename ShapeType>
-std::vector<layout> assign_inst::calc_output_layouts(const assign_node& /*node*/, const kernel_impl_params& impl_param) {
-    auto desc = impl_param.typed_desc<assign>();
-    auto input_layout = impl_param.get_input_layout(0);
-    auto dt = desc->output_data_types[0].value_or(input_layout.data_type);
-
-    ov::op::v6::Assign op;
-
-    std::vector<ShapeType> output_shapes = { ShapeType{} };
-    std::vector<ShapeType> input_shapes = {
-        input_layout.get<ShapeType>()
-    };
-
-    ov::op::v6::shape_infer(&op, input_shapes, output_shapes);
-    return {{output_shapes[0], dt, format::get_default_format(output_shapes[0].size())}};
 }
 
 std::string assign_inst::to_string(const assign_node& node) {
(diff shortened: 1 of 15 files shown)


src/plugins/intel_gpu/src/graph/assign.cpp
```diff
@@ -19,6 +21,23 @@ layout assign_inst::calc_output_layout(const assign_node& node, kernel_impl_para
     return impl_param.typed_desc<assign>()->output_layout;
 }
 
+template<typename ShapeType>
+std::vector<layout> assign_inst::calc_output_layouts(const assign_node& /*node*/, const kernel_impl_params& impl_param) {
```
vladimir-paramuzov, line 25: Can't it be implemented via `return forward_input0_shape<ShapeType>(impl_param);` ?

src/plugins/intel_gpu/src/graph/assign.cpp
```diff
@@ -19,6 +21,23 @@ layout assign_inst::calc_output_layout(const assign_node& node, kernel_impl_para
...
+    auto desc = impl_param.typed_desc<assign>();
+    auto input_layout = impl_param.get_input_layout(0);
```
vladimir-paramuzov, line 27: Please add explicit template instantiation for PartialShape

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -1449,8 +1449,16 @@ void network::allocate_primitive_instance(program_node const& node) {
         if (node.is_type<data>())
             _data_outputs.push_back(inst);
     }
-    if (std::dynamic_pointer_cast<assign_inst>(inst) || std::dynamic_pointer_cast<read_value_inst>(inst))
+    if (std::dynamic_pointer_cast<assign_inst>(inst) || std::dynamic_pointer_cast<read_value_inst>(inst)) {
```
vladimir-paramuzov, line 1452: Use `node.is_type<>()` for consistency

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -1508,4 +1516,57 @@ void network::assign_variables_memories(variables_states_map &&variables_memorie
...
+            }
+        }
+    }
+}
+
+void network::update_variable_memory(const std::string& variable_id, const cldnn::layout& layout) {
+    auto it = _variables_states.find(variable_id);
+    if (it != _variables_states.end()) {
+        if (it->second->memory->get_layout() != layout) {
+            it->second->set_memory(get_engine().allocate_memory(layout, false));
```
vladimir-paramuzov, line 1534: I think memory can be reused in some cases

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -1508,4 +1516,57 @@ void network::assign_variables_memories(variables_states_map &&variables_memorie
...
+                }
+            }
+        }
+    } else {
+        cldnn::network::VariableState::Ptr variable_state = std::make_shared<cldnn::network::VariableState>(get_engine().allocate_memory(layout, false));
+        _variables_states.insert({variable_id, variable_state});
+        for (auto primitive : _variable_state_primitives) {
+            if (const auto& memory_state_primitive = std::dynamic_pointer_cast<memory_state::variable>(primitive)) {
+                if (!variable_id.compare(memory_state_primitive->variable_id())) {
+                    primitive->set_output_memory(variable_state->memory, false);
```
vladimir-paramuzov, line 1549: That part duplicates code above. Can it be refactored to something like this:
```
auto it = _variables_states.find(variable_id);
if (it == _variables_states.end()) {
 // alloc & insert
}

// common_update
```
?

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -1508,4 +1516,57 @@ void network::assign_variables_memories(variables_states_map &&variables_memorie
...
+
+void network::update_variable_memory(const std::string& variable_id, const cldnn::layout& layout) {
+    auto it = _variables_states.find(variable_id);
+    if (it != _variables_states.end()) {
+        if (it->second->memory->get_layout() != layout) {
+            it->second->set_memory(get_engine().allocate_memory(layout, false));
+            for (auto primitive : _variable_state_primitives) {
+                if (const auto& memory_state_primitive = std::dynamic_pointer_cast<memory_state::variable>(primitive)) {
+                    if (!variable_id.compare(memory_state_primitive->variable_id())) {
+                        primitive->set_output_memory(it->second->memory, false);
```
vladimir-paramuzov, line 1538: Also, I'd propose to think about changing state implementation a little bit.
As I understand, currently Virables are not a part of graph and stored separately in network. But I think it can be added as a `mutable_data` primitive, so read/assign primitives will be connected with it to guarantee correct processing order.

I assume that with such a change we won't need to iterate over state primitives and update each of them with new mem ptr, you'll just update mutable_data once and then common code will propagate the info.

src/plugins/intel_gpu/src/graph/assign.cpp
andrew-k-park: Yes, Since `ReadValue` and `Assign` just propagate the input shape as it is, I updated it to use `forward_input0_shape` in header

src/plugins/intel_gpu/src/graph/network.cpp
andrew-k-park: updated

src/plugins/intel_gpu/src/graph/network.cpp
andrew-k-park: Added to compare and reuse existing allocated variable buffers

src/plugins/intel_gpu/src/graph/network.cpp
andrew-k-park: Refactored this part to avoid duplication

src/plugins/intel_gpu/src/graph/network.cpp
vladimir-paramuzov: I think empty tensor (or tensor with single element) would be fine.

vladimir-paramuzov (approved):
Overall LGTM. I suggest trying it e2e with https://github.com/openvinotoolkit/openvino/pull/17749/ to ensure that everything is OK.
Comment concerning replacing variables map with mutable_data node can be applied in another PR

andrew-k-park:
> Overall LGTM. I suggest trying it e2e with #17749 to ensure that everything is OK. Comment concerning replacing variables map with mutable_data node can be applied in another PR
@vladimir-paramuzov As a result of running the stateful dolly version by applying the corresponding PR(https://github.com/openvinotoolkit/openvino/pull/17749), `read_model` step passes, but a runtime error occurs during ngraph pass execution(probably during `ConvertPrecision` pass) in `compile_model` step. Therefore, should it be checked on ngraph transform component side?

PR #18092, opened 2023-06-15
Decompose/flatten tuple inputs

slyalin:
Decomposes (flatten) input tuples at inputs in PT FE recursively down to raw tensors.
Allow passing tuples and list as `example_input` argument of `convert_model`.

TODO: Naming of new inputs.


eaidova (approved):

src/frontends/pytorch/src/transforms/prim_list_tuple_unpack_replacer.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace pass {
+
+bool DecomposeListTupleParameters::run_on_model(const std::shared_ptr<Model>& model) {
+    bool at_least_one_decomposed = false;
+    std::queue<std::shared_ptr<ov::op::v0::Parameter>> parameters;
+    for (auto par : model->get_parameters()) {
```
rkazants, line 21:
```suggestion
    for (const auto& par : model->get_parameters()) {
```

rkazants: exclude extra copy

src/frontends/pytorch/src/transforms/prim_list_tuple_unpack_replacer.cpp
```diff
@@ -0,0 +1,98 @@
...
+        auto parameter = parameters.front();
+        parameters.pop();
+        auto consumers = parameter->get_output_target_inputs(0);
+        size_t num_outputs = 0;  // number of outputs in each unpack consumer should match
+        bool all_unpacks = true;
+
+        // collects all outputs per each consumer operation for this tuple/list Parameter
+        std::vector<OutputVector> consumer_outputs;
+
+        for (auto consumer : consumers) {
```
rkazants, line 34:
```suggestion
        for (const auto& consumer : consumers) {
```

src/frontends/pytorch/src/transforms/prim_list_tuple_unpack_replacer.hpp
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+
+#include "openvino/pass/graph_rewrite.hpp"
+#include "openvino/pass/pass.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace pass {
+
```
rkazants, line 14: please provide description for this transformation and its idea with unpack operation

src/frontends/pytorch/src/transforms/prim_list_tuple_unpack_replacer.cpp
```diff
@@ -0,0 +1,98 @@
...
+namespace pytorch {
+namespace pass {
+
+bool DecomposeListTupleParameters::run_on_model(const std::shared_ptr<Model>& model) {
+    bool at_least_one_decomposed = false;
+    std::queue<std::shared_ptr<ov::op::v0::Parameter>> parameters;
+    for (auto par : model->get_parameters()) {
+        parameters.push(par);
+    }
+    while (!parameters.empty()) {
```
rkazants, line 24: can we use `for` instead and avoid extra copies by using `front()` method?

src/frontends/pytorch/src/transforms/prim_list_tuple_unpack_replacer.cpp
```diff
@@ -0,0 +1,98 @@
...
+            continue;
+        }
+
+        for (size_t i = 0; i < num_outputs; ++i) {
+            // Merged partial shape and element type among all the consumers of i-th result of unpack ops
+            PartialShape ps = PartialShape::dynamic();
+            element::Type et = element::dynamic;
+            std::set<Input<Node>> inputs;
+
+            for (auto outputs : consumer_outputs) {
```
rkazants, line 67:
```suggestion
            for (const auto& outputs : consumer_outputs) {
```

src/frontends/pytorch/src/transforms/prim_list_tuple_unpack_replacer.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+            auto new_parameter = std::make_shared<ov::op::v0::Parameter>(et, ps);
+
+            for (auto input : inputs) {
+                auto names = input.get_tensor().get_names();
+                input.replace_source_output(new_parameter->output(0));
+                new_parameter->output(0).add_names(names);
+            }
+
+            // TODO: Assign correct names
```
rkazants, line 85: ?

rkazants (commented):
Still in more deep review.

src/frontends/pytorch/src/transforms/prim_list_tuple_unpack_replacer.cpp
slyalin: Not addressed in this PR.

rkazants (approved):

src/frontends/pytorch/src/transforms/prim_tuple_unpack_parameter_replacer.cpp
```diff
@@ -0,0 +1,97 @@
...
+            auto new_parameter = std::make_shared<ov::op::v0::Parameter>(et, ps);
+
+            for (auto input : inputs) {
+                auto names = input.get_tensor().get_names();
+                input.replace_source_output(new_parameter->output(0));
+                new_parameter->output(0).add_names(names);
+            }
+
+            // TODO: Assign correct names
+            model->add_parameters({new_parameter});
```
mvafin, line 85: I don't think this will preserve order of inputs for the case when tuple is followed by a non tuple input.

mvafin (changes requested):

src/frontends/pytorch/src/transforms/prim_tuple_unpack_parameter_replacer.cpp
```diff
@@ -0,0 +1,97 @@
...
+            auto node = consumer.get_node()->shared_from_this();
+            auto tuple_unpack = cast_fw_node(node, "prim::TupleUnpack");
+            if (!tuple_unpack) {
+                all_unpacks = false;
+                break;
+            }
+            if (num_outputs == 0) {
+                num_outputs = node->get_output_size();
+            } else if (num_outputs != node->get_output_size()) {
+                std::cerr << "[ PT FE WARNING ] Unpack node " << node
```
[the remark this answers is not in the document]
slyalin: Indeed we need to inform user about this in some way. OPENVINO_DEBUG is not shown it is useless for a regular user. Can we attach a warning to a node and print it when this node is left unconverted? If not I don't know about other methods to report a problem.

src/frontends/pytorch/src/transforms/prim_tuple_unpack_parameter_replacer.cpp
```diff
@@ -0,0 +1,97 @@
...
+                all_unpacks = false;
+                break;
+            }
+            consumer_outputs.push_back(node->outputs());
+        }
+
+        if (!all_unpacks || consumer_outputs.empty()) {
+            // if at least one consumer is not an unpack-like op or there are not matching number of unpacked objects,
+            // we cannot replace other unpacks even if they exist, leaving Unpack-op(s) in the graph for this Parameter
+            continue;
```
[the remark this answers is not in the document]
slyalin: Do we have a way to report such warnings?

src/frontends/pytorch/src/transforms/prim_tuple_unpack_parameter_replacer.cpp
slyalin: Real thing! Thanks for catching it. This is copied-and-pasted from a similar `Result`s flattening transformation at src/frontends/pytorch/src/transforms/prim_list_tuple_construct_replacer.cpp. Just added a couple of tests and the problem is really reproduced for both transformations. Will fix both.

src/frontends/pytorch/src/transforms/prim_tuple_unpack_parameter_replacer.cpp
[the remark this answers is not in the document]
slyalin: What do you mean by "temporary solution"? It looks to be permanent, nothing else we can do in this particular transformation.

src/frontends/pytorch/src/transforms/prim_tuple_unpack_parameter_replacer.cpp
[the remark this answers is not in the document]
mvafin: You can use `add_exception_to_fw_node` helper function to add that exception into TupleUnpack node and it will be reported as the reason why it wasn't decomposed. If you still want to leave it as console message please change "PT FE" to "PyTorch Frontend", because pt fe might confuse user.

src/frontends/pytorch/src/transforms/prim_tuple_unpack_parameter_replacer.cpp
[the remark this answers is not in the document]
mvafin: You can add message to unpack node using `add_exception_to_fw_node` helper function to be shown as a reason why it wasn't decomposed. By temporary I meant that eventually we want to support tuples in a better way

src/frontends/pytorch/src/transforms/prim_tuple_unpack_parameter_replacer.cpp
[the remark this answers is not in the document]
slyalin: Done. Also handled the case when not all the consumers have prim::TupleUnpack types to attach the diagnostics to those nodes that have type prim::TupleUnpack. Tested with manual experiments, dedicated test wasn't added (don't know how to do that).

mvafin (approved):

PR #18093, opened 2023-06-15
[ARM CPU] Add parallel reference power operation

allnes:
continue work [PR#17581](https://github.com/openvinotoolkit/openvino/pull/17581)


src/plugins/intel_cpu/src/nodes/eltwise.cpp
```diff
@@ -1634,6 +1634,50 @@ class EltwiseRefExecutor : public Eltwise::IEltwiseExecutor {
             });
             return;
         }
+        if (_opData.algo == Algorithm::EltwisePowerDynamic) {
```
dmitry-gorokhov, line 1637: I would expect addition of PowerStatic optimized code, as long as this algorithm much more common.

src/plugins/intel_cpu/tests/functional/single_layer_tests/classes/eltwise.cpp
```diff
@@ -190,7 +191,7 @@ const std::vector<ngraph::helpers::EltwiseTypes>& eltwiseOpTypesBinInp() {
 const std::vector<ngraph::helpers::EltwiseTypes>& eltwiseOpTypesDiffInp() {
         static const std::vector<ngraph::helpers::EltwiseTypes> eltwiseOpTypesDiffInp = { // Different number of input nodes depending on optimizations
         #if defined(OPENVINO_ARCH_X86) || defined(OPENVINO_ARCH_X86_64)
-                ngraph::helpers::EltwiseTypes::POWER,                   //TODO: Fix CVS-111880
+                ngraph::helpers::EltwiseTypes::POWER,
```
dmitry-gorokhov, line 194: POWER is still tested only on x86

dmitry-gorokhov (changes requested):

src/plugins/intel_cpu/tests/functional/single_layer_tests/classes/eltwise.cpp
allnes: @dmitry-gorokhov this was disabled because of the executor check (now some tests failed because ref != acl), @alvoron can you please describe how to return the ref check back on the implementation

src/plugins/intel_cpu/src/nodes/eltwise.cpp
allnes: @dmitry-gorokhov corrected

dmitry-gorokhov: So the next step is to enable Conversion to PowerStatic for ARM platform: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/transformations/cpu_opset/convert_to_cpu_specific_opset.hpp#L39 (by making this transformation common). After that you will be able to remove optimized code block for PowerDynamic.
Also please peform benchmark launch on whole set of models to make sure PowerStatic transormation enablement doesn't introduce any regressions.

dmitry-gorokhov (approved):
LGTM

PR #18098, opened 2023-06-16
removing redundant statement in `ov::descriptor::Tensor::Tensor`

kai-waang:
### Details:
 - removing `m_name_it = m_names.cend();`, which seems to be redundant because in `set_names` the iterator is immediately set to `m_names.cbegin()`.

### Tickets:
 - *ticket-id*


src/core/src/descriptor/tensor.cpp
```diff
@@ -15,7 +15,6 @@ ov::descriptor::Tensor::Tensor(const element::Type& element_type,
     : m_element_type(element_type),
       m_partial_shape(pshape),
       m_shape_changed(true) {
-    m_name_it = m_names.cend();
```
rkazants, line 18: why this is redundant? We just initialize this in the constructor that is not bad practice. Also, the stuff happened in other constructors of `Tensor`.

kai-waang: but it will be set to `m_names.cbegin()` when `set_names` is called.

ilyachur (approved):

PR #18100, opened 2023-06-16
[CPU] ONNX Unique tests failing on CPU

nshchego:
### Details:
 - *Fix for invalid sorting in sliced case*

### Tickets:
 - *101008, 103964*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/unique.cpp
+++ b/src/plugins/intel_cpu/src/nodes/unique.cpp
@@ -2,11 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-#include <string>
-#include <vector>
-
 #include "unique.hpp"
-#include <ngraph/opsets/opset1.hpp>
+
+#include "ie_parallel.hpp"
+#include <openvino/op/unique.hpp>
 #include <utils/shape_inference/shape_inference_internal_dyn.hpp>
 
 using namespace InferenceEngine;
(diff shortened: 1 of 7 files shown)


tomdol:
The ONNX tests got fixed by this PR so you need to enable them to make the ONNX CI pass. Please remove the usages of this xfail https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/tests/test_onnx/test_backend.py#L199 and its definition because after your PR it does not apply to any test any more.

dmitry-gorokhov:
@antonvor Please review this PR

antonvor (approved):

src/plugins/intel_cpu/src/nodes/unique.cpp
```diff
@@ -318,178 +315,171 @@ void Unique::slicedTensorExec() {
...
-                    }
-                    if (equal) {
-                        inToOutTmpPtr[b1] = b2;
-                    }
-                }
-            }
+            memcpy(inToOutPtr, inToOutTmp.data(), axisDim * sizeof(int));
+        }
+        if (definedOutputs[OCCURRENCES_NUM]) {
+            memcpy(occurNPtr, occurTmp.data(), uniqueLenIB);
```
maxnick, line 482: Two questions here.
1. Is not it better to use `cpu_memcpy` to avoid some warnings from MSVC?
2. Does it make sense to use parallel copy here?

nshchego: Done

maxnick (approved):

PR #18117, opened 2023-06-16
[PT FE] Extend aten::pad transformation to accept aten::reflection_pad2d

mmikolajcz:
### Details:
 - *Extend existing aten::pad transformation to accept aten::reflection_pad2d*
 - *Operator aten::reflection_pad2d doesn't have layer tests - this seem to be old implementation of aten::pad in torch, couldn't reproduce it with current pytorch version in tests. Tested manually using TorchScript model containing this op that was converted using torch 1.8*

### Tickets:
 - *112882*


eaidova (approved):

src/frontends/pytorch/src/transforms/prim_list_construct_pad.cpp
```diff
@@ -61,19 +61,45 @@ PrimListConstructPadReplacer::PrimListConstructPadReplacer() {
...
+                    mode = attrs.at("string_value");
+                }
+            }
+        } else if ((pad_op = cast_fw_node(m.get_match_root(), "aten::reflection_pad2d"))) {
+            mode = "reflect";
+            input_node = pad_op->input_value(0);
+            padding = pad_op->input_value(1);
+            // Pad value is used only for constant pad, fill with 0 as placeholder.
+            pad_value = v0::Constant::create(element::f32, Shape{}, {0});
+            rt_info = {pad_op, input_node.get_node_shared_ptr(), padding.get_node_shared_ptr()};
```
mvafin, line 93: Why do you create this NodeVector? I can't it being used anywhere. In fact if you meant to use it to copy rt info it is not needed, because all created operation are handled by NodeRegistry here.

mmikolajcz: Thanks, deleted. I've added it before changes with NodeRegistry were introduced and forgot to delete it while solving merge conflicts

mvafin (approved):

PR #18119, opened 2023-06-16
Gna plugin more meaningful memory dump

jnowicki-intel:
### Details:
 -  current implementation: allows displaying aggregate sum across all layers and operations of memory used for each operation type (weights, biases etc.) and memory type (read only, scratch etc.). Useful for debugging differences between IR and GNA model representations.
 - in future: IR to GNA layer mapping (which operations transformed into what operations), refactoring dumping code

example output at end of file (synthetic example):
Gna2MemoryTagReadOnly activationFunction: 4616 bytes.
Gna2MemoryTagReadOnly biases: 1024 bytes.
Gna2MemoryTagReadOnly filters: 14576 bytes.
Gna2MemoryTagReadOnly inputs: 11288 bytes.
Gna2MemoryTagReadOnly weights: 85424 bytes.


src/plugins/intel_gna/src/log/dump.cpp
```diff
@@ -440,11 +440,38 @@ inline void DumpCharArray(std::ostream& dumpFile, const char* carray, size_t cou
     dumpFile << "\n";
 }
 
+class Gna2Stats {
```
sirzabek, line 443:
```suggestion
class GnaMemStats {
```

src/plugins/intel_gna/src/log/dump.cpp
```diff
@@ -440,11 +440,38 @@ inline void DumpCharArray(std::ostream& dumpFile, const char* carray, size_t cou
...
+        if (size == 0)
+            return;
+        const std::string key = memory_tag + " " + operation_type;
+        if (stats_.count(key) == 0) {
+            stats_[key] = size;
+        } else {
+            stats_[key] += size;
+        }
+    }
+    std::string GetFormattedStats() {
```
sirzabek, line 455:
```suggestion
    std::string GetFormattedMemStats() {
```

src/plugins/intel_gna/src/log/dump.cpp
sirzabek: I suggest to change the name to avoid confusion with some performance or other statistics available from GNA library and also with this class being related to GNA library headers (which use the Gna2 prefix).

sirzabek:
Great work! Can we have this extended to generate per layer type stats as well (e.g. Convolutions, Affine, AffineDiagonal, etc.)? This is very useful during analysis. Then count of each layer type should be added as well.

src/plugins/intel_gna/src/log/dump.cpp
```diff
@@ -440,11 +440,63 @@ inline void DumpCharArray(std::ostream& dumpFile, const char* carray, size_t cou
     dumpFile << "\n";
 }
 
+class GnaMemStats {
+public:
+    void Add(const std::string memory_tag,
+             const std::string operation_type,
```
sirzabek, line 446:
```suggestion
             const std::string operand_name,
```

src/plugins/intel_gna/src/log/dump.cpp
```diff
@@ -440,11 +440,63 @@ inline void DumpCharArray(std::ostream& dumpFile, const char* carray, size_t cou
...
+             const std::string operation_name,
+             const uint32_t size) {
+        if (size == 0)
+            return;
+        const std::pair<std::string, std::string> key =
+            std::make_pair(operation_name, memory_tag + " " + operation_type);
+        if (stats_.count(key) == 0) {
+            stats_[key] = size;
+        } else {
+            stats_[key] += size;
```
sirzabek, line 456: Can this ever happen for a combination of operation type, number, operand name and memory tag type? Isn't it unique?

jnowicki-intel: true

sirzabek (approved):

PR #18130, opened 2023-06-18
Don't eliminate Transpose when 0 size of order_value

riverlijunjie:
### Details:
 - Remove invalid test case of transpose

### Tickets:
 - *ticket-id*


src/common/transformations/src/transformations/common_optimizations/nop_elimination.cpp
```diff
@@ -701,6 +701,11 @@ pass::EliminateTranspose::EliminateTranspose() {
         }
 
         const auto& order_values = order_const->cast_vector<int64_t>();
+        // Don't eliminate Transpose when 0 size of order_value, which will make single TransPose ops to be eliminated
+        // when preprocess is applied by default in case of legacy ov api + plugin api 2.0. In which Convert ops will be
+        // added by preprocess but Transpose will be removed.
+        if (order_values.size() == 0)
+            return false;
```
rkazants, line 708: Not quite clear how it is possible to get empty `input_order` value. Is it for a scalar?

rkazants: Can you please elaborate more what is the pre-processing applied (mean/scale and input)?

rkazants:
Please add test

src/common/transformations/src/transformations/common_optimizations/nop_elimination.cpp
riverlijunjie: We can see test case, for example smoke_Transpose4D tests:
https://github.com/openvinotoolkit/openvino/blob/5574acc6ca0d5a32bfdf4214bb4b891d901041cf/src/plugins/intel_cpu/tests/functional/shared_tests_instances/single_layer_tests/transpose.cpp#L39-L45

There is `{}` in `order4D`, which will cause `order_values.size() == 0` as above.

While for Legacy api + plugin api 2.0, we can see preprocess is added by default:
https://github.com/openvinotoolkit/openvino/blob/5574acc6ca0d5a32bfdf4214bb4b891d901041cf/src/inference/src/dev/core_impl.cpp#L600-L608

It will cause new ops are added into graph, it will cause Transpose single layer tests become non-single layer and Transpose will be eliminated.

riverlijunjie:
> Please add test

Current test case has contained such case, this issue is triggered for legacy ov api + cpu plugin api 2.0.

rkazants:
> > Please add test
> 
> Current test case has contained such case, this issue is triggered for legacy ov api + cpu plugin api 2.0.

We have separate test suite for transformations but I have double that the test case is valid. Please double-check.

@itikhono, fyi.

riverlijunjie:
> > > Please add test
> > 
> > 
> > Current test case has contained such case, this issue is triggered for legacy ov api + cpu plugin api 2.0.
> 
> We have separate test suite for transformations but I have doubt that the test case is valid. Please double-check.
> 
> @itikhono, fyi.

@itikhono @rkazants how do you think about this PR? remove possible invalid `{}` test or check the empty order? We need a decision to make this PR can be closed or merged, thanks!

rkazants:
> > > > Please add test
> > > 
> > > 
> > > Current test case has contained such case, this issue is triggered for legacy ov api + cpu plugin api 2.0.
> > 
> > 
> > We have separate test suite for transformations but I have doubt that the test case is valid. Please double-check.
> > @itikhono, fyi.
> 
> @itikhono @rkazants how do you think about this PR? remove possible invalid `{}` test or check the empty order? We need a decision to make this PR can be closed or merged, thanks!

I think the test case is not quite valid because Transpose with empty order contradicts Transpose specification. Also, we don't have any real model with such Transpose, right? So I propose just to remove the test case and the fix is not needed.

riverlijunjie:
> > > > > Please add test
> > > > 
> > > > 
> > > > Current test case has contained such case, this issue is triggered for legacy ov api + cpu plugin api 2.0.
> > > 
> > > 
> > > We have separate test suite for transformations but I have doubt that the test case is valid. Please double-check.
> > > @itikhono, fyi.
> > 
> > 
> > @itikhono @rkazants how do you think about this PR? remove possible invalid `{}` test or check the empty order? We need a decision to make this PR can be closed or merged, thanks!
> 
> I think the test case is not quite valid because Transpose with empty order contradicts Transpose specification. Also, we don't have any real model with such Transpose, right? So I propose just to remove the test case and the fix is not needed.

Ok, let's remove such invalid test cases.

rkazants (approved):

riverlijunjie:
@rkazants  can we merge this PR?

rkazants:
> @rkazants can we merge this PR?

merged

PR #18137, opened 2023-06-19
Optimize transpose reference implementation

mateusztabaka:
Ticket: CVS-105967


src/core/reference/src/runtime/opt_kernel/reshape.cpp
```diff
@@ -73,26 +96,39 @@ void reshape_in3(const char* in,
...
-                           elem_size);
+    size_t num_elements = shape_size(in_shape);
+    if (num_elements <= get_threshold()) {
+        size_t i, j, k;
+        size_t* in_indexes[3];
+        in_indexes[in_axis_order[0]] = &i;
+        in_indexes[in_axis_order[1]] = &j;
+        in_indexes[in_axis_order[2]] = &k;
+
+        for (i = 0; i < out_shape[0]; i++) {
```
praasz, line 107: Can you explain why  pointers to indexes are used?
The alternative solution could be like in each loop the sub-offset could be calculated per iteration like:
```suggestion
        for (i = 0; i < out_shape[0]; i++) {
            size_t sub_offset_i = i * in_shape[1];
```

or 

```suggestion
        size_t sub_offset_i = 0;
        for (i = 0; i < out_shape[0]; i++) {
           // add in_shape every iteration to avoid multiplication and it should be eq to i * in_shape[1]
            sub_offset_i += in_shape[1];
```

Then in `copy_element` just sum all of them:
```cpp
copy_element(out, in +(sub_offset_i + sub_offset_j + ..., elem_size);
```
I think working fundamental types directly should be faster then using access to table and then via pointer to value  (indirect access).

But maybe current solution better(faster)?

praasz (approved):

src/core/reference/src/runtime/opt_kernel/reshape.cpp
```diff
@@ -41,58 +42,107 @@ void reshape_in1(const char* in,
...
+}
+
+static inline void copy_element(char* out, const char* in, size_t elem_size) {
+#define CASE(type)                                                          \
+    case sizeof(type):                                                      \
+        *reinterpret_cast<type*>(out) = *reinterpret_cast<const type*>(in); \
+        break;
+
+    switch (elem_size) {
+        CASE(int32_t)
```
itikhono, line 57: maybe it's better to add a minor comment here, that float, double and so on types are also covered with this switch-case and this is expected.

itikhono (approved):

PR #18138, opened 2023-06-19
Handle sequence ops with non constant W, R, B inputs

mateusztabaka:
Ticket: CVS-49207


src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
```diff
@@ -293,80 +294,24 @@ void Transformations::PreLpt(const std::vector<ov::element::Type>& defaultPrecis
...
-        [isSequencePrimitiveSupported](const_node_ptr &node) -> bool {
-            return isSequencePrimitiveSupported(node);
+        [](const_node_ptr &node) -> bool {
+            const auto& data_pshape = node->get_input_partial_shape(0);
+            // WA: dynamic shapes make impossible to check seq_len due to shapeOf subgraphs
+            // but the sequence is still supported in CPU and doesn't need to be decomposed
+            if (data_pshape.is_dynamic())
+                return true;
+            std::string msg;
+            return node::RNN::isSupportedOperation(node, msg);
```
itikhono, line 305: I believe we also need to update functional tests to cover this case with Parameters
maybe these
tests/functional/shared_test_classes/src/single_layer/lstm_cell.cpp

itikhono (approved):

itikhono:
Have you checked GPU?

src/plugins/intel_cpu/src/nodes/rnn.cpp
```diff
@@ -255,25 +256,57 @@ bool RNN::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, std::s
...
-            direction = ov::as_type_ptr<const ov::op::v0::LSTMSequence>(op)->get_direction();
-        } else if (ov::is_type<ov::op::v5::LSTMSequence>(op)) {
-            direction = ov::as_type_ptr<const ov::op::v5::LSTMSequence>(op)->get_direction();
-        } else if (ov::is_type<ov::op::v5::RNNSequence>(op)) {
-            direction = ov::as_type_ptr<const ov::op::v5::RNNSequence>(op)->get_direction();
+        size_t seqLenIdx = 2;
+        if (auto gru_seq = ov::as_type_ptr<const ov::op::v5::GRUSequence>(op)) {
+            direction = gru_seq->get_direction();
+        } else if (auto lstm_seq = ov::as_type_ptr<const ov::op::v0::LSTMSequence>(op)) {
+            if (lstm_seq->get_activations() != std::vector<std::string>{"sigmoid", "tanh", "tanh"}) {
```
EgorDuplensky, line 282: Isn't it check at line 2687

src/plugins/intel_cpu/src/nodes/rnn.cpp
```diff
@@ -255,25 +256,57 @@ bool RNN::isSupportedOperation(const std::shared_ptr<const ov::Node>& op, std::s
         }
 
         auto rnnCellBase = ov::as_type_ptr<const ov::op::util::RNNCellBase>(op);
-        if (rnnCellBase && rnnCellBase->get_clip() != 0.f) {
-            errorMessage = "Clipping is not supported for RNN primitive.";
-            return false;
+        if (rnnCellBase) {
```
EgorDuplensky, line 259: I believe we can add an early return if cast to RNNCellBase is failed.

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
EgorDuplensky: Isn't it covered by conformance tests?

EgorDuplensky:
@mateusztabaka Have we checked that conformance tests are passing with this change?

src/plugins/intel_cpu/src/nodes/rnn.cpp
mateusztabaka: It would be an issue for v0::LSTMSequence since it doesn't inherit from RNNCellBase.

src/plugins/intel_cpu/src/nodes/rnn.cpp
mateusztabaka: Line 268 checks for v5::LSTMSequence. Here we do that for v0::LSTMSequence.

mateusztabaka:
> @mateusztabaka Have we checked that conformance tests are passing with this change?

They're passing now. There was some issue with GRUSequence test, but it's working now.

mateusztabaka:
> Have you checked GPU?

I've added tests for GPU. Some of them are failing because of result mismatch. They're disabled for further investigation:  https://github.com/openvinotoolkit/openvino/pull/18138/files#diff-08c8f907e1dcb5695ec429778494212b3257ee8b73c6c59045cb95b8d57274c7R66.

EgorDuplensky (approved):

itikhono:
Have we checked the models from the ticket?

itikhono (approved):

mateusztabaka:
> Have we checked the models from the ticket?

The model works with or without my change.

itikhono:
@EgorDuplensky do we have some "real" model to verify this change?

maxnick:
@v-Golubev , could you please review the LPT part?

EgorDuplensky:
> @EgorDuplensky do we have some "real" model to test this change?
I don't think so.
The initial idea was to fix the conformance tests.

v-Golubev (approved):
LPT part LGTM

maxnick (approved):
CPU part LGTM

maxnick:
@iefode , could you please take a look at the changes in the common test classes?

iefode (approved):
LGTM from IE test infra perspective

src/common/transformations/include/transformations/utils/utils.hpp
```diff
@@ -221,6 +221,9 @@ TRANSFORMATIONS_API bool is_dequantization_subgraph(const Output<Node>& node);
 TRANSFORMATIONS_API bool can_eliminate_eltwise_node(const std::shared_ptr<Node>& eltwise,
                                                     const Output<Node>& constant,
                                                     const Output<Node>& non_constant_input);
+
+TRANSFORMATIONS_API bool is_constfoldable(const ov::Output<ov::Node>& output);
```
v-Golubev, line 225: @mateusztabaka I have one minor comment regarding the function name. As for me, this name is a bit misleading.
The function checks that output is on constant path, not constfoldable. In some cases `output` can be on constant path but not be constfoldable. E.g. if constant folding for some node on the path was disabled by the [other transformation](https://github.com/openvinotoolkit/openvino/blob/master/src/common/transformations/src/transformations/low_precision/mark_dequantization_subgraph.cpp#L79)
Maybe it's good idea to rename this function? What do you think?

mateusztabaka: Right, I renamed the function to is_on_constant_path.

PR #18141, opened 2023-06-19
[PyOV] Expose api to Model class

akuporos:
### Details:
 - Exposed `add_parameters`, `add_results`, `add_sinks`, `get_sinks`, `remove_result`, `remove_parameter` and `remove_sink` methods of the Model class
 - Implement `Assign` operation on pybind side

### Tickets:
 - 113329


tomdol (approved):

src/bindings/python/src/openvino/runtime/opset10/__init__.py
```diff
@@ -11,7 +11,7 @@
 from openvino.runtime.opset1.ops import add
 from openvino.runtime.opset1.ops import asin
 from openvino.runtime.opset4.ops import asinh
-from openvino.runtime.opset3.ops import assign
+from openvino.runtime.opset6.ops import assign
```
jiwaszki, line 14: @tomdol @mitruska is it OK to change already established opset versions?

tomdol: In fact this change aligns the Python opsets with C++ where the opset10 contains Assign v6, not v3 https://github.com/openvinotoolkit/openvino/blob/master/src/core/include/openvino/opsets/opset10_tbl.hpp#L161

src/bindings/python/src/pyopenvino/graph/model.cpp
```diff
@@ -696,6 +707,130 @@ void regclass_graph_Model(py::module m) {
...
+        "remove_sink",
+        [](ov::Model& self, const py::object& node) {
+            if (py::isinstance<ov::op::v6::Assign>(node)) {
+                auto sink = std::dynamic_pointer_cast<ov::op::Sink>(node.cast<std::shared_ptr<ov::op::v6::Assign>>());
+                self.remove_sink(sink);
+            } else if (py::isinstance<ov::Node>(node)) {
+                auto sink = std::dynamic_pointer_cast<ov::op::Sink>(node.cast<std::shared_ptr<ov::Node>>());
+                self.remove_sink(sink);
+            } else {
+                throw py::type_error("Incorrect argument type.");
```
p-wysocki, line 747: Info about expected type would probably be useful from the user's point of view.

p-wysocki (approved):
LGTM. Do we need separate tests for the new `assign` binding?

src/bindings/python/src/pyopenvino/graph/model.cpp
akuporos: extended message

PR #18154, opened 2023-06-20
[DOCS] Add FrontEnd API note

tsavina:
### Details:
 - *item1*
 - *...*

### Tickets:
 - CVS-111588


docs/MO_DG/prepare_model/convert_model/Convert_Model_From_ONNX.md
```diff
@@ -7,6 +7,8 @@ Introduction to ONNX
 
 `ONNX <https://github.com/onnx/onnx>`__ is a representation format for deep learning models that allows AI developers to easily transfer models between different frameworks. It is hugely popular among deep learning tools, like PyTorch, Caffe2, Apache MXNet, Microsoft Cognitive Toolkit, and many others.
 
+.. note:: ONNX models are supported via FrontEnd API. You may skip conversion to IR and read models directly by OpenVINO runtime API. Refer to the :doc:`inference example <openvino_docs_OV_UG_Integrate_OV_with_your_application>` for more details.
```
rkazants, line 10: We can add here that in advanced cases (like new custom inputs/outputs in case of model pruning, adding pre-processing, Python conversion extensions) user still needs to use `convert_model`. Now it seems that user does not need `convert_model` at all. And it is better to mention exceptional cases.

rkazants (approved):

PR #18158, opened 2023-06-20
[GPU] Not to add sync if the node belongs to shape of subgraph

yeonbok:
### Details:
- If the node is part of shape of subgraph, no need to wait for dependent node because it will be cpu impl

### Tickets:
 - N/A
 -

### DIFF
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -270,6 +270,8 @@ void primitive_inst::update_shape() {
             continue;
         }
         auto& dep = _node->get_dependency(i);
+        if (dep.is_in_shape_of_subgraph())
+            continue;
         auto dep_id = dep.id();
         // exclude fused node from memory_deps
         if (_node->is_fused_dep(i)) {
@@ -285,7 +287,7 @@ void primitive_inst::update_shape() {
         has_runtime_deps = true;
     }
 
-    if (has_runtime_deps && !get_node().is_in_shape_of_subgraph()) {
+    if (has_runtime_deps) {
         if (!dependencies_events.empty() && queue_type == QueueTypes::out_of_order) {
             _network.get_stream().wait_for_events(dependencies_events);
         } else if (queue_type == QueueTypes::in_order) {
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -285,7 +285,7 @@ void primitive_inst::update_shape() {
         has_runtime_deps = true;
     }
 
-    if (has_runtime_deps) {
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -285,7 +285,7 @@ void primitive_inst::update_shape() {
         has_runtime_deps = true;
     }
 
-    if (has_runtime_deps) {
+    if (has_runtime_deps && !get_node().is_in_shape_of_subgraph()) {
```
vladimir-paramuzov, line 288: Not sure that we can really have such condition here. What if we have GPU impl in shape_of subgraph?

sshlyapn: @vladimir-paramuzov currently we use only CPU implementations in subgraphs, so this optimization looks acceptable to me

sshlyapn (approved):

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
vladimir-paramuzov: @sshlyapn That doesn't make me more confident in this change. What if a new model that we haven't checked yet has new op on the shape-of subgraph w/o cpu impl? I expect that in such case we'll face sporadic accuracy issue which will be hard to troubleshoot.

sshlyapn: @vladimir-paramuzov  Hmm, yeah, you are right, it makes sense in case of complex patterns when single shape-of subgraph is splitted by some new operation like this:
image

I think we may additionally check if runtime dependency is in shape-of subgraph and then do not set `has_runtime_deps`. Seems it will be enough to keep improvements from this patch and avoid such problematic cases and , because as of now only nodes with cpu impls may be included to shape-of subgraph.

https://github.com/openvinotoolkit/openvino/blob/52b9df4a6d7cf80833c0e441c11fb80f12d6bfce/src/plugins/intel_gpu/src/graph/graph_optimizer/mark_shape_of_subgraphs.cpp#L59-L65

yeonbok: Thanks for the discussion! Or, how about not to insert the clFinish if the predecessor's impl is cpu ?

yeonbok: Of course the above suggestion can be applied at the same time too.

vladimir-paramuzov: > Or, how about not to insert the clFinish just if the runtime dep predecessor's impl is cpu ?

It could be possible, but please note that such solution may be impacted by some changes in the plugin, e.g. if we implement async execution of cpu impls or use different impl types for static and dynamic kernels

yeonbok: Updated the PR.

PR #18175, opened 2023-06-21
[TF FE] Handle control dependencies to have correct outputs

rkazants:
**Details:** Handle control dependencies to have correct a number of outputs. Nodes having control flow dependents can be terminating.

**Ticket:** 111548

### DIFF
--- a/src/frontends/tensorflow/src/input_model.cpp
+++ b/src/frontends/tensorflow/src/input_model.cpp
@@ -214,6 +214,14 @@ void InputModel::InputModelTFImpl::load_places() {
                                              producer_op_name,
                                              producer_output_port_name,
                                              producer_output_port_idx);
+                if (is_conditional_edge(producer_op_name)) {
+                    // exclude "^" mark indicating (execution) conditional dependency
+                    // for example, "^sub_op" means dependency on a producer node with a name "sub_op"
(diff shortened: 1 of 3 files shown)


src/frontends/tensorflow/src/input_model.cpp
```diff
@@ -214,6 +214,14 @@ void InputModel::InputModelTFImpl::load_places() {
                                              producer_op_name,
                                              producer_output_port_name,
                                              producer_output_port_idx);
+                if (is_conditional_edge(producer_op_name)) {
+                    // exclude "^" mark indicating (execution) conditional dependency
+                    // for example, "^sub_op" means dependency on a producer node with a name "sub_op"
+                    // if a node has dependent operation nodes and has no data consumers,
+                    // this node is not terminating and will not output to the Result node
+                    producer_op_name = producer_op_name.substr(1);
```
popovaan, line 222: So we handle control flow edges as usual edges? Will there be no problem with cyclic control flow dependencies?

rkazants: We handle control dependencies quite the same as data edges but control dependencies do not have input/output ports numbers. They only represent connection between two operation nodes.
The cyclic dependencies for this kind of edge is not possible.

PR #18176, opened 2023-06-21
[PT FE] Add translation for aten::fake_quantize_per_tensor_affine and aten::fake_quantize_per_channel_affine

mmikolajcz:
### Details:
 - *Add translation for aten::fake_quantize_per_channel_affine*
 - *Add translation for aten::fake_quantize_per_tensor_affine*

### Tickets:
 - *110801*


src/frontends/pytorch/src/op/fake_quantize.cpp
```diff
@@ -0,0 +1,102 @@
...
+    auto input_node = context.get_input(0);
+    auto scale = std::make_shared<v0::Convert>(context.get_input(1), element::f32);
+    auto zero_point = std::make_shared<v0::Convert>(context.get_input(2), element::f32);
+    auto out_low_input = context.get_input(3);
+    auto out_high_input = context.get_input(4);
+    auto out_low = std::make_shared<v0::Convert>(out_low_input, element::f32);
+    auto out_high = std::make_shared<v0::Convert>(out_high_input, element::f32);
+    // Set default levels value to 256. This would be correct for bounds like: 0-255 and -128-127
+    int levels = 256;
+    if (std::dynamic_pointer_cast<v0::Constant>(out_low_input.get_node_shared_ptr()) &&
```
mvafin, line 39: I think that should be an exception if we cant calculate levels when inputs are not constants

mvafin: Also please use get_input_from_visible_context function to get constants, or you can even get inputs 4 and 5 as constant values from the very beginning, because if they aren't constant we can not represent them as openvino FakeQuantize anyway

src/frontends/pytorch/src/op/fake_quantize.cpp
```diff
@@ -0,0 +1,102 @@
...
+    auto new_axis = std::make_shared<v1::Select>(is_less, axis_rank, axis);
+    auto new_shape = std::make_shared<v3::ScatterElementsUpdate>(ones, new_axis, const_neg_1, const_0);
+    auto scale_bc = std::make_shared<v1::Reshape>(scale, new_shape, false);
+    auto zero_point_bc = std::make_shared<v1::Reshape>(zero_point, new_shape, false);
+    auto out_low_normalized = std::make_shared<v1::Subtract>(out_low, zero_point_bc);
+    auto out_high_normalized = std::make_shared<v1::Subtract>(out_high, zero_point_bc);
+    auto bound_a = std::make_shared<v1::Multiply>(scale_bc, out_low_normalized);
+    auto bound_b = std::make_shared<v1::Multiply>(scale_bc, out_high_normalized);
+    auto bound_high = std::make_shared<v1::Maximum>(bound_a, bound_b);
+    auto bound_low = std::make_shared<v1::Minimum>(bound_a, bound_b);
```
rkazants, line 94: please add comment for each step because now it is not clear what you do on each step. For others it will be hard to maintain.

src/frontends/pytorch/src/op/fake_quantize.cpp
```diff
@@ -0,0 +1,102 @@
...
+    int levels = 256;
+    if (std::dynamic_pointer_cast<v0::Constant>(out_low_input.get_node_shared_ptr()) &&
+        std::dynamic_pointer_cast<v0::Constant>(out_high_input.get_node_shared_ptr())) {
+        // If bounds are both constant, calculate levels value depending on values.
+        auto out_low_const = context.const_input<int64_t>(4);
+        auto out_high_const = context.const_input<int64_t>(5);
+        levels = abs(out_high_const - out_low_const) + 1;
+    }
+    auto shape_rank = get_shape_rank(context, input_node);
+    auto shape = std::get<0>(shape_rank);
```
rkazants, line 75:
```suggestion
```

rkazants: Looks unused variable

src/frontends/pytorch/src/op/fake_quantize.cpp
rkazants, line 39: you can use `ov::as_type_ptr` instead of `std::dynamic_pointer_cast`. It will be shorter

src/frontends/pytorch/src/op/fake_quantize.cpp
mmikolajcz: Yes, thanks, removed

src/frontends/pytorch/src/op/fake_quantize.cpp
mmikolajcz: Removed this cast, now constants are retrieved as @mvafin suggested.

src/frontends/pytorch/src/op/fake_quantize.cpp
```diff
@@ -0,0 +1,94 @@
...
+OutputVector translate_fake_quantize_per_tensor_affine(const NodeContext& context) {
+    num_inputs_check(context, 5, 5);
+    auto input_node = context.get_input(0);
+    auto scale = std::make_shared<v0::Convert>(context.get_input(1), element::f32);
+    auto zero_point = std::make_shared<v0::Convert>(context.get_input(2), element::f32);
+    auto out_low = context.const_input<int64_t>(3);
+    auto out_high = context.const_input<int64_t>(4);
+    // Calculate levels value - distance between bounds.
+    auto levels = std::abs(out_high - out_low) + 1;
+    auto out_low_const = v0::Constant::create(element::f32, Shape{1}, {out_low});
```
mvafin, line 35: Instead of creating constants here, better would be to use initial inputs obtained by `get_input` that way if FakeQantize will be supporting dynamic levels less changes would be needed and also using initial nodes would make graph easier to match with initial pytorch graph

mvafin:
Generally looks good, one comment

src/frontends/pytorch/src/op/fake_quantize.cpp
mmikolajcz: Done

mvafin (approved):

eaidova (approved):

PR #18181, opened 2023-06-22
[GPU] Memory preallocation optimization

sshlyapn:
### Details:
 - Added MemoryUsageTracker to collect and predict the next possible shapes. As of now it works in 2 modes: preallocation for the next N iterations (applicable only if per dimension diff between shapes is equal or less than 2 - it's enough to cover LLMs cases and avoid huge unneeded memory pre-allocations in other scenarios), and if it is not possible to predict the next N iterations, it uses N% buffer increase (configurable with ov::intel_gpu::buffers_preallocation_ratio property with default value equal to 1.1; 1.0 or smaller value may disable this feature)
 - Fixed CPU impls to use real shapes instead of buffers sizes

### Tickets:
 - 112929, 112930


src/plugins/intel_gpu/include/intel_gpu/runtime/memory.hpp
```diff
@@ -258,4 +259,29 @@ inline ov::Tensor make_tensor(layout l, void* memory_pointer) {
...
+    enum class math_op { SUM, SUB, MUL };
+    ov::Shape shapes_math(const ov::Shape& shape1, const ov::Shape& shape2, math_op op);
+    void add_shape(const std::string& id, const ov::Shape& shape);
+
+    std::map<std::string, std::deque<ov::Shape>> _shapes_info;
+    const engine* _engine;
+
+    const float _buffers_preallocation_ratio = 1.0f;
+    static constexpr size_t _max_deque_size = 3;
+    static constexpr size_t _next_iters_preallocation_count = 10;
```
yeonbok, line 283: Can you add some debug configs to control these parameters for new features?

yeonbok: Also, it is not directly relevant to this PR but it would be good to have another debug config option for "disable shape_of_subgraph_cpu_impl", which will be sometimes good to localize the issue. (It can be added by different PR of course. Just wanted to note)

src/plugins/intel_gpu/src/runtime/memory.cpp
```diff
@@ -45,4 +45,107 @@ std::unique_ptr<surfaces_lock> surfaces_lock::create(engine_types engine_type, s
...
+    // buffer can be reused
+    if (can_reuse_buffer)
+        return {false, {}};
+
+    // Check if there is enough data for prediction
+    auto& shapes = _shapes_info[id];
+    const auto shapes_num = shapes.size();
+
+    // Number of shapes used for iterations mode predictions
+    const auto min_shapes_num = 3;
```
yeonbok, line 99: Can we unify this parameter with max_deque_size?

src/plugins/intel_gpu/src/runtime/memory.cpp
```diff
@@ -45,4 +45,107 @@ std::unique_ptr<surfaces_lock> surfaces_lock::create(engine_types engine_type, s
     }
 }
 
+ov::Shape MemoryUsageTracker::shapes_math(const ov::Shape& shape1, const ov::Shape& shape2, math_op op) {
```
vladimir-paramuzov, line 48: I think several operators(=,-,*) for ov::Shape would look better. I.e.

```cpp
namespace ov {
static ov::Shape operator+(const ov::Shape& s1, const ov::Shape& s2) { }
static ov::Shape operator-(const ov::Shape& s1, const ov::Shape& s2) { }
static ov::Shape operator*(const ov::Shape& s1, const ov::Shape& s2) { }
} // namespace ov

...

        for (size_t i = 0; i < min_shapes_num - 1; ++i) {
            auto result = shapes[shapes_num - i - 1] - shapes[shapes_num - i - 2];
```

src/plugins/intel_gpu/include/intel_gpu/runtime/memory.hpp
```diff
@@ -258,4 +259,29 @@ inline ov::Tensor make_tensor(layout l, void* memory_pointer) {
...
 
+struct MemoryUsageTracker {
+public:
+    MemoryUsageTracker(const engine* engine, float buffers_preallocation_ratio)
+        : _engine(engine)
+        , _buffers_preallocation_ratio(buffers_preallocation_ratio) {
+        static_assert(_max_deque_size >= 2, "[GPU] Deque is supposed to contain at least 2 elements for prediction");
+    }
+
+    std::pair<bool, ov::Shape> predict_preallocated_shape_size(const std::string& id, const ov::Shape& current_shape, bool can_reuse_buffer);
```
vladimir-paramuzov, line 270: Please add a comment that explains what bool in the result pair means exactly

src/plugins/intel_gpu/include/intel_gpu/runtime/memory.hpp
```diff
@@ -258,4 +259,29 @@ inline ov::Tensor make_tensor(layout l, void* memory_pointer) {
...
+    bool can_preallocate(size_t desired_buffer_size);
+
+private:
+    enum class math_op { SUM, SUB, MUL };
+    ov::Shape shapes_math(const ov::Shape& shape1, const ov::Shape& shape2, math_op op);
+    void add_shape(const std::string& id, const ov::Shape& shape);
```
vladimir-paramuzov, line 276: Does it need to works with shapes at all? Won't buffer byte size be enough for tracking & prediction?

src/plugins/intel_gpu/include/intel_gpu/runtime/memory.hpp
sshlyapn: @yeonbok I have added "OV_GPU_MemPreallocationOptions" debug option, please take a look

src/plugins/intel_gpu/src/runtime/memory.cpp
sshlyapn: Good idea, thanks

src/plugins/intel_gpu/include/intel_gpu/runtime/memory.hpp
sshlyapn: Sounds good, but I think shapes themself still may be useful for LLMs cases detection (for example shapes [128, 128] -> [256, 128] -> [192, 256] may be considered as sequential by buffers sizes (16384, 32768, 49152), but this is not a sequence in terms of iterations). So, it looks like we need to collect shapes as well

vladimir-paramuzov: Does it really matter? If you change different dims of the shape on each iteration, but buffer size increase is same on each iter, then why we need to ensure that only specific dim is updated?

sshlyapn: @vladimir-paramuzov, such a sequence is most likely is some kind of successful combination of circumstances, isn't? I'm not sure we can rely a lot on it without extra conditions, otherwise we run the risk of allocation extra memory, which will never be used, especially taking into account the fact that we don't have mechanism for memory usage reduction. I'd suggest to keep shape usage for LLMs cases as an additional more strict rule for initial implementation...
It would be great to track and detect mentioned above cases as well, but it looks like this requires wider buffer size/shape changes history collection to make sure the trend is upward, and then we don't need to have exactly the same size change between iterations; and also it would be great to have mechanism for reducing memory buffer sizes if the last N iterations do not reach maximum size

vladimir-paramuzov (approved):

PR #18183, opened 2023-06-22
Add explicit converts for Parameter and Result in ConvertPrecision tr…

mateusztabaka:
Ticket: CVS-111453


src/common/transformations/include/transformations/convert_precision.hpp
```diff
@@ -81,22 +81,27 @@ class ov::pass::ConvertPrecision : public ov::pass::ModelPass {
     ConvertPrecision(ov::element::Type_t from,
                      ov::element::Type_t to,
                      type_to_fuse_map additional_type_to_fuse_map = {},
-                     bool keep_precision_sensitive_in_fp32 = false)
+                     bool keep_precision_sensitive_in_fp32 = false,
+                     bool convert_input_output_precision = true)
```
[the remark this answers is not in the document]
mateusztabaka: This allows me to merge it as a standalone PR - so I don't have to change plugins in other repos. Then plugins can opt in for that feature whenever they're ready for it.

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -470,17 +523,33 @@ bool fuse_type_to_eye_v9(const std::shared_ptr<ngraph::Node>& node, const precis
...
+        if (convert_input_precision) {
+            param->set_element_type(to);
+            param->validate_and_infer_types();
+            changed = true;
+        } else {
+            auto param_consumers = param->output(0).get_target_inputs();
+            auto convert = std::make_shared<opset4::Convert>(param, to);
+            for (auto& input : param_consumers) {
+                const auto consumer = input.get_node();
+                if (ov::is_type<ov::op::v0::Result>(consumer) || ov::is_type<ov::op::v0::Convert>(consumer)) {
```
itikhono, line 544: what about Sink operations (Assign op)?

src/common/transformations/include/transformations/convert_precision.hpp
itikhono, line 85: do we plan to delete `convert_input_output_precision` mode? or is it useful for some cases?

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -173,20 +179,33 @@ bool convert_function_precision(
...
     bool is_changed,
-    bool is_subgraph) {
+    bool is_subgraph,
+    bool convert_input_output_precision) {
     bool is_output_precision_changed = false;
 
-    auto ops = f->get_ordered_ops();
+    ov::element::TypeVector orig_result_types;
+    if (!convert_input_output_precision) {
+        const auto& results = f->get_results();
```
itikhono, line 188: do we need to handle Sink ops here?

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -252,6 +272,37 @@ bool convert_function_precision(
         }
     }
 
+    if (is_changed && !convert_input_output_precision) {
+        auto& results = f->get_results();
```
itikhono, line 276: the same for Sink ops?

src/common/transformations/src/transformations/convert_precision.cpp
```diff
@@ -252,6 +272,37 @@ bool convert_function_precision(
...
+        for (size_t i = 0; i < results.size(); i++) {
+            auto& result = results[i];
+            if (result->get_input_element_type(0) != orig_result_types[i]) {
+                auto result_input = result->input_value(0);
+                const auto convert = std::make_shared<ov::op::v0::Convert>(result_input, orig_result_types[i]);
+                if (result_input.get_node()->get_output_size() > 1) {
```
itikhono, line 282: is it possible to use op::util::create_ie_output_name function here?

itikhono (approved):

nkogteva (approved):

src/common/transformations/src/transformations/convert_precision.cpp
mateusztabaka: I'd prefer not to - if tensor associated with 'result_input' has legacy name set - create_ie_output_name will return it instead of this node friendly name. So I prefer to update node friendly name and and tensor legacy name separately.

src/common/transformations/src/transformations/convert_precision.cpp
mateusztabaka: I don't think we should handle them in ConvertPrecision. For example if we'd like keep Assign original precision (e.g. f64) its ReadValue counterpart will already have its input precision changed to f32 and we'd get a mismatch.

PR #18189, opened 2023-06-22
Introduce ov element type visitor

praasz:
### Details:
 - Simplify evaluate implementation and code code maintenance by automatic code generation over switch case and macro
 - Applied visitor reduce openvinolib.so by around 27 KiB (Ubuntu 20.04, gcc 9.4, release build no LTO).

### Tickets:
 N/A


src/core/shape_inference/include/element_visitor.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+namespace ov {
+namespace element {
+
+/**
+ * @brief Defines supported elements for applying the element visitor action.
+ *
+ * @tparam List of supported ov::element types.
+ *
+ * The apply function check input element type, if is on the list apply function of visitor for specific type
```
t-jankowski, line 20: Does it regard only input elements? Or may it be used for just element type?

src/core/shape_inference/include/element_visitor.hpp
```diff
@@ -0,0 +1,118 @@
...
+ * if not found apply default action.
```
t-jankowski, line 21:
```suggestion
 * The apply function checks -input- element type whether it is on the list ?of? apply function of visitor for specific type
 * if not found default action is applied.
```

src/core/shape_inference/include/element_visitor.hpp
```diff
@@ -0,0 +1,118 @@
...
+struct NoAction<void> {
+    using result_type = void;
+
+    static void visit() {}
+};
+
+/**
+ * @brief Helper visitor which throws ov::Exception for not supported element type.
+ *
+ * @tparam R Type of return type (used to be compatible with others call operator in Visitor).
```
t-jankowski, line 102:
```suggestion
 * @tparam R Type of return type (used to be compatible with other call operator in Visitor).
```

src/core/shape_inference/include/element_visitor.hpp
```diff
@@ -0,0 +1,118 @@
...
+     */
+    template <class Visitor, class... Args>
+    static auto apply(Type_t et, Args&&... args) -> typename Visitor::result_type {
+        return (et == ET) ? Visitor::template visit<ET>(std::forward<Args>(args)...)
+                          : IfTypeOf<Others...>::template apply<Visitor>(et, std::forward<Args>(args)...);
+    }
+};
+
+/**
+ * @brief Helper visitor which define no action for not supported type.
```
t-jankowski, line 73:
```suggestion
 * @brief Helper visitor which defines no action for not supported type.
```

src/core/shape_inference/include/element_visitor.hpp
praasz: Comment updated.

src/core/shape_inference/include/element_visitor.hpp
praasz: Comment updated.

src/core/shape_inference/include/element_visitor.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * @brief Helper visitor which defines no action for not supported type.
+ *
+ * @tparam R     Type of return value.
+ * @tparam value Default value returned.
+ */
+template <class R, R... value>
+struct NoAction {
+    static_assert(sizeof...(value) < 2, "There should no more then one result value.");
```
t-jankowski, line 84:
```suggestion
    static_assert(sizeof...(value) < 2, "There should be no more than one result value.");
```

src/core/shape_inference/include/element_visitor.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    using result_type = R;
+
+    static constexpr result_type visit() {
+        return {value...};
+    }
+};
+
+/**
+ * @brief Helper visitor which define no action for not supported type if result is void type.
```
t-jankowski, line 94:
```suggestion
 * @brief Helper visitor which defines no action for not supported type if result is void type.
```

src/core/src/op/round.cpp
```diff
@@ -15,50 +16,48 @@ using namespace std;
...
-// function used by COPY_TENSOR
-template <element::Type_t ET>
-inline bool copy_tensor(const HostTensorPtr& arg0, const HostTensorPtr& out, const size_t count) {
-    runtime::reference::copy(arg0->get_data_ptr<ET>(), out->get_data_ptr<ET>(), count);
-    return true;
-}
+class Evaluate : public ov::element::NoAction<bool> {
+    template <element::Type_t ET>
+    static constexpr bool is_floating() {
+        return (ET == element::f16) || (ET == element::f32) || (ET == element::bf16);
```
t-jankowski, line 23: What about f64? At least a comment why not.

src/plugins/intel_cpu/tests/unit/shape_inference_test/element_visitor_test.cpp
```diff
@@ -0,0 +1,104 @@
+// Copyright (C) 2023 Intel Corporation
```
t-jankowski, line 1: Is plugin right place for these tests?

praasz: Moved to core

mlukasze:
@ilyachur & @mitruska could you review it, please?

src/core/src/op/round.cpp
```diff
@@ -15,50 +16,48 @@ using namespace std;
...
+    }
 
+public:
+    using ov::element::NoAction<bool>::visit;
+    template <element::Type_t ET, typename std::enable_if<!is_floating<ET>()>::type* = nullptr>
+    static result_type visit(const HostTensorPtr& arg0,
+                             const HostTensorPtr& out,
+                             const size_t count,
+                             const op::v5::Round::RoundMode) {
+        memcpy(out->get_data_ptr(), arg0->get_data_ptr(), out->get_size_in_bytes());
```
ilyachur, line 33: From my point of view will be better to have this differentiation inside `ngraph::runtime::reference::round()` function. Inside this function we need to select the optimal implementation.

ilyachur (approved):
The idea LGTM

src/core/src/op/round.cpp
praasz: Agree, created ticket to refactor reference implementation and apply this change.

PR #18190, opened 2023-06-22
[PT FE]: fix constant folding for dequantization

eaidova:
### Details:
 - *added transformation for tuple construct + tuple unpack pattern, occurs in non-frozen models*
 - *added mark dequantize subgraph transformation on frontend side*

### Tickets:
 - *CVS-113593*

### DIFF
--- a/src/frontends/pytorch/src/frontend.cpp
+++ b/src/frontends/pytorch/src/frontend.cpp
@@ -157,7 +157,8 @@ void FrontEnd::normalize(const std::shared_ptr<ov::Model>& model) const {
     manager.register_pass<ov::pass::ConvertConvertLike>();
     manager.register_pass<ov::frontend::pytorch::pass::AtenIndexToSelect>();
 
-    manager.register_pass<ov::pass::MarkDequantizationSubgraph>(element::TypeVector{element::u8});
+    manager.register_pass<ov::pass::MarkDequantizationSubgraph>(
+        element::TypeVector{element::u8, element::i8, element::u4, element::i4});
     manager.register_pass<ov::pass::ConstantFolding>();
     manager.register_pass<ov::pass::PushConstantToSubgraph>();
     manager.register_pass<ov::pass::UnrollIf>();
--- a/src/frontends/pytorch/src/frontend.cpp
+++ b/src/frontends/pytorch/src/frontend.cpp
@@ -16,6 +16,7 @@
 #include "transformations/common_optimizations/remove_multi_subgraph_op_dangling_params.hpp"
 #include "transformations/common_optimizations/reverse_shape_and_type_infer.hpp"
 #include "transformations/control_flow/unroll_if.hpp"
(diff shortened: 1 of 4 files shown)


src/frontends/pytorch/src/frontend.cpp
```diff
@@ -155,6 +157,7 @@ void FrontEnd::normalize(const std::shared_ptr<ov::Model>& model) const {
     manager.register_pass<ov::pass::ConvertConvertLike>();
     manager.register_pass<ov::frontend::pytorch::pass::AtenIndexToSelect>();
 
+    manager.register_pass<ov::pass::MarkDequantizationSubgraph>(element::TypeVector{element::u8});
```
slyalin, line 160: Why only u8? List all types that we can support in theory (u8, i8, i4, u4).

src/frontends/pytorch/src/transforms/tuple_unpack_replacer.cpp
```diff
@@ -0,0 +1,50 @@
...
+        auto tuple_construct = cast_fw_node(input_node, "prim::TupleConstruct");
+        if (!tuple_construct) {
+            return false;
+        }
+        for (const auto& input : input_node->inputs()) {
+            const auto& out = input.get_source_output();
+            outputs.push_back(out);
+            to_copy_rt.push_back(out.get_node_shared_ptr());
+        }
+        copy_runtime_info({tuple_unpack, input_node}, to_copy_rt);
```
slyalin, line 36: Not sure that this is a correct RT transfer. `to_copy_rt` contains all nodes that provides inputs to `TupleConstruct`, why do we need to fuse RT into all of them? Looks like there is no any other place suitable for that but can we just lose this info? What are the implications?

src/frontends/pytorch/src/transforms/tuple_unpack_replacer.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "openvino/pass/pattern/matcher.hpp"
+#include "openvino/pass/pattern/op/wrap_type.hpp"
+#include "utils.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace pass {
+
+PrimTupleUnpackReplacer::PrimTupleUnpackReplacer() {
```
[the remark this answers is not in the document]
eaidova: added

src/frontends/pytorch/src/frontend.cpp
eaidova: added more precisions

src/frontends/pytorch/src/transforms/tuple_unpack_replacer.cpp
eaidova: I think there is no more impack here if we will no copy info as there is no additional context between packing and unpacking of these nodes, so I removed copying

slyalin (approved):

PR #18193, opened 2023-06-22
[IE TESTS][CONFORMANCE] Change data_generation initialization to avoid incorrect value

iefode:
### Details:
 - *Change data_generation initialization to avoid incorrect value*
 - *Add w/a to generate correct input to replace constants in conformance*

### Tickets:
 - *ticket-id*


sbalandi (approved):

iefode:
@EgorDuplensky FYI.
I have made these changes to avoid the following situation:
we need to genereate in [0.08, 0.18]
In this case: start_from is 0, range is also 0 but it is incorrect. Start from will be 0 in any cases
With updated algo: start_from = 8, range = 10, resolution = 100 -> all data values is inside [0.08, 0.18)

src/tests/ie_test_utils/common_test_utils/data_utils.hpp
```diff
@@ -195,16 +195,22 @@ template<class T>
...
     const uint32_t k_range = k * range; // range with respect to k
     random.Generate(k_range);
 
     if (start_from < 0 && !std::is_signed<T>::value) {
         start_from = 0;
     }
-
     for (std::size_t i = 0; i < size; i++) {
-        pointer[i] = static_cast<T>(start_from + static_cast<T>(random.Generate(k_range)) / k);
+        pointer[i] = static_cast<T>((start_from + static_cast<T>(random.Generate(k_range))) / k);
```
EgorDuplensky, line 213: Now this function is not trivial, and requires some preprocessing of the start_from argument (I would propose to add a description comment).
Is it possible to allow to specify start_from as a float value instead? Or this would require much more changes?

iefode: from high level nothing does change (check the code). Generators work with int numbers. My suggestion is use such construction. Moreover, the changes which we merged before is not so trivial :)

PR #18209, opened 2023-06-23
[GNA] Fix additional warnings introduced after removal of warning suppression

rjeziers:
### Details:
Fixed warnings introduced after this PR: https://github.com/openvinotoolkit/openvino/pull/17841

### Tickets:
 - 93715

PR #18216, opened 2023-06-23
[GNA]Fix conversion warning in gna plugin

praasz:
### Details:
 - Fix warnings as error on windows build

### Tickets:
N/A


rjeziers (commented):
Should be already fixed with https://github.com/openvinotoolkit/openvino/pull/18209

praasz:
> Should be already fixed with #18209
If not reported by CI here are still missing some casts. I got some wraning as error on local build so it could at some time make issue on CIs

ilya-lavrenov (approved):

PR #18228, opened 2023-06-25
[GPU] memory reuse for dynamic models

e-ddykim:
### Details:
 - This PR enables memory reuse for dynamic models.
   * updated `user_requesting_mem_reuse_false` not to return true when dynamic models
   * updated `user_requesting_mem_reuse_false` to use `primitive_inst` instead of `program_node` in runtime
   * updated memory_pool.release_memory() to compare buffer_ptrs 

### Tickets:
 - 91383

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/cpu/shape_of.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/cpu/shape_of.cpp
@@ -40,8 +40,6 @@ struct shape_of_impl : public typed_primitive_impl<shape_of> {
         OV_ITT_SCOPED_TASK(ov::intel_gpu::itt::domains::intel_gpu_plugin, "shape_of::execute_impl");
         auto& stream = instance.get_network().get_stream();
 
-        auto ev = stream.create_user_event(false);
-
         auto output_mem_ptr = instance.output_memory_ptr();
 
         auto output_dt = instance.get_impl_params()->get_output_layout().data_type;
@@ -60,9 +58,13 @@ struct shape_of_impl : public typed_primitive_impl<shape_of> {
             OPENVINO_THROW("[GPU] Couldn't execute shape_of operation: unsupported output data type (", output_dt , ")");
         }
 
-        ev->set();
-
-        return ev;
+        if (events.size() > 1) {
+            return stream.group_events(events);
+        } else if (events.size() == 1) {
+            return events[0];
+        } else {
+            return stream.create_user_event(true);
+        }
     }
 
     void init_kernels(const kernels_cache& , const kernel_impl_params&) override {}
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -1086,7 +1086,7 @@ memory::ptr primitive_inst::allocate_output(engine& _engine, memory_pool& pool, 
                         const kernel_impl_params& impl_params, uint32_t net_id, bool is_internal, size_t idx, bool reset, bool is_output_buffer,
                         bool runtime_alloc, memory* curr_memory) {
     auto get_memory_from_pool = [&](engine& _engine, const layout& layout, const primitive_id id, std::set<primitive_id> dependencies,
-            allocation_type type, bool reusable, bool reset = true, memory* curr_memory) {
+            allocation_type type, bool reusable, bool reset = true, memory* curr_memory = nullptr) {
         OPENVINO_ASSERT(!layout.is_dynamic() || layout.has_upper_bound(), "[GPU] Can't allocate output for dynamic layout without upper bound");
         // Use layout with max tensor for dynamic shape with upper bound
         auto static_layout = cldnn::layout(layout.get_partial_shape().get_max_shape(), layout.data_type, layout.format, layout.data_padding);
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -1082,8 +1082,9 @@ static bool user_requesting_mem_reuse_false(const program_node& node, const prim
     return false;
 }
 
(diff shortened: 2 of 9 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1054,30 +1056,44 @@ event::ptr primitive_inst::update_weights() {
...
-static bool user_requesting_mem_reuse_false(const program_node& node) {
-    for (auto& user : node.get_users()) {
-        if (user->is_dynamic())
-            return true;
-        if ((user->get_selected_impl() != nullptr) && (user->get_selected_impl()->can_reuse_memory == false)) {
-            return true;
-        } else if (user->get_selected_impl() == nullptr) {
-            if (user_requesting_mem_reuse_false(*user)) {
+static bool user_requesting_mem_reuse_false(const program_node& node, const primitive_inst* prim, bool runtime_alloc) {
+    if (prim == nullptr || runtime_alloc == false) {
```
yeonbok, line 1060: Is there any case where  runtime_alloc == true && prim == nullptr?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1054,30 +1056,44 @@ event::ptr primitive_inst::update_weights() {
...
+    } else {
+        for (auto& user : prim->get_user_insts()) {
+            if ((user->get_impl() != nullptr) && (user->get_impl()->can_reuse_memory == false)) {
+                return true;
+            } else if (user->get_impl() == nullptr) {
+                if (user_requesting_mem_reuse_false(user->get_node(), user.get(), runtime_alloc)) {
+                    return true;
+                }
             }
         }
```
yeonbok, line 1079: ah... I just realized that the "user impl" is not decided yet when allocating current inst's output. 
I think maybe we need to adjust the conditions further. Let's think about it later...

yeonbok:
Could you add a unittest to check whether the intended memory is reused?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
yeonbok: And now, I think we don't need "user_requesting_mem_reuse_false" at all because it should be dealt by need_reset_output_memory, if the user requires memory reset. (e.g., we can add this function to some types of reduce)

src/plugins/intel_gpu/src/runtime/memory_pool.cpp
```diff
@@ -55,13 +55,12 @@ void memory_pool::release_memory(memory* mem, const primitive_id& id, uint32_t n
...
-        auto range = _non_padded_pool.equal_range(_layout.bytes_count());
-        auto it = range.first;
+        auto it = _non_padded_pool.lower_bound(_layout.bytes_count());
 
-        while (it != range.second && it != _non_padded_pool.end()) {
+        while (it != _non_padded_pool.end()) {
             if (it->second._network_id == network_id &&
                 it->second._type == type &&
-                it->second._memory.get() == mem) {
+                it->second._memory->buffer_ptr() == mem->buffer_ptr()) {
```
sshlyapn, line 63: `buffer_ptr()` is available only for USM buffers, so we need to handle OpenCL buffers independently

src/plugins/intel_gpu/src/graph/impls/cpu/shape_of.cpp
```diff
@@ -60,9 +58,13 @@ struct shape_of_impl : public typed_primitive_impl<shape_of> {
             OPENVINO_THROW("[GPU] Couldn't execute shape_of operation: unsupported output data type (", output_dt , ")");
         }
 
-        ev->set();
-
-        return ev;
+        if (events.size() > 1) {
+            return stream.group_events(events);
```
sshlyapn, line 62: Could you please clarify, why do we need to propagate input events to the next operations in case of shape_of? Wouldn't it cause negative performance impact?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
e-ddykim: I updated it not to use `user_requesting_mem_reuse_false` in the case of dynamic models, as you guided. Thank you.

src/plugins/intel_gpu/src/graph/impls/cpu/shape_of.cpp
e-ddykim: Previously, `shape_of` did not wait for dependent events, causing accuracy issues.  GPU plugin's synchronization logic assumes that when an event completed, all previous events have already finished. Prior to this PR, the `shape_of` made a custom event and set it as finished without checking the status of its dependent events. As a result, the users of `shape_of` were assuming that all previous events had completed and proceeding, when in fact there were still events that hadn't completed. So, I updated it to return the dependent events to ensure that `shape_of`'s users are executed after `shape_of`'s dependent events are finished

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1097,7 +1100,7 @@ memory::ptr primitive_inst::allocate_output(engine& _engine, memory_pool& pool,
     if (total_device_input_mem_size > _engine.get_device_info().max_global_mem_size)
         usm_device_allocatable = false;
 
-    bool memory_reuse_by_user = !user_requesting_mem_reuse_false(_node);
+    bool memory_reuse_by_user = _node.is_dynamic() ? !reset : !user_requesting_mem_reuse_false(_node);
```
yeonbok, line 1103: is_dynamic checks both input and output, though allocation need consider only output at each time. So I think we only check is_dynamic_output_layout().

e-ddykim: I replaced it with `is_dynamic_output_layout()` as you reviewed. Thank you.

e-ddykim:
> Could you add a unittest to check whether the intended memory is reused?

I added a unit test for memory reuse of dynamic models.

yeonbok (approved):

PR #18229, opened 2023-06-25
 to be checked

SANTHOSH-MAMIDISETTI:
### Details:
 - *Draft PR so that it can be checked *
 

### Tickets:
 - *ticket-id (not mentioned)*

### DIFF
--- a/src/frontends/onnx/frontend/src/op/com.microsoft/ONNX_Frontend_extension.cpp
+++ b/src/frontends/onnx/frontend/src/op/com.microsoft/ONNX_Frontend_extension.cpp
@@ -1,0 +1,110 @@
+#include "ONNX_Frontend_extension.hpp"
+#include <iostream>
+
+namespace onnx_opset {
+    namespace op {
+        void Pad(const float* input_data, const int64_t* pads_data, float* output_data, const std::vector<int64_t>& input_shape, const std::vector<int64_t>& output_shape, const std::string& mode, float constant_value) {
+            int64_t input_rank = input_shape.size();
+            int64_t output_rank = output_shape.size();
+            int64_t num_axes = input_rank;  // Number of axes to pad
+            int64_t channels = input_shape[1];  // Number of channels
+
+            // Iterate over each element in the output tensor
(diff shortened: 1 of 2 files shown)


src/frontends/onnx/frontend/src/op/com.microsoft/ONNX_Frontend_extension.cpp
```diff
@@ -0,0 +1,110 @@
+#include "ONNX_Frontend_extension.hpp"
+#include <iostream>
+
+namespace onnx_opset {
+    namespace op {
+        void Pad(const float* input_data, const int64_t* pads_data, float* output_data, const std::vector<int64_t>& input_shape, const std::vector<int64_t>& output_shape, const std::string& mode, float constant_value) {
```
mbencer, line 6: Thank you for contibution! Be aware that translation ONNX op to OpenVINO representation has be to be a function in shape:
```
OutputVector pad(const Node& node) {
return OneOrMoreOutoutOfOvSubgraph
}
```
In particular we can't use input tensors data (like `input_data` in your case) if not constants.
Please look at https://github.com/openvinotoolkit/openvino/blob/50897e86e67b623b96e6c9a815a1606f88620b8f/src/frontends/onnx/frontend/src/op/pad.cpp#L68

SANTHOSH-MAMIDISETTI: > Thank you for contribution! Be aware that translation ONNX op to OpenVINO representation has be to be a function in shape:
> 
> ```
> OutputVector pad(const Node& node) {
> return OneOrMoreOutoutOfOvSubgraph
> }
> ```
> 
> In particular we can't use input tensors data (like `input_data` in your case) if not constants. Please look at
> 
> https://github.com/openvinotoolkit/openvino/blob/50897e86e67b623b96e6c9a815a1606f88620b8f/src/frontends/onnx/frontend/src/op/pad.cpp#L68

Thank you for your review @mbencer , I will definitely correct it and let you know . Secondly I would like to know if the implementation logic ,return values and such are 
correct ? If not could you  please point me what went wrong . Thanks!

PR #18247, opened 2023-06-26
[GPU] oneDNN3.3pc integration.

hyunback:
Supports new formats from oneDNN3.3pc requires.

### Tickets:
 - *109146*


hyunback:
Will plan v3.3

PR #18275, opened 2023-06-28
[PyOV] Memory flow control with share_inputs and share_outputs

jiwaszki:
### Details:
 - Added flags that controls both inputs and outputs sharing of inference calls, with potential to reduce memory copy overhead.
 - Updated docstrings for functions.
 - Added testcases that validates memory sharing and positional/keyword solution.
 - `share_outputs` is a new argument.
 - `share_inputs` is taking place of `shared_memory`.
 - Deprecated `shared_memory` in a way that deprecation notice is only raised when called directly with a keyword. Positional arguments are transparently using new `share_inputs`. Examples:
 ```python
 In [3]: a = compiled_model([arr_1, arr_2], shared_memory=True)
<ipython-input-3-0c33129d1c80>:1: FutureWarning: `shared_memory` is deprecated and will be removed in 2024.0. Value of `shared_memory` is going to override `share_inputs` value. Please use only `share_inputs` explicitly.
  a = compiled_model([arr_1, arr_2], shared_memory=True)

In [4]: a = compiled_model([arr_1, arr_2], shared_memory=False)
<ipython-input-4-a0494d0b5d8a>:1: FutureWarning: `shared_memory` is deprecated and will be removed in 2024.0. Value of `shared_memory` is going to override `share_inputs` value. Please use only `share_inputs` explicitly.
  a = compiled_model([arr_1, arr_2], shared_memory=False)

In [5]: a = compiled_model([arr_1, arr_2], True)

In [6]: a = compiled_model([arr_1, arr_2], True, False)

In [7]: a = compiled_model([arr_1, arr_2], share_inputs=True, share_outputs=True)
 ```
- TODO: in a separate PR update documentation page of Python Features.

### Tickets:
 - *107861*


ilya-lavrenov (approved):

src/bindings/python/src/openvino/runtime/ie_api.py
```diff
@@ -22,10 +23,30 @@
 )
 
 
+def _deprecated_memory_arg(shared_memory: bool, share_inputs: bool) -> bool:
+    if shared_memory is not None:
+        warnings.warn(
+            "`shared_memory` is deprecated and will be removed in 2024.0. "
+            "Value of `shared_memory` is going to override `share_inputs` value. "
```
yury-gorbachev, line 30: If person is using share_inputs then he is probably aware that this exists and also aware that it replaced shared_memory, so this information about overriding probably does not add much value here. Instead, if user uses this old flag - probably he is not aware that there is new one.
Could we just say - we recommend to use share_inputs instead?

yury-gorbachev (approved):

src/bindings/python/src/openvino/runtime/ie_api.py
jiwaszki: > this information about overriding probably does not add much value here

This information is only shown when `shared_memory` is explicitly used. Which actually make sense as only users of the old flag will be notified. (+ there is `"Please use only share_inputs explicitly."` at the end of message)

If the new flag `share_inputs` or positional argument is used, there is nothing to notify about. Function is ran without warning.

Overall the whole problem boils down to the priority of flags which do the same thing. In summary:

- (until deprecation) `shared_memory` will trigger a warning about overriding.
- anything like `infer(..., shared_memory=True, share_inputs=False)` wouldn't be consider undefined behavior.
- both `infer(..., True)` and `infer(..., share_inputs=True)` would run without warnings.

akuporos (approved):

p-wysocki (approved):

PR #18283, opened 2023-06-29
Modify the condition making batch interpretation true/false

davidsnam-intel:
### Details:
 - *When the user is Convert for Constant node, and tensor is 1d, Set needBatchInterpretation to true*

### Tickets:
 - *111516*

### DIFF
--- a/src/plugins/intel_gpu/src/plugin/ops/constant.cpp
+++ b/src/plugins/intel_gpu/src/plugin/ops/constant.cpp
@@ -106,12 +106,22 @@ static void CreateConstantOp(Program& p, const std::shared_ptr<ngraph::op::v0::C
         }
     };
 
+    auto is_all_inputs_1d = [&] (ov::Node* op) -> bool {
+        for (size_t i = 0; i < op->get_input_size(); i++) {
+            auto& in_shape = op->get_input_partial_shape(i);
+            if (in_shape.size() > 1)
+                return false;
+        }
+        return true;
+    };
+
     auto is_convert_into_binary_eltwise = [&] (ov::Node* op) -> bool {
         if (ngraph::is_type<ngraph::op::v0::Convert>(op)) {
             for (size_t i = 0; i < op->get_output_size(); ++i) {
                 auto convertUsers = op->get_output_target_inputs(i);
                 for (auto user : convertUsers) {
-                    if (is_binary_eltwise(user.get_node())) {
+                    if (is_binary_eltwise(user.get_node()) &&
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gpu/src/plugin/ops/constant.cpp
```diff
@@ -109,7 +109,8 @@ static void CreateConstantOp(Program& p, const std::shared_ptr<ngraph::op::v0::C
         } else if (ngraph::op::is_binary_elementwise_arithmetic(outOp) ||
                    ngraph::op::is_binary_elementwise_logical(outOp) ||
                    ngraph::op::is_binary_elementwise_comparison(outOp) ||
-                   ngraph::is_type<ngraph::op::v0::SquaredDifference>(outOp)) {
+                   ngraph::is_type<ngraph::op::v0::SquaredDifference>(outOp) ||
+                   ngraph::is_type<ngraph::op::v0::Convert>(outOp)) {
```
yeonbok, line 113: This is applied too broad range of cases...

Instead, I suggest to do like, 
 if "outOp" is convert, than check whetherits predecessor is one of "binary_elementwise_arithmetic, binary_elementwise_logical, binary_elementwise_comparison, squareddifference" and then do the original logic

src/plugins/intel_gpu/src/plugin/ops/constant.cpp
```diff
@@ -106,17 +140,10 @@ static void CreateConstantOp(Program& p, const std::shared_ptr<ngraph::op::v0::C
...
+                   is_convert_into_binary_eltwise(outOp) ||
                    ngraph::is_type<ngraph::op::v0::SquaredDifference>(outOp)) {
-            bool all_inputs_1d = true;
-            for (size_t j = 0; j < outOp->get_input_size(); j++) {
-                auto& in_shape = outOp->get_input_partial_shape(j);
-                if (in_shape.size() > 1)
-                    all_inputs_1d = false;
-            }
-            consts[op].needsBatchInterpretation = all_inputs_1d && constDims.size() == 1;
+            consts[op].needsBatchInterpretation = is_all_inputs_1d(outOp) && constDims.size() == 1;
```
yeonbok, line 146: `else if ((is_binary_eltwise(outOp) || (ngraph::is_type<ngraph::op::v0::SquareDifference>(outOp)) && is_all_inputs_1d(outOp)) || is_convert_into_binary_eltwise(outOp)) {
    const[op].needsBatchInterpretation = constDims.size() == 1;
}`

yeonbok (approved):

PR #18311, opened 2023-06-30
[AUTO] AUTO can't fall back if target device throws the exception before compiling model via AUTO

yangwang201911:
### Details:
 - Update AUTO candidate device filtering logic.
 - Add AUTO handling the exception logic when generating the target device infomation.

### Tickets:
 - 114170


src/plugins/auto/tests/unit/get_device_list.cpp
```diff
@@ -134,25 +137,29 @@ const std::vector<Params> testConfigs = {Params{" ", " "},
                                          Params{"CPU,VPUX", "CPU,VPUX"},
                                          Params{"CPU,-VPUX", "CPU"},
                                          Params{"CPU,-INVALID_DEVICE", "CPU"},
+                                         Params{"CPU,INVALID_DEVICE", "CPU"},
+                                         Params{"INVALID_DEVICE", ""},
```
songbell, line 141: add {"INVALID_DEVICE,GPU,CPU", ""},{"INVALID_DEVICE,CPU", ""},

yangwang201911: This test case is just used for check the logic of filtering the valid devices from the inputting property `ov::device::priorities`. 
From my understanding,  `"GPU,CPU" `will be the expected final valid device candidate list for the case `"INVALID_DEVICE,GPU,CPU"` as the specified devices candidate list.
Could you help explain why the empty device list is returned for the mentioned 2 test cases.

songbell: not empty, just add "INVALID_DEVICE,GPU,CPU" this kind of test

yangwang201911: Got it.

src/plugins/auto/src/plugin.cpp
```diff
@@ -796,8 +807,9 @@ std::string Plugin::get_device_list(const ov::AnyMap& properties) const {
                         // The device is the device with default device ID (eg. GPU.0) and
                         // its wide name (eg. GPU) has been in device candidate list.
                         continue;
-                    // Add user specified device into candidate list
-                    devices_merged.push_back(device);
+                    std::string realName = parsed.get_device_name();
+                    if (is_any_dev(realName, device_list))
+                        devices_merged.push_back(device);
```
songbell, line 812: this is a must change?

yangwang201911: Yes, this logic will filter out the device which is not in the registered list.
You mean we can just enable this filtering logic in the function `AUTO::parse_meta_devices()` only?
As discussed with @peterchen-intel , we implemented this logic in both function AUTO::get_device_list() and AUTO::parse_meta_devices() for double insurance.

yangwang201911: I found there is a test suit for` CachingTest` that needs to load network to` AUTO:mock.0,mock.1,....`.
@peterchen-intel , so should we remove this filtering logic from `AUTO::get_device_list()` and just try to catch the exception within the `AUTO::parse_meta_devices()` function?

yangwang201911: Have reverted the logic of `AUTO:get_device_list()`.

src/plugins/auto/tests/unit/parse_meta_device_test.cpp
```diff
@@ -142,6 +149,7 @@ const std::vector<ConfigParams> testConfigs = {
 
     ConfigParams {"CPU(-1),GPU,VPUX",  {}, true},
     ConfigParams {"CPU(NA),GPU,VPUX",  {}, true},
+    ConfigParams {"INVALID_DEVICE",  {}, false},
```
songbell, line 152: can you add INVALID_DEVICE,CPU, btw, why the second {} is empty here?

src/plugins/auto/tests/unit/parse_meta_device_test.cpp
```diff
@@ -134,6 +136,11 @@ const std::vector<ConfigParams> testConfigs = {
              {"GPU.0", {}, -1, "", std::string(igpuFullDeviceName) + "_0", 1},
              {"GPU.1", {}, -1, "", std::string(dgpuFullDeviceName) + "_1", 1},
              {"CPU", {}, -1, "", "CPU_", 2}}, false},
+    ConfigParams {"VPUX,GPU,INVALID_DEVICE",
+         {{"VPUX", {}, -1, "", "VPUX_", 0},
```
wangleis, line 140: In 2023.1 release, we should use "VPU" in test case.

src/plugins/auto/tests/unit/parse_meta_device_test.cpp
yangwang201911: Because an exception will be occurred as expected while getting property from INVALID_DEVICE.. An empty list of meta devices will be returned.

src/plugins/auto/tests/unit/parse_meta_device_test.cpp
yangwang201911: I need to rename all of the VPUX to VPU in this file if we changed here to pass this added test case for VPU instead of VPUX.  is this our expected behavior?

src/plugins/auto/tests/unit/parse_meta_device_test.cpp
yangwang201911: btw, the second {} means nth because the 3rd parameter `true` means an exception will be expected.

peterchen-intel: Covered by  [PR#17949](https://github.com/openvinotoolkit/openvino/pull/17949)

src/plugins/auto/tests/unit/parse_meta_device_test.cpp
peterchen-intel: @wangleis  Being covered by https://github.com/openvinotoolkit/openvino/pull/17949

songbell (approved):

PR #18312, opened 2023-06-30
[MO] add support for scalar shapes into cli_parser.py

pavel-esir:
#### Details:
 - corrected regular expression so that scalar shape in `--input_shape=inp[]` or `--input_shape=input1[1,100,100,3],input2[]`
 - `--input` argument already has support of scalar input shapes

#### Tickets:
 - xxx-99609


tools/mo/openvino/tools/mo/utils/cli_parser.py
```diff
@@ -1490,15 +1490,15 @@ def split_inputs(input_str):
...
+    first_digit_reg = r'([0-9 ]*|-1|\?|{})'.format(range_reg)
     next_digits_reg = r'(,{})*'.format(first_digit_reg)
     tuple_reg = r'((\({}{}\))|(\[{}{}\]))'.format(first_digit_reg, next_digits_reg,
                                                   first_digit_reg, next_digits_reg)
 
     full_reg = r'^{}(\s*,\s*{})*$|^$'.format(tuple_reg, tuple_reg)
     if not re.match(full_reg, argv_input_shape):
         raise Error('Input shape "{}" cannot be parsed. ' + refer_to_faq_msg(57), argv_input_shape)
-    return re.findall(r'[(\[]([0-9,\.\? -]+)[)\]]', argv_input_shape)
+    return re.findall(r'[(\[]([0-9,\.\? -]*)[)\]]', argv_input_shape)
```
popovaan, line 1501: Do we need convert_model() to support following syntax?
convert_model(input_shape=[])
convert_model(input=InputCutInfo(shape=[]))

tools/mo/openvino/tools/mo/utils/cli_parser.py
```diff
@@ -1579,7 +1579,7 @@ def get_placeholder_shapes(argv_input: str, argv_input_shape: str, argv_batch=No
         # clean inputs from values for freezing
         inputs_without_value = list(map(lambda x: x.split('->')[0], inputs))
         placeholder_shapes = dict(zip_longest(inputs_without_value,
-                                              map(lambda x: PartialShape(x) if x else None, shapes)))
+                                              map(lambda x: PartialShape(x) if (x or x == '') else None, shapes)))
```
rkazants, line 1582: why is this change needed? Please test this case as well.

pavel-esir: This line is needed exactly for the cases with inputs with scalars and all new test cases test this line as well. E.g. if there are inputs `[12,4,1],[]` they will be split into 2 values with `'12,4,1'` and `''` and for the later case we need not `None` but `PartialShape([])` or equivalent `PartialShape('')`

image

tools/mo/openvino/tools/mo/utils/cli_parser.py
pavel-esir: `convert_model` is already supporting inputs with scalar shapes. Added an test case for that

rkazants (approved):

PR #18325, opened 2023-07-02
Fix python segment fault issue of plugin api 2.0

riverlijunjie:
### Details:
 - Fix python segment fault issue of plugin api 2.0

### Tickets:
 - 114261


peterchen-intel:
If framework = pt, we should make all trust_remote_code=True
```
$ python3 ./public/benchmark.py -f pt -d cpu -m /mnt/data3_1878/huggingface/hub/mpt-7b-chat -id databricks/dolly-v2-12b
==SUCCESS FOUND==: use_case: text_gen, model_name: mpt-7b-chat
load text model from model path:/mnt/data3_1878/huggingface/hub/mpt-7b-chat
Loading /mnt/data3_1878/huggingface/hub/mpt-7b-chat requires to execute some code in that repo, you can inspect the content of the repository at https://hf.co//mnt/data3_1878/huggingface/hub/mpt-7b-chat. You can dismiss this prompt by passing `trust_remote_code=True`
```

src/inference/src/dev/icompiled_model.cpp
```diff
@@ -48,34 +48,47 @@ ov::ICompiledModel::ICompiledModel(const std::shared_ptr<const ov::Model>& model
...
-                auto fake_param = std::make_shared<ov::op::v0::Parameter>(result->get_output_element_type(0),
-                                                                          result->get_output_partial_shape(0));
-                const std::string res_name = ov::op::util::create_ie_output_name(result->input_value(0));
+            param->output(0).get_tensor().set_names({param_name});
+            new_param->set_element_type(param->get_element_type());
+            new_param->set_layout(param->get_layout());
+            new_param->output(0).get_rt_info() = param->output(0).get_rt_info();
+            new_param->validate_and_infer_types();
+            m_inputs.emplace_back(
+                ov::Output<const ov::Node>{new_param->output(0).get_node(), param->output(0).get_index()});
```
songbell, line 68: add new nodes here in m_inputs/outputs may introduce other potential issues? like if operate 2.0 public API with legacy 1.0 GPU plugin, this nodes will get duplicated. @ilyachur probably, we need to handle the lifecyle with another proper fix?

ilyachur: Another possible fix is to hold `ov::Model` but in this case we also will hold the full original model inside the CompiledModel.
At first sight the creation new inputs/outputs looks good to me. Will take a look

songbell: @riverlijunjie  port tensors should be reused here:
https://github.com/riverlijunjie/openvino-1/blob/36fedf8bb8aed0236452dd1319723347df7ccaf9/src/inference/src/dev/isync_infer_request.cpp#L266

src/inference/src/dev/icompiled_model.cpp
```diff
@@ -48,34 +48,47 @@ ov::ICompiledModel::ICompiledModel(const std::shared_ptr<const ov::Model>& model
...
-                result->output(0).get_tensor().add_names({res_name});
-                m_outputs.emplace_back(
-                    ov::Output<const ov::Node>{result->output(0).get_node(), result->output(0).get_index()});
             }
-        } else {
-            m_inputs = model->inputs();
-            m_outputs = model->outputs();
+            new_result->output(0).get_tensor().add_names({res_name});
+            auto r = std::dynamic_pointer_cast<ov::op::v0::Result>(new_result);
+            OPENVINO_ASSERT(r, "Internal error. set outputs failure casting output copy to Result");
```
ilyachur, line 88: If I understand right, here you always have Result operation. Propose to avoid unreachable checks. Will be better to work with result operation here.

songbell (approved):

ilyachur (approved):

src/inference/src/dev/icompiled_model.cpp
```diff
@@ -48,34 +48,48 @@ ov::ICompiledModel::ICompiledModel(const std::shared_ptr<const ov::Model>& model
...
+            fake_param->validate_and_infer_types();
+            auto new_result = result->copy_with_new_inputs({fake_param});
+            new_result->set_friendly_name(result->get_friendly_name());
+            if (add_operation_names) {
                 OPENVINO_ASSERT(!m_plugin->is_new_api() || leaf_names.find(res_name) == leaf_names.end() ||
                                     result->output(0).get_names().find(res_name) != result->output(0).get_names().end(),
                                 "Model operation names have collisions with tensor names.",
                                 " Please use MO to generate new IR version, it should allow to avoid the issue");
                 leaf_names.insert(res_name);
-                result->output(0).get_tensor().add_names({res_name});
```
songbell, line 72: @ilyachur why previously need to call add_names(ie_output_name) here?

ilyachur: Good catch! we need to add it for backward compatibility with old API. @riverlijunjie Please revert this change

ilyachur (changes requested):

src/inference/src/dev/icompiled_model.cpp
```diff
@@ -48,34 +48,48 @@ ov::ICompiledModel::ICompiledModel(const std::shared_ptr<const ov::Model>& model
...
+            const auto& param_name = param->get_friendly_name();
+            auto new_param = ov::as_type_ptr<ov::op::v0::Parameter>(param->copy_with_new_inputs({}));
+            new_param->set_friendly_name(param_name);
+            if (add_operation_names) {
                 OPENVINO_ASSERT(!m_plugin->is_new_api() || leaf_names.find(param_name) == leaf_names.end() ||
                                     param->output(0).get_names().find(param_name) != param->output(0).get_names().end(),
                                 "Model operation names have collisions with tensor names.",
                                 " Please use MO to generate new IR version, it should allow to avoid the issue");
                 leaf_names.insert(param_name);
-                param->output(0).get_tensor().add_names({param_name});
```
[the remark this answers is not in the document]
riverlijunjie: yes, we do need revert it, else it will bring some problem to compatible with IR 10.

src/inference/src/dev/icompiled_model.cpp
```diff
@@ -48,34 +48,47 @@ ov::ICompiledModel::ICompiledModel(const std::shared_ptr<const ov::Model>& model
...
+            fake_param->validate_and_infer_types();
+            auto new_result = result->copy_with_new_inputs({fake_param});
+            new_result->set_friendly_name(result->get_friendly_name());
+            if (add_operation_names) {
                 OPENVINO_ASSERT(!m_plugin->is_new_api() || leaf_names.find(res_name) == leaf_names.end() ||
                                     result->output(0).get_names().find(res_name) != result->output(0).get_names().end(),
                                 "Model operation names have collisions with tensor names.",
                                 " Please use MO to generate new IR version, it should allow to avoid the issue");
                 leaf_names.insert(res_name);
                 result->output(0).get_tensor().add_names({res_name});
```
ilyachur, line 85: I think we can add names only for new result. and we shouldn't modify original model

riverlijunjie: If not add such names, it will lead to some test failure:  legacy public api + plugin api 2.0 test import model.

`smoke_BehaviorTests/OVCompiledGraphImportExportTest.importExportedIENetwork/targetDevice=CPU_elementType=i8_`

It is caused by name missing when model export.

ilyachur (approved):

PR #18328, opened 2023-07-03
[Do-not-review]test post commits - Bell/fix onnx multi output failure

songbell:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


src/inference/src/dev/icompiled_model.cpp
```diff
@@ -48,34 +48,47 @@ ov::ICompiledModel::ICompiledModel(const std::shared_ptr<const ov::Model>& model
...
                 leaf_names.insert(param_name);
-                param->output(0).get_tensor().add_names({param_name});
-                m_inputs.emplace_back(
-                    ov::Output<const ov::Node>{param->output(0).get_node(), param->output(0).get_index()});
             }
-            for (const auto& result : model->get_results()) {
-                auto fake_param = std::make_shared<ov::op::v0::Parameter>(result->get_output_element_type(0),
-                                                                          result->get_output_partial_shape(0));
-                const std::string res_name = ov::op::util::create_ie_output_name(result->input_value(0));
+            // param->output(0).get_tensor().set_names({param_name});
```
songbell, line 62: @riverlijunjie I comment the set_names here, otherwise the port check will fail

riverlijunjie: If we comment the set_names, the `onnx_duplicated_output_name` can pass, but as we discussed that the graph after transformation still be different with  the master branch's result during we run below test case:
./ov_onnx_frontend_tests --gtest_filter=INTERPRETER.onnx_duplicated_output_name

1. original model:
image

2. master branch's transformation result:

image


3. This pr's transformation results:
image

PR #18208, opened 2023-06-23
[CPU] Fixed naming in SwapConvertTranspose pass

a-sidorova:
### Details:
 - *The CPU Plugin Pass `SwapConvertTranspose` swaps the nodes but doesn’t change the output name.  The PR adds the updates of names and refactors the code a little bit*
 - *Added unit test that verifies: names, graph and runtime info*

### Tickets:
 - *113703*

PR #18329, opened 2023-07-03
Fix friendly names in SwapConvertTranspose transformation

mateusztabaka:
Ticket: CVS-113703

### DIFF
--- a/src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/swap_convert_transpose.cpp
+++ b/src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/swap_convert_transpose.cpp
@@ -32,11 +32,12 @@ ov::intel_cpu::SwapConvertTranspose::SwapConvertTranspose() {
         transposeInputs[0] = convert->input_value(0);
         auto newTranspose = transpose->clone_with_new_inputs(transposeInputs);
         ngraph::copy_runtime_info(transpose, newTranspose);
-        newTranspose->set_friendly_name(transpose->get_friendly_name());
+        newTranspose->set_friendly_name(convert->get_friendly_name());
 
         ngraph::OutputVector convertInputs = convert->input_values();
         convertInputs[0] = newTranspose;
         auto newConvert = convert->clone_with_new_inputs(convertInputs);
+        newConvert->set_friendly_name(transpose->get_friendly_name());
         ngraph::replace_node(transpose, newConvert);
         return true;
     };


src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/swap_convert_transpose.cpp
```diff
@@ -32,11 +32,12 @@ ov::intel_cpu::SwapConvertTranspose::SwapConvertTranspose() {
         transposeInputs[0] = convert->input_value(0);
         auto newTranspose = transpose->clone_with_new_inputs(transposeInputs);
         ngraph::copy_runtime_info(transpose, newTranspose);
-        newTranspose->set_friendly_name(transpose->get_friendly_name());
+        newTranspose->set_friendly_name(convert->get_friendly_name());
```
rkazants, line 35: why should we use `convert` name if outputs of `convert` and new `transpose` are different?

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/swap_convert_transpose.cpp
```diff
@@ -32,11 +32,12 @@ ov::intel_cpu::SwapConvertTranspose::SwapConvertTranspose() {
...
 
         ngraph::OutputVector convertInputs = convert->input_values();
         convertInputs[0] = newTranspose;
         auto newConvert = convert->clone_with_new_inputs(convertInputs);
+        newConvert->set_friendly_name(transpose->get_friendly_name());
```
rkazants, line 40: can we have test for this transformation to check node names preserving?

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/swap_convert_transpose.cpp
a-sidorova: The transformation broke the output names on new API 2.0 for CPU Plugin. Let's take a look at the simple test `./ov_cpu_func_tests --gtest_filter=smoke_Transpose2D/TransposeLayerTest.CompareWithRefs/IS=\(2.10\)_inputOrder=\(\)_netPRC=I8_inPRC=UNSPECIFIED_outPRC=UNSPECIFIED_inL=ANY_outL=ANY_trgDev=CPU*`
We have the following transformation pipeline (`Convert` is inserted after new preprocessing):
image
The model expects output with the name `Transpose_3` but after the pass `SwapConvertTranspose` the model has output node with name `Convert_1696`. To avoid these cases we should preserve the output name after transformation.

Thank you for the question! 🙂

a-sidorova:
@mateusztabaka I have already the fix https://github.com/openvinotoolkit/openvino/pull/18208 with one approve

PR #18341, opened 2023-07-03
[PT FE] Add aten::scatter and inplace for aten::sub translation

mmikolajcz:
### Details:
 - *Add translations for `aten::scatter` and `aten::scatter_`*
 - *Allow for `aten::sub_` inplace translation*
 - *Add test cases*

### Tickets:
 - *110807*


src/frontends/pytorch/src/op/scatter.cpp
```diff
@@ -0,0 +1,61 @@
...
+using namespace ov::op;
+
+OutputVector translate_scatter(const NodeContext& context) {
+    // Out-of-place schema
+    // "aten::scatter.src(Tensor self, int dim, Tensor index, Tensor src) -> Tensor"
+    // "aten::scatter.value(Tensor self, int dim, Tensor index, Scalar value) -> Tensor"
+
+    // Inplace schema
+    // "aten::scatter_.src(Tensor(a!) self, int dim, Tensor index, Tensor src) -> Tensor(a!)"
+    // "aten::scatter_.value(Tensor(a!) self, int dim, Tensor index, Scalar value) -> Tensor(a!)"
```
eaidova, line 27: it is not full list of schemas, there is any plans to support reduction?
 
 aten::scatter_.src(Tensor(a!) self, int dim, Tensor index, Tensor src) -> Tensor(a!):
  aten::scatter_.value(Tensor(a!) self, int dim, Tensor index, Scalar value) -> Tensor(a!):
  aten::scatter_.reduce(Tensor(a!) self, int dim, Tensor index, Tensor src, *, str reduce) -> Tensor(a!):
  aten::scatter_.value_reduce(Tensor(a!) self, int dim, Tensor index, Scalar value, *, str reduce) -> Tensor(a!):

mmikolajcz: Yes, sorry, I've missed those schemas and though that all reductions are handled by separate aten::scatter_reduce op. I've already verified that reduce is working, just need to fix few cases that are still failing and do some cleanup in tests, it should be ready soon

mmikolajcz: Added support for reduce schema, however, looking on all schemas there might still be few unsupported (attributes dimname, out) and some tests needed to be skipped due to issues with prim::Constant/prim::GetAttr

eaidova (approved):

src/frontends/pytorch/src/op/scatter.cpp
eaidova: dim name is rarely used and as I do not see it in any our models, so it is ok

src/frontends/pytorch/src/op/scatter.cpp
```diff
@@ -0,0 +1,77 @@
...
+    // aten::scatter_.value_reduce(Tensor(a!) self, int dim, Tensor index, Scalar value, *, str reduce) -> Tensor(a!):
+    num_inputs_check(context, 4, 5);
+    auto input = context.get_input(0);
+    auto dim = context.get_input(1);
+    auto index = context.mark_node(std::make_shared<v0::Convert>(context.get_input(2), element::i32));
+    auto src = context.get_input(3);
+
+    auto reduction = v12::ScatterElementsUpdate::Reduction::NONE;
+    auto input_num = context.get_input_size();
+    if (input_num > 4 && !context.input_is_none(input_num - 1)) {
```
mvafin, line 40:
```suggestion
    if (!context.input_is_none(input_num - 1)) {
```
Don't really need to check number of inputs that is verified inside `input_is_none`

mvafin (approved):
Non-critical comment

src/frontends/pytorch/src/op/scatter.cpp
mmikolajcz: We need to check if schema contains reduce parameter, without this check `input_num > 4`, in cases without reduce param, last input would be src and it would fail due to trying to cast it to string.
As for setting reduce input index directly instead of relying on last element, it seem that index depends on input combination, couldn't figure out exact rules how it depends on it, last element was easiest simplest solution

PR #18344, opened 2023-07-03
[Pad-12] Negative pads_begin/end tests for Pad reference implementation

mitruska:
### Details:
 - Tests with negative values in pads_begin and pads_end inputs for Pad op reference implementation 
 (common for Pad-1 and Pad-12)
 
*Negative values for pads_begin/end have been already supported by Pad-1 reference implementation, so the code and tests are common for both Pad-1 and Pad-12.

### Tickets:
 - 110242


src/plugins/template/tests/functional/op_reference/pad.cpp
```diff
@@ -1003,6 +1003,219 @@ std::vector<PadParams> generateParams() {
...
+        PadParams(
+            reference_tests::Tensor(ET, {3, 4}, std::vector<T>{
+                1, 2, 3, 4,
+                5, 6, 7, 8,
+                9, 10, 11, 12
+            }),
+            reference_tests::Tensor(ET_INT, {2}, std::vector<T_INT>{-1, 0}),
+            reference_tests::Tensor(ET_INT, {2}, std::vector<T_INT>{0, -2}),
+            reference_tests::Tensor(ET, {2, 2}, std::vector<T>{
+                5, 6, 9, 10
```
a-sidorova, line 1078: Minor
```suggestion
                5, 6,
                9, 10
```
And below

a-sidorova (approved):

src/plugins/template/tests/functional/op_reference/pad.cpp
mitruska: Sure, done:)

src/plugins/template/tests/functional/op_reference/pad.cpp
```diff
@@ -1003,6 +1004,223 @@ std::vector<PadParams> generateParams() {
...
+                9, 10, 11, 12
+            }),
+            reference_tests::Tensor(ET_INT, {2}, std::vector<T_INT>{2, 3}),
+            reference_tests::Tensor(ET_INT, {2}, std::vector<T_INT>{-1, -2}),
+            reference_tests::Tensor(ET, {4, 5}, std::vector<T>{
+                0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0,
+                0, 0, 0, 1, 2,
+                0, 0, 0, 5, 6,
+                0, 0, 0, 9, 10,
```
t-jankowski, line 1115: It 5x5 tensor, not 4x5. Remove `0,0,0,9,10` what is dropped. Did this test passed?

t-jankowski (changes requested):
1. Add a test with negative padding on 'middle' axis, e.g. [0, -1, 0]
2. What about such case: data 2x2, paddings [-1, 0] [-1, 0] ?

mitruska:
> 1. Add a test with negative padding on 'middle' axis, e.g. [0, -1, 0]

Done
> 2. What about such case: data 2x2, paddings [-1, 0] [-1, 0] ?

The result will be an empty tensor of shape [0, 2], test added.

t-jankowski (approved):

PR #18355, opened 2023-07-04
[CI] Introduce Linux pipeline in GitHub Actions

akashchi:
### Details:
This PR introduces the Linux pipeline that is executed in the GitHub Actions environment. It is similar to the Linux Azure pipeline in terms of stages: OV Build, Install; C++ tests, Python Tests, CPU Functional tests.

This pipeline utilizes the GitHub-provided runners both for building and testing. Also, it uses the GitHub Actions cache to speed up the build stage of the pipeline.

### Tickets:
 - *113015*


tools/mo/unit_tests/mo/frontend_ngraph_test.py
```diff
@@ -33,7 +33,7 @@ def test_frontends():
     status = subprocess.run(args, env=os.environ)
     assert not status.returncode
 
-
+@pytest.mark.skip(reason="Ticket - 113358")
```
rkazants, line 36: What is the exact problem with it? Let us discuss offline before skipping.

rkazants (changes requested):
Solid number of skipped tests and I worry about it

tools/mo/unit_tests/mo/frontend_ngraph_test.py
rkazants: @pavel-esir, did you have a chance to look at this? 
@akashchi, did you build moc_frontend for these tests and register a path to it in `LD_LIBRARY`?

tests/layer_tests/tensorflow_tests/test_tf_LRN.py
```diff
@@ -27,6 +27,7 @@ def create_lrn_net(self, input_shape, depth_radius, bias, alpha, beta):
     @pytest.mark.parametrize("params", test_data_basic)
     @pytest.mark.precommit_tf_fe
     @pytest.mark.nightly
+    @pytest.mark.skip(reason="Sporadically failed. Need further investigation. Ticket - 113362")
```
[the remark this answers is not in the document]
akashchi: I have introduced a separate mark (`@pytest.mark.skipif(os.getenv("GITHUB_ACTIONS"))`) as suggested using the default `GITHUB_ACTIONS` env variable as per https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables.

tests/layer_tests/tensorflow_tests/test_tf_SplitV.py
```diff
@@ -33,6 +33,7 @@ def create_splitv_net(self, value_shape, size_splits_values, axis_value):
     @pytest.mark.parametrize("params", test_data_basic)
     @pytest.mark.precommit_tf_fe
     @pytest.mark.nightly
+    @pytest.mark.skip(reason="Ticket - 113359")
```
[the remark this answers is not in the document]
akashchi: The ticket is currently assigned to Golov, Mikhail. I suggest referring to him.

tools/mo/unit_tests/mo/frontend_ngraph_test.py
akashchi: @pavel-esir highlighted that these tests are skipped in Azure due to the absence of the mock_api:
image

[Example pipeline](https://dev.azure.com/openvinoci/dldt/_build/results?buildId=590660&view=logs&j=982848f2-7f8d-5e57-730a-8df1e5f991ba&t=640d3dcd-a1dd-5e6b-6d22-d23b2eb859ae)

I am unsure why they are executed in the GH Actions pipeline as I was copying the build commands from the Azure pipeline.

akashchi: @rkazants could you please link me to the example command used for building `moc_frontend` and registering its path in `LD_LIBRARY`?

pavel-esir: I think @akashchi was right that skipped them. @rkazants i think they can be definitely skipped at the moment. They cannot be patched quickly, they need to be rewritten.

>pavel-esir, did you have a chance to look at this?

yes, i did look at them. And in Andrey's environment openvino_mock_mo_frontend was successfully build and placed correctly inside `bin/intel64/Release/python`(without version) and for this very reason they were not skipped. 

But in the Azure they are placed wrongly somewhere else and since `openvino_mock_mo_frontend` was not found they are silently skipped:

https://github.com/openvinotoolkit/openvino/blob/5d399faa642cfafa6ee995bf689ad2dfb79bd7e3/tools/mo/unit_tests/mo/moc_frontend/moc_extractor_test_actual.py#L14-L26


If i build all `pybind_mock_frontend`, `openvino_mock_mo_frontend `, `mock_mo_python_api` et all. and look at the messages of the exact failed tests `test_frontends` (which calls `frontend_ngraph_test_actual`) and `test_moc_extractor` (which calls `moc_extractor_test_actual`) they are very outdated. Eg. they are called with 3 arguments including `fem`
https://github.com/openvinotoolkit/openvino/blob/0c9abf43a91fbe2dc95974b6cb2651bd883d1b12/tools/mo/unit_tests/mo/frontend_ngraph_test_actual.py#L98
but the current tools.mo.main has only 2 input arguments https://github.com/openvinotoolkit/openvino/blob/5d399faa642cfafa6ee995bf689ad2dfb79bd7e3/tools/mo/openvino/tools/mo/main.py#L19

even if removed redundant `fem` then other error occurs, mock frontends changed their names
image

Since they actually have been implicitly skipped for a long time they have become a little bit stale. Therefore if we want to keep this tests we need some time to fix/rewrite them.

rkazants (approved):
Looks good to me from MO and TF FE perspective

src/bindings/python/src/pyopenvino/CMakeLists.txt
```diff
@@ -114,9 +114,7 @@ if(OpenVINO_SOURCE_DIR OR OpenVINODeveloperPackage_FOUND)
     install(DIRECTORY ${OpenVINOPython_SOURCE_DIR}/src/openvino
             DESTINATION ${OV_CPACK_PYTHONDIR}
             COMPONENT ${OV_CPACK_COMP_PYTHON_OPENVINO}_${pyversion}
-            ${OV_CPACK_COMP_PYTHON_OPENVINO_EXCLUDE_ALL}
-            USE_SOURCE_PERMISSIONS
-            PATTERN "test_utils" EXCLUDE)
+            ${OV_CPACK_COMP_PYTHON_OPENVINO_EXCLUDE_ALL})
```
ilya-lavrenov, line 117: looks like after this change, tests code will be a part of the product. We should avoid it

.github/workflows/linux.yml
```diff
@@ -0,0 +1,714 @@
...
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install python dependencies
+        run: |
+          # For Python API
+          python3 -m pip install --upgrade pip
+          python3 -m pip install -r ${{ env.OPENVINO_REPO }}/src/bindings/python/wheel/requirements-dev.txt
```
akladiev, line 487: Ideally, each type of tests (and builds) that have their own requirement files should use their own Python virtual environment as well (this was a problem for Azure recently - MO MXNet unit tests required older NumPy version than other frameworks, while in Azure we're installing deps for all frameworks to the same place (system), causing dependency overwriting - now we're testing all frameworks on an older NumPy, instead of just MXNet, that is wrong).

If the goal is to achieve feature-parity with Azure - we can fix that later.

src/bindings/python/src/pyopenvino/CMakeLists.txt
akashchi: Seems like it was a debug change, I've removed it, [the pipeline is passing](https://github.com/openvinotoolkit/openvino/actions/runs/5505603283).

ababushk (approved):

PR #18379, opened 2023-07-05
Resolve tensor sharing issue when there are same name output port name

riverlijunjie:
### Details:
 - In some case, model has 2 or more same name input/output ports, they aslo have the same precision and partial_shape. Compiled_model will share the same ov::Descriptor::Tensor pointer and ov::Tensor between multiple such ports.
- Considered solving python segment fault issue to create separated input/output ports, which also need handle such tensor shared case, this patch will do it.

### Tickets:
 - *ticket-id*


src/core/include/openvino/core/node_output.hpp
```diff
@@ -64,6 +64,8 @@ class OPENVINO_API Output<Node> {
     descriptor::Tensor& get_tensor() const;
     /// \return A shared point to the tensor ptr for this output.
     std::shared_ptr<descriptor::Tensor> get_tensor_ptr() const;
+    /// \return Set new tensor shared pointer to this output
```
songbell, line 67:
```suggestion
    /// \return Set new tensor desc pointer to this output
```

src/inference/src/dev/icompiled_model.cpp
```diff
@@ -48,6 +48,20 @@ ov::ICompiledModel::ICompiledModel(const std::shared_ptr<const ov::Model>& model
             }
         }
 
+        auto find_tensor_desc =
```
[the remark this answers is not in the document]
riverlijunjie: Good idea!

src/inference/src/dev/icompiled_model.cpp
```diff
@@ -64,6 +65,18 @@ ov::ICompiledModel::ICompiledModel(const std::shared_ptr<const ov::Model>& model
...
+            if (tensor_map.count(old_tensor)) {
+                new_param->output(0).set_tensor_ptr(tensor_map[old_tensor]);
+            } else {
+                auto tensor_desc =
+                    std::make_shared<ov::descriptor::Tensor>(param->get_output_element_type(0),
+                                                             param->get_output_partial_shape(0),
+                                                             new_param->output(0).get_tensor().get_names());
+                tensor_desc->clone_from(param->output(0).get_tensor());
+                new_param->output(0).set_tensor_ptr(tensor_desc);
+                tensor_map[old_tensor] = tensor_desc;
```
ilyachur, line 78: Why do you need set_tensor here?
I thought that you can simple call:
```
tensor_map[old_tensor] = new_param->output(0).get_tensor_ptr();
```

riverlijunjie: sure, it is reasonable, let me try whether there are any test failures.

songbell (approved):

ilyachur (approved):

PR #18391, opened 2023-07-05
[GNA] Added Reshape support to TS patterns

mryzhov:
### Details:
 - Corrected patterns in TS sinking to support Reshapes between Transpose and Split/Concat

### Tickets:
 - *114277*

### DIFF
--- a/src/plugins/intel_gna/src/common/graph_utils.hpp
+++ b/src/plugins/intel_gna/src/common/graph_utils.hpp
@@ -81,6 +81,13 @@ inline bool get_constant_value(const std::shared_ptr<ngraph::opset8::Constant>& 
     return true;
 }
 
+/**
+ * @brief Checks if 2 shapes are the same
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gna/src/common/graph_utils.hpp
```diff
@@ -679,6 +679,13 @@ inline bool has_n_consumers(const std::shared_ptr<ov::Node>& node, size_t n_cons
     return node->output(0).get_target_inputs().size() == n_consumers;
 }
 
+/**
+ * @brief Checks if 2 shapes are the same
+ */
+inline bool are_shapes_equal(const ov::Shape& shape_1, const ov::Shape& shape_2) {
```
sirzabek, line 685: A suggestion to look for places when it can be readily used and use it.

sirzabek (approved):

PR #18393, opened 2023-07-05
[GPU] Fix issues in runtime buffer fusing for oneDNN concat and auto padding support for conv

andrew-k-park:
### Details:
 - Fix to let concat onednn impl check `can_be_optimized` in impl_param instead of that in node
 - Add `auto_pad` attribute for creating conv primitive and apply auto padding to kernel param for conv ocl impl

### Tickets:
 - 114379


src/plugins/intel_gpu/src/graph/graph_optimizer/compile_graph.cpp
```diff
@@ -58,6 +58,11 @@ void compile_graph::run(program& p) {
             can_select_impl = false;
         }
 
+        // TODO: Remove this WA once we have shape agnostic conv kernl with specified auto_pad attributes
+        if (node->is_type<convolution>() && node->is_dynamic() && !node->as<convolution>().use_explicit_padding()) {
+            can_select_impl = false;
+        }
```
yeonbok, line 64: for deconv, binary_convolution, too, right?

yeonbok (approved):

src/plugins/intel_gpu/src/graph/graph_optimizer/compile_graph.cpp
andrew-k-park: Currently, only convolution specifies auto_pad attribute from ngraph op. deconvolution or binary_convolution is always explicit padding.

PR #18399, opened 2023-07-05
Fix the legacy zero point issue.

luweizhou2016:
### Details:
 - The issue should has been introduced when migrating to ONEDNN 3.1.  The PR#17082 in OV changes the post ops pattern and issue appeared. Fork legacy input zero point compensation need to update conflict register code flow as signed_input mechanism. 

    const Xbyak::Reg64 reg_bias_data = r12;
    const Xbyak::Reg64 reg_comp_data = r12;
    const Xbyak::Reg64 reg_ptr_dst_scale = r12;

### Tickets:
 - *CVS-114070*


luweizhou2016:
Forked onednn PR:https://github.com/openvinotoolkit/oneDNN/pull/203

dmitry-gorokhov (approved):

dmitry-gorokhov:
@luweizhou2016  The fix itself LGTM. 
Hopefully we can extend exisitng [quantized Conv with zero-points test](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/tests/functional/subgraph_tests/src/conv_with_zero_point_fuse.cpp) to cover such scenario. Can be done in follow-up PRs.

PR #18403, opened 2023-07-06
[Snippets][CPU] Enabled MHA tokenization for quant and bf16 cases

a-sidorova:
### Details:
 - *Enable Snippets tokenization on quantized models and float models with `inference_precision=bf16`*
 - *Disabled custom MHA Fusion pass*
 - *Added `Fill` op after `VectorBuffer` to fill the correct default value for accumulator pattern. Before `ReduceMax` had zeros in buffers but should have `-float_max`*
 - *Fixed buffer allocation size*
 - *Updated subgraph CPU tests for MHA*

### Tickets:
 - *112973, 105339*
 
 ### TODO:
- [x] *Add check for matmul shapes since blocking by k, n dimensions is supported only for f32*
- [x] *Performance validation results are attached to the ticket 112973*
- [x] *Accuracy validation results are attached to the ticket 112973*

### DIFF
--- a/src/common/snippets/src/lowered/pass/allocate_buffers.cpp
+++ b/src/common/snippets/src/lowered/pass/allocate_buffers.cpp
@@ -17,6 +17,7 @@ void AllocateBuffers::propagate_offset(const LinearIR& linear_ir, const Expressi
     // to correctly read and write data because all Buffers has the common data pointer on buffer scratchpad
 
     const auto buffer = ov::as_type_ptr<op::Buffer>(buffer_expr->get_node());
+    buffer->set_offset(static_cast<int64_t>(offset));
 
     // Propagate to up: in Store. Buffer can have only one Store
     {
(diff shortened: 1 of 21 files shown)


src/plugins/intel_cpu/src/emitters/x64/jit_snippets_emitters.cpp
```diff
@@ -1452,6 +1430,27 @@ void FillEmitter::emit_isa(const std::vector<size_t> &in, const std::vector<size
...
+template <typename Vmm>
+void FillEmitter::fill_full(const Vmm& dst_vmm) const {
+    // Optimized impl for zero
+    if (fill_value == int32_t(0x0)) {
+        h->uni_vpxor(dst_vmm, dst_vmm, dst_vmm);
+    } else {
+        Xmm xmm = Xmm(dst_vmm.getIdx());
+        h->mov(Reg64(aux_gpr_idxs[0]), fill_value);
+        h->vmovq(xmm, Reg64(aux_gpr_idxs[0]));
+        h->vbroadcastss(dst_vmm, xmm);
```
IvanNovoselov, line 1448: Broadcast can also read from memory, can't it?
If so, are you sure it's more efficient to perform `mov` + `vmovq` instead of reading from table as you do in `fill_tail`?

src/common/snippets/src/lowered/pass/allocate_buffers.cpp
```diff
@@ -57,34 +58,40 @@ void AllocateBuffers::propagate_offset(const LinearIR& linear_ir, const Expressi
...
 bool AllocateBuffers::run(LinearIR& linear_ir) {
     OV_ITT_SCOPED_TASK(ov::pass::itt::domains::SnippetsTransform, "Snippets::AllocateBuffers");
+    // [113664] The pass contains two main logics: it defines which of buffers can be inplace (use the same memory) and
+    // allocate memory of needed size. It should be splitted into several passes and updated in bounds of the ticket 113664.
 
-    size_t offset = 0;
+    size_t offset = 0, wsp_offset = 0;
+    size_t prev_data_size = 0, current_data_size = 0;
     std::set<ExpressionPtr> allocated_buffers;
+    bool wsp_allocated = false;
```
IvanNovoselov, line 67: Minor: could you please provide explanation on what `wsp` means? Please add a comment, or better provide a self-explanatory names to `wsp_offset` and `wsp_allocated`

IvanNovoselov (approved):

src/common/snippets/src/lowered/pass/allocate_buffers.cpp
a-sidorova: This name is used in `Brgemm` AMX case. I renamed the variable and added the comment with the ticket. I believe we will rewrite Buffer logic asap 🙂 
Thank you

a-sidorova:
@IvanNovoselov thank you for the review!

@dmitry-gorokhov could you take a look please as well? Thank you!

dmitry-gorokhov (approved):

PR #18405, opened 2023-07-06
fix get_profiling_info tests not run in ci

songbell:
### Details:
 - profiling tests not tested in CI

### Tickets:
 - *ticket-id*


songbell:
replaced by https://github.com/openvinotoolkit/openvino/pull/18408/files

PR #18406, opened 2023-07-06
[PyOV] Limit `numpy` version to 1.23 for MO mxnet

p-wysocki:
### Details:
 - the newest version of mxnet is 1.9.1: https://pypi.org/project/mxnet/1.9.1/#history
 - numpy==1.24.0 deprecated `np.bool` symbol: https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations
 - mxnet==1.9.1 uses it, causing errors in CI pipelines: https://github.com/apache/mxnet/blob/v1.9.1/python/mxnet/numpy/utils.py#L37
 - we can't upgrade mxnet, so we have to limit numpy version for MO mxnet

### Tickets:
 - N/A

### DIFF
--- a/tools/mo/requirements_mxnet.txt
+++ b/tools/mo/requirements_mxnet.txt
@@ -1,5 +1,5 @@
 -c ../constraints.txt
-numpy>=1.16.6,<1.26
+numpy>=1.16.6,<1.24
 mxnet
 networkx
 defusedxml


rkazants (approved):

akladiev:
As far as I understand, since the conflicting NumPy dependency originates from the newly released TensorFlow, and in its further releases we have a potential risk of conflicting with NNCF that doesn't support newer NumPy's, this PR is more like a workaround. @p-wysocki, @mlukasze, should we create a ticket to restrict TF version instead?

p-wysocki:
@akladiev I believe that the "clean" solution would be to have MO tested in separate Python environments, per framework. That way `mxnet` can be tested separately with its own, old dependency versions, without an impact on the rest of the product. I think that it's the correct choice over restricting TF version, which may negatively impact users' experience.

pavel-esir:
@p-wysocki thanks for the PR! I added ticket number into description because there was a ticket where this problems was reported

PR #18410, opened 2023-07-06
OVC tool leftovers

popovaan:
Description:

Duplicated OVC/MO common functionality
Added OVC Pylint test
Fixed IR version in OVC cli tool

Ticket: 114255

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update


.github/workflows/mo.yml
```diff
@@ -52,6 +52,10 @@ jobs:
...
         working-directory: tools/mo
 
-      - name: Pylint
+      - name: Pylint-MO
         run: pylint -d C,R,W openvino/tools/mo
         working-directory: tools/mo
+
+      - name: Pylint-OVC
+        run: pylint -d C,R,W openvino/tools/ovc
+        working-directory: tools/ovc
```
rkazants, line 61: can you please double-check this one because it seems to generate empty report for some reason?

rkazants (commented):
can you please add layer test running mo/ovc tools and mo/ovc `convert_model`?
I know some of them already exists. Do we have any gaps here?

Other looks good to me.

popovaan:
> can you please add layer test running mo/ovc tools and mo/ovc `convert_model`? I know some of them already exists. Do we have any gaps here?
> 
> Other looks good to me.

OVC tool test is here:
https://github.com/openvinotoolkit/openvino/blob/8f529df763f6bba38af85748d8d24f495bb673f1/tests/layer_tests/mo_python_api_tests/test_ovc_cli_tool.py#L76

convert_model() is imported from openvino.runtime for all layer tests which is the same as importing from tools.ovc:
https://github.com/openvinotoolkit/openvino/blob/8f529df763f6bba38af85748d8d24f495bb673f1/tests/layer_tests/common/utils/common_utils.py#L50C24-L50C24
https://github.com/openvinotoolkit/openvino/blob/8f529df763f6bba38af85748d8d24f495bb673f1/tests/layer_tests/common/layer_test_class.py#L62

convert_model() from tools.mo is used only for tests where use_legacy_frontend is used. So it looks like no gaps here.

.github/workflows/mo.yml
popovaan: Yes, I missed the config files for pylint, so the check was not complete. Added same configs as in MO.

rkazants (approved):

PR #18422, opened 2023-07-07
[GPU] Update DepthToSpace to use nGraph shape inference

andrew-k-park:
### Details:
 - Update `DepthToSpace` to use nGraph shape inference for requirement on dynamic model

### Tickets:
 - 100229

### DIFF
--- a/src/plugins/intel_gpu/src/graph/depth_to_space.cpp
+++ b/src/plugins/intel_gpu/src/graph/depth_to_space.cpp
@@ -3,6 +3,7 @@
 //
 
 #include "depth_to_space_inst.h"
+#include "depth_to_space_shape_inference.hpp"
 
 #include "primitive_type_base.h"
 #include "intel_gpu/runtime/error_handler.hpp"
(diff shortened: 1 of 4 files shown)


e-ddykim (approved):
Looks good to me

ahnyoung-paul (approved):

src/plugins/intel_gpu/src/graph/depth_to_space.cpp
```diff
@@ -20,11 +21,6 @@ layout depth_to_space_inst::calc_output_layout(depth_to_space_node const& node,
 
     const size_t block_size = desc->block_size;
 
-    if (block_size < 2)
-        CLDNN_ERROR_MESSAGE(desc->id,
-                            "Invalid depthToSpace block_size value (should equal at least two). Actual block size is" +
-                                std::to_string(block_size));
-
```
yeonbok, line 27: Why can we remove this assertion?

yeonbok: Hmm anyway the spec seems not have such a limitation..

andrew-k-park: Yes, This assertion seems to be a legacy limitation. According to the spec, the default block_size is 1, and I was able to test that case(block_size=1) when I removed this assertion.

PR #18428, opened 2023-07-07
temporarily disable to unblock CI

songbell:
### Details:
 - temporaribly disable to unblock CI, debug WIP

### Tickets:
 - [106625]


songbell:
close, fix is on the way

PR #18461, opened 2023-07-10
String tensors in tf.Graph decoder

popovaan:
Description: 
Implemented passing of string tensors to TF FE using vector of string.
Added test.

Ticket: 113210

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests

Documentation:


tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
```diff
@@ -587,6 +587,27 @@ def __call__(self, input1):
     return model, model_ref, {'example_input': example_input}
 
 
+def create_keras_layer_with_string_tensor(tmp_dir):
+    import tensorflow as tf
+
+    x = tf.keras.Input(shape=[1], name="input", dtype=tf.dtypes.int64)
```
rkazants, line 593:
```suggestion
    x = tf.keras.Input(shape=[1], name="input", dtype=tf.int64)
```

tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
```diff
@@ -587,6 +587,27 @@ def __call__(self, input1):
...
+    var = tf.Variable("Text_1")
+    const = tf.constant("Text_2")
```
rkazants, line 595: please set dtype explicitly

tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
```diff
@@ -587,6 +587,27 @@ def __call__(self, input1):
...
+    hash = tf.strings.to_hash_bucket([const, var], 3)
+
+    model = tf.keras.Model(inputs=[x], outputs=[hash + x])
+    example_input = tf.constant(3)
```
rkazants, line 599: it seems that `example_input` is not needed here.

tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
```diff
@@ -587,6 +587,27 @@ def __call__(self, input1):
...
+
+    param = ov.opset8.parameter([], dtype=np.int32)
+    conv = ov.opset8.convert(param, np.int64)
+    const = ov.opset8.constant([0, 2], dtype=np.int64)
```
rkazants, line 603: I am not sure that we test the required logic. I propose to create a graph with two sub-graphs: 1. Add(param1_float, param2_float) 2. Const(string)

We have transformation removing Const->Result this way the resulted graph will be with Add operation and two flow params

src/bindings/python/src/openvino/frontend/tensorflow/node_decoder.py
```diff
@@ -44,26 +46,47 @@ def tf_attr_to_ov(attr):
     return OVAny(tf_attr_to_numpy(attr))
 
 
+def try_decode(line: bytes):
+    assert isinstance(line, bytes), "Expected bytes, got object of type {}".format(type(line))
+    try:
+        return True, line.decode()
+    except ValueError:
+        return False, line
```
rkazants, line 54: why do we need this logic? why can't we decode each string tensor element more universally?

popovaan: This logic is needed to decode bytes to string. Bytes in general can fail to decode, if it is not a string or there is some unknown encoding. For this reason try-catch is used.
If we fail to decode it, then list of bytes objects will be returned from decoder.

What do mean by decoding each string tensor element more universally?

tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
popovaan: Done.

tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
popovaan: Applied.

tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
popovaan: Added test with such graph.

src/bindings/python/src/openvino/frontend/tensorflow/node_decoder.py
popovaan: In offline discussion it was decided to replace `decode()` with `str()`. Done.

src/bindings/python/src/openvino/frontend/tensorflow/node_decoder.py
```diff
@@ -44,26 +46,45 @@ def tf_attr_to_ov(attr):
...
```
rkazants, line 54: agreed to remove it and do not change bytestream

rkazants (approved):
LGTM, please remove decoding for elements of string tensor

src/bindings/python/src/openvino/frontend/tensorflow/node_decoder.py
popovaan: Removed.

PR #18463, opened 2023-07-10
Fix -api sync for single -data_shape

Wovchena:
Tickets 111187 and 111185

I wasn’t able to find C++ equivalent of Python’s `info.original_shape.is_static`. Later I realized that it shouldn’t be considered because -shape cmd arg should have higher priority for shape inference than model’s shape. So I removed it from Python.

Replace

`if allow_inference_only_or_sync and batch_size.is_dynamic:`

with

`if allow_inference_only_or_sync and batch_size.is_dynamic:`

to reset batch_size to static in case of dynamic shape with single -data_shape


samples/cpp/benchmark_app/utils.cpp
```diff
@@ -108,6 +108,17 @@ std::vector<float> split_float(const std::string& s, char delim) {
     return result;
 }
 
+bool can_measure_as_static(const std::vector<benchmark_app::InputsInfo>& app_input_info) {
+    for (const benchmark_app::InputsInfo& info : app_input_info) {
+        for (const auto& pair : info) {
+            if (pair.second.partialShape.is_dynamic() && app_input_info.size() > 1) {
```
nkogteva, line 114: Please correct me, if I'm wrong: we can measure dynamic model as static only when 1 group of shapes are given (--data_shape parameter). But how shape can be dynamic? We have to check it first during command line arguments parsing.

nkogteva: > I think I don't understand. `partialShape` value does come from command line arguments parsing:
> 
> https://github.com/openvinotoolkit/openvino/blob/7568a494c8b1d995fad77156fcdcd2aeec726afb/samples/cpp/benchmark_app/utils.cpp#L609

In this case I also don't understand what you are trying to check here. This function is already called for dynamic model. What is the condition to execute dynamic model in sync mode?

Wovchena: If there's partial shape and -data_shape sets more than one shape the model can't be used for sync. Everything else is fine for sync

Wovchena: This came form Python. Your idea of only app_input_info.size() == 1 is probably a sufficient condition. I will check that

nkogteva (approved):

PR #18464, opened 2023-07-10
[PyOV] Extend Tensor API

akuporos:
### Details:
 - support `copy_to` and ctors from output port

### Tickets:
 - 104225


src/bindings/python/src/pyopenvino/core/tensor.cpp
```diff
@@ -133,6 +133,14 @@ void regclass_Tensor(py::module m) {
             py::arg("type"),
             py::arg("shape"));
 
+    cls.def(py::init<ov::Output<ov::Node>>(), py::arg("node_port"));
+
+    cls.def(py::init([](ov::Output<ov::Node>& port, py::array& array) {
+                return ov::Tensor(port, const_cast<void*>(array.data(0)));
+            }),
+            py::arg("node_port"),
+            py::arg("array"));
```
jiwaszki, line 142:
```suggestion
    cls.def(py::init<const ov::Output<const ov::Node>>(), py::arg("node_port"));

    cls.def(py::init([](const ov::Output<const ov::Node>& port, py::array& array) {
                return ov::Tensor(port, const_cast<void*>(array.data(0)));
            }),
            py::arg("node_port"),
            py::arg("array"));
```

There is a difference with classes `Output<const Node>` and `Output<Node>`. The outside qualifier is less (but still) important. However, there is significant difference with `<const Node>`, it comes usually from `CompiledModel`. The second `Ouput<Node>` comes usually from `Model`.

So according to:
https://github.com/openvinotoolkit/openvino/blob/3330cd69a7496ede1a55ec3feee649b8b8c8687c/src/core/include/openvino/core/node_output.hpp#L118
https://github.com/openvinotoolkit/openvino/blob/3330cd69a7496ede1a55ec3feee649b8b8c8687c/src/core/include/openvino/runtime/tensor.hpp#L146
https://github.com/openvinotoolkit/openvino/blob/3330cd69a7496ede1a55ec3feee649b8b8c8687c/src/core/include/openvino/runtime/tensor.hpp#L156

I believe that both constructors need ports from compiled models to work with. So this is required change for the constructors.

src/bindings/python/tests/test_runtime/test_tensor.py
```diff
@@ -148,6 +149,25 @@ def test_init_with_numpy_copy_memory(ov_type, numpy_dtype):
...
+    tensor3 = Tensor(param3.output(0))
+    tensor4 = Tensor(param3.output(0), ones_arr)
+    assert tensor1.shape == param1.shape
+    assert tensor1.element_type == param1.get_element_type()
+    assert tensor2.shape == param2.shape
+    assert tensor2.element_type == param2.get_element_type()
+    assert tensor3.shape == ov.Shape([0])
+    assert tensor3.element_type == param3.get_element_type()
+    assert tensor4.shape == ov.Shape([0])
+    assert tensor4.element_type == param3.get_element_type()
```
jiwaszki, line 168: According to change with `<const Node>`, this tests should be rewritten as well.

src/bindings/python/src/pyopenvino/core/tensor.cpp
```diff
@@ -247,6 +255,25 @@ void regclass_Tensor(py::module m) {
...
+        )");
+
+    cls.def(
+        "copy_to",
+        [](ov::Tensor& self, ov::Tensor& dst) {
+            return self.copy_to(dst);
+        },
+        py::arg("target_tensor"),
+        R"(
+        Copy tensor, destination tensor should have the same element type and shape
```
jiwaszki, line 274:
```suggestion
        Copy tensor's data to a destination tensor. The destination tensor should have the same element type and shape.
```

src/bindings/python/src/pyopenvino/core/tensor.cpp
```diff
@@ -247,6 +255,25 @@ void regclass_Tensor(py::module m) {
             Sets Tensor's shape.
         )");
 
+    cls.def("is_continuous",
+            &ov::Tensor::is_continuous,
+            R"(
+            Reports whether the tensor is continuous or not.
+
+            :return: true if tensor is continuous
```
jiwaszki, line 263:
```suggestion
            :return: True if the tensor is continuous, otherwise False.
```

src/bindings/python/src/pyopenvino/core/tensor.cpp
```diff
@@ -247,6 +255,25 @@ void regclass_Tensor(py::module m) {
             Sets Tensor's shape.
         )");
 
+    cls.def("is_continuous",
+            &ov::Tensor::is_continuous,
```
jiwaszki, line 259: Are there any tests for it?

src/bindings/python/src/pyopenvino/core/tensor.cpp
```diff
@@ -133,6 +133,14 @@ void regclass_Tensor(py::module m) {
             py::arg("type"),
             py::arg("shape"));
 
+    cls.def(py::init<ov::Output<ov::Node>>(), py::arg("node_port"));
+
+    cls.def(py::init([](ov::Output<ov::Node>& port, py::array& array) {
+                return ov::Tensor(port, const_cast<void*>(array.data(0)));
+            }),
+            py::arg("node_port"),
```
jiwaszki, line 141:
```suggestion
    cls.def(py::init<ov::Output<ov::Node>>(), py::arg("port"));

    cls.def(py::init([](ov::Output<ov::Node>& port, py::array& array) {
                return ov::Tensor(port, const_cast<void*>(array.data(0)));
            }),
            py::arg("port"),
```
Now I feel like keeping port would be OK... let's align it.

src/bindings/python/src/pyopenvino/core/tensor.cpp
```diff
@@ -133,6 +133,14 @@ void regclass_Tensor(py::module m) {
             py::arg("type"),
             py::arg("shape"));
 
+    cls.def(py::init<ov::Output<ov::Node>>(), py::arg("node_port"));
+
+    cls.def(py::init([](ov::Output<ov::Node>& port, py::array& array) {
+                return ov::Tensor(port, const_cast<void*>(array.data(0)));
```
jiwaszki, line 139: Both functions missing documentation strings.

jiwaszki, line 139: I have some concerns about this. As previous overload is OK, just creates empty Tensor with given element type and shape. This one might introduce issues with memory.

What happens if passed array is of shape `[2, 2]` let's say `[[1, 2], [3, 4]]` and port has a shape `[2]`? Is result just `[1, 2]`?
So what in the case of port of shape `[5]`? Is the result `[1, 2, 3, 4, 0.000412345154151]`?

What happens if port has mismatched type with array? Let's assume port is `ov.Type.i32` and array is `np.int8`. Is it going to reinterpret data as `int32`? Or is it going to "pack" data into `int32`?

All of this needs to be clarified, tested and documented.

As for shapes there can be a warning that shapes does not match and data will be cropped/extended. (personally I would like a warning about cropping and error about extended as it goes out-of-bounds)

Also warnings about mismatching types would be great. Maybe a combination of both when overall byte size is not matching?

CC @p-wysocki what do you think?

jiwaszki (changes requested):

src/bindings/python/src/pyopenvino/core/tensor.cpp
jiwaszki: After discussion, let's keep both const and non-const constructors.

src/bindings/python/src/pyopenvino/core/tensor.cpp
jiwaszki: Let's remove this function and add it if necessary (request is coming).

src/bindings/python/src/pyopenvino/core/tensor.cpp
akuporos: removed method as discussed

src/bindings/python/src/pyopenvino/core/tensor.cpp
akuporos: removed

src/bindings/python/src/pyopenvino/core/tensor.cpp
akuporos: updated

src/bindings/python/src/pyopenvino/core/tensor.cpp
akuporos: updated

src/bindings/python/src/pyopenvino/core/tensor.cpp
akuporos: added

src/bindings/python/src/pyopenvino/core/tensor.cpp
akuporos: added constructors with const-output argument

src/bindings/python/tests/test_runtime/test_tensor.py
akuporos: added one more test for const-output

src/bindings/python/src/pyopenvino/core/tensor.cpp
akuporos: added warnings for mismatched elements and shape.

src/bindings/python/src/pyopenvino/core/common.cpp
```diff
@@ -227,6 +227,31 @@ ov::Tensor tensor_from_pointer(py::array& array, const ov::Shape& shape, const o
...
+    }
+
+    if (array_helpers::is_contiguous(array)) {
+        if (port_shape_size > array_shape_size) {
+            OPENVINO_THROW("Shape of the port exceeds shape of the array.");
+        }
+        if (port_shape_size < array_shape_size) {
+            PyErr_WarnEx(PyExc_RuntimeWarning,
+                         "Shape of the port less that shape of the array. Passing data will be cropped.",
+                         1);
```
jiwaszki, line 247:
```suggestion
            PyErr_WarnEx(PyExc_RuntimeWarning,
                         "Shape of the port is smaller than shape of the array. Passed data will be cropped.",
                         1);
```

src/bindings/python/tests/test_runtime/test_tensor.py
```diff
@@ -148,6 +149,62 @@ def test_init_with_numpy_copy_memory(ov_type, numpy_dtype):
...
+    with pytest.raises(RuntimeError) as e:
+        Tensor(param2.output(0), ones_arr)
+    assert "Shape of the port exceeds shape of the array." in str(e.value)
+
+
+def test_init_with_output_port_different_types():
+    param1 = ops.parameter(ov.Shape([2, 2]), dtype=np.int8)
+    data = np.random.rand(2, 2).astype(np.float32)
+    with pytest.warns(RuntimeWarning):
+        Tensor(param1.output(0), data)
```
jiwaszki, line 205: A behavior of casting could be tested here.

src/bindings/python/tests/test_runtime/test_tensor.py
```diff
@@ -148,6 +149,62 @@ def test_init_with_numpy_copy_memory(ov_type, numpy_dtype):
...
+def test_init_with_node_output_port():
+    param1 = ops.parameter(ov.Shape([1, 3, 2, 2]), dtype=np.float64)
+    param2 = ops.parameter(ov.Shape([1, 3, 32, 32]), dtype=np.float64)
+    param3 = ops.parameter(ov.PartialShape.dynamic(), dtype=np.float64)
+    ones_arr = np.ones(shape=(1, 3, 32, 32), dtype=np.float64)
+    tensor1 = Tensor(param1.output(0))
+    tensor2 = Tensor(param2.output(0), ones_arr)
+    tensor3 = Tensor(param3.output(0))
+    with pytest.warns(RuntimeWarning):
+        tensor4 = Tensor(param3.output(0), ones_arr)
```
jiwaszki, line 161: Should dynamic shapes be thrown as an warning? If the shape is dynamic, it does not really matter what is an array's shape. Can the check be adjusted to not throw warning when the shape is dynamic?

src/bindings/python/src/pyopenvino/core/tensor.cpp
```diff
@@ -242,6 +300,16 @@ void regclass_Tensor(py::module m) {
             Sets Tensor's shape.
         )");
 
+    cls.def(
```
ilyachur, line 303: Please also add `is_continuous()` method

src/bindings/python/src/pyopenvino/core/common.cpp
```diff
@@ -227,6 +227,31 @@ ov::Tensor tensor_from_pointer(py::array& array, const ov::Shape& shape, const o
...
+        }
+        return ov::Tensor(port, const_cast<void*>(array.data(0)));
+    }
```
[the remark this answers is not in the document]
akuporos: done

src/bindings/python/src/pyopenvino/core/common.cpp
```diff
@@ -227,6 +227,31 @@ ov::Tensor tensor_from_pointer(py::array& array, const ov::Shape& shape, const o
...
 }
 
+ov::Tensor tensor_from_pointer(py::array& array, const ov::Output<const ov::Node>& port) {
+    auto array_type = array_helpers::get_ov_type(array);
+    auto array_shape_size = ov::shape_size(array_helpers::get_shape(array));
+    auto port_element_type = port.get_element_type();
+    auto port_shape_size = ov::shape_size(port.get_partial_shape().is_dynamic() ? ov::Shape{0} : port.get_shape());
+
+    if (array_type != port_element_type) {
+        PyErr_WarnEx(PyExc_RuntimeWarning, "Type of the array and the port are mismatched.", 1);
```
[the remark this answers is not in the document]
akuporos: updated

src/bindings/python/src/pyopenvino/core/common.cpp
akuporos: updated

src/bindings/python/tests/test_runtime/test_tensor.py
akuporos: added

src/bindings/python/src/pyopenvino/core/tensor.cpp
akuporos: added

src/bindings/python/tests/test_runtime/test_tensor.py
akuporos: updated

jiwaszki (approved):

PR #18467, opened 2023-07-11
Luwei/brg avx2

luweizhou2016:
### Details:
 - *Add brgemm AVX2 kernel  support for FC and Matmul*

### Tickets:
 - *CVS-105756*


src/plugins/intel_cpu/tests/functional/single_layer_tests/convolution.cpp
```diff
@@ -220,7 +220,8 @@ TEST_P(ConvolutionLayerCPUTest, CompareWithRefs) {
 
     if (!priority.empty()) {
         // Skip tests for brgconv convolution where kernel size = 1x1
-        if (priority[0] == "brgconv_avx512" || priority[0] == "brgconv_avx512_amx") {
+        if (priority[0] == "brgconv_avx512" || priority[0] == "brgconv_avx512_amx" ||
+            priority[0] == "brgconv_avx2") {
```
luo-cheng2021, line 224: using `one_of(priority[0], "brgconv_avx512", "brgconv_avx512_amx", "brgconv_avx2")` may be concise.

luo-cheng2021 (approved):
LGTM.

luweizhou2016:
Applied review comments. Thx!

src/plugins/intel_cpu/src/node.cpp
```diff
@@ -152,9 +152,6 @@ Node::Node(const std::shared_ptr<ngraph::Node>& op,
                 str != "cpu:unknown")
                 IE_THROW() << "Unsupported CPU implementation " << str << " for node " << getName();
         }
-        // add default primitive priorities as a fallback for the custom ones
```
dmitry-gorokhov, line 155: Could you please ellaborate why you deleted this code?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -688,7 +690,7 @@ void FullyConnected::initSupportedPrimitiveDescriptors() {
     for (auto& desc : descs) {
         auto first_desc = dnnl::primitive_desc(DnnlExtensionUtils::clone_primitive_desc(desc.get()));
 
-        const bool first_match = customImplPriorities.empty();
+        const bool first_match = !customImplPriorities.empty();
```
dmitry-gorokhov, line 693: The same: could you please share details thy this change is needed?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -538,6 +538,8 @@ const std::vector<impl_desc_type>& FullyConnected::getDefaultImplPriority() {
         impl_desc_type::gemm_any,
         impl_desc_type::gemm,
         impl_desc_type::jit_gemm,
+        //Lower down brgemm avx2 primitive priority than jit_gemm for perf issue.
+        impl_desc_type::brgemm_avx2,
```
dmitry-gorokhov, line 542: Is brgemm_avx2 shows lower performance both for Convolution and FC/MatMul primitives. My understanding was that is mostly applicable for Convolutions. is that correct?

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -347,6 +347,9 @@ const std::vector<impl_desc_type>& Convolution::getDefaultImplPriority() {
         impl_desc_type::jit_avx2_dw,
         impl_desc_type::jit_avx2_1x1,
         impl_desc_type::jit_avx2,
+        //Lower down brgemm avx2 primitive priority than jit avx2 for perf issue.
```
dmitry-gorokhov, line 350: If we have ticket to track brgemm vs jit perf issue I would ask you to add the number here.
Same for other nodes

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
```diff
@@ -229,10 +229,12 @@ std::vector<CPUSpecificParams> filterSpecificParams() {
     return specificParams;
 }
 
-std::vector<CPUSpecificParams> filterSpecificParams_Brgemm() {
+std::vector<CPUSpecificParams> filterSpecificParams_Brgemm(bool tryBrgAVX2 = false) {
```
dmitry-gorokhov, line 232: Why cannot we add brgemm_avx2 unconditionally? My expections is that brgemm coverage between avx512 and avx2 should be identical

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
```diff
@@ -759,7 +761,7 @@ const auto fullyConnectedParams2D_Brgemm_nightly = ::testing::Combine(::testing:
 const auto testParams2D_Brgemm_nightly = ::testing::Combine(fullyConnectedParams2D_Brgemm_nightly,
                                              ::testing::Values(MatMulNodeType::FullyConnected),
                                              ::testing::ValuesIn(fusingParamsSet2D_nightly),
-                                             ::testing::ValuesIn(filterSpecificParams_Brgemm()));
+                                             ::testing::ValuesIn(filterSpecificParams_Brgemm(true)));
```
dmitry-gorokhov, line 764: In current file only FC case is covered with brgemm_avx2?  Can we extend on matmul case as well?

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/matmul_quantized_subgraph.cpp
```diff
@@ -0,0 +1,161 @@
...
+                ASSERT_EQ(node->get_output_element_type(0), outType);
+                ASSERT_EQ(node->get_input_element_type(0), inType);
+            }
+        }
+    }
+};
+
+TEST_P(MatmulBrgemmInt8Test, CompareWithRefs) {
+    // only cover avx2_vnni
+    if (InferenceEngine::with_cpu_x86_avx512_core() || !InferenceEngine::with_cpu_x86_avx2_vnni())
```
dmitry-gorokhov, line 130: This test looks pretty generic and defintetelly useful. Can we extend for avx512/AMX isas as well (maybe in separate PR)?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
luweizhou2016: customImplPriorities vector is the implement parsed via Ngraph node RTinfo.  

When there is no customImplPriorities:
Current first_match behavior in master branch would only add one dnnl::primitive_desc for one memory layout type and this  dnnl::primitive_desc implement type is based on ONEDNN list priority. For example, for fp32 current master code would only add one implement type(brgemm av512) for nhwc layout. Such as, For INT8 model, it is found the implement type list priority can't work when we want kernel type jit_gemm priority higher than avx2 brgemm. Because only  brgemm_avx2 is added into the dnnl::primitive_desc for AVX2_VNNI ISA platforms. 

I recalled the former code would add all the supported dnnl::primitive_desc implement type(brgemm av512, brgemm avx2 and  other) into the supportedPrimitiveDescriptor list. So this changes would added all supported into.  Only by this way, selectPreferPrimitiveDescriptor() can work as OV priority list.
When there is customImplPriorities parsed from RT info:
No need to add all supported dnnl::primitive_des. Only added the implement type (can be found in the  customImplPriorities ) into list.

So that is why made this changes.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
luweizhou2016: It should be. Already prioritize brgemm avx2 higher than jit gemm for FC/Matmal. Internal test shows INT8 precision has some perf&latency GEOMEAN gain. Others precision GEOMEAN doesn't almost  changes. Updated.

src/plugins/intel_cpu/src/nodes/conv.cpp
luweizhou2016: Added.

src/plugins/intel_cpu/src/node.cpp
luweizhou2016: customImplPriorities vector is the implement parsed via Ngraph node RTinfo. Current master code produce the prioritylist by customImplPriorities + default priority list.   If having customImplPriorities, no need to add all supported dnnl::primitive_desc. Just found one.  So in this case, no need to add default priority list as fall back list. If can't found the implement type which customImplPriorities specified. Current code would use the  first_desc  as  fallback

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
luweizhou2016: brgemm_matmul_t can't support avx2 in stock onednn. That is why the reason we have to pass on argument into to decide whether to add the brgemm_avx2 into list.

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
luweizhou2016: Same as above. Matmul brgemm can support INT8 inference on avx2_vnni, but can't support fp32 inference on avx2 ISA. That is the onednn limitation.

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/matmul_quantized_subgraph.cpp
luweizhou2016: Would be added in the separate PR further.

luweizhou2016:
@dmitry-gorokhov , Applied.

src/plugins/intel_cpu/src/node.cpp
dmitry-gorokhov: customImplPriorities vector is used not only by tests, but also can be set-up by modifing ov::model itself. This is not officialy documented functionality but still used to fix some execution issue on customer side w/o OV recompilation. 
So the idea is to specify some custom priority list (like prioritizing some impls over others), but if, for some specific operations, primitives from custom list cannot be used (like some parameters are not supported) we will fallback on default OV priorities. In other words custom + default lists merge required to relax limitations on custom impls applicability.

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
dmitry-gorokhov: Got it, thanks. Please add this info as comment here.

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -775,7 +778,7 @@ void Convolution::initSupportedPrimitiveDescriptors() {
         auto& desc = descs[dIdx];
         auto first_desc = dnnl::primitive_desc(DnnlExtensionUtils::clone_primitive_desc(desc.get()));
 
-        const bool first_match = customImplPriorities.empty();
+        const bool first_match = !customImplPriorities.empty();
```
EgorDuplensky, line 781: The idea of the first_match logic was to avoid extra primitive_desc creation (create only first_match primitive_desc)
As you also found out, if we change the primitive priority in getDefaultImplPriority we no longer able to work with first_match and have to create all the primitive_descs because of limitations of oneDNN api.
So even we do not like to extend the fork with additional changes, I think the more proper fix will be to actually change the primitive priority in oneDNN's implementation priority lists in our fork.
And discard all the changes made in initSupportedPrimitiveDescriptors().
@dmitry-gorokhov FYI

luweizhou2016: @EgorDuplensky I have some concerns on this. Once the first_match is true, It means inside one input&output layout pair, only one implement type primitive is chosen based on onednn list priority.  Also means that OV CPU priority list can not take effect inside one specified layout pair.
But across different input&output layout pairs, OV priority list is working based on onednn list priority.
It would make things a little complicated and confusing.  For INT8,  most of layout is NHWC, it means OV priority can't work. But for FP32, OV priority seems work as expected, but the hidden reason is different layout are supported by FP32.
So personally, I would like decouple OV list and stock onednn list priority.

src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul.cpp
luweizhou2016: Applied

src/plugins/intel_cpu/src/node.cpp
luweizhou2016: restore custom + default  back.

luweizhou2016:
@dmitry-gorokhov @EgorDuplensky , I have just removed the first_match and iterate all the ONEDNN implement types. Only by this way, OV priority list can work and decoupled with ONEDNN priority list. Thx!

src/plugins/intel_cpu/src/nodes/conv.cpp
EgorDuplensky: The problem is that having a separate priority list inside the plugin is not free.
oneDNN does not expose the possibility of changing the priorities of the implementations.
So we have to create all the possible primitive implementations and only then choose one based on our own priorities.
And this step has significant impact on compile_model time (you can verify it by measuring the compile_model time after your changes)
From my opinion we should consider this as unacceptable way of working with the primitive implementations (I mean creating all the possible primitive_desc implementations)

luweizhou2016: Agreed on the impact on loading time overhead. 
But my understanding is that  our own priority list in conv/matmul/fc is designed to not follow the onenn default priority list. That is why in below selectPreferPrimitiveDescriptor()  the OV priority list  would be the outmost loop. 
https://github.com/openvinotoolkit/openvino/blob/e2c3d0fe7318921984855c132f00baed997d2c3b/src/plugins/intel_cpu/src/node.cpp#L273

 To ensure our priority work, we have to accept the extra overhead. Also the extra overhead should be much less than expected in my understanding, because more primitive descriptors are constructed but no generating kernel. Most of the overhead should be in generating kernel in my memory.

Above all, I think the **key point** is whether we want to keep OV node priority work in the future. If
 NOT, we can just change forked ONEDNN list. But if yes, maybe need to accept this iteration overhead.   Maybe need @dmitry-gorokhov also give some comments on this. 

Some internal perf test data  uploaded https://jira.devtools.intel.com/browse/CVS-105756 for data. But don't know why have loading time gain for most precision.  I have re-triggered the test and see.

EgorDuplensky: The custom priority list at least provides a possibility to completely skip a particular implementation (as it was done for brgemm_avx2)
> Most of the overhead should be in generating kernel in my memory.
In general yes, it is true, generating kernel is much heavier operation, but there are cases where creation of many primitive_descs  also impacts loadtime (especially brgemm ones).

luweizhou2016: @EgorDuplensky , Yes. I ran the test manually today to check loading time. Removing first_match did increase the loading time as expected. I just aligned with @dmitry-gorokhov  and will remove the brgconv avx2 support from the convolution priority list and keep brgemm avx2 support for FC and Matmul.  Then no need to change onednn code.  Will update PR soon.

luweizhou2016:
@dmitry-gorokhov @EgorDuplensky . Updated with disabling brgconv avx2 implement type.

EgorDuplensky (approved):

EgorDuplensky:
> @dmitry-gorokhov @EgorDuplensky . Updated with disabling brgconv avx2 implement type.

Could you please create a PR for oneDNN fork and specify the cross links between the PRs

luweizhou2016:
> > @dmitry-gorokhov @EgorDuplensky . Updated with disabling brgconv avx2 implement type.
> 
> Could you please create a PR for oneDNN fork and specify the cross links between the PRs in their description

@EgorDuplensky , There is no onednn fork changed because brgemm_conv_avx2 is not in OV priority list. So would not choose the onednn brgemm_conv_avx2 kernel on avx2 platform.

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -367,7 +367,7 @@ const std::vector<impl_desc_type>& Convolution::getDefaultImplPriority() {
     return priorities;
 }
 
-const bool Convolution::isBrgConvAvailable = dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx512_core);
+const bool Convolution::isBrgConvAvailable = dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx2);
```
dmitry-gorokhov, line 370: I guess that condition should be restored back. As well as other changes inside conv.cpp

luweizhou2016: Yes. Will updated.

dmitry-gorokhov (approved):

EgorDuplensky:
> > > @dmitry-gorokhov @EgorDuplensky . Updated with disabling brgconv avx2 implement type.
> > 
> > 
> > Could you please create a PR for oneDNN fork and specify the cross links between the PRs in their description
> 
> @EgorDuplensky , There is no onednn fork changed because brgemm_conv_avx2 is not in OV priority list. So would not choose the onednn brgemm_conv_avx2 kernel on avx2 platform.

Ok, I see. Thanks

PR #18468, opened 2023-07-11
Do not return default context from the compiled model

ilyachur:
### Details:
 - GPU compiled model returns the context also in case if compiled model was created without context

### Tickets:
 - CVS-114983


ilya-lavrenov (approved):

src/plugins/proxy/src/plugin.cpp
```diff
@@ -378,11 +393,25 @@ std::shared_ptr<ov::IRemoteContext> ov::proxy::Plugin::get_default_context(const
...
+    auto device_model = get_core()->import_model(model, device, device_config);
+    ov::SoPtr<ov::IRemoteContext> device_context;
+    ov::RemoteContext remote_context;
+    try {
+        device_context = {device_model->get_context(), device_model._so};
+        remote_context = ov::proxy::RemoteContext::make_context(ov::SoPtr<ov::IRemoteContext>(
+            std::make_shared<ov::proxy::RemoteContext>(device_context, dev_name, dev_idx, has_dev_idx, is_new_api),
+            device_model._so));
+    } catch (const ov::NotImplemented&) {
+    }
```
nkogteva, line 412: It seems code is duplicated in 2 places. Maybe create function for `RemoteContext` creation?

src/plugins/proxy/src/plugin.cpp
```diff
@@ -14,12 +14,30 @@
...
 
+ov::RemoteContext create_proxy_context(const ov::SoPtr<ov::ICompiledModel>& compiled_model,
+                                       const std::string dev_name,
+                                       size_t dev_idx,
+                                       bool has_dev_idx,
+                                       bool is_new_api) {
+    ov::SoPtr<ov::IRemoteContext> device_context;
+    ov::RemoteContext remote_context;
+    try {
+        device_context = {compiled_model->get_context(), compiled_model._so};
```
songbell, line 32: @ilyachur here proxy plugin can construct the context, but auto/batch plugin can't access this constructor.. why icompiled_model base class does not hold IRemoteContext instead of RemoteContext?

ilyachur: @songbell Good catch!

Thank you. I updated the API and provide helpers which can be used for meta plugins

ilya-lavrenov (approved):

nkogteva (approved):

src/plugins/proxy/src/plugin.cpp
```diff
@@ -299,15 +300,35 @@ ov::Any ov::proxy::Plugin::get_property(const std::string& name, const ov::AnyMa
...
+    auto is_new_api = get_core()->is_new_api();
+    ov::SoPtr<ov::IRemoteContext> device_context;
+    ov::SoPtr<ov::IRemoteContext> remote_context;
+    try {
+        device_context = compiled_model->get_context();
+        if (!device_context._so)
+            device_context._so = compiled_model._so;
+        remote_context =
+            std::make_shared<ov::proxy::RemoteContext>(device_context, dev_name, dev_idx, has_dev_idx, is_new_api);
+    } catch (const ov::NotImplemented&) {
```
songbell, line 318: is this exception enough? if underlying hardware still 1.0, I suppose still IE exceptions

src/plugins/proxy/src/plugin.cpp
```diff
@@ -299,15 +300,35 @@ ov::Any ov::proxy::Plugin::get_property(const std::string& name, const ov::AnyMa
...
+    const ov::AnyMap& properties) const {
+    auto dev_name = get_device_name();
+    auto dev_idx = get_device_from_config(properties);
+    auto has_dev_idx = is_device_in_config(properties);
+    auto is_new_api = get_core()->is_new_api();
+    ov::SoPtr<ov::IRemoteContext> device_context;
+    ov::SoPtr<ov::IRemoteContext> remote_context;
+    try {
+        device_context = compiled_model->get_context();
+        if (!device_context._so)
```
songbell, line 314: when can we hit this if clause?

ilyachur: Hardware plugins return context without so, but in case if undert proxy we use auto batch or another meta plugins, we already will have so object inside

src/plugins/proxy/src/plugin.cpp
ilyachur: Yes it is ok, in case of legacy API it is ok if method will throw an exception.

PR #18470, opened 2023-07-11
[GPU] fix compilation context to use kernel_impl_params as key

e-ddykim:
### Details:
 - This PR fixes the async compilation context to use the kernel_impl_params as the keys to overcome hash collisions.

### Tickets:
 - 114982

### DIFF
--- a/src/plugins/intel_gpu/src/graph/include/compilation_context.hpp
+++ b/src/plugins/intel_gpu/src/graph/include/compilation_context.hpp
@@ -7,14 +7,15 @@
 #include "openvino/runtime/threading/cpu_streams_executor.hpp"
 #include <functional>
 #include <memory>
+#include "intel_gpu/graph/kernel_impl_params.hpp"
 
 namespace cldnn {
 
 class ICompilationContext {
 public:
     using Task = std::function<void()>;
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/src/graph/include/compilation_context.hpp
```diff
@@ -7,14 +7,15 @@
...
+#include "intel_gpu/graph/kernel_impl_params.hpp"
 
 namespace cldnn {
 
 class ICompilationContext {
 public:
     using Task = std::function<void()>;
-    virtual void push_task(size_t key, Task&& task) = 0;
-    virtual void remove_keys(std::vector<size_t>&& keys) = 0;
+    virtual void push_task(kernel_impl_params key, Task&& task) = 0;
```
yeonbok, line 17: Can't we make this header to use template and do actual specialization in the cpp code for more general impl of the abstract class?

yeonbok (approved):

src/plugins/intel_gpu/src/graph/include/compilation_context.hpp
e-ddykim: As discussed offline, I'm not sure about how to apply template to the static creation method.

PR #18473, opened 2023-07-11
do streams calculation before UpToCpuSpecificOpSet() in CPU plugin

wangleis:
### Details:
 - *do streams calculation before UpToCpuSpecificOpSet() in CPU plugin*
 - *add WA to skip saving throughput mode information during latency mode inference and export/import test case with different performance hint

### Tickets:
 - *ticket-id*

### DIFF
--- a/tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
+++ b/tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
@@ -19,6 +19,17 @@ def get_pytorch_decoder(model, input_shape, example_inputs, args):
     except Exception as e:
         log.error("PyTorch frontend loading failed")
         raise e
+    try:
+        import nncf
+        from nncf.torch.nncf_network import NNCFNetwork
+        from packaging import version
+
+        if isinstance(model, NNCFNetwork):
+            if version.parse(nncf.__version__) < version.parse("2.6"):
+                raise RuntimeError(
+                    "NNCF models produced by nncf<2.6 are not supported directly. Please export to ONNX first.")
+    except:
+        pass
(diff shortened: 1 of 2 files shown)


src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -460,6 +460,20 @@ Engine::LoadExeNetworkImpl(const InferenceEngine::CNNNetwork &network, const std
...
+
+    // update the props after the perf mode translated to configs
+    // TODO: Clarify the behavior of SetConfig method. Skip eng_config or not?
+    Config conf = engConfig;
+
+    conf.readProperties(config);
+    if (is_cpu_map_available()) {
+        GetPerformanceStreams(conf, nGraphFunc);
+    }
+
```
sunxiaoxia2022, line 476: This change can cause the number of streams is different from previous version.

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -460,6 +460,20 @@ Engine::LoadExeNetworkImpl(const InferenceEngine::CNNNetwork &network, const std
     DEBUG_LOG(PrintableModel(*nGraphFunc, "org_"));
 
     Transformations transformations(nGraphFunc, enableLPT, inferencePrecision, isLegacyAPI(), snippetsMode, engConfig);
```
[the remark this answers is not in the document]
wangleis: Updated. This update will change the nGraph based of streams calculation, will check the performance impact again.

src/plugins/intel_cpu/src/plugin.cpp
wangleis: Thank you for reminding. This is recorded as a known issue in review.

wangleis:
Close this PR and follow up this change in https://github.com/openvinotoolkit/openvino/pull/18911 without WA

PR #18475, opened 2023-07-11
rework remote context which causes auto tests sporadic failure on CI

songbell:
### Details:
 - refactor of remote_context in auto will be in cvs-115052

### Tickets:
 - 106625


songbell:
@ilya-lavrenov ilya, can we merge this first to avoid such failures block ci progress? I will fix the code and enable the tests after context wrapper merged in https://github.com/openvinotoolkit/openvino/pull/18468

src/plugins/intel_gpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
```diff
@@ -137,6 +137,8 @@ std::vector<std::string> disabledTestPatterns() {
             // For some strange reason (bug?) output format cannot have a rank greater than 4 for dynamic shape case,
             // because it crashes in some random places during "reorder_inputs" pass.
             R"(.*UniqueLayerDynamicGPUTest.*\(\d*\.\d*\.\d*\.\d*\.\d*\).*axis.*)",
+            // cvs-115052
+            R"(.*smoke_Multi.*canCreateContextThenRequestThenBlobsAndInfer.*)",
```
ilyachur, line 141: Why do you start to skip new tests?

songbell: the current implementation of remote context is not correct in auto plugin, I will re-work remote context and re-enable these test after https://github.com/openvinotoolkit/openvino/pull/18468, this PR is just to temporarily disable the remote context function in Auto(which currently will introduce sporadic failures) in case the sporadic failures blocking too many CI builds.

ilya-lavrenov (approved):

PR #18385, opened 2023-07-05
Consolidate common test utils targets

olpipi:
### Details:
 - consolidate common_test_utils and ngraph_test_util targets

### Tickets:
 - [CVS-111358](https://jira.devtools.intel.com/browse/CVS-111358)

PR #18477, opened 2023-07-11
[Snippets] ExtractReshapesFromMHA transformation

v-Golubev:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


v-Golubev:
@a-sidorova @IvanNovoselov could you please take a look? Thanks!

src/common/snippets/src/pass/extract_reshapes_from_mha.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include <openvino/opsets/opset1.hpp>
+
+#include "openvino/core/rt_info.hpp"
+#include "openvino/op/util/binary_elementwise_arithmetic.hpp"
+#include "openvino/pass/pattern/op/wrap_type.hpp"
+#include "snippets/itt.hpp"
+#include "snippets/snippets_isa.hpp"
+#include "snippets/pass/mha_tokenization.hpp"
+
```
IvanNovoselov, line 15: Minor: it seems like the following headers are not needed:
* #include "openvino/op/util/binary_elementwise_arithmetic.hpp"
* #include "snippets/snippets_isa.hpp"
If it is indeed so, can we remove them maybe?

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
```diff
@@ -666,32 +668,33 @@ void Transformations::MainSnippets(void) {
...
-                const auto needed_num_of_threads = 12lu;
-                const auto is_unsupported_parallel_work_amount =
-                    parallel_get_num_threads() / 2 > parallel_work_amount &&
-                    static_cast<size_t>(parallel_work_amount) < needed_num_of_threads;
-                return is_unsupported_parallel_work_amount;
-            },
-            snippets::pass::TokenizeMHASnippets);
+        auto mha_callback = [&](const std::shared_ptr<const ov::Node>& n) -> bool {
+            // Tranformation callback is called on MatMul0
+            if (!is_supported_matmul(n))
```
IvanNovoselov, line 673: We have `is_supported_matmul()` in TokenizeMHASnippets callback, and then `is_matmul_supported()` inside the transformation. Do you think we can unite these checks?

IvanNovoselov (approved):

src/common/snippets/src/pass/extract_reshapes_from_mha.cpp
v-Golubev: Good idea, done. Thanks!

src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
v-Golubev: `ov::snippets::pass::TokenizeMHASnippets::is_matmul_supported` contains checks only for the first matmul in MHA pattern whereas `is_supported_matmul()` lambda contains requirements for both matmuls. So I think we don't need to unite these checks.
However, it's good idea to rename static `TokenizeMHASnippets` method to `is_matmul0_supported`: it is done

BTW after offline discussion with @a-sidorova CPU callback's part was refactored: `snippets::pass::ExtractReshapesFromMHA` callback now don't check child matmul because it's not guarantied that we find it in while loop (because the second matmul is not defined in the transformation matcher)

a-sidorova (approved):

v-Golubev:
@a-sidorova @IvanNovoselov thanks!
@dmitry-gorokhov could you please make the final review?

dmitry-gorokhov (approved):

dmitry-gorokhov:
Lets merge after https://github.com/openvinotoolkit/openvino/pull/18385

PR #18489, opened 2023-07-11
[GPU] Dynamism support for Proposal

andrew-k-park:
### Details:
 - Update `Proposal` to use ngraph shape inference
 - Align proposal cpu impl to ngraph reference
 - Add static and dynamic TCs for ov_gpu_func_tests and ov_gpu_unit_tests
 - Fix issue where proposal outputs was not selected as lockable memory type for dynamic TCs

### Tickets:
 - 109742, 100236


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1106,6 +1107,10 @@ memory::ptr primitive_inst::allocate_output(engine& _engine, memory_pool& pool,
     // For outputs, cpu prim we want to have lockable alloc type
     // Also if the successor of a node is an cpu, then memory needs to be lockable.
     bool is_cpu = _node.get_selected_impl() ? _node.get_selected_impl()->is_cpu() : false;
+    // WA: Since proposal only have cpu impl, if selected_impl of node is empty
+    //     We need to set it to be selected as lockable memory type manually
+    if (_node.is_type<proposal>() && !_node.get_selected_impl())
+        is_cpu = true;
```
yeonbok, line 1113: why this (impl is cpu but _node.get_selected_impl() is empty) can happen?

andrew-k-park: update to select impl for dynamic or static shape during compile_graph as discussed

yeonbok (approved):

PR #18494, opened 2023-07-12
[GPU] PriorBox dynamic shape support

wilson-seok:
### Details:
 - Add dynamic shape support for priorbox/priorboxclustered

### Tickets:
 - 114498
 - 100195

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
@@ -60,8 +60,6 @@ struct kernel_impl_params {
     std::vector<std::map<size_t, primitive_id>> io_output_maps = {};
     std::vector<size_t> output_size;
     std::vector<size_t> img_size;
-    mutable tensor input1{};
-    mutable tensor input2{};
 
     kernel_impl_params() : prog(nullptr), strm(nullptr), desc(nullptr), unique_id(0) {}
 
@@ -134,17 +132,6 @@ struct kernel_impl_params {
     bool is_type() const {
         return std::static_pointer_cast<const PType>(desc)->type == PType::type_id();
     }
-
-    void set_input1_tensors(tensor& tensor) const {
-        input1 = tensor;
-    }
-
-    void set_input2_tensors(tensor& tensor) const {
-        input2 = tensor;
-    }
-
-    tensor get_input1_tensors() const { return input1; }
-    tensor get_input2_tensors() const { return input2; }
 
 virtual primitive_type_id type() const { return desc->type; }
 
--- a/src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
@@ -58,7 +58,7 @@ struct kernel_impl_params {
     std::vector<std::shared_ptr<program>> inner_progs = {};
     std::vector<std::shared_ptr<network>> inner_nets = {};
     std::vector<std::map<size_t, primitive_id>> io_output_maps = {};
-    // TODO : Thsese values are temporarily added for prior box. Such values decided at runtime shape infer and primitive creation will be handled with more generalized way in the near future.
+    // TODO : These values are temporarily added for prior box. Such values decided at runtime shape infer and primitive creation will be handled with more generalized way in the near future.
     std::vector<size_t> output_size;
     std::vector<size_t> img_size;
 
--- a/src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
@@ -58,7 +58,8 @@ struct kernel_impl_params {
     std::vector<std::shared_ptr<program>> inner_progs = {};
     std::vector<std::shared_ptr<network>> inner_nets = {};
     std::vector<std::map<size_t, primitive_id>> io_output_maps = {};
-
+    std::vector<size_t> output_size;
+    std::vector<size_t> img_size;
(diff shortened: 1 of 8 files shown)


src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
```diff
@@ -131,6 +134,17 @@ struct kernel_impl_params {
...
+        input1 = tensor;
+    }
+
+    void set_input2_tensors(tensor& tensor) const {
+        input2 = tensor;
+    }
+
+    tensor get_input1_tensors() const { return input1; }
+    tensor get_input2_tensors() const { return input2; }
+
```
yeonbok, line 147: 1. At creation of impl, we only need the "values". Why should we save it as tensor? 
(e.g., we can store the dim sizes std::vector<size_t> which contains width, height. )
2. Also I think we can generalize such runtime data in future using BinaryOutputBuffer as a next step

wilson-seok: Got it! I modified it and binarybuffer will be implemented in next phase.

src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
```diff
@@ -58,6 +58,8 @@ struct kernel_impl_params {
     std::vector<std::shared_ptr<program>> inner_progs = {};
     std::vector<std::shared_ptr<network>> inner_nets = {};
     std::vector<std::map<size_t, primitive_id>> io_output_maps = {};
+    std::vector<size_t> output_size;
+    std::vector<size_t> img_size;
```
yeonbok, line 62: Currently these values are only used for prior_box. It may cause confusion from in/output_layouts. So please add comments like 
// TODO : Thsese values are temporarily added for prior box. Such values decided at runtime shape infer and primitive creation will be handled with more generalized way in the near future.

src/plugins/intel_gpu/include/intel_gpu/primitives/prior_box.hpp
```diff
@@ -135,10 +209,82 @@ struct prior_box : public primitive_base<prior_box> {
...
+        attrs.step_heights = step_height;
+        attrs.step = step;
+        attrs.offset = offset;
+        attrs.variances = variance;
+        return attrs;
+    }
+
     /// @brief Spatial size of generated grid with boxes.
-    tensor output_size{};
+    mutable cldnn::tensor output_size{};
```
yeonbok, line 285: why should it be mutable? We dont change primitive's variables after creation.

src/plugins/intel_gpu/include/intel_gpu/graph/kernel_impl_params.hpp
vladimir-paramuzov: Why is it needed at all? Can't values of the corresponding tensors be read when it's needed instead of this WA?

yeonbok: Umm yeah we can read again in the ocl impl creation for these values but it will affect performance. Isn't it better to read and store one time at calc_output_layout and then use them at ocl impl creation?

yeonbok: There are such values read or calculated at calc_output_layout such as crop's offset and these values. In my thought, we can store such "runtime parameters" as a binary format in impl_param and then parse it at each primitive type on demand.

wilson-seok: Updated comment. Thanks!

src/plugins/intel_gpu/include/intel_gpu/primitives/prior_box.hpp
wilson-seok: Removed 'mutable'. Thanks!

src/plugins/intel_gpu/include/intel_gpu/primitives/prior_box.hpp
```diff
@@ -5,10 +5,13 @@
 #pragma once
 
 #include "primitive.hpp"
+#include "openvino/op/prior_box.hpp"
+#include "openvino/op/prior_box_clustered.hpp"
 
 #include <cmath>
 #include <vector>
 #include <limits>
+#include <map>
```
yeonbok, line 14: do we need this map in this header?

src/plugins/intel_gpu/src/graph/prior_box.cpp
```diff
@@ -437,6 +440,74 @@ layout prior_box_inst::calc_output_layout(prior_box_node const& node, kernel_imp
...
+            p_param->img_size.push_back(static_cast<size_t>(img_width));
+            p_param->img_size.push_back(static_cast<size_t>(img_height));
+        } else { //int32_t
+            auto output_height = reinterpret_cast<int32_t*>(output_size_lock.data())[0];
+            auto output_width = reinterpret_cast<int32_t*>(output_size_lock.data())[1];
+            auto img_height = reinterpret_cast<int32_t*>(img_size_lock.data())[0];
+            auto img_width = reinterpret_cast<int32_t*>(img_size_lock.data())[1];
+
+            p_param->output_size.push_back(static_cast<size_t>(output_width));
+            p_param->output_size.push_back(static_cast<size_t>(output_height));
```
yeonbok, line 483: if this is called multiple times, output_size and img_size will have increased data ..Please handle it.

wilson-seok: Yes, I updated it to use only 2 elements. Also I added multiple image sizes in functional test case. Thanks!

yeonbok (approved):

PR #18501, opened 2023-07-12
[PT FE]: multiple fixes for models from optimum testing scope

eaidova:
### Details:
 - *supported aten::round, aten::multiply ops*
 - *aten::lift* based ops added to skipping mode, they used for in autograd functions in backward mode*

### Tickets:
 - *ticket-id*

verified on ibert and convert models families


src/frontends/pytorch/src/op_table.cpp
```diff
@@ -280,6 +281,9 @@ const std::map<std::string, CreatorFunction> get_supported_ops() {
         {"aten::leaky_relu", op::translate_1to1_match_2_inputs<opset10::PRelu>},
         {"aten::leaky_relu_", op::inplace_op<op::translate_1to1_match_2_inputs<opset10::PRelu>>},
         {"aten::len", op::translate_len},
+        {"aten::lift", op::skip_node},
```
mvafin, line 284: Could you please add some comments why we skip lift ops and what they do

src/frontends/pytorch/src/op_table.cpp
```diff
@@ -280,6 +281,10 @@ const std::map<std::string, CreatorFunction> get_supported_ops() {
         {"aten::leaky_relu", op::translate_1to1_match_2_inputs<opset10::PRelu>},
         {"aten::leaky_relu_", op::inplace_op<op::translate_1to1_match_2_inputs<opset10::PRelu>>},
         {"aten::len", op::translate_len},
+        // lift ops is torchscript specific ops responsible for tensors coping with guarantee of new memory allocation
```
slyalin, line 284: Does it mean that successive operations that consumes the result of lift* ops shouldn't use in-place semantics?

PR #18502, opened 2023-07-12
[PyOV] py wrappers for clone/deepcopy model

akuporos:
### Details:
 - extend test to check model type for cloned/copied model
 - add py wrappers for clone/deepcopy

### Tickets:
 - 115096


src/bindings/python/src/openvino/runtime/ie_api.py
```diff
@@ -2,7 +2,8 @@
 # Copyright (C) 2018-2023 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Any, Iterable, Union, Optional
+from __future__ import annotations
```
jiwaszki, line 5: Why are future features needed, what is the reason? Isn't `"CompiledModel"` good enough?

src/bindings/python/src/openvino/runtime/utils/util.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
```
jiwaszki, line 5: Same here.

src/bindings/python/src/openvino/runtime/utils/util.py
```diff
@@ -0,0 +1,16 @@
...
+from openvino._pyopenvino.util import clone_model as clone_model_base
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from openvino.runtime import Model
```
jiwaszki, line 11: Isn't this introducing some circular dependency normally? It only triggers with mypy checks, right?

jiwaszki (changes requested):

src/bindings/python/src/openvino/runtime/utils/util.py
akuporos: yes, this import is trigger only with mypy checks (another way -- suppress annotation check)

src/bindings/python/src/openvino/runtime/utils/util.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def clone_model(model: Model) -> Model:
+    from openvino.runtime import Model
+    return Model(clone_model_base(model))
```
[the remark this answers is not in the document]
akuporos: Sure, I'll deprecate `clone_model`, but it is still need to return `Model` python wrapper. We don't know if someone use this function.

src/bindings/python/src/openvino/runtime/ie_api.py
akuporos: I just like it to be in one style :)
Otherwise, returning type for `clone` and `deepcopy` also should be take into quotes.

src/bindings/python/src/openvino/runtime/utils/util.py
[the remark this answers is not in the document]
akuporos: deprecated `clone_model`

src/bindings/python/src/openvino/runtime/utils/util.py
akuporos: removed

src/bindings/python/src/openvino/runtime/ie_api.py
akuporos: removed __future__

jiwaszki (approved):

PR #18508, opened 2023-07-12
Update low precision transformations to work with Pad-12

evkotov:
### Details:
 - update transformation to accept v12::Pad with positive indexes

### Tickets:
 - 114251


src/common/low_precision_transformations/src/pad.cpp
```diff
@@ -152,16 +154,35 @@ bool PadTransformation::transform(TransformationContext& context, ngraph::patter
     return true;
 }
 
+namespace {
+    bool hasNegativeIndexes(const std::shared_ptr<ov::op::util::PadBase>& pad) {
```
itikhono, line 158: do we have a util method for this?

src/common/low_precision_transformations/src/markup_precisions.cpp
```diff
@@ -13,6 +13,7 @@
 #include <ngraph/opsets/opset4.hpp>
 #include <ngraph/opsets/opset5.hpp>
 #include <ngraph/opsets/opset6.hpp>
+#include "openvino/opsets/opset12.hpp"
```
itikhono, line 16: I believe we need a ticket to update LPT transformations to ov namespace

itikhono (approved):

PR #18509, opened 2023-07-12
[WA] performance issue for plugin api 2.0

riverlijunjie:
### Details:
 - WA for performance issue for plugin api 2,0

### Tickets:
 -  Related to 115223


src/inference/src/dev/isync_infer_request.cpp
```diff
@@ -110,6 +110,7 @@ const std::shared_ptr<const ov::ICompiledModel>& ov::ISyncInferRequest::get_comp
 }
 
 ov::ISyncInferRequest::FoundPort ov::ISyncInferRequest::find_port(const ov::Output<const ov::Node>& port) const {
+    // This function is hotspot, need optimization.
     auto check_nodes = [](const ov::Node* node1, const ov::Node* node2) {
```
riverlijunjie, line 114: If we simply the condition of find port, there will be failure of onnx tests:

```
[ RUN      ] INTERPRETER.onnx_model_generate_proposals
unknown file: Failure
C++ exception with description "Exception from src/inference/src/infer_request.cpp:217:
Check 'found_port.found()' failed at src/inference/src/dev/isync_infer_request.cpp:179:
Cannot find tensor for port rpnrois/sink_port_0[0]:f32[..6,4]

" thrown in the test body.
[  FAILED  ] INTERPRETER.onnx_model_generate_proposals (75 ms)
```

src/inference/src/dev/iasync_infer_request.cpp
```diff
@@ -201,6 +201,12 @@ void ov::IAsyncInferRequest::check_state() const {
     }
 }
 
+void ov::IAsyncInferRequest::check_cancelled_state() const {
+    std::lock_guard<std::mutex> lock{m_mutex};
+    if (m_state == InferState::CANCELLED)
```
ilya-lavrenov, line 206: why `check_state` is not enough?

riverlijunjie: check_state() will check BUSY as well, while in cpu plugin it will only check CANCELLED for several times in a single infer(), we notice at this time the state always BUSY, so check_state() will throw BUSY exception to block infer().
So cpu plugin need method to only check CANCELLED state during infer() is BUSY state.

src/inference/src/dev/isync_infer_request.cpp
```diff
@@ -251,7 +252,10 @@ void ov::ISyncInferRequest::check_tensor(const ov::Output<const ov::Node>& port,
                     " expecting ",
                     port.get_shape(),
                     ".");
-    OPENVINO_ASSERT(tensor.is<ov::RemoteTensor>() || tensor.data() != nullptr, "Tensor data equal nullptr!");
+    // It seems tensor.is<ov::RemoteTensor>() is heavy function, it impacts performance greatly.
+    // OPENVINO_ASSERT(tensor.is<ov::RemoteTensor>() || tensor.data() != nullptr, "Tensor data equal nullptr!");
+    auto remote_tensor = std::dynamic_pointer_cast<ov::IRemoteTensor>(tensor._impl);
```
riverlijunjie, line 257: Test result shows `tensor.is<ov::RemoteTensor>()` is the most factor to impact performance.

image

src/inference/src/dev/iasync_infer_request.cpp
ilyachur: Who will use this method? `check_state` is called in all method, does it mean that CPU will override all default methods?

ilyachur (approved):

PR #18525, opened 2023-07-13
Fix IRDFT for 1D complex input in debug mode

mbencer:
Ticket: CVS-112743

### DIFF
--- a/src/core/reference/src/runtime/reference/utils/fft_common.cpp
+++ b/src/core/reference/src/runtime/reference/utils/fft_common.cpp
@@ -17,7 +17,7 @@ namespace runtime {
 namespace reference {
 namespace fft_common {
 std::vector<int64_t> reverse_shape_of_emulated_complex_tensor(const ngraph::Shape& shape) {
-    assert(shape.size() >= 2);
+    assert(shape.size() >= 1);
     std::vector<int64_t> reversed_shape(shape.begin(), shape.end() - 1);
     std::reverse(reversed_shape.begin(), reversed_shape.end());
     return reversed_shape;


src/core/reference/src/runtime/reference/utils/fft_common.cpp
```diff
@@ -17,7 +17,7 @@ namespace runtime {
 namespace reference {
 namespace fft_common {
 std::vector<int64_t> reverse_shape_of_emulated_complex_tensor(const ngraph::Shape& shape) {
-    assert(shape.size() >= 2);
+    assert(shape.size() >= 1);
```
rkazants, line 20: 1D compex tensor should contain two dimensions, should not it?

rkazants: is test possible to write?

mbencer: @rkazants From the current codebase, we have failing for example `smoke_RDFT_CPU_1D/RDFTTestCPU.CompareWithRefs/prec=f32_IS0=[]_TS0=((9.2))_constAxes=true_axes=((0))_isInverse=true_primitive=jit_avx512`
In this case input shape is `{9,2}`, but `reverse_shape_of_emulated_complex_tensor` is used with input `outer_extended_shape` which is `{2}` in this case.

But agree, that change in `reverse_shape_of_emulated_complex_tensor` can be a bit confusion, so I changed a way of calculating `outer_extended_shape`

praasz (approved):

t-jankowski (approved):

PR #18527, opened 2023-07-13
Partial Values and Labels visualization and update in label equality checks

jane-intel:
1. Visualize upper-lower bounds and labels as an option of VisualizeTree pass
2. Improve logic around equality of two dimensions. Previously if we would have tree dimensions A, B and C and would've set A equal to B and B equal to C -- then there would be no evidence that A equal C. With current implementation -- it has this evidence and can provide it without problem.


ilyachur:
What is the main purpose of this PR?

src/core/dev_api/openvino/core/dimension_tracker.hpp
```diff
@@ -28,56 +28,46 @@ class DimensionTracker {
         OPENVINO_ASSERT(label != no_label, "Can not set zero as label for dimension -- it is reserved for no label");
         d.m_label = label;
     }
-
+    static bool has_label(const ov::Dimension& d);
     static label_t get_label(const ov::Dimension& d) {
```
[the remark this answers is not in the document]
jane-intel: It wan't working before I exported the symbols. Now it works with implementations in cpp -- I pushed the changes

src/core/dev_api/openvino/core/dimension_tracker.hpp
```diff
@@ -16,68 +16,43 @@ namespace ov {
...
-    const ValTable& get_value_equivalence_table() const {
-        return value_table_of_equivalence;
-    }
+    explicit TableOfEquivalence(label_t label = 1) : current_label(label) {};
+    void set_as_equal(const ov::Dimension& lhs, const ov::Dimension& rhs);
+    bool are_equal(const ov::Dimension& lhs, const ov::Dimension& rhs);
 
+    const EqTable& get_equivalence_table() const;
+    const ValTable& get_value_equivalence_table() const;
+    label_t get_next_label();
```
praasz, line 53: Assume this next label is to have unique labels, is that correct?
If this will be unique labels maybe it should be unique for instances of `TableOfEquivalence` or per instance is also OK?

jane-intel: It is okay to have unique labels per instance of `TableOfEquivalence` because we have limited scopes of code which label the dimensions and then they clear up all the artifacts of this usage afterwards. So since this is highly controllable usage -- for now this is enough.

praasz (approved):
Consider update PR description to provide purpose o these changes.

PR #18534, opened 2023-07-13
[PT FE] Support bfloat16 constants

mvafin:
### Details:
 - *Support bfloat16 constants*

### Tickets:
 - *110820*


src/bindings/python/src/openvino/frontend/pytorch/decoder.py
```diff
@@ -153,7 +159,6 @@ def prepare_example_inputs(inputs, input_signature):
                     inputs = ordered_inputs
             if isinstance(inputs, torch.Tensor):
                 inputs = [inputs]
```
mvafin, line 161:
```suggestion
                inputs = [inputs]

```

PR #18535, opened 2023-07-13
[GPU] Add debug config for disabled async compilation

andrew-k-park:
### Details:
 - Add new debug config(OV_GPU_DisableAsyncCompilation) to disable async compilation
 - Add debug log for layout calculation

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -541,6 +541,13 @@ bool primitive_inst::update_impl() {
         }
         if (!cached_impl) {
             if (_dynamic_impl) {
+                auto use_async_compilation = [&]() {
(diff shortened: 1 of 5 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -649,6 +651,14 @@ bool primitive_inst::has_inner_networks() const {
     return (_impl_params->inner_nets.size() > 0);
 }
 
+bool primitive_inst::use_async_compilation() const {
```
yeonbok, line 654: I think this need not be a public member function of primitive_inst. Could  you add this as a lambda function above?

andrew-k-park: applied as comment

PR #18553, opened 2023-07-14
Update selection logic for 3rd part devices within AUTO plugin.

yangwang201911:
### Details:
 - Enable 3rd part devices support for selection logic of AUTO.

### Tickets:
 - 114776


yangwang201911:
Incorrect source of this branch.

PR #18557, opened 2023-07-14
[AUTO] Clean up the logic of AUTO device priorities

yangwang201911:
### Details:
 - Update default candidate list priority: dGPU > iGPU > 3rd part/other devices> CPU
 
### Tickets:
 - 114766


src/plugins/auto/src/plugin.cpp
```diff
@@ -624,8 +627,8 @@ std::list<DeviceInformation> Plugin::get_valid_device(
         devices.splice(devices.end(), VPUX);
     }
     devices.splice(devices.end(), iGPU);
-    devices.splice(devices.end(), MYRIAD);
     devices.splice(devices.end(), CPU);
+    devices.splice(devices.end(), Others);
```
songbell, line 631: where do we place GNA?

src/plugins/auto/src/plugin.cpp
```diff
@@ -575,91 +575,72 @@ std::list<DeviceInformation> Plugin::get_valid_device(
...
-        }
-        if (item.device_name.find("MYRIAD") == 0) {
-            MYRIAD.push_back(item);
-            continue;
-        }
-        if (item.device_name.find("VPUX") == 0) {
-            VPUX.push_back(item);
+    for (auto&& device_info : meta_devices) {
+        if (device_info.device_priority > 0)
+            is_default_list = false;
```
songbell, line 595: we don't need to judge this flag for each meta_devices, once is enough?

songbell: btw, why do we need to know if is_default_list? default or not default, we already know from device_priority, isn't it?

yangwang201911: Let's keep this logic as discussed in the meeting.

src/plugins/auto/src/plugin.cpp
```diff
@@ -575,91 +575,73 @@ std::list<DeviceInformation> Plugin::get_valid_device(
...
-            MYRIAD.push_back(item);
-            continue;
-        }
-        if (item.device_name.find("VPU") == 0) {
-            VPU.push_back(item);
+    for (auto&& device_info : meta_devices) {
+        if (device_info.device_priority > 0)
+            is_default_list = false;
+        // check if device support this model precision
+        if (!is_supported_model(device_info.device_name) && meta_devices.size() > 1)
```
peterchen-intel, line 597:
```suggestion
        // Add device to valid_devices if it is the only device in meta_devices, no matter it supports required precision or not.
        if (!is_supported_model(device_info.device_name) && meta_devices.size() > 1)
```
Do I understand correctly?

src/plugins/auto/src/plugin.cpp
```diff
@@ -575,91 +575,73 @@ std::list<DeviceInformation> Plugin::get_valid_device(
...
-                LOG_DEBUG_TAG("get property :%s for %s failed ", "DEVICE_TYPE", item.device_name.c_str());
+                LOG_DEBUG_TAG("get property :%s for %s failed ", "DEVICE_TYPE", device_info.device_name.c_str());
             }
             if (device_type == "integrated") {
-                iGPU.push_back(item);
+                iGPU.push_back(device_info);
             } else if (device_type == "discrete") {
-                dGPU.push_back(item);
+                dGPU.push_back(device_info);
             } else {
```
songbell, line 614: all these dGPU,iGPU vectors seems no use when is_default_list is false?

songbell (approved):

src/plugins/auto/src/plugin.cpp
yangwang201911: Yes,  It's convenient to kept this logic that saved the device info into these dGPU, iGPU, CPU vectors during the iteration. Eg, `meta_devices ` may look like as below. 
`{
{"GPU.0",..,0}, //device priority is 0
{"GPU.1",..,0},//device priority is 0
{"GPU.2",..,0},//device priority is 0
{"GPU.3",..,0},//device priority is 0
{"CPU",..,1},//device priority is 1
}`

wangleis (approved):

PR #18429, opened 2023-07-07
Hetero 2.0

vurusovs:
### Details:
 - Combination of 
   - https://github.com/openvinotoolkit/openvino/pull/18392
   - https://github.com/openvinotoolkit/openvino/pull/17814
   - https://github.com/openvinotoolkit/openvino/pull/17605/

### Tickets:
 - CVS-109770

PR #18559, opened 2023-07-14
Hetero 2.0 leftovers

vurusovs:
### Details:
 - Several changes after https://github.com/openvinotoolkit/openvino/pull/18429

### Tickets:
 - CVS-109770

### DIFF
--- a/src/plugins/hetero/src/compiled_model.cpp
+++ b/src/plugins/hetero/src/compiled_model.cpp
@@ -401,7 +401,6 @@ ov::hetero::CompiledModel::CompiledModel(const std::shared_ptr<ov::Model>& model
                                                        subgraph._sinks,
                                                        subgraph._parameters,
                                                        m_name + '_' + std::to_string(id));
-        m_compiled_submodels[id].model = subFunctions[id]->clone();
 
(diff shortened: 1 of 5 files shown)


nkogteva (approved):

src/plugins/hetero/src/compiled_model.cpp
```diff
@@ -411,7 +409,7 @@ ov::hetero::CompiledModel::CompiledModel(const std::shared_ptr<ov::Model>& model
         auto device_config = metaDevices[m_compiled_submodels[id].device];
         device_config[ov::cache_dir.name()] = "";
 
-        m_compiled_submodels[id].compiled_model = plugin->get_core()->compile_model(m_compiled_submodels[id].model,
+        m_compiled_submodels[id].compiled_model = plugin->get_core()->compile_model(subFunctions[id]->clone(),
```
ilyachur, line 412: Why do you need to clone the model here? I think we will have another copy inside the plugin compile model

vurusovs: Done in https://github.com/openvinotoolkit/openvino/pull/18576

PR #18612, opened 2023-07-18
Revert "Remove extra model from CompiledModelDesc" 

vurusovs:
### Details:
 - Model returned by `get_runtime_model()` not useful for import/export


nkogteva (commented):
Does it working? It seems we have no tests which cover this scenario

vurusovs:
> Does it working? It seems we have no tests which cover this scenario

Yes, it works. No, we have `NoCacheMetric` mocked tests that validates it. The only problem is that mocked `get_runtime_model()` returns correct ov::Model which can be read by test

nkogteva (approved):

PR #18622, opened 2023-07-18
SharedOpOptimization

jane-intel:
### Details:
 - *Optimizes shared operations: Slice, Tile, Reshape, Concat, GatherElements*

DLB was triggered to get performance numbers and to prevent unexpected backward incompatibilities
UPD: DLB couldn't compile due to public headers of myriad were deleted from new api.

### Tickets:
 - *Enhances performance of several LLMs and other models*


src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -0,0 +1,142 @@
...
+        if (lhs_constant->get_shape() != rhs_constant->get_shape() ||
+            ov::shape_size(lhs_constant->get_shape()) > 10)  // protection for not comparing huge constants
+            return false;
+        if (lhs_constant->get_element_type() != rhs_constant->get_element_type())
+            return false;
+        if (lhs_constant->get_element_type().is_integral()) {
+            if (lhs_constant->cast_vector<int64_t>() != rhs_constant->cast_vector<int64_t>())
+                return false;
+        } else if (lhs_constant->get_element_type().is_real()) {
+            if (lhs_constant->cast_vector<double>() != rhs_constant->cast_vector<double>())
```
t-jankowski, line 83: Since type equality is asserted these `cast`s bring no value. Look at https://github.com/openvinotoolkit/openvino/pull/17694/files#diff-e6fd768fedc8b143cfb95dfe0eefd599ed0cc5e3455767955d64707819bc7ef4

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+        auto slice_0 = make_slice(data, 1, 2, 3, 3);
+        auto slice_2 = make_slice(data, 1, 3, 3, 3);
+
+        auto concat = std::make_shared<v0::Concat>(OutputVector{slice_0, slice_0, slice_2, slice_0, slice_0}, 0);
+
+        auto result = std::make_shared<v0::Result>(concat);
+        model_ref = std::make_shared<ov::Model>(ResultVector{result}, ParameterVector{data});
+    }
+}
```
t-jankowski, line 56: It'd be good to also test:
- mixed op type nodes with the same source(s),
- `Result` producers are not replaced while the others are,
- multi sources in case of concat.

t-jankowski (changes requested):
Look at CVS-105138 and [PR 17694](https://github.com/openvinotoolkit/openvino/pull/17694)

jane-intel:
I remove do_not_merge since this PR doesn't introduce degradation.

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -53,4 +55,181 @@ TEST_F(TransformationTestsF, SharedSlice) {
...
+        auto data = std::make_shared<v0::Parameter>(element::f32, PartialShape{-1});
+        auto post_constant = v0::Constant::create(element::f32, Shape{1}, std::vector<float>{3.14});
+
+        auto concat_0 = std::make_shared<v0::Concat>(OutputVector{pre_constant_0, data, post_constant}, 0);
+        auto concat_1 = std::make_shared<v0::Concat>(OutputVector{pre_constant_1, data, post_constant}, 0);
+
+        auto result_0 = std::make_shared<v0::Result>(concat_0);
+        auto result_1 = std::make_shared<v0::Result>(concat_1);
+        model_ref = std::make_shared<ov::Model>(ResultVector{result_0, result_1}, ParameterVector{data});
+    }
```
t-jankowski, line 234: This is not needed, [reference model is cloned if not defined](https://github.com/openvinotoolkit/openvino/blob/f470825fafb2a298172de84ef4485184f0c7f167/src/tests/test_utils/common_test_utils/src/ngraph_test_utils.cpp#L61-L63).

t-jankowski (changes requested):

t-jankowski (approved):

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+        auto transpose_0_0 = make_transpose(slice_0, {0, 2, 3, 1});
+
+        auto tile_2_0 = make_tile(slice_2, {1, 2, 3, 4});
+        auto transpose_2_0 = make_transpose(slice_2, {0, 2, 3, 1});
+
+        auto concat = std::make_shared<v0::Concat>(
+            OutputVector{// source from slice 0
+                         tile_0_0,
+                         transpose_0_0,
+                         tile_0_0,
```
slyalin, line 158: Is it good to have the same outputs duplicated with multiple Result nodes.

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -0,0 +1,135 @@
...
+        auto rhs_constant = get_constant_from_source(rhs->input_value(i));
+        OPENVINO_SUPPRESS_DEPRECATED_END
+        if (!lhs_constant || !rhs_constant)
+            return false;
+        if (lhs_constant->get_element_type() != rhs_constant->get_element_type())
+            return false;
+        const auto& lhs_shape = lhs_constant->get_shape();
+        if (lhs_shape != rhs_constant->get_shape() || shape_size(lhs_shape) > 10)
+            return false;
+        size_t bytes = (shape_size(lhs_shape) * lhs_constant->get_element_type().bitwidth() + 8 - 1) / 8;
```
slyalin, line 78: Constant op has `get_byte_size` method. Doesn't it work in this case?

slyalin (approved):

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -0,0 +1,145 @@
...
+using namespace ov;
+using namespace ov::op;
+
+bool shared_node_optimization(const shared_ptr<Model>& model,
+                              const unordered_map<Node::type_info_t, bool (*)(const Node*, const Node*)>& rules) {
+    bool rewritten = false;
+
+    for (const auto& op : model->get_ordered_ops()) {
+        // Recursively apply transformation for sub-graph based operations
+        if (auto sub_graph_node = dynamic_pointer_cast<op::util::SubGraphOp>(op)) {
```
itikhono, line 27: what about MultiSubGraphOp (If)?

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -0,0 +1,142 @@
...
+        if (lhs->input_value(i) == rhs->input_value(i))
+            continue;
+        auto lhs_constant = as_type_ptr<v0::Constant>(lhs->get_input_node_shared_ptr(i));
+        auto rhs_constant = as_type_ptr<v0::Constant>(rhs->get_input_node_shared_ptr(i));
+        if (!lhs_constant || !rhs_constant)
+            return false;
+        if (lhs_constant->get_element_type() != rhs_constant->get_element_type())
+            return false;
+        const auto& lhs_shape = lhs_constant->get_shape();
+        if (lhs_shape != rhs_constant->get_shape() || shape_size(lhs_shape) > 10)
```
itikhono, line 85: why > 10?
could you add comment here?

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -0,0 +1,142 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+    return rewritten;
+}
+
+bool inputs_from_same_source_or_equal_constants(const Node* lhs, const Node* rhs) {
```
itikhono, line 71: do we have a constant comparator function somewhere in common place? op::utils?
Should we add ==operator to Constant class?

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+using namespace std;
+using namespace ov;
+using namespace ov::op;
+
+bool shared_node_optimization(const shared_ptr<Model>& model,
+                              const unordered_map<Node::type_info_t, bool (*)(const Node*, const Node*)>& rules) {
+    bool rewritten = false;
+
+    for (const auto& op : model->get_ordered_ops()) {
```
itikhono, line 24: If I remember correctly, we decided to change the order of the traversal for this transformation (starting from Parameters and Constant and moving to Results and Sinks direction)

@t-jankowski could you remind us what was the reason for that?

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
t-jankowski: It's not there yet. I made it common, but it's in on-hold PR #17694:
[src/common/transformations/include/transformations/utils/utils.hpp](https://github.com/openvinotoolkit/openvino/pull/17694/files#diff-3919034fc619fa9085663dfb6fbf07c63946f27a4d916d3a4fb49356d3e9b9c3R226)
[src/common/transformations/src/transformations/utils/utils.cpp](https://github.com/openvinotoolkit/openvino/pull/17694/files#diff-e6fd768fedc8b143cfb95dfe0eefd599ed0cc5e3455767955d64707819bc7ef4R345-R357)

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
t-jankowski: _could you remind us what was the reason for that?_
MSB transformation from PR #17694 removes (by merging or fusing) nodes from the graph and only after gathers next nodes top-down for further scanning. It's to avoid running multiple scanners on dangling node.
That's the case here indeed, but rather harmless - [if (target_inputs.size() <= 1) continue;](https://github.com/openvinotoolkit/openvino/pull/18622/files#diff-cff48c943726972afbe6e7ebeccc97205c793d1367845d2e3af3765327f0ada1R34-R35).

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
jane-intel: I can add the comment later -- right now green pre-commit is rare

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
jane-intel: Since this isn't available yet -- I resolve the comment

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
jane-intel: My logic was designed with current topological order -- no need to alter it in any way.

itikhono (approved):

PR #18625, opened 2023-07-18
[GPU] Dynamism support for DetectionOutput

andrew-k-park:
### Details:
 - Updates for DetectionOutput v8 support
 - Update DetectionOutput to use ngraph shape inference
 - Fix issues for detection output cpu impl selection on dynamic shape case
 - Add dynamic and shape inference TCs for ov_gpu_func_tests and ov_gpu_unit_tests

### Tickets:
 - 109741


src/plugins/intel_gpu/src/plugin/ops/detection_output.cpp
```diff
@@ -27,14 +27,13 @@ static cldnn::prior_box_code_type PriorBoxCodeFromString(const std::string& str)
     return cldnn::prior_box_code_type::corner;
 }
 
-static void CreateDetectionOutputOp(Program& p, const std::shared_ptr<ngraph::op::v0::DetectionOutput>& op) {
-    validate_inputs_count(op, {3});
+static void CreateCommonDetectionOutputOp(Program& p,
+                                          const std::shared_ptr<ngraph::Node>& op,
+                                          const ov::op::util::DetectionOutputBase::AttributesBase& attrs,
+                                          int num_classes) {
     auto inputs = p.GetInputInfo(op);
```
yeonbok, line 34: Could you add an exception message if there are optional inputs that we are not support yet? 
E.g., if there are optional inputs (4/5th) :
"[GPU] Optional 4th and 5th inputs are not supported by GPU plugin yet.

andrew-k-park: `validate_inputs_count` already handle exceptions, but does that mean separate exception message should be handled?

yeonbok (approved):

PR #18637, opened 2023-07-19
[GPU] Add per iteration performance profiling mode

sshlyapn:
### Details:
 - Added per iteration performance profiling mode via `OV_GPU_DumpProfilingDataPerIter` environment variable
 - Replaced memory buffer's layout with actual primitive's layout in log_memory_to_file() function

### Tickets:
 - *ticket-id*


yeonbok:
I think this is good to have, but I'd like to suggest some change on the behavior: Currently the data is accumulated in network and flushed at the end of the execution. However if we run big models those data becomes big too, so such data needed to be flushed if possible. So how about throw away those data at the end of each iteration if we run with per iter mode?

p-durandin (approved):

PR #18656, opened 2023-07-19
New save_model function (C++, Python)

slyalin:
A new method `save_model` is introduced as a simpler alternative to `ov::serialize` and better aligned with new `ovc.convert_model`. There are two differences from `ov::serialize`:

1. FP16 compression of weights is applied by default to output IR.
2. There is no bin file name parameter, it is always formed from xml file name.

This function is designed to be a complementary part for a new version of `convert_model` which means that save_model(convert_model(model)) is identical to `ovc` call.

TODO: More efficient fp16 compression should be implemented on-the-fly when data is serialized to a file, not as a separate pass that requires extra memory allocation and model clone as it is implemented now. No better alternative is available in OV at this moment.


src/core/src/graph_util.cpp
```diff
@@ -847,3 +850,15 @@ void ov::serialize(const std::shared_ptr<const ov::Model>& m,
...
+
+void ov::save_model(const std::shared_ptr<const ov::Model>& m, const std::string& xml_path, bool compress_to_fp16) {
+    ov::pass::Manager manager;
+    if (compress_to_fp16) {
+        manager.register_pass<ov::pass::MarkPrecisionSensitiveConstants>();
+        manager.register_pass<ov::pass::CompressFloatConstants>();
+    }
+    manager.register_pass<ov::pass::FusedNamesCleanup>();
+    manager.register_pass<ov::pass::Serialize>(xml_path, "");
+    auto cloned = ov::clone_model(*m);  // TODO: Implement on-the-fly compression in pass::Serialize
```
slyalin, line 862: Replace by `m->clone()`.

src/core/src/graph_util.cpp
```diff
@@ -847,3 +850,15 @@ void ov::serialize(const std::shared_ptr<const ov::Model>& m,
...
+    manager.run_passes(std::const_pointer_cast<ov::Model>(cloned));
```
slyalin, line 863: `const_pointer_cast` is not needed.

ilyachur (approved):

jiwaszki (approved):
LGTM, @akuporos can you take a second look?

src/bindings/python/src/pyopenvino/pyopenvino.cpp
```diff
@@ -157,6 +160,38 @@ PYBIND11_MODULE(_pyopenvino, m) {
...
+        "save_model",
+        [](std::shared_ptr<ov::Model>& model,
+           const py::object& xml_path,
+           bool compress_to_fp16) {
+            ov::save_model(model,
+                          Common::utils::convert_path_to_string(xml_path),
+                          compress_to_fp16);
+        },
+        py::arg("model"),
+        py::arg("output_model"),
```
akuporos, line 173: Wouldn't it better to name this argument `xml_path` to be aligned with `serialize`?
Or `output_path` -- it's more self-explainable

CC @jiwaszki

slyalin: This new naming is explained by:
1. Align with a similar parameter in new `ovc` tool, this is also `output_model` that is complimentary to `input_model`.
2. Hiding details about IR implementation: `xml`. We can change xml to something else in the future when go away from xml format.

slyalin: We can adjust naming in a separate PR. Will collect more feedback about it. But even now it has "story" underneath as explained.

PR #18658, opened 2023-07-20
enable auto handle GPU.invalid_index case

songbell:
### Details:
 - for case like AUTO:GPU.100,GPU.0, give auto the authority to select a valid device

### Tickets:
 - *CVS-116252*


ilyachur:
Am I right that this PR fixes CVS-115307?

src/plugins/proxy/src/plugin.cpp
```diff
@@ -451,7 +451,8 @@ std::string ov::proxy::Plugin::get_primary_device(size_t idx) const {
     if (devices.empty())
         // Return low level device name in case of no devices wasn't found
         return m_device_order.at(0);
-    OPENVINO_ASSERT(devices.size() > idx);
+    if (devices.size() <= idx)
+        OPENVINO_THROW("devie with index", idx, " wasn't found!");
```
ilyachur, line 455: Typo in the message: devie -> device
It can be:
```suggestion
    OPENVINO_ASSERT(devices.size() > idx, "Device with the  index", idx, " wasn't found!");
```

src/plugins/auto/src/plugin.cpp
```diff
@@ -248,7 +250,11 @@ std::vector<DeviceInformation> Plugin::parse_meta_devices(const std::string& pri
                     full_device_name = get_core()->get_property(device_name_with_id, ov::device::full_name);
                 } catch (ov::Exception&) {
                     LOG_DEBUG_TAG("get full device name failed for ", device_name_with_id.c_str());
+                OPENVINO_SUPPRESS_DEPRECATED_START
+                } catch (InferenceEngine::Exception&) {
```
ilyachur, line 254: Why do we need old exception here?

songbell:
> Am I right that this PR fixes [CVS-115307](https://jira.devtools.intel.com/browse/CVS-115307)?
not sure., the error msg looks same... but I need to understand the device name in the jira

src/plugins/auto/src/plugin.cpp
songbell: ilya, I just tried to align to ov::exception in plugin calls, but seems introduce several case failures in different components,  so shall we merge this first? I can try in another PR to fix the exceptions below:
https://github.com/openvinotoolkit/openvino/blob/master/src/inference/src/dev/plugin.cpp#L18

wangleis (approved):

src/plugins/auto/src/plugin.cpp
ilyachur: Yes let's do it separatelly.

ilyachur (approved):

PR #18172, opened 2023-06-21
[AUTO BATCH PLUGIN] enable api 2.0 for auto batch plugin

zhaixuejun1993:
### Details:
 - enable api 2.0 for auto batch plugin
 - migrate unit test of auto batch plugin api 2.0

### Tickets:
 - 111940

PR #18660, opened 2023-07-20
[AUTO BATCH PLUGIN] fix review comments

zhaixuejun1993:
### Details:
 - Add api quary_status()
 - Fix comments of 18172

### Tickets:
 - 115955


src/plugins/auto_batch/src/compiled_model.cpp
```diff
@@ -217,7 +211,23 @@ ov::Any CompiledModel::get_property(const std::string& name) const {
             return m_compiled_model_without_batch->get_property(name);
         } else if (name == ov::loaded_from_cache) {
             return m_compiled_model_without_batch->get_property(ov::loaded_from_cache.name());
+        } else if (name == ov::supported_properties) {
+            return std::vector<ov::PropertyName>{
```
riverlijunjie, line 215:
```suggestion
            return std::vector<std::string>{
                       ov::optimal_number_of_infer_requests.name(), 
                       ...
```

src/plugins/auto_batch/src/async_infer_request.cpp
```diff
@@ -54,6 +54,14 @@ std::vector<ov::ProfilingInfo> AsyncInferRequest::get_profiling_info() const {
         return m_request_without_batch->get_profiling_info();
 }
 
+std::vector<ov::SoPtr<ov::IVariableState>> AsyncInferRequest::query_state() const {
+    check_state();
+    if (SyncInferRequest::eExecutionFlavor::BATCH_EXECUTED == m_sync_request->m_batched_request_status)
```
riverlijunjie, line 59: Is there thread race condition risk? because m_sync_request->m_batched_request_status maybe updated in a separated thread.

zhaixuejun1993: I think no. Before query, it will be locked at:
https://github.com/openvinotoolkit/openvino/blob/b2d361fcca75d419b15e55b8db3fb34eb3fc7e4c/src/inference/src/dev/iasync_infer_request.cpp#L192-L202

src/plugins/auto_batch/src/compiled_model.cpp
```diff
@@ -59,8 +59,9 @@ CompiledModel::GetWorkerInferRequest() const {
     if (!batch_id) {  // need new request
         m_worker_requests.push_back(std::make_shared<WorkerInferRequest>());
         auto workerRequestPtr = m_worker_requests.back().get();
-        workerRequestPtr->_infer_request_batched = {m_compiled_model_with_batch->create_infer_request(),
-                                                    m_compiled_model_with_batch._so};
+        workerRequestPtr->_infer_request_batched._ptr = m_compiled_model_with_batch->create_infer_request();
+        if (workerRequestPtr->_infer_request_batched._so == nullptr)
```
songbell, line 63: won't it always be nullptr? seem no one assign value to it

zhaixuejun1993: can saw the discussion:
https://github.com/openvinotoolkit/openvino/pull/18172#discussion_r1267531918

src/plugins/auto_batch/src/async_infer_request.cpp
ilyachur: This mutex will be unlocked after return from check_state() method

src/plugins/auto_batch/src/compiled_model.cpp
```diff
@@ -179,13 +180,6 @@ ov::Any CompiledModel::get_property(const std::string& name) const {
     if (it != m_config.end()) {
         return it->second;
     } else {
-        // find config key among networks config keys
-        auto modelSupportedProperties = m_compiled_model_without_batch->get_property(ov::supported_properties.name());
```
ilyachur, line 183: Why did you remove this code?

zhaixuejun1993: Didn't remove this, just move this to:
https://github.com/openvinotoolkit/openvino/blob/0555a492313a666c1a7d7e8cef4c0db26407ccdc/src/plugins/auto_batch/src/compiled_model.cpp#L222-L230
The original is a bug, when call get_property for `ov::supported_properties`, which will return the properties from HW plugin compiler model not for auto-batch plugin compiler model. So add the following code:
https://github.com/openvinotoolkit/openvino/blob/0555a492313a666c1a7d7e8cef4c0db26407ccdc/src/plugins/auto_batch/src/compiled_model.cpp#L214-L222

riverlijunjie (approved):
LGTM

src/plugins/auto_batch/src/compiled_model.cpp
```diff
@@ -217,7 +211,23 @@ ov::Any CompiledModel::get_property(const std::string& name) const {
...
+                ov::PropertyName{ov::auto_batch_timeout.name(), ov::PropertyMutability::RO}};
         } else {
+            // find config key among networks config keys
+            auto modelSupportedProperties =
+                m_compiled_model_without_batch->get_property(ov::supported_properties.name());
+            for (auto&& property : modelSupportedProperties.as<std::vector<ov::PropertyName>>()) {
+                if (property == name) {
+                    return m_compiled_model_without_batch->get_property(property);
+                }
+            }
```
songbell, line 230: this pr does not solve timeout issue right? e.g
batch_compiled_model.set_property(auto_batch_timeout(200))
then : batch_compiled_model.get_property(auto_batch_timeout)
do you plan in a seperate PR?

src/plugins/auto_batch/src/async_infer_request.cpp
songbell: shall be no race issue here from my point of view
only that the profiling info /state status will not be the data for the batched infer request instead of individual request.
better to understand the usage scenario to decide if we need to match the batched profilinginfo /status back to individual profiling info/ state info.

src/plugins/auto_batch/src/compiled_model.cpp
```diff
@@ -217,7 +211,23 @@ ov::Any CompiledModel::get_property(const std::string& name) const {
...
+            return std::vector<ov::PropertyName>{
+                ov::PropertyName{ov::optimal_number_of_infer_requests.name(), ov::PropertyMutability::RO},
+                ov::PropertyName{METRIC_KEY(SUPPORTED_METRICS), ov::PropertyMutability::RO},
+                ov::PropertyName{ov::model_name.name(), ov::PropertyMutability::RO},
+                ov::PropertyName{METRIC_KEY(SUPPORTED_CONFIG_KEYS), ov::PropertyMutability::RO},
+                ov::PropertyName{ov::execution_devices.name(), ov::PropertyMutability::RO},
+                ov::PropertyName{ov::auto_batch_timeout.name(), ov::PropertyMutability::RO}};
         } else {
+            // find config key among networks config keys
+            auto modelSupportedProperties =
```
songbell, line 224: em... won't it be better that we support ov::device::property here? then we can get more intuitive parameters from underlying compiled_model parameters

ilyachur (approved):

PR #18675, opened 2023-07-20
[ShapeInference] Additional check for the shape of DetectionOutput input

mitruska:
### Details:
 - Additional check for the shape of DetectionOutput input to catch the incompatible shape before plugin check
 https://github.com/openvinotoolkit/openvino/blob/25b0013b8f1587190a1ef041b559f01b92b3e058/src/plugins/intel_cpu/src/nodes/detection_output.cpp#L104
 

### Tickets:
 - 88466


src/core/shape_inference/include/detection_output_shape_inference.hpp
```diff
@@ -168,6 +168,20 @@ void shape_infer_base(const DetectionOutputBase* op,
                                   ".");
             num_prior_boxes = box_logits_pshape_2nd_dim / (num_loc_classes * 4);
         }
+
+        if (num_prior_boxes > 0 && num_loc_classes > 0) {
+            NODE_VALIDATION_CHECK(op,
```
praasz, line 173: Please refactor this validation to reduce number arguments. generating this kind of validation has negative impact on binary size.
e.g:
the `but must be compatible with (num_prior_boxes * num_loc_classes * 4) = (",`, can be removed  as check `box_logits_pshape[1].compatible(num_prior_boxes * num_loc_classes * 4)` will converted to string.

Instead interupt string with values try make "explanation", current value, expect value etc.

` box_logits_pshape[1],` printing it could be replaced by new macro `NODE_SHAPE_INFER_CHECK`

src/core/tests/type_prop/detection_output.cpp
```diff
@@ -1143,3 +1143,143 @@ TEST(type_prop_layers, detection_output_v8_default_ctor) {
...
+        EXPECT_HAS_SUBSTRING(error.what(),
+                             std::string("The second dimension of the first input (box logits) is not compatible. "
+                                         "Current value: 384, expected value: 512"));
+    } catch (...) {
+        FAIL() << "Unknown exception was thrown";
+    }
+}
+
+TEST(type_prop_layers, detection_output_v8_incompatible_dynamic_num_prior_boxes_normalized_true_shareloc_true) {
+    try {
```
praasz, line 1236: Consider use `OV_EXPECT_THROW` instead `try..catch`,

mitruska: Ok, done

src/core/shape_inference/include/detection_output_shape_inference.hpp
mitruska: Ok, I wanted this error message to be as much informative as possible, but I have shortened it.

jane-intel (approved):

praasz (approved):

PR #18678, opened 2023-07-20
Align mmap offset with page size for Linux

mbencer:
Ticket: CVS-115745


src/common/util/src/os/lin/lin_mmap_object.cpp
```diff
@@ -79,11 +79,19 @@ class MapHolder : public MappedMemory {
...
-            m_data = mmap(nullptr, m_size, prot, MAP_PRIVATE, m_handle.get(), offset);
+        const size_t page_size = sysconf(_SC_PAGE_SIZE);
+        // align offset with page size
+        const size_t aligned_offset = (offset / page_size) * page_size;
+        const size_t aligned_size = (offset % page_size) + m_size;
+        // move pointer to the expected data (after alignment with page size)
+        const size_t offset_delta = offset - aligned_offset;
+
+        if (aligned_size > 0) {
+            m_data = mmap(nullptr, aligned_size, prot, MAP_PRIVATE, m_handle.get(), aligned_offset);
```
vurusovs, line 90: According to mmap man:
1. `For a file that is not a multiple of the page size, the remaining bytes in the partial page at the end of the mapping are zeroed when mapped, and modifications to that region are not written out to the file`
2. `offset must be a multiple of the page size as returned by sysconf(_SC_PAGE_SIZE).`

Looks like the issue is in offset only. Let's try to leave size as it was before

vurusovs (changes requested):

src/common/util/src/os/lin/lin_mmap_object.cpp
mbencer: @vurusovs I think, that aligning size is needed, but I had missing unit test to confirm it.
Let's consider case, where we want to mmap memory on the border of two page size, like:
image
In such a case we have:
- `aligned_offset` = `0`
- `m_size` = `16`
so we are mmaping `Page 1` and have access only to 8 bytes (from Page 1).

In this case `aligned_size` will be `4104`, so mmap will be applied on `Page 1` and `Page 2` (expected place in the `m_data` will be determined by `offset_delta`

vurusovs (approved):

PR #18679, opened 2023-07-20
[shape infer]Review Reshape class for shape inference aspects

praasz:
### Details:
 - Check static and dynamic dimension and label propagation
 - Review template implementation of `shape_infer` for reshape.
 - Expand testing

### Tickets:
 - 90592
 - 90689
 - 90726
 - 90763
 - 90800
 - 90837
 - 90874
 - 90911


src/core/shape_inference/include/reshape_shape_inference.hpp
```diff
@@ -0,0 +1,283 @@
...
+            auto label_iter = labels.begin();
+            for (auto& d : shape) {
+                if (*label_iter != no_label) {
+                    DimensionTracker::set_label(d, *label_iter);
+                }
+                ++label_iter;
+            }
+        }
+    }
+}
```
jane-intel, line 169: What if value 0 or -1 has a label -- we don't want this label to override labels that we deduced from input shape

praasz: What I tested the label should not be overwritten, but if you got some special cases I will test them and check if updated implementation handles it correctly.

jane-intel: Pseudo code:
```
input = Parameter([1, 1, 100]); // label Dimension 100 with label 1

pattern = Parameter([1]);
pattern.output(0).get_tensor().set_value_labels({2});
tensor_m1 = Tensor({-1});
pattern.output(0).get_tensor().set_lower_bound(tensor_m1);
pattern.output(0).get_tensor().set_upper_bound(tensor_m1);

test = Reshape(input, pattern, false);
ASSERT(DimensionTracker::get_label(test.get_output_partial_shape(0)[0]) == 2);
```

You can achieve the same with zero, but the outcome of labels resolving would depend on special_zero attribute of Reshape operation.

jane-intel (commented):
I find code state very good. I only have one question / concern.

src/core/shape_inference/include/reshape_shape_inference.hpp
praasz: Updated minus one dimension calculation and label propagation.
See test groups with for details:
- `reshape_label_propagation_minus_one_no_special_zero_case_X`
- `reshape_label_propagation_minus_one_special_zero_case_X`
- `reshape_tricky_label_propagation_for_auto_batch_case_X`

The tests pass also on master version except two cases `reshape_label_propagation_minus_one_special_zero_case_6` and `reshape_tricky_label_propagation_for_auto_batch_case_4`. The difference is that on master the label is not propagated but it can be.

src/core/tests/type_prop/reshape.cpp
```diff
@@ -672,33 +667,550 @@ TEST(type_prop, reshape_dynamic_value_and_label_propagation) {
...
+    auto out_shape = PartialShape{{3, 5}, 0, 1, 0};
+    set_shape_labels(param_shape, 10);
+    set_shape_labels(out_shape, 20);
+
+    const auto data = make_shared<op::v0::Parameter>(element::f32, param_shape);
+    const auto out = make_shared<op::v0::Parameter>(element::f32, out_shape);
+    const auto shape_of = make_shared<op::v3::ShapeOf>(out);
+    auto dims_from_out_shape = make_shared<op::v1::Gather>(shape_of,
+                                                           op::v0::Constant::create(element::i64, {4}, {0, 1, 2, 3}),
+                                                           op::v0::Constant::create(element::i64, {}, {0}));
```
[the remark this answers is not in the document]
praasz: I think is just result of copy-paste. Will replace gather.

src/core/tests/type_prop/reshape.cpp
```diff
@@ -672,33 +667,550 @@ TEST(type_prop, reshape_dynamic_value_and_label_propagation) {
...
+    EXPECT_EQ(r->get_output_partial_shape(0), PartialShape::dynamic());
+}
+
+TEST(type_prop, reshape_when_pattern_has_scalar_shape_only) {
+    auto param = make_shared<op::v0::Parameter>(element::f32, Shape{3, 4});
+    auto shape_pattern = make_shared<op::v0::Parameter>(element::u64, PartialShape{});
+    auto r = make_shared<op::v1::Reshape>(param, shape_pattern, false);
+
+    EXPECT_EQ(r->get_element_type(), element::f32);
+    EXPECT_EQ(r->get_output_partial_shape(0), PartialShape());
```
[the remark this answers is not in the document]
praasz: I think in this test the output rank should be {0,1}, because value of scalar is unknown.
But if scalar value will be constant and 1 then input shape never become sacalar and shape inference can throw exception.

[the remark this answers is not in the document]
praasz: Done.

src/core/shape_inference/include/reshape_shape_inference.hpp
```diff
@@ -0,0 +1,363 @@
...
+
+            NODE_SHAPE_INFER_CHECK(op,
+                                   input_shapes,
+                                   backward_compatible_check || in_out_elements_equal,
+                                   "Requested output shape ",
+                                   output_shape,
+                                   " is incompatible with input shape");
+        }
+    } else if (pattern_shape_rank.is_static()) {
+        std::cout << "temp\n";
```
jane-intel, line 346:
```suggestion
```

src/core/shape_inference/include/reshape_shape_inference.hpp
```diff
@@ -0,0 +1,363 @@
...
+
+    const T& get_dynamic_out() const {
+        return out.second;
+    }
+
+    void calculate() {
+        // dimensions compare to remove same from product calculation
+        auto dim_full_eq = [](const T& lhs, const T& rhs) -> bool {
+            return (lhs == rhs) && DimensionTracker::get_label(lhs) == DimensionTracker::get_label(rhs) &&
+                   (lhs.is_static() || DimensionTracker::has_label(lhs));
```
jane-intel, line 87: two static dims may be equal but with different labels -- shouldn't we handle them as equal anyways?

praasz: Agree that for product calculation the  static shape with different labels has can be skipped. 
The intention is use these static dimensions into product calculation because of label propagation. ( in some cases when resolve label for '-1' the label can be passed and not always lost).

src/core/shape_inference/include/reshape_shape_inference.hpp
```diff
@@ -0,0 +1,360 @@
...
+        }
+    } else if (pattern_shape_rank.is_static()) {
+        if (pattern_shape_rank.get_length() == 0) {
+            NODE_SHAPE_INFER_CHECK(op,
+                                   input_shapes,
+                                   input_rank.compatible(0),
+                                   "Input must be scalar as pattern is scalar!");
+        } else {
+            output_shape =
+                PartialShape::dynamic(Rank(pattern_shape[0].get_min_length(), pattern_shape[0].get_max_length()));
```
jane-intel, line 351: rank is not stored in PartialShape, so you could've coded a bit simpler:
```suggestion
                PartialShape::dynamic(Rank(pattern_shape[0]));
```

jane-intel (approved):

jane-intel:
Please confirm if reshape and dynamism tests have no failures connected to this PR

src/core/shape_inference/include/reshape_shape_inference.hpp
praasz: The `Rank(pattern_shape[0])` cannot be used as there is compile error rank cannot be created from `StaticShape` class.

src/common/transformations/tests/utils/convert_precision.cpp
```diff
@@ -2147,7 +2147,7 @@ TEST(TransformationTests, align_mixed_fp16_fp32_with_parameter_for_shape_1) {
         auto upscale_const = ov::op::v0::Constant::create(element::f32, Shape{1}, {2.0f});
         auto mul_1 = make_shared<ov::op::v1::Multiply>(shape_input, upscale_const);
         auto axis_const = ov::op::v0::Constant::create(element::i64, Shape{1}, {0});
-        auto final_float_shape = make_shared<ov::op::v1::ReduceProd>(mul_1, axis_const);
+        auto final_float_shape = make_shared<ov::op::v1::ReduceProd>(mul_1, axis_const, true);
```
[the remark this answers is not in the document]
praasz: Is related to check in (shape_infer)[https://github.com/openvinotoolkit/openvino/blob/380b86b756c058ac29677d4a6ec94254c5e402b7/src/core/shape_inference/include/reshape_shape_inference.hpp#L336]
The `final_float_shape ` after reduction become scalar and `convert_to_f16` is 1D and this check will throw exception that input must be also scalar.

So changed to keep dimension and `final_float_shape` is 1D, unless check in reshape is not correct.

jane-intel:
Enabling auto-merge per confirmation with @p-durandin

PR #18692, opened 2023-07-20
[PT FE] Add helper for regular ops on quantized values

mvafin:
### Details:
 - *Add helper for regular ops on quantized values*

### Tickets:
 - *None*


eaidova (approved):

src/frontends/pytorch/src/op_table.cpp
```diff
@@ -376,13 +377,13 @@ const std::map<std::string, CreatorFunction> get_supported_ops() {
         {"aten::sin_", op::inplace_op<op::translate_1to1_match_1_inputs<opset10::Sin>>},
         {"aten::sinh", op::translate_1to1_match_1_inputs_with_fp32_type_alignment<opset10::Sinh>},
         {"aten::sinh_", op::inplace_op<op::translate_1to1_match_1_inputs<opset10::Sinh>>},
-        {"aten::size", op::translate_size},
-        {"aten::slice", op::translate_slice},
+        {"aten::size", op::quantizable_op<op::translate_size>},
```
mmikolajcz, line 380: I think this change for aten::size can cause problems in models. Adding FakeQuantize would change return dtype of size from integer to float, and later, cause mixed dtypes when it is merged into ListConstruct with other int to be an argument of aten::view

mmikolajcz: I think similar case could be for aten::len and comparison operators returning boolean but didn't checked that in models

PR #18699, opened 2023-07-21
[dGPU] Enable user scratchpad mode for onednn primitive.

jade-cho:
+ Set scratchpad mode to dnnl::scratchpad_mode::user to manage memory in GPU Plugin.

### Tickets:
 - *96507*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
+++ b/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
@@ -470,6 +470,7 @@ protected:
         }
 
         if (_scratchpad_md.get_size() != 0) {
+            // onednn primitive can have only 1 scratchpad memory.
             auto scratchpad = instance.get_intermediates_memories()[0];
             args.insert({DNNL_ARG_SCRATCHPAD, scratchpad->get_onednn_memory(_scratchpad_md, 0)});
         }
--- a/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
+++ b/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
@@ -41,6 +41,7 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
     PrimDescType _pd;
     PrimType _prim;
     std::unordered_map<uint32_t, std::unordered_map<int, dnnl::memory>> _args;
+    dnnl::memory::desc _scratchpad_md;
     bool _enable_profiling = false;
 
     typed_primitive_onednn_impl(const engine& engine,
@@ -53,6 +54,24 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
         _attrs(attrs),
         _pd(pd) {
             _enable_profiling = config.get_property(ov::enable_profiling);
+
+            _scratchpad_md = _pd.scratchpad_desc();
+
+            GPU_DEBUG_GET_INSTANCE(debug_config);
+            GPU_DEBUG_IF(!debug_config->dump_profiling_data.empty()) {
+                _enable_profiling = true;
+            }
+
+            GPU_DEBUG_IF(debug_config->verbose >= 5) {
+                if (_scratchpad_md.get_size() > 0) {
+                    static long long total = 0;
+                    int64_t size = _scratchpad_md.get_size() / 1048576;
+                    total += size;
+                    GPU_DEBUG_LOG << "primitive kind: " << static_cast<int>(_pd.get_kind()) << ", scratchpad: " << size
+                        << "MB, total " << total << "MB" << std::endl;
+                }
+            }
+
             build_primitive(config);
         }
 
@@ -189,7 +208,7 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
 
         if (has_attrs) {
             {
-                dnnl::scratchpad_mode _scratchpad_mode = dnnl::scratchpad_mode::library;
+                dnnl::scratchpad_mode _scratchpad_mode = dnnl::scratchpad_mode::user;
                 ib >> make_data(&_scratchpad_mode, sizeof(dnnl::scratchpad_mode));
                 _attrs->set_scratchpad_mode(_scratchpad_mode);
             }
@@ -450,6 +469,11 @@ protected:
             args.insert({DNNL_ARG_DST, output.get_onednn_memory(_pd.dnnl::primitive_desc_base::dst_desc(0), offset)});
         }
 
+        if (_scratchpad_md.get_size() != 0) {
+            auto scratchpad = const_cast<cldnn::memory*>(instance.get_intermediates_memories()[0].get());
+            args.insert({DNNL_ARG_SCRATCHPAD, scratchpad->get_onednn_memory(_scratchpad_md, 0)});
+        }
+
         configure_post_ops_arguments(instance, args);
 
         return args;
@@ -510,6 +534,12 @@ protected:
         }
 
         return event;
+    }
+
+    std::vector<layout> get_internal_buffer_layouts_impl() const override {
+        if (_scratchpad_md.get_size() == 0)
+            return {};
+        return {{cldnn::data_types::u8, format::bfyx, {1, 1, 1, (tensor::value_type)(_scratchpad_md.get_size())}}};
     }
 };
 
--- a/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
+++ b/src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
@@ -62,13 +62,13 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
                 _enable_profiling = true;
             }
 
-            GPU_DEBUG_IF(debug_config->verbose >= 5) {
+            GPU_DEBUG_IF(debug_config->verbose >= 4) {
                 if (_scratchpad_md.get_size() > 0) {
-                    static long long total = 0;
+                    static std::atomic_llong total{0};
                     int64_t size = _scratchpad_md.get_size() / 1048576;
                     total += size;
-                    GPU_DEBUG_LOG << "primitive kind: " << static_cast<int>(_pd.get_kind()) << ", scratchpad: " << size
-                        << "MB, total " << total << "MB" << std::endl;
+                    GPU_DEBUG_TRACE_DETAIL << " [scratchpad] kind: " << static_cast<int>(_pd.get_kind())
+                        << ", " << size << "MB, total " << total << "MB" << std::endl;
                 }
             }
 
@@ -470,7 +470,7 @@ protected:
         }
 
         if (_scratchpad_md.get_size() != 0) {
-            auto scratchpad = const_cast<cldnn::memory*>(instance.get_intermediates_memories()[0].get());
+            auto scratchpad = instance.get_intermediates_memories()[0];
             args.insert({DNNL_ARG_SCRATCHPAD, scratchpad->get_onednn_memory(_scratchpad_md, 0)});
         }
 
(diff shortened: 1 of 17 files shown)


src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
```diff
@@ -53,10 +54,23 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
...
             GPU_DEBUG_GET_INSTANCE(debug_config);
             GPU_DEBUG_IF(!debug_config->dump_profiling_data.empty()) {
                 _enable_profiling = true;
             }
+
+            GPU_DEBUG_IF(debug_config->verbose >= 5) {
+                if (_scratchpad_md.get_size() > 0) {
+                    static long long total = 0;
+                    int64_t size = _scratchpad_md.get_size() / 1048576;
+                    total += size;
```
vladimir-paramuzov, line 68: That's thread unsafe code

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
```diff
@@ -53,10 +54,23 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
...
         _pd(pd) {
             _enable_profiling = config.get_property(ov::enable_profiling);
+            _scratchpad_md = _pd.scratchpad_desc();
+
             GPU_DEBUG_GET_INSTANCE(debug_config);
             GPU_DEBUG_IF(!debug_config->dump_profiling_data.empty()) {
                 _enable_profiling = true;
             }
+
+            GPU_DEBUG_IF(debug_config->verbose >= 5) {
```
vladimir-paramuzov, line 64: I'd propose to have `4` being max verbosity level

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
```diff
@@ -454,6 +468,11 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
             args.insert({DNNL_ARG_DST, output.get_onednn_memory(_pd.dnnl::primitive_desc_base::dst_desc(0), offset)});
         }
 
+        if (_scratchpad_md.get_size() != 0) {
+            auto scratchpad = const_cast<cldnn::memory*>(instance.get_intermediates_memories()[0].get());
```
vladimir-paramuzov, line 472: I think `get_onednn_memory` can be marked as `const` instead to avoid this const cast

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
```diff
@@ -515,6 +534,12 @@ struct typed_primitive_onednn_impl : public typed_primitive_impl<PType> {
 
         return event;
     }
+
+    std::vector<layout> get_internal_buffer_layouts_impl() const override {
+        if (_scratchpad_md.get_size() == 0)
+            return {};
+        return {{cldnn::data_types::u8, format::bfyx, {1, 1, 1, (tensor::value_type)(_scratchpad_md.get_size())}}};
```
vladimir-paramuzov, line 541: Use layout c-tor with PartialShape instead

src/plugins/intel_gpu/src/graph/program_node.cpp
```diff
@@ -935,6 +935,10 @@ void program_node::init_onednn_primitive_attributes() {
     // Added this for debug purposes only
     size_t empty_mem = 0xff;
 
+    // Change scratchpad mode to user
+    if (attrs->get_scratchpad_mode() == dnnl::scratchpad_mode::library)
+        attrs->set_scratchpad_mode(dnnl::scratchpad_mode::user);
```
vladimir-paramuzov, line 940: Can it be done unconditionally?

src/plugins/intel_gpu/tests/unit/module_tests/network_test.cpp
```diff
@@ -208,4 +210,47 @@ TEST(network_test, has_proper_event_for_in_order_queue_onednn) {
...
+
+    network net2(net1.get_program(), 1);
+    net2.set_input_data("input", input_mem);
+    net2.execute();
+
+    auto fc1 = net1.get_primitive("fc_prim");
+    auto fc2 = net2.get_primitive("fc_prim");
+
+    // Check if the addresses of scratchpad memory are different when num of stream is 2.
+    std::cout << "Address 1st network: " << fc1->get_intermediates_memories()[0]->buffer_ptr()
```
vladimir-paramuzov, line 251: Remove it

src/plugins/intel_gpu/tests/unit/module_tests/network_test.cpp
```diff
@@ -208,4 +210,47 @@ TEST(network_test, has_proper_event_for_in_order_queue_onednn) {
...
+    network net1(engine, topology, config);
+    net1.set_input_data("input", input_mem);
+    net1.execute();
+
+    network net2(net1.get_program(), 1);
+    net2.set_input_data("input", input_mem);
+    net2.execute();
+
+    auto fc1 = net1.get_primitive("fc_prim");
+    auto fc2 = net2.get_primitive("fc_prim");
```
vladimir-paramuzov, line 248: I think total used memory size should be checked as well in such test

src/plugins/intel_gpu/src/graph/program_node.cpp
jade-cho: No problem. Because there is no change if onednn primitive has no scratchpad.

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
jade-cho: It is changed to atomic variable. Anyway, I think thread-safe is not important because it is for debugging feature.

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
jade-cho: Changed.

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
jade-cho: `get_onednn_memory` is marked as `const`.

src/plugins/intel_gpu/src/graph/impls/onednn/primitive_onednn_base.h
jade-cho: Changed.

src/plugins/intel_gpu/tests/unit/module_tests/network_test.cpp
jade-cho: Removed

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -998,6 +998,11 @@ memory::ptr primitive_inst::allocate_internal_buffer(size_t idx, bool reset) {
         alloc_type = engine.get_lockable_preferred_memory_allocation_type();
     }
     GPU_DEBUG_LOG << "=> allocate to " << alloc_type << std::endl;
+
+    // Reuse intermediate buffer like output buffer.
+    if (_node->get_preferred_impl_type() == impl_types::onednn) {
+        return _network.get_memory_pool().get_intermediate_memory(layout, _node->id(), _network.get_id(), _node->get_memory_dependencies(), alloc_type);
```
isanghao, line 1004: why do we treat onednn case in a special way compared to cldnn primitives?

src/plugins/intel_gpu/src/runtime/memory_pool.cpp
```diff
@@ -240,6 +240,33 @@ memory::ptr memory_pool::get_memory(const layout& layout,
...
+            it->second._type == type &&
+            !has_conflict(it->second._users, restrictions, network_id)) {
+            it->second._users.insert(memory_user(id, network_id));
+            auto ret_mem = _engine->reinterpret_buffer(*it->second._memory, layout);
+            return ret_mem;
+        } else {
+            ++it;
+        }
+    }
+    GPU_DEBUG_LOG << "[" << id << ": output]" << std::endl;
```
isanghao, line 260: is it a necessary log?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -998,7 +998,9 @@ memory::ptr primitive_inst::allocate_internal_buffer(size_t idx, bool reset) {
         alloc_type = engine.get_lockable_preferred_memory_allocation_type();
     }
     GPU_DEBUG_LOG << "=> allocate to " << alloc_type << std::endl;
-    return engine.allocate_memory(layout, alloc_type, reset);
+
+    // Reuse intermediate buffer like output buffer.
+    return _network.get_memory_pool().get_intermediate_memory(layout, _node->id(), _network.get_id(), _node->get_memory_dependencies(), alloc_type);
```
sshlyapn, line 1003: There is `bool reset` function parameter which is supposed to reset memory if needed, so I think we need to handle it to avoid misunderstanding and problebs in the future
Also, `network::allocate_primitives()` uses `prim->allocate_internal_buffers()` with default `true` value

src/plugins/intel_gpu/src/runtime/memory_pool.cpp
```diff
@@ -240,6 +240,36 @@ memory::ptr memory_pool::get_memory(const layout& layout,
     }
 }
 
+memory_ptr memory_pool::get_intermediate_memory(const layout& layout,
```
yeonbok, line 243: 1) Why do we need to reuse intermediate memory with a separate handle? 
2) Can it handle the case where a primitive has multiple intermediates memories?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
jade-cho: It has changed.

src/plugins/intel_gpu/src/runtime/memory_pool.cpp
jade-cho: It has changed.

src/plugins/intel_gpu/src/runtime/memory_pool.cpp
jade-cho: 1. A has not been recycled so far, and I wanted to prevent unnecessary side effects by adding one struct object and one function.
2. I will check this out.

jade-cho: I've checked that this function and a separate handle are not needed. It is enough to use `memory_pool::get_memory()` function. I will remove them.

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
jade-cho: `memory_pool::get_memory()` is used to allocate memory at `prim->allocate_internal_buffer(..)`. So, I remained reset argument.

src/plugins/intel_gpu/src/runtime/memory_pool.cpp
jade-cho: Own id is added to mem deps. And `memory_pool::get_memory()` will be used to intermediate memories. Anyway, there is no multiple intermediate memory issue. It checked with DetectionOutput unit test.

src/plugins/intel_gpu/src/graph/program_node.cpp
isanghao: I think code was not changed after comment. could you please follow-up?

src/plugins/intel_gpu/tests/unit/module_tests/network_test.cpp
isanghao: I think it is OK not to check total size as this test is more about the accuracy issue that happens from buffer sharing betwen multiple streams.

isanghao (commented):
LGTM. Let me check performance.

src/plugins/intel_gpu/src/graph/program_node.cpp
jade-cho: Sorry for the confusion. I said that there is no need for conditions. Using user mode has no effect on onednn primitives, which do not use scratchpad anyway. Are there any plans to support library mode?

isanghao: I think what Vladimir meant is to remove this condition `    if (attrs->get_scratchpad_mode() == dnnl::scratchpad_mode::library)`
But I think this is not a big deal at this moment.

isanghao (approved):

PR #18700, opened 2023-07-21
Fix GPU stress tests under the proxy

ilyachur:
### Details:
 - Fixed stress tests. Mark unavailable devices globally per process
 - Update cached proxy devices only if proxy properties were changed

### Tickets:
 - CVS-115255


ilyachur:
@ilya-lavrenov @nkogteva Please take a look. This solution solves the issue with stress tests on GPU.
Proof: https://github.com/intel-innersource/frameworks.ai.openvino.ci.product-configs/pull/2526

src/plugins/proxy/src/plugin.cpp
```diff
@@ -22,6 +23,26 @@
 
 namespace {
 
+bool compare_containers(const std::vector<std::string>& c1, const std::vector<std::string> c2) {
```
nkogteva, line 26:
```suggestion
bool compare_containers(const std::vector<std::string>& c1, const std::vector<std::string>& c2) {
```

src/plugins/proxy/src/plugin.cpp
```diff
@@ -22,6 +23,26 @@
...
+    if (c1.size() != c2.size())
+        return false;
+    for (size_t i = 0; i < c1.size(); i++) {
+        if (c1.at(i) != c2.at(i))
+            return false;
+    }
+    return true;
+}
+
+bool compare_containers(const std::unordered_set<std::string>& c1, const std::unordered_set<std::string> c2) {
```
nkogteva, line 36:
```suggestion
bool compare_containers(const std::unordered_set<std::string>& c1, const std::unordered_set<std::string>& c2) {
```

nkogteva (approved):
As WA it is LGTM. Is real cause stilll unknown?

ilyachur:
> As WA it is LGTM. Is real cause stilll unknown?

It is not a WA. In old gcc dlopen is not thread safe, we run stress tests on the ubuntu 20, which has such issue.

Comments will fix in the next PR

PR #18708, opened 2023-07-21
Fixed PT build

ilyachur:
### Details:
 - Fixed build after https://github.com/openvinotoolkit/openvino/pull/18616

### Tickets:
 - *ticket-id*


eaidova:
it is not correect. Should I resubmit with proper fix?

PR #18711, opened 2023-07-21
Advanced `mmap` test for IR FE

vurusovs:
### Tickets:
 - CVS-114566


src/frontends/ir/tests/frontend_test_mmap.cpp
```diff
@@ -0,0 +1,161 @@
...
+        GetProcessMemoryInfo(GetCurrentProcess(),&pmc, pmc.cb);
+        return pmc;
+    }
+
+    size_t getVmRSSInKB() {
+        return getMemoryInfo().WorkingSetSize / 1024;
+        }
+
+    size_t getRssFileInKB() {
+
```
vurusovs, line 21: Implement, use `VirtualQueryEx` and `MEMORY_BASIC_INFORMATION`

vurusovs: Will be implemented later

mbencer (approved):

src/frontends/ir/tests/frontend_test_mmap.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "frontend_test.hpp"
+#include "openvino/opsets/opset1.hpp"
+
+class IRFrontendMMapTestsAdvanced : public ::testing::TestWithParam<bool>, public IRFrontendTestsImpl {
+protected:
+    size_t binsize;
+
+    void SetUp() override {
+        size_t CONST_SIZE = 10000000; /*~ 77 MB for f64 to guarantee size is more than size of mapped libraries*/
+        auto parameter = std::make_shared<ov::opset1::Parameter>(ov::element::f64, ov::Shape{CONST_SIZE});
```
slyalin, line 15: Why not to use more regularly used data types like `f3`2? What's the reason to use `f64`? Any specific value for this mmap functionality with that type?

src/frontends/ir/tests/frontend_test_mmap.cpp
```diff
@@ -0,0 +1,106 @@
...
+        constant->set_friendly_name("value1");
+        auto add = std::make_shared<ov::opset1::Add>(parameter, constant);
+        add->set_friendly_name("Add");
+        auto result = std::make_shared<ov::opset1::Result>(add);
+        result->set_friendly_name("output");
+        auto model = std::make_shared<ov::Model>(ov::NodeVector{result}, ov::ParameterVector{parameter});
+
+        auto filePrefix = CommonTestUtils::generateTestFilePrefix();
+        xmlFileName = filePrefix + "_IrFrontendTestModel.xml";
+        binFileName = filePrefix + "_IrFrontendTestModel.bin";
```
slyalin, line 29: Stop forming bin file name from xml file name. It is done automatically inside serialize if you leave bin_path with its default value.

src/frontends/ir/tests/frontend_test_mmap.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "openvino/opsets/opset1.hpp"
+
+class IRFrontendMMapTestsAdvanced : public ::testing::TestWithParam<bool>, public IRFrontendTestsImpl {
+protected:
+    size_t binsize;
+
+    void SetUp() override {
+        size_t CONST_SIZE = 10000000; /*~ 77 MB for f64 to guarantee size is more than size of mapped libraries*/
+        auto parameter = std::make_shared<ov::opset1::Parameter>(ov::element::f64, ov::Shape{CONST_SIZE});
+        parameter->set_friendly_name("input");
```
slyalin, line 16: Where do we use set friendly name? Looks useless. Remove all `set_friendly_name` calls.

src/frontends/ir/tests/frontend_test_mmap.cpp
```diff
@@ -0,0 +1,106 @@
...
+    auto rss_mapped_read = CommonTestUtils::getRssFileInKB();
+    auto compiled_model = core.compile_model(model, "CPU");
+    auto rss_mapped_compiled = CommonTestUtils::getRssFileInKB();
+
+    // Mappings size (RssFile) should increase at least on .bin size
+    bool is_weights_mapped = (rss_mapped_compiled - rss_mapped_read) > binsize;
+    EXPECT_TRUE(is_mmap == is_weights_mapped);
+}
+
+TEST_P(IRFrontendMMapTestsAdvanced, fe_read_and_compile_model) {
```
slyalin, line 68: Why do we need this test if everything that is related to mmap can be configured via `read_model` and the same FEs are used underneath? I think previous test should be applied to different frameworks that are relevant: ONNX and TF in the future.

slyalin (changes requested):

src/frontends/ir/tests/CMakeLists.txt
```diff
@@ -4,11 +4,18 @@
 
 set(TARGET_NAME ov_ir_frontend_tests)
 
+list(APPEND IR_TESTS_DEPENDENCIES openvino_ir_frontend openvino_intel_cpu_plugin)
+if (NOT UNIX OR NOT ENABLE_INTEL_CPU)
+    list(REMOVE_ITEM IR_TESTS_DEPENDENCIES openvino_intel_cpu_plugin)
```
ilyachur, line 9: It will be better to add new dependencies instead of excluding:
```
if(ENABLE_INTEL_CPU)
    list(APPEND DEPENDENCIES openvino_intel_cpu_plugin)
    list(APPEND COMPILE_DEFINITIONS ENABLE_INTEL_CPU)
endif()
```

BTW why do you use CPU plugin for tests? Can you use template plugin?

src/frontends/ir/tests/CMakeLists.txt
```diff
@@ -4,11 +4,18 @@
...
+    set(EXCLUDED_TESTS ${CMAKE_CURRENT_SOURCE_DIR}/frontend_test_mmap.cpp)
+endif()
+
 ov_add_test_target(
         NAME ${TARGET_NAME}
         ROOT ${CMAKE_CURRENT_SOURCE_DIR}
+        EXCLUDED_SOURCE_PATHS ${EXCLUDED_TESTS}
```
ilyachur, line 16: You can use ifdef to exclude some tests if macro is not defined

src/frontends/ir/tests/frontend_test_mmap.cpp
vurusovs: There was no special intention to use `f64`, so replaced with `f32`. Thanks!

src/frontends/ir/tests/frontend_test_mmap.cpp
vurusovs: Done, thanks!

src/frontends/ir/tests/frontend_test_mmap.cpp
vurusovs: Removed

src/frontends/ir/tests/frontend_test_mmap.cpp
vurusovs: Replaced with test where by default `FE` reads .bin

src/frontends/ir/tests/CMakeLists.txt
vurusovs: Thanks for suggestion! Anyway, CPU dependency is removed

src/frontends/ir/tests/CMakeLists.txt
vurusovs: Thanks for suggestion! Anyway, CPU dependency is removed

src/frontends/ir/tests/frontend_test_mmap.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // trigger plugin loading in order to map libraries and not affect memory values during compilation
+    core.get_versions("CPU");
+    core.set_property(ov::enable_mmap(GetParam()));
+
+    auto rss_init = CommonTestUtils::getVmRSSInKB();
+    auto model = core.read_model(xmlFileName, binFileName);
+    auto rss_read = CommonTestUtils::getVmRSSInKB();
+
+    // RAM should increase on at least 90% of .bin size. 10% is a proposed error that cover file system cache
+    size_t bin_in_RAM = binsize * 0.9;
```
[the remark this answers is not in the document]
vurusovs: As discussed:
1. Test is run in a separate process, so memory behavior became independent from other tests
2. `compile_model` is removed, so `RssFile` metric is also gone (it shows how which mapped memory is loaded)

Anyway, I decided to leave RSS delta instead of full RSS metric because makes test not dependent from other components (libraries size, Core etc.). If some prereading of .bin will be implemented out of `read_model` scope, it will be found by the test with `enable_mmap(false)`

src/frontends/ir/tests/frontend_test_mmap.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "common_test_utils/file_utils.hpp"
+#include "frontend_test.hpp"
+#include "openvino/opsets/opset1.hpp"
+
+class IRFrontendMMapTestsAdvanced : public ::testing::TestWithParam<bool>, public IRFrontendTestsImpl {
+protected:
+    size_t binsize;
+
+    void SetUp() override {
+        size_t CONST_SIZE = 10000000; /*~ 77 MB for f64 to guarantee size is more than size of mapped libraries*/
```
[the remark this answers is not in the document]
vurusovs: Done

src/frontends/ir/tests/frontend_test_mmap.cpp
```diff
@@ -0,0 +1,121 @@
...
+        auto model = core.read_model(xmlFileName);
+        auto rss_read = ov::test::utils::getVmRSSInKB();
+
+        bool is_weights_read = (rss_read - rss_init) > REF_RSS;
+        if (is_mmap == is_weights_read) {
+            std::cerr << "Test failed: mmap is " << (is_mmap ? "enabled" : "disabled") << ", but weights are "
+                      << (is_weights_read ? "read" : "not read") << " in RAM" << std::endl;
+            exit(1);
+        }
+        std::cerr << "Test passed" << std::endl;
```
ilyachur, line 61: Why do you use `cerr` for notification message?
Moreover I think this message is duplicated in 67 line :)

src/tests/test_utils/common_test_utils/src/common_utils.cpp
```diff
@@ -34,12 +45,70 @@ std::string generateTestFilePrefix() {
...
+                break;
+            }
+        }
+        fclose(file);
+    }
+    return result;
+}
+
+size_t getVmSizeInKB() {
+    return getSystemDataByName(const_cast<char*>("VmSize:"));
```
ilyachur, line 103: As I understand it doesn't work on macOS, should we add implementations which will throw exception?

ilyachur (approved):
In general LGTM

src/frontends/ir/tests/frontend_test_mmap.cpp
vurusovs: It's a design limitation of test in separate subprocess. Because `ASSERT_` and `EXPECT_` can't be handled, 2 processes communicate via stdout or stderr, and stdout isn't presented correctly in logs, so stderr is used. Duplicated message is a `regex` that means expected message that means test is correct

The reason why separate process used is to prevent any memory side effects from previous tests

src/tests/test_utils/common_test_utils/src/common_utils.cpp
vurusovs: Yes, we need to add separate implementation for MacOS. For the moment that implementation will return -1 on MacOS, and I guess better way is to throw `not implemented`. Will be done in next PR

slyalin (approved):

PR #18718, opened 2023-07-21
[CPU] MatMul: move i8 compressed constants constant folding to the plugin

v-Golubev:
### Details:
 - *Reused LPT pass for ConstantFold pass disabling for decompression subgraph*
 - *GraphOptimizer: added FuseFCAndWeightsDecompression transformation*
 - *Transformation pipeline adapted to the MatMuls with compressed weights*
 - *Added MoveFCReshapeToWeights CPU transformation*

### Tickets:
 - *CVS-113732*

### DIFF
--- a/src/plugins/intel_cpu/src/graph_optimizer.cpp
+++ b/src/plugins/intel_cpu/src/graph_optimizer.cpp
@@ -75,6 +75,10 @@ void GraphOptimizer::ApplyCommonGraphOptimizations(Graph &graph) {
     FuseConvMatmulFCDeconvAndDQScales(graph);
     graph.RemoveDroppedNodes();
 
+    OV_ITT_SCOPE_NEXT(FIRST_INFERENCE, taskChain, "FuseMatmulAndWeightsDecompression");
+    FuseFCAndWeightsDecompression(graph);
+    graph.RemoveDroppedNodes();
+
     OV_ITT_SCOPE_NEXT(FIRST_INFERENCE, taskChain, "FuseConvolutionAndBias");
     FuseConvolutionMatMulDeconvAndBias(graph);
     graph.RemoveDroppedNodes();
@@ -274,6 +278,107 @@ void GraphOptimizer::FuseConvMatmulFCDeconvAndDQScales(Graph &graph) {
             graph.RemoveEdge(p_edge);
             graph.DropNode(mul);
         }
+    }
+}
+
+void GraphOptimizer::FuseFCAndWeightsDecompression(Graph &graph) {
+    auto& graphNodes = graph.GetNodes();
+
+    auto tryFuseWeightsDecompression = [&](NodePtr node) {
+        // todo: draw pattern
+
+        // pattern matching
+        if (node->getType() != Type::FullyConnected)
+            return false;
+
+        auto multiplyNode = node->getParentEdgesAtPort(1)[0]->getParent();
+        if (!multiplyNode->isConstant())
+            return false;
+
+        if (multiplyNode->getType() != Type::Eltwise || multiplyNode->getAlgorithm() != Algorithm::EltwiseMultiply)
+            return false;
+        if (multiplyNode->getChildEdges().size() != 1)
+            return false;
+
+        auto multiplyConstNode = multiplyNode->getParentEdgesAtPort(1)[0]->getParent();
+        if (multiplyConstNode->getType() != Type::Input)
+            return false;
+        if (multiplyConstNode->getChildEdges().size() != 1)
+            return false;
+
+        auto subtractNode = multiplyNode->getParentEdgesAtPort(0)[0]->getParent();
+        if (subtractNode->getType() != Type::Eltwise || subtractNode->getAlgorithm() != Algorithm::EltwiseSubtract)
+            return false;
+        if (subtractNode->getChildEdges().size() != 1)
+            return false;
+
+        auto subtractConstNode = subtractNode->getParentEdgesAtPort(1)[0]->getParent();
+        if (subtractConstNode->getType() != Type::Input)
+            return false;
+        if (subtractConstNode->getChildEdges().size() != 1)
+            return false;
+
+        auto convertNode = subtractNode->getParentEdgesAtPort(0)[0]->getParent();
+        if (convertNode->getType() != Type::Convert)
+            return false;
+        if (convertNode->getChildEdges().size() != 1)
+            return false;
+
+        auto matmulConstNode = convertNode->getParentEdgesAtPort(0)[0]->getParent();
+        if (matmulConstNode->getType() != Type::Input)
+            return false;
+        if (matmulConstNode->getChildEdges().size() != 1)
+            return false;
+
+        // precision limitations
+        if (multiplyConstNode->getOriginalOutputPrecisionAtPort(0) != Precision::FP32)
+            return false;
+        if (subtractConstNode->getOriginalOutputPrecisionAtPort(0) != Precision::FP32)
+            return false;
+        if (matmulConstNode->getOriginalOutputPrecisionAtPort(0) != Precision::U8)
+            return false;
+
+        // shape limitations
+        auto weightsShape = node->getInputShapeAtPort(1);
+        if (weightsShape != matmulConstNode->getOutputShapeAtPort(0))
+            return false;
+
+        VectorDims expectedDims = {weightsShape.getDims()[0], 1};
+        if (multiplyConstNode->getOutputShapeAtPort(0).getDims() != expectedDims)
+            return false;
+        if (subtractConstNode->getOutputShapeAtPort(0).getDims() != expectedDims)
+            return false;
+
+        // decompression fusion
+        auto fcNode = dynamic_cast<node::FullyConnected*>(node.get());
+        if (fcNode == nullptr)
+            IE_THROW() << "Cannot cast to FullyConnected node";
+
+        fcNode->fuseDecompressionMultiply(multiplyConstNode);
+        fcNode->fuseDecompressionSubtract(subtractConstNode);
+
+        fcNode->addOriginalLayer(multiplyNode->getOriginalLayers());
+        fcNode->addOriginalLayer(subtractNode->getOriginalLayers());
+        fcNode->addOriginalLayer(convertNode->getOriginalLayers());
+
+        auto subtractConstEdge = subtractConstNode->getChildEdges()[0].lock();
+        graph.RemoveEdge(subtractConstEdge);
+
+        auto multiplyConstEdge = multiplyConstNode->getChildEdges()[0].lock();
+        graph.RemoveEdge(multiplyConstEdge);
+
+        graph.DropNode(convertNode);
+        graph.DropNode(subtractNode);
+        graph.DropNode(multiplyNode);
+
+        fcNode->setOriginalInputPrecisionAtPort(1, matmulConstNode->getOriginalOutputPrecisionAtPort(0));
+
+        return true;
+    };
+
+    for (size_t i = 0; i < graphNodes.size(); i++) {
+        auto fcNode = graphNodes[i];
+        tryFuseWeightsDecompression(fcNode);
     }
 }
 
--- a/src/plugins/intel_cpu/src/graph_optimizer.cpp
+++ b/src/plugins/intel_cpu/src/graph_optimizer.cpp
@@ -282,103 +282,93 @@ void GraphOptimizer::FuseConvMatmulFCDeconvAndDQScales(Graph &graph) {
 }
 
 void GraphOptimizer::FuseFCAndWeightsDecompression(Graph &graph) {
+    auto expectedNode = [](NodePtr node, Type expectedType) {
+        return node->getType() == expectedType && node->getChildEdges().size() == 1;
+    };
+
     auto& graphNodes = graph.GetNodes();
-
-    auto tryFuseWeightsDecompression = [&](NodePtr node) {
-        // todo: draw pattern
-
-        // pattern matching
+    for (size_t i = 0; i < graphNodes.size(); i++) {
+        auto node = graphNodes[i];
         if (node->getType() != Type::FullyConnected)
-            return false;
+            continue;
 
-        auto multiplyNode = node->getParentEdgesAtPort(1)[0]->getParent();
+        const auto multiplyNode = node->getParentEdgesAtPort(1)[0]->getParent();
         if (!multiplyNode->isConstant())
-            return false;
+            continue;
 
-        if (multiplyNode->getType() != Type::Eltwise || multiplyNode->getAlgorithm() != Algorithm::EltwiseMultiply)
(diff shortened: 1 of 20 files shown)


v-Golubev:
@itikhono Could you please take a look on transformations part? Thanks

v-Golubev:
@dmitry-gorokhov could you please review CPU part? Thanks

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/convert_matmul_to_fc.cpp
```diff
@@ -74,7 +77,7 @@ ov::intel_cpu::ConvertMatMulToFC::ConvertMatMulToFC() {
                 shape_b_aligned.insert(shape_b_aligned.begin(), 1);
             }
 
-            if (matmul->get_transpose_a() && rank_a != 1) {
```
dmitry-gorokhov, line 77: Just wondering why this condition was deleted?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -1075,6 +1076,28 @@ bool FullyConnected::useSparseWeightsDecompression() {
...
+void FullyConnected::fuseDecompressionMultiply(const NodePtr& constData) {
+    fuseDecompressionConstant(constData, decompressionMultiply);
+}
+
+void FullyConnected::fuseDecompressionSubtract(const NodePtr& constData) {
+    fuseDecompressionConstant(constData, decompressionSubtract);
+}
+
+void FullyConnected::fuseDecompressionConstant(const NodePtr& constData, std::vector<float>& decompressionValues) {
+    auto *constInputNode = dynamic_cast<node::Input *>(constData.get());
```
dmitry-gorokhov, line 1088: in fact the same code block is used in three different places inside the plugin already (node.cpp, rnn.cpp, fullyconnected.cpp).
What about introduction of castToVector method into Input node API (same as cast_vector inside Constant ngraph op)?

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -277,6 +277,99 @@ void GraphOptimizer::FuseConvMatmulFCDeconvAndDQScales(Graph &graph) {
...
+
+        const auto multiplyNode = with_transpose ? parent->getParentEdgesAtPort(0)[0]->getParent() : parent;
+        if (!expectedNode(multiplyNode, Type::Eltwise) || multiplyNode->getAlgorithm() != Algorithm::EltwiseMultiply ||
+            !multiplyNode->isConstant())
+            continue;
+
+        const auto multiplyConstNode = multiplyNode->getParentEdgesAtPort(1)[0]->getParent();
+        if (!expectedNode(multiplyConstNode, Type::Input))
+            continue;
+
```
dmitry-gorokhov, line 303: Would propose to add `CPU_GRAPH_OPTIMIZER_SCOPE(FuseFCAndWeightsDecompression);` line here to enable conditional compilation optimization for this pass.

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -277,6 +277,99 @@ void GraphOptimizer::FuseConvMatmulFCDeconvAndDQScales(Graph &graph) {
...
+        if (!expectedNode(convertNode, Type::Convert))
+            continue;
+        const auto weightsNode = convertNode->getParentEdgesAtPort(0)[0]->getParent();
+        if (!expectedNode(weightsNode, Type::Input))
+            continue;
+
+        // Precision limitations
+        if (multiplyConstNode->getOriginalOutputPrecisionAtPort(0) != Precision::FP32)
+            continue;
+        if (weightsNode->getOriginalOutputPrecisionAtPort(0) != Precision::U8)
```
dmitry-gorokhov, line 326: Can you add vector with supported weights precision in the beggining of the pass and check it here? 
The expection is list of supported weights precision will be expanded in nearest future, so to minimize required changes for the developers it is reasonable to implement it as parameter.

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/matmul_weights_decompression.cpp
```diff
@@ -0,0 +1,220 @@
...
+            auto transpose_constant = ov::opset10::Constant::create(ov::element::i32, {rank}, order);
+            auto transpose = std::make_shared<ov::opset10::Transpose>(matmul_weights, transpose_constant);
+            matmul_weights = transpose;
+        }
+        auto matMul = builder::makeMatMul(params[0], matmul_weights);
+        ov::Shape bias_const_shape{static_cast<size_t>(matMul->get_output_partial_shape(0).rbegin()->get_length())};
+        auto bias_const = ngraph::builder::makeConstant<float>(data_precision, bias_const_shape, {}, true);
+        auto bias = std::make_shared<ov::opset10::Add>(matMul, bias_const);
+
+        return std::make_shared<ov::Model>(matMul, params, "MatmulWeightsDecompression");
```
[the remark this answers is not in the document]
dmitry-gorokhov: 3. Would also ask you to make weights precision parameter of the test

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/convert_matmul_to_fc.cpp
```diff
@@ -133,25 +135,28 @@ ov::intel_cpu::ConvertMatMulToFC::ConvertMatMulToFC() {
...
+            auto reshape_shape = ngraph::opset1::Constant::create(ngraph::element::i32, ngraph::Shape{ 2 }, reshape_shape_values);
             fc_input_b = ov::op::util::make_try_fold<ngraph::opset1::Reshape>(fc_input_b, reshape_shape, false);
             if (!std::dynamic_pointer_cast<ngraph::opset1::Constant>(fc_input_b.get_node_shared_ptr())) {
                 new_ops.push_back(reshape_shape);
             }
             new_ops.push_back(fc_input_b.get_node_shared_ptr());
         }
 
+        // Weights normalization
+        if (!matmul->get_transpose_b()) {
```
itikhono, line 154: is "!" supposed to be here?
or it should be
if(matmul->get_transpose_b()) ?
please double check

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/convert_matmul_to_fc.cpp
```diff
@@ -133,25 +135,28 @@ ov::intel_cpu::ConvertMatMulToFC::ConvertMatMulToFC() {
...
-        if (!matmul->get_transpose_b()) {
-            fc_input_b = create_transpose(fc_input_b, matmul->get_friendly_name() + "/transpose_b");
-        }
-
         if (rank_b != 2) {
             ngraph::Dimension K = *(shape_b_aligned.rbegin() + 1);
             NGRAPH_CHECK(K.is_static());
-            std::vector<int64_t> reshape_shape_values = { -1ll, static_cast<int64_t>(K.get_length()) };
-            auto reshape_shape = ngraph::opset1::Constant::create(ngraph::element::i64, ngraph::Shape{ 2 }, reshape_shape_values);
+            std::vector<int64_t> reshape_shape_values = {K.get_length()};
```
itikhono, line 141: minor: 
maybe it's better from readability perspective
std::vector<int64_t> reshape_shape_values(2);
auto k_len =  K.get_length();
matmul->get_transpose_b() ? reshape_shape_values = { -1, k_len} : reshape_shape_values  = {k_len, -1};

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/move_fc_reshape_to_weights.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include <openvino/pass/pattern/op/wrap_type.hpp>
+#include <openvino/pass/pattern/op/or.hpp>
+#include <openvino/opsets/opset1.hpp>
+
+#include "itt.hpp"
+
+ov::intel_cpu::MoveFCReshapeToWeights::MoveFCReshapeToWeights() {
+    MATCHER_SCOPE(MoveFCReshapeToWeights);
+    using namespace ov::pass::pattern;
+    auto weights_m = wrap_type<ov::opset1::Constant>(consumers_count(1));
```
itikhono, line 17: the current recommendation is to use the concrete operation versions, e.g.
ov::op::v0::Constant instead of ov::opset1::Constant due to CVS-91401 
and do not include opset headers (<openvino/opsets/opset1.hpp>)

In short, if we include the opset header, it adds ~5mb to obj file size, there are lots of transformations (N) and every transformation corresponds to its own obj file (N * 5mb), so the final lib size can exceed 4gb limit on Win platform in debug configuration

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/move_fc_reshape_to_weights.cpp
```diff
@@ -0,0 +1,102 @@
...
+        ov::Shape expected_weights_shape(3, 1);
+        expected_weights_shape[1] = fc_input_shape[with_transpose ? 1 : 0];
+        expected_weights_shape[2] = fc_input_shape[with_transpose ? 0 : 1];
+        if (weights->get_output_shape(0) != expected_weights_shape)
+            return false;
+
+        auto squeeze_constant = [](const std::shared_ptr<ov::Node>& node) {
+            const auto constant = ov::as_type_ptr<ov::opset1::Constant>(node);
+            auto shape = constant->get_shape();
+            OPENVINO_ASSERT(shape[0] == 1);
```
itikhono, line 83: usually we try to avoid using exceptions/asserts in the transformations code.
can we check the shapes first and if they contain shape[0] == 1, then do not apply the transformation at all and return false immediately?

src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp
```diff
@@ -396,6 +397,27 @@ bool isSuitableChildForFusingSumActivation(const std::shared_ptr<const Node> &no
...
+            multiply = consumer;
+        }
+    }
+    if (!multiply)
+        return false;
+    const auto consumers = multiply->get_output_target_inputs(0);
+    if (consumers.size() != 1)
+        return false;
+    const auto consumer = consumers.begin()->get_node()->shared_from_this();
+    return ov::is_type<ov::opset1::MatMul>(consumer) && ov::op::util::is_on_constant_path(multiply);
```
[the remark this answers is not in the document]
v-Golubev: I agree. The logic has been simplified, thanks

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/matmul_weights_decompression.cpp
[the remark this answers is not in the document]
v-Golubev: The test class has been extended as per suggestions. Thank you

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/matmul_weights_decompression.cpp
```diff
@@ -0,0 +1,220 @@
...
+        function = initSubgraph(inputDynamicShapes, netType, ov::element::u8, transpose_weights, decompression_sub, reshape_on_decompression);
+    }
+};
+
+TEST_P(MatmulWeightsDecompression, CompareWithRefs) {
+    run();
+    std::map<std::string, std::string> additional_config = std::get<4>(GetParam());
+    const size_t expected_count = additional_config[PluginConfigParams::KEY_ENFORCE_BF16] == PluginConfigParams::YES ? 1 : 0;
+    CheckNumberOfNodesWithType(compiledModel, "Convert", expected_count);
+    CheckNumberOfNodesWithType(compiledModel, "Eltwise", expected_count);
```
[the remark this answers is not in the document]
v-Golubev: The check is added, thanks

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
v-Golubev: I agree with the idea. But can we do that in a separate PR? I took a look on ngraph cast_vector realization: it is not trivial, and it may take extra time to do smth the same in CPU Input node

src/plugins/intel_cpu/src/graph_optimizer.cpp
v-Golubev: Added, thanks

src/plugins/intel_cpu/src/graph_optimizer.cpp
v-Golubev: Added is `supportedWeightsPrecisions`. But as std::set, not std::vector. In my opinion, this is a more appropriate structure to describe the supported weights precisions

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/convert_matmul_to_fc.cpp
v-Golubev: Yes, this version is correct. CPU FullyConnected operation is converted from the MatMul node with `transpose_a=false` and `transpose_b=true` so we create a separate transpose if `transpose_b=false`

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/convert_matmul_to_fc.cpp
v-Golubev: I agree, your version is more readable. But I changed it a little bit: hope the readability is not affected by this.

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/move_fc_reshape_to_weights.cpp
v-Golubev: Thank you for the explanation. I refactored the both affected by this PR transformations to match your suggestions

src/plugins/intel_cpu/src/transformations/cpu_opset/common/pass/move_fc_reshape_to_weights.cpp
v-Golubev: Actually we already check it upper so I just removed this useless assert

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
dmitry-gorokhov: Agree. Lets follow-up in separate PR

itikhono (approved):

dmitry-gorokhov (approved):

PR #18722, opened 2023-07-22
[GPU] fix accuracy issue in Dolly 

yeonbok:
### Details:
 - Revert partial change in PR18596 to fix acc issue in dolly for dGPU
 - The issue happens by removing sync before shapeof 
 - The sync issue only happens in a complicated network execution, cannot reproduce with unittest


wilson-seok (approved):

e-ddykim (approved):

kelvinchoi-intel (approved):

ahnyoung-paul (approved):

p-durandin (commented):
Should we add test case?

yeonbok:
> Should we add test case?

Cannot add, the sync issue happens in a large network. This commit is just reverting previous change.

PR #18728, opened 2023-07-24
Respect secondary property of performance hint

songbell:
### Details:
 - do not overwrite secondary property hints
 - enable new test class with real ie core

### Tickets:
 - *ticket-id*


src/plugins/auto/src/plugin.cpp
```diff
@@ -172,6 +171,10 @@ std::vector<DeviceInformation> Plugin::parse_meta_devices(const std::string& pri
     auto get_device_config = [&] (const DeviceName & device_with_id) {
         auto device_config = get_core()->get_supported_property(device_with_id, properties);
         set_default_hint(device_with_id, device_config, properties);
+        // only in case of cumulative, update to tput for hardware device
+        auto && iter = device_config.find(ov::hint::performance_mode.name());
+        if (iter != device_config.end() && iter->second.as<std::string>() == "CUMULATIVE_THROUGHPUT")
```
peterchen-intel, line 176: Will we still get "CUMULATIVE_THROUGHPUT" instead of ov::hint::PerformanceMode::THROUGHPUT ? Per my understanding legacy API -> wrapper -> AUTO Plugin API 2.0 ?

songbell: yes, ov::Any with string type will be there if use legacy API, 
e.g  we can do this: ov::Any(ov::hint::PerformanceMode::THROUGHPUT) == "THROUGHPUT"
but not this: ov::Any("THROUGHPUT") == ov::hint::PerformanceMode::THROUGHPUT

songbell:
@wangleis can we merge this?

yangwang201911 (approved):

wangleis (approved):

PR #18745, opened 2023-07-24
[Snippets] Optimized second attempt of SplitDimensionM optimization

a-sidorova:
### Details:
 - *SplitDimensionM optimization has 2 tries to reshape body. First step splits dimension `M` of MHA pattern into two `batch_dim` and `new_dim_m` to have `batch * batch_dim % threads == 0` - all threads have the same work amount. If it's not possible, we have Second step (second attempt). On the master branch this stage tries to make `new_dim_m` as the most value that `new_dim_m < optimal_m_dim`. It's not optimal way since not all threads have work anyway. The PR updates this step. Now the algorithm tries to select such values so that as many threads as possible have work (see the comment for `[ Second Step ]`)*
 - *Removed `optimal_m_dim` to increase work thread amount*

### Tickets:
 - *115284*

### DIFF
--- a/src/common/snippets/src/pass/common_optimizations.cpp
+++ b/src/common/snippets/src/pass/common_optimizations.cpp
@@ -277,7 +277,7 @@ void CommonOptimizations::SplitDimensionM(const std::shared_ptr<ov::snippets::op
                 const auto shape_const = ov::as_type_ptr<ov::op::v0::Constant>(broadcast->input_value(1).get_node_shared_ptr());
                 OPENVINO_ASSERT(shape_const, "SplitDimensionM expects Broadcast with Constant output shape");
                 const auto new_shape = get_updated_shape(shape_const->cast_vector<size_t>(), true);
-                broadcast->set_argument(1, std::make_shared<ov::op::v0::Constant>(ov::element::i32, ov::Shape{new_shape.size()}, new_shape));
+                broadcast->set_argument(1, std::make_shared<ov::op::v0::Constant>(shape_const->get_element_type(), ov::Shape{new_shape.size()}, new_shape));
             }
         }
         subgraph->validate_and_infer_types();
--- a/src/common/snippets/src/pass/common_optimizations.cpp
+++ b/src/common/snippets/src/pass/common_optimizations.cpp
@@ -109,26 +109,22 @@ void CommonOptimizations::SplitDimensionM(const std::shared_ptr<ov::snippets::op
     const auto lcm = get_lcm(batch_dim, optimal_parallelism_work_amount);  // LCM(b, nthrs)
     const auto batch_dim_multiplier = lcm / batch_dim;  // LCM(b, nthrs) / b
     const auto needed_new_dim = m_dim / batch_dim_multiplier;  // m / (LCM(b, nthrs) / b) - needed factors of dimension m
-    if (batch_dim_multiplier * needed_new_dim == m_dim) {
+    if (batch_dim_multiplier * needed_new_dim == m_dim && is_optimized(batch_dim_multiplier)) {
         batch_m_dim = batch_dim_multiplier;
         new_m_dim = needed_new_dim;
-    }
-
-    // [ Second Step ]
-    // If we couldn't optimally split on the previous step, try the second step.
-    // The algorithm finds the more optimal parallelism work amount [batch_dim * batch_m_dim],
-    // where batch_m_dim is divisor of dimension M.
-    // The optimal parallelism work amount means the case when as many threads as possible have work
-    // For example, there are 8 threads and shape [5, 384, 32]
-    //              768 = [2 x 192] = [3 x 128] = [4 x 96] = [6 x 64]
-    //               - [5, 2, 192, 32] - WA = 10 = 8 + 2 (6 threads calculates once and 2 threads twice)
-    //               - [5, 3, 128, 32] - WA = 15 = 8 + 7 (all threads have 2 kernel except one thread) <- the most optimal case
-    //               - [5, 4,  96, 32] - WA = 20 = 8 x 2 + 4
-    //               - [5, 6,  64, 32] - WA = 30 = 8 x 3 + 6
-    // The most optimal and possible case is [5, 3, 128, 32] - almost all threads executes kernel twice
-    if (!is_optimized(batch_m_dim)) {
-        batch_m_dim = 1;
-        new_m_dim = m_dim;
+    } else {
+        // [ Second Step ]
+        // If we couldn't optimally split on the previous step, try the second step.
+        // The algorithm finds the more optimal parallelism work amount [batch_dim * batch_m_dim],
+        // where batch_m_dim is divisor of dimension M.
+        // The optimal parallelism work amount means the case when as many threads as possible have work
+        // For example, there are 8 threads and shape [5, 384, 32]
+        //              768 = [2 x 192] = [3 x 128] = [4 x 96] = [6 x 64]
+        //               - [5, 2, 192, 32] - WA = 10 = 8 + 2 (6 threads calculates once and 2 threads twice)
+        //               - [5, 3, 128, 32] - WA = 15 = 8 + 7 (all threads have 2 kernel except one thread) <- the most optimal case
+        //               - [5, 4,  96, 32] - WA = 20 = 8 x 2 + 4
+        //               - [5, 6,  64, 32] - WA = 30 = 8 x 3 + 6
+        // The most optimal and possible case is [5, 3, 128, 32] - almost all threads executes kernel twice
         // Heuristic value for a quick exit from the algorithm.
         // The value shows the number of threads in percentages that perform the most equal work
         const auto optimal_thread_num_percent = 0.8;
--- a/src/common/snippets/src/pass/common_optimizations.cpp
+++ b/src/common/snippets/src/pass/common_optimizations.cpp
@@ -19,21 +19,35 @@ namespace ov {
 namespace snippets {
 namespace pass {
 
-bool CommonOptimizations::CanOptimizeParallelWA(const std::shared_ptr<const ov::Node>& node, size_t minimal_concurrency) {
-    if (!ov::is_type<ov::op::v0::MatMul>(node))
+namespace {
+size_t get_lcm(size_t a, size_t b) {
+    std::function<size_t(size_t, size_t)> get_gcd;
+    get_gcd = [&get_gcd](size_t a, size_t b) {
+        if (b == 0)
+            return a;
+        return get_gcd(b, a % b);
+    };
+    return a / get_gcd(a, b) * b;
+}
+
+bool is_supported_matmul_for_split_dim_m_optimization(const std::shared_ptr<const ov::Node>& node) {
+    const auto matmul = ov::as_type_ptr<const ov::op::v0::MatMul>(node);
+    return matmul && !matmul->get_transpose_a() && !matmul->is_dynamic() && node->get_shape().size() == 3; // It's needed only for 3D MHA patterns
+}
+}  // namespace
+
+bool CommonOptimizations::CanOptimizeParallelWA(const std::shared_ptr<const ov::Node>& node, size_t concurrency) {
+    if (!is_supported_matmul_for_split_dim_m_optimization(node))
         return false;
-    // It's needed only for 3D MHA patterns
     const auto mm_shape = node->get_shape();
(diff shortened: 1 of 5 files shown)


src/common/snippets/src/pass/common_optimizations.cpp
```diff
@@ -19,21 +19,35 @@ namespace ov {
...
+
+bool CommonOptimizations::CanOptimizeParallelWA(const std::shared_ptr<const ov::Node>& node, size_t concurrency) {
+    if (!is_supported_matmul_for_split_dim_m_optimization(node))
         return false;
-    // It's needed only for 3D MHA patterns
     const auto mm_shape = node->get_shape();
-    if (mm_shape.size() != 3)
-        return false;
     const auto current_parallel_work_amount =
         std::accumulate(mm_shape.rbegin() + 2, mm_shape.rend(), size_t(1), std::multiplies<size_t>());
```
IvanNovoselov, line 44: Minor: looks like an unnecessary complication. We should either rely on the checks inside `is_supported_matmul_for_split_dim_m_optimization` and assume the shape rank to be 3, then we can just take a single dim value. Alternatively, we can write a more general code, so the checks in `is_supported_matmul_for_split_dim_m_optimization` could be updated independently. In this case, we need to check that the shape rank>=2.

src/common/snippets/src/pass/common_optimizations.cpp
```diff
@@ -68,82 +82,87 @@ void CommonOptimizations::SplitDimensionM(const std::shared_ptr<ov::snippets::op
 
     const auto mm_shape = matmul0->get_shape();
     const auto m_dim = get_dim_M(mm_shape);  // M
-    const auto n_dim = mm_shape.back(); // N
+    const auto batch_dim =
+        std::accumulate(mm_shape.rbegin() + 2, mm_shape.rend(), size_t(1), std::multiplies<size_t>());  // B (batch)
```
IvanNovoselov, line 86: The same as with `CanOptimizeParallelWA` probably worth checking the rank

src/common/snippets/src/pass/common_optimizations.cpp
```diff
@@ -68,82 +82,87 @@ void CommonOptimizations::SplitDimensionM(const std::shared_ptr<ov::snippets::op
...
     // When this optimization will be moved into Subgraph and blocking param will be implemented as dependents of shapes,
     // need to implement common way (for all backends) to calculate optimal value of M dimension
     const auto optimal_m_dim = 32 * 2;
-    const auto optimal_parallelism_work_amount = minimal_concurrency;
-    if (m_dim <= optimal_m_dim)
+    const auto optimal_parallelism_work_amount = concurrency;
+    // We skip optimization
+    //  - if the current M dimension is not enough for splitting
+    //  - if the current batch is optimal for concurrency
+    if (m_dim <= optimal_m_dim || batch_dim % optimal_parallelism_work_amount == 0)
```
IvanNovoselov, line 95: It seems really strange that `optimal_m_dim` is hard-coded and doesn't depend on the number of threads or the actual selected block size. For example, if the shape is `[4, 60, 512]` and there are 16 threads available, it would still make sense to reshape into `[16, 15, 512]` to load all the threads. 
Moreover, is the `M` dim is too small, it's better to disable M-blocking (we don't need it in this case), and load all the threads instead.
What do you think?

src/common/snippets/src/pass/common_optimizations.cpp
```diff
@@ -68,82 +82,87 @@ void CommonOptimizations::SplitDimensionM(const std::shared_ptr<ov::snippets::op
...
         return;
 
-    const auto batch_dim =
-        std::accumulate(mm_shape.rbegin() + 2, mm_shape.rend(), size_t(1), std::multiplies<size_t>());  // B (batch)
     size_t batch_m_dim = 1;
     size_t new_m_dim = m_dim;
 
+    auto is_optimized = [&](size_t batch_m_dim, size_t new_m_dim) {
+        return batch_m_dim > 1 && new_m_dim >= static_cast<size_t>(optimal_m_dim);
```
IvanNovoselov, line 102: To add a point to the prev comment: `optimal_m_dim` defines our stopping criterion, so the choice of its value could affect performance significantly. We probably need to be more careful with it.

src/common/snippets/src/pass/common_optimizations.cpp
```diff
@@ -68,82 +82,87 @@ void CommonOptimizations::SplitDimensionM(const std::shared_ptr<ov::snippets::op
...
     const auto batch_dim_multiplier = lcm / batch_dim;  // LCM(b, nthrs) / b
     const auto needed_new_dim = m_dim / batch_dim_multiplier;  // m / (LCM(b, nthrs) / b) - needed factors of dimension m
-
-    auto is_optimized = [&](size_t batch_m_dim, size_t new_m_dim) {
-        return batch_m_dim != 1 && new_m_dim >= static_cast<size_t>(optimal_m_dim);
-    };
-
     if (batch_dim_multiplier * needed_new_dim == m_dim) {
         batch_m_dim = batch_dim_multiplier;
         new_m_dim = needed_new_dim;
```
IvanNovoselov, line 119: Minor: it seems a bit confusing that the Steps are not mutually exclusive: even if we performed decomposition on the first step, we might still fall-through and attempt to decompose again in the second step.
Can we just add smth like `&& batch_dim_multiplier != 1` to the first `if` and replace the second `if` with `else if`?

src/common/snippets/src/pass/common_optimizations.cpp
```diff
@@ -68,82 +82,87 @@ void CommonOptimizations::SplitDimensionM(const std::shared_ptr<ov::snippets::op
...
-                    dim = res;
-                } else {
-                    div++;
-                }
+        batch_m_dim = 1;
+        new_m_dim = m_dim;
+        // Heuristic value for a quick exit from the algorithm.
+        // The value shows the number of threads in percentages that perform the most equal work
+        const auto optimal_thread_num_percent = 0.8;
+        size_t opt_remainder = batch_m_dim % optimal_parallelism_work_amount;
```
IvanNovoselov, line 140: `opt_remainder` always equals to `1`, since `batch_m_dim` is 1. Can we just assign 1?

src/common/snippets/src/pass/common_optimizations.cpp
a-sidorova: We decided to leave it so in the [previous](https://github.com/openvinotoolkit/openvino/pull/18160#discussion_r1261054283) review. Your comment makes a sense, right. Thanks a lot for the idea!
As you know we implemented this optimization for special model with specific shapes. If we get new model or new shapes, we can adjust the condition. Unfortunately we don't have enough time to additionally validate this idea before release. At the moment we have a good performance and I think it's enough still. Moreover, this code was before. I tried to just update second step to not break working optimization of first step in bounds of the current PR. Thank you!

src/common/snippets/src/pass/common_optimizations.cpp
a-sidorova: The function `is_optimized` was just moved in the file. I didn't change this logic and didn't break the previous logic. I think if we have more shapes, more models, we would be able to adjust the logic more carefully or flexible. But we have what we have. Thanks!

src/common/snippets/src/pass/common_optimizations.cpp
a-sidorova: Honestly I don't understand why the steps are not exclusive.

Let's take a look at the example:

We try to optimize shapes at first step. If we find `LCM` and `m_dim % batch_dim_multiplier == 0`, we update `batch_m_dim` and `new_m_dim` here. 
Then we go to the line 134 `if (!is_optimized(batch_m_dim, new_m_dim))`. This check is just to verify that first step has already optimized. If not we try do it again using second step.

Just to remind: `is_optimized`: `return batch_m_dim > 1 && new_m_dim >= static_cast<size_t>(optimal_m_dim);`

There may be the following cases:
- We find `LCM` but unfortunately  `m_dim % batch_dim_multiplier != 0` - we don't update `batch_m_dim` and `new_m_dim` (they save previous values `1` and `m_dim`). So `is_optimized` returns `false` and we try second step. It's OK;
- We find `LCM` and  `m_dim % batch_dim_multiplier == 0`. Great! Let's think that `batch_dim_multiplier  = 1` and `new_m_dim = m_dim`  (it's possible). Anyway `is_optimized` returns `false` since `batch_m_dim = batch_dim_multiplier = 1` not `> 1`. Well, go to the second step.
- Like the previous case but `batch_dim_multiplier  = m_dim` and `new_m_dim = 1`  (it's possible as well). Anyway `is_optimized` returns `false` since `new_m_dim =  1` not `> optimal_m_dim`. Well, got the second step.
- The great case when We find `LCM`,  `m_dim % batch_dim_multiplier == 0` and `is_optimized` returns `true`. We skip second step.

I hope it's clearer now. I just show that if we optimized shapes at first step, we skip second step. Or maybe I didn't understand you point. In this case could you please eloborate? Thank you for the comment!

src/common/snippets/src/pass/common_optimizations.cpp
a-sidorova: Oh, great catch! Thanks a lot! There should be `size_t opt_remainder = (batch * batch_m_dim) % optimal_parallelism_work_amount;` (like on 145 line) because `opt_remainder` is just the maximum of possible remainders. Firstly we have shape `[batch, batch_m_dim, new_m_dim, smth]` where `batch_m_dim = 1`. So at the beginning we have parallel_work_amount `[batch, 1]`. It means that only `(batch * 1) % optimal_parallelism_work_amount` threads will work. After that we try to increase `batch_m_dim` and calculate new `remainder` as `batch_dim * divisor_0 % optimal_parallelism_work_amount` (`divisor_0` is possible new `batch_m_dim`). If now the current `remainder` is bigger than `opt_remainder`, we update `opt_remainder` and save `divisor_0` and `divisor_1` since they're potential  `batch_m_dim` and `new_m_dim`.
Thanks!

src/common/snippets/src/pass/common_optimizations.cpp
IvanNovoselov: Ok, but what if we need to support 2D shapes in future? 🙃

IvanNovoselov: Discussed offline. Agreed to leave as-is

src/common/snippets/src/pass/common_optimizations.cpp
IvanNovoselov: Discussed offline. Agreed to allow for splitting the only M block (remove `optimal_m_dim`). Need to perform validation.

src/common/snippets/src/pass/common_optimizations.cpp
IvanNovoselov: Discussed offline. Agreed to allow for splitting the only M block (remove `optimal_m_dim`). Need to perform validation.

src/common/snippets/src/pass/common_optimizations.cpp
IvanNovoselov: > I hope it's clearer now. I just show that if we optimized shapes at first step, we skip second step.
Indeed! So my suggestion is just to make it more obvious by using `else if`. Otherwise you have to dig into the logic of these checks, and it reduces readability.

src/common/snippets/src/pass/common_optimizations.cpp
a-sidorova: As we discussed offline, I removed `optimal_m_dim` and validated on model set. The results are attached to the ticket

src/common/snippets/src/pass/common_optimizations.cpp
a-sidorova: As we discussed offline, I removed `optimal_m_dim` and validated on model set. The results are attached to the ticket

src/common/snippets/src/pass/common_optimizations.cpp
a-sidorova: Changed. I really hope it's obvious now 🙂 Thanks!

IvanNovoselov (approved):
👍

a-sidorova:
@IvanNovoselov thanks a lot for the detailed review and suggested ideas!

@dmitry-gorokhov may I ask you to take a look too? Thank you!

dmitry-gorokhov (approved):

PR #18747, opened 2023-07-24
Use of share_memory parameter in tf.Graph decoder.

popovaan:
Description: 
Used "share_memory" parameter to turn on/off shared memory in tf.Graph decoder.

Ticket: 114971

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update

### DIFF
--- a/tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
+++ b/tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
@@ -688,7 +688,7 @@ class TestMoConvertTF(CommonMOConvertTest):
 
     def test_zero_copy(self, ie_device, precision, ir_version, temp_dir):
         import tensorflow as tf
-        from openvino.tools.mo import convert_model
+        from openvino.tools.ovc import convert_model
         from openvino.runtime import compile_model
         class LayerModel(tf.Module):
             def __init__(self):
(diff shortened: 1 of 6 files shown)


tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
```diff
@@ -688,7 +688,7 @@ def test_unnamed_saved_model_dir(self, ie_device, precision, ir_version, temp_di
 
     def test_zero_copy(self, ie_device, precision, ir_version, temp_dir):
         import tensorflow as tf
-        from openvino.tools.mo import convert_model
+        from openvino.tools.ovc import convert_model
```
mvafin, line 691: Do we fully switch to ovc in tests? Shouldn't we test both at least during transition period?

popovaan: OVC is the main tool that will be developed, MO will be frozen and deprecated. So it is preferable to test OVC when possible. 
Currently most tests in CI use tools.mo.convert_model. Only ovc unit tests and part of MO python API tests are switched to OVC.
Ideally we need to test both, but not sure that we should do it in CI and double CI time. It would be good to be able run tools.mo tests on demand when we change legacy functionality. Probably it is a task for validation team.

mvafin (approved):

PR #18749, opened 2023-07-24
[ONNX FE] Do not skip Constants for subgraphs

mbencer:
Ticket: 113610
Such optimization in ONNX FE blocks using additional operation after conversion (like `MakeStateful`).
Eventually, it will be applied during `compile` step while `PushConstantToSubgraph` so it shouldn't cause any difference during inference.


mbencer:
The change is blocked by CVS-116547 (Template plugin calculate incorrect results for some cases (like `INTERPRETER.onnx_if_with_only_indentity_in_else_branch`) if we skip const optmization on ONNX FE level

PR #18751, opened 2023-07-24
[Coverity] fix uninitialized scalar field

andrei-cv:
### Details:
 - *fix uninitialized scalar field*

### Tickets:
 - *CVS-116194*
 - *CVS-112759*
 - *CVS-116198*


src/plugins/intel_gpu/src/graph/layout_optimizer.cpp
```diff
@@ -1204,22 +1204,7 @@ format layout_optimizer::get_expected_format(quantize_node const& node) {
...
-                    expected = format::bfyx;
-                else
-                    expected = format::bs_fs_yx_bsv16_fsv4;
-
-                auto conv_output_layout = conv.get_output_layout();
-                auto weights_layout = conv.weights().get_output_layout().convert_to_weights_layout(conv.get_primitive()->grouped_weights_shape);
-                format expected_conv_fmt = get_expected_format(conv);
-                if (expected == format::bfyx && expected_conv_fmt == format::bs_fs_yx_bsv32_fsv32 && layout.feature() % 32 == 0)
-                    expected = expected_conv_fmt;
-            }
```
p-durandin, line 1221: @sshlyapn please take a look. It looks this block was not used sometimes but it can be useful for large batches on dGPU

sshlyapn: It looks like as a legacy code, as of now we are requesting required format from OneDNN directly instead of such heuristics

sshlyapn (approved):

PR #18753, opened 2023-07-24
add smoke test for CPU map on MacOS

wangleis:
### Details:
 - *add smoke test for CPU map on MacOS*

### Tickets:
 - *ticket-id*


xufang-lisa (approved):

sunxiaoxia2022 (approved):

src/inference/tests/unit/cpu_map_parser/parser_macos.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+        ASSERT_EQ(test_data._processors, test_processors);
+        ASSERT_EQ(test_data._numa_nodes, test_numa_nodes);
+        ASSERT_EQ(test_data._sockets, test_sockets);
+        ASSERT_EQ(test_data._cores, test_cores);
+        ASSERT_EQ(test_data._proc_type_table, test_proc_type_table);
+    }
+};
+
+MacOSCpuMapTestCase test_case_arm = {
```
dmitry-gorokhov, line 55: As usual the ask is to add comments for each parameter for the first test case. The comments should explain is it input or expected output of the test and also highlights its meaning. Otherwise it is quite hard to understand what is being tested

PR #18767, opened 2023-07-25
fix get property cannot get correct auto_batch_timeout

panhaiqi:
### Details:
 - *fix get property cannot get correct auto_batch_timeout*

### Tickets:
 - **


src/plugins/auto_batch/src/compiled_model.cpp
```diff
@@ -177,6 +177,10 @@ void CompiledModel::set_property(const ov::AnyMap& properties) {
 
 ov::Any CompiledModel::get_property(const std::string& name) const {
     auto it = m_config.find(name);
+    if (name == ov::auto_batch_timeout.name()) {
```
ilyachur, line 180: Can we add tests for this change?

PR #18768, opened 2023-07-25
[GPU] Add onednn reorder selection available at bfzyx, b_fs_zyx_fsv32.

hyunback:
Add to use onednn Reorder at bfzyx, b_fs_zyx_fsv32(5dims)

allowInputReordring is available in case float data_type.
It's the failure of unit-test unt-test. allowInputReordring white-list format is for gen9_common_conv_fwd_data_f16/f32. But convolution_kernel_b_fs_zyx_fsv16_imad also support zyx_fsv16 format and it cause an accuracy issue without padding. So we need to add the condition for only gen9_common_conv_kernel (float input only).
    
    

### Tickets:
 - *116368*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/convolution.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/convolution.cpp
@@ -195,10 +195,12 @@ public:
         }
 
         auto format = impl_param.get_output_layout().format;
-        if (format == format::b_fs_zyx_fsv16 ||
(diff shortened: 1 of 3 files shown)


isanghao (changes requested):
please check functional issues

hyunback:
Just to check whether CI can be pass.

isanghao:
Performance was good on dGPU, but I think we need further review for iGPU case. So let's postpone merging this PR beyond CF.

PR #18774, opened 2023-07-25
Fix issue with cleaning up after build openvino-dev wheels

RaguzovaTatyana:
### Details:
 - Modified custom clean up function in openvino-dev setup file

### Tickets:
 - 89845
 - 115352


artanokhov (approved):

tools/openvino_dev/setup.py
```diff
@@ -170,23 +170,23 @@ def run(self):
 class CustomClean(clean):
     """Clean up staging directories"""
 
-    def clean(self, install_cfg):
+    def clean(self,):
```
p-wysocki, line 173:
```suggestion
    def clean(self):
```

tools/openvino_dev/setup.py
```diff
@@ -170,23 +170,23 @@ def run(self):
...
             for path in paths:
                 if path.is_file() and path.exists():
                     path = path.parent
                 self.announce(f'Cleaning: {path}', level=log.INFO)
-                shutil.rmtree(path)
+                if os.path.exists(path):
+                    shutil.rmtree(path)
+
+    def run(self):
+        self.clean()
```
p-wysocki, line 189: Can this logic somehow be simplified or made more readable? Right now it reads:
`self.clean()`
`clean.run(self)`

p-wysocki (approved):
LGTM, I left two minor comments.

PR #18784, opened 2023-07-25
[Coverity] fix coverity

andrei-cv:
### Details:
 - *Fix coverity *

### Tickets:
 - *CVS-112758*
 - *CVS-114053*
 - *CVS-116203*
 - *CVS-112779*
 - *CVS-112785*


src/plugins/intel_gpu/include/intel_gpu/runtime/memory.hpp
```diff
@@ -123,7 +123,8 @@ struct simple_attached_memory : memory {
 
 template <class T, mem_lock_type lock_type = mem_lock_type::read_write>
 struct mem_lock {
-    explicit mem_lock(memory::ptr mem, const stream& stream) : _mem(mem), _stream(stream), _ptr(reinterpret_cast<T*>(_mem->lock(_stream, lock_type))) {}
+    explicit mem_lock(memory::ptr mem, const stream& stream) : _mem(std::move(mem)), _stream(stream),
+                                        _ptr(reinterpret_cast<T*>(_mem->lock(_stream, lock_type))) {}
```
Lyamin-Roman, line 127: nit: alignment

src/plugins/intel_gpu/src/graph/concatenation.cpp
```diff
@@ -154,7 +154,7 @@ concatenation_inst::typed_primitive_inst(network& network, concatenation_node co
             auto nodes_list = stack.front();
             stack.pop_front();
 
-            for (auto processed_nodes : *nodes_list) {
+            for (auto& processed_nodes : *nodes_list) {
```
Lyamin-Roman, line 157: May use a const ref in cases of changing copies to a reference in non-trivial sections of code. To avoid potential errors with data changes.

src/plugins/intel_gpu/src/graph/include/implementation_map.hpp
```diff
@@ -128,7 +128,7 @@ class implementation_map {
 
     static void add(impl_types impl_type, factory_type factory,
                     const std::vector<data_types>& types, const std::vector<format::type>& formats) {
-        add(impl_type, factory, combine(types, formats));
+        add(impl_type, std::move(factory), combine(types, formats));
```
Lyamin-Roman, line 131: Please add std::move for factory to all `add(..)` functions

src/plugins/intel_gpu/src/graph/lstm.cpp
```diff
@@ -48,7 +48,7 @@ std::string lstm_inst::to_string(lstm_node const& node) {
     lstm_info.add("weights id", weights_id);
     lstm_info.add("recurrent id", recurrent_id);
     lstm_info.add("bias id", bias_id);
-    lstm_info.add("peepholes id", peepholes_id);
+    lstm_info.add("peepholes id", std::move(peepholes_id));
```
Lyamin-Roman, line 51: Does it make sense to move small strings? And if it makes sense, then why `peepholes_id` moved only?
Same question for other changes in `to_string(..)` method

src/plugins/intel_gpu/src/graph/program.cpp
```diff
@@ -180,7 +180,7 @@ program::program(engine& engine_ref,
     : _engine(engine_ref),
       _stream(_engine.create_stream(config)),
       _config(config),
-      _task_executor(task_executor),
+      _task_executor(std::move(task_executor)),
```
Lyamin-Roman, line 183: Please add `std::move` for `task_executor` in other ctr too

src/plugins/intel_gpu/src/plugin/ops/gather.cpp
```diff
@@ -102,7 +102,8 @@ void CreateGatherOpBase(Program& p, const std::shared_ptr<T>& op, const int64_t
         const auto& indices_node = indices.get_node_shared_ptr();
         auto indices_constant = std::dynamic_pointer_cast<ngraph::op::v0::Constant>(indices_node);
         float result = 0.f;
-        ov::op::util::get_single_value(indices_constant, result);
+        if (!ov::op::util::get_single_value(indices_constant, result))
```
Lyamin-Roman, line 105: It's better to use `OPENVINO_ASSERT` instead `IE_THROW()`

Lyamin-Roman:
```suggestion
        OPENVINO_ASSERT(ov::op::util::get_single_value(indices_constant, result),
                        "Unsupported indices node in ", op->get_friendly_name(), " (", op->get_type_name(), ")");
```

src/plugins/intel_gpu/include/intel_gpu/graph/fused_primitive_desc.hpp
```diff
@@ -21,7 +21,7 @@ class NodeFuseParams {
 };
 
 struct fused_primitive_desc {
-    explicit fused_primitive_desc(std::shared_ptr<const primitive> prim) : desc(prim) {}
+    explicit fused_primitive_desc(std::shared_ptr<const primitive> prim) : desc(std::move(prim)) {}
```
sshlyapn, line 24: To reduce the number of `std::move`s and improve code readability I'd suggest to use passing by const reference instead of passing by value in all cases where it is possible. And then we will have only single copy c-tr call (as when `std::move` using) and better code design to my mind
```suggestion
    explicit fused_primitive_desc(const std::shared_ptr<const primitive>& prim) : desc(prim) {}
```

Lyamin-Roman (approved):
LGTM, but I think that std::move for small strings (ex: in lstm_inst::to_string method) is redundant

PR #18792, opened 2023-07-26
[GPU] Add new debug config OV_GPU_DisableMemoryReuse

yeonbok:
### Details:
 - DisableMemoryReuse config for debug purpose

### Tickets:
 - 115838

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
@@ -122,6 +122,7 @@ public:
     int disable_async_compilation;                  // Disable async compilation
     int disable_dynamic_impl;                       // Disable dynamic implementation
     int disable_runtime_buffer_fusing;              // Disable runtime buffer fusing
+    int disable_memory_reuse;                   // Disable memmory reuse among layers
     std::set<int64_t> dump_iteration;               // Dump n-th execution of network.
     std::vector<std::string> load_layers_raw_dump;  // List of layers to load dumped raw binary and filenames
     static const debug_configuration *get_instance();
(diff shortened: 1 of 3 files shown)


src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
```diff
@@ -120,6 +120,7 @@ class debug_configuration {
     int disable_async_compilation;              // Disable async compilation
     int disable_dynamic_impl;                   // Disable dynamic implementation
     int disable_runtime_buffer_fusing;          // Disable runtime buffer fusing
+    int disable_memory_reuse;                   // Disable memmory reuse among layers
```
andrew-k-park, line 123: nit:
```suggestion
    int disable_memory_reuse;                   // Disable memory reuse among layers
```

e-ddykim (approved):

andrew-k-park (approved):

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1180,6 +1180,10 @@ memory::ptr primitive_inst::allocate_output(engine& _engine, memory_pool& pool,
     if (_node.is_in_shape_of_subgraph())
         reusable_across_network = false;
 
+    GPU_DEBUG_GET_INSTANCE(debug_config);
+    GPU_DEBUG_IF(debug_config->disable_memory_reuse) {
+        reusable_across_network = false;
```
sshlyapn, line 1185: Wouldn't it be better to force `ov::intel_gpu::enable_memory_pool` property in `void ExecutionConfig::apply_debug_options`?

yeonbok: At first I thought it too, but maybe it will hide pool's allocation behavior itself, so I thought "wehther to reuse or not" and "whether to use pool or not" should be handled separately. how do you think?

sshlyapn: @yeonbok hmm, yeah, it makes sense, thanks for explanation!

yeonbok: Thanks!

PR #18797, opened 2023-07-26
[GPU] Do not add post reorder for inputs if data type does not change 

andrew-k-park:
### Details:
 - Do not add post type conversion reorder for inputs if mean value is not specified and the data type does not change

### Tickets:
 - 116261

### DIFF
--- a/src/plugins/intel_gpu/src/plugin/graph.cpp
+++ b/src/plugins/intel_gpu/src/plugin/graph.cpp
@@ -367,6 +367,10 @@ std::shared_ptr<ngraph::Function> Graph::GetExecGraphInfoByPrimitivesInfo(std::v
             auto param = std::make_shared<ngraph::op::Parameter>(out_et, out_pshape);
             params.push_back(param);
             return_node = param;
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/plugin/graph.cpp
```diff
@@ -367,6 +367,10 @@ std::shared_ptr<ngraph::Function> Graph::GetExecGraphInfoByPrimitivesInfo(std::v
             auto param = std::make_shared<ngraph::op::Parameter>(out_et, out_pshape);
             params.push_back(param);
             return_node = param;
+            // create additional result node if parameter is output without post reorder
+            if (is_output) {
```
yeonbok, line 371: Where did you observed such case parameter == is_output?

andrew-k-park: image
In some test case of FakeQuantizeTransformation, there was case where parameter was also output directly by not inserting the post reorder.

yeonbok (approved):

PR #18809, opened 2023-07-27
[AUTO Plugin] Update document for testing AUTO plugin from benchmark app

yangwang201911:
### Details:
 - Update guide document to show the detailed steps of testing AUTO plugin from benchmark app

### Tickets:
 - 115641


src/plugins/auto/src/auto_compiled_model.cpp
```diff
@@ -106,8 +106,10 @@ ov::Any AutoCompiledModel::get_property(const std::string& name) const {
         const unsigned int default_num_for_latency = 1u;
         unsigned int real = 0;
         if (m_scheduler->m_compile_context[ACTUALDEVICE].m_is_already) {
-            real = m_scheduler->m_compile_context[ACTUALDEVICE].
-                m_compiled_model->get_property(name).as<unsigned int>();
+            real = m_scheduler->m_compile_context[ACTUALDEVICE].m_compiled_model->get_property(name).as<unsigned int>();
```
peterchen-intel, line 109: Please exclude these debug changes.

src/plugins/auto/docs/tests.md
```diff
@@ -49,4 +49,270 @@ Note: Google Test filter = *AutoReleaseHelperTest*cpuLoadFailure_accelerateorLoa
...
+```
+
+### Device selection fallback of the AUTO
+
+This section will show the fallback of device selection within the AUTO plugin if the device with high priority doesn't support the precision of the inputting model. For example, CPU supports both FP16 and FP32 precision model, while GNA doesn't FP32 precision. Although GNA has higher priority, AUTO plugin will ultimately choose the CPU plugin to load model with FP32 precision.  
+
+AUTO will select GNA if no other device is specified in the device candidate list.
+
+```bash
+openvino/bin/intel64/Release$ ./benchmark_app -m openvino/src/core/tests/models/ir/add_abc.xml -d AUTO:GNA -t 10
```
peterchen-intel, line 261: Any reason we don't use GPU as example?

yangwang201911: GPU supports the model precision with both FP16 and FP32, which means selection fallback will not happen here if we specified `-d AUTO:GPU,CPU` for the model with precision FP32(add_abc.xml). GNA doesn't supports FP32 and we can see the selection fallback here when inputting `-d AUTO:GNA,CPU` for FP32 model.

wangleis (approved):
This is test introduction base on benchmark app which is close to user application.

PR #18811, opened 2023-07-27
[GPU] remove unroll to void the register spilling issue.

sungeunk:
### Tickets:
 - 112952


isanghao:
This is a too aggressive rolling of loop. Let's wait for fix from compiler team :(

PR #18813, opened 2023-07-27
TorchFX caching fix

cavusmustafa:
- Signature generation updates for in-memory caching and IR saving/loading in TorchFX backend.
- Added function to clear all caches in TorchFX backend.

### DIFF
--- a/src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/execute.py
+++ b/src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/execute.py
@@ -153,6 +153,7 @@ def openvino_execute_partitioned(gm: GraphModule, *args, executor_parameters=Non
                                                          model_hash_str=model_hash_str)
     return partitioned_modules[signature](*args)
 
+
 def clear_caches():
     global partitioned_modules
     global compiled_cache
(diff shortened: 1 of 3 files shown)


slyalin (approved):

src/bindings/python/src/openvino/frontend/pytorch/torchdynamo/execute.py
```diff
@@ -143,9 +143,19 @@ def openvino_execute_partitioned(gm: GraphModule, *args, executor_parameters=Non
...
         partitioned_modules[signature] = partition_graph(gm, use_python_fusion_cache=use_python_fusion_cache,
                                                          model_hash_str=model_hash_str)
     return partitioned_modules[signature](*args)
+
+def clear_caches():
+    global partitioned_modules
+    global compiled_cache
+
+    compiled_cache.clear()
+    partitioned_modules.clear()
```
slyalin, line 161: I hope we will leverage a callable, which we are returning from the backend, as a point where we are attaching all items required for caching and will get rid off this global caches completely. Don't really want to expose `reset` and `clear_caches` functions at the user-level because it explicitly exposes some global states in our solution that is not very aligned with what `torch.compile` provides (it literally returns an optimized model object without any hint about global caches).

mvafin (approved):

mvafin:
build_jenkins

PR #18818, opened 2023-07-27
update streams calculation for latency mode on MacOS

wangleis:
### Details:
 - *update streams calculation for latency mode on MacOS to use all CPU on big core/small core platform*

### Tickets:
 - *ticket-id*


xufang-lisa (approved):

sunxiaoxia2022 (approved):

src/plugins/intel_cpu/src/cpu_streams_calculation.cpp
```diff
@@ -410,10 +410,16 @@ int get_model_prefer_threads(const int num_streams,
             const int int8_threshold = 4;  // ~relative efficiency of the VNNI-intensive code for Big vs Little cores;
             const int fp32_threshold = 2;  // ~relative efficiency of the AVX2 fp32 code for Big vs Little cores;
             // by default the latency case uses (faster) Big cores only, depending on the compute ratio
+#ifdef __APPLE__
+            if ((proc_type_table.size() == 1) && (proc_type_table[0][EFFICIENT_CORE_PROC] > 0)) {
```
dmitry-gorokhov, line 414: In fact this change returns us to the state of 23.0 release, so good to go as is.
However as follow-up step we need to find best hieristics assuming different mac configurations (like 8+2).

dmitry-gorokhov (approved):

PR #18824, opened 2023-07-27
[CPU] Extensions shape inference issue fix

v-Golubev:
### Details:
 - *Fixed `get_port_mask()` behavior of `NgraphShapeInfer`: only port mask which was set by plugin node is returned, mask from `m_shape_infer` is ignored*

### Tickets:
 - *CVS-116464*

### DIFF
--- a/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_ngraph.hpp
+++ b/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_ngraph.hpp
@@ -33,7 +33,7 @@ public:
         return m_shape_infer->get_pads_end();
     }
     port_mask_t get_port_mask() const override {
-        return m_port_mask & m_shape_infer->get_port_mask();
+        return m_port_mask;
     }
 private:
     std::shared_ptr<IStaticShapeInfer> m_shape_infer;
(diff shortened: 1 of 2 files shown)


praasz (approved):

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_ngraph.hpp
```diff
@@ -33,7 +33,7 @@ class NgraphShapeInfer : public IShapeInfer {
         return m_shape_infer->get_pads_end();
     }
     port_mask_t get_port_mask() const override {
-        return m_port_mask & m_shape_infer->get_port_mask();
```
dmitry-gorokhov, line 36: @v-Golubev  @praasz  I am not sure I fully understand the fix.
Lets imagine we don't have own CPU shape inference abstraction layer (`IShapeInfer`) and directly use static shape inference abstraction layer (`IStaticShapeInfer`). 
So for custom operation `make_shape_inference` should return instance of `ShapeInferFallback` which always return zero port mask. However since we don't know semantics of custom op in advance shouldn't it return full port mask to make sure all the data will be accessible?
In other words this patch just hides incorrect behavior of static shape inference. Please correct me if my understanding is wrong.

v-Golubev: @dmitry-gorokhov I might be wrong but `ShapeInferFallback` can be used not only for custom ops but also for the ops for which we don't have plugin shape inference. And for this category of nodes we can say which inputs are required for output shapes computation . In this case, we don't need to set full mask.
So looks like we need to rely on `m_port_mask` which is set in plugin node constructors

dmitry-gorokhov (approved):
Approved as reasonable solution on CPU plugin level, but need to follow-up on static shape inference behavior

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference_ngraph.hpp
praasz: I think I introduce this change too soon (as it pass pre-commit test I left it but it seems not all scenarios are tested).
The intention is to use mask from `IStaticShapeInfer` as for supported operators this value should be correct. 
But it looks like when `ShapeInferFallback` used it (it returns mask `0` now) this approach is not correct. I need to do more  investigation what value should be in fallback (e.g. be full mask and then masked with input mask or has possibility to be set by plugin).

PR #18827, opened 2023-07-27
[PT FE] Partially disable freezing for int8 and uint8 weights

mvafin:
### Details:
 - *Partially disable freezing for int8 and uint8 weights*

### Tickets:
 - *115883*


AlexKoff88:
We should consider adding a test here, e.g. using this [model](https://huggingface.co/hf-internal-testing/tiny-random-OPTModel). It is created for testing and should be fast.

mvafin:
> We should consider adding a test here, e.g. using this [model](https://huggingface.co/hf-internal-testing/tiny-random-OPTModel). It is created for testing and should be fast.

There is already a test for it https://github.com/openvinotoolkit/openvino/blob/223461e4620533024e4985a04172b960bd2c4c6e/tests/layer_tests/mo_python_api_tests/test_mo_convert_pytorch.py#L789. I am not sure if we want to have hf dependency in tests

slyalin (approved):

PR #18832, opened 2023-07-28
Apply global mutex only for old glibc

ilyachur:
### Details:
 - Fixed leftovers after https://github.com/openvinotoolkit/openvino/pull/18700
 - Use global mutex only for old gcc
 - Enable CoreThreading tests for macOS and other platforms which don't have gcc

### Tickets:
 - *ticket-id*


nkogteva (approved):

ilya-lavrenov (approved):

vurusovs (approved):
Could you add some detailed comment about why mutex is added?

PR #18835, opened 2023-07-28
Fix Reduce node bug regarding fusing with low precision

xuchen-intel:
### Details:
 - *Fix Reduce node bug regarding fusing with low precision.*

### Tickets:
 - *CVS-115506*
 - *CVS-116115*
 - *CVS-116792*

### DIFF
--- a/src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
+++ b/src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
@@ -6,6 +6,7 @@
 #include "shared_test_classes/single_layer/reduce_ops.hpp"
 #include "test_utils/cpu_test_utils.hpp"
 #include "test_utils/fusing_test_utils.hpp"
+#include "lpt_ngraph_functions/common/builders.hpp"
 
 using namespace InferenceEngine;
 using namespace CPUTestUtils;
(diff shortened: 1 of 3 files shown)


xuchen-intel:
@dmitry-gorokhov Hi Dmitry, could you please take a look?

dmitry-gorokhov:
@xuchen-intel Is it possible to extend tests to cover the issue?

xuchen-intel:
> @xuchen-intel Is it possible to extend tests to cover the issue?

@dmitry-gorokhov Hi Dmitry, I tried to extend the test cases. It seems that the current test infrastructure does not support this scenario of fusing with low precision (i.e. Reduce node output precision fp32, fusing output precision int8). Please correct me, if I'm wrong. I had checked the fusing test cases, the output precision of fusing is always fp32, so the code branch with issue is not activated. Further discussions are always welcomed. Thanks dmitry!

dmitry-gorokhov:
> > @xuchen-intel Is it possible to extend tests to cover the issue?
> 
> @dmitry-gorokhov Hi Dmitry, I tried to extend the test cases. It seems that the current test infrastructure does not support this scenario of fusing with low precision (i.e. Reduce node output precision fp32, fusing output precision int8). Please correct me, if I'm wrong. I had checked the fusing test cases, the output precision of fusing is always fp32, so the code branch with issue is not activated. Further discussions are always welcomed. Thanks dmitry!

Two ideas comes to my mind:
1. Try to force model ouput precision to int8. We do have optimization that propagated model output precision to the last node https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/graph.cpp#L328
2. Implement Subgraph test (like Reduce -> FQ -> Tranpose to prevent dequantization fusion into Reduce).

xuchen-intel:
> > > @xuchen-intel Is it possible to extend tests to cover the issue?
> > 
> > 
> > @dmitry-gorokhov Hi Dmitry, I tried to extend the test cases. It seems that the current test infrastructure does not support this scenario of fusing with low precision (i.e. Reduce node output precision fp32, fusing output precision int8). Please correct me, if I'm wrong. I had checked the fusing test cases, the output precision of fusing is always fp32, so the code branch with issue is not activated. Further discussions are always welcomed. Thanks dmitry!
> 
> Two ideas comes to my mind:
> 
> 1. Try to force model ouput precision to int8. We do have optimization that propagated model output precision to the last node https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/graph.cpp#L328
> 2. Implement Subgraph test (like Reduce -> FQ -> Tranpose to prevent dequantization fusion into Reduce).

@dmitry-gorokhov Great suggestions! I've added test cases by appending FQ with Tranpose node, which reproduced the issue in the first place. Thanks Dmitry!

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
```diff
@@ -513,6 +546,13 @@ INSTANTIATE_TEST_SUITE_P(
         ReduceCPULayerTest::getTestCaseName
 );
 
+INSTANTIATE_TEST_SUITE_P(
```
dmitry-gorokhov, line 549: Ideally we should explain why such custom test cases is needed. For example here we test post-ops fusing case with U8/I8 precision on output since Transpose prevents dequantization part to be fused back into Reduce.

dmitry-gorokhov (approved):

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/x64/reduce.cpp
xuchen-intel: Good point! Description added. Thanks Dmitry!

PR #18836, opened 2023-07-28
[GPU] Allow StridedSlice as predecessor for in place concat

andrew-k-park:
### Details:
 - Allow StridedSlice as predecessor for in place concat optimization during runtime
 - Enable padding support for StridedSlice

### Tickets:
 - 112029


yeonbok:
Could you add dynamic buffer fusing test w/ strided_slice too?

andrew-k-park:
> Could you add dynamic buffer fusing test w/ strided_slice too?

Added prepare_buffer_fusing TC w/ strided_slice on ov_gpu_unit_tests

yeonbok (approved):

PR #18843, opened 2023-07-28
"static_shape" parameter fix.

popovaan:
Root cause analysis: 
Shape of sub graphs are not folded during constant folding, even if 'static_shape' parameter is turned on.

Solution: 
Pass 'static_shape' value to apply_moc_transformations() and set following flag:
**cf** - determines if model shapes should be used during transformations. Set it true if "static_shape" is on.

Change OVC behavior as if "static_shape" is always on. 

Ticket: 109919

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update


jane-intel (approved):

tools/mo/openvino/tools/mo/back/offline_transformations.py
```diff
@@ -88,9 +88,9 @@ def apply_user_transformations(func: object, transforms: list):
         available_transformations[name](func, **args)
 
 
-def apply_moc_transformations(func: object):
+def apply_moc_transformations(func: object, static_shape):
```
rkazants, line 91:
```suggestion
def apply_moc_transformations(func: object, static_shape: bool):
```

rkazants: Does anyone use this particular function? I mean POT or other tools since we are changing signature of this function?

tools/ovc/openvino/tools/ovc/moc_frontend/offline_transformations.py
```diff
@@ -89,7 +89,7 @@ def apply_user_transformations(func: object, transforms: list):
 
 def apply_moc_transformations(func: object):
     from openvino._offline_transformations import apply_moc_transformations  # pylint: disable=import-error,no-name-in-module
-    apply_moc_transformations(func, cf=False, smart_reshape=True)
+    apply_moc_transformations(func, cf=True, smart_reshape=False)
```
rkazants, line 92: do we need this wrapper for `apply_moc_transformations`? Why can't we directly call it?

tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
```diff
@@ -611,6 +611,84 @@ def __call__(self, input1, input2):
 
     return model, model_ref, {}
 
+def shape_of_const_fold_test(temp_dir):
```
rkazants, line 614: Why these functions are not added into `class TestMoConvertTF(CommonMOConvertTest)`?

tools/mo/openvino/tools/mo/back/offline_transformations.py
popovaan: Yes, they are most likely used by optimum-intel. Should we create a new method with new signature?

popovaan: Used `apply_moc_transformations()` directly from `openvino._offline_transformations`, so now we don't need to change this method.

tools/ovc/openvino/tools/ovc/moc_frontend/offline_transformations.py
popovaan: No, we don't need it in OVC. Removed it.

tests/layer_tests/mo_python_api_tests/test_mo_convert_tf.py
popovaan: I didn't commit the line where they were added to list of tests. Corrected it, thanks!

rkazants (approved):

PR #18858, opened 2023-07-28
Load extensions

slyalin:
Enable using custom operations loaded from extension library in Python by adding new add_extension method to NodeFactory class. Needed to support tokenizer PR in openvino-contrib: https://github.com/openvinotoolkit/openvino_contrib/pull/687.

TODO:
- Add tests
- Decide whether legacy part of Python API needs to be changed
- Add diagnostics in the second NodeFactory::create method if user tries to create op in that way which is impossible now


slyalin:
Here is how to use it in a simple example that uses 3 custom ops from the library with ops added in https://github.com/openvinotoolkit/openvino_contrib/pull/687.

image

src/bindings/python/tests/test_graph/test_node_factory.py
```diff
@@ -92,3 +94,27 @@ def test_node_factory_validate_missing_arguments():
...
+    model = Model([identity], [data])
+    compiled = compile_model(model)
+    tensor = np.array([[3, 4]], dtype=np.float32)
+    result = compiled(tensor)
+
+    # TODO: There is an issue with life time of objects, free resources explicitly
+    # otherwise segfault will occur. Workaround: create factory as a global variable.
+    del compiled
+    del model
+    del identity
```
slyalin, line 118: Looks like I don't know how to hold the library until all the users disappear. @ilyachur, looks like this is a common problem of extensions, how can I solve it?

slyalin: I tried to hold `SOExtension` object in `rt_info` for the node created from that extension, but it didn't help presumably because the operation object itself needs to be destroyed before that RT info tag is destroyed, but real sequence is obviously reversed.

slyalin: I can try to temporarily hold the library in `ov::Model` destructor while all the nodes are being destroyed. But it won't help if someone holds a shared pointer to the op from the library somewhere else, outside the normal `ov::Model` graph, or in case when such a node is destroyed during the transformations -- too many places where we need to care about it.

Another solution is to collect them in a global static object, but it would mean extension libraries are never unloaded (not a big problem for "normal" applications though).

@jiwaszki, @ilya-lavrenov, @ilyachur, would appreciate your suggestions.

apaniukov:
I checked this branch on tokenizers PR, everything works fine:
https://github.com/apaniukov/openvino_contrib/blob/239acc469260d02673d501ec885a51ffdd953ced/modules/custom_operations/user_ie_extensions/tokenizer/tokenizer_pipeline.py#L16

slyalin:
Not sure that the added test is really running in the pre-commit, but CI is passed and this test is passed on my local system when running manually.

PR #18859, opened 2023-07-29
[GPU] Skip reorder at runtime if data type and format are not changed

yeonbok:
### Details:
 - Skip reorder at runtime if it is needless to run
### Tickets:
 - 116688

### DIFF
--- a/src/plugins/intel_gpu/src/graph/include/reorder_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/reorder_inst.h
@@ -88,7 +88,7 @@ public:
     void update_output_memory() override;
     bool requires_reinterpret() const {
         if (input_memory().get_layout() != _impl_params->get_output_layout()) {
-            return true;
+            _req_reinterpr = true;
         }
         return _req_reinterpr;
     }
--- a/src/plugins/intel_gpu/src/graph/include/reorder_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/reorder_inst.h
@@ -87,8 +87,6 @@ public:
 
     void update_output_memory() override;
     bool requires_reinterpret() const {
-        if (!get_node().is_dynamic())
-            return _req_reinterpr;
         if (input_memory().get_layout() != _impl_params->get_output_layout()) {
             return true;
         }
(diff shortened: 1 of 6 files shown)


src/plugins/intel_gpu/tests/unit/dynamic_execution/skip_redundant_reorder_at_runtime.cpp
```diff
@@ -0,0 +1,52 @@
...
+    auto weight_mem = engine.allocate_memory({{2, 32}, data_types::f32, format::bfyx});
+    std::vector<float> weight_data(weight_mem->get_layout().count());
+    std::iota(weight_data.begin(), weight_data.end(), 1.0f);
+    set_values(weight_mem, weight_data);
+
+    auto input_l = layout{ov::PartialShape::dynamic(2), data_types::f32, format::bfyx};
+    topology topology(input_layout("input", input_l),
+                      data("weight", weight_mem),
+                      fully_connected("fc", input_info("input"), {"weight"}, "", data_types::f32),
+                      reorder("reorder", input_info("fc"), format::bfyx, data_types::f32)); /*output padding*/
```
kelvinchoi-intel, line 33: How about adding another reorder which is not output? I suggest this because output reorder is sometimes handled as specific case.

yeonbok: Skipping output reorder is my intention for dynamic case.

src/plugins/intel_gpu/src/graph/include/reorder_inst.h
```diff
@@ -86,7 +86,12 @@ class typed_primitive_inst<reorder> : public typed_primitive_inst_base<reorder>
     bool has_mean() const { return !get_typed_desc<reorder>()->mean.empty(); }
 
     void update_output_memory() override;
-    bool requires_reinterpret() const { return _req_reinterpr; }
+    bool requires_reinterpret() const {
+        if (input_memory().get_layout() != _impl_params->get_output_layout()) {
+            return true;
```
e-ddykim, line 91:
```suggestion
            _req_reinterpr = true;
```
Is this possible to change it as above for consistency?

yeonbok: For static shapes, this is already checked and set at build time.
(At that time, the check using _impl_params is same as the original result too)
For dynamic shape, it should be checked at runtime using _impl_params. 
But basically in my opinion, there is no problem even if we do reinterpretion always ..

yeonbok: Ah I thought you asked about consistency but you suggested just a source level change. No problem, I will apply it.

e-ddykim (approved):

PR #18861, opened 2023-07-29
[GPU] Check if padding is supported for static shape when buffer fusing

davidsnam-intel:
### Details:
 - *is_padding_supported() should be checked even for static shape when buffer fusing*

### Tickets:
 - *112977*

### DIFF
--- a/src/plugins/intel_gpu/tests/unit/passes/prepare_buffer_fusing_test.cpp
+++ b/src/plugins/intel_gpu/tests/unit/passes/prepare_buffer_fusing_test.cpp
@@ -861,19 +861,6 @@ TEST(prepare_buffer_fusing, test_checking_padding_supported) {
     auto in_layout2 = layout{ ov::PartialShape{2, 72, 57, 57}, data_types::f16, format::fs_b_yx_fsv32};
     auto in_layout3 = layout{ ov::PartialShape{2, 144, 57, 57}, data_types::f16, format::fs_b_yx_fsv32};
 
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/tests/unit/passes/prepare_buffer_fusing_test.cpp
```diff
@@ -852,6 +855,50 @@ TEST(prepare_buffer_fusing, test_implicit_crop_and_outerpadding_deconv) {
...
+    auto in_layout3 = layout{ ov::PartialShape{2, 144, 57, 57}, data_types::f16, format::fs_b_yx_fsv32};
+
+    auto in_mem1 = engine.allocate_memory(in_layout1);
+    auto in_mem2 = engine.allocate_memory(in_layout2);
+    auto in_mem3 = engine.allocate_memory(in_layout3);
+
+    random_generator rg(GET_SUITE_NAME);
+    auto in_data1 = rg.generate_random_4d<FLOAT16>(2, 36, 57, 57, -1, 1);
+    auto in_data2 = rg.generate_random_4d<FLOAT16>(2, 72, 57, 57, -1, 1);
+    auto in_data3 = rg.generate_random_4d<FLOAT16>(2, 144, 57, 57, -1, 1);
```
yeonbok, line 871: Very minor thing, but if you are not actually run the network but just want to check the nodes' optimization, we dont need to generate these values and set to memory.

yeonbok (approved):

PR #18863, opened 2023-07-30
update stream info for mixing MAIN_CORE_PROC and HYPER_THREADING_PROC

wangleis:
### Details:
 - *refactoring get_streams_info_table() for code reuse*
 - *update stream info for mixing MAIN_CORE_PROC and HYPER_THREADING_PROC*

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_cpu/src/cpu_streams_calculation.cpp
+++ b/src/plugins/intel_cpu/src/cpu_streams_calculation.cpp
@@ -38,30 +38,49 @@ std::vector<std::vector<int>> get_streams_info_table(const int input_streams,
     std::vector<std::vector<int>> proc_socket_table;
 
     int n_streams = 0;
+    int n_threads = 0;
     int n_threads_per_stream = 0;
-
-    auto update_mix_stream_info = [&]() {
-        stream_info[NUMBER_OF_STREAMS] = 0;
-        int n_threads = stream_info[THREADS_PER_STREAM];
-        for (int n = MAIN_CORE_PROC; n <= HYPER_THREADING_PROC; n++) {
-            if (0 != proc_type_table[0][n]) {
-                stream_info[PROC_TYPE] = n;
-                if (n_threads <= proc_type_table[0][n]) {
-                    stream_info[THREADS_PER_STREAM] = n_threads;
-                    streams_info_table.push_back(stream_info);
-                    break;
-                } else {
-                    stream_info[THREADS_PER_STREAM] = proc_type_table[0][n];
-                    streams_info_table.push_back(stream_info);
-                    n_threads -= proc_type_table[0][n];
-                }
-            }
-        }
-    };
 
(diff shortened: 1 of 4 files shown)


src/plugins/intel_cpu/tests/unit/streams_info/streams_info_table_test.cpp
```diff
@@ -90,7 +94,15 @@ StreamsCalculationTestCase _2sockets_104cores_latency_platform_3 = {
      {52, 26, 0, 26, 1, 0},
      {52, 26, 0, 26, 2, 1},
      {52, 26, 0, 26, 3, 1}},
-    {{1, MAIN_CORE_PROC, 208, -1, -1}},
+    {{1, ALL_PROC, 208, -1, -1},
```
sunxiaoxia2022, line 97: There is no ECore used in this situation. Is ALL_PROC also used in this case?

wangleis: Now ALL_PROC is used for all mixed streams, even if the core type is same but numa nodes or sockets are mixed. Then the following function can use it to identify mix stream and parse detail threads on each numa node. 

Changing to MAIN_CORE_PROC is doable. May I know if there is any benefit for StreamExecutor?

src/plugins/intel_cpu/src/cpu_streams_calculation.cpp
```diff
@@ -73,19 +93,33 @@ std::vector<std::vector<int>> get_streams_info_table(const int input_streams,
...
-            proc_socket_table[one_proc_info[PROC_SOCKET_ID]][proc_type] -=
-                stream_info[NUMBER_OF_STREAMS] * stream_info[THREADS_PER_STREAM];
+    auto check_threads_per_stream = [&]() {
+        int count = 0;
+        while (count < n_streams) {
+            for (int n_type = MAIN_CORE_PROC; n_type <= HYPER_THREADING_PROC; n_type++) {
+                count += static_cast<int>(proc_type_table[0][n_type] / n_threads_per_stream);
+            }
+            if (((count == 0) && (proc_type_table.size() > 1)) || (count >= n_streams)) {
+                return;
```
riverlijunjie, line 113: It seem `count >= n_streams` is duplicated with `while (count < n_streams)`

riverlijunjie: `((count == 0) && (proc_type_table.size() > 1)` should be abnormal condition, do we need handle it rather than return directly?

wangleis: updated

xufang-lisa (approved):

sunxiaoxia2022 (approved):

riverlijunjie (approved):

dmitry-gorokhov (approved):

PR #18865, opened 2023-07-31
[CPU] Share reorders

usstq:
### Details:
 - redundant reorders (from same source node with same destination layout & precision) are frequently observed in real models, `ShareReorders()` is introduced into `ApplyImplSpecificGraphOptimizations` to optimize them.
 - with above change, reorder node now can have multiple children which fails `MergeTransposeAndReorder()` transformation's assumption, so enhance `MergeTransposeAndReorder()` to handle cases where the [transpose+reorder] operation has multiple children.
 - this PR also allows @jane-intel 's PR https://github.com/openvinotoolkit/openvino/pull/18775 to pass CPU functional test: `FuseTransposeAndReorderTest4`, to further optimize the performance for cases where transpose nodes (before reorder node) can be also shared.

### Tickets:
 - CVS-127706


dmitry-gorokhov:
@usstq Two general comments:
1. We need to add subgraph test that will verify the behavior based on runtime graph
2. Haven't you analyzed possibility to implement such kind of optimization as transformation inside this block: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/graph_optimizer.cpp#L190. Smt like horizontal reorsers fusion. Maybe it is not that straitforward as first glance, but from my understanding separation of this optimization logic makes the pipeline more reliable. 

@maxnick Would be nice if you can take a look as well

yuxu42:
@zhangYiIntel Could you please take a review? Thanks!

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2489,14 +2493,45 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
+        auto finalReorder = reorderNode;
         if (inPrec != outPrec) {
             auto reorderInDesc2 = reorderOutDesc;
             auto reorderOutDesc2 = outDesc;
 
             std::string reorderLayerName2 = reorderNode->getName() + "_" +
                                     Reorder::getReorderArgs(*reorderInDesc2, *reorderOutDesc2) + "_" + childChildNode->getName();
 
-            graph.InsertReorder(reorderNode->getChildEdgeAt(0), reorderLayerName2, *reorderInDesc2, *reorderOutDesc2, false);
+            finalReorder = graph.InsertReorder(reorderNode->getChildEdgeAt(0), reorderLayerName2, *reorderInDesc2, *reorderOutDesc2, false);
```
zhangYiIntel, line 2504: Why not just extend the insertReoder to insert reorders to all the children ? Inserting one reorder then connecting to others needs more checks.

zhangYiIntel (approved):
LGTM, please add tests to cover the reordering merging case and the case 2 in TransposeReorder

src/plugins/intel_cpu/src/graph_optimizer.cpp
usstq: good idea, but `insertReoder`'s API limited this possibility, I will try to avoid using `insertReoder` and refactor this logic according to your idea!

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2415,6 +2414,9 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
         auto parentParentConstNode = parentNode->getParentEdgesAtPort(1)[0]->getParent();
         auto childChildNode = childNode->getChildEdgeAt(0)->getChild();
 
+        // pp => transpose0 => reorder0 => cc0, cc1, ...
```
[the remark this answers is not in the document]
usstq: `FuseTransposeAndReorderTest5` with shared transpose is added

src/plugins/intel_cpu/src/graph_optimizer.cpp
usstq: this block of code is refactored according to your comment, with the help of slightly adjusted `Graph::InsertNode` which allows initializing node w/o creating parent & child edges, .

usstq:
> @usstq Two general comments:
> 
> 1. We need to add subgraph test that will verify the behavior based on runtime graph
> 2. Haven't you analyzed possibility to implement such kind of optimization as transformation inside this block: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/graph_optimizer.cpp#L190. Smt like horizontal reorsers fusion. Maybe it is not that straitforward as first glance, but from my understanding separation of this optimization logic makes the pipeline more reliable.
> 
> @maxnick Would be nice if you can take a look as well

1. FuseTransposeAndReorderTest5 is added as a behavior test.
2. Move `reuseReorderForReadonlyChildren` into a new CPU graph transform is possible, however requires more complex checking for inplace conflict, imagine some of the sibling reorders may be inserted by logic at the end of `Graph::ResolveEdgeConflicts()` to solve inplace conflict and cannot be removed or shared, and it's not easy to detect that. 
do reoder sharing here has no such concern. and it eliminates the redundant reorders from the beginning make following transform easier.

usstq:
@maxnick Can you help to review this PR? thanks!

usstq:
@dmitry-gorokhov Can you help to review this PR since Maksim is on vocation.

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1598,31 +1630,42 @@ bool Graph::InsertNode(EdgePtr edge, NodePtr node, bool initNode) {
...
+            resolveInPlaceDirection(node);
+        }
+        for (auto& node : new_nodes) {
+            node->initOptimalPrimitiveDescriptor();
+        }
+    }
+    for (auto& node : new_nodes) {
+        graphNodes.push_back(node);
+    }
+    node->updateConstantType();
```
jane-intel, line 1651: node is unavailable in this scope. did you mean smth else?

usstq: Sorry, fixed.

dmitry-gorokhov:
Looks like worth to update the PR after https://github.com/openvinotoolkit/openvino/pull/21847 merge. Graph modification interface is changed.
@EgorDuplensky Please also take a look on this PR

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -450,6 +450,37 @@ void Graph::ResolveEdgeConflicts() {
         InsertReorder(edge, layerName, edge->getInputDesc(), edge->getOutputDesc(), isOptimized);
     };
 
+    auto reuseReorderForReadonlyChildren = [&](EdgePtr& edge) {
```
EgorDuplensky, line 453: Can we have it as an optimization in scope of ApplyImplSpecificGraphOptimizations path?
I believe in this case we don't impact Mergetransposeandreorder optimization.
So, we perform Mergetransposeandreorder first and then we perform reorders optimization.
Of visa versa, depending on the potential benefits.

usstq: > Can we have it as an optimization in scope of ApplyImplSpecificGraphOptimizations path?
I believe in this case we don't impact Mergetransposeandreorder optimization. So, we perform MergeTransposeAndReorder first and then we perform reorders optimization.

One purpose of this PR is to enable @jane-intel 's https://github.com/openvinotoolkit/openvino/pull/18775  which means there will be multiple children reorders from same transpose node (with her PR), and your expectation `don't impact Mergetransposeandreorder` will be disappointed. we have to change `MergeTransposeAndReorder` to allow such topology anyway.

>If the problem is that we do not want to touch Reorders which resolve in-place conflicts, can we have that as a path, i.e. ResolveInPlaceConflicts() and run it after the optimization you have introduced?

If we take a close look at `Graph::ResolveEdgeConflicts`, we can recognize that it's actually composed of two transformation passes just like any other pass inside `optimizer.ApplyImplSpecificGraphOptimizations` (although it's name didn't suggest that):
 - first pass : checking every pair of parent-child and insert reorder if their memory expectations are inconsistent.
 - second pass: resolve inplace conflict.

my change is just adding another optimization into first pass : reusing existing reorder.

@dmitry-gorokhov & you both expressed same concern about this change, because you didn't consider `ResolveEdgeConflicts` as a pass, just like any other passes inside `ApplyImplSpecificGraphOptimizations`, but to me `ResolveEdgeConflicts` is a pass, we can even split it into 2 function and move them into `ApplyImplSpecificGraphOptimizations` if we like.

EgorDuplensky: From my opinion, even this might be a more verbose change / implementation, it follows the general principle of the transformation pipeline - do one thing at a time.
This approach is simply more scalable and testable.
As a theoretical example, we might want to start testing this `ShareReorders` logic.
And to isolate the logic, we could create a graph with reorders explicitly inserted into the outputs of the same port.
And this would be impossible to test your optimization without executing `ResolveEdgeConflicts` call, when the idea of the test is not to test resolving of the edge conflicts, but `ShareReorders`
Regarding `ResolveEdgeConflicts`, I agree it is very similar to the graph optimizations paths. The only difference is it is actually mandatory (so we cannot really call it "optimization").
So, we can consider the whole scenario as:
1) run mandatory path first `ResolveEdgeConflicts` (after we split it of course)
2) optimize / clean up the graph after using optimization path `ShareReorders`

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -450,6 +450,37 @@ void Graph::ResolveEdgeConflicts() {
...
+        if (edge->inPlace(Edge::LOOK_DOWN))
+            return false;
+
+        const auto& dstDesc = edge->getOutputDesc();
+        for (auto siblingEdge : edge->getParent()->getChildEdgesAtPort(edge->getInputNum())) {
+            if (siblingEdge == edge)
+                continue;
+            auto reorder = std::dynamic_pointer_cast<node::Reorder>(siblingEdge->getChild());
```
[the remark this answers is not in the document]
usstq: OK, changed

src/plugins/intel_cpu/src/graph.cpp
usstq: OK, reasonable. I changed according to your suggestion.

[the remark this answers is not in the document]
usstq: OK, moved into `ApplyImplSpecificGraphOptimizations` as an optimization pass with necessary clean-up function calls.

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1579,21 +1611,21 @@ bool Graph::InsertNode(EdgePtr edge, NodePtr node, bool initNode) {
...
-        node->initSupportedPrimitiveDescriptors();
-        node->filterSupportedPrimitiveDescriptors();
-        node->selectOptimalPrimitiveDescriptor();
-        resolveInPlaceDirection(node);
-        node->initOptimalPrimitiveDescriptor();
-    }
+    AddNode(node);
 
+    if (initNode)
+        InitNewNodes({node});
```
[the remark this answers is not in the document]
usstq: According to your suggestion, I suppose the subgraph to be inserted has already been connected into the Graph by calling `Graph::CreateEdge` (recently introduced) before calling `Graph::InsertSubgraph`, so what `InsertSubgraph` is supposed to do is simply initialize the nodes contained in the subgraph which is exactly what `InitNewNodes` does.

If you mean `InsertSubgraph()` would also be responsible for connecting new nodes into subgraph and also connect the subgraph into main graph, I would say there is no simple way to provide the desired subgraph-topology information to `InsertSubgraph()`, I would rather call the flexible API `Graph::CreateEdge` to build the subgraph topology into the main graph, and then call `InitNewNodes()` to initialized newly created nodes in desired order.

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2091,6 +2094,73 @@ void GraphOptimizer::FuseEltwiseAndSimple(Graph &graph) {
...
+
+        Reorder* reorder = dynamic_cast<Reorder*>(node.get());
+        if (reorder == nullptr)
+            OPENVINO_THROW("Cannot get reorder layer ", node->getName());
+
+        if (!reorder->getShareable())
+            return nullptr;
+
+        // inplace children may write to data, cannot be safely shared
+        auto reorderConsumers = reorder->getChildEdgesAtPort(0);
```
EgorDuplensky, line 2110: So, we have two main cases where we insert reorders into the graph generically:
1) Layout reorders
2) Inpace conflicts

From my understanding this optimization covers only Layout reorders. Is it correct?
If so, can we have a following pipeline, instead of checking for inplace case every time:
1) Insert Layout reorders
2) Share reorders
3) Insert Inplace-conflict reorders.

usstq: Oh, sorry, my concern is invalid, you are correct!
Sharing reorders w/o considering inplace is good idea, since afterward ResolveInplaceConflicts() will handle inplace conflicts, thanks for your explanations! code is updated, please review again!

yuxu42:
@EgorDuplensky Could you please review again? Thanks!

EgorDuplensky (approved):

src/plugins/intel_cpu/src/graph.cpp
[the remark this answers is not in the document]
maxnick: The main problem here is the fact that the graph has two main states, which are different from the transformation perspective: Uninitialized and Initialized. The `Graph::CreateEdge`, `Graph::AddNode` and other similar ops, allow us to easily manipulate the graph topology, but completely ignores the initialized state. Therefore, if we had made it correctly, they wouldn't have been allowed to be called to the Initialized (or partially initialized) graph, to avoid possible errors (as the current implementations of these calls invalidate the graph initialized state in general case). But, the initialized and unutilized states are not really separated and the graph object may be in both or even in-between. What your propose here is just a further step in this wrong direction. This method `InitNewNodes(const std::vector<NodePtr>& new_nodes)` disclose some internal complexity (as it become a public method) of the graph without binding it with all the other internal stuff properly, e.g. one may call initialization to nodes that are not a part of the graph, or to the nodes that are partially in the graph, so nothing prevents it from misuse. Also let's look at its purpose. You introduced this method, which has a decent impact on the graph interface, only to serve in a single transformation, which seems to be too much, and may be even resolved inside this transformation itself. On the other hand we already have `InsertNode` methods, which are beyond graph topology modification methods ( `Graph::CreateEdge`, `Graph::AddNode`) and consider the graph initialization specifics, though cumbersome (via the `initNode`) flag. So, to do it properly, we have to either extend the existing `InsertNode` action to a subgraph (when we insert a subgraph in-between two nodes) or somehow modify the graph topology modification methods ( `Graph::CreateEdge`, `Graph::AddNode` etc.) so that they respect the initialization state properly.

maxnick (changes requested):

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -491,6 +496,39 @@ void Graph::ResolveEdgeConflicts() {
...
+        std::string layerName = basicLayerName;
+        int idx = 0;
+        while (uniqueLayerNames.find(layerName) != uniqueLayerNames.end()) {
+            idx++;
+            layerName = basicLayerName + "_" + std::to_string(idx);
+        }
+        uniqueLayerNames.insert(layerName);
+
+        // optimized flag indicate that just desc update w/o actual physical memory movement.
+        return InsertReorder(edge, layerName, edge->getInputDesc(), edge->getOutputDesc(), isOptimized);
```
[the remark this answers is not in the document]
usstq: removed

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -517,13 +555,11 @@ void Graph::ResolveEdgeConflicts() {
...
     };
 
-    numberOfEdges = graphEdges.size(); //update the total number
-
     for (ptrdiff_t i = 0; i < numberOfEdges; i++) {
         auto edge = graphEdges[i];
         if (needReorder(edge)) {
             constexpr bool optimizedReorder = false;
-            insertReorder(edge, optimizedReorder);
+            auto newReorder = insertReorder(edge, optimizedReorder);
```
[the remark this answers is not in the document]
usstq: removed

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -491,6 +496,39 @@ void Graph::ResolveEdgeConflicts() {
...
+        int idx = 0;
+        while (uniqueLayerNames.find(layerName) != uniqueLayerNames.end()) {
+            idx++;
+            layerName = basicLayerName + "_" + std::to_string(idx);
+        }
+        uniqueLayerNames.insert(layerName);
+
+        // optimized flag indicate that just desc update w/o actual physical memory movement.
+        return InsertReorder(edge, layerName, edge->getInputDesc(), edge->getOutputDesc(), isOptimized);
+    };
```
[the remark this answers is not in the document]
usstq: extracted!

jane-intel:
The PR dates back to Jul 31, 2023. Is there any ETA?

src/plugins/intel_cpu/src/graph.h
```diff
@@ -190,6 +190,8 @@ class Graph {
         return internalStateNodes;
     }
 
+    void resolveInPlaceDirection(const NodePtr& node) const;
```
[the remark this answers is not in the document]
usstq: agree, it's more like Node's member function, just like `getSupportedDescriptors`/`initSupportedPrimitiveDescriptors`/`...`, so I moved it into Node

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/fuse_transpose_reorder.cpp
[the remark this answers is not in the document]
usstq: covered!

src/plugins/intel_cpu/tests/unit/graph/merge_transpose_reorder_test.cpp
```diff
@@ -151,8 +148,24 @@ class MergeTransposeReorderIsOptimizedCPUTest : public ::testing::Test {
...
 
     std::unique_ptr<Graph> m_graph;
-}; // class MergeTransposeReorderIsOptimizedCPUTest
+}; // class MergeTransposeReordersCPUTest
+
+// upstream node or downstream node is inPlaced thereby the inserted Reorder cannot be optimized.
+TEST_F(MergeTransposeReordersCPUTest, smoke_Run_MergeTransposeReorders_isOptimized) {
+    CreateGraph(1, Edge::LOOK::LOOK_DOWN);
+    CheckTransposeCount(0);
+    CheckReorderOptimized(std::string("_fake"), false);  // the fused node is of name "reshape_abcd_acdb_fake"
```
maxnick, line 157: Hm, this is an interesting outcome. Should we name reorders which are not "optimized" with the "fake" postfix? I think we shouldn't.

src/plugins/intel_cpu/tests/unit/graph/merge_transpose_reorder_test.cpp
```diff
@@ -151,8 +148,24 @@ class MergeTransposeReorderIsOptimizedCPUTest : public ::testing::Test {
...
+TEST_F(MergeTransposeReordersCPUTest, smoke_Run_MergeTransposeReorders_shared) {
+    CreateGraph(3, 0);
+    CheckTransposeCount(0);
+    CheckReorderOptimized(std::string("_fake"), true);
+}
 
-TEST_F(MergeTransposeReorderIsOptimizedCPUTest, smoke_Run_MergeTransposeReorder_isOptimized) {
-    Validate();
+// 3 inplace consumers cannot share reorders thus transpose is not fused with reorders
+TEST_F(MergeTransposeReordersCPUTest, smoke_Run_MergeTransposeReorders_notFused) {
```
maxnick, line 168: Please add expected reorders count check.

src/plugins/intel_cpu/tests/unit/graph/merge_transpose_reorder_test.cpp
maxnick, line 1: Please add a description that there may be more than one Dummy->Output blocks.

src/plugins/intel_cpu/tests/unit/graph/merge_transpose_reorder_test.cpp
usstq: I don't want to change this convention in my PR since:
 - I'm not sure whether there is other test relying on this naming
 - if we remove _fake, we don't known how to find this reorder (since there are many reorders inserted by other pass/stages)

src/plugins/intel_cpu/tests/unit/graph/merge_transpose_reorder_test.cpp
usstq: added

src/plugins/intel_cpu/tests/unit/graph/merge_transpose_reorder_test.cpp
usstq: OK

maxnick (approved):

maxnick:
Merge after internal validation.

maxnick:
@dmitry-gorokhov , the PR is ready to be merged. The PR passed the internal accuracy/performance validation.

PR #18867, opened 2023-07-31
Return a dynamic layout for empty dependent memory of border

steve-y:
### Details:
 - *Return a dynamic layout for empty dependent memory of border in dynamic model*

### Tickets:
 - *110706*


wilson-seok:
@steve-y Please add either func test or unit test for the change.

steve-y:
> @steve-y Please add either func test or unit test for the change.

Yes, the unit test for the change is added. Thanks for your comment.

src/plugins/intel_gpu/src/graph/border.cpp
```diff
@@ -65,6 +65,10 @@ std::vector<layout> border_inst::calc_output_layouts(border_node const& /*node*/
     };
 
     auto& memory_deps = impl_param.memory_deps;
+    if (desc->pad_value == 0 && (is_begin_mem || is_end_mem) && memory_deps.empty()) {
+        return {layout{ShapeType::dynamic(static_cast<int64_t>(input_shapes[0].size())), input0_layout.data_type, input0_layout.format}};
+    }
+
```
vladimir-paramuzov, line 71: [random spot] Could you add shape infer test for border?

vladimir-paramuzov: Also please have a look at the smoke_GPUPadDynamic2D/PadLayerGPUTest.CompareWithRefs/IS=[?.?]_TS=(5.36)_(3.16)_Prc=f32_padsBegin=(0.1)_padsEnd=(1.2)_constantInput=CONSTANT/PARAMETER/CONSTANT_PadMode=edge test. With PR https://github.com/openvinotoolkit/openvino/pull/18925 this (and some other) test fails due to nullptr passed to make_tensor. So I assume that the shape infer impl has logical error, thus please ensure that such case is covered by shape infer unit tests

src/plugins/intel_gpu/src/graph/border.cpp
```diff
@@ -65,6 +65,10 @@ std::vector<layout> border_inst::calc_output_layouts(border_node const& /*node*/
     };
 
     auto& memory_deps = impl_param.memory_deps;
+    if (desc->pad_value == 0 && (is_begin_mem || is_end_mem) && memory_deps.empty()) {
```
yeonbok, line 68: I think we dont need to check value, 
but what we need to check is the memory dep count is equal to the required  # of memory deps.
i.e., 
if begin is from mem, end is constant => we need 2 mem deps 
if only begin is from mem, and end is constant => we need 1 mem dep 

So, please change it like 
if num_required_mem_dep < actual_mem_dep, return dynamic shape.

src/plugins/intel_gpu/src/graph/border.cpp
steve-y: > [random spot] Could you add shape infer test for border?

Yes, the shape infer test is added. **pad_test_params** in the test is used, and I'm not sure if it is proper.
Could you review the test please?

vladimir-paramuzov: I think the case that I mentioned above is still not covered and not fixed. I think it can be covered to some degree by changing border type for `pad_test_non_constant_input_end` test case to `EDGE` value. Another uncovered thing is the case when `value` input has `Constant` type in ngraph, but mode != Constant. In that case `data` primitive is created for that argument and it may behave a little bit differently comparing to `input_layout`

PR #18869, opened 2023-07-31
Review NMS shape inference

praasz:
### Details:
 - Review static and dynamic shape propagation
 - Review label propagation
 - Check interval values and label propagation on inputs.
 
### Tickets:
 - 97239

### DIFF
--- a/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
+++ b/src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
@@ -66,6 +66,7 @@
 #include "lstm_sequence_shape_inference.hpp"
 #include "matmul_shape_inference.hpp"
 #include "max_pool_shape_inference.hpp"
+#include "nms_shape_inference.hpp"
 #include "one_hot_shape_inference.hpp"
 #include "pad_shape_inference.hpp"
 #include "prior_box_clustered_shape_inference.hpp"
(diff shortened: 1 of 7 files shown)


src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
```diff
@@ -399,6 +401,7 @@ const IStaticShapeInferFactory::TRegistry IStaticShapeInferFactory::registry{
     _OV_OP_SHAPE_INFER_MASK_REG(opset9::Eye, ShapeInferTA, util::bit::mask(0, 1, 3)),
     _OV_OP_SHAPE_INFER_MASK_REG(opset9::GridSample, ShapeInferTA, util::bit::mask()),
     _OV_OP_SHAPE_INFER_MASK_REG(opset9::IRDFT, ShapeInferTA, util::bit::mask(1, 2)),
+    _OV_OP_SHAPE_INFER_MASK_REG(opset9::NonMaxSuppression, ShapeInferTA, util::bit::mask(2)),
```
jane-intel, line 404: NMS shouldn't be registered here -- Only plugin will decide output shape of the NMS operation -- it is dynamic and depends on input data.

jane-intel (changes requested):

src/plugins/intel_cpu/src/utils/shape_inference/shape_inference.cpp
praasz: OK, I'll remove it.

I've just follow the description in spec [NMS_9](https://github.com/praasz/openvino/blob/099339a960ce9619a40543b3d684f91740f866d7/docs/ops/sort/NonMaxSuppression_9.md?plain=1#L86).

The CPU plugin should calculate accurate shape value depends on data and providing upper bound is not required?

praasz: Removed v5 and v9 as they produce dynamic output.

PR #18874, opened 2023-07-31
[GNA] Fixed absent node_id crash

mryzhov:
### Details:
 - Markup function uses ngraph function get_ordered_ops which doesn't return layers inside the tensor operator. This fix takes care about that.

### Tickets:
 - *N/A*


src/plugins/intel_gna/src/transformations/fuse_conv_bias_activation.cpp
```diff
@@ -23,6 +24,7 @@
 #include "openvino/pass/pattern/op/or.hpp"
 #include "openvino/pass/pattern/op/pattern.hpp"
 #include "openvino/pass/pattern/op/wrap_type.hpp"
+#include "openvino/pass/serialize.hpp"
```
sirzabek, line 27: Is it required?

sirzabek (approved):
Please add ticket number to general PR description.

PR #18879, opened 2023-07-31
[CPU][ARM] MLAS Transpose executor

alvoron:
MLAS PR to enable ARM build: https://github.com/openvinotoolkit/mlas/pull/3


src/plugins/intel_cpu/CMakeLists.txt
```diff
@@ -33,8 +33,13 @@ else()
 endif()
 set(OV_CPU_ARM_TARGET_ARCH ${OV_CPU_ARM_TARGET_ARCH_DEFAULT})
 
-# enbale mlas for X86 cpus only
-ie_dependent_option(ENABLE_MLAS_FOR_CPU "MLAS GEMM for OpenVINO CPU Plugin" ON "X86 OR X86_64" OFF)
+if(X86 OR X86_64)
+    set(ENABLE_MLAS_FOR_CPU_DEFAULT ON)
+else()
+    set(ENABLE_MLAS_FOR_CPU_DEFAULT OFF)
```
dmitry-gorokhov, line 39: Since we already enable MLAS Transpose as first priority we can set ON by default for ARM as well. For other architectures (like RV64) it should be OFF.

src/plugins/intel_cpu/src/nodes/executors/mlas/mlas_transpose.cpp
```diff
@@ -0,0 +1,299 @@
...
+                    }
+                    output_for_first_writer += bytes_per_write;
+                }
+                output_data += writes_per_loop * bytes_per_write;
+            }
+        }
+    }
+}
+
+void MlasTransposeExecutor::TransposeSingleAxisInwards(const MemoryCPtr& input, const MemoryPtr& output, size_t from, size_t to) {
```
dmitry-gorokhov, line 215: maybe I missed smt, but I don't see any parallelization in the code. Seems like we will need to evaluate this as follow-up task.

src/plugins/intel_cpu/src/nodes/executors/mlas/mlas_transpose.cpp
```diff
@@ -0,0 +1,309 @@
...
+        DEBUG_LOG("MLAS Transpose executor supports moving single axis only");
+        return false;
+    }
+    return true;
+}
+
+bool MlasTransposeExecutorBuilder::isSupported(const TransposeParams& transposeParams,
+                                               const std::vector<MemoryDescPtr>& srcDescs,
+                                               const std::vector<MemoryDescPtr>& dstDescs) const {
+    if (!srcDescs[0]->hasLayoutType(LayoutType::ncsp)) {
```
dmitry-gorokhov, line 297: 1. dst layout should be the same is input one.
2. supported precision sizes (getPrecision.size()) are 1,2,4,8. We have to check this.

src/plugins/intel_cpu/src/nodes/executors/mlas/mlas_transpose.cpp
```diff
@@ -0,0 +1,309 @@
...
+            }
+        }
+    }
+}
+
+void MlasTransposeExecutor::TransposeSingleAxisInwards(const MemoryCPtr& input, const MemoryPtr& output, size_t from, size_t to) {
+    const auto& input_shape = input->getShape();
+    const auto& input_dims = input_shape.getDims();
+
+    const auto element_size = dnnl::memory::data_type_size(input->getDataType());
```
dmitry-gorokhov, line 219: still there

PR #18884, opened 2023-07-31
[PyOV] Add Range-4 as part of opset12

jiwaszki:
### Details:
 - Add `opset12.range` as it was missing since `opset4`.

### Tickets:
 - *116684*


src/bindings/python/tests/test_graph/test_sequence_processing.py
```diff
@@ -5,27 +5,47 @@
...
+def test_range_1():
     start = 5
     stop = 35
     step = 5
 
-    node = ov.range(start, stop, step)
+    node = ov.opset11.range(start, stop, step)
     assert node.get_output_size() == 1
     assert node.get_type_name() == "Range"
     assert list(node.get_output_shape(0)) == [6]
```
mitruska, line 32: Worth to add expected output type also for v0::range test.

src/bindings/python/src/openvino/runtime/opset12/__init__.py
```diff
@@ -120,7 +120,7 @@
 from openvino.runtime.opset1.ops import prior_box_clustered
 from openvino.runtime.opset1.ops import psroi_pooling
 from openvino.runtime.opset4.ops import proposal
-from openvino.runtime.opset1.ops import range
+from openvino.runtime.opset12.ops import range
```
mitruska, line 123: If `def range` will be moved to `opset4/ops.py`:
```suggestion
from openvino.runtime.opset4.ops import range
```

mitruska (approved):
Approved current version, but left few suggestions to consider (can be discussed and applied separately).

src/bindings/python/tests/test_graph/test_sequence_processing.py
jiwaszki: Tbh I do not want to change this test now.

src/bindings/python/src/openvino/runtime/opset12/ops.py
```diff
@@ -118,3 +119,29 @@ def group_normalization(
             "epsilon": epsilon,
         },
     )
+
+
+@nameable_op
+def range(
```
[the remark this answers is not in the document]
jiwaszki: Done

src/bindings/python/src/openvino/runtime/opset12/__init__.py
jiwaszki: Done

PR #18891, opened 2023-08-01
[GPU] updated to enqueue only fully_connected layers for async build

e-ddykim:
### Details:
 - This PR updates the async build for dynamic models to build convolution, fully_connected, and softmax layers.
 - Current kernel build process has negative impact on inference tasks. So we need to select effective kernels and to build them.

### Tickets:
 - 103753

### DIFF
--- a/src/plugins/intel_gpu/src/graph/compilation_context.cpp
+++ b/src/plugins/intel_gpu/src/graph/compilation_context.cpp
@@ -13,7 +13,7 @@ namespace cldnn {
 class CompilationContext : public ICompilationContext {
 public:
     CompilationContext(ov::threading::IStreamsExecutor::Config task_executor_config) : _task_executor_config(task_executor_config) {
-        _task_executor_config._streams = 4;
+        _task_executor_config._streams = 1;
         _task_executor = std::make_shared<ov::threading::CPUStreamsExecutor>(_task_executor_config);
     }
 
--- a/src/plugins/intel_gpu/src/graph/compilation_context.cpp
+++ b/src/plugins/intel_gpu/src/graph/compilation_context.cpp
@@ -13,7 +13,6 @@ namespace cldnn {
 class CompilationContext : public ICompilationContext {
 public:
     CompilationContext(ov::threading::IStreamsExecutor::Config task_executor_config) : _task_executor_config(task_executor_config) {
-        _task_executor_config._streams = 1;
         _task_executor = std::make_shared<ov::threading::CPUStreamsExecutor>(_task_executor_config);
     }
 
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -590,12 +590,12 @@ bool primitive_inst::update_impl() {
         }
         if (!cached_impl) {
             if (_dynamic_impl) {
-                auto use_async_compilation = [&]() {
+                auto use_async_compilation = [&updated_params_no_dyn_pad]() {
                     GPU_DEBUG_GET_INSTANCE(debug_config);
                     GPU_DEBUG_IF(debug_config->disable_async_compilation) {
                         return false;
                     }
-                    return true;
+                    return updated_params_no_dyn_pad.desc->type_string().compare("fully_connected") == 0;
                 };
                 if (use_async_compilation()) {
                     auto& compilation_context = get_network().get_program()->get_compilation_context();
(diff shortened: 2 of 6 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -569,12 +569,12 @@ bool primitive_inst::update_impl() {
...
         if (!cached_impl) {
             if (_dynamic_impl) {
-                auto use_async_compilation = [&]() {
+                auto use_async_compilation = [&updated_params_no_dyn_pad]() {
                     GPU_DEBUG_GET_INSTANCE(debug_config);
                     GPU_DEBUG_IF(debug_config->disable_async_compilation) {
                         return false;
                     }
-                    return true;
+                    return updated_params_no_dyn_pad.desc->type_string().compare("fully_connected") == 0;
```
vladimir-paramuzov, line 577: Why comparing strings?
```
updated_params_no_dyn_pad.is_type<fully_connected>()
```
it can also be a member of primitive_inst
```cpp
bool primitive_inst::use_async_compilation() const {
    return _node->is_type<fully_connected>();
}
```

src/plugins/intel_gpu/src/graph/compilation_context.cpp
```diff
@@ -13,7 +13,7 @@ namespace cldnn {
 class CompilationContext : public ICompilationContext {
 public:
     CompilationContext(ov::threading::IStreamsExecutor::Config task_executor_config) : _task_executor_config(task_executor_config) {
-        _task_executor_config._streams = 4;
+        _task_executor_config._streams = 1;
```
yeonbok, line 16: Could you please make a change so that this value is set from the caller? 
i.e., set _streams in task_executor_config from the caller side.

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -590,10 +590,12 @@ bool primitive_inst::update_impl() {
                                 return;
                         }
 
-                        auto impl = _node->type()->choose_impl(*_node, updated_params_no_dyn_pad);
                         if (!can_be_optimized()) {
-                            auto kernels = _program->get_kernels_cache().compile(updated_params_no_dyn_pad, impl->get_kernels_source());
-                            impl->set_kernels(kernels);
+                            auto impl = _node->type()->choose_impl(*_node, updated_params_no_dyn_pad);
+                            if (impl->get_kernels_source().size() > 0) {
```
yeonbok, line 595: when impl->get_kernels_source().size() == 0  though it is not can_be_optimized?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
e-ddykim: Oh, I'll update it as you guided. Thank you!

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
e-ddykim: This is the case of onednn_impl.

src/plugins/intel_gpu/src/graph/compilation_context.cpp
e-ddykim: Okay, I'll update it. Thank you.

yeonbok (approved):

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -478,6 +479,16 @@ event::ptr primitive_inst::realloc_if_needed() {
     return ev;
 }
 
+bool primitive_inst::use_async_compilation() {
+    GPU_DEBUG_GET_INSTANCE(debug_config);
+    GPU_DEBUG_IF(debug_config->disable_async_compilation) {
+        return false;
+    }
+    return (_node->is_type<convolution>() ||
+            _node->is_type<fully_connected>() ||
```
vladimir-paramuzov, line 488: btw, shouldn't gemm be a part of this list?

src/plugins/intel_gpu/tests/unit/test_cases/gemm_gpu_test.cpp
```diff
@@ -1499,7 +1499,7 @@ INSTANTIATE_TEST_SUITE_P(gemm_gpu, gemm_onednn_ndims, ::testing::ValuesIn(std::v
     gemm_onednn_test_params{ CASE_GEMM_ONEDNN_I8_6D },
 }));
 
-TEST(gemm_onednn, impl_replacement_with_cldnn) {
+TEST(gemm_onednn, DISABLED_impl_replacement_with_cldnn) {
```
vladimir-paramuzov, line 1502: Why disabled? I think such test still make sense, gemm may be replaced with FC primitive if needed.

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
e-ddykim: My test results didn't show no benefit when `gemm` is added to the async compilation queue. I'll test it again with a long running scenario such as the `softmax` case.

e-ddykim: I updated the async compilation queue to include `gemm`.

src/plugins/intel_gpu/tests/unit/test_cases/gemm_gpu_test.cpp
e-ddykim: I reverted this change. Thank you.

vladimir-paramuzov (approved):

PR #18893, opened 2023-08-01
[CPU] Fix jit compilation exception on sse for mvn

chenhu-wang:
### Details:
 - *Valid tails number is [0, 15] on avx512, [0, 7] on avx2 and [0, 3] on sse41. Generic 8/4/2/1 tiles processing for tails should have conditional start. On AVX2, should be 4/2/1. On SSE, should be 2/1. Otherwise, for example, load 8 tails on sse is not allowed*

### Tickets:
 - *CVS-116378*


chenhu-wang:
Current single layer test case could cover it, but not run on SSE4. Local SDE show this could fix the model. and all single layer test become passed on sse4.

dmitry-gorokhov (approved):

PR #18925, opened 2023-08-02
[GPU] Use ov::Tensor intead of ngraph::HostTensor

vladimir-paramuzov:
### Details:
 - use ov::Tensor for shape inference instead of deprecated HostTensor
 - replace map with unordered_map to match available tensor accessor specializations

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/runtime/tensor_accessor.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/runtime/tensor_accessor.hpp
@@ -1,0 +1,82 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include "openvino/runtime/tensor.hpp"
+#include "tensor_data_accessor.hpp"
+
+#include "memory.hpp"
+#include "layout.hpp"
+
+namespace cldnn {
+
+inline ov::Tensor make_tensor(const layout& l, void* memory_pointer) {
+    ov::element::Type et = data_type_to_element_type(l.data_type);
+
+    return ov::Tensor(et, l.get_shape(), memory_pointer);
+}
+
+struct TensorsContainer final {
+    using MemoryMap = std::unordered_map<size_t, cldnn::memory::ptr>;
+    using TensorsMap = std::unordered_map<size_t, ov::Tensor>;
+
+    explicit TensorsContainer(const cldnn::stream* stream) : m_stream(stream) { }
+
+    ~TensorsContainer() {
+        for (auto& mem : m_memories) {
+            mem.second->unlock(*m_stream);
+        }
+    }
+
+    void emplace(size_t port, cldnn::memory::ptr mem) {
+        m_memories.emplace(port, mem);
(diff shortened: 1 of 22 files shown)


yeonbok (approved):

src/plugins/intel_gpu/include/intel_gpu/runtime/tensor_accessor.hpp
```diff
@@ -0,0 +1,82 @@
...
+        return m_tensors.at(port);
+    }
+
+private:
+    const cldnn::stream* m_stream;
+    MemoryMap m_memories;
+    TensorsMap m_tensors;
+};
+
+class TensorAccessor final : public ov::ITensorAccessor {
```
praasz, line 64: Can you check this implementation of tensor accessor for GPU [MemoryAccessor](https://github.com/praasz/openvino/blob/b83389a5576c3641366de739789acff696881ae4/src/plugins/intel_gpu/src/graph/include/memory_accessor.hpp#L15)
I've done it more like concept but maybe instead make new one, this idea can be merged into this proposal  or vice versa?

Here is example of usage [CLDNN and core for tensor map](https://github.com/praasz/openvino/blob/b83389a5576c3641366de739789acff696881ae4/src/plugins/intel_gpu/src/graph/resample.cpp#L69)
It uses `cldnn` memories and also `core` tensor accessor on operator properties converted to tensor map.

This `MemoryAccessor` check data in `cldnn` memories and if not found use callback e.g. `core` tensor accessor or any custom callback.
[custom lambda example](https://github.com/praasz/openvino/blob/b83389a5576c3641366de739789acff696881ae4/src/plugins/intel_gpu/src/graph/tile.cpp#L56)

vladimir-paramuzov: @praasz Looks like the idea is pretty much similar to what I've done - simplify lock()/unclok() for GPU memory and selection between different dependency sources. 

The key difference is that I've added custom container type which manages cldnn::memory and ov::Tensors internally, and tensor accessor is pretty simple, while in your implementation tensor accessor itself manages cldnn memory + callback provides fallback to ov::Tensors if needed. Don't see any huge advantage of one impl vs another.

Also, your implementation has a good feature of on demand memory lock, though likely the logic is a little bit incorrect as access to multiple tensors sequentially invalidates all pointers except the last one due to `unlock_current_data()` call. So I've added similar feature for my impl, but it just saves the port ids of locked memorys for correct unlock on container d-tor.

vladimir-paramuzov:
@yeonbok, @praasz Do you have any further comments for this patch? If no, I'll merge it
@steve-y Could you check if this PR also fixes the issue from https://github.com/openvinotoolkit/openvino/pull/18867 ?

src/plugins/intel_gpu/include/intel_gpu/runtime/tensor_accessor.hpp
praasz: Ok, my main point was to have one solution instead of two for GPU, but in generał this PR is fine for me.

steve-y:
> @steve-y Could you check if this PR also fixes the issue from #18867 ?

Yes, border works as expected. Thanks.

PR #18943, opened 2023-08-02
[ShapeInference] GenerateProposals - shape infer review and tests

mitruska:
### Details:
 - No new features, just additional tests (labels and dynamic dimensions ranges) and update validation checks to use `NODE_SHAPE_INFER_CHECK` instead of `NODE_VALIDATION_CHECK`
 
\* GenerateProposals shape inference produces always dynamic num_rois dimension, so it's not registered in CPU shape infer map for usage with StaticShape
### Tickets:
 - 97233


src/core/shape_inference/include/generate_proposals_shape_inference.hpp
```diff
@@ -22,106 +22,120 @@ std::vector<TRShape> shape_infer(const GenerateProposals* op, const std::vector<
...
-
-        NODE_VALIDATION_CHECK(op,
-                              deltas_shape[3].compatible(scores_shape[3]),
-                              "Width for inputs 'input_deltas' and 'input_scores' should be "
-                              "equal. Got: ",
-                              deltas_shape[3],
-                              scores_shape[3]);
+        NODE_SHAPE_INFER_CHECK(op,
+                               input_shapes,
+                               deltas_shape[0].compatible(scores_shape[0]),
```
praasz, line 74: I think, the merges above could be used here instead compatible check.

src/core/shape_inference/include/generate_proposals_shape_inference.hpp
```diff
@@ -22,106 +22,120 @@ std::vector<TRShape> shape_infer(const GenerateProposals* op, const std::vector<
...
+        NODE_SHAPE_INFER_CHECK(op,
+                               input_shapes,
+                               anchors_shape[2].compatible(scores_shape[1]),
+                               "Anchor number for inputs 'input_anchors' and 'input_scores' should be "
+                               "equal. Got: ",
+                               anchors_shape[2],
+                               scores_shape[1]);
     }
 
     auto num_rois = Dimension(0, (num_batches * op->get_attrs().post_nms_count).get_max_length());
```
praasz, line 141: for template usage this conversion should be applied for PartialShape only.

src/core/shape_inference/include/generate_proposals_shape_inference.hpp
```diff
@@ -22,106 +22,120 @@ std::vector<TRShape> shape_infer(const GenerateProposals* op, const std::vector<
     const auto& scores_shape = input_shapes[3];
     const auto im_info_shape_rank = im_info_shape.rank();
     auto num_batches = Dimension::dynamic();
```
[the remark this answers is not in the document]
mitruska: Currently it is possible to build GenerateProposals shape_infer only with PartialShape class, because `num_rois` dimension is always dynamic by design. For other shape types compilation error is better than runtime exception, so incorrect usage can be detected earlier.
It can be refactored to make it non template function. To be discussed.

src/core/shape_inference/include/generate_proposals_shape_inference.hpp
mitruska: The whole shape infer for GenerateProposals is compatible with PartialShape class only, as it always produces dynamic shape.

src/core/shape_inference/include/generate_proposals_shape_inference.hpp
mitruska: Yes, seems like Dimension::merge and check can be done at once, but I would like to avoid unnecessary changes in this PR.
Let's merge current version and apply other suggestions separately.

praasz (approved):

PR #18948, opened 2023-08-02
Update ResolveNameCollisions to handle all friendly names, not only auto-generated

itikhono:
### Details:
 - Added an additional mode "resolve_all_names" to handle all friendly names, not only auto-generated 
 - Updated tests

### Tickets:
 - *115909*


src/common/transformations/src/transformations/resolve_names_collisions.cpp
```diff
@@ -43,19 +43,38 @@ bool ov::pass::ResolveNameCollisions::run_on_model(const std::shared_ptr<ov::Mod
...
+                // we don't need to change name of the first op in this list.
+                // the expected result: Name, Name_1, Name_2 etc
+                if (cnt == 0) {
+                    cnt++;
+                    continue;
+                }
+                update_name(op, cnt);
+            }
+        } else {
+            int64_t cnt = 2;
```
itikhono, line 68: maybe cnt = 1 is better

src/common/transformations/include/transformations/resolve_names_collisions.hpp
```diff
@@ -18,11 +18,19 @@ namespace pass {
  * Every transformation call can change the graph structure and create some additional operations,
  * autogenerated name is used if new operation doesn't have friendly name.
  * This transformations should be called after the transformation pipeline in order to fix names collisions.
+ *
+ *  There is also an additional mode "resolve_all_names", the logic of which is the same,
```
rkazants, line 22:
```suggestion
 * There is also an additional mode "resolve_all_names", the logic of which is the same,
```

rkazants: minor

vladimir-paramuzov (approved):

src/common/transformations/src/transformations/resolve_names_collisions.cpp
```diff
@@ -43,19 +43,38 @@ bool ov::pass::ResolveNameCollisions::run_on_model(const std::shared_ptr<ov::Mod
...
-            bool is_autogenerated = op->m_friendly_name.empty();
-            if (!is_autogenerated) {
-                continue;
-            }
+        auto update_name = [&](ov::Node* op, int64_t& cnt) {
             // add a prefix "_counter" to the autogenerated name to make it unique.
-            auto new_name = op->get_friendly_name() + "_" + std::to_string(cnt++);
+            const auto& old_name = op->get_friendly_name();
+            auto new_name = old_name + "_" + std::to_string(cnt++);
             while (visited_nodes.find(new_name) != visited_nodes.end()) {
```
rkazants, line 50: should not we update a map of `visited_nodes` with a new name?

src/common/transformations/src/transformations/resolve_names_collisions.cpp
rkazants: Looks strange that you are using different initial value for counter. Also, I think you can use the same logic for both cases `m_resolve_all_names` that wiil determine if check `op->m_friendly_name.empty()`

src/common/transformations/src/transformations/resolve_names_collisions.cpp
itikhono: Good point! I believe we don't need and that's why:
Let's say  "visited_nodes" map initially contains
[prefix1] = node1, node2, node3
[prefix2] = node4, node5
, where prefix1 != prefix2

After applying the transformation for "prefix1" and adding the new elements to "visited_nodes" map
[prefix1] = node1
[prefix2] = node4, node5
[prefix3 = prefix1 + id] = node2
[prefix4 = prefix1 + id] = node3
where prefix3 != prefix4 and prefix3 != prefix2, prefix4 != prefix2

The question is: is it possible to get prefix1 + id from prefix2 by adding id?
I don't think so, because prefix1 != prefix2

src/common/transformations/src/transformations/resolve_names_collisions.cpp
itikhono: I will resolve it in the next PR

rkazants (approved):

mateusztabaka (approved):

PR #18949, opened 2023-08-02
[GPU] Disabling redundant copying of constant weights

Lyamin-Roman:
### Details:
 - *Skipped precision reinterpret_cast for big constants on ConvertPrecision pass and replaced to Converts*
 - *Added reorder_transfer pass*

### Tickets:
 - *105424*


src/plugins/intel_gpu/src/graph/program.cpp
```diff
@@ -613,7 +608,17 @@ void program::post_optimize_graph(bool is_internal) {
...
     set_layout_optimizer_attributes(lo);
-    apply_opt_pass<post_optimize_weights>(rf);
+
+    bool optimize_data = _config.get_property(ov::intel_gpu::optimize_data);
+
+    if (!is_internal) {
+        apply_opt_pass<reorder_transfer>();
+
+        if (optimize_data) {
+            apply_opt_pass<fuse_constant_transposes>();
```
vladimir-paramuzov, line 618: Does reorder_transfer and fuse_constant_transposes need to happen on post-optimize stage? Can't reorder be transfered before fusions, and fuse_constant_transposes be a part of prepare_primitive_fusing pass?

src/common/transformations/src/transformations/fp16_compression/mark_decompression_convert_constant_folding.cpp
```diff
@@ -74,3 +76,35 @@ pass::KeepConstAndDecompression::KeepConstAndDecompression() {
     auto m = std::make_shared<pattern::Matcher>(node_pattern, matcher_name);
     register_matcher(m, callback);
 }
+
+pass::KeepConstantsPrecisionAndAddConverts::KeepConstantsPrecisionAndAddConverts() {
```
vladimir-paramuzov, line 80: Please add few tests for this transformation

src/plugins/intel_gpu/src/graph/graph_optimizer/fuse_constant_transposes.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "program_helpers.h"
+
+#include "permute_inst.h"
+#include "fully_connected_inst.h"
+
+using namespace cldnn;
+
+format fuse_constant_transposes::convert_weights_format_by_order(format fmt, const std::vector<uint16_t>& order) const {
+    format weights_fmt = from_weights_layout(to_weights_layout(fmt, false));
+    std::string old_order = weights_fmt.order();
```
vladimir-paramuzov, line 15: Please use `_order` from format traits instead of this string order

src/plugins/intel_gpu/src/runtime/format.cpp
```diff
@@ -344,4 +344,15 @@ const std::vector<std::pair<size_t, int>> format::per_axis_block_size(format fmt
 
     return sizes_for_dims;
 }
+
+format format::find_format(std::string order, const std::vector<std::pair<size_t, int>>& block_sizes) {
```
vladimir-paramuzov, line 348: Please add tests for this method

src/plugins/intel_gpu/src/graph/graph_optimizer/fuse_constant_transposes.cpp
```diff
@@ -0,0 +1,75 @@
...
+    auto itr = p.get_processing_order().begin();
+    while (itr != p.get_processing_order().end()) {
+        auto& node = *itr++;
+
+        if (!node->is_type<permute>())
+            continue;
+
+        auto& permute_node = node->as<permute>();
+
+        if (!is_matmul_weights_path(&permute_node) || !permute_node.get_dependency(0).is_type<data>())
```
vladimir-paramuzov, line 53: Why is it limited to weights path only? I suppose it may be useful for other cases as well. For example, after https://github.com/openvinotoolkit/openvino/pull/19516 same transform can be applied to deconvolution weights

src/plugins/intel_gpu/src/graph/graph_optimizer/fuse_constant_transposes.cpp
```diff
@@ -0,0 +1,75 @@
...
+            continue;
+
+        auto& prev_const = permute_node.get_dependency(0).as<data>();
+        const auto& permute_order = permute_node.get_primitive()->permute_order;
+
+        format new_fmt = convert_weights_format_by_order(prev_const.get_output_layout().format, permute_order);
```
vladimir-paramuzov, line 59: Would be good to catch exception here and skip such nodes

src/plugins/intel_gpu/src/graph/graph_optimizer/reorder_transfer.cpp
```diff
@@ -0,0 +1,65 @@
...
+        if (!is_simple_reorder)
+            continue;
+
+        layout input_layout = reorder_node.get_input_layout();
+        layout output_layout = reorder_node.get_output_layout();
+        bool only_precision_changed = input_layout.get_partial_shape() == output_layout.get_partial_shape() &&
+                                      input_layout.format == output_layout.format &&
+                                      input_layout.data_type != output_layout.data_type;
+        if (!only_precision_changed)
+            continue;
```
yeonbok, line 37: How about add a handle in reorder_node to return whether it is a simple_reorder_with_only_precision_change ? 
I think that condition can be utilized in other locations such as remove redundant reorder too.

src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing.cpp
```diff
@@ -1181,6 +1182,82 @@ void prepare_primitive_fusing::fuse_simple_primitives(program &p) {
...
+
+        const auto& new_mem = p.get_engine().reinterpret_buffer(prev_const.get_attached_memory(), updated_const_layout);
+
+        auto new_const_prim = std::make_shared<data>(prev_const.id() + "_fused_with_transpose", new_mem);
+        auto& new_const_node = p.get_or_create(new_const_prim);
+
+        p.replace(prev_const, new_const_node);
+        new_const_node.recalc_output_layout(false);
+    }
+}
```
yeonbok, line 1259: I'd like to be aligned with the basic idea,
Is my understanding correct? 
- The basic idea is to faking constant which is not actually transposed, but setting the expected transposed layout. 
- And the actual transpose will be done at the post weight optimization 

However I have curiosity can this be applied to random transpose order? 
Can this be applied for orders like (2, 1, 3, 0) ?
Can't it be applied to a transpose with only two dimensions exchanging? (e.g., 0, 1, 3, 2)

src/common/transformations/src/transformations/fp16_compression/mark_decompression_convert_constant_folding.cpp
```diff
@@ -74,3 +76,35 @@ pass::KeepConstAndDecompression::KeepConstAndDecompression() {
...
+
+    matcher_pass_callback callback = [=](pattern::Matcher& m) {
+        auto const_node = m.get_match_root();
+
+        if (transformation_callback(const_node)) {
+            return false;
+        }
+
+        auto constant_target_inputs = const_node->get_output_target_inputs(0);
+        auto convert = std::make_shared<ov::op::v0::Convert>(const_node, const_node->get_element_type());
```
vladimir-paramuzov, line 92: Seems like you always insert this dummy `Convert` op for matmuls, so I wonder what happens in case of FP16 model being loaded with FP16 inference precision (i.e. when conversion is not needed)? Is this convert eliminated on early transformation stages (in common optimizations for example)?

src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing.cpp
vladimir-paramuzov: It won't work for some combinations. But if the format is not found for given permutation, then such transpose will not be merged and will be folder later.

src/plugins/intel_gpu/src/runtime/format.cpp
```diff
@@ -344,4 +344,30 @@ const std::vector<std::pair<size_t, int>> format::per_axis_block_size(format fmt
...
+               format::is_grouped(trait.first) == is_grouped &&
+               format::is_image_2d(trait.first) == is_image_2d &&
+               format::is_winograd(trait.first) == is_winograd &&
+               format::is_nv12(trait.first) == is_nv12;
+    };
+    const auto& finded_format = std::find_if(format_traits_map.begin(), format_traits_map.end(), filter);
+    OPENVINO_ASSERT(finded_format != format_traits_map.end(), "[GPU] Cannot find a format with the specified parameters");
+
+    const auto& finded_format_from_end = std::find_if(format_traits_map.rbegin(), format_traits_map.rend(), filter);
+    OPENVINO_ASSERT(finded_format->first == finded_format_from_end->first, "[GPU] Cannot find a format. Specified parameters are ambiguous");
```
sshlyapn, line 368: Wouldn't it be better to iterate over format_traits_map manually only once and count the number of occurrences?

src/common/transformations/src/transformations/fp16_compression/mark_decompression_convert_constant_folding.cpp
Lyamin-Roman: Yes, looks like this Convert must be removed in CommonOptimizations->MOCTransformations->SharedOpOptimization pass

vladimir-paramuzov (approved):

PR #18960, opened 2023-08-03
[PT FE]: support aten::view_as and aten::std

eaidova:
### Details:
 - *support aten::std. aten::std_mean and aten::view_as ops*
tested on wavelm and vaw2vec-conformer models

### Tickets:
 - *CVS-117016*
 - *CVS-117015*


mvafin (approved):
Please fix test

PR #18965, opened 2023-08-03
[PT FE] Improved MaxPool convert by PyTorch FE

praasz:
### Details:
 - Improved conversion of `MaxPool` operator by `PyTorch` frontend, when sliding windows go off bound.
   Example:
   input shape: [1,3,5,5], kernel=2, stride=2, pad=1,dilation=1, ceil_mode
   - `PyTorch` result shape [1,3,3,3] 
   - `OV` result shape after conversion [1,3,4,4]
   This change makes conversion to be aligned with `PyTorch` result.

### Tickets:
 - #18731

### DIFF
--- a/src/frontends/pytorch/src/op/max_poolnd.cpp
+++ b/src/frontends/pytorch/src/op/max_poolnd.cpp
@@ -84,8 +84,7 @@ OutputVector translate_max_poolnd(const NodeContext& context) {
         const auto shape_of_mp = context.mark_node(std::make_shared<v3::ShapeOf>(mp, element::i32));
         const auto gth_out_dims = context.mark_node(std::make_shared<v8::Gather>(shape_of_mp, dim_idxs, zero));
         const auto out_sub_one = context.mark_node(std::make_shared<v1::Subtract>(gth_out_dims, one));
-        const auto stride_node =
-            use_kernel ? context.get_input(1).get_node_shared_ptr() : context.get_input(2).get_node_shared_ptr();
+        const auto stride_node = use_kernel ? context.get_input(1) : context.get_input(2);
         const auto out_mul_stride = context.mark_node(std::make_shared<v1::Multiply>(out_sub_one, stride_node));
 
         // if (in_dim + pad) > ((out_dim - 1) * stride) sliding window in bound use end padding.
--- a/src/frontends/pytorch/src/op/max_poolnd.cpp
+++ b/src/frontends/pytorch/src/op/max_poolnd.cpp
@@ -3,7 +3,18 @@
 //
 
 #include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/op/add.hpp"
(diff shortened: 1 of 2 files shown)


mmikolajcz (approved):

src/frontends/pytorch/src/op/max_poolnd.cpp
```diff
@@ -42,8 +54,55 @@ OutputVector translate_max_poolnd(const NodeContext& context) {
...
+        // PyTorch allows sliding window go off bound, which leads to this accommodation.
+        // More detail on https://github.com/pytorch/pytorch/issues/57178
+        const auto zero = context.mark_node(v0::Constant::create(element::i32, Shape{}, {0}));
+        const auto one = context.mark_node(v0::Constant::create(element::i32, Shape{}, {1}));
+        const auto two = context.mark_node(v0::Constant::create(element::i32, Shape{}, {2}));
+
+        const auto padding =
+            context.input_is_none(3)
+                ? context.mark_node(std::make_shared<v0::Constant>(element::i32, Shape{pads.size()}, 0))
+                : context.get_input(3).get_node_shared_ptr();
```
mvafin, line 69:
```suggestion
        const auto padding =
            context.input_is_none(3)
                ? context.mark_node(std::make_shared<v0::Constant>(element::i32, Shape{pads.size()}, 0))->output(0)
                : context.get_input(3);
```
It is not good to retrieve node from input. That node can have multiple outputs, but you will use only first output next, that may create incorrect graph.

src/frontends/pytorch/src/op/max_poolnd.cpp
```diff
@@ -42,8 +54,55 @@ OutputVector translate_max_poolnd(const NodeContext& context) {
...
+        const auto in_left_padded = context.mark_node(std::make_shared<v1::Add>(gth_in_dims, padding));
+
+        // gather output spatial dims and prepare it for compare as values (out_dim - 1) * stride
+        const auto mp = context.mark_node(
+            std::make_shared<v8::MaxPool>(input, strides, dilations, pads, pads, kernel, rounding_type));
+        const auto shape_of_mp = context.mark_node(std::make_shared<v3::ShapeOf>(mp, element::i32));
+        const auto gth_out_dims = context.mark_node(std::make_shared<v8::Gather>(shape_of_mp, dim_idxs, zero));
+        const auto out_sub_one = context.mark_node(std::make_shared<v1::Subtract>(gth_out_dims, one));
+        const auto stride_node =
+            use_kernel ? context.get_input(1).get_node_shared_ptr() : context.get_input(2).get_node_shared_ptr();
```
mvafin, line 88:
```suggestion
        const auto stride_node = use_kernel ? context.get_input(1) : context.get_input(2);
```

mvafin (changes requested):

mvafin (approved):

PR #18973, opened 2023-08-03
[CPU][ARM] Perf fixes for FP16 precision

EgorDuplensky:
### Details:
 - Enable float16_t for eltwise reference executor to avoid extra reorders
 - Allow forcing FP16 precision for input and output nodes via input/output info
 - Enable fp16 precision for func tests on arm

### Tickets:
 - 117049

### TBD:
- [x] Perf regression check
- [x] Tests update


EgorDuplensky:
@dmitry-gorokhov Ready for review. Could you please take a look?

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1743,44 +1745,58 @@ void Graph::EnforceInferencePrecision() {
...
+                    return true;
 
-            for (size_t i = 0; i < node->getOriginalOutputsNumber(); i++) {
-                if (node->getOriginalOutputPrecisionAtPort(i) == Precision::FP32)
-                    node->setOriginalOutputPrecisionAtPort(i, inferPrec);
-            }
+                return false;
+            };
+
+            if (keepOriginalInputPrecision(node, i))
```
EgorDuplensky, line 1786: perhaps better to pass a parent instead of idx

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1743,44 +1745,58 @@ void Graph::EnforceInferencePrecision() {
...
+                continue;
+
+            if (node->getOriginalInputPrecisionAtPort(i) != Precision::FP32)
```
dmitry-gorokhov, line 1789: Maybe better to include into `keepOriginalInputPrecision`

src/plugins/intel_cpu/src/graph.cpp
dmitry-gorokhov: Idx is ok for me. Only suggestion is to rename `keepOriginalInputPrecision` parameter as `inPort` for better understanding.

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -1685,21 +1694,14 @@ bool Graph::InsertNode(NodePtr parent, NodePtr child, NodePtr node, int parentPo
...
-        inferPrec = InferenceEngine::Precision::FP16;
-        break;
-    default:
-        return;
-        break;
-    }
+
+    const auto inferPrec = convertPrecision(getConfig().inferencePrecision);
+
+    if (inferPrec == Precision::FP32)
```
dmitry-gorokhov, line 1703: Maybe better to check allowed precisions: {BF16, FP16}?

src/plugins/intel_cpu/src/utils/precision_support.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "precision_support.h"
+
+#include "ie_precision.hpp"
+#include "cpu/x64/cpu_isa_traits.hpp"
+#include "openvino/core/visibility.hpp"
+
+namespace ov {
+namespace intel_cpu {
+
+bool isSupported(const InferenceEngine::Precision& precision) {
```
dmitry-gorokhov, line 14: 1. Does it make sence to rename into `isSupportedInferencePrecision`? My understanding is that it covers only inference precision use-cases.
2. If 1st item is applicable function should return false by default.
3. Can we reuse this utility in Config::readProperties?

src/plugins/intel_cpu/src/graph.cpp
EgorDuplensky: Make sense,
I will also create keepOriginalOutputPrecision then just to align things.

src/plugins/intel_cpu/src/graph.cpp
EgorDuplensky: My idea was actually to get rid of this check after the release
And replace the checks
`if (node->getOriginalInputPrecisionAtPort(i) != Precision::FP32)`
with
`if (!node->getOriginalInputPrecisionAtPort(i).is_float())`
So it would be possible to force FP32 precision for BF16 and FP16 models, which is impossible at the moment
What do you think?

src/plugins/intel_cpu/src/graph.cpp
EgorDuplensky: Will be updated

EgorDuplensky: Done

src/plugins/intel_cpu/src/graph.cpp
EgorDuplensky: Done,
Haven't created keepOriginalOutputPrecision since it looks a bit to heavy for a simple check

EgorDuplensky:
@dmitry-gorokhov The post commit check was executed for the PR, multiple tests are failed.
The 3rd commit contains test fixes
@alvoron FYI

src/plugins/intel_cpu/tests/functional/single_layer_tests/classes/eltwise.cpp
```diff
@@ -83,6 +85,11 @@ void EltwiseLayerCPUTest::SetUp() {
     ov::test::utils::OpType opType;
     ov::AnyMap additionalConfig;
     std::tie(shapes, eltwiseType, secondaryInputType, opType, netType, inType, outType, targetDevice, additionalConfig) = basicParamsSet;
+    // we have to change model precision as well, otherwise inference precision won't affect single-node graph
+    // due to enforce inference precision optimization for the eltwise as first node of the model
+    if (ov::element::Type(netType).is_real() && additionalConfig.count(ov::hint::inference_precision.name())) {
```
EgorDuplensky, line 90: Another option would be to disable optimizations in EnforceInferencePrecision method for single layer tests

dmitry-gorokhov (approved):

PR #18979, opened 2023-08-04
fix at docs/dev/build_webassembly.md

DmitriyValetov:
Just docs guide fixes


ilya-lavrenov (approved):

jlb6740:
Hi @DmitriyValetov, how do you use this Wasm build? I'd like to do inferencing via openvino compiled to Wasm and I assumed this build allowed for that. If it doesn't allow for that, what does it enable?

PR #18980, opened 2023-08-04
[GPU] restore perf for customer model.

sungeunk:
### Tickets:
 - 116423


isanghao (changes requested):
As discussed offline, it will have negative perf impact on stable diffusion network.

PR #18983, opened 2023-08-04
[GPU] update parsing of CL_DEVICE_IP_VERSION_INTEL (GMD_ID)

sungeunk:
### Tickets:
 - 113779

### DIFF
--- a/src/plugins/intel_gpu/src/runtime/ocl/ocl_device.cpp
+++ b/src/plugins/intel_gpu/src/runtime/ocl/ocl_device.cpp
@@ -119,12 +119,20 @@ device_type get_device_type(const cl::Device& device) {
     return unified_mem ? device_type::integrated_gpu : device_type::discrete_gpu;
 }
 
-gfx_version parse_version(cl_uint ver) {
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/runtime/ocl/ocl_device.cpp
```diff
@@ -119,12 +119,20 @@ device_type get_device_type(const cl::Device& device) {
...
+        {
+            uint32_t revision : 6;
+            uint32_t reserved : 8;
+            uint32_t release : 8;
+            uint32_t architecture : 10;
+        };
+    };
 
-    return {major, minor, revision};
+    GMDID gmd_id = {gmdid};
```
vladimir-paramuzov, line 134: I assume that such parsing will not work for older driver versions. Ideally we need a backward compatible logic here. What if we try to parse id with a new format, then check if the results is meaningful (e.g. check that arch in a range [0, 30] or it's one of [9,10,11,12] list - something like that), and fallback to the previous logic otherwise.

PR #18986, opened 2023-08-04
[GPU] Extended itt profiling for GPU including device timeline

vladimir-paramuzov:
### Details:
 - Added itt timestamps for kernels execution based on event profiling info. 
 - Added more itt timestamps for dynamic pipeline
 - Automatically enable profiling property in case if ITT collection is enabled
 - Bump default OCL version to 210 to be able to use `getDeviceAndHostTimer` API
 - Example how it looks like:
image
- Device timeline doesn't work correctly for dGPU for now, so it's disabled


src/plugins/intel_gpu/src/runtime/ocl/ocl_event.cpp
```diff
@@ -103,6 +104,32 @@ bool ocl_event::get_profiling_info_impl(std::list<instrumentation::profiling_int
     return true;
 }
 
+std::pair<uint64_t, uint64_t> ocl_event::get_host_timestamps(const stream& s) const {
```
sshlyapn, line 107: How about adding `get_host_timestamps()` implementation to `ocl_events` as well?

PR #18993, opened 2023-08-04
[PT FE] Use weight share switch in frontend

mvafin:
### Details:
 - *Use weight share switch in frontend*

### Tickets:
 - *None*


eaidova (approved):

tests/layer_tests/mo_python_api_tests/test_mo_convert_pytorch.py
```diff
@@ -1065,6 +1065,35 @@ def test_mo_import_from_memory(self, create_model, ie_device, precision, ir_vers
...
+                self.data = torch.tensor([1, 2, 3, 4])
+        
+            def forward(self, x):                
+                return x + self.data
+
+        data_model = DataModel()
+        test_input = np.array([1, 2, 3, 4])
+
+        # Convert model to OV
+        ov_model = convert_model(data_model, input=([4], Type.i32), share_weights=False)
```
popovaan, line 1085: Shouldn't here be share_weights=True?

popovaan, line 1085: Can we add two cases with share_weights=True and share_weights=False?

mvafin: No, we check here that weight are not shared

mvafin: Added

popovaan (approved):

PR #18999, opened 2023-08-04
[CPU] MergeTransposeReorder extending

v-Golubev:
### Details:
 - *MergeTransposeReorder: reverse operations order support*
 - *MergeTransposeReorder: reshape between Transpose and Reorder is supported with the following limitations:*
 
    - Direct order (Transpose -> Reshape -> Reorder): reshape which separates one of the dimension on 2 consecutive ones
    - reverse order: (Reorder -> Reshape -> Transpose): reshape which fuses 2 consecutive dimensions into one

 - Common code, which performs the fusion, is moved to a separate function.

### Tickets:
 - *CVS-113363*

### DIFF
--- a/src/common/transformations/include/transformations/utils/utils.hpp
+++ b/src/common/transformations/include/transformations/utils/utils.hpp
@@ -259,12 +259,14 @@ TRANSFORMATIONS_API std::vector<Input<Node>> get_node_target_inputs(const std::s
 TRANSFORMATIONS_API std::shared_ptr<Node> node_to_get_shape_value_of_indices_from_shape_node(
     const std::shared_ptr<Node>& shape_node,
     const std::vector<size_t>& indices,
-    const std::vector<std::shared_ptr<Node>>& copy_rt_info_from = {});
+    const std::vector<std::shared_ptr<Node>>& copy_rt_info_from = {},
+    const ov::element::Type& shape_path_precision = ov::element::i64);
 
 TRANSFORMATIONS_API std::shared_ptr<Node> node_to_get_shape_value_of_indices_from_shape_source(
     const Output<Node>& shape_source,
     const std::vector<size_t>& indices,
-    const std::vector<std::shared_ptr<Node>>& copy_rt_info_from = {});
+    const std::vector<std::shared_ptr<Node>>& copy_rt_info_from = {},
+    const ov::element::Type& shape_path_precision = ov::element::i64);
 
 TRANSFORMATIONS_API bool is_dequantization_subgraph(const Output<Node>& node);
 
--- a/src/plugins/intel_cpu/src/graph_optimizer.cpp
+++ b/src/plugins/intel_cpu/src/graph_optimizer.cpp
@@ -2913,8 +2913,7 @@ void GraphOptimizer::MergeReorderAndTranspose(Graph &graph) {
     };
 
     auto isSuitableReorder = [](NodePtr node) {
-        return node->getType() == Type::Reorder
-                && !node->isDynamicNode();
+        return node->getType() == Type::Reorder && node->getChildEdges().size() == 1 && !node->isDynamicNode();
     };
 
     auto updateOrder = [](const VectorDims& originalOrder, NodePtr reshape) {
--- a/src/plugins/intel_cpu/src/graph_optimizer.cpp
+++ b/src/plugins/intel_cpu/src/graph_optimizer.cpp
@@ -2565,10 +2565,10 @@ void GraphOptimizer::FusePerformedAsScaleShiftAndFakeQuantize(Graph &graph) {
     }
 }
 
-bool GraphOptimizer::checkAscendingSummaryOrder(const VectorDims& transposeOrder,
-                                                const VectorDims& layoutOrder,
-                                                const VectorDims& reorderInOrder,
-                                                const VectorDims& reorderOutOrder) {
+bool GraphOptimizer::checkAscendingFinalOrder(const VectorDims& transposeOrder,
+                                              const VectorDims& layoutOrder,
+                                              const VectorDims& reorderInOrder,
+                                              const VectorDims& reorderOutOrder) {
     if (transposeOrder.size() != layoutOrder.size() || layoutOrder.size() != reorderInOrder.size() ||
         reorderInOrder.size() != reorderOutOrder.size()) {
         return false;
@@ -2802,7 +2802,7 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph& graph) {
 
         // Further logic works with transpose order without Reshape.
         // If there is a Reshape node, which splits one of the dimensions into 2 consecutive ones,
-        // the order must be updated as like Transpose is done after Reshape
+        // the order must be updated as if Transpose is done after Reshape
         // Example. For this sequence:
         // [1,12,5] -> Transpose(0,2,1) -> Reshape(1,5,3,4) -> [1,5,3,4]
         // updated order must be (0,3,1,2):
@@ -2810,7 +2810,7 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph& graph) {
         // - dim idxes which was greater then 1, increments by 1
         const auto& reshapeInShape = reshape->getInputShapeAtPort(0).getDims();
         const auto& reshapeOutShape = reshape->getOutputShapeAtPort(0).getDims();
-        const size_t separationDimIdx = [&]() {
+        const size_t splitDimIdx = [&]() {
             for (size_t i = 0; i < reshapeInShape.size(); ++i) {
                 if (reshapeInShape[i] != reshapeOutShape[i]) {
                     for (size_t j = 0; j < originalOrder.size(); ++j) {
@@ -2819,20 +2819,20 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph& graph) {
                     }
                 }
             }
-            OPENVINO_THROW("separationDimIdx can not be found");
+            OPENVINO_THROW("splitDimIdx can not be found");
         }();
 
         auto transformedOrder = originalOrder;
         auto insertIt = transformedOrder.end();
         for (auto it = transformedOrder.begin(); it != transformedOrder.end(); ++it) {
             auto& elem = *it;
-            if (elem > separationDimIdx) {
+            if (elem > splitDimIdx) {
                 elem++;
-            } else if (elem == separationDimIdx) {
+            } else if (elem == splitDimIdx) {
                 insertIt = it + 1;
             }
         }
-        transformedOrder.insert(insertIt, separationDimIdx + 1);
+        transformedOrder.insert(insertIt, splitDimIdx + 1);
         return transformedOrder;
     };
 
@@ -2876,7 +2876,7 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph& graph) {
         auto& inOrder = inBlockedDesc->getOrder();
         auto& outOrder = outBlockedDesc->getOrder();
 
-        if (checkAscendingSummaryOrder(transposeOrder, layoutOrder, inOrder, outOrder)) {
+        if (checkAscendingFinalOrder(transposeOrder, layoutOrder, inOrder, outOrder)) {
             mergeTransposeReshapeReorder(graph, transposeNode, reshapeNode, reorderNode, false);
         }
     }
@@ -2994,7 +2994,7 @@ void GraphOptimizer::MergeReorderAndTranspose(Graph &graph) {
         auto& inOrder = inBlockedDesc->getOrder();
         auto& outOrder = outBlockedDesc->getOrder();
 
-        if (checkAscendingSummaryOrder(transposeOrder, layoutOrder, inOrder, outOrder)) {
+        if (checkAscendingFinalOrder(transposeOrder, layoutOrder, inOrder, outOrder)) {
             mergeTransposeReshapeReorder(graph, transposeNode, reshapeNode, reorderNode, true);
         }
     }
--- a/src/plugins/intel_cpu/src/graph_optimizer.cpp
+++ b/src/plugins/intel_cpu/src/graph_optimizer.cpp
@@ -201,6 +201,9 @@ void GraphOptimizer::ApplyImplSpecificGraphOptimizations(Graph &graph) {
     MergeTransposeAndReorder(graph);
     graph.RemoveDroppedNodes();
 
+    MergeReorderAndTranspose(graph);
+    graph.RemoveDroppedNodes();
+
     graph.RemoveDroppedEdges();
 }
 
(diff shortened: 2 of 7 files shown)


v-Golubev:
@ceciliapeng2011 Could you please take a look? Thanks in advance

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2398,10 +2401,165 @@ void GraphOptimizer::FusePerformedAsScaleShiftAndFakeQuantize(Graph &graph) {
...
+            break;
+        }
+    }
+    if (!edge) {
+        IE_THROW() << "Parent node '" << parentNode->getName() << "' has invalid edges.";
+    }
+
+    // transposeNode support blocked input & non-blocked output, in the case, the Reorder after Transpose cannot be optimized
+    std::vector<int> srcPerm;
+    if (!reverseOrder) {
```
ceciliapeng2011, line 2517: I think this permute step check is asked for both case of Transpose+Reorder and Reorder+Transpose. We need corresponding unit tests to trigger the conditions for reverseOrder==True case, and address it here.

ceciliapeng2011 (changes requested):

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2424,196 +2582,233 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
 
-        // newTransposeOrder - Transpose layout-aware permutation
-        auto newTransposeOrder = VectorDims(transposeOrder.size());
-        for (size_t i = 0; i < newTransposeOrder.size(); i++) {
-            newTransposeOrder[i] = layoutOrder[transposeOrder[revLayoutOrder[i]]];
-        }
+    auto isSuitableReorder = [](NodePtr node) {
+        return node->getType() == Type::Reorder
+                && node->getChildEdges().size() == 1
+                && !node->isDynamicNode();   // TODO [DS]: enable for dynamic shapes when inPlace in the dynamic case is available (CVS-74863)
```
ceciliapeng2011, line 2608: It looks we can fuse the Reorder and Transpose node with dynamic shapes now. We can remove the limitation if I understand correctly. @dmitry-gorokhov @maxnick Correct me.

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2424,196 +2582,233 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
-        auto reorderOutDesc = outDesc->cloneWithNewPrecision(inPrec);
+void GraphOptimizer::MergeReorderAndTranspose(Graph &graph) {
+    auto& graphNodes = graph.GetNodes();
 
-        std::string reorderlayerName = parentParentNode->getName() + "_" +
-                Reorder::getReorderArgs(*reorderInDesc, *reorderOutDesc) + "_" + "fake";
+    auto isSuitableTranspose = [](NodePtr node) {
+        return node->getType() == Type::Transpose
+                && node->getChildEdges().size() == 1
+                && !node->isDynamicNode();
```
ceciliapeng2011, line 2704: Sounds we can support dynamic case now. No need of such limitation.

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2424,196 +2582,233 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
-                isOptimized = false;
-                // inDesc should be permuted before calling reorder
-                auto & ord = transposeNode->getOrder();
-                srcPerm = std::vector<int>(ord.size());
-                for (size_t i = 0; i < ord.size(); i++) {
-                    srcPerm[ord[i]] = i;
+    auto isSuitableReorder = [](NodePtr node) {
+        return node->getType() == Type::Reorder
+                && node->getChildEdges().size() == 1
+                && !node->isDynamicNode();
```
ceciliapeng2011, line 2730: again, no such limitation.

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2424,196 +2582,233 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
 
-        DEBUG_LOG("mergeTransposeAndReorder ", parentNode->getName(), " and ", childNode->getName(), " -> ", reorderlayerName);
+    auto isSuitableReshape = [](NodePtr node) {
+        if (node->getChildEdges().size() != 1)
+            return false;
+        // Reshape supported only in one case: if two consecutive input dims are merged into 1
+        const auto& inShape = node->getInputShapeAtPort(0).toPartialShape();
+        const auto& outShape = node->getOutputShapeAtPort(0).toPartialShape();
+        if (inShape.size() - outShape.size() != 1)
+            return false;
```
ceciliapeng2011, line 2714: Would you please elaborate this limitation? In my mind, Reshape operator only manipulates shape, and keeps memory data unchanged. So there should be no restriction to it here.

ceciliapeng2011:
Generally it is a feasible PR, but we can make it more applicable by removing some restrictions of the pattern. And can you please add a test case to cover "reorder + transpose" which has no "reshape" in the middle?

src/plugins/intel_cpu/src/graph_optimizer.cpp
ceciliapeng2011: I thought twice. It looks not a simple job to enable the dynamic shape case, as Reorder uses a passthrough shapeinfer. But it is feasible when the Reorder node is optimized. One possible way is to keep reorder of how the reorder is transformed and apply it to shapeinfer.

src/plugins/intel_cpu/src/graph_optimizer.cpp
v-Golubev: As described in the comment, _**"transposeNode support blocked input & non-blocked output, in the case, the Reorder after Transpose cannot be optimized"**_. But we cannot have the opposite situation (non-blocked input but blocked output) (I saw [Transpose::initSupportedPrimitiveDescriptors](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/nodes/transpose.cpp#L98)). Based on this I concluded that in case of reverse order we don't need this check. Am I wrong?

src/plugins/intel_cpu/src/graph_optimizer.cpp
v-Golubev: We have this limitation for reshape because of the `updateOrder` logic. Please see [the comment with an example](https://github.com/openvinotoolkit/openvino/pull/18999/files/fcea9306098ac506f318550a39cd3c1b1fba0026#diff-4596e5068af72535c11480ba99ca338f9097ce55ea043d33a46cddff28c81aedR2738) (it is like a swap of Reshape and Transpose in a graph). This swap-like logic can't be applied to an arbitrary Reshape - so we have to have such limitations (The same is true for the reverse operations order)

src/plugins/intel_cpu/src/graph_optimizer.cpp
v-Golubev: As I mentioned in [the comment](https://github.com/openvinotoolkit/openvino/pull/18999#discussion_r1389651862), we can support Reshape only in some specific cases. And to check that the Reshape satisfies this strict conditions, we need to know input/output shapes.
However, it is probably possible to support Reshape with one dynamic dimension in input/output shapes: in this case we still can understand whether the Reshape satisfies the conditions or not. What do you think?

v-Golubev:
@ceciliapeng2011 thanks a lot for the review. I will try to avoid dynamic shapes limitations if possible.
Regarding tests: we already [have the tests](https://github.com/openvinotoolkit/openvino/blob/fcea9306098ac506f318550a39cd3c1b1fba0026/src/plugins/intel_cpu/tests/functional/subgraph_tests/src/fuse_transpose_reorder.cpp) for "reorder + transpose" without reshape. In this PR, I especially added a separate test class with more complicated subgraph from the real model

ceciliapeng2011:
File fuse_transpose_reorder.cpp are tests of "transpose+reorder" without reshape. And thanks to your new unit test file merge_transpose_reorder.cpp, do love it. It is covering a case of "Transpose+Reshape+Reorder" and another case of "Reorder+Reshape+Transpose", in a combined fashion. So we are still missing "reorder+transpose" w/o reshape, aren't we?

src/plugins/intel_cpu/src/graph_optimizer.cpp
ceciliapeng2011: Considering two situations -
1.  (some layout A) -> Reorder -> (blocked layout B) -> Transpose (non-blocked layout C) https://github.com/openvinotoolkit/oneDNN/blob/2ead5d4fe5993a797d9a7a4b8b5557b96f6ec90e/src/cpu/reorder/cpu_reorder_regular_f32_f32.cpp#L82

In the context of the pattern and current implementation of Reorder and Transpose, there is no reverse order for this situation.

Blocked layout B could only be ncsp8c or ncsp16c, and non-blocked layout C is ncsp; layout A could only be nspc.

2. But the second situation is still possiblely asked for the reverse order. Please refer to src/plugins/intel_cpu/tests/unit/graph/merge_transpose_reorder_test.cpp.

src/plugins/intel_cpu/src/graph_optimizer.cpp
ceciliapeng2011: Yes, the dynamic shape restrictions cannot be easily relaxed. I won't insist on this. Let's consider it probably in a sperate story.

v-Golubev:
@ceciliapeng2011 I applied your comments, could you please review again?

Regarding "reorder+transpose" tests: this case is actually covered by [fuse_transpose_reorder.cpp](https://github.com/openvinotoolkit/openvino/pull/18999/files#diff-40867183141ab095523f40749882ea0602b78841f3e7c92262e8aedbe0dbaccd). Please see my changes in this file: I had to change expected transpose count after MergeReorderTranspose implementation.

src/plugins/intel_cpu/src/graph_optimizer.cpp
ceciliapeng2011: LGTM

ceciliapeng2011 (approved):

v-Golubev:
Internal performance validation passed, no degradations, several improvements (in addition to the target topology from the ticket) detected. All details can be found in the ticket

dmitry-gorokhov:
@maxnick Please review the PR

src/plugins/intel_cpu/tests/functional/custom/subgraph_tests/src/merge_transpose_reorder.cpp
```diff
@@ -0,0 +1,176 @@
...
+        CheckNumberOfNodesWithType(compiledModel, "Transpose", non_optimized_reshape_transpose_count);
+        CheckNumberOfNodesWithType(compiledModel, "Reshape", non_optimized_reshape_transpose_count);
+        size_t fake_reorder_count = 0;
+        for (const auto& node : compiledModel.get_runtime_model()->get_ops()) {
+            const auto& rtInfo = node->get_rt_info();
+            auto it = rtInfo.find(ov::exec_model_info::LAYER_TYPE);
+            IE_ASSERT(rtInfo.end() != it);
+            if (it->second.as<std::string>() == "Reorder" && node->get_friendly_name().find("_fake") != std::string::npos) {
+                fake_reorder_count++;
+            }
```
maxnick, line 140: Such type of a check is not really reliable as the name may be changed by some reason. Could you please elaborate on the reason why we need to count only "fake" reorders? Can we count all the reorders?

src/plugins/intel_cpu/tests/unit/graph/merge_transpose_reorder_test.cpp
```diff
@@ -16,63 +18,97 @@
...
+using LOOK = Edge::LOOK;
+struct MergeTransposeReorderTestParams {
+    enum Result {IS_OPTIMIZED, NOT_OPTIMIZED};
+    MergeTransposeReorderTestParams(const ov::Shape& testShape,
+                                    ov::element::Type_t testPrec,
+                                    LayoutType firstNodeLayout,
+                                    LOOK firstNodeInplaceDirection,
+                                    LayoutType secondNodeLayout,
+                                    LOOK secondNodeInplaceDirection,
+                                    Result test_result)
```
maxnick, line 30: do we really need the constructor while it is possible to use the aggregate initialization?

src/plugins/intel_cpu/tests/unit/graph/merge_transpose_reorder_test.cpp
```diff
@@ -16,63 +18,97 @@
...
+        : testShape(testShape),
+          testPrec(testPrec),
```
maxnick, line 32: Do we need `testPrec`, while `f32` is always used?

maxnick (changes requested):

src/common/transformations/include/transformations/utils/utils.hpp
```diff
@@ -236,12 +236,14 @@ TRANSFORMATIONS_API std::vector<Input<Node>> get_node_target_inputs(const std::s
 TRANSFORMATIONS_API std::shared_ptr<Node> node_to_get_shape_value_of_indices_from_shape_node(
     const std::shared_ptr<Node>& shape_node,
     const std::vector<size_t>& indices,
-    const std::vector<std::shared_ptr<Node>>& copy_rt_info_from = {});
+    const std::vector<std::shared_ptr<Node>>& copy_rt_info_from = {},
+    const ov::element::Type& shape_path_precision = ov::element::i64);
```
maxnick, line 240: May be I missed something, but I didn't see how this new parameter is used in this PR. Do we really need this change?

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2389,190 +2528,233 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
-        auto outBlockedDesc = reorderNode->getSelectedPrimitiveDescriptor()->getConfig().outConfs[0].getMemDesc()->as<BlockedMemoryDesc>();
-
-        auto& inOrder = inBlockedDesc->getOrder();
-        auto& outOrder = outBlockedDesc->getOrder();
-
-        if (transposeOrder.size() != layoutOrder.size() || layoutOrder.size() != inOrder.size() || inOrder.size() != outOrder.size()) {
+        // Reshape supported only in one case: if one of the input dims is split into 2 consecutive dims
+        const auto& inShape = node->getInputShapeAtPort(0).toPartialShape();
+        const auto& outShape = node->getOutputShapeAtPort(0).toPartialShape();
+        if (outShape.is_dynamic() || (outShape.size() - inShape.size() != 1))
```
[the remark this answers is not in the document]
v-Golubev: Agree with you, conversion to PartialShape has been removed

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2389,190 +2528,233 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
-                    reorderOrder[i] = j;
-                    continue;
+    auto updateOrder = [](std::shared_ptr<Transpose> transpose, NodePtr reshape) {
+        const auto originalOrder = transpose->getOrder();
+        if (!reshape)
+            return originalOrder;
+
+        // Further logic works with transpose order without Reshape.
+        // If there is Reshape node which separates one of the dimension on 2 consecutive ones,
+        // the order must be updated correspondigly: Transpose order like moved after reshape
```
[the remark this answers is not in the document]
v-Golubev: I tried to rephrase it, please take a look

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2389,190 +2528,233 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
+        // Example. For this sequence:
+        // [1,12,5] -> Transpose(0,2,1) -> Reshape(1,5,3,4) -> [1,5,3,4]
+        // updated order must be (0,3,1,2):
+        // - dim with idx=1 is separated on 2 ones: 1 and 2
+        // - dim idxes which was greater then 1, increments by 1
+        const auto& reshapeInShape = reshape->getInputShapeAtPort(0).toPartialShape();
+        const auto& reshapeOutShape = reshape->getOutputShapeAtPort(0).toPartialShape();
```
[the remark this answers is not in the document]
v-Golubev: Done, thanks

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2363,10 +2366,146 @@ void GraphOptimizer::FusePerformedAsScaleShiftAndFakeQuantize(Graph &graph) {
...
+        isOptimized = false;
+        // inDesc should be permuted before calling reorder
+        auto& ord = castedTranspose->getOrder();
+        srcPerm = std::vector<int>(ord.size());
+        for (size_t i = 0; i < ord.size(); i++) {
+            srcPerm[ord[i]] = i;
+        }
+    }
+
+    std::string reorderlayerName = nodeBeforeSequence->getName() + "_" + Reorder::getReorderArgs(*reorderInDesc, *reorderOutDesc) + "_fake";
```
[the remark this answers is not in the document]
v-Golubev: We shouldn't I suppose. I updated the logic: the postfix is added only in case of optimal reorder. Thanks for the suggestion

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2389,190 +2528,233 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
+            continue;
         }
 
-        // check that Transpose+Reorder is the identical permutation
-        for (size_t i = 0; i < summaryOrder.size(); i++) {
-            if (summaryOrder[i] != i) {
-                return false;
+        CPU_GRAPH_OPTIMIZER_SCOPE(MergeTransposeAndReorder_ParentNode);
+
+        auto childNode = parentNode->getChildEdgeAt(0)->getChild();
```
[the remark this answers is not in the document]
v-Golubev: Done, thanks

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2363,10 +2366,146 @@ void GraphOptimizer::FusePerformedAsScaleShiftAndFakeQuantize(Graph &graph) {
...
+        graph.RemoveEdge(reshapeNode->getParentEdgeAt(1));
+
+    // to prevent inPlace conflict we must check that the memory reference is unidirectional or
+    // inPlace memory is not used
+    const auto parentInPlace = parentNode->getParentEdgeAt(0)->inPlace(Edge::LOOK_UP);
+    const auto& childEdges = childNode->getChildEdgesAtPort(0);
+    const auto childInPlace = std::any_of(childEdges.begin(), childEdges.end(), [](const EdgePtr& edge) {
+        return edge->inPlace(Edge::LOOK_DOWN);
+    });
+    bool isOptimized = !(parentInPlace && childInPlace);
```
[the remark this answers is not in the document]
v-Golubev: We must perform this check before detaching nodes, because we need to analyze the edges which will be removed. So I can't move this check. I added an explanatory comment in the code

src/plugins/intel_cpu/tests/functional/custom/subgraph_tests/src/merge_transpose_reorder.cpp
```diff
@@ -0,0 +1,176 @@
...
+           |
+     GroupConvolution
+           |
+        Reshape(3D)
+           |
+     Transpose(0, 2, 1)
+           |
+         MatMul
+*/
+using MergeTransposeReorderTestParams = std::tuple<InputShape, ElementType, size_t>;
```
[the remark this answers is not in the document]
v-Golubev: We can easily remove this parameter. Done, thanks

src/plugins/intel_cpu/tests/functional/custom/subgraph_tests/src/merge_transpose_reorder.cpp
v-Golubev: Yes, I agree: the checks are done in a more reliable way now

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2389,190 +2528,233 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
+        if (checkAscendingSummaryOrder(transposeOrder, layoutOrder, inOrder, outOrder)) {
+            mergeTransposeReshapeReorder(graph, transposeNode, reshapeNode, reorderNode, false);
+        }
+    }
+}
 
-        auto reorderInDesc = inDesc;
-        auto reorderOutDesc = outDesc->cloneWithNewPrecision(inPrec);
+void GraphOptimizer::MergeReorderAndTranspose(Graph &graph) {
+    auto& graphNodes = graph.GetNodes();
```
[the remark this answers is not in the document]
v-Golubev: `MergeReorderAndTranspose` and `MergeTransposeAndReorder` passes look similar, but they are quite different in checks part (please take a look at `updateOrder` lambda for example). So I would prefer to have 2 separate transformations in order to avoid a mess in checks.
The part, which performs the fusion, has already moved to a separate function and reused in both transformations.

src/common/transformations/include/transformations/utils/utils.hpp
v-Golubev: `node_to_get_shape_value_of_indices_from_shape_node` is used in function `node_to_get_shape_value_of_indices_from_shape_source`, which in turn is used to create shapeof subgraphs in [merge_transpose_reorder.cpp cpu tests](https://github.com/openvinotoolkit/openvino/pull/18999/files?diff=split&w=0#diff-dc6f3e5783581766a8831fdf0309aecc32f9762b69fad8d2af8567da0185ad32R104). I added `shape_path_precision` parameter to create the shapeof subgraph in i32 precision and avoid unnecessary conversion i64->i32 converstion.

v-Golubev:
@maxnick your comments have been applied. Could you please take a look?

v-Golubev:
@usstq I noticed that my changes had a lot of conflicts with PR #18865. Can I ask you to take a look at this PR in order to make sure I didn't break your logic when fixing the merge conflicts?

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2596,224 +2778,232 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
-        auto* reorderNode = dynamic_cast<Reorder*>(childNode.get());
-        if (!transposeNode || !reorderNode) {
-            return false;
+    auto updateOrder = [](const VectorDims& originalOrder, NodePtr reshape) {
+        if (!reshape)
+            return originalOrder;
+
+        // Further logic works with transpose order without Reshape.
+        // If there is a Reshape node, which splits one of the dimensions into 2 consecutive ones,
+        // the order must be updated as like Transpose is done after Reshape
```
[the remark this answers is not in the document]
v-Golubev: Corrected, thanks

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2596,224 +2778,232 @@ void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
...
+        // Example. For this sequence:
+        // [1,12,5] -> Transpose(0,2,1) -> Reshape(1,5,3,4) -> [1,5,3,4]
+        // updated order must be (0,3,1,2):
+        // - dim with idx=1 is split into 2 parts: 1 and 2
+        // - dim idxes which was greater then 1, increments by 1
+        const auto& reshapeInShape = reshape->getInputShapeAtPort(0).getDims();
+        const auto& reshapeOutShape = reshape->getOutputShapeAtPort(0).getDims();
+        const size_t separationDimIdx = [&]() {
```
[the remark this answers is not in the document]
v-Golubev: Done

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -2570,10 +2573,189 @@ void GraphOptimizer::FusePerformedAsScaleShiftAndFakeQuantize(Graph &graph) {
     }
 }
 
-void GraphOptimizer::MergeTransposeAndReorder(Graph &graph) {
+bool GraphOptimizer::checkAscendingSummaryOrder(const VectorDims& transposeOrder,
```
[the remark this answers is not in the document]
v-Golubev: Renamed

usstq:
> @usstq I noticed that my changes had a lot of conflicts with PR #18865. Can I ask you to take a look at this PR in order to make sure I didn't break your logic when fixing the merge conflicts?

Sure, my PR has a testcase and you didn't break it, so I guess it should be fine, but I will take a look

usstq (approved):
Sorry I didn't have time to go through all logics here, but I checked & ran the testcase & my testcase from PR#18865 and they looks good. My only request is to make the comments more clear by adding more examples with clear notations

maxnick (approved):
LGTM.
@v-Golubev , please resolve the comments from @usstq 
This changes impact overall performance, thus internal performance validation is required.

v-Golubev:
@maxnick internal performance validation is finished. Please check the results in the ticket

PR #19003, opened 2023-08-04
Refactor Parameter builders for tests

olpipi:
### Details:
 - Separate makeParams and makeDynamicParams into a separate header
 - Refactor these functions 
 - Change namespace

### Tickets:
 - CVS-117139


olpipi:
Names I changed are still used in openvino_contrib repo. So I need to add a WA to fix build

src/tests/ngraph_helpers/ngraph_functions/include/ngraph_functions/node_builders/param_vector.hpp
```diff
@@ -0,0 +1,34 @@
...
+//
+
+#include <vector>
+
+#include "openvino/op/parameter.hpp"
+
+namespace ov {
+namespace test {
+namespace utils {
+namespace builder {
```
ilyachur, line 12: Let's reduce number of layers in namespaces
```suggestion
namespace builder {
```

src/tests/ngraph_helpers/ngraph_functions/include/ngraph_functions/node_builders/param_vector.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+ov::ParameterVector makeParams(const ov::element::Type &type, const std::vector<std::vector<size_t>> &shapes);
```
ilyachur, line 14: Why cannot we remove these methods and in tests use `std::make_shared<Parameter>()` in the cycle?
@iefode ??

src/tests/ngraph_helpers/ngraph_functions/include/ngraph_functions/node_builders/param_vector.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+ov::ParameterVector makeDynamicParams(const ov::element::Type &type, const std::vector<ov::PartialShape> &shapes);
+
+ov::ParameterVector makeDynamicParams(const std::vector<ov::element::Type>& types, const std::vector<ov::PartialShape>& shapes);
+
+ov::ParameterVector makeParams(const ov::element::Type &type, const std::vector<std::pair<std::string, std::vector<size_t>>> &inputs);
```
ilyachur, line 20: Also let's use snake_case for naming

src/tests/ngraph_helpers/ngraph_functions/include/ngraph_functions/node_builders/param_vector.hpp
olpipi: `ov::test::builder` will conflict with `ngraph::builder` because a lot of files contains simultaneous `using namespace` for `ov::test` and `ngraph`

We can do this when `ngraph::builder` is removed

src/tests/ngraph_helpers/ngraph_functions/include/ngraph_functions/node_builders/param_vector.hpp
iefode: @ilyachur  I am absolutely ok to remove the all builders :)

iefode: @olpipi Let's move to use OV api instead of builders. My suggestion is to remove theam

PR #19014, opened 2023-08-07
Fix of tensor names loosing for cf=True flag in MOC transformations.

popovaan:
Root cause analysis: 
Names are lost after `validate_nodes_and_infer_types()`. 
In Result `validate_and_infer_types()` output tensor pointer is replaced with input tensor pointer : https://github.com/openvinotoolkit/openvino/blob/648351ba391ce55e17bcc6e758d824a9220f4711/src/core/src/op/result.cpp#L34
So if some tensor name was set in Result output tensor, it becomes lost. 
If we try to set tensor name again after `validate_nodes_and_infer_types()`, it works, as pointers are the same at this point.

The issue appeared when `cf=True` was turned on in MOC transformations.
`cf=True` turns on shape-off constant folding and removes `validate_nodes_and_infer_types()` in the end of MOC transformations pipeline: https://github.com/openvinotoolkit/openvino/blob/648351ba391ce55e17bcc6e758d824a9220f4711/src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp#L258

Solution: 
Add `validate_and_infer_types()` in the end of MOC pipeline if `cf=True`.

Ticket: CVS-117068

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update

### DIFF
--- a/src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
+++ b/src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
@@ -254,7 +254,7 @@ bool ov::pass::MOCTransformations::run_on_model(const std::shared_ptr<ngraph::Fu
         // Restore original shapes to the nGraph Function
         for (auto&& param : f->get_parameters()) {
             param->set_partial_shape(input_shapes.at(param.get()));
-        }   
+        }
     }
     f->validate_nodes_and_infer_types();
 
--- a/src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
+++ b/src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
@@ -254,13 +254,9 @@ bool ov::pass::MOCTransformations::run_on_model(const std::shared_ptr<ngraph::Fu
         // Restore original shapes to the nGraph Function
         for (auto&& param : f->get_parameters()) {
             param->set_partial_shape(input_shapes.at(param.get()));
-        }
-        f->validate_nodes_and_infer_types();
-    } else {
-        for (auto&& result : f->get_results()) {
-            result->validate_and_infer_types();
-        }
+        }   
     }
+    f->validate_nodes_and_infer_types();
 
     return false;
 }
--- a/src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
+++ b/src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
@@ -257,6 +257,11 @@ bool ov::pass::MOCTransformations::run_on_model(const std::shared_ptr<ngraph::Fu
         }
         f->validate_nodes_and_infer_types();
     }
+    else {
+        for (auto&&  result: f->get_results()) {
+            result->validate_and_infer_types();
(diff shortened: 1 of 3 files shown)


src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
```diff
@@ -256,6 +256,10 @@ bool ov::pass::MOCTransformations::run_on_model(const std::shared_ptr<ngraph::Fu
             param->set_partial_shape(input_shapes.at(param.get()));
         }
         f->validate_nodes_and_infer_types();
+    } else {
```
itikhono, line 259: could you add a comment explaining why do we need it here?

itikhono:
Not so sure that this is a correct fix. 
I believe Validation pass should work after Constant folding if CF actually happens and returns the value "true" from the function

src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
```diff
@@ -256,6 +256,10 @@ bool ov::pass::MOCTransformations::run_on_model(const std::shared_ptr<ngraph::Fu
...
+        for (auto&& result : f->get_results()) {
```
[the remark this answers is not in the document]
popovaan: The solution was changed to run extra validate_nodes_and_infer_types().

src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
popovaan: Removed it.

popovaan:
> Not so sure that this is a correct fix. I believe Validation pass should work after Constant folding if CF actually happens and returns the value "true" from the function

As discussed added validate_nodes_and_infer_types() at the end of MOC pipeline.

src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
```diff
@@ -255,8 +255,8 @@ bool ov::pass::MOCTransformations::run_on_model(const std::shared_ptr<ngraph::Fu
         for (auto&& param : f->get_parameters()) {
             param->set_partial_shape(input_shapes.at(param.get()));
         }
-        f->validate_nodes_and_infer_types();
     }
+    f->validate_nodes_and_infer_types();
```
itikhono, line 259: could you add c++ test?
with Add->Add->Add + LinOp fusion transformation and check that we can set the names

popovaan: Done.

src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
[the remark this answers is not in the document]
popovaan: Added cpp test as well.

itikhono (approved):

rkazants (approved):

PR #19020, opened 2023-08-07
[TF FE][TF Hub] Support InvertPermutation and CheckNumerics operations

rkazants:
**Details:** Support InvertPermutation and CheckNumerics operations

**Tickets:** 111315, 111258


src/frontends/tensorflow/src/op_table.cpp
```diff
@@ -169,6 +171,7 @@ const std::map<std::string, CreatorFunction> get_supported_ops() {
         {"Iterator", CreatorFunction(translate_iterator_op)},
         {"IteratorGetNext", CreatorFunction(translate_iterator_get_next_op)},
         {"IteratorV2", CreatorFunction(translate_iterator_op)},
+        {"InvertPermutation", CreatorFunction(translate_invert_permutation_op)},
```
mvafin, line 174: You forgot to add file with this function

rkazants: Added, thanks

mvafin (approved):

PR #19021, opened 2023-08-07
Round ref implementation select optimal solution

praasz:
### Details:
 - The reference implementation of Round operator select optimal solution depends on data type instead of Round's evaluate function.
 - Refactor evaluate function to use `Tensor` instead `HostTensor`.

### Tickets:
 - 114677


src/core/reference/include/openvino/reference/round.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+namespace ov {
+template <class T>
+constexpr bool is_floating_point() {
+    using U = typename std::decay<T>::type;
+    return std::is_floating_point<U>::value || std::is_same<float16, U>::value || std::is_same<bfloat16, U>::value;
+}
+
+namespace reference {
+void set_round_mode(const op::v5::Round::RoundMode& mode);
```
t-jankowski, line 21: Consider passing it by value, it's enum.

src/core/reference/include/openvino/reference/round.hpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/**
+ * @brief Reference implementation of Round operator.
+ *
+ * Used when T is OpenVINO floating type.
+ *
+ * @param arg    Input tensor with data to round.
+ * @param out    Output tensor with rounding result.
+ * @param count  Number of elements in input tensor.
```
t-jankowski, line 54: These are not tensors here, I'd call it buffer or array.

src/core/reference/include/openvino/reference/round.hpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/**
+ * @brief Reference implementation of Round operator.
+ *
+ * Used when T is OpenVINO integral type.
+ *
+ * @param arg    Input tensor with data to round.
+ * @param out    Output tensor with rounding result.
+ * @param count  Number of elements in input tensor.
```
t-jankowski, line 73: As above .. tensor vs buffer

t-jankowski (approved):

src/core/reference/include/openvino/reference/round.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * @brief Rounding algorithm for ov::op::v5::Round::RoundMode::HALF_AWAY_FROM_ZERO.
+ *
+ * @tparam T     Value type.
+ * @param value  Value for rounding.
+ * @return       Rounded value.
+ */
+template <typename T>
+T round_half_away_zero(T value) {
+    return std::round(value);
+}
```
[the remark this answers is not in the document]
praasz: Yes, to select correct overload.

src/core/reference/include/openvino/reference/round.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * @param mode   Rounding mode.
+ */
+template <typename T, typename std::enable_if<ov::is_floating_point<T>()>::type* = nullptr>
+void round(const T* arg, T* out, const size_t count, const op::v5::Round::RoundMode& mode) {
+    set_round_mode(mode);
+    const auto round_algo =
+        (mode == op::v5::Round::RoundMode::HALF_TO_EVEN) ? round_to_nearest_even<T> : round_half_away_zero<T>;
+
+    std::transform(arg, arg + count, out, round_algo);
+}
```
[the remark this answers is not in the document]
praasz: Restore is just a waste. To correct use functions which  depends on this round mode it must be set to correct value before usage restoration is not required.

src/core/reference/src/runtime/reference/round.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include "openvino/reference/round.hpp"
+
+namespace ov {
+namespace reference {
+
+void set_round_mode(const op::v5::Round::RoundMode& mode) {
+    if (mode == op::v5::Round::RoundMode::HALF_TO_EVEN) {
+        std::fesetround(FE_TONEAREST);
+    }
+}
```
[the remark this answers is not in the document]
praasz: Is not required to set, as std::round is used which not depends on this flag.

src/core/reference/include/openvino/reference/round.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include <cfenv>
+#include <cmath>
+#include <type_traits>
+
+#include "openvino/op/round.hpp"
+
+namespace ov {
+template <class T>
+constexpr bool is_floating_point() {
```
ilyachur, line 15: It can be common util function

src/core/reference/include/openvino/reference/round.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * Used when T is OpenVINO floating type.
+ *
+ * @param arg    Input tensor with data to round.
+ * @param out    Output tensor with rounding result.
+ * @param count  Number of elements in input tensor.
+ * @param mode   Rounding mode.
+ */
+template <typename T, typename std::enable_if<ov::is_floating_point<T>()>::type* = nullptr>
+void round(const T* arg, T* out, const size_t count, const op::v5::Round::RoundMode& mode) {
+    set_round_mode(mode);
```
ilyachur, line 59: Should we restore previous round mode at the end of this function?

ilyachur (commented):
In general LGTM

src/core/reference/include/openvino/reference/round.hpp
praasz: Where should be stored in project in (e.g element_type_traits)?

src/core/reference/include/openvino/reference/round.hpp
praasz: Yes, done as simple guard which can be reused in other reference implementations which change rounding direction.

src/core/reference/include/openvino/reference/round.hpp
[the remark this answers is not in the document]
praasz: Changed.

praasz:
> The original rounding mode is not restored after `std::fesetround`, so it can affect other floating point operations.
> 
> Please ensure that existing test cases for Round reference implementation are enough to detect any regression related to proposed changes in the rounding calculations: [/src/plugins/template/tests/functional/op_reference/round.cpp](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/template/tests/functional/op_reference/round.cpp)

The current tests detect wrong rounding.

src/core/reference/include/openvino/reference/round.hpp
ilyachur: It can be a dedicated header in the reference library.

src/core/reference/src/runtime/reference/utils/round_guard.cpp
```diff
@@ -0,0 +1,11 @@
+#include "openvino/reference/round_guard.hpp"
+
+namespace ov {
+RoundGuard::RoundGuard(int mode) : m_prev_round_mode{std::fegetround()} {
```
ilyachur, line 4: What if we call different rounds from different threads? Will it work?

src/core/reference/include/openvino/reference/round.hpp
praasz: OK, Will remember about it and do it next time or if there will any new changes applied into this branch (because CI pass so don't want trigger re-run because of new source file)

mitruska (approved):

PR #19025, opened 2023-08-07
[CPU] Replace memory_order_relaxed with release/acquire in UpdateNodesBase class

alvoron:
There are various issues on ARM related to uninitialised operation executor.
The article https://www.arangodb.com/2021/02/cpp-memory-model-migrating-from-x86-to-arm/ says `memory_order_relaxed` does not guarantee that the initialization is executed before `store` on ARM.
The solution is to use `memory_order_acquire` and `memory_order_release` instead of `memory_order_relaxed`.


alvoron:
The second approach is also being validated: https://github.com/openvinotoolkit/openvino/pull/19028

alvoron:
The second approach was chosen, so this PR is closed.

PR #19028, opened 2023-08-07
[CPU] Replace memory_order_relaxed with release/acquire in UpdateNodesBase class

alvoron:
There are various issues on ARM related to uninitialised operation executor.
The article https://www.arangodb.com/2021/02/cpp-memory-model-migrating-from-x86-to-arm/ says memory_order_relaxed does not guarantee that the initialization is executed before store on ARM.
Tickets: 
CVS-116674
CVS-115839
CVS-113372
CVS-112750


alvoron:
It seems it's enough to change memory order for load operation only, I wasn't able to reproduce the issue with this fix as well.

PR #19035, opened 2023-08-08
[GPU] Fix premature termination of compilation task in tests

Lyamin-Roman:
### Tickets:
 - *117093*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/compilation_context.cpp
+++ b/src/plugins/intel_gpu/src/graph/compilation_context.cpp
@@ -7,6 +7,7 @@
 #include <mutex>
 #include <atomic>
 #include <unordered_set>
+#include <future>
 #include "intel_gpu/runtime/utils.hpp"
 
 namespace cldnn {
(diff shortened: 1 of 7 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -618,6 +619,13 @@ bool primitive_inst::update_impl() {
                             cache.add(updated_params_no_dyn_pad, impl->clone());
                         }
                     });
+                } else {
+                    auto impl = _node->type()->choose_impl(*_node, updated_params_no_dyn_pad);
+                    if (!can_be_optimized()) {
+                        auto kernels = prog->get_kernels_cache().compile(updated_params_no_dyn_pad, impl->get_kernels_source());
+                        impl->set_kernels(kernels);
+                        cache.add(updated_params_no_dyn_pad, impl->clone());
+                    }
```
vladimir-paramuzov, line 628: I don't get the point of these changes. Async compilation is used only if we have shape agnostic impl for current node. It means that if async compilation is disabled for some reason, shape agnostic kernel must be used. 
"Sync" compilation is supposed to happen as a fallback for primitives w/o dynamic impl (and it's covered below)

src/plugins/intel_gpu/src/graph/compilation_context.cpp
```diff
@@ -51,14 +51,13 @@ class CompilationContext : public ICompilationContext {
...
             return;
-
-        _stop_compilation = true;
         {
             std::lock_guard<std::mutex> lock(_mutex);
             if (_task_executor != nullptr)
                 _task_executor.reset();
             _task_keys.clear();
         }
+        _stop_compilation = true;
```
vladimir-paramuzov, line 60: Why?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
Lyamin-Roman: Removed this change

PR #17162, opened 2023-04-24
[cpu] add unit test for custom shape infer 

tiger100256-hu:
### Details:
 - *Extract custom shape infer classes into a separate folder*
     adaptive_pooling, color_convert, eltwise, fullyconnected, gather, matmul, ngram, one_hot, priorbox_clustered
     priorbox, reshape, shapeof, strided_slice, subgraph, transpose
     
 - *ngraph test case also can be used to test custom shape infer*
 - *fix a bug in custom unsqueeze shape infer when there is same patterns, for example {2, -1, -1}* 
 - *fix a bug in custom squeeze shape infer when the patterns is {}*
 -  *add test cases*
 - replace IE_THROW with OPENVINO_THROW in custom shape infer

### Tickets:
 - *108946*

PR #19037, opened 2023-08-08
[CPU] enable skipped custom shape infer test case

tiger100256-hu:
### Details:
 - *enable skipped test case in https://github.com/openvinotoolkit/openvino/pull/17162#discussion_r1256190017*
 - *remove default stride test for strideslice，already set port mask, so stride data should not be null*
 - *remove shapeof OD test, CPU ShapeOfShapeInfer don't support it*
 -  *add NgraphShapeInfer branch when rank0!=rank1 in matmul node*
 - *skip the eltwise none boardcast test and create ticket 122351 for it*
### Tickets:
 - *118109*


yuxu42:
@maxnick Could you please take a review? Thanks!

src/plugins/intel_cpu/tests/unit/shape_inference_test/custom_shape_infer/binary_elementwise_arithmetic.cpp
```diff
@@ -83,7 +83,11 @@ TYPED_TEST_P(CpuShapeInferenceTest_BEA, shape_inference_aubtob_none) {
 }
 
 TYPED_TEST_P(CpuShapeInferenceTest_BEA, shape_inference_aubtob_none_incompatible_shapes) {
-    GTEST_SKIP() << "Skipping test, please check CVS-108946";
+    GTEST_SKIP() << "Skipping test, eltwise only implemented numpy type boardcast";
```
maxnick, line 86: Eltwise shape infer most definitely doesn't support broadcasting types other then numpy. But this may be resolved through switching to the ngraph shape inference when other broadcasting types are specified. This can be done in the shape inference factory. Another question is, whether the Node itself support different broadcasting.

src/plugins/intel_cpu/tests/unit/shape_inference_test/custom_shape_infer/matmul.cpp
```diff
@@ -82,63 +84,60 @@ class CPUMatMulTest : public TestWithParam<matmul_test_params_t> {
...
                                 make_tuple(StaticShape({3}), StaticShape({3})),
                                 make_tuple(StaticShape({5, 2}), StaticShape({2, 6})),
-                                make_tuple(StaticShape({2, 1, 2}), StaticShape({2, 6})),
                                 make_tuple(StaticShape({10, 8, 9, 2}), StaticShape({10, 8, 2, 8})),
                                 make_tuple(StaticShape({3, 1, 4, 3, 4}), StaticShape({3, 2, 1, 4, 1}))),
+                                // make_tuple(StaticShape({1}), StaticShape({1, 3})),
+                                // make_tuple(StaticShape({1}), StaticShape({1, 1, 3})),
+                                // make_tuple(StaticShape({3, 1}), StaticShape({1})),
+                                // make_tuple(StaticShape({3, 2, 1}), StaticShape({1})),
+                                // make_tuple(StaticShape({2, 1, 2}), StaticShape({2, 6})),
```
maxnick, line 140: If such test cases are not supported, and most importantly are not supposed to be supported, then it does make sense to remove them? What do you think?

src/plugins/intel_cpu/src/shape_inference/custom/matmul.cpp
```diff
@@ -29,7 +29,7 @@ Result MMShapeInfer::infer(
     if (rankA == 1 && rankB == 1 && shapeA[0] == shapeB[0]) {
         return {{m_shapeY}, ShapeInferStatus::success};
     }
-
+    OPENVINO_ASSERT(m_out_rank >= 2);
```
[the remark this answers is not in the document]
tiger100256-hu: ok, I will do it

src/plugins/intel_cpu/tests/unit/shape_inference_test/custom_shape_infer/matmul.cpp
tiger100256-hu: I totally agree with that we should support it, from the comment https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/shape_inference/custom/matmul.cpp#L27 , only some rank cases are implemented. I don't have the background information, so I don't know if need to do it

src/plugins/intel_cpu/tests/unit/shape_inference_test/custom_shape_infer/binary_elementwise_arithmetic.cpp
tiger100256-hu: when I added code  to do shape infer for different broadcast property , I found that op lost the broadcast property when converted to cpu graph.
1. sometimes op is converted to powerstatic, but powerstatic lost the broadcast property.
2. when op fuse, there is some issue, for example  Tanh fuse with Add op, the Tanh's broadcast type is null, and
    Add's broadcast type is numpy, so the broadcast type should be numpy, but if check the broadcast type of Tanh, it' s null
I will look deep into it and check how to fix it

maxnick: Ok then, probably we can make a ticket and skip the tests for now.

src/plugins/intel_cpu/tests/unit/shape_inference_test/custom_shape_infer/matmul.cpp
maxnick: From the comment you mentioned, it follows that for the cases of different ranks we have to fallback on the ngraph implementation. It can be done on the factory level pretty easy.

src/plugins/intel_cpu/src/shape_inference/custom/matmul.cpp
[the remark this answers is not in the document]
tiger100256-hu: fixed

src/plugins/intel_cpu/tests/unit/shape_inference_test/custom_shape_infer/binary_elementwise_arithmetic.cpp
tiger100256-hu: already add a ticket for it and skipped the test case

src/plugins/intel_cpu/tests/unit/shape_inference_test/custom_shape_infer/matmul.cpp
tiger100256-hu: add NgrapShapeInfer branch when rank0 != rank1

maxnick (approved):

PR #19039, opened 2023-08-08
[GPU] 7-dimension only supports plain format

kelvinchoi-intel:
### Details:
 - *7-dimension only supports plain format*
 - *Skip concat_input_order opt in case dependancy is dynamic*
 - *Add plain 7d 8d case for jitter pitch size*

### Tickets:
 - *116385*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/concat_input_order.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/concat_input_order.cpp
@@ -118,8 +118,7 @@ void concat_input_order::run(program& p) {
         // 4. Not already aligned
         // 5. Users can accept shuffled features
         // 6. No fused primitives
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/concat_input_order.cpp
```diff
@@ -140,12 +140,20 @@ void concat_input_order::run(program& p) {
         bool single_format = true;
         std::vector<tensor::value_type> feature_sizes;
         feature_sizes.reserve(inputs_count);
+        bool continue_outer = false;
```
yeonbok, line 143: shouldnt we add some other condition in line 122 that input is dynamic shape?

kelvinchoi-intel: Done.

yeonbok (approved):

PR #19040, opened 2023-08-08
[DOCS] Improve NNCF workflow images

sgolebiewski-intel:
Updating the diagram presenting the NNCF workflow. This PR regards the CVS-111571.


helena-intel:
This is a great improvement! I have a few comments:
- "for better inference results it is recommended to provide a quantized model in OpenVINO format" is not clear to me. It is only in the PTQ section, and not clear whether this is important for the source or target model, and what "better inference results" means specifically. I'm thinking that if this is important enough to have as part of the picture (instead of a section in the docs) it should be very clear what it means. 
- The "middle" PyTorch TF and ONNX items in the PTQ part are disconnected. You start with this source model, do convert_model or read_model and end up with an ov.Model. That is true, but not related to PTQ, it is just model conversion. Should these items be connected to  the "quantized python model object"?
- For PTQ we have a quantized python model object->serialize->IR. For QAT optimized python object->convert_model->optimized model. No serialize. I recommend aligning these. Finetuning models with NNCF can take many hours so you definitely want to save the optimized model.

sgolebiewski-intel:
> * "for better inference results it is recommended to provide a quantized model in OpenVINO format" is not clear to me. It is only in the PTQ section, and not clear whether this is important for the source or target model, and what "better inference results" means specifically. I'm thinking that if this is important enough to have as part of the picture (instead of a section in the docs) it should be very clear what it means.

The phrase has been agreed upon with @alexsu52 who provided feedback and insight into the matter.  Another suggestion:

"It is recommended to use ov.Model() as input for nncf.quantize() method to obtain a quantized model for inference in OpenVINO"

> * The "middle" PyTorch TF and ONNX items in the PTQ part are disconnected. You start with this source model, do convert_model or read_model and end up with an ov.Model. That is true, but not related to PTQ, it is just model conversion. Should these items be connected to  the "quantized python model object"?

I believe the sole purpose of presenting the conversion process in this diagram was to highlight the origin of the recommended input for PTQ. I assume that keeping only "ov.Model" at the start should be self-explanatory enough and if there's a decision to keep conversion "out of the way" then I will provide immediate corrections.

> * For PTQ we have a quantized python model object->serialize->IR. For QAT optimized python object->convert_model->optimized model. No serialize. I recommend aligning these. Finetuning models with NNCF can take many hours so you definitely want to save the optimized model.

Again, I have removed the serialize part from QAT per Alexander Suslov's request. @alexsu52, can you relate?

helena-intel:
> "It is recommended to use ov.Model() as input for nncf.quantize() method to obtain a quantized model for inference in OpenVINO"

I like that better! When we have several methods to do things, users will ask "which one should I use", so if we can give a clear recommendation, that's good. But users will still ask why we recommend this, so it would be good to address that (probably not by adding more text to the image itself).

> Again, I have removed the serialize part from QAT per Alexander Suslov's request.

I don't disagree with reducing complexity in the diagram and treating serialization as an implementation detail (explained in the docs itself but not the diagram). But then my suggestion is to remove it from the PTQ part too.

sgolebiewski-intel:
> I like that better! When we have several methods to do things, users will ask "which one should I use", so if we can give a clear recommendation, that's good. But users will still ask why we recommend this, so it would be good to address that (probably not by adding more text to the image itself).

I totally agree. Using extensive wording within the images, especially in the diagrams, is not the way. What about a change in the preceding paragraph and removing the sentence from the image:

image


"

> > Again, I have removed the serialize part from QAT per Alexander Suslov's request.
> 
> I don't disagree with reducing complexity in the diagram and treating serialization as an implementation detail (explained in the docs itself but not the diagram). But then my suggestion is to remove it from the PTQ part too.

I will align PTQ and QAT workflows.

helena-intel:
> I totally agree. Using extensive wording within the images, especially in the diagrams, is not the way. What about a change in the preceding paragraph and removing the sentence from the image:

I like that! I still would like a why though. Or a link to a why. My reason to recommend starting with an ov.Model is that that allows you to easily try AccuracyAware quantization if regular PTQ does not result in good enough accuracy (and it allows fast_bias_correction, which is not available for all source frameworks). That's reason enough for me! But I'm now wondering if that is what we mean or if apart from that ov.Model is still recommended. I hope @alexsu52 can clarify.

sgolebiewski-intel:
> > I totally agree. Using extensive wording within the images, especially in the diagrams, is not the way. What about a change in the preceding paragraph and removing the sentence from the image:
> 
> I like that! I still would like a why though. Or a link to a why. My reason to recommend starting with an ov.Model is that that allows you to easily try AccuracyAware quantization if regular PTQ does not result in good enough accuracy (and it allows fast_bias_correction, which is not available for all source frameworks). That's reason enough for me! But I'm now wondering if that is what we mean or if apart from that ov.Model is still recommended. I hope @alexsu52 can clarify.

It definitely sounds like a reasonable justification for this choice. Let us wait for some more feedback

alexsu52:
> > I totally agree. Using extensive wording within the images, especially in the diagrams, is not the way. What about a change in the preceding paragraph and removing the sentence from the image:
> 
> I like that! I still would like a why though. Or a link to a why. My reason to recommend starting with an ov.Model is that that allows you to easily try AccuracyAware quantization if regular PTQ does not result in good enough accuracy (and it allows fast_bias_correction, which is not available for all source frameworks). That's reason enough for me! But I'm now wondering if that is what we mean or if apart from that ov.Model is still recommended. I hope @alexsu52 can clarify.

Hi @helena-intel, @sgolebiewski-intel. The general justification of using ov.Model instead of a framework model as input of `nncf.quantize(...)` is following:
- The user probably already has an FP32 OpenVINO model, as the user needs to test and measure the performance of the FP32 model.
- Using OpenVINO model excludes additional step is conversion of optimized model to OpenVINO. So the user has less steps in the pipeline => more robust result.
- If NNCF works with the native model representation for OpenVINO runtime, better accuracy and performance can be achieved because NNCF works with a model representation that includes all the general optimizations made during the conversion step.

alexsu52 (commented):
Looks like there is an error in the PTQ part. What do you mean under OpenVINO Runtime? `result = model(input)` If yes, then Pytorch, ONNX and TF optimized model require additional step `convert_model(...)` before using OpenVINO runtime. So if you divide these flows, it will be clear that using ov.Model as input for `nncf.quantize(...)` is preferable.

sgolebiewski-intel:
> Looks like there is an error in the PTQ part. What do you mean under OpenVINO Runtime? `result = model(input)` If yes, then Pytorch, ONNX and TF optimized model require additional step `convert_model(...)` before using OpenVINO runtime. So if you divide these flows, it will be clear that using ov.Model as input for `nncf.quantize(...)` is preferable.

I have made the corrections and have added the information you have provided in the https://github.com/openvinotoolkit/openvino/pull/19040/commits/703d23c0362c4db0988ee06632bd652ef7c7b33d commit.

alexsu52:
> > Looks like there is an error in the PTQ part. What do you mean under OpenVINO Runtime? `result = model(input)` If yes, then Pytorch, ONNX and TF optimized model require additional step `convert_model(...)` before using OpenVINO runtime. So if you divide these flows, it will be clear that using ov.Model as input for `nncf.quantize(...)` is preferable.
> 
> I have made the corrections and have added the information you have provided in the [703d23c](https://github.com/openvinotoolkit/openvino/commit/703d23c0362c4db0988ee06632bd652ef7c7b33d) commit.

Thanks! I kindly recommend to move ov.Model flow in PTQ higher than framework models flow.

docs/optimization_guide/model_optimization_guide.md
```diff
@@ -18,7 +18,10 @@ Model optimization is an optional offline step of improving the final model perf
...
+Post-training Quantization is the fastest way to optimize a model and should be applied first, but it is limited in terms of achievable accuracy-performance trade-off.
+Therefore, it is recommended to use ``ov.Model()`` as input for ``nncf.quantize()`` method to obtain a quantized model for inference in OpenVINO (see the diagram below). Using a model in the OpenVINO format excludes the additional step of converting an optimized model. It means there is fewer steps in the custom training pipeline and that gives more robust results, especially when there is a need to test and measure the performance of an FP32 precision model. Moreover, the native model representation, that includes all the general optimizations made during the conversion step, may result in better accuracy and performance.
+
+However, in case of unsatisfactory accuracy or performance after Post-training Quantization, Training-time Optimization can be used as an option.
```
alexsu52, line 24: Probably, such wording will be useful:

Post-training Quantization is the fastest way to optimize a model and should be applied first, but it is limited in terms of achievable accuracy-performance trade-off. The recommend steps to obtain OpenVINO quantized model is to convert the framework model into OpenVINO model and ensure that the model works correctly in OpenVINO, for example through calculating the model metrics. Then feed OpenVINO model to ``nncf.quantize()` method to get the quantized model.

In case of unsatisfactory accuracy or performance after Post-training Quantization, Training-time Optimization can be used as an option.

cc` @MaximProshin

sgolebiewski-intel, line 24:
```suggestion
Post-training Quantization is the fastest way to optimize a model and should be applied first, but it is limited in terms of achievable accuracy-performance trade-off. The recommended approach to obtain OpenVINO quantized model is to convert a model from original framework to ``ov.Model`` and ensure that the model works correctly in OpenVINO, for example, through calculating the model metrics. Then, ``ov.Model`` can be used as input for the ``nncf.quantize()` method to get the quantized model (see the diagram below).

In case of unsatisfactory accuracy or performance after Post-training Quantization, Training-time Optimization can be used as an option.
```

sgolebiewski-intel: I have updated the article with the https://github.com/openvinotoolkit/openvino/pull/19040/commits/fa5a4bd792525553897c40b6f7b2758586b54be5 commit.

alexsu52 (approved):
LGTM

zKulesza (approved):

PR #19061, opened 2023-08-08
CompressQuantizeWeights - handle decompressed inputs to FakeQuantize

mateusztabaka:
Decompression attribute (that is present in models with FP16 precision) prevents the weights to be constantfolded. Weights constantfolding is required by CompressQuantizeWeights to compress the weights to low precision format.

Ticket: CVS-117310

### DIFF
--- a/src/common/transformations/tests/utils/compress_quantize_weights.cpp
+++ b/src/common/transformations/tests/utils/compress_quantize_weights.cpp
@@ -10,6 +10,7 @@
 #include <ngraph/opsets/opset8.hpp>
 #include <ngraph/pass/manager.hpp>
 #include <transformations/init_node_info.hpp>
+#include <transformations/rt_info/decompression.hpp>
 #include <transformations/utils/utils.hpp>
 
 #include "common_test_utils/ngraph_test_utils.hpp"
@@ -41,8 +42,10 @@ class CompressQuantizeWeightsTests
         std::tie(param, data_prc) = GetParam();
         {
             std::shared_ptr<Node> data = opset8::Constant::create(data_prc, param.shape, param.weights);
-            if (data_prc == element::f16)
+            if (data_prc == element::f16) {
                 data = std::make_shared<opset8::Convert>(data, element::f32);
+                ov::mark_as_decompression(data);
+            }
             auto input_low = opset8::Constant::create(element::f32, Shape{}, {param.in_low});
             auto input_high = opset8::Constant::create(element::f32, Shape{}, {param.in_high});
             auto output_low = opset8::Constant::create(element::f32, Shape{}, {param.out_low});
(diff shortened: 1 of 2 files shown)


pavel-esir (approved):

src/common/transformations/tests/utils/compress_quantize_weights.cpp
```diff
@@ -159,6 +162,41 @@ TEST_F(TransformationTestsF, CompressQuantizeWeightsWithDequantizationSubgraph)
     comparator.enable(FunctionsComparator::CmpValues::ACCURACY);
 }
 
+TEST_F(TransformationTestsF, CompressQuantizeWeightsWithDequantizationSubgraphFP16) {
```
mvafin, line 165: Will this test fail without the changes? I don't understand how this mark affects this transformation.

itikhono: yes, it will fail
we use "get_constant_from_source" in the transformation code, and this method checks
image
the highlighted method calls "evaluate_both_bounds" method which calls "could_propagate" method:
image
and "could_propagate" checks that 
image
it's not that obvious at all. 
@pavel-esir could you comment why do we need "is_decompession" check inside the methods above?

pavel-esir: I added `is_decompression` check inside `ov::could_propagate` in order to disable constant folding for `Const+Convert (with rt_info decompression)` pattern and to do CF on plugin side. I placed this stopper here instead of placing it inside of `ConstantFolding` because some transformations call `get_constant_from_source` and this in the end called expensive `Convert::evaluate`.

At first glance this solution looked ok: this solution removes marking with this attribute calls evaluate and then returns back attribute. But considering that there are question i will double check locally how this solution works

itikhono: It seems quite odd that we have different mechanisms for disabling const_folding:
"disable_const_folding" attribute and this "decompression" attribute (which will be transformed to "disable_const_folding" attribute" via `KeepConstAndDecompression`)

itikhono (approved):

src/common/transformations/tests/utils/compress_quantize_weights.cpp
itikhono: I believe this is ok as a quick fix, but we need to discuss it to make the mechanism more clear.
It is necessary to get rid of ambiguity in the functionality of attributes and clearly define their purpose.

pavel-esir: i tried with `disable_const_folding` but it didn't work for all cases. When `ov::get_constant_from_source()` is called even if `disable_const_folding` is set it didn't prevent from calling `Convert::evaluate()` therefore i added this skipping by `decompression`.

mvafin (approved):

src/common/transformations/tests/utils/compress_quantize_weights.cpp
mvafin: Can we have other transformations that need to use `get_constant_from_source `, but wouldn't be able to?

itikhono: ticket: CVS-117367

PR #19068, opened 2023-08-08
[GPU] Add reorder if the input format of eltwise is different from the output format

andrew-k-park:
### Details:
 - Add reorder if the input format of eltwise is different from the output format during `add_required_reorders` pass

### Tickets:
 - 110281


yeonbok:
As talked offline, this fix only addresses eltwise user of dft. 
We'll make a more general fix which can fix the essential issue after the CF.

yeonbok (approved):

PR #19071, opened 2023-08-08
[ARM CPU] Remove configure from exec func in eltwise, reduce and pooling

allnes:
On our executor's conception:
we call once arm_compute::configure in init() function and call few times arm_compute::run in exec() function

In our cases for eltwise, reduce and pooling we created lambda functions for arm_compute::configure and arm_compute::run in init(), but call every time in exec

this problem could be affect performance and create some problems with threading and streams

This PR fix it and correct realisation with our concept


dmitry-gorokhov:
@allnes We need detailed explanation why this PR neeede in the desctiption. Link to the JIRA (if exists) also required.

allnes:
> @allnes We need detailed explanation why this PR neeede in the desctiption. Link to the JIRA (if exists) also required.

@dmitry-gorokhov wrote, please check

PR #19087, opened 2023-08-09
[GPU] RandomUniform new shape inference for dynamic models support

sshlyapn:
### Details:
 - RandomUniform new shape inference

### Tickets:
 - 117252

### DIFF
--- a/src/plugins/intel_gpu/src/graph/include/random_uniform_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/random_uniform_inst.h
@@ -9,6 +9,17 @@
 
 namespace cldnn {
 
+template <>
+struct typed_program_node<random_uniform> : public typed_program_node_base<random_uniform> {
+    using parent = typed_program_node_base<random_uniform>;
+
+public:
+    using parent::parent;
+
+    program_node& input(size_t index = 0) const { return get_dependency(index); }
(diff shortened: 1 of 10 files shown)


src/plugins/intel_gpu/src/graph/include/random_uniform_inst.h
```diff
@@ -9,6 +9,17 @@
...
 
+template <>
+struct typed_program_node<random_uniform> : public typed_program_node_base<random_uniform> {
+    using parent = typed_program_node_base<random_uniform>;
+
+public:
+    using parent::parent;
+
+    program_node& input(size_t index = 0) const { return get_dependency(index); }
+    std::vector<size_t> get_shape_infer_dependencies() const override { return {0}; }
```
vladimir-paramuzov, line 20: Shouldn't dep 1 and 2 be also here?

sshlyapn: They are some kind of "optional" dependencies (min and max value inputs) - shape infer function may check these values for correctness if they present, but they are not required for shape inference itself. However I tried to avoid additional synchronization work in case of runtime min and max inputs, so I haven't included them here

vladimir-paramuzov (approved):

PR #19090, opened 2023-08-09
[tests] switch layer tests to FP16 on pre-commit

pavel-esir:
### Details:
 - switch layer tests to FP16 on on pre-commit
 - add some layer tests cases with precision sensitive subgraphs

### Tickets:
 - [CVS-116556](https://jira.devtools.intel.com/browse/CVS-116556)


src/common/transformations/src/transformations/common_optimizations/remove_useless_convert_like.cpp
```diff
@@ -0,0 +1,42 @@
...
+//
+
+#include "transformations/common_optimizations/remove_useless_convert_like.hpp"
+
+#include <memory>
+#include <vector>
+
+#include "itt.hpp"
+#include "openvino/core/rt_info.hpp"
+//#include <ngraph/pattern/op/wrap_type.hpp>
```
pavel-esir, line 12: remove this

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
```diff
@@ -143,6 +144,7 @@ void TransformationsPipeline::apply(std::shared_ptr<ov::Model> func) {
         }
 
         manager.register_pass<ov::pass::InitNodeInfo>();
+        manager.register_pass<ov::pass::VisualizeTree>("initial.svg");
```
vladimir-paramuzov, line 147: remove it

src/common/transformations/src/transformations/common_optimizations/remove_useless_convert_like.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "openvino/pass/pattern/op/wrap_type.hpp"
+#include "transformations/utils/utils.hpp"
+
+using namespace ov::op;
+
+ov::pass::RemoveUselessConvertLike::RemoveUselessConvertLike() {
+    MATCHER_SCOPE(RemoveUselessConvertLike);
+
+    const auto const_pattern = pattern::any_input();
```
vladimir-paramuzov, line 21: Not sure what is the original intention, but the word "const" here looks confusing as I'd expect Constant op in the pattern, but it doesn't exits.

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
```diff
@@ -201,6 +203,8 @@ void TransformationsPipeline::apply(std::shared_ptr<ov::Model> func) {
         manager.register_pass<ov::pass::MVN6Decomposition>();
         manager.register_pass<ov::pass::BroadcastTransition>();
 
+        manager.register_pass<ov::pass::ConstantFolding>();
+        manager.register_pass<ov::pass::RemoveUselessConvertLike>();
```
vladimir-paramuzov, line 207: Why it can't be called after convert precision pass (in common optimizations for example)?

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
```diff
@@ -201,6 +203,8 @@ void TransformationsPipeline::apply(std::shared_ptr<ov::Model> func) {
         manager.register_pass<ov::pass::MVN6Decomposition>();
         manager.register_pass<ov::pass::BroadcastTransition>();
 
+        manager.register_pass<ov::pass::ConstantFolding>();
```
vladimir-paramuzov, line 206: I think this check requires testing on wider models scope. Have you run anything to ensure that it doesn't impact the common scenarios?

vladimir-paramuzov (dismissed):

tests/layer_tests/pytorch_tests/pytorch_layer_test_class.py
```diff
@@ -82,6 +81,7 @@ def use_torch_compile_backend():
                 trace_model = kwargs.get('trace_model', False)
                 freeze_model = kwargs.get('freeze_model', True)
                 model, converted_model = self.convert_directly_via_frontend(model, torch_inputs, trace_model, dynamic_shapes, ov_inputs, freeze_model)
+                openvino.runtime.save_model(converted_model, "out.xml")
```
pavel-esir, line 84:
```suggestion
```

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
pavel-esir: oups, sorry for that. removed

src/common/transformations/src/transformations/common_optimizations/remove_useless_convert_like.cpp
pavel-esir: agree, old name was confusing. in the current soslution i removed any c++ changes

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
pavel-esir: for some reason i was getting cumbersome graph with a lot of `ConvertLike`s and they were blocking mixed precision marking. image

Such graph was visible only on GPU, CPU called ConstantFold multiple time and in their `transformation_pipeline.cpp` graph looked good. That was the reason why i added it only for GPU. But now i completely removed this c++ code. If we call the new utility `openvino.tools.ovc.convert_movel` then graph is normal without redundant ConvertLikes. 

So now only python code is in this PR.

src/plugins/intel_gpu/src/plugin/transformations_pipeline.cpp
pavel-esir: answered above

tests/layer_tests/pytorch_tests/pytorch_layer_test_class.py
```diff
@@ -286,8 +285,6 @@ def get_params(ie_device=None, precision=None):
 
     test_args = []
     for element in itertools.product(ie_device_params, precision_params):
-        if element[0] == 'CPU' and element[1] == 'FP16':
```
pavel-esir, line 289: without this tests for fp16 compressed IRs on `CPU` were skipped

mvafin (approved):

rkazants (approved):

.github/workflows/linux.yml
```diff
@@ -617,6 +617,7 @@ jobs:
           python3 -m pytest ${{ env.LAYER_TESTS_INSTALL_DIR }}/pytorch_tests -m precommit --junitxml=${{ env.INSTALL_TEST_DIR }}/TEST-pytorch.xml
         env:
           TEST_DEVICE: CPU
+          TEST_PRECISION: FP16
```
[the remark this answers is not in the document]
pavel-esir: it's `5e-2`. i'll try to decrease it to `1e-3` to see what happens

PR #19096, opened 2023-08-09
vpu 3rd party dependencies

jacekpawlak:
### Details:
 - added 3rd party VPU dependencies to OV 

### Tickets:
 - [115199]


mlyashko:
Please change the flabuffers from DZ ones to https://github.com/google/flatbuffers. Corresponding difference was found, and Protex project was updated

jacekpawlak:
Continued in: https://github.com/openvinotoolkit/openvino/pull/19222

PR #19097, opened 2023-08-09
Fix boxes dim calculation when scores dynamic rank

praasz:
### Details:
 - Improve upper bound calculation in case when:
   * scores has dynamic rank
   * get minimum selected boxes from number of boxes dimension or from `max_output_boxes_per_class` input.

### Tickets:
 - 117923


src/core/shape_inference/include/nms_shape_inference.hpp
```diff
@@ -184,24 +185,13 @@ std::vector<TRShape> shape_infer(const Node* op,
...
-            const auto min_selected_boxes =
-                std::min(num_boxes.get_length(), static_cast<V>(max_out_boxes_per_class_val));
-            selected_boxes = static_output ? TDim{min_selected_boxes} : TDim{0, min_selected_boxes};
-        } else if (scores_rank.is_static() && num_boxes.get_max_length() != -1 &&
-                   scores_shape[0].get_max_length() != -1 && scores_shape[1].get_max_length() != -1) {
+        if (const auto max_out_boxes_per_class = get_input_const_data_as<TRShape, int64_t>(op, 2, ta)) {
+            const auto& num_boxes = boxes_shape[1];
             const auto min_selected_boxes =
-                std::min(num_boxes.get_max_length(), static_cast<V>(max_out_boxes_per_class_val));
+                std::min(num_boxes.get_max_length(), static_cast<V>(max_out_boxes_per_class->front()));
```
mitruska, line 194: If the `num_boxes` hasn't got upper bound, the `get_max_length` will return `-1`, so it will be taken as minimum even if the `max_out_boxes_per_class` is provided.
Consider test case with `boxes_shape` static rank, fully dynamic dimension for `num_boxes`, with provided max_output_boxes_per_class as Constant input.

praasz: If `num_box` has `-1` as upper bound the `min_selected_boxes` cannot be evaluated even if `max_out_boxes_per_class ` is constant, because when static shape provided it can be lower or greater than `max_out_boxes_per_class`, take `-1` is correct.
Regarding the test will check and add if required.

src/core/shape_inference/include/nms_shape_inference.hpp
```diff
@@ -257,7 +245,7 @@ std::vector<TRShape> shape_infer(const NonMaxSuppression* op,
     const auto& boxes_shape = input_shapes[0];
     const auto& scores_shape = input_shapes[1];
 
-    auto output_shapes = std::vector<TRShape>{TRShape{TDim(-1), 3}};
+    auto output_shapes = std::vector<TRShape>{TRShape{TDim(dim::inf_bound), 3}};
     if (boxes_shape.rank().is_static()) {
         const auto max_out_boxes_per_class = get_input_const_data_as<TRShape, int64_t>(op, 2, ta);
         const auto max_out_class_boxes = max_out_boxes_per_class ? max_out_boxes_per_class->front() : dim::inf_bound;
```
mitruska, line 251: Shouldn't the changes be applied also for v4 nms?

src/core/tests/type_prop/non_max_suppression.cpp
```diff
@@ -369,6 +369,29 @@ TYPED_TEST_P(NMSDynamicOutputTest, interval_shapes_labels) {
...
+    const auto scores = make_shared<op::v0::Parameter>(element::f32, scores_shape);
+    const auto max_output_boxes_per_class = op::v0::Constant::create(element::i16, Shape{}, {5});
+    const auto iou_threshold = make_shared<op::v0::Parameter>(element::f32, Shape{});
+    const auto score_threshold = make_shared<op::v0::Parameter>(element::f32, Shape{});
+
+    const auto op = this->make_op(boxes, scores, max_output_boxes_per_class, iou_threshold, score_threshold);
+
+    EXPECT_THAT(op->outputs(),
+                ElementsAre(Property("Indicies shape", &Output<Node>::get_partial_shape, PartialShape({-1, 3})),
+                            Property("Scores shape", &Output<Node>::get_partial_shape, PartialShape({-1, 3})),
```
mitruska, line 388: Following my previous suggestion, in this test case, the expected `selected_boxes` upper bound could be set to `(max_output_boxes_per_class * num_batches * num_classes_upper_bound)` so `(5 * 2 * 5) = 50`. 
```suggestion
                ElementsAre(Property("Indicies shape", &Output<Node>::get_partial_shape, PartialShape({{0, 50}, 3})),
                            Property("Scores shape", &Output<Node>::get_partial_shape, PartialShape({{0, 50}, 3})),
```

src/core/shape_inference/include/nms_shape_inference.hpp
praasz: Can you explain?

src/core/tests/type_prop/non_max_suppression.cpp
praasz: This behavior is to be compatible with previous implementation and selected boxes will calculated if  `num_boxes` has upper bound.

src/core/shape_inference/include/nms_shape_inference.hpp
mitruska: The code for shape_infer in v4 namespace is similar, but the changes in this PR are applied only to the common part.

praasz: The v4 is different and do calculation of selected boxes only if required dimension are static so there is no need to changed.
It could be changed but inference will not work for this version in same way (not sure if any regression can pop-up). The v5 and v9 introduced more dynamic approach.

mitruska: But the case with separate input rank checks is similar and redundant, 
the calculations below `if(boxes_shape.rank().is_static())` 
have sense only when `if(scores_shape.rank().is_static())` is true, 
so it could be a single `if(boxes_shape.rank().is_static() && scores_shape.rank().is_static())`

src/core/shape_inference/include/nms_shape_inference.hpp
```diff
@@ -257,20 +245,17 @@ std::vector<TRShape> shape_infer(const NonMaxSuppression* op,
...
+
+    if (const auto max_out_boxes_per_class = get_input_const_data_as<TRShape, int64_t>(op, 2, ta)) {
+        if (boxes_shape.rank().is_static() && scores_shape.rank().is_static()) {
+            const auto& num_boxes = boxes_shape[1];
+            if (num_boxes.is_static()) {
+                auto& selected_boxes = output_shapes[0][0];
+                selected_boxes = std::min(num_boxes.get_length(), static_cast<V>(max_out_boxes_per_class->front()));
+                selected_boxes *= scores_shape[0].get_max_length();
+                selected_boxes *= scores_shape[1].get_max_length();
+            }
```
mitruska, line 258: In fact `get_input_const_data_as<TRShape, int64_t>(op, 2, ta)` is the more expensive part, 
I would suggest to call as the last check.
```suggestion
    auto output_shapes = std::vector<TRShape>{TRShape{TDim(dim::inf_bound), 3}};

    if (boxes_shape.rank().is_static() && scores_shape.rank().is_static()) {
        if (const auto max_out_boxes_per_class = get_input_const_data_as<TRShape, int64_t>(op, 2, ta)) {
            const auto& num_boxes = boxes_shape[1];
            if (num_boxes.is_static()) {
                auto& selected_boxes = output_shapes[0][0];
                selected_boxes = std::min(num_boxes.get_length(), static_cast<V>(max_out_boxes_per_class->front()));
                selected_boxes *= scores_shape[0].get_max_length();
                selected_boxes *= scores_shape[1].get_max_length();
            }
```

mitruska (approved):

src/core/tests/type_prop/non_max_suppression.cpp
mitruska: Additional case when the bounds are set can be beneficial, but I approve current changes without this potential improvement.

jane-intel (approved):

PR #19101, opened 2023-08-09
[PT FE] aten::concat

eaidova:
### Details:
 - *support aten::concat*
 verified with stable diffusion xl from optimum path

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/frontends/pytorch/src/op/cat.cpp
+++ b/src/frontends/pytorch/src/op/cat.cpp
@@ -44,7 +44,11 @@ OutputVector translate_cat(const NodeContext& context) {
     num_inputs_check(context, 2, 3);
     const auto&& list_elems = get_list_as_outputs(context.get_input(0));
     auto axis = context.const_input<int64_t>(1);
-    return translate_cat_common(context, list_elems, axis);
+    auto out = translate_cat_common(context, list_elems, axis);
+    if (!context.input_is_none(2)){
+        context.mutate_input(2, out[0]);
+    }
+    return out;
 };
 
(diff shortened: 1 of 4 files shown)


src/frontends/pytorch/src/op/cat.cpp
```diff
@@ -41,7 +41,7 @@ OutputVector translate_cat_common(const NodeContext& context,
 
 OutputVector translate_cat(const NodeContext& context) {
     // This translator is only needed to get axis as constant from external scope
-    num_inputs_check(context, 2, 2);
+    num_inputs_check(context, 2, 3);
```
mvafin, line 44: What 3rd input means for `aten::concat` case?

tests/layer_tests/pytorch_tests/test_cat.py
```diff
@@ -7,60 +7,82 @@
...
+        
 
     @pytest.mark.nightly
     @pytest.mark.precommit
-    def test_cat(self, ie_device, precision, ir_version):
-        self._test(aten_cat(), None, ["aten::cat", "prim::ListConstruct"],
-                   ie_device, precision, ir_version)
+    @pytest.mark.parametrize("out", [False, True])
+    def test_cat(self, out, ie_device, precision, ir_version):
+        self._test(aten_cat(out), None, ["aten::cat", "prim::ListConstruct"],
```
mvafin, line 65: I don't see any `aten::concat` here.

mvafin (approved):

PR #19133, opened 2023-08-10
Fixed description for `convert_partially`

slyalin:
The original comment is not correct for some time already.

### DIFF
--- a/src/frontends/common/include/openvino/frontend/frontend.hpp
+++ b/src/frontends/common/include/openvino/frontend/frontend.hpp
@@ -83,9 +83,10 @@ public:
     virtual void convert(const std::shared_ptr<ov::Model>& partially_converted) const;
 
     /// \brief Convert only those parts of the model that can be converted leaving others
-    /// as-is. Converted parts are not normalized by additional transformations; normalize
-    /// Model or another form of convert Model should be called to finalize the
-    /// conversion process.
+    /// as-is wrapped by FrameworkNode. Converted parts are normalized by additional
+    /// transformations like it is done in convert method. If part of the graph cannot be
+    /// converted, it is not guaranteed that the converted regions are complitelly normalized.
+    /// Normalize should be call for each completelly converted parts individually in this case.
     /// \param model Input model
     /// \return partially converted OV Model
     virtual std::shared_ptr<ov::Model> convert_partially(const InputModel::Ptr& model) const;
--- a/src/frontends/common/include/openvino/frontend/frontend.hpp
+++ b/src/frontends/common/include/openvino/frontend/frontend.hpp
@@ -86,7 +86,7 @@ public:
     /// as-is wrapped by FrameworkNode. Converted parts are normalized by additional
     /// transformations like it is done in convert method. If part of the graph cannot be
     /// converted, it is not guaranteed that the converted regions are complitelly normalized.
-    /// Normalize should be call for each completelly converted parts individually in this case.
+    /// Normalize should be called for each completely converted parts individually in this case.
     /// \param model Input model
     /// \return partially converted OV Model
     virtual std::shared_ptr<ov::Model> convert_partially(const InputModel::Ptr& model) const;
--- a/src/frontends/common/include/openvino/frontend/frontend.hpp
+++ b/src/frontends/common/include/openvino/frontend/frontend.hpp
@@ -85,7 +85,7 @@ public:
     /// \brief Convert only those parts of the model that can be converted leaving others
     /// as-is wrapped by FrameworkNode. Converted parts are normalized by additional
     /// transformations like it is done in convert method. If part of the graph cannot be
-    /// converted, it is not guaranteed that the converted regions are complitelly normalized.
+    /// converted, it is not guaranteed that the converted regions are completely normalized.
     /// Normalize should be called for each completely converted parts individually in this case.
     /// \param model Input model
     /// \return partially converted OV Model


src/frontends/common/include/openvino/frontend/frontend.hpp
```diff
@@ -83,9 +83,10 @@ class FRONTEND_API FrontEnd {
     virtual void convert(const std::shared_ptr<ov::Model>& partially_converted) const;
 
     /// \brief Convert only those parts of the model that can be converted leaving others
-    /// as-is. Converted parts are not normalized by additional transformations; normalize
-    /// Model or another form of convert Model should be called to finalize the
-    /// conversion process.
+    /// as-is wrapped by FrameworkNode. Converted parts are normalized by additional
+    /// transformations like it is done in convert method. If part of the graph cannot be
+    /// converted, it is not guaranteed that the converted regions are complitelly normalized.
+    /// Normalize should be call for each completelly converted parts individually in this case.
```
mvafin, line 89:
```suggestion
    /// Normalize should be called for each completelly converted parts individually in this case.
```

src/frontends/common/include/openvino/frontend/frontend.hpp
```diff
@@ -83,9 +83,10 @@ class FRONTEND_API FrontEnd {
     virtual void convert(const std::shared_ptr<ov::Model>& partially_converted) const;
 
     /// \brief Convert only those parts of the model that can be converted leaving others
-    /// as-is. Converted parts are not normalized by additional transformations; normalize
-    /// Model or another form of convert Model should be called to finalize the
-    /// conversion process.
+    /// as-is wrapped by FrameworkNode. Converted parts are normalized by additional
+    /// transformations like it is done in convert method. If part of the graph cannot be
+    /// converted, it is not guaranteed that the converted regions are complitelly normalized.
```
rkazants, line 88:
```suggestion
    /// converted, it is not guaranteed that the converted regions are completely normalized.
```

itikhono: what does it mean "Converted parts" here? 
if it means that we are going to run "normalize" transformations for the part which were wrapped by FrameworkNodes, then we can't mention "like it is done in convert method" here.  Because, for example, "convert" method of TF FE doesn't call  "normalize" for them. Should we fix "convert" method in TF FE?

It's quite strange to see that
auto input_model = fe->load(path);
fe->convert(input_model) // fails
fe->convert_partially(input_model) // works well and the model fully normalized

src/frontends/common/include/openvino/frontend/frontend.hpp
```diff
@@ -83,9 +83,10 @@ class FRONTEND_API FrontEnd {
     virtual void convert(const std::shared_ptr<ov::Model>& partially_converted) const;
```
itikhono, line 83: Should we run normalize method inside "convert" (not this one, in line 79)
 /// \brief Completely convert and **normalize entire Model**, throws if it is not
/// possible

Does this description mean that if we know how to convert model, maybe not with 1:1 fw->ov mapping, but with n:1 or n:n, we should do it?

src/frontends/common/include/openvino/frontend/frontend.hpp
slyalin: @itikhono, `fe->convert_partially(input_model)` will give you a partially converted model. The parts of the model that consist of converted operations ("converted parts") will be normalized by regular normalization passed. But as the model cannot be converted completely then it is not "fully normalized" in your terminology, because normalization passes don't know how to deal with unknown FrameworkNodes and it may lead to under-normalizations. For example, your love -- transpose sinking -- cannot transfer Transpose over FrameworkNode, because the semantics of this node is unknown. FrameworkNode cannot produce output shape and element_type, then it may break some transformations that they won't do their job for the entire model.

It already works in this way.

src/frontends/common/include/openvino/frontend/frontend.hpp
slyalin: To run or not to run `normalize` inside any other method is an FE internal detail. This doesn't need to be specified. Each method has described behavior from the user side -- only this matters. Let me just give you a couple of examples: (1) Imagine you don't have any code in `normalize`, why do you need to call it inside `convert`? (2) `convert` may be implemented in a way that runs the same normalization passed, which `normalize` does, immediately after the conversion like matcher passes that can be run in a pipeline in a single model graph pass. Technically it is possible and it doesn't require running `normalize` as a method inside `convert`.

src/frontends/common/include/openvino/frontend/frontend.hpp
itikhono: in the case above, 
auto input_model = fe->load(path);
fe->convert(input_model) // fails because there are some nodes which are not supported for 1:1 fw -> ov mapping, but can be normalized
fe->convert_partially(input_model) // works well and the model fully normalized
after "convert_partially" method call we will get the model without Framework nodes, these nodes will be converted to TensorIterator and so on, but "convert" method doesn't run the part of transformations in TF FE, which can help to get rid of FrameworkNodes

jane-intel (approved):

mvafin (approved):

slyalin:
FE aligned with new description: Paddle, Tensorflow, Tensorflow Lite, ONNX
Not aligned FEs: Pytorch. Will be aligned (@mvafin)

rkazants (approved):

slyalin:
Alignment PR for Pytorch FE: https://github.com/openvinotoolkit/openvino/pull/19147.

PR #19147, opened 2023-08-11
[PT FE] Fix behavior of convert_partially method to align with new description

mvafin:
### Details:
 - *Fix behavior of convert_partially method to align with new description*

### Tickets:
 - *117687*


eaidova (approved):

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -119,8 +119,20 @@ FrontEnd::FrontEnd() {
...
 }
 
+std::shared_ptr<Model> FrontEnd::convert_denormalized(const ov::frontend::InputModel::Ptr& model) const {
+    FRONT_END_GENERAL_CHECK(std::dynamic_pointer_cast<pytorch::InputModel>(model), "Invalid input model");
+    try {
+        TranslateSession translate_session(model, m_op_translators, m_telemetry);
+        return translate_session.get_converted_model();
+    } catch (const std::runtime_error& e) {
+        std::cerr << "[ ERROR ] Unexpected error while converting pytorch model: " << e.what() << '\n';
+        std::cerr << "Rethrowing.\n";
```
slyalin, line 129: Can we avoid printing any messages? Being integrated into `torch.compile` it may frequently print such messages.

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -145,15 +157,13 @@ void FrontEnd::convert(const std::shared_ptr<Model>& partiallyConverted) const {
...
+    auto partial_model = convert_denormalized(model);
     try {
-        TranslateSession translate_session(model, m_op_translators, m_telemetry);
-        return translate_session.get_converted_model();
-    } catch (const std::runtime_error& e) {
-        std::cerr << "[ ERROR ] Unexpected error while converting pytorch model: " << e.what() << '\n';
-        std::cerr << "Rethrowing. Misleading error message from pybind11 may come next. TODO.";
-        throw;
+        normalize(partial_model);
+    } catch (...) {
```
slyalin, line 163: I think it is overkill, `normalize` shouldn't throw an exception in any case related to not completely supported graphs, it should just leave graph not completely normalized. If exception is thrown then it is something serious that is not related to unsupported operations. Then it is probably better to let it re-throw and then, on the caller side, decide what we should do with that. For example, if `convert_partially` is being used in `torch.compile` they should have try/except block to handle an exception correctly -- but it is a decision on their level, not here.

Such try/catch here in this place in code can be treated as a temporary workaround for known bugs inside `normalize` transformations and shouldn't be kept for in the long term. Is this the case?

slyalin (changes requested):

src/frontends/pytorch/src/frontend.cpp
mvafin: It is not overkill. `normalize` will throw if after some transformation validation can't be done.

src/frontends/pytorch/src/frontend.cpp
mvafin: okay, can delete that

src/frontends/pytorch/src/frontend.cpp
slyalin: But it shouldn't be normal because `normalize` will leave the graph in a wrong state. Even if the graph is partially converted it should pass validation.

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -144,16 +146,16 @@ void FrontEnd::convert(const std::shared_ptr<Model>& partiallyConverted) const {
     FRONT_END_NOT_IMPLEMENTED(convert);
 }
 
-std::shared_ptr<Model> FrontEnd::convert_partially(const ov::frontend::InputModel::Ptr& model) const {
+std::shared_ptr<Model> FrontEnd::convert_partially(const ov::frontend::InputModel::Ptr& model) const {
```
mvafin, line 149:
```suggestion
std::shared_ptr<Model> FrontEnd::convert_partially(const ov::frontend::InputModel::Ptr& model) const {
```

src/frontends/pytorch/src/frontend.cpp
mvafin: Changed as we discussed offline to suppress exception only when model can be validated.

slyalin (approved):

PR #19157, opened 2023-08-11
Fix of confusing logs in convert_model().

popovaan:
Root cause analysis: 
When nncf is imported it shows confusing logs from MO and OVC convert_model().

Solution: 
Check if nncf is in imported modules before importing.

Ticket: 117543

Code:
* [x]  Comments
* [x]  Code style (PEP8) - N/A

Validation:
* [x]  Unit tests - N/A

Documentation:

### DIFF
--- a/tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
+++ b/tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
@@ -19,11 +19,11 @@ def get_pytorch_decoder(model, input_shape, example_inputs, args):
         raise e
     try:
         if 'nncf' in sys.modules:
-            from nncf.torch.nncf_network import NNCFNetwork
+            from nncf.torch.nncf_network import NNCFNetwork # pylint: disable=undefined-variable
             from packaging import version
 
             if isinstance(model, NNCFNetwork):
-                if version.parse(nncf.__version__) < version.parse("2.6"):
+                if version.parse(nncf.__version__) < version.parse("2.6"): # pylint: disable=undefined-variable
                     raise RuntimeError(
                         "NNCF models produced by nncf<2.6 are not supported directly. Please upgrade nncf or export to ONNX first.")
     except:
--- a/tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
+++ b/tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
@@ -20,14 +20,15 @@ def get_pytorch_decoder(model, input_shape, example_inputs, args):
         log.error("PyTorch frontend loading failed")
         raise e
     try:
-        import nncf
-        from nncf.torch.nncf_network import NNCFNetwork
-        from packaging import version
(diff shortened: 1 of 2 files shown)


tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
```diff
@@ -20,14 +18,15 @@ def get_pytorch_decoder(model, input_shape, example_inputs, args):
...
-                    "NNCF models produced by nncf<2.6 are not supported directly. Please export to ONNX first.")
+        if 'nncf' in sys.modules:
+            import nncf
+            from nncf.torch.nncf_network import NNCFNetwork
+            from packaging import version
+
+            if isinstance(model, NNCFNetwork):
+                if version.parse(nncf.__version__) < version.parse("2.6"):
+                    raise RuntimeError(
+                        "NNCF models produced by nncf<2.6 are not supported directly. Please export to ONNX first.")
```
slyalin, line 29:
```suggestion
                        "NNCF models produced by nncf<2.6 are not supported directly. Please upgrade nncf version or export to ONNX first.")
```

slyalin: @mvafin, @alexsu52, please check whether it is what we agreed on.

tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
```diff
@@ -20,14 +18,15 @@ def get_pytorch_decoder(model, input_shape, example_inputs, args):
...
     try:
-        import nncf
-        from nncf.torch.nncf_network import NNCFNetwork
-        from packaging import version
-
-        if isinstance(model, NNCFNetwork):
-            if version.parse(nncf.__version__) < version.parse("2.6"):
-                raise RuntimeError(
-                    "NNCF models produced by nncf<2.6 are not supported directly. Please export to ONNX first.")
+        if 'nncf' in sys.modules:
```
rkazants, line 21: I didn't get it. How does it help to suppress confusing logs?
We get confusing logs during `import nncf` and check `if 'nncf' in sys.modules` will pass.

tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
```diff
@@ -20,14 +18,15 @@ def get_pytorch_decoder(model, input_shape, example_inputs, args):
...
+            import nncf
```
rkazants, line 22: if `nncf` is already imported, why should we import it again? It will lead to emitting the confusing logs, won't it?

rkazants (approved):

slyalin (approved):

tools/mo/openvino/tools/mo/moc_frontend/pytorch_frontend_utils.py
mvafin: I checked and python fails if there is no import

PR #19171, opened 2023-08-14
Fixed 'output_model' logic in OVC.

popovaan:
Root cause analysis: 
Output model is deduced incorrectly if saved model directory is used with separator in the end.
Solution: 
Split input directory in cycle until non-zero name is found.
If no name can be deduced from input directory, set model name to "model".

Ticket: _CVS-117820_

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests

Documentation:

### DIFF
--- a/tests/layer_tests/ovc_python_api_tests/test_ovc_cli_tool.py
+++ b/tests/layer_tests/ovc_python_api_tests/test_ovc_cli_tool.py
@@ -72,7 +72,7 @@ class TestOVCTool(CommonMOConvertTest):
             tf_net = sess.graph_def
 
         # save model to .pb and return path to the model
-        return save_to_pb(tf_net, tmp_dir)
+        return save_to_pb(tf_net, tmp_dir, 'model2.pb')
 
     def create_tf_saved_model_dir(self, temp_dir):
         import tensorflow as tf
@@ -85,7 +85,7 @@ class TestOVCTool(CommonMOConvertTest):
         y = tf.nn.sigmoid(tf.nn.relu(x1 + x2))
         keras_net = tf.keras.Model(inputs=[x1, x2], outputs=[y])
 
-        tf.saved_model.save(keras_net, temp_dir + "/model")
+        tf.saved_model.save(keras_net, temp_dir + "/test_model")
 
         shape = PartialShape([-1, 1, 2, 3])
         param1 = ov.opset8.parameter(shape, name="Input1:0", dtype=np.float32)
@@ -97,7 +97,7 @@ class TestOVCTool(CommonMOConvertTest):
         parameter_list = [param1, param2]
         model_ref = Model([sigm], parameter_list, "test")
 
-        return temp_dir + "/model", model_ref
+        return temp_dir + "/test_model", model_ref
 
 
     def test_ovc_tool(self, ie_device, precision, ir_version, temp_dir, use_new_frontend, use_old_api):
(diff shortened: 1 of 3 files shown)


tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -871,7 +871,13 @@ def get_model_name_from_args(argv: argparse.Namespace):
...
         return output_dir
 
-    input_model_name = os.path.splitext(os.path.split(input_model)[1])[0]
+    input_model_split = os.path.split(input_model)
+    while len(input_model_split[0]) > 0 and len(input_model_split) > 1 and (input_model_split[1] == '' or input_model_split[1] == '.'):
+        input_model_split = os.path.split(input_model_split[0])
+
+    input_model_name = os.path.splitext(input_model_split[1])[0]
+    if input_model_name == '.' or input_model_name == '':
+        input_model_name = "model"
```
rkazants, line 880: I think this logic is overcomplicated and it is easy to make a mistake here. Why don't we use `basename` method?

popovaan: Rewrote this code using `basename` and added comments. Please take a look.

tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -871,7 +871,22 @@ def get_model_name_from_args(argv: argparse.Namespace):
...
         return output_dir
 
-    input_model_name = os.path.splitext(os.path.split(input_model)[1])[0]
+    # get file name
+    input_model_name = os.path.split(input_model)[1]
+
+    # get base directory name if file name is empty
+    while input_model_name == '' and len(input_model) > 0:
+        input_model_name = os.path.basename(input_model)
+        input_model = os.path.split(input_model)[0]
```
rkazants, line 880: can we do this much simpler:

```
input_model_name = os.path.basename(input_model)
if input_model_name == '':
    input_model_name = os.path.basename(os.path.dirname(input_model))
```

tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -871,7 +871,22 @@ def get_model_name_from_args(argv: argparse.Namespace):
...
+
+    # remove extension if exists
+    input_model_name = os.path.splitext(input_model_name)[0]
```
rkazants, line 883: pay attention to that `os.path.splittext` can raise exception of `input_model_name` is still ''

tools/ovc/openvino/tools/ovc/cli_parser.py
popovaan: If input_model is file, for example `../dir/some_model.pb`, then this code will return `dir.xml`
But we need in this case `some_model.xml`.
So I first try to set input_model_name with filename in line 875. And only if it is empty I use directory name.

tools/ovc/openvino/tools/ovc/cli_parser.py
popovaan: `os.path.splittext()` always returns tuple with two  elements. I checked the result of `os.path.splitext('')`
It returns:` ('', '')`
So this line looks safe.

tools/ovc/openvino/tools/ovc/cli_parser.py
rkazants: `input_model_name = os.path.basename(input_model)` will return `some_model.pb`. So 875 line is not needed as well.

rkazants: `while` loop looks strange, please believe)

popovaan: Agree. Removed loop.

tests/layer_tests/ovc_python_api_tests/test_ovc_cli_tool.py
```diff
@@ -74,6 +74,31 @@ def create_tf_model(self, tmp_dir):
...
+
+        input_names = ["Input1", "Input2"]
+        input_shape = [1, 2, 3]
+
+        x1 = tf.keras.Input(shape=input_shape, name=input_names[0])
+        x2 = tf.keras.Input(shape=input_shape, name=input_names[1])
+        y = tf.nn.sigmoid(tf.nn.relu(x1 + x2))
+        keras_net = tf.keras.Model(inputs=[x1, x2], outputs=[y])
+
+        tf.saved_model.save(keras_net, temp_dir + "/model")
```
rkazants, line 88: nit: better to use non-default name, for example, "/test_model"

rkazants (approved):

tests/layer_tests/ovc_python_api_tests/test_ovc_cli_tool.py
popovaan: Done.

PR #19173, opened 2023-08-14
[GPU] Resolve fp16 overflow of mul

byungilm:
+ Fixed black output image by Nan output.
+ Resolved fp16 overflow of gemm primitive before softmax
+ Added fused post ops of clamp activation not to get inf which caused Nan output
+ Added new pass modify_fused_ops

### Details:
 - 

### Tickets:
 - CVS-112747, CVS-114209

### DIFF
--- a/src/plugins/intel_gpu/src/graph/program.cpp
+++ b/src/plugins/intel_gpu/src/graph/program.cpp
@@ -593,6 +593,10 @@ void program::pre_optimize_graph(bool is_internal) {
     // check if there exists some layout incompatibilities and add an reorder node if required
     apply_opt_pass<add_required_reorders>();
 
+    // Modify fused post operation to resolve overflow of fp16 output by adding clamp activation
+    // Currently, 'gemm-softmax' case is applied for clamping
+    apply_opt_pass<clamp_fp16_output>();
+
     // add optimization attributes for onednn primitives
     apply_opt_pass<add_onednn_optimization_attributes>();
 
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/clamp_fp16_output.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/clamp_fp16_output.cpp
@@ -1,0 +1,41 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "pass_manager.h"
+#include "program_node.h"
+
+#include "gemm_inst.h"
+#include "reshape_inst.h"
+#include "softmax_inst.h"
+
+using namespace cldnn;
+
+void clamp_fp16_output::run(program& p) {
+    for (auto& node : p.get_processing_order()) {
+        // Add clamp activation to avoid inf result which causes Nan output
+        if (node->is_type<gemm>() && !node->is_output() && node->get_output_layout().data_type == data_types::f16) {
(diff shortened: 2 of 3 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/modify_fused_ops.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "gemm_inst.h"
+#include "reshape_inst.h"
+#include "softmax_inst.h"
+#include <impls/onednn/utils.hpp>
+#endif
+
+using namespace cldnn;
+
+void modify_fused_ops::run(program& p) {
+#ifdef ENABLE_ONEDNN_FOR_GPU
```
isanghao, line 18: why do we have this `ifdef`? I think we need to remove it because it is necessary for cldnn as well.

src/plugins/intel_gpu/src/graph/graph_optimizer/modify_fused_ops.cpp
```diff
@@ -0,0 +1,46 @@
...
+    for (auto& node : p.get_processing_order()) {
+        // Add clamp activation to avoid inf result which causes Nan output
+        if (node->is_type<gemm>() && !node->is_output() && node->get_output_layout().data_type == data_types::f16) {
+            auto user = node->get_users().front();
+            // Reshape could be added in CreateMatMulOp : check a user node of the Reshape
+            if (user->is_type<reshape>() && user->id().find("_cldnn_out_reshape") != std::string::npos)
```
isanghao, line 24: why do we need `user->id().find("_cldnn_out_reshape")`?

src/plugins/intel_gpu/src/graph/program.cpp
```diff
@@ -592,6 +592,9 @@ void program::pre_optimize_graph(bool is_internal) {
     // check if there exists some layout incompatibilities and add an reorder node if required
     apply_opt_pass<add_required_reorders>();
 
+    // Modify fused post operation before initialization of onednn attributes and compilation
```
isanghao, line 595: * I think the name and description are too general. What about `clamp_gemm_fp16_output`?
* This is not only for onednn but also for cldnn.

byungilm: Modified to clamp_fp16_output.
+ Removed comment about onednn

src/plugins/intel_gpu/src/graph/program.cpp
```diff
@@ -593,6 +593,9 @@ void program::pre_optimize_graph(bool is_internal) {
     // check if there exists some layout incompatibilities and add an reorder node if required
     apply_opt_pass<add_required_reorders>();
 
+    // Modify fused post operation to clamp fp16 output before compilation
```
hyunback, line 596: It always process gemm and softmax pattern only.
Naming for clamp_fp16_output and comment are too general. At least could you update the comment explicitly shows for clamping gemm and softmax case?

isanghao:
No perf difference --this ~/log/pr/2023-08-14--20-18-PR19173.report --ref ~/log/2023-08-14--23-47.report

src/plugins/intel_gpu/src/graph/program.cpp
byungilm: Ok, I will add it.

isanghao (approved):

PR #19182, opened 2023-08-14
[GPU] updated to add reorder after parameter when it is connected to quantize

e-ddykim:
### Details:
 - Quantize impl has performance regression when its output format is different from its input format.
 - This PR partially reverts https://github.com/openvinotoolkit/openvino/pull/18797

### Tickets:
 - 117544

### DIFF
--- a/src/plugins/intel_gpu/src/plugin/ops/parameter.cpp
+++ b/src/plugins/intel_gpu/src/plugin/ops/parameter.cpp
@@ -162,6 +162,16 @@ static void CreateParameterOp(Program& p, const std::shared_ptr<ngraph::op::v0::
         return surface_input_found;
     };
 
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gpu/src/plugin/ops/parameter.cpp
```diff
@@ -162,6 +162,15 @@ static void CreateParameterOp(Program& p, const std::shared_ptr<ngraph::op::v0::
         return surface_input_found;
     };
 
+    std::function<bool(const std::shared_ptr<ov::Node>&)> connected_to_quantize =
```
yeonbok, line 165: Could you add some comments here for future record?

yeonbok (approved):

PR #19184, opened 2023-08-14
Improvements and fixes in OVC convert_model

popovaan:
Description: 
Added tuple support in "input" parameter for setting of multiple inputs.
Removed support of {<type>} syntax from OVC tool.

Ticket: -

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update

### DIFF
--- a/tests/layer_tests/ovc_python_api_tests/test_complex_params.py
+++ b/tests/layer_tests/ovc_python_api_tests/test_complex_params.py
@@ -91,6 +91,18 @@ class TestComplexParams(CommonMOConvertTest):
         {'params_test': {'input': [ov.Type.f32, ov.Type.f32]},
          'params_ref': {'input': 'Input1{f32},Input2{f32}'}},
         {'params_test': {'input': [([1, 3, -1, -1], ov.Type.i32), ov.Type.i32, ov.Type.i32]},
+         'params_ref': {'input': 'Input1[1,3,?,?]{i32},Input2{i32},Input3{i32}'}},
+        {'params_test': {'input': (PartialShape([2, 3, 4]), [2, 3, 4], [Dimension(2), Dimension(3), Dimension(4)])},
+         'params_ref': {'input_shape': "[2,3,4],[2,3,4],[2,3,4]", 'input': 'Input1,Input2,Input3'}},
+        {'params_test': {'input': (PartialShape([1, 3, -1, -1]), [1, 3, -1, -1])},
+         'params_ref': {'input_shape': "[1,3,?,?],[1,3,?,?]", 'input': 'Input1,Input2'}},
+        {'params_test': {'input': ((2, 3, 4), [2, 3, 4], (Dimension(2), Dimension(3), Dimension(4)))},
+         'params_ref': {'input_shape': "[2,3,4],[2,3,4],[2,3,4]", 'input': 'Input1,Input2,Input3'}},
+        {'params_test': {'input': (np.int32, Type(np.int32), np.int32)},
+         'params_ref': {'input': 'Input1{i32},Input2{i32},Input3{i32}'}},
+        {'params_test': {'input': (ov.Type.f32, ov.Type.f32)},
+         'params_ref': {'input': 'Input1{f32},Input2{f32}'}},
+        {'params_test': {'input': (([1, 3, -1, -1], ov.Type.i32), ov.Type.i32, ov.Type.i32)},
          'params_ref': {'input': 'Input1[1,3,?,?]{i32},Input2{i32},Input3{i32}'}}
     ]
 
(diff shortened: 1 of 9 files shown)


tests/layer_tests/ovc_python_api_tests/test_complex_params.py
```diff
@@ -91,6 +91,18 @@ def create_tf_model_single_input_output(tmp_dir):
         {'params_test': {'input': [ov.Type.f32, ov.Type.f32]},
          'params_ref': {'input': 'Input1{f32},Input2{f32}'}},
         {'params_test': {'input': [([1, 3, -1, -1], ov.Type.i32), ov.Type.i32, ov.Type.i32]},
+         'params_ref': {'input': 'Input1[1,3,?,?]{i32},Input2{i32},Input3{i32}'}},
```
slyalin, line 94: I don't understand how these tests are passing if we remove `{type}` syntax.

tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -90,6 +89,27 @@ def single_input_to_input_cut_info(input: [str, tuple, list, PartialShape, Type,
 
     raise Exception("Unexpected object provided for input. Expected tuple, Shape, PartialShape, Type or str. Got {}".format(type(input)))
 
+def is_single_input(input: [str, tuple, list]):
```
slyalin, line 92: Type annotation is a bit confusing, this is not a list of specified types.

tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -592,198 +590,17 @@ def parse_input_value(input_value: str):
     ----------
     input_value
         string with a specified node name, shape, value and data_type.
```
slyalin, line 592:
```suggestion
        string with a specified node name and shape.
```

tools/ovc/openvino/tools/ovc/help.py
```diff
@@ -17,15 +17,12 @@ def get_convert_model_help_specifics():
...
                  'in converted model will match the order of '
                  'specified inputs. The shape is specified as comma-separated list. '
-                 'The data type of input node is specified in braces and can have one of '
-                 'the values: f64, f32, f16, i64, i32, u8, boolean. If data type is not '
-                 'specified explicitly then data type is taken from the '
-                 'original node data type. Example, to set `input_1` input '
-                 'with shape [1,100] and float32 type, and `sequence_len` input '
-                 'with int32 type \"input_1[1,100]{f32},sequence_len{i32}\".'},
+                 'Example, to set `input_1` input with shape [1,100] and `sequence_len` input '
+                 'with shape [1,?]: \"input_1[1,100],sequence_len[1,?]\", where "?" is dynamic dimension. '
```
slyalin, line 24:
```suggestion
                 'with shape [1,?]: \"input_1[1,100],sequence_len[1,?]\", where "?" is a dynamic dimension, '
                 'which means that such a dimension can be specified later in the runtime. '
                 'If the dimension is set as an integer (like 100 in [1,100]), such a dimension is not supposed '
                 'to be changed later, during a model conversion it is treated as a static value. '
```

tools/ovc/openvino/tools/ovc/cli_parser.py
popovaan: Corrected type annotation and added comment to the method.

tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -90,6 +89,32 @@ def single_input_to_input_cut_info(input: [str, tuple, list, PartialShape, Type,
 
     raise Exception("Unexpected object provided for input. Expected tuple, Shape, PartialShape, Type or str. Got {}".format(type(input)))
 
+def is_single_input(input: [tuple, list, str, type, Type, PartialShape, Shape]):
```
rkazants, line 92: Am I correct that this function should accept only list or tuple type because below you expect iterable object?
So we may need to correct signature of this routine:

```
def is_single_input(input: [tuple, list]):
```

tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -90,6 +89,32 @@ def single_input_to_input_cut_info(input: [str, tuple, list, PartialShape, Type,
...
+    :param input: list or tuple of input parameters or input shape or input name.
+    :return: True if input has parameters for single input, otherwise False.
+    """
+    name = None
+    inp_type = None
+    shape = None
+    for val in input:
+        if isinstance(val, str):
+            if name is not None:
+                return False
```
rkazants, line 104: Shouldn't we raise an exception that user input is not correct since user specifies multiple names?

tools/ovc/openvino/tools/ovc/cli_parser.py
popovaan: Yes, replaced with your suggestion.

rkazants (approved):

slyalin (approved):

PR #19185, opened 2023-08-14
[CPU] Remove skip for the first inPlace base edge

maxnick:
### Details:
We need to remove skipping the first inPlace edge connected to the port in the base edge searching algo. This skip breaks the logic of searching the base edge in the case when a Result and InPlace nodes are connected to the same port. The reasons of having this skip are not so obvious, and at the moment this skip looks obsolete.

### Tickets:
 - 117593


maxnick:

dmitry-gorokhov (approved):

PR #19188, opened 2023-08-14
[CPU][ARM][FP16] Use Transpose executor for layout Reorder

EgorDuplensky:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


EgorDuplensky:
will be split into two separate PRs

PR #19194, opened 2023-08-15
[GPU] KV cache optimization

yeonbok:
### Details:
 -  If the user sets the next input  with previous output, we can set the current input primitive with the corresponding previous output's usm memory,  so that we can skip copying data from user blob to current input primitive.

### Tickets:
 - 116112

### DIFF
--- a/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
@@ -212,6 +212,12 @@ public:
     void build_deps();
     void do_runtime_skip_reorder();
     void do_runtime_in_place_concat();
+    void force_realloc_mem() {
+        force_realloc = true;
+    }
+    bool is_forced_realloc_mem() const {
+        return force_realloc;
+    }
     void configure_shape_of_dependencies();
 
     memory::ptr fused_memory(size_t dep_id) const {
@@ -277,6 +283,7 @@ protected:
 
     bool update_shape_done_by_other = false;
     bool allocation_done_by_other = false;
+    bool force_realloc = false;
     std::unique_ptr<kernel_impl_params> _impl_params;
     std::unique_ptr<primitive_impl> _impl;
     std::unique_ptr<primitive_impl> _dynamic_impl = nullptr;
--- a/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
@@ -212,12 +212,10 @@ public:
     void build_deps();
     void do_runtime_skip_reorder();
     void do_runtime_in_place_concat();
-    void set_output_buffer_used_for_next_input() {
-        _impl_params->output_buffer_used_for_next_input = true;
+    void set_output_buffer_used_for_next_input(const primitive_id& input_prim) {
+        _impl_params->output_buffer_used_for_next_input = input_prim;
     }
-    bool is_output_buffer_used_for_next_input() const {
-        return _impl_params->output_buffer_used_for_next_input;
-    }
+
     void configure_shape_of_dependencies();
 
     memory::ptr fused_memory(size_t dep_id) const {
--- a/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
@@ -212,11 +212,11 @@ public:
     void build_deps();
     void do_runtime_skip_reorder();
     void do_runtime_in_place_concat();
-    void force_realloc_mem() {
-        _impl_params->force_realloc = true;
+    void set_output_buffer_used_for_next_input() {
+        _impl_params->output_buffer_used_for_next_input = true;
     }
-    bool is_forced_realloc_mem() const {
-        return _impl_params->force_realloc;
+    bool is_output_buffer_used_for_next_input() const {
+        return _impl_params->output_buffer_used_for_next_input;
     }
     void configure_shape_of_dependencies();
 
--- a/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
@@ -213,10 +213,10 @@ public:
     void do_runtime_skip_reorder();
     void do_runtime_in_place_concat();
     void force_realloc_mem() {
-        force_realloc = true;
+        _impl_params->force_realloc = true;
     }
     bool is_forced_realloc_mem() const {
-        return force_realloc;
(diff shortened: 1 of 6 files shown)


src/plugins/intel_gpu/src/graph/include/primitive_inst.h
```diff
@@ -212,6 +212,10 @@ class primitive_inst {
     void build_deps();
     void do_runtime_skip_reorder();
     void do_runtime_in_place_concat();
+    void set_output_buffer_used_for_next_input(const primitive_id& input_prim) {
+        _impl_params->output_buffer_used_for_next_input = input_prim;
+    }
```
sshlyapn, line 217: How about having some mapping at the network level to configure such dependencies, instead of using kernel_impl_params? Wouldn't that be more straightforward? And also this approach would allow to reset this map every iteration and avoid output's buffer reallocation in case of changes in output's reuse policy (if there will be any...)

src/plugins/intel_gpu/src/graph/network.cpp
```diff
@@ -771,10 +771,26 @@ event::ptr network::set_input_data(const primitive_id& id, memory::ptr data) {
     }
 
     auto input = std::static_pointer_cast<input_layout_inst>(primitive_inst);
-
     return input->set_data(data);
 }
 
+void network::set_output_buffer_used_for_next_input(const primitive_id& out_prim, const primitive_id& in_prim) {
+    // Note that there is an assmption that if the output buffer of previous iteration is used for the input of current iteration
```
sshlyapn, line 778: nit: assmption -> assumption

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -414,7 +414,20 @@ event::ptr primitive_inst::realloc_if_needed() {
         return ev;
     }
 
-    bool can_reuse_buffer = _outputs[0] && actual_layout.count() <= max_output_layout_size;
+    // If the node's output memory is used as next input memory for the same node,
+    // the buffer should be allocated newly every time to prevent data corruption
+    bool output_used_as_its_input_at_next_iter = false;
+    if (_impl_params->output_buffer_used_for_next_input != "") {
+        for (auto dep : _deps) {
```
sshlyapn, line 421: Shouldn't we set `output_used_as_its_input_at_next_iter` immediately if `_impl_params->output_buffer_used_for_next_input != ""`, rather than only if its dependencies are marked as well?

src/plugins/intel_gpu/src/plugin/infer_request.cpp
```diff
@@ -855,7 +855,35 @@ void InferRequest::prepare_input(const cldnn::primitive_id& inputName, Blob::Ptr
     };
 
     auto _nw_ptr = m_graph->GetNetwork();
-    if (input_layout.is_dynamic()) {
```
sshlyapn, line 858: [random spot]
It seems that we should limit the usage of this optimization to scenarios involving single stream and single infer request only, isn't it?

src/plugins/intel_gpu/src/plugin/infer_request.cpp
```diff
@@ -855,7 +855,35 @@ void InferRequest::prepare_input(const cldnn::primitive_id& inputName, Blob::Ptr
...
+                auto prev_output_mem = internal_outputs.at(outputID).get_memory(false);
+                auto prev_output_layout = internal_outputs.at(outputID).get_layout();
+                GPU_DEBUG_TRACE_DETAIL << "Use previous output memory of " << outputID
+                                       << " as current input memory for " << internalInputName << std::endl;
+                if (prev_output_layout == in_layout) {
+                    if (prev_output_mem->get_layout() != prev_output_layout) {
+                        prev_output_mem =
+                            m_graph->get_engine().reinterpret_buffer(*prev_output_mem, prev_output_layout);
+                    }
+                    _nw_ptr->set_output_buffer_used_for_next_input(outputID, internalInputName);
```
vladimir-paramuzov, line 880: That implementation looks quite easy, but honestly, I'd prefer to have buffer swap within infer request only (_device_output and _device_input swap), so the reset of the flow will be exactly the same as for other dynamic cases. That would be much more easy to maintain as swapping logic will be isolated in request.

Likely we'll need an ability to set output memory from request to network for dynamic models as a prerequisite

yeonbok: My initial impl was only swap memory, but there were multiple issues needed to be hanlded internally:
-  if the out mem is to be reused as input : it should not be limited to usm_host & should not allow its memory to be reused
- Also if the output is to be used as input of the same node (e.g., concat), it should allocate new memory for the output 

So the current impl is made.. but basically I also agree that we need to seek for better solution. 

>>Likely we'll need an ability to set output memory from request to network for dynamic models as a prerequisite

The output shape is decided at runtime, but how can we  set output memory from infer request ?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
yeonbok: I think we only need to force new memory for output memory only if the memory is to be used by same node. 
i.e., 
inmem -> concat -> outmem
where outmem is to be used as inmem in the next iter. 
In such case, inmem is to be same as outmem if the memory is not to be allocated newly, 
(e.g., becuase current shape is smaller than previous shape), so that the input data will be currupted.

src/plugins/intel_gpu/src/graph/network.cpp
yeonbok: Fixed, thanks~

src/plugins/intel_gpu/src/graph/include/primitive_inst.h
yeonbok: I also thought about it, but I needed to pass something to the allocate_output of network pointer or some other flag because it is a static function. So I did this for less change. How do you think?

src/plugins/intel_gpu/src/plugin/infer_request.cpp
yeonbok: hmm why so?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
sshlyapn: Agree, thanks for explanation!

src/plugins/intel_gpu/src/plugin/infer_request.cpp
yeonbok: Ah yes, you are right. I will update with fix! thanks!

PR #19203, opened 2023-08-15
Reverted agressive static shape in ovc

slyalin:
Reverted aggressive application of CF in OVC.


rkazants (approved):

slyalin:
This new is better: https://github.com/openvinotoolkit/openvino/pull/19223

PR #19204, opened 2023-08-15
[Coverity] fix coverity

andrei-cv:
### Details:
 - *fix coverity*

### Tickets:
 - *CVS-112753*
 - *CVS-112763*
 - *CVS-112760*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/primitive_base.hpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/primitive_base.hpp
@@ -224,9 +224,9 @@ protected:
     }
 
     kernel_arguments_data get_arguments_impl(const typed_primitive_inst<PType>& instance) const override {
-        for (size_t k = 0; k < _kernels.size(); ++k) {
+        if (_kernels.size()) {
             auto args = get_arguments(instance);
-            args.scalars = &_kernel_data.kernels[k].params.scalars;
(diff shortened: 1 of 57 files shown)


src/plugins/intel_gpu/src/graph/impls/ocl/primitive_base.hpp
```diff
@@ -224,9 +224,9 @@ struct typed_primitive_impl_ocl : public typed_primitive_impl<PType> {
     }
 
     kernel_arguments_data get_arguments_impl(const typed_primitive_inst<PType>& instance) const override {
-        for (size_t k = 0; k < _kernels.size(); ++k) {
+        if (_kernels.size()) {
```
vladimir-paramuzov, line 227: Why?

andrei-cv: The loop makes only one step, if _kernels.size()>0. Return in line 235.

vladimir-paramuzov: @e-ddykim Is it expected behavior?

andrei-cv: It seems that yes.

vladimir-paramuzov (approved):

PR #19206, opened 2023-08-16
[CPU] Add OV_CPU_DUMP_MODELS and OV_CPU_PROFILE

usstq:
### Details:
 -  add *.cpp option to `OV_CPU_EXEC_GRAPH_PATH`  to dump model in pseudo-code style text file (and IR format):
    - this helps developing big models when viewing IR using netron takes so much time and almost unfeasible.
    - the pseudo-code can be used directly as pattern for ngraph transformation (with following PRs implementing the API)

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_cpu/src/utils/debug_capabilities.cpp
+++ b/src/plugins/intel_cpu/src/utils/debug_capabilities.cpp
@@ -7,16 +7,21 @@
 #include <oneapi/dnnl/dnnl.hpp>
 #include "memory_desc/blocked_memory_desc.h"
 #include "onednn/iml_type_mapper.h"
+#include "debug_capabilities.h"
 #ifdef CPU_DEBUG_CAPS
 
-#include "debug_capabilities.h"
 #include "node.h"
 #include "edge.h"
+#include "graph.h"
 #include <iomanip>
+#include <cfloat>
+#include <climits>
 #include "nodes/input.h"
 #include "nodes/eltwise.h"
 #include "snippets/op/subgraph.hpp"
 #include <ie_ngraph_utils.hpp>
+#include <ngraph/pass/manager.hpp>
+#include <openvino/pass/serialize.hpp>
 
 namespace dnnl {
 namespace impl {
@@ -154,6 +159,85 @@ std::ostream & operator<<(std::ostream & os, const Edge& edge) {
     return os;
 }
 
+template<typename DT>
+void dump_tensor(std::ostream& os, void * ptr, const VectorDims& dims) {
+    DT * p = reinterpret_cast<DT*>(ptr);
+    auto rank = dims.size();
+    os << "{";
+
+    if (rank > 1) os << "\n\t";
+
+    auto last_dim_size = dims[rank - 1];
+
+    auto sz = shape_size(dims);
+    std::stringstream ss;
+    int lines = 0;
+    for (size_t i = 0; i < sz; i++) {
+        if (ss.tellp() < 256)
+            ss << p[i] << ",";
+
+        if ((i % last_dim_size) == (last_dim_size - 1)) {
+            os << lines << " : " <<  ss.str() << "...\n\t";
+            ss.str("");
+            lines++;
+            if (lines > 16) {
+                os << "... ... ... ... \n\t";
+                break;
+            }
+        }
+    }
+    os << "}";
+}
+
+std::ostream& operator<<(std::ostream& os, const Memory& mem) {
+    auto nbytes = mem.getSize();
+    void* ptr = mem.getData();
+    auto& dims = mem.getStaticDims();
+    switch (mem.getDataType()) {
+    case dnnl::memory::data_type::bf16:
+        dump_tensor<ov::bfloat16>(os, ptr, dims);
+        break;
+    case dnnl::memory::data_type::f32:
+        dump_tensor<float>(os, ptr, dims);
+        break;
+    case dnnl::memory::data_type::s32:
+        dump_tensor<int32_t>(os, ptr, dims);
+        break;
+    case dnnl::memory::data_type::s8:
+        dump_tensor<int8_t>(os, ptr, dims);
+        break;
+    case dnnl::memory::data_type::u8:
+        dump_tensor<uint8_t>(os, ptr, dims);
+        break;
+    }
+    return os;
+}
+
+bool print_mem_content(std::ostream & os, Node& node) {
+    bool printed = false;
+    if (node.getType() == intel_cpu::Type::Input && node.isConstant()) {
+        if (auto input_node = reinterpret_cast<intel_cpu::node::Input *>(&node)) {
+            auto pmem = input_node->getMemoryPtr();
+            void * data = pmem->getData();
+            auto shape = pmem->getDesc().getShape().getDims();
+            if (shape_size(shape) <= 8) {
+                os << "[";
+                auto type = InferenceEngine::details::convertPrecision(pmem->getDesc().getPrecision());
+                auto tensor = std::make_shared<ngraph::runtime::HostTensor>(type, shape, data);
+                auto constop = std::make_shared<ngraph::op::Constant>(tensor);
+                const char * comma = "";
+                for (auto & v : constop->get_value_strings()) {
+                    os << comma << v;
+                    comma = ",";
+                }
+                os << "]";
+                printed = true;
+            }
+        }
+    }
+    return printed;
+}
+
 std::ostream & operator<<(std::ostream & os, const Node &c_node) {
     Node & node = const_cast<Node &>(c_node);
     const int align_col = 50;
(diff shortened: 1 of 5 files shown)


dmitry-gorokhov:
@usstq  Regarding "OV_CPU_DUMP_MODELS" option. We already have pretty well defined debug logic for graph serialization. This logic allows to dunp graphs after diffrent pipeline stage in different formats. From serialization formats perspective currently we support: cout and xml. 
So I would suggest to reuse existing serialization infrastructure by extending formats list with "cpp". 

@EgorDuplensky  FYI

dmitry-gorokhov:
@usstq  @EgorDuplensky For `OV_CPU_PROFILE` I want to align on the next steps.
I think the feature itself is very useful as long as it provide more convinient (and less overheadable :)) way to get traces comparing with [existing approach](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/docs/performance_analysis_ITT_counters.md) via ITT counters. 
Concern I have is that we create new separate profiling infrastructure (including new PROFILE macro) in addition to existing [perf counters](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/perf_count.h). Moreover we have PR that extends existing infra to cover dynamic shape stages and perform different way of data summarization. The PR requires some remaining fixes which I have locally so hopefully we will be able to productize that PR soon. @usstq Would reccomend you to take a look: https://github.com/openvinotoolkit/openvino/pull/15301.
My view on that we should have unifiend way for collecting performane counters, but can have different post-proccesing logic. Post-processing will include options to dump counters in: 1) agregated tables format and 2) traces format (json).

usstq:
> @usstq Regarding "OV_CPU_DUMP_MODELS" option. We already have pretty well defined debug logic for graph serialization https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/docs/debug_capabilities/graph_serialization.md. This logic allows to dump graphs after different pipeline stage in different formats. From serialization formats perspective currently we support: cout and xml. So I would suggest to reuse existing serialization infrastructure by extending formats list with "cpp".
> 
> @EgorDuplensky FYI

Not exactly, as the [document](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/docs/debug_capabilities/graph_serialization.md) claimed, there are few limitations:
 1. OV_CPU_EXEC_GRAPH_PATH can only dump CPU execution graph, ngraph has it's own environment variable to control how to dump.
 2. when being used for LLM model,  all methods above suffers from terribly slow speed when open the dumpped xml in netron.

That's why I add this new format which use text

usstq:
> @usstq Regarding "OV_CPU_DUMP_MODELS" option. We already have pretty well defined debug logic for graph serialization https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/docs/debug_capabilities/graph_serialization.md. This logic allows to dump graphs after different pipeline stage in different formats. From serialization formats perspective currently we support: cout and xml. So I would suggest to reuse existing serialization infrastructure by extending formats list with "cpp".
> 
> @EgorDuplensky FYI

I can do that for CPU execution graph part, but this OV_CPU_DUMP_MODELS also controls

usstq:
> @usstq Regarding "OV_CPU_DUMP_MODELS" option. We already have pretty well defined debug logic for graph serialization https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/docs/debug_capabilities/graph_serialization.md. This logic allows to dump graphs after different pipeline stage in different formats. From serialization formats perspective currently we support: cout and xml. So I would suggest to reuse existing serialization infrastructure by extending formats list with "cpp".
> 
> @EgorDuplensky FYI

Sure, will use OV_CPU_EXEC_GRAPH_PATH instead

usstq:
> @usstq @EgorDuplensky For `OV_CPU_PROFILE` I want to align on the next steps. I think the feature itself is very useful as long as it provide more convenient (and less overheadable :)) way to get traces comparing with [existing approach](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/docs/performance_analysis_ITT_counters.md) via ITT counters. Concern I have is that we create new separate profiling infrastructure (including new PROFILE macro) in addition to existing [perf counters](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/perf_count.h). Moreover we have PR that extends existing infra to cover dynamic shape stages and perform different way of data summarization. The PR requires some remaining fixes which I have locally so hopefully we will be able to productize it soon. @usstq Would recommend you to take a look: #15301. My view on that we should have unifiend way for collecting performane counters (so reuse infra from PR 15301), but can have different post-proccesing logic. Post-processing will include options to dump counters in: 1) agregated tables format and 2) traces format (json).

I think this Profiler is different from existing one, existing one is included in release (as part of OpenVINO API), but what I proposed here is just a option needs to manually enabled for OV developer, so it shouldn't be aggregated into same infra-structure ( you can see perf counters is always enabled, but OV_CPU_PROFILE is not).

I don't want to make it included in the release because that would require OpenVINO API changes and other-plugins may also need to support chrome-tracing style profiler, so I'd prefer to leave it as developer's helper.

usstq:
@dmitry-gorokhov  Updated, please review again, Thx!

src/plugins/intel_cpu/src/utils/debug_capabilities.cpp
```diff
@@ -596,7 +658,250 @@ std::ostream & operator<<(std::ostream & os, const dnnl::memory::format_tag form
...
+
+    std::ofstream ofs(file_name);
+    if (!ofs) {
+        std::cout << "Error opening file " << file_name << " for output" << std::endl;
+        return false;
+    }
+    dump_cpp_style(ofs, model);
+    ofs.close();
+
+    ov::serialize(model, file_name + ".xml", "/dev/null");
```
EgorDuplensky, line 882: I think we should not silently create an xml file as well.
OV_CPU_DUMP_IR supports specifying multiple formats.

usstq:
@EgorDuplensky the output looks like this:

```
    auto div_Divide = GenPattern<opset1::MatMul>({add_Add, Multiply_113447}, "f32[?,32,?,?]", {{"transpose_a", 0}, {"transpose_b", 1}});   //  __module.model.model.layers.20.self_attn/aten::div/Divide(__module.model.model.layers.20.self_attn/aten::add/Add, Multiply_113447)
    auto add_Add_3864 = GenPattern<opset1::Add>({div_Divide, __module_model_model_aten::add_Add_157}, "f32[?,32,?,?]", {{"auto_broadcast", "numpy"}});   //  __module.model.model.layers.20.self_attn/aten::add/Add_3864(__module.model.model.layers.20.self_attn/aten::div/Divide, __module.model.model/aten::add/Add_157)
    auto softmax_Softmax = GenPattern<opset1::Softmax>({add_Add_3864}, "f32[?,32,?,?]", {{"axis", 3}});   //  __module.model.model.layers.20.self_attn/aten::softmax/Softmax(__module.model.model.layers.20.self_attn/aten::add/Add_3864)
    auto transpose_Transpose_3760 = GenPattern<opset1::Transpose>({view_Reshape_3746, {0, 2, 1, 3}}, "f32[?,32,?,128]");   //  __module.model.model.layers.20.self_attn/aten::transpose/Transpose_3760(__module.model.model.layers.20.self_attn/aten::view/Reshape_3746, __module.model.model.layers.20.self_attn/aten::transpose/ScatterElementsUpdate_3759)

```

this conforms with c++ syntax, with another following-up PR, it can be used as pattern matching template with minimal editing. 

each OP is described as input list/output ele-type & partial shape/attributes map, with orginal ngraph friendly name as comment to easy cross-check.

small 1d constant node (like `{0, 2, 1, 3}`) is directly inline-ed as initializer_list to increase the readibility.

usstq:
> CPU_DEBUG_CAP_TRANSFORMATION_SCOPE

Does it support dump the model in between registered pass? when developing transformation we would like to insert it into an existing `ov::pass::Manager` instance, so we need this.

```cpp
    ov::pass::Manager manager;
    manager.set_per_pass_validation(false);
    CPU_REGISTER_PASS_COMMON(manager, ov::pass::InitNodeInfo);
    CPU_REGISTER_PASS_COMMON(manager, ov::pass::MarkShapeOfSubgraphs);

    //******************** can we dump the model here? ********************

    CPU_REGISTER_PASS_COMMON(manager, ov::pass::KeepConstAndDecompressionForMatMul);
```

EgorDuplensky:
> > CPU_DEBUG_CAP_TRANSFORMATION_SCOPE
> 
> Does it support dump the model in between registered pass? (when developing transformation we would like to insert it into an existing `ov::pass::Manager` instance, and knowing what the model looks like in between passes is required)
> 
> ```c++
>     ov::pass::Manager manager;
>     manager.set_per_pass_validation(false);
>     CPU_REGISTER_PASS_COMMON(manager, ov::pass::InitNodeInfo);
>     CPU_REGISTER_PASS_COMMON(manager, ov::pass::MarkShapeOfSubgraphs);
> 
>     //******************** can we dump the model here? ********************
> 
>     CPU_REGISTER_PASS_COMMON(manager, ov::pass::KeepConstAndDecompressionForMatMul);
> ```

I see your point.
I think then we need to extend/refactor CPU_DEBUG_CAP_TRANSFORMATION_SCOPE, so it will also use a pass manager in order to serialize the model. So instead of calling it before the manager (as we do it at the moment), we should just register it as a first and the last pass of the particular scope (lpt, prelpt, etc). And also it should be possible to register it between other passes with some "custom" scope.
The idea is to generalize the serialization logic and have a common approach instead of multiple ones.
Since it will require quite a major re-implementation of the current approach, I guess we can do it in scope of the following PR.
@usstq What do you think?

PR #19209, opened 2023-08-16
[GPU] Try to use softmax_gpu_ref when types are mismatched (#19209)

steve-y:
### Details:
 - *Try to use softmax_gpu_ref when the input type of softmax that tries to fuse the fakequantize is not same as the output type of the fakequantize*

### Tickets:
 - *113414*

### DIFF
--- a/src/plugins/intel_gpu/tests/unit/fusions/softmax_fusion_test.cpp
+++ b/src/plugins/intel_gpu/tests/unit/fusions/softmax_fusion_test.cpp
@@ -67,9 +67,11 @@ public:
 /* ----------------------------------------------------------------------------------------------------- */
 #define CASE_SOFTMAX_FP32_1 {1, 15, 4, 5}, data_types::f32, format::bfyx, 1, data_types::f32, format::bfyx
 #define CASE_SOFTMAX_FP32_2 {1, 15, 4, 5}, data_types::f32, format::bfyx, 3, data_types::f32, format::bfyx
+#define CASE_SOFTMAX_FP32_3 {1, 15, 1, 1}, data_types::f32, format::bfyx, 1, data_types::f32, format::bfyx
 
 #define CASE_SOFTMAX_FP16_1 {1, 15, 4, 5}, data_types::f16, format::bfyx, 1, data_types::f16, format::bfyx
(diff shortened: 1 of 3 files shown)


src/plugins/intel_gpu/tests/unit/fusions/softmax_fusion_test.cpp
```diff
@@ -71,6 +71,9 @@ class SoftmaxPrimitiveFusingTest : public ::BaseFusingTest<softmax_test_params>
 #define CASE_SOFTMAX_FP16_1 {1, 15, 4, 5}, data_types::f16, format::bfyx, 1, data_types::f16, format::bfyx
 #define CASE_SOFTMAX_FP16_2 {1, 15, 4, 5}, data_types::f16, format::bfyx, 3, data_types::f16, format::bfyx
 
+#define CASE_SOFTMAX_BF_FP32_1 {1, 15, 1, 1}, data_types::f32, format::bfyx, 1, data_types::f32, format::bfyx
```
wilson-seok, line 74: What is the intension of those CASE_SOFTMAX_BF? Please comment detail behavior of those CASE_SOFTMAX_BF.. test cases compared with existing ones.

yeonbok: I also think "BF" is not a good naming here because it can be confused with bf16. 
I dont think we need a new name,... just CASE_SOFTMAX_FP32_XXX is enough.
BTW can these TCs cover the intended issue? (float input and i8 output)

steve-y: > I also think "BF" is not a good naming here because it can be confused with bf16. I dont think we need a new name,... just CASE_SOFTMAX_FP32_XXX is enough. BTW can these TCs cover the intended issue? (float input and i8 output)

The testcase names were modified as you commented. And yes, it was tested as expected and without it the test was failed. Thanks for your comment.

steve-y: > What is the intension of those CASE_SOFTMAX_BF? Please comment detail behavior of those CASE_SOFTMAX_BF.. test cases compared with existing ones.

The test case name was modified as Taylor commented. Thanks for your comment.

wilson-seok (approved):

yeonbok (approved):

PR #19219, opened 2023-08-16
[GPU] add check condition of input dynamic shape in conv fusing

wilson-seok:
### Details:
 - Add check condition of input dynamic shape in conv fusing

### Tickets:
 - 101235

### DIFF
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing.cpp
@@ -494,7 +494,7 @@ void prepare_primitive_fusing::fuse_simple_primitives(program &p) {
             if (_lo.get_optimization_attributes().use_onednn_impls == 1)
                 return true;
 
-            if (node.get_output_layout().is_dynamic()) {
+            if (node.get_output_layout().is_dynamic() || node.get_input_layout().is_dynamic()) {
                 return true;
             }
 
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing.cpp
```diff
@@ -494,7 +494,7 @@ void prepare_primitive_fusing::fuse_simple_primitives(program &p) {
             if (_lo.get_optimization_attributes().use_onednn_impls == 1)
                 return true;
 
-            if (node.get_output_layout().is_dynamic()) {
+            if (node.get_output_layout().is_dynamic() || node.get_input_layout().is_dynamic()) {
```
vladimir-paramuzov, line 497: Why do we need to prohibit fusions with convolution in dynamic case at all? Where the limitation comes from?

wilson-seok: As I know, it is policy for these days. We still have lots of get_dims() in prepare_primitive_fusing and should replace them to partial shape compatible ones. Before it, dynamic case with convolution is prohibited temporarily to avoid runtime error in get_dims(). @yeonbok Could you confirm it?

yeonbok: This is not "prohibit" but "allowing" always because the function is conv_supports_fusings. 
The basic policy is let it support, and unfuse at runtime if it is illegal.
Also currently we are enforcing blocked format for dynamic shape convolution, so most of the cases the below condition hits. 
However,  we may need more conditions like is_grouped_conv.. 

But adding current change is okay I think, because it is aligned with the original idea of the return true if output is dynamic .

yeonbok (approved):

src/plugins/intel_gpu/src/graph/graph_optimizer/prepare_primitive_fusing.cpp
wilson-seok: @vladimir-paramuzov Is this OK and can we go forward to merge?

vladimir-paramuzov (approved):

PR #19220, opened 2023-08-16
Handle unspecified type in wrap_tensor function

itikhono:
### Details:
 - The original model contains Parameter node with "unspecified" type. ConstantFolding tries to create ViewTensor via wrap_tensor function with "unspecified" type. ViewTensor ctor contains an assert for this case. So "unspecified" type was handled in wrap_tensor function.

### Tickets:
 - *CVS-111263*


rkazants (approved):

andrei-kochin (changes requested):
Please wait for release branch

jane-intel (approved):

andrei-kochin (approved):

PR #19233, opened 2023-08-16
Fix TransposeSinking when the data input is broadcasted

itikhono:
### Details:
 - *Handle the case when the input with the target transpose op is broadcasted
     The inserted Transpose after TS had only 2 values in the order input.
```
    Input1 (2, 3) -> Transpose -----
                                       Add (out: 1, 2, 3) -> 
    Input2 (1, 1, 1) ----------------- 
```
 

### Tickets:
 - *CVS-111267*


andrei-kochin (changes requested):
Please wait for release branch

evkotov (approved):

andrei-kochin (approved):

PR #19141, opened 2023-08-11
Fixed checks in snippets_mark_skipped for MatMul weights

antonvor:
Issue: 117681

TODO: add test

PR #19239, opened 2023-08-17
FIxed is_on_constant_path() using in all places

antonvor:
Issues: 117928, 117058
FIx for https://github.com/openvinotoolkit/openvino/pull/19141

PR to release branch: https://github.com/openvinotoolkit/openvino/pull/19256

### DIFF
--- a/src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp
+++ b/src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp
@@ -254,7 +254,7 @@ bool isSuitableChildForFusingMatMul(const std::shared_ptr<const Node> &node, con
     ov::PartialShape matmul_shape;
     for (const auto &parent_out : node->input_values()) {
         const auto parent = parent_out.get_node_shared_ptr();
-        if (ov::op::util::is_on_constant_path(parent)) {
+        if (ov::op::util::is_on_constant_path(parent->output(0))) {
             bias_shape = parent_out.get_shape();
             num_non_const_inputs++;
         } else {
--- a/src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp
+++ b/src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp
@@ -254,7 +254,7 @@ bool isSuitableChildForFusingMatMul(const std::shared_ptr<const Node> &node, con
     ov::PartialShape matmul_shape;
     for (const auto &parent_out : node->input_values()) {
         const auto parent = parent_out.get_node_shared_ptr();
-        if (ov::op::util::is_on_constant_path(parent->output(0))) {
+        if (ov::op::util::is_on_constant_path(parent_out)) {
             bias_shape = parent_out.get_shape();
             num_non_const_inputs++;
         } else {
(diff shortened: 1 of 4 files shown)


antonvor:
@itikhono @dmitry-gorokhov take a look please?

src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp
```diff
@@ -254,7 +254,7 @@ bool isSuitableChildForFusingMatMul(const std::shared_ptr<const Node> &node, con
     ov::PartialShape matmul_shape;
     for (const auto &parent_out : node->input_values()) {
         const auto parent = parent_out.get_node_shared_ptr();
-        if (ov::op::util::is_on_constant_path(parent)) {
+        if (ov::op::util::is_on_constant_path(parent->output(0))) {
```
dmitry-gorokhov, line 257: Shouldn't it be `parent_out` instead?
As far as I understand there is no guarantee `parent` is connected with `node` via 0th port.

v-Golubev:
@antonvor [this line](https://github.com/openvinotoolkit/openvino/blob/95300ac5bfc3f60ded81aa3425b550421d88efbe/src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp#L268) should be also fixed in the same way: we should call `is_on_constant_path` for output, not node

src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp
antonvor: changed to `parent_out`, thanks

antonvor:
@v-Golubev thank you for your comment, fixed
https://github.com/openvinotoolkit/openvino/pull/19239/files#diff-c19b63cc114defd4e10f8dc600407fc4c3d6a5e38d111116a732bc5017a28623R268

itikhono:
Could you fix other "is_on_constant_path" usages? maybe in the next PR 
image

src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp
```diff
@@ -254,7 +254,7 @@ bool isSuitableChildForFusingMatMul(const std::shared_ptr<const Node> &node, con
     ov::PartialShape matmul_shape;
     for (const auto &parent_out : node->input_values()) {
         const auto parent = parent_out.get_node_shared_ptr();
-        if (ov::op::util::is_on_constant_path(parent)) {
```
itikhono, line 257: should we add some test for this case?

itikhono (approved):

antonvor:
> Could you fix other "is_on_constant_path" usages? maybe in the next PR image

I fixed check in `ConvertMatMulToFC()` pass only, let's update other places in separate PR

src/plugins/intel_cpu/src/transformations/snippets/x64/pass/snippets_mark_skipped.cpp
antonvor: added `SplitMatMulConcat` subgraph test for this case

v-Golubev (approved):

PR #19244, opened 2023-08-17
Fix uninit members in default `GroupNormalization()`

vurusovs:
### Tickets:
 - CVS-117984

### DIFF
--- a/src/core/src/op/group_normalization.cpp
+++ b/src/core/src/op/group_normalization.cpp
@@ -10,6 +10,8 @@
 #include "openvino/core/validation_util.hpp"
 
 namespace ov {
+op::v12::GroupNormalization::GroupNormalization() : m_num_groups{0}, m_epsilon{0} {}
+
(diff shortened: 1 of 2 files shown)


ilya-lavrenov (approved):

src/core/src/op/group_normalization.cpp
```diff
@@ -10,6 +10,8 @@
 #include "openvino/core/validation_util.hpp"
 
 namespace ov {
+op::v12::GroupNormalization::GroupNormalization() : m_num_groups{0}, m_epsilon{0} {}
```
t-jankowski, line 13: Consider setting default values to members in class definition -> One file changed and ctor remains `default`.

t-jankowski: I don't find it as an argument.

jane-intel (approved):

jane-intel:
This goes to the master, so I merge it

PR #19257, opened 2023-08-18
[GNA] Fix memory leak in GNA plugin.

marcinkusm:
* Disabled transformation introducing memory leak.

### Details:
* Disabled transformation introducing memory leak.

### Tickets:
 - *116912*


src/plugins/intel_gna/src/gna_transformations_pipeline.cpp
```diff
@@ -208,10 +209,9 @@ void TransformationsPipeline::apply_legacy(const InferenceEngine::CNNNetwork& ne
     }
 
     passes->registerPass<InsertSplitAligningFilterPass>();
+    // TODO enable legacy pass instead of disabled one.
```
sirzabek, line 212: I don't think we need todos in this PR.

sirzabek (approved):

PR #19259, opened 2023-08-18
classification_sample_async: state that the samples support NCHW model layout only

Wovchena:
Ticket 107409


ilyachur (approved):
Should we port this PR to the release branch?

PR #19261, opened 2023-08-18
[CPU] Fix deconvolution default primitive search algo

maxnick:
### Details:
The current implementation contains a bug in the fall back primitive search algorithm. This led to selecting a reference implementation instead of optimized one as the fallback oneDNN primitive.

### Tickets:
 - 118178


maxnick:
@EgorDuplensky , FIY

dmitry-gorokhov (approved):

PR #19266, opened 2023-08-18
[GNA] Fix memory leak in insert_copy_layer.cpp

marcinkusm:
### Details:
 - Added cleanup transformation for inert copy layer transformations

### Tickets:
 - *116912*

### DIFF
--- a/src/plugins/intel_gna/src/gna_transformations_pipeline.cpp
+++ b/src/plugins/intel_gna/src/gna_transformations_pipeline.cpp
@@ -192,6 +192,8 @@ void TransformationsPipeline::apply(const std::shared_ptr<ov::Model>& model,
     manager.register_pass<ov::intel_gna::pass::InsertCopyBeforeConcatLayer>();
     manager.register_pass<ov::intel_gna::pass::HandleMultiConnectedLayerToConcatAndMemory>();
     manager.register_pass<ov::intel_gna::pass::HandleNonFunctionalSubgraphs>();
+    manager.register_pass<ov::intel_gna::pass::HandleNonFunctionalSubgraphsCleanup>();
+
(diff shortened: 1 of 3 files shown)


src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
```diff
@@ -20,6 +20,9 @@ using namespace ov::intel_gna::pass;
 using namespace ov::intel_gna::graph_utils;
 using namespace ov::opset9;
 
+static const std::string kNonCompProperty("non_compute_node");
+static const std::string kResultProprty("result_vector");
```
sirzabek, line 24:
```suggestion
static const std::string kResultProperty("result_vector");
```

sirzabek (changes requested):
Looks more like workaround than a fix.

sirzabek (approved):

mblaszk (approved):

src/plugins/intel_gna/src/gna_transformations_pipeline.cpp
```diff
@@ -192,6 +192,8 @@ void TransformationsPipeline::apply(const std::shared_ptr<ov::Model>& model,
     manager.register_pass<ov::intel_gna::pass::InsertCopyBeforeConcatLayer>();
     manager.register_pass<ov::intel_gna::pass::HandleMultiConnectedLayerToConcatAndMemory>();
     manager.register_pass<ov::intel_gna::pass::HandleNonFunctionalSubgraphs>();
+    manager.register_pass<ov::intel_gna::pass::HandleNonFunctionalSubgraphsCleanup>();
```
mryzhov, line 195: it is better to hide cleanup transformations from the plugin transformations pipeline. Why it can not be added to transformation pipeline?

PR #19274, opened 2023-08-18
[GNA] Fix memory leak in insert_copy_layer

marcinkusm:
### Details:
- enabled transformation for insert_copu_layer.cpp
- added new cleanup transformation

### Tickets:
 - *116912*

### DIFF
--- a/src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
+++ b/src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
@@ -20,6 +20,9 @@ using namespace ov::intel_gna::pass;
 using namespace ov::intel_gna::ngraph_util;
 using namespace ov::opset9;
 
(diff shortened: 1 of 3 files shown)


src/plugins/intel_gna/src/transformations/insert_copy_layer.cpp
```diff
@@ -20,6 +20,9 @@ using namespace ov::intel_gna::pass;
 using namespace ov::intel_gna::ngraph_util;
 using namespace ov::opset9;
 
+static const std::string kNonCompProperty("non_compute_node");
+static const std::string kResultProprty("result_vector");
```
sirzabek, line 24:
```suggestion
static const std::string kResultProperty("result_vector");
```

sirzabek (approved):
It looks more like a workaround than a fix.

PR #19276, opened 2023-08-18
Torchfx compile bugfix

cavusmustafa:
- BugFix in TorchFX
- Missing core object in get_device()


slyalin (changes requested):
Take your changes only.

PR #19279, opened 2023-08-18
[CPU] Fix convolution plus sum layout alignment

maxnick:
### Details:
Since in the case of fusing sum into convolution the output memory is used as the accumulator, the DNNL executor have to handle this fact properly. This PR introduces the convolution executor specialization for this case.

### Tickets:
 - 117937

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/conv.cpp
+++ b/src/plugins/intel_cpu/src/nodes/conv.cpp
@@ -1362,6 +1362,16 @@ void Convolution::prepareParams() {
         if (!reorderConvDesc)
             return nullptr;
 
+        if (key.attr.get()->post_ops_.count(dnnl::impl::primitive_kind::sum)) {
+            return std::make_shared<ConvolutionSumExecutor>(
+                reorderConvDesc,
+                key.inp0->getDnnlDesc(),
+                key.inp1->getDnnlDesc(),
+                key.out->getDnnlDesc(),
+                engine,
+                key.constWeight);
+        }
+
         return std::make_shared<ConvolutionExecutor>(
             reorderConvDesc,
             key.inp0->getDnnlDesc(),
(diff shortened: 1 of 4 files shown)


maxnick:
@dmitry-gorokhov, could you please review?

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1440,6 +1450,42 @@ Convolution::ConvolutionExecutor::ConvolutionExecutor(const dnnl::primitive_desc
     }
 }
 
+Convolution::ConvolutionSumExecutor::ConvolutionSumExecutor(const dnnl::primitive_desc& pd,
```
dmitry-gorokhov, line 1453: Looks like the only difference with common `ConvolutionExecutor` is additional `inputReorder` for dst memory. Anyway takes some time to get it from the code. I would ask you to add some description for the executor.

src/plugins/intel_cpu/src/nodes/conv.cpp
```diff
@@ -1440,6 +1450,42 @@ Convolution::ConvolutionExecutor::ConvolutionExecutor(const dnnl::primitive_desc
...
+        if (primArgs.count(inReorder.first)) {
+            dnnl::memory memDst(inReorder.second.getDstDesc(), strm.get_engine());
+            inReorder.second.exec(primArgs[inReorder.first], memDst, strm);
+            primArgs[inReorder.first] = memDst;
+        } else {
+            IE_THROW() << "DnnlExecutor has reorder for input " << inReorder.first << ", but doesn't have source memory";
+        }
+    }
+    execPrim.execute(strm, primArgs);
+    outputReorders.at(DNNL_ARG_DST).exec(primArgs.at(DNNL_ARG_DST), outputMem, strm);
```
dmitry-gorokhov, line 1486: Is there a guarantee that outputReorders is not empty in any case?

src/plugins/intel_cpu/src/nodes/conv.cpp
maxnick: Done

src/plugins/intel_cpu/src/nodes/conv.cpp
maxnick: Good catch! In general case the `outputReorders` map may be empty. A dedicated check has been added.

dmitry-gorokhov (approved):

PR #19302, opened 2023-08-21
[GPU] update the data-type of primitive from ops

sungeunk:
From vladimir, "That will require manual handling of ops that are decomposed to multiple primitives to ensure that output types are correct for intermediate tensors. And then calc_output_layout(s) impl can be changed to always set this optional_data_type value as output precision instead of propagation of input data type."

### problem:
The problem node has different data-type of output_layout and memory_impl because it's miss-match of data-type between program_node and primitive.

At remove_redundant_reorders pass, the reorder that converts from f32 to f16 is fused to the concat.
It updates dt of node to f16 only. No update primitive dt(f32).
-  Before,        const_data -> ... -> concat(prim:f32, node:f32) -> reorder(prim:f16, node:f16) -> ... -> permuteA1(prim:f32, node:f16)
-  After fusing,  const_data -> ... -> concat(prim:f32, node:f16) -> ... -> permuteA1(prim:f32, node:f16)

At propagate_constants pass, it creates some sub-graphs to calculate the propagated nodes.
In sub_graph, the data-type of concat will be f32 because the data-type of primitive is f32.
The finally created data has the f32 data-type, and when replaced with the main-graph node,
output_data_type is f16 and memory_impl is f32, and miss-matching occurs.
-  sub-graph      const_data -> ... -> concat(prim:f32, node:f32) -> ... -> permuteA2(prim:f32, node:f32)
-  main-graph     REPLACE [ const_data -> ... -> concat(prim:f32, node:f16) -> ... -> permuteA1(prim:f32, node:f16) ] to [ permuteA2(prim:f32, node:f16) ]

### Tickets:
 - 117387

### DIFF
--- a/src/plugins/intel_gpu/src/graph/tile.cpp
+++ b/src/plugins/intel_gpu/src/graph/tile.cpp
@@ -17,6 +17,8 @@ namespace cldnn {
 GPU_DEFINE_PRIMITIVE_TYPE_ID(tile)
 
 layout tile_inst::calc_output_layout(tile_node const& node, kernel_impl_params const& impl_param) {
+    assert(static_cast<bool>(impl_param.desc->output_data_types[0]) == false &&
+           "Output data type forcing is not supported for tile_node!");
     auto desc = impl_param.typed_desc<tile>();
 
     auto input_layout = impl_param.get_input_layout();
--- a/src/plugins/intel_gpu/src/graph/tile.cpp
+++ b/src/plugins/intel_gpu/src/graph/tile.cpp
@@ -17,8 +17,6 @@ namespace cldnn {
 GPU_DEFINE_PRIMITIVE_TYPE_ID(tile)
 
 layout tile_inst::calc_output_layout(tile_node const& node, kernel_impl_params const& impl_param) {
-    assert(static_cast<bool>(impl_param.desc->output_data_types[0]) == false &&
-           "Output data type forcing is not supported for tile_node!");
     auto desc = impl_param.typed_desc<tile>();
 
     auto input_layout = impl_param.get_input_layout();
--- a/src/plugins/intel_gpu/src/plugin/program_builder.cpp
+++ b/src/plugins/intel_gpu/src/plugin/program_builder.cpp
@@ -523,6 +523,9 @@ void ProgramBuilder::add_primitive(const ov::Node& op, std::shared_ptr<cldnn::pr
 
     prim->origin_op_name = op.get_friendly_name();
     prim->origin_op_type_name = op.get_type_name();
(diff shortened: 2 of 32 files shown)


src/plugins/intel_gpu/src/graph/tile.cpp
```diff
@@ -17,8 +17,6 @@ namespace cldnn {
 GPU_DEFINE_PRIMITIVE_TYPE_ID(tile)
 
 layout tile_inst::calc_output_layout(tile_node const& node, kernel_impl_params const& impl_param) {
-    assert(static_cast<bool>(impl_param.desc->output_data_types[0]) == false &&
-           "Output data type forcing is not supported for tile_node!");
```
vladimir-paramuzov, line 21: [Random spot] Have you considered overriding output layout data type somewhere here https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_gpu/src/graph/include/primitive_type_base.h#L119 in case if it's available? Likely it will also help to find the issues with incorrect type passed from Op converters mentioned in the other comment, as looks like currently most of calc_output_layout(s) impls just ignore those optional data types.

src/plugins/intel_gpu/src/plugin/program_builder.cpp
```diff
@@ -523,6 +523,9 @@ void ProgramBuilder::add_primitive(const ov::Node& op, std::shared_ptr<cldnn::pr
 
     prim->origin_op_name = op.get_friendly_name();
     prim->origin_op_type_name = op.get_type_name();
+    size_t output_size = std::min(prim->output_data_types.size(), op.get_output_size());
+    for (size_t i = 0; i < output_size; ++i)
+        prim->output_data_types[i] = cldnn::element_type_to_data_type(op.get_output_element_type(i));
```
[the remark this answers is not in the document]
sungeunk: How about does it set the output data-type before calling add_primitive?
There are some cases to set the output dt as src/plugins/intel_gpu/src/plugin/ops/non_max_suppression.cpp.

In program_builder, it's hard to set the dt as a common logic. The logic to set the dt should be implemented when it's needed.

src/plugins/intel_gpu/src/graph/tile.cpp
sungeunk: I've modified ops/transpose to set the output data-type. It needs to update graph/permute only.
Common logic to check the output dt seems unnecessary.

isanghao:
LGTM. Let me check performance and let's have a review from another person.

vladimir-paramuzov (approved):

isanghao (approved):

PR #19312, opened 2023-08-21
[CPU] Number of streams calculation heuristic moved after LPT

v-Golubev:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*

PR #19313, opened 2023-08-21
[CPU] Number of streams calculation heuristic moved after LPT

v-Golubev:
### Details:
Earlier, streams calculation logic has been moved before all transformations. This led to performance degradations in `hint=tput` mode on some INT8 models.
In this PR, number of streams calculation heuristic moved after LPT in plugin pipeline. This allows to take into account the precision of the weights when heuristic is applied.

### Tickets:
 - *CVS-117971*

PR to release branch: #19312

### DIFF
--- a/src/plugins/intel_cpu/src/plugin.cpp
+++ b/src/plugins/intel_cpu/src/plugin.cpp
@@ -499,6 +499,9 @@ Engine::LoadExeNetworkImpl(const InferenceEngine::CNNNetwork &network, const std
 
     DEBUG_LOG(PrintableModel(*nGraphFunc, "org_"));
 
+    Transformations transformations(nGraphFunc, enableLPT, inferencePrecision, isLegacyAPI(), snippetsMode, engConfig);
+    transformations.UpToLpt();
+
     if (!is_cpu_map_available()) {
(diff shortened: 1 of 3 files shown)


src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -499,6 +499,9 @@ Engine::LoadExeNetworkImpl(const InferenceEngine::CNNNetwork &network, const std
 
     DEBUG_LOG(PrintableModel(*nGraphFunc, "org_"));
 
+    Transformations transformations(nGraphFunc, enableLPT, inferencePrecision, isLegacyAPI(), snippetsMode, engConfig);
```
dmitry-gorokhov, line 502: So now we create transformation pipeline with engine config instead of merged engine + compilation config. This is ok since it doesn't affect transformations behavior, but in fact logically incorrect. We will need to refactor this in follow-up PRs. 
The idea is to merge engine and compilation configs via `readProperties` call and use the merged one everywhere during compilation. It will also allow to get rid of `getInferencePrecision`, `getSnippetsMode` and other similar functions.

dmitry-gorokhov (approved):

PR #19322, opened 2023-08-22
[CPU] optimize initEdge to avoid addtional convert/reorder nodes

riverlijunjie:
### Details:
 - optimize initEdge to avoid additional convert/reorder nodes
 - Avoid introduce unsupported precision(fp64) to reorder node in case of  ov::pass::ConvertPrecision with convert_input_output_precision=false

### Tickets:
 - 118020 115823

### DIFF
--- a/src/plugins/intel_cpu/src/dnnl_extension_utils.cpp
+++ b/src/plugins/intel_cpu/src/dnnl_extension_utils.cpp
@@ -90,6 +90,32 @@ InferenceEngine::Precision DnnlExtensionUtils::DataTypeToIEPrecision(memory::dat
     }
 }
 
+bool DnnlExtensionUtils::isSupportedPrecision(const InferenceEngine::Precision& prec) {
+    switch (prec) {
+        case InferenceEngine::Precision::FP32:
+        case InferenceEngine::Precision::I32:
+        case InferenceEngine::Precision::I8:
+        case InferenceEngine::Precision::U8:
+        case InferenceEngine::Precision::BOOL:
+        case InferenceEngine::Precision::BIN:
+        case InferenceEngine::Precision::BF16:
+            return true;
+        case InferenceEngine::Precision::FP16:
+#if defined(OPENVINO_ARCH_X86) || defined(OPENVINO_ARCH_X86_64)
+            return true;
+#elif defined(OPENVINO_ARCH_ARM64) || defined(OPENVINO_ARCH_ARM)
+#    if defined(OV_CPU_ARM_ENABLE_FP16)
+            return true;
+#    else
+            return false;
+#    endif
+#endif
+        default:
+            return false;
+    }
+    return false;
+}
+
 Dim DnnlExtensionUtils::convertToDim(const dnnl::memory::dim &dim) {
     return dim == DNNL_RUNTIME_DIM_VAL ?  Shape::UNDEFINED_DIM : static_cast<size_t>(dim);
 }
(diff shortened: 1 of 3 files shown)


riverlijunjie:
> I have a few questions regarding the proposed solution:
> 
> 1. The root cause of the behavior we're trying to fix with this PR is still unclear. Let's take the first  example
>    image
>    The first reorder is FP32->BF16, but does it change the layout? Moreover, It looks like CumSum reported I32 precision on its input, so why do we have BF16 here? Then, why do we have the second reorder right before the CumSum node? It doesn't change precision. Does it change the layout? The convert node has been either inserted by the logic that covers cases when oneDNN reorder doesn't support some precision conversion combinations or it was already there on the ngraph function level. Which reason is relevant here?
> 2. The second example
>    image
>    The abcd->acdb reorders are rather a bug in the (Convert?) node, there is no clear reason of changing the layout.

@maxnick The above strange behaviors are brought after ov::pass::ConvertPrecision (convert_input_output_precision=false) is applied. We can check the exec_graph and found the first reorder will change layout from abcd to acdb, and the second reorder will change layout from acdb to abcd, it is very strange. While after we apply this PR, this strange behavior will be eliminated.
As we discussed, the root cause comes from ov::pass::ConvertPrecision (convert_input_output_precision=false), why it has added such strange Convert node to make initEdge() insert additional unnecessary reorder nodes.

abcd->acdb reorders also exists in the first example, it should be triggered by the strange Convert nodes during transformation.

maxnick:
> > I have a few questions regarding the proposed solution:
> > 
> > 1. The root cause of the behavior we're trying to fix with this PR is still unclear. Let's take the first  example
> >    image
> >    The first reorder is FP32->BF16, but does it change the layout? Moreover, It looks like CumSum reported I32 precision on its input, so why do we have BF16 here? Then, why do we have the second reorder right before the CumSum node? It doesn't change precision. Does it change the layout? The convert node has been either inserted by the logic that covers cases when oneDNN reorder doesn't support some precision conversion combinations or it was already there on the ngraph function level. Which reason is relevant here?
> > 2. The second example
> >    image
> >    The abcd->acdb reorders are rather a bug in the (Convert?) node, there is no clear reason of changing the layout.
> 
> @maxnick The above strange behaviors are brought after ov::pass::ConvertPrecision (convert_input_output_precision=false) is applied. We can check the exec_graph and found the first reorder will change layout from abcd to acdb, and the second reorder will change layout from acdb to abcd, it is very strange. While after we apply this PR, this strange behavior will be eliminated. As we discussed, the root cause comes from ov::pass::ConvertPrecision (convert_input_output_precision=false), why it has added such strange Convert node to make initEdge() insert additional unnecessary reorder nodes.
> 
> abcd->acdb reorders also exists in the first example, it should be triggered by the strange Convert nodes during transformation.

The root cause of the appearance of the excessive abcd->acdb and acdb->abcd reorders is not in the transformations but rather in the broken plugins logic (I suspect that the convert node reports nspc layout when it shouldn't do it). So this is the first thing we have to resolve.
Second the fp32->fp64 reorder. We have some protecting logic that prevent instancing reorders with unsupported precisions, please check https://github.com/openvinotoolkit/openvino/blob/350c4d2363c54f96d4d2bebb73f16aa56d22364a/src/plugins/intel_cpu/src/graph.cpp#L604-L624. Actually this is the reason of inserting `Ceiling_3511_FP32_FP64`. But this convert doesn't help, since this logic was oriented to unsupported precision conversion combinations, rather than unsupported precisions itself, since the convert precision pass protected the plugin from having unsupported precisions on the CPU graph level. Now the paradigm has been changed. And here we have two solutions, one is to extend the logic, so that it insert converts and reorders in a proper order. Another approach, is to hide all these specifics inside the reorder node. If some of the precisions is not supported, the reorder node must perform the convert implicitly, and only then call oneDNN implementation with proper precisions, so that hiding the reorder specifics behind the node abstraction, thus properly separating the responsibility.
So, first of all, you'd have never encountered this problem with several at first glance unnecessary reorder/convert nodes, if the plugin logic hadn't inserted these unnecessary abcd<->acdb reorders. But, we still have to support reorders with precisions, that are not supported by oneDNN.

PR #19327, opened 2023-08-22
[CPU][Compile model] Optimize Compile Model

xipingyan:
### Details:
 - *Change CPU plugin denormals default config to {"CPU_DENORMALS_OPTIMIZATION":"YES"}*
 - *Condition 1: it only work for THREADING: TBB/TBB_AUTO*
 - *Condition 2: if customer set this config, the customer's settings will be adopted*
 - TODO: update OV doc description: CVS-121292

### Tickets:
 - *CVS-110424*


dmitry-gorokhov:
@xipingyan The idea itself looks good! Given each thread has its own registers state, that should allow avoid impact on user code.
However we need to understand that enabling ftz flags impacts not only weights of the model (as it works with current solution), but activations as well. In general it may lead to more significant accuracy impact. 
So in order to make it work by default we need to do full-blown accuracy tests and evaluate the impact. Ideally that should be done both for x86 and ARM.

xipingyan:
> @xipingyan The idea itself looks good! Given each thread has its own registers state, that should allow avoid impact on user code. However we need to understand that enabling ftz flags impacts not only weights of the model (as it works with current solution), but activations as well. In general it may lead to more significant accuracy impact. So in order to make it work by default we need to do full-blown accuracy tests and evaluate the impact. Ideally that should be done both for x86 and ARM.

Thanks @dmitry-gorokhov , Besides the problem you mentioned, I am not sure if it works in tbb thread. So I still need a deeper analysis.

xipingyan:
For tbb, I just tested: I can get same value based `on _mm_getcsr` in tbb threads. @dmitry-gorokhov .
So the last problem is accuary verification.

yuxu42:
@usstq Could you please take a review? Thanks!

xipingyan:
Add accuracy result: (http://10.67.108.202:8080/accuracy/219/215)
Here is fp32 top 10:
<html>
<body>


Model | Framework | Precision | result | metric | target value | ref value | D-value(tar-ref) | standard value | D-value(tar-standard)
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
wavernn-upsampler | onnx | FP32 | PASSED | mae@upsample_mels@mean | 0.000000000000000 | 0.000000000000000 | 0 | 0.000000 | 0
yolo_v4 | tf | FP32 | PASSED | map | 0.711723437380425 | 0.711723437380425 | 0 | 0.711700 | 2.3437380424984E-5
i3d-flow | tf | FP32 | PASSED | accuracy@top1 | 0.346666666666667 | 0.346666666666667 | 0 | 0.346667 | -3.3333333299135E-7
i3d-flow | tf | FP32 | PASSED | accuracy@top5 | 0.570000000000000 | 0.570000000000000 | 0 | 0.570000 | 0
dpn-68 | onnx | FP32 | PASSED | accuracy@top1 | 0.758900000000000 | 0.758900000000000 | 0 | 0.758940 | -3.9999999999929E-5
dpn-68 | onnx | FP32 | PASSED | accuracy@top5 | 0.927000000000000 | 0.927000000000000 | 0 | 0.927080 | -7.9999999999969E-5
efficientnet-b0 | tf | FP32 | PASSED | accuracy@top1 | 0.756940000000000 | 0.756940000000000 | 0 | 0.757000 | -6.000000000006E-5
efficientnet-b0 | tf | FP32 | PASSED | accuracy@top5 | 0.927620000000000 | 0.927620000000000 | 0 | 0.927600 | 2.000000000002E-5
t2t-vit-7 | onnx | FP32 | PASSED | accuracy@top1 | 0.715580000000000 | 0.715580000000000 | 0 | 0.715500 | 7.9999999999969E-5
t2t-vit-7 | onnx | FP32 | PASSED | accuracy@top5 | 0.908280000000000 | 0.908280000000000 | 0 | 0.908300 | -2.000000000002E-5
densenet-121 | tf | FP32 | PASSED | accuracy@top1 | 0.730940000000000 | 0.730940000000000 | 0 | 0.730900 | 4.000000000004E-5


</body>
</html>

dmitry-gorokhov:
@xipingyan What definetelly needs to be done here is some functional test, that creates simple subgraph with some math (like multiplication operation) which will be launched both via OV CPU and reference template plugin (usual func test behavior). After that we need to check that denormals are flushed to zero in OV CPU case and kept untouched for ngraph reference case. Such test will prove that OV CPU configuration doesn't affect user code.

src/plugins/intel_cpu/src/exec_network.cpp
```diff
@@ -126,6 +128,18 @@ ExecNetwork::ExecNetwork(const InferenceEngine::CNNNetwork &network,
         do {
             for (auto&& task : tasks) {
                 task = [this] {
+                    // SSE runtime check is needed for some ATOM machine, which is x86-64 but w/o SSE
```
dmitry-gorokhov, line 131: 1. I would say it is better to wrap this code block into `ExecNetwork::GetGraph();` itself.
2. We need to make sure StreamExecutor will execute the task not in main thread. From my understanding this is not guaranteed in general (OMP, SEQ modes?). @wangleis CAn you help us find out most reliable way to unsure separate thread will be created for the task execution?

xipingyan: For question 2: **OMP mode**, I just verified: OMP and TBB are same, `ExecNetwork::GetGraph()` is executed in different thread. @wangleis , @dmitry-gorokhov

src/plugins/intel_cpu/src/exec_network.cpp
```diff
@@ -126,6 +128,18 @@ ExecNetwork::ExecNetwork(const InferenceEngine::CNNNetwork &network,
...
+                    static Xbyak::util::Cpu cpu;
+                    if (cpu.has(Xbyak::util::Cpu::tSSE)) {
+                        if (this->_cfg.denormalsOptMode == Config::DenormalsOptMode::DO_On) {
```
wangleis, line 134: @xipingyan This function is used in scenarios where the application and CPU inference are run in different threads. So SEQ should not support it. 

Then Please add different default setting of _cfg.denormalsOptMode for TBB/OMP based on macro definition `OV_THREAD_TBB`, `OV_THREAD_OMP` and `OV_THREAD_TBB_AUTO`. 

And please add test case in ov_cpu_unit_tests to cover this logic.

src/plugins/intel_cpu/src/exec_network.cpp
xipingyan: For question 1: I can't wrap them into `GetGraph`, because `GetGraph` is const function, and we need to update value "`_cfg.DAZOn`".
If we wrap them into `GetGraph`, we will get **error**:
```
openvino/src/plugins/intel_cpu/src/exec_network.cpp: In member function ‘ov::intel_cpu::ExecNetwork::GraphGuard::Lock ov::intel_cpu::ExecNetwork::GetGraph() const’:
openvino/src/plugins/intel_cpu/src/exec_network.cpp:170:30: error: assignment of member ‘ov::intel_cpu::Config::DAZOn’ in read-only object
  170 |             this->_cfg.DAZOn = denormals_as_zero(true);
```

src/plugins/intel_cpu/src/exec_network.cpp
xipingyan: hi @wangleis @dmitry-gorokhov ,
When I added unit tests for it, I found a new problem.
If the customer calls `start_async`, current solution works fine, but if the customer calls` infer`, it will not work because inference runs in the customer's main thread when `infer` is called.

wangleis: @xipingyan The behavior you mentioned is right. 

If this function need to cover `infer()`, I propose to keep the function in main threads as well. And just skip the function in main threads and enable it in sub threads by default. Then `start_async()` will enable this function by default and `infer()` can enable this function by user setting. 

Please document this different behavior between `start_async()` and `infer()`.

xipingyan: Updated and added unit test. Please help review it again.

src/plugins/intel_cpu/tests/unit/denormals_optimize_test.cpp
```diff
@@ -0,0 +1,191 @@
...
+        if (pConst3)
+            delete[] pConst3;
+
+        // Recover the default setting to avoid impacting other tests.
+        ov::intel_cpu::flush_to_zero(false);
+        ov::intel_cpu::denormals_as_zero(false);
+    }
+
+    void SetUp() override {
+        initBuf(pConst1, shape_size(inpShape), std::numeric_limits<float>::min());
```
riverlijunjie, line 41: Is there any reason to apply float::min?

src/inference/dev_api/openvino/runtime/threading/istreams_executor.hpp
```diff
@@ -118,6 +118,8 @@ class OPENVINO_RUNTIME_API IStreamsExecutor : virtual public ITaskExecutor {
         bool _cpu_reservation = false;
         bool _streams_changed = false;
 
+        bool optDenormalsForTBB = true;
```
riverlijunjie, line 121: Suggest use snake_case naming style.

src/plugins/intel_cpu/tests/unit/denormals_optimize_test.cpp
xipingyan: min float * 0.5, we will get less value than min float, so system will use subnormal to express it.
If we set denormal optimization to true, subnormal will be as zero.
Else, system will use subnormal to express this value(less min float).

xipingyan:
Thanks @riverlijunjie for help review.
I have a strange ci fail issue. (Tests on Linux (Ubuntu 20.04, Python 3.11) / CXX_Unit_Tests)
Error is that we can't use CPU device, do you know it? Because this is CPU plugin unit test.

riverlijunjie:
> Thanks @riverlijunjie for help review. I have a strange ci fail issue. (Tests on Linux (Ubuntu 20.04, Python 3.11) / CXX_Unit_Tests) Error is that we can't use CPU device, do you know it? Because this is CPU plugin unit test.

It seems it is mostly unlike unit tests because the tests will call pubic OV API to do inference, unit test applies mock core to do tests, maybe CPU device is not registered in some case. Putting the tests into func_tests maybe more reasonable, it can avoid CPU unregistered problem (shared lib build).

src/inference/dev_api/openvino/runtime/threading/istreams_executor.hpp
xipingyan: Done

xipingyan:
> > Thanks @riverlijunjie for help review. I have a strange ci fail issue. (Tests on Linux (Ubuntu 20.04, Python 3.11) / CXX_Unit_Tests) Error is that we can't use CPU device, do you know it? Because this is CPU plugin unit test.
> 
> It seems it is mostly unlikely unit tests because the tests will call pubic OV API to do inference, unit test applies mock core to do tests, maybe CPU device is not registered in some case. Putting the tests into func_tests maybe more reasonable, it can avoid CPU unregistered problem (shared lib build).

Thanks @riverlijunjie , I have moved test to cpu_functional_test, it seems to be more reasonable. And all test passed. 
Please help review it again. @riverlijunjie @wangleis

src/plugins/intel_cpu/tests/unit/denormals_optimize_test.cpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    void TearDown() override {}
+
+    ov::Shape inpShape = ov::Shape{1, 5, 12, 64};
+    ov::element::Type rtPrc = ov::element::f32;
+
+    float* pConst1 = new float[shape_size(inpShape)];
+    float* pConst2 = new float[shape_size(inpShape)];
+    float* pConst3 = new float[shape_size(inpShape)];
```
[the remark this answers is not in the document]
xipingyan: Done

src/plugins/intel_cpu/src/exec_network.cpp
xipingyan: Test had been added cpu functional test.

riverlijunjie (approved):
LGTM

wangleis (approved):

src/inference/dev_api/openvino/runtime/threading/istreams_executor.hpp
```diff
@@ -139,7 +140,8 @@ class OPENVINO_RUNTIME_API IStreamsExecutor : virtual public ITaskExecutor {
                int threads = 0,
                PreferredCoreType threadPreferredCoreType = PreferredCoreType::ANY,
                std::vector<std::vector<int>> streamsInfoTable = {},
-               bool cpuReservation = false)
+               bool cpuReservation = false,
+               bool opt_denormals_for_tbb = true)
```
xipingyan, line 144: I have to add this param in this constructor in order to fix CI issue.
`lto1: internal compiler error: in add_symbol_to_partition_1, at lto/lto-partition.c:153`

yuxu42:
Hi @dmitry-gorokhov Could you please take a look? 2023.2 has branch out.

wenjiew:
@xipingyan Are the test cases fully added? Thanks!

xipingyan:
> @xipingyan Are the test cases fully added? Thanks!

@wenjiew Yes, @chenhu-wang also added a test about data movement. https://github.com/openvinotoolkit/openvino/pull/21033

wenjiew:
Let us plan this for 2024.1 release. Thanks!

PR #19336, opened 2023-08-22
[packaging] APT/YUM: Added few conflict version for dot-releases

artanokhov:


ilya-lavrenov (approved):
Need to port to master

PR #19342, opened 2023-08-22
Avoid creating new threads when converting legacy inference request to API 2.0

zhaixuejun1993:
### Details:
 - Fix 2 new threads created in 1.0 req convert to 2.0 req issue

### Tickets:
 - 115961

### DIFF
--- a/src/inference/src/dev/icompiled_model_wrapper.cpp
+++ b/src/inference/src/dev/icompiled_model_wrapper.cpp
@@ -10,7 +10,7 @@
 
 InferenceEngine::ICompiledModelWrapper::ICompiledModelWrapper(
     const std::shared_ptr<InferenceEngine::IExecutableNetworkInternal>& model)
-    : ov::ICompiledModel(nullptr, ov::legacy_convert::convert_plugin(model->_plugin)),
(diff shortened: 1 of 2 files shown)


src/inference/src/dev/icompiled_model_wrapper.cpp
```diff
@@ -10,7 +10,7 @@
 
 InferenceEngine::ICompiledModelWrapper::ICompiledModelWrapper(
     const std::shared_ptr<InferenceEngine::IExecutableNetworkInternal>& model)
-    : ov::ICompiledModel(nullptr, ov::legacy_convert::convert_plugin(model->_plugin)),
+    : ov::ICompiledModel(nullptr, ov::legacy_convert::convert_plugin(model->_plugin), nullptr, nullptr),
```
songbell, line 13: why?

songbell: this is not needed in wrapper right? won't they be released when compiled model destructed?

ilyachur: Yes we don't need to create these task and callback executors, because new API just wraps old objects

ilyachur (approved):

PR #19351, opened 2023-08-23
[GPU] Fix accuracy issue

ahnyoung-paul:
### Details:
 - *[scatter_update] Use input index for input buffer instead of output index*
 - *[concat cpu impl] Sync input layout and mem_ptr when input host tensor creation*
 - *Add unit tests for scatter_update and concat cpu impl*

### Tickets:
 - *100495*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/cpu/concat.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/cpu/concat.cpp
@@ -69,16 +69,17 @@ struct concatenation_impl : public typed_primitive_impl<concatenation> {
         std::vector<memory::ptr> input_mem_ptrs;
         for (size_t i = 0; i < instance.dependencies().size(); i++) {
             auto& dep = instance.dependencies().at(i);
-            if (dep.first->get_output_layout().count() > 0)
-                input_mem_ptrs.push_back(instance.dep_memory_ptr(i));
+            if (dep.first->get_output_layout().count() > 0) {
+                auto mem_ptr = instance.dep_memory_ptr(i);
+                input_host_tensors.push_back(make_tensor(params->input_layouts[i], mem_ptr->lock(stream, mem_lock_type::read)));
+                // push mem_ptr to input_mem_ptr to unlock after processing
+                input_mem_ptrs.push_back(mem_ptr);
+            }
         }
 
         auto output_mem_ptr = instance.output_memory_ptr();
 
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/src/graph/impls/cpu/concat.cpp
```diff
@@ -69,17 +69,18 @@ struct concatenation_impl : public typed_primitive_impl<concatenation> {
...
         for (size_t i = 0; i < instance.dependencies().size(); i++) {
             auto& dep = instance.dependencies().at(i);
-            if (dep.first->get_output_layout().count() > 0)
-                input_mem_ptrs.push_back(instance.dep_memory_ptr(i));
+            if (dep.first->get_output_layout().count() > 0) {
+                auto mem_ptr = instance.dep_memory_ptr(i);
+                input_host_tensors.push_back(make_tensor(params->input_layouts[i], mem_ptr->lock(stream, mem_lock_type::read)));
+                // push mem_ptr to input_mem_ptr to unlock after processing
+                input_mem_ptrs.push_back(mem_ptr);
+            }
```
e-ddykim, line 77: I don't understand what is changed in this part.

ahnyoung-paul: When concat has 5 inputs and 3rd input are empty layout, previous code create input_host_tensor with 3rd input layout 4th input memory pointer. I fixed this issue.

yeonbok: What I think is that 
when one of the  dep nodes is  empty tensor
then it was not added to input_mem_ptrs. 
e.g.,there are three inputs dep0, dep1, dep2 and dep1 is empty, then input_mem_ptrs has only dep0 and dep2. 
However in the original impl
line 80 is iterating for those two and making host tensor from input_layouts(0) and input_layouts(1) , 
not input_layouts(0) and input_laytous(2)

ahnyoung-paul: Yes, @yeonbok is right. thanks.

yeonbok (approved):

src/plugins/intel_gpu/src/graph/impls/cpu/concat.cpp
e-ddykim: Oh I see, thank you.

e-ddykim (approved):

PR #19353, opened 2023-08-23
Moved type prop tests to new API from g to z

ilyachur:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


src/core/tests/type_prop/irdft.cpp
```diff
@@ -14,11 +14,14 @@
 // limitations under the License.
```
vurusovs, line 14: Out of PR scope - should the license be replaced with shorter version?

vurusovs (approved):

PR #19354, opened 2023-08-23
[GPU] Memory reuse false for dynamic and null impl case

kelvinchoi-intel:
### Details:
 - *Memory reuse false for dynamic and null impl case*

### Tickets:
 - *116385*

### DIFF
--- a/src/plugins/intel_gpu/tests/unit/test_cases/memory_test.cpp
+++ b/src/plugins/intel_gpu/tests/unit/test_cases/memory_test.cpp
@@ -14,6 +14,7 @@
 #include <intel_gpu/primitives/reshape.hpp>
 #include <intel_gpu/primitives/crop.hpp>
 #include <intel_gpu/primitives/eltwise.hpp>
+#include <intel_gpu/primitives/grid_sample.hpp>
 #include <fully_connected_inst.h>
 
 using namespace cldnn;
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/tests/unit/test_cases/memory_test.cpp
```diff
@@ -595,6 +596,58 @@ class memory_pool: public ::testing::Test {
...
+
+        auto input_actual_layout1 = layout{ ov::PartialShape{ 2, input_f, 1, 2 }, data_types::f32, format::bfyx};
+        auto input_actual_layout2 = layout{ ov::PartialShape{ 1, input_f, 1, 2 }, data_types::f32, format::bfyx};
+        auto input_data1 = engine.allocate_memory(input_actual_layout1);
+        auto input_data2 = engine.allocate_memory(input_actual_layout2);
+        auto fc_weights_data = engine.allocate_memory({ ov::PartialShape{ 2, input_f, 1, 1 }, data_types::f32, format::bfyx});
+
+        set_values(input_data1, { -0.5f, 2.0f, 0.5f, -0.5f, 2.0f, 0.5f });
+        set_values(input_data2, { 0.5f, -2.0f, -0.5f, -0.5f, 2.0f, 0.5f,
+                                -0.5f, 2.0f, 0.5f, -0.5f, 2.0f, 0.5f });
```
e-ddykim, line 617: The sizes of data and layouts of `input_data1` and `input_data2` are not matched.
Additionally, this test doesn't check the correctness of the results.
Please update this test to check it.

e-ddykim (changes requested):

src/plugins/intel_gpu/tests/unit/test_cases/memory_test.cpp
kelvinchoi-intel: Layout size issue is fixed.
Result value doesn't check because the goal of this test is cheking 'not reuse mem' so it only checks memomy buffer pointer value.

e-ddykim (approved):

PR #19357, opened 2023-08-23
[tests] save into different file in compression_test.py

pavel-esir:
### Details:
 - port https://github.com/openvinotoolkit/openvino/pull/19356 into master
 - sporadic failures were observed when temporary IRs are saved with the same name. Need to save into a file with unique temporary name to fix that.

### Tickets:
 - [CVS-118122](https://jira.devtools.intel.com/browse/CVS-118122)


akuporos (approved):

slyalin:
@andrei-kochin, can we merge it without extra approvals?

PR #19358, opened 2023-08-23
Fix incorrect output for float to bf16 in avx2 isa

riverlijunjie:
### Details:
 - Fix incorrect output for float to bf16 in avx2 isa

### Tickets:
 - 118588, 117486

### DIFF
--- a/src/plugins/intel_cpu/src/emitters/x64/jit_bf16_emitters.hpp
+++ b/src/plugins/intel_cpu/src/emitters/x64/jit_bf16_emitters.hpp
@@ -78,7 +78,6 @@ private:
                 h->vpermq(Ymm(aux.getIdx()), Ymm(aux.getIdx()), 0xD8); //11 01 10 00
                 h->vextracti128(out, Ymm(aux.getIdx()), 0);
             } else {
-                h->pshufd(Xmm(aux.getIdx()), Xmm(aux.getIdx()), 0xD8); //11 01 10 00
                 h->uni_vmovups(out, aux);
             }
         }
--- a/src/plugins/intel_cpu/src/emitters/x64/jit_bf16_emitters.hpp
+++ b/src/plugins/intel_cpu/src/emitters/x64/jit_bf16_emitters.hpp
@@ -75,8 +75,10 @@ private:
             h->uni_vpackusdw(aux, aux, aux);
 
             if (host_isa_ == dnnl::impl::cpu::x64::cpu_isa_t::avx2) {
+                h->vpermq(Ymm(aux.getIdx()), Ymm(aux.getIdx()), 0xD8); //11 01 10 00
                 h->vextracti128(out, Ymm(aux.getIdx()), 0);
             } else {
+                h->pshufd(Xmm(aux.getIdx()), Xmm(aux.getIdx()), 0xD8); //11 01 10 00
                 h->uni_vmovups(out, aux);
             }
         }
--- a/src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
+++ b/src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
@@ -241,6 +241,10 @@ std::vector<std::string> disabledTestPatterns() {
 #endif
 
     if (!InferenceEngine::with_cpu_x86_avx512_core()) {
+        // on platforms which do not support bfloat16, we are disabling bf16 tests since there are no bf16 primitives,
+        // tests are useless on such platforms
+        retVector.emplace_back(R"(.*(BF|bf)16.*)");
+        retVector.emplace_back(R"(.*bfloat16.*)");
         // MatMul in Snippets uses BRGEMM that is supported only on AVX512 platforms
         // Disabled Snippets MHA tests as well because MHA pattern contains MatMul
         retVector.emplace_back(R"(.*Snippets.*MHA.*)");
--- a/src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
+++ b/src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
@@ -241,10 +241,6 @@ std::vector<std::string> disabledTestPatterns() {
 #endif
 
     if (!InferenceEngine::with_cpu_x86_avx512_core()) {
-        // on platforms which do not support bfloat16, we are disabling bf16 tests since there are no bf16 primitives,
-        // tests are useless on such platforms
-        retVector.emplace_back(R"(.*(BF|bf)16.*)");
-        retVector.emplace_back(R"(.*bfloat16.*)");
         // MatMul in Snippets uses BRGEMM that is supported only on AVX512 platforms
         // Disabled Snippets MHA tests as well because MHA pattern contains MatMul
         retVector.emplace_back(R"(.*Snippets.*MHA.*)");
--- a/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
+++ b/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
@@ -272,20 +272,19 @@ struct jit_variable_load_store_test_kernel {
         size_t size;
     };
 
-    template<size_t N, bool is_src>
+    template<size_t N, size_t M, bool is_src>
     void test() {
         kernel_impl<N, is_src> kernel;
         kernel.init();
-
-        const size_t size = 3;
+        ASSERT_GE(N, M);
 
         std::array<SrcT, N> src {};
         std::array<DstT, N> result {};
 
-        Params args = { src.data(), result.data(), size };
+        Params args = { src.data(), result.data(), M };
 
         src.fill(static_cast<SrcT>(42));
-        for (size_t i = 0; i < size; ++i) {
+        for (size_t i = 0; i < M; ++i) {
             src[i] = static_cast<SrcT>(i);
         }
 
@@ -293,7 +292,7 @@ struct jit_variable_load_store_test_kernel {
 
         std::array<DstT, N> expected_result {};
 
-        for (size_t i = 0; i < size; ++i) {
+        for (size_t i = 0; i < M; ++i) {
             expected_result[i] = static_cast<DstT>(i);
         }
 
@@ -325,52 +324,52 @@ TEST(JitKernel, variable_load_and_store) {
     {
         jit_variable_load_store_test_kernel<uint8_t, float> kernel;
         if (mayiuse(cpu_isa_t::avx512_core)) {
-            kernel.test<16, false>();
+            kernel.test<16, 11, false>();
         }
         if (mayiuse(cpu_isa_t::avx2)) {
-            kernel.test<8, false>();
+            kernel.test<8, 5, false>();
         }
         if (mayiuse(cpu_isa_t::sse41)) {
-            kernel.test<4, false>();
+            kernel.test<4, 3, false>();
         }
     }
 
     {
         jit_variable_load_store_test_kernel<int8_t, int8_t> kernel;
         if (mayiuse(cpu_isa_t::avx512_core)) {
-            kernel.test<16, false>();
+            kernel.test<16, 11, false>();
         }
         if (mayiuse(cpu_isa_t::avx2)) {
-            kernel.test<8, false>();
+            kernel.test<8, 5, false>();
         }
         if (mayiuse(cpu_isa_t::sse41)) {
-            kernel.test<4, false>();
+            kernel.test<4, 3, false>();
         }
     }
 
     {
         jit_variable_load_store_test_kernel<float, bfloat16_t> kernel;
         if (mayiuse(cpu_isa_t::avx512_core)) {
-            kernel.test<16, true>();
+            kernel.test<16, 11, true>();
         }
         if (mayiuse(cpu_isa_t::avx2)) {
-            kernel.test<8, true>();
+            kernel.test<8, 5, true>();
         }
         if (mayiuse(cpu_isa_t::sse41)) {
-            kernel.test<4, true>();
+            kernel.test<4, 3, true>();
         }
     }
 
     {
         jit_variable_load_store_test_kernel<int32_t, bfloat16_t> kernel;
         if (mayiuse(cpu_isa_t::avx512_core)) {
-            kernel.test<16, true>();
+            kernel.test<16, 11, true>();
         }
         if (mayiuse(cpu_isa_t::avx2)) {
-            kernel.test<8, true>();
+            kernel.test<8, 5, true>();
         }
         if (mayiuse(cpu_isa_t::sse41)) {
-            kernel.test<4, true>();
+            kernel.test<4, 3, true>();
         }
     }
 }


src/plugins/intel_cpu/src/emitters/x64/jit_bf16_emitters.hpp
```diff
@@ -75,8 +75,10 @@ class jit_uni_vcvtneps2bf16 : public jit_emitter {
             h->uni_vpackusdw(aux, aux, aux);
 
             if (host_isa_ == dnnl::impl::cpu::x64::cpu_isa_t::avx2) {
+                h->vpermq(Ymm(aux.getIdx()), Ymm(aux.getIdx()), 0xD8); //11 01 10 00
                 h->vextracti128(out, Ymm(aux.getIdx()), 0);
             } else {
+                h->pshufd(Xmm(aux.getIdx()), Xmm(aux.getIdx()), 0xD8); //11 01 10 00
```
chenhu-wang, line 81: Is this line really needed? as we are fixing cross lane alignment.

chenhu-wang:
Is it possible to add test for this failure?

src/plugins/intel_cpu/src/emitters/x64/jit_bf16_emitters.hpp
riverlijunjie: SSE4.1 should also have this problem, but float->bf16 tests are disabled in non-avx512.

riverlijunjie:
> Is it possible to add test for this failure?

In non-avx512 ISA, all bf16 C++ tests are disabled: https://github.com/openvinotoolkit/openvino/blob/f306007e59bc4573414408253aeb4e0d19f4bc67/src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp#L243-L247

But there are several python test cases touch it on non-AVX512 ISA platform:
image
image
image
image

Do you think it is necessary to add separated float->bf16 test cases for non-avx512 ISA?

src/plugins/intel_cpu/src/emitters/x64/jit_bf16_emitters.hpp
chenhu-wang: I don't think it's needed on sse41, 4 bf16 data already locate the lower half of XMM at this point, as I mentioned the bug exist cross lane alignment, there is single lane on sse41.

chenhu-wang:
> > Is it possible to add test for this failure?
> 
> In non-avx512 ISA, all bf16 C++ tests are disabled:
> 
> https://github.com/openvinotoolkit/openvino/blob/f306007e59bc4573414408253aeb4e0d19f4bc67/src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp#L243-L247
> 
> But there are several python test cases touch it on non-AVX512 ISA platform: image image image image
> 
> Do you think it is necessary to add separated float->bf16 test cases for non-avx512 ISA?

1. The correctness should be guarantee inside plugin test first. python test may change in the future lead not to cover it.
2. This is live code path that could be executed such in case described in ticket. 
3. If we want support bf16 on avx2 and ss41, we need enable all the test, remove bf16 in skip_test_config. But this should be separate activity.
So at current I think we should add float->bf16 test inside cpu plugin, and then test on avx2 and sse41 with SED locally. Make sure it's passed after enable bf16 test on avx2/sse41.

chenhu-wang:
> 

convert test is necessary as long as this code is used and executed on avx2 in any case.
For generic bf16 on avx2 test enable, it's depend on requirement and arch decision. for example whether bf16 precision could give perf boost on avx2. what's the effort to enable it. The future isa instruction introduced on avx2 etc. If needed, we need update and fix to pass all the test. cc @dmitry-gorokhov

src/plugins/intel_cpu/src/emitters/x64/jit_bf16_emitters.hpp
riverlijunjie: oh, I see, SSE4.1 use Xmm reg, which is 128bit, and the lower 64bit has got 4 bf16 data.

riverlijunjie: I verify SSE4.1 branch with Intel(R) Penryn CPU, with or without this line the float to bf16 all pass, let's delete this line.

riverlijunjie:
The bug only is reproduced in case of ov::pass::ConvertPrecision with convert_input_output_precision=false, which is required by API 2.0 migration, we can see below test failure if enable bf16 test in avx2 ISA:

image

After apply this PR, the test can be passed:

image

chenhu-wang:
> The bug only is reproduced in case of ov::pass::ConvertPrecision with convert_input_output_precision=false, which is required by API 2.0 migration, we can see below test failure if enable bf16 test in avx2 ISA:
> 
> image
> 
> After apply this PR, the test can be passed:
> 
> image

In short, existing test could cover it, but not enabled on CI. local test show this pr could fix it. LGTM now.

chenhu-wang (approved):

maxnick:
I'm agree with @chenhu-wang , we need to cover this change with SL test. Though on AVX2 all the BF16 test are skipped, it doesn't mean we can't adjust the skipping regular expr so that to run some necessary BF16 coversion tests on AVX2.

a-sidorova:
As author of this emulator emitter for non-AVX512 cases, let me tell you about testing.
When I were developing this emulator, I needed to test it for sure. For that I used `jit_kernel_tests` to test [store](https://github.com/openvinotoolkit/openvino/blob/198da893d494bd6565953d63d1211e19530f3e95/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp#L352) emitter with `jit_uni_vcvtneps2bf16` [inside](https://github.com/openvinotoolkit/openvino/blob/198da893d494bd6565953d63d1211e19530f3e95/src/plugins/intel_cpu/src/emitters/x64/jit_load_store_emitters.cpp#L598).
To be honest, I don't know - these tests are executed in CI or not. But the name of test doesn't contain `bf16` or something relevant so in general, these tests should not be skipped.

I guess @riverlijunjie you can try to launch them if needed.


cc @chenhu-wang @maxnick

riverlijunjie:
> As author of this emulator emitter for non-AVX512 cases, let me tell you about testing. When I were developing this emulator, I needed to test it for sure. For that I used `jit_kernel_tests` to test [store](https://github.com/openvinotoolkit/openvino/blob/198da893d494bd6565953d63d1211e19530f3e95/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp#L352) emitter with `jit_uni_vcvtneps2bf16` [inside](https://github.com/openvinotoolkit/openvino/blob/198da893d494bd6565953d63d1211e19530f3e95/src/plugins/intel_cpu/src/emitters/x64/jit_load_store_emitters.cpp#L598). To be honest, I don't know - these tests are executed in CI or not. But the name of test doesn't contain `bf16` or something relevant so in general, these tests should not be skipped.
> 
> I guess @riverlijunjie you can try to launch them if needed.
> 
> cc @chenhu-wang @maxnick

@a-sidorova thank your clarification, I confirm `jit_kernel_tests` does test float->bf16 on AVX platform as below:
https://github.com/openvinotoolkit/openvino/blob/198da893d494bd6565953d63d1211e19530f3e95/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp#L352-L360

But it only test and compare the first 3 data:
https://github.com/openvinotoolkit/openvino/blob/198da893d494bd6565953d63d1211e19530f3e95/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp#L280-L298

We can see that size=3 is applied to set input data and compared expected data, while this issue occurred from 4 data: 0-3 is correct and 4-7 data is wrong and duplicated with 0-3 data.

I have changed the tests as below (N=8):
```
        Params args = { src.data(), result.data(), N };

        src.fill(static_cast<SrcT>(42));
        for (size_t i = 0; i < N; ++i) {
            src[i] = static_cast<SrcT>(i);
        }

        kernel(args);

        std::array<DstT, N> expected_result {};

        for (size_t i = 0; i < N; ++i) {
            expected_result[i] = static_cast<DstT>(i);
        }
```
And the tests will fail:

image

riverlijunjie:
@a-sidorova could you help check below change whether makes sense? This change can cover this PR's test case.

```
diff --git a/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp b/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
index 53576c629b..d077e027b6 100644
--- a/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
+++ b/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
@@ -282,10 +282,10 @@ struct jit_variable_load_store_test_kernel {
         std::array<SrcT, N> src {};
         std::array<DstT, N> result {};
 
-        Params args = { src.data(), result.data(), size };
+        Params args = { src.data(), result.data(), N };
 
         src.fill(static_cast<SrcT>(42));
-        for (size_t i = 0; i < size; ++i) {
+        for (size_t i = 0; i < N; ++i) {
             src[i] = static_cast<SrcT>(i);
         }
 
@@ -293,7 +293,7 @@ struct jit_variable_load_store_test_kernel {
 
         std::array<DstT, N> expected_result {};
 
-        for (size_t i = 0; i < size; ++i) {
+        for (size_t i = 0; i < N; ++i) {
             expected_result[i] = static_cast<DstT>(i);
         }
```

riverlijunjie:
After we applied above jit_kernel_test changes and this PR, test is pass:

image

a-sidorova:
> @a-sidorova could you help check below change whether makes sense? This change can cover this PR's test case.
> 
> ```
> diff --git a/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp b/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
> index 53576c629b..d077e027b6 100644
> --- a/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
> +++ b/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
> @@ -282,10 +282,10 @@ struct jit_variable_load_store_test_kernel {
>          std::array<SrcT, N> src {};
>          std::array<DstT, N> result {};
>  
> -        Params args = { src.data(), result.data(), size };
> +        Params args = { src.data(), result.data(), N };
>  
>          src.fill(static_cast<SrcT>(42));
> -        for (size_t i = 0; i < size; ++i) {
> +        for (size_t i = 0; i < N; ++i) {
>              src[i] = static_cast<SrcT>(i);
>          }
>  
> @@ -293,7 +293,7 @@ struct jit_variable_load_store_test_kernel {
>  
>          std::array<DstT, N> expected_result {};
>  
> -        for (size_t i = 0; i < size; ++i) {
> +        for (size_t i = 0; i < N; ++i) {
>              expected_result[i] = static_cast<DstT>(i);
>          }
> ```

I'd suggest to add just new template parameter `length` to allow developers to test different vector register length (existing parameter `N`) with different value count in it (new parameter `length` - existing local variable `size`) instead of reusing `N` as length of data. Because firstly this test is to validate load and store emitters work. And it's important to have opportunity to test tail registers as well.

riverlijunjie:
Thank @a-sidorova , is it ok of changes is below?

```
diff --git a/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp b/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
index 53576c629b..a851e0b4c5 100644
--- a/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
+++ b/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
@@ -272,20 +272,19 @@ struct jit_variable_load_store_test_kernel {
         size_t size;
     };
 
-    template<size_t N, bool is_src>
+    template<size_t N, size_t M, bool is_src>
     void test() {
         kernel_impl<N, is_src> kernel;
         kernel.init();
-
-        const size_t size = 3;
+        ASSERT_GE(N, M);
 
         std::array<SrcT, N> src {};
         std::array<DstT, N> result {};
 
-        Params args = { src.data(), result.data(), size };
+        Params args = { src.data(), result.data(), M };
 
         src.fill(static_cast<SrcT>(42));
-        for (size_t i = 0; i < size; ++i) {
+        for (size_t i = 0; i < M; ++i) {
             src[i] = static_cast<SrcT>(i);
         }
 
@@ -293,7 +292,7 @@ struct jit_variable_load_store_test_kernel {
 
         std::array<DstT, N> expected_result {};
 
-        for (size_t i = 0; i < size; ++i) {
+        for (size_t i = 0; i < M; ++i) {
             expected_result[i] = static_cast<DstT>(i);
         }
 
@@ -325,52 +324,52 @@ TEST(JitKernel, variable_load_and_store) {
     {
         jit_variable_load_store_test_kernel<uint8_t, float> kernel;
         if (mayiuse(cpu_isa_t::avx512_core)) {
-            kernel.test<16, false>();
+            kernel.test<16, 11, false>();
         }
         if (mayiuse(cpu_isa_t::avx2)) {
-            kernel.test<8, false>();
+            kernel.test<8, 5, false>();
         }
```

a-sidorova:
> Thank @a-sidorova , is it ok of changes is below?
> 
> ```
> diff --git a/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp b/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
> index 53576c629b..a851e0b4c5 100644
> --- a/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
> +++ b/src/plugins/intel_cpu/tests/unit/jit_kernel_test.cpp
> @@ -272,20 +272,19 @@ struct jit_variable_load_store_test_kernel {
>          size_t size;
>      };
>  
> -    template<size_t N, bool is_src>
> +    template<size_t N, size_t M, bool is_src>
>      void test() {
>          kernel_impl<N, is_src> kernel;
>          kernel.init();
> -
> -        const size_t size = 3;
> +        ASSERT_GE(N, M);
>  
>          std::array<SrcT, N> src {};
>          std::array<DstT, N> result {};
>  
> -        Params args = { src.data(), result.data(), size };
> +        Params args = { src.data(), result.data(), M };
>  
>          src.fill(static_cast<SrcT>(42));
> -        for (size_t i = 0; i < size; ++i) {
> +        for (size_t i = 0; i < M; ++i) {
>              src[i] = static_cast<SrcT>(i);
>          }
>  
> @@ -293,7 +292,7 @@ struct jit_variable_load_store_test_kernel {
>  
>          std::array<DstT, N> expected_result {};
>  
> -        for (size_t i = 0; i < size; ++i) {
> +        for (size_t i = 0; i < M; ++i) {
>              expected_result[i] = static_cast<DstT>(i);
>          }
>  
> @@ -325,52 +324,52 @@ TEST(JitKernel, variable_load_and_store) {
>      {
>          jit_variable_load_store_test_kernel<uint8_t, float> kernel;
>          if (mayiuse(cpu_isa_t::avx512_core)) {
> -            kernel.test<16, false>();
> +            kernel.test<16, 11, false>();
>          }
>          if (mayiuse(cpu_isa_t::avx2)) {
> -            kernel.test<8, false>();
> +            kernel.test<8, 5, false>();
>          }
> ```

I think it's OK 😃

a-sidorova (approved):

maxnick:
@riverlijunjie , do I correctly understand, that moving to API 2.0 will trigger this code path through the entire OV pipeline with the existing `SetBlobCPU` tests?

riverlijunjie:
> @riverlijunjie , do I correctly understand, that moving to API 2.0 will trigger this code path through the entire OV pipeline with the existing `SetBlobCPU` tests?

It seems current master branch also can trigger this issue if bf16 tests are not skipped in non-avx512 platform.

maxnick:
> > @riverlijunjie , do I correctly understand, that moving to API 2.0 will trigger this code path through the entire OV pipeline with the existing `SetBlobCPU` tests?
> 
> It seems current master branch also can trigger this issue if bf16 tests are not skipped in non-avx512 platform.

By they are skipped (: Should we enable them?

riverlijunjie:
> > > @riverlijunjie , do I correctly understand, that moving to API 2.0 will trigger this code path through the entire OV pipeline with the existing `SetBlobCPU` tests?
> > 
> > 
> > It seems current master branch also can trigger this issue if bf16 tests are not skipped in non-avx512 platform.
> 
> But they are skipped (: Should we enable them?

This PR's issue has been covered by jit_kernel_test, so enable bf16 test on non-avx512 should not be too necessary?

maxnick (approved):

wangleis (approved):

PR #19360, opened 2023-08-23
[CPU] Fixed has_subnormals behavior for negative zero values

dmitry-gorokhov:
### Details:
 - *Context:* CPU plugin has logic that check if constant tensors contain denormal values. In case denormal values are found constant tensor will be copied and denormal values will be zeroed.
 - *Issue:* denormals check logic has incorrectly handled negative zero values (-0.0f) assuming them to be denormal. That has led to significant amount of "false-positive" answers and as a results to redundant memory copies. The last thing extremly affects LLMs in terms of memory consumption and compilation time.
 - The PR fixes denormals check logic behavior

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/input.cpp
+++ b/src/plugins/intel_cpu/src/nodes/input.cpp
@@ -114,7 +114,7 @@ protected:
         uni_vmovdqu(b, a);                   // b = a
         uni_vmovdqu(c, a);                   // c = a
         uni_vpand(b, b, mantissa_mask);      // b = a & 00000000011111111111111111111111
-        uni_vpcmpeqd(b, b, zero);            // if (a == 0) b = 1 else b = 0
+        uni_vpcmpeqd(b, b, zero);            // if (b == 0) b = 1 else b = 0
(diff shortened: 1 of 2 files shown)


dmitry-gorokhov:
@maxnick Could you please take a look?

luweizhou2016:
LGTM. @dmitry-gorokhov  Just one discussion maybe needed here:

Seems @xipingyan  also has another draft PR https://github.com/openvinotoolkit/openvino/pull/19327/commits 
to move CPU denormal config into child threads, which should avoid denormal config effects to customer user app and lower down the compiling time.  so should we approach this PR#19327 as a long term solution? Thx!

dmitry-gorokhov:
> LGTM. @dmitry-gorokhov Just one discussion maybe needed here:
> 
> Seems @xipingyan also has another draft PR https://github.com/openvinotoolkit/openvino/pull/19327/commits to move CPU denormal config into child threads, which should avoid denormal config effects to customer user app and lower down the compiling time. so should we approach this PR#19327 as a long term solution? Thx!

Thanks for heads up! Lets continue discussion in the PR you are referring to.

luweizhou2016 (approved):

PR #19361, opened 2023-08-23
[CPU] Fixed has_subnormals behavior for negative zero values

dmitry-gorokhov:
### Details:
 - *Context:* CPU plugin has logic that check if constant tensors contain denormal values. In case denormal values are found constant tensor will be copied and denormal values will be zeroed.
 - *Issue:* denormals check logic has incorrectly handled negative zero values (-0.0f) assuming them to be denormal. That has led to significant amount of "false-positive" answers and as a results to redundant memory copies. The last thing extremly affects LLMs in terms of memory consumption and compilation time.
 - The PR fixes denormals check logic behavior

Cherry-picks: https://github.com/openvinotoolkit/openvino/pull/19360

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/input.cpp
+++ b/src/plugins/intel_cpu/src/nodes/input.cpp
@@ -92,19 +92,20 @@ protected:
         pop(rsi);
     }
 
-    void check_subnormals(const Xbyak::Reg64& src, const Xbyak::Ymm &mask, const Xbyak::Ymm &zero) {
+    void check_subnormals(const Xbyak::Reg64& src, const Xbyak::Ymm &exponent_mask, const Xbyak::Ymm &mantissa_mask, const Xbyak::Ymm &zero) {
         auto a = ymm1;
         auto b = ymm2;
         auto c = ymm3;
 
         vmovdqu(a, yword[src]);         // load 8 floats
-        vpcmpeqd(b, a, zero);           // if (a == 0) b = 1 else b = 0
-        vpand(c, a, mask);              // c = a & 01111111100000000000000000000000
+        vpand(b, a, mantissa_mask);     // b = a & 00000000011111111111111111111111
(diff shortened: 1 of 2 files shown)


src/plugins/intel_cpu/src/nodes/input.cpp
```diff
@@ -92,30 +92,32 @@ struct jit_has_subnormals_base : public jit_generator {
...
 
         uni_vmovdqu(a, xword[src]);          // load 4 floats
         uni_vmovdqu(b, a);                   // b = a
         uni_vmovdqu(c, a);                   // c = a
-        uni_vpcmpeqd(b, b, zero);               // if (a == 0) b = 1 else b = 0
-        uni_vpand(c, c, mask);                  // c = a & 01111111100000000000000000000000
-        uni_vpcmpeqd(c, c, zero);               // if (c == 0) c = 1 else c = 0
-        uni_vtestps(b, c);                    // if ((!b & c) == 0) CF = 1 else CF = 0
+        uni_vpand(b, b, mantissa_mask);      // b = a & 00000000011111111111111111111111
+        uni_vpcmpeqd(b, b, zero);            // if (a == 0) b = 1 else b = 0
```
luweizhou2016, line 117: About comments. should be "if (b == 0) b = 1 else b = 0" ?

luweizhou2016 (approved):
LGTM. Just minor comments update.

src/plugins/intel_cpu/src/nodes/input.cpp
dmitry-gorokhov: Agree. Better to fix the comment.
Will do on master branch PR, just to avoid restarting the CI here.

dmitry-gorokhov: @luweizhou2016 Your comment is addressed here: https://github.com/openvinotoolkit/openvino/pull/19360/commits/178e3e29bcc6ba4372cfb8928e020419ec3bc0ac

PR #19375, opened 2023-08-23
[ARM CPU] Enable deconvolution SLT tests

allnes:
Ticket: 118716


src/plugins/intel_cpu/src/nodes/executors/acl/acl_deconv.cpp
```diff
@@ -36,12 +36,16 @@ ACLDeconvTensorInfo getACLDeconvTensorInfo(const DeconvAttrs& deconvAttrs,
...
-    unsigned int pad_r =
-            (deconvAttrs.paddingR.size() > 1) ? static_cast<unsigned int>(deconvAttrs.paddingR.at(1)) : static_cast<unsigned int>(deconvAttrs.paddingR.at(0));
-    unsigned int pad_t = static_cast<unsigned int>(deconvAttrs.paddingL.at(0));
-    unsigned int pad_b = static_cast<unsigned int>(deconvAttrs.paddingR.at(0));
+    int temp_pad_l = (deconvAttrs.paddingL.size() > 1) ? static_cast<int>(deconvAttrs.paddingL.at(1)) : static_cast<int>(deconvAttrs.paddingL.at(0));
+    int temp_pad_r = (deconvAttrs.paddingR.size() > 1) ? static_cast<int>(deconvAttrs.paddingR.at(1)) : static_cast<int>(deconvAttrs.paddingR.at(0));
+    int temp_pad_t = static_cast<int>(deconvAttrs.paddingL.at(0));
+    int temp_pad_b = static_cast<int>(deconvAttrs.paddingR.at(0));
+
+    unsigned int pad_l = temp_pad_l < 0 ? 0 : temp_pad_l;
```
EgorDuplensky, line 44: Just wondering if it is correct, that we can get negative padding here?
It seems the previous version was not expecting this at all.
Does it mean that the previous version was working incorrectly?
Do we have a test for it?

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/arm/convolution_backprop_data.cpp
```diff
@@ -0,0 +1,458 @@
...
+                {{15, 15}}
+        },
+        DeconvInputData{
+                InputShape{{{1, 10}, 12, 7, 7}, {{ 1, 12, 7, 7}, { 2, 12, 7, 7}, { 3, 12, 7, 7}}},
+                ngraph::helpers::InputLayerType::CONSTANT,
+                {{15, 15}}
+        },
+};
+
+INSTANTIATE_TEST_SUITE_P(smoke_Deconv_2D_Planar_FP32, DeconvolutionLayerCPUTest,
```
EgorDuplensky, line 65: It seems we only have separate test instances for Arm and x64
Does it mean we do not have any common instances, which are covered by both Arm and x64?

src/plugins/intel_cpu/tests/functional/single_layer_tests/classes/convolution_backprop_data.cpp
```diff
@@ -0,0 +1,252 @@
...
+        InferenceEngine::PluginConfigParams::YES) {
+        inType = outType = prec = ElementType::bf16;
+        rel_threshold = 1e-2f;
+    } else {
+        inType = outType = prec;
+    }
+
+    selectedType = makeSelectedTypeStr(selectedType, prec);
+
+    if  (inputShape.second.size() > 1 &&
```
EgorDuplensky, line 216: Ideally we should create a separate instances for the input shapes which are supported by x64 and not supported by acl, thus we avoid this check.
But this approach have downsides:
1) It is not clear (at least for me) why acl does not support this particular configuration. This is not like acl does not support some specific algorithm, rank, or something. It is hard to justify the separation for such limitations.
2) If we separate the instances and will test only the "proper" shapes for acl, than we never notice, if the limitation is gone and will continue test the limited configuration of the shapes.

@dmitry-gorokhov Could you please provide you opinion on this one?

src/plugins/intel_cpu/src/nodes/executors/acl/acl_deconv.cpp
EgorDuplensky: Agreed to simply return isSupported=false for the negative paddings

allnes: @EgorDuplensky  done

src/plugins/intel_cpu/tests/functional/single_layer_tests/classes/convolution_backprop_data.cpp
dmitry-gorokhov: I agree with the concern about separate shape instances per backend
Just an idea: can we pass std::function as test parameter which will initilize expected ptimitive type based on input shapes (and possible other arguments)? So we still have separate instances per backend, but all parameters are defined in common part, except srd::function implementation which is backend specific.

EgorDuplensky: We could do it this way, but this contradicts a bit with the initial idea to test a particular implementation regardless of it's priority.
So from my understanding in general we still want to force a particular implementation, but there can be exceptions, when a back-end does not support some specific (tricky?) configuration, which we can handle by some "fallback" logic, like some function, as you mentioned.
Ideally we could localize this logic in some static function in the source code (since it is also required there as part of "isSupported()" anyway), and use it in scope of the tests as well.

EgorDuplensky: @maxnick Could please provide your pov

maxnick: We must test the code that is designed to work. Thus, the question is whether these parameter sets are expected to work? If yes, but in the reality they don't, then this is a bug and we create a ticket and disable the test. If it's not expected that some parameters sets are supported by ACL, then we don't have to declare such test case instances and test other implementations.
Another question is how to conveniently generate test cases considering some backend specifics? Actually we can develop generator functions per backend, which generates test case parameter sets considering all the necessary backend specifics (if any).

allnes: @EgorDuplensky  @dmitry-gorokhov @maxnick corrected, added tests inskip list

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/arm/convolution_backprop_data.cpp
allnes: @EgorDuplensky done

src/plugins/intel_cpu/tests/functional/test_utils/filter_cpu_info.hpp
```diff
@@ -0,0 +1,92 @@
...
+        ///  - For GRU cells, the gates order is update, reset and output gate.
+        ldgo = abcd,
+    } cpu_memory_format_t;
+
+using CPUSpecificParams =  std::tuple<
+    std::vector<cpu_memory_format_t>, // input memomry format
+    std::vector<cpu_memory_format_t>, // output memory format
+    std::vector<std::string>,         // priority
+    std::string                       // selected primitive type
+>;
```
EgorDuplensky, line 87: Why do we moved these definitions as well? They are using more like common utils, not only in the context of filtering CPU info.
I propose to move it back

src/plugins/intel_cpu/tests/functional/specific_tests.cmake
```diff
@@ -40,6 +40,17 @@ if(DEFINED ENABLE_CPU_SUBSET_TESTS_PATH)
     ${CMAKE_CURRENT_SOURCE_DIR}/test_utils/fusing_test_utils.cpp
     ${CPU_SUBSET_TEST_ABS_PATH})
 
+  message(FATAL_ERROR ${X86_64})
```
EgorDuplensky, line 43: I guess this should be cleaned up

src/plugins/intel_cpu/tests/functional/target_per_test.cmake
```diff
@@ -17,6 +17,17 @@ function(create_target_per_test_for_directory TEST_DIR TARGET_PREFIX)
     ${CMAKE_CURRENT_SOURCE_DIR}/test_utils/fusing_test_utils.cpp
   )
 
+  message(FATAL_ERROR ${X86_64})
```
EgorDuplensky, line 20: same here

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/arm/convolution_backprop_data.cpp
EgorDuplensky: @allnes The idea was to implement it as the common instances and use a filter function to drop unnecessary implementation names

PR #19380, opened 2023-08-24
[C API] Fix the BOOLEAN in ov_element_type conflict with the typedefine in Windows.h

zhaixuejun1993:
### Details:
 - Fix conflict of "BOOLEAN", after include header file "Windows.h"

### Tickets:
 - 117890

### DIFF
--- a/src/bindings/c/include/openvino/c/openvino.h
+++ b/src/bindings/c/include/openvino/c/openvino.h
@@ -16,9 +16,10 @@
 #pragma once
 
 #ifdef _WINDOWS_
-#pragma message("The BOOLEAN define in ov_element_type_e conflict with Windows.h BOOLEAN define. Here redifine the BOOLEAN of Windows.h to WIN_BOOLEAN & the BOOLEAN of ov_element_type_e to OV_BOOLEAN.")
+#    pragma message( \
+        "The BOOLEAN define in ov_element_type_e conflict with Windows.h BOOLEAN define. Here redifine the BOOLEAN of Windows.h to WIN_BOOLEAN & the BOOLEAN of ov_element_type_e to OV_BOOLEAN.")
 typedef BOOLEAN WIN_BOOLEAN;
-#define BOOLEAN OV_BOOLEAN
+#    define BOOLEAN OV_BOOLEAN
 #endif
 
 #include "openvino/c/auto/properties.h"
--- a/src/bindings/c/include/openvino/c/openvino.h
+++ b/src/bindings/c/include/openvino/c/openvino.h
@@ -15,12 +15,12 @@
  **/
 #pragma once
 
-#define BOOLEAN OV_BOOLEAN
 #ifdef _WINDOWS_
 #    pragma message( \
         "The BOOLEAN define in ov_element_type_e conflict with Windows.h BOOLEAN define. Here redifine the BOOLEAN of Windows.h to WIN_BOOLEAN & the BOOLEAN of ov_element_type_e to OV_BOOLEAN.")
 typedef BOOLEAN WIN_BOOLEAN;
 #endif
+#define BOOLEAN OV_BOOLEAN
 
 #include "openvino/c/auto/properties.h"
 #include "openvino/c/ov_common.h"
--- a/src/bindings/c/include/openvino/c/openvino.h
+++ b/src/bindings/c/include/openvino/c/openvino.h
@@ -15,6 +15,12 @@
  **/
 #pragma once
 
+#ifdef _WINDOWS_
+#pragma message("The BOOLEAN define in ov_element_type_e conflict with Windows.h BOOLEAN define. Here redifine the BOOLEAN of Windows.h to WIN_BOOLEAN & the BOOLEAN of ov_element_type_e to OV_BOOLEAN.")
+typedef BOOLEAN WIN_BOOLEAN;
+#define BOOLEAN OV_BOOLEAN
+#endif
+
 #include "openvino/c/auto/properties.h"
 #include "openvino/c/ov_common.h"
 #include "openvino/c/ov_compiled_model.h"
--- a/src/bindings/c/include/openvino/c/openvino.h
+++ b/src/bindings/c/include/openvino/c/openvino.h
@@ -17,8 +17,7 @@
 
 #ifdef _WINDOWS_
 #    pragma message( \
-        "The BOOLEAN define in ov_element_type_e conflict with Windows.h BOOLEAN define. Here redifine the BOOLEAN of Windows.h to WIN_BOOLEAN & the BOOLEAN of ov_element_type_e to OV_BOOLEAN.")
-typedef BOOLEAN WIN_BOOLEAN;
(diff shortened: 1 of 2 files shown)


src/bindings/c/include/openvino/c/ov_common.h
```diff
@@ -167,10 +167,11 @@ typedef enum {
  * @ingroup ov_base_c_api
  * @brief This enum contains codes for element type.
  */
+
 typedef enum {
     UNDEFINED = 0U,  //!< Undefined element type
     DYNAMIC,         //!< Dynamic element type
-    BOOLEAN,         //!< boolean element type
+    BIN,             //!< boolean element type
```
ilyachur, line 174: We cannot do it. It is not a backward compatible change

riverlijunjie: Backward compatibility is a problem, but also there is BOOLEAN redefine issue if windows.h is referenced in the same C++/C files. Is there any good idea to solve both problem?

zhaixuejun1993:
@ilya-lavrenov @ilyachur  @riverlijunjie  Pls help review.
image

riverlijunjie:
Please resolve code style issue.

BTW, just discussed with @zhaixuejun1993  there is a reference about deprecating enumeration data name, could we also adopt similar method if there are more name conflict? @ilyachur 

https://github.com/openvinotoolkit/openvino/blob/c5843cf5d63b5f5695c707db652aebecc007c92f/src/core/include/openvino/op/interpolate.hpp#L47-L57

src/bindings/c/include/openvino/c/openvino.h
```diff
@@ -15,6 +15,12 @@
  **/
 #pragma once
 
+#ifdef _WINDOWS_
+#pragma message("The BOOLEAN define in ov_element_type_e conflict with Windows.h BOOLEAN define. Here redifine the BOOLEAN of Windows.h to WIN_BOOLEAN & the BOOLEAN of ov_element_type_e to OV_BOOLEAN.")
+typedef BOOLEAN WIN_BOOLEAN;
+#define BOOLEAN OV_BOOLEAN
+#endif
```
ilya-lavrenov, line 22: can we use it in some test?

riverlijunjie: It is compilation conflict error? Is there any similar test cases for reference?

ilya-lavrenov: I supposed we can emulate situation when Windows .h is used and we used our WA suggestions (OV_BOOLEAN) and it worked.

BTW, is such code cross-platform? E.g. I used OV_BOOLNEAN on Windows, what will it compile on Linux?

In general, I want to use usage scenario of our WA.

zhaixuejun1993: OV_BOOLNEAN is only for Windows, it will not compile on Linux, BOOLEAN for Linux.

src/bindings/c/tests/ov_windows_conflict_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+//
+
+/*
+ * This test only for Windows, which caused by the BOOLEAN define conflict between Windows.h and openvino.h
+ * (ov_element_type_e). Sollution: 1) check Windows.h was used or not in openvino.h for windows OS 2) if YES, will
+ * redefine the "BOOLEAN" from Windows.h to "WIN_BOOLEAN" and also redefine the "BOOLEAN" from openvino.h to
+ * "OV_BOOLEAN"
+ */
+
+#if defined(_MSC_VER)
```
ilya-lavrenov, line 12:
```suggestion
#if defined(_WIN32)
```

src/bindings/c/tests/ov_windows_conflict_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+#    include <Windows.h>
```
ilya-lavrenov, line 13:
```suggestion
#    include <windows.h>
```

src/bindings/c/tests/ov_windows_conflict_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+#    include "ov_test.hpp"
+TEST(ov_windows_conflict_test, ov_windows_boolean_conflict) {
+    EXPECT_NO_THROW(ov_element_type_e element_type =
+                        OV_BOOLEAN);           // The BOOLEAN from ov_element_type_e will be replaced by OV_BOOLEAN
+    EXPECT_NO_THROW(WIN_BOOLEAN win_boolean);  // The BOOLEAN from Windows.h will be replaced by WIN_BOOLEAN
```
ilya-lavrenov, line 19:
```suggestion
    WIN_BOOLEAN win_boolean;  // The BOOLEAN from Windows.h will be replaced by WIN_BOOLEAN
```

src/bindings/c/tests/ov_windows_conflict_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+ * redefine the "BOOLEAN" from Windows.h to "WIN_BOOLEAN" and also redefine the "BOOLEAN" from openvino.h to
+ * "OV_BOOLEAN"
+ */
+
+#if defined(_MSC_VER)
+#    include <Windows.h>
+
+#    include "ov_test.hpp"
+TEST(ov_windows_conflict_test, ov_windows_boolean_conflict) {
+    EXPECT_NO_THROW(ov_element_type_e element_type =
```
ilya-lavrenov, line 17:
```suggestion
    ov_element_type_e element_type =
```

src/bindings/c/include/openvino/c/openvino.h
zhaixuejun1993: Maybe we can do this define "#define BOOLEAN OV_BOOLEAN" for Windows & Linux together, which means:

**- scenario 1: On windows OS, without include windows.h, the following statements work fine (BOOLEAN = OV_BOOLEAN)**
ov_element_type_e element_type = OV_BOOLEAN;
ov_element_type_e element_type = BOOLEAN;

**- scenario 2: On windows OS, with include windows.h, the following statements work**
ov_element_type_e element_type = OV_BOOLEAN;
ov_element_type_e element_type = BOOLEAN;
WIN_BOOLEAN win_boolean;

**- scenario 3: On Linux OS, the following statements work fine (BOOLEAN = OV_BOOLEAN)**
ov_element_type_e element_type = OV_BOOLEAN;
ov_element_type_e element_type = BOOLEAN;

@ilya-lavrenov  how do u think?

ilya-lavrenov: I agree.
But I still think we don't need to define `WIN_BOOLEAN`

zhaixuejun1993: Ok! Let me update

peterchen-intel: @zhaixuejun1993  One option is to add it to message that user can "typedef BOOLEAN WIN_BOOLEAN;" here if user need to use Windows BOOLEAN.

wangleis: Agree to not define `WIN_BOOLEAN` for Windows. Let's clarify the message that Window BOOLEAN cannot be used in this file. If the user does need to use Window BOOLEAN, we expect that the user should use it in a separate file.

ilya-lavrenov (approved):

PR #19390, opened 2023-08-24
[TF Hub][TF FE] Support TensorListLength and TensorListResize operations

rkazants:
**Details:** Support TensorListLength and TensorListResize operations

**Tickets:** 111326, 111327


src/frontends/tensorflow_common/src/op/tensor_list_operations.cpp
```diff
@@ -132,29 +146,82 @@ OutputVector translate_tensor_list_push_back_op(const NodeContext& node) {
...
+    default_op_checks(node, 1, {"TensorListLength"});
+    auto input_handle = node.get_input(0);
+
+    // create auxiliary constants
+    auto zero_const = make_shared<v0::Constant>(element::i32, Shape{1}, 0);
+    auto one_const = make_shared<v0::Constant>(element::i32, Shape{1}, 1);
+
+    // compute the current length of the list
+    auto tensor_list_shape = make_shared<v3::ShapeOf>(input_handle, element::i32);
+    auto list_length = make_shared<v8::Slice>(tensor_list_shape, zero_const, one_const, one_const);
```
mvafin, line 219: What will be result for empty list? In pytorch frontend for similar op we had to use `ReduceSum` to return 0 for empty list https://github.com/openvinotoolkit/openvino/blob/498731f8fd773fbd8e35f953522f7721f726bfc9/src/frontends/pytorch/src/op/len.cpp#L28

rkazants: Empty list means to have the first dimension to be zero `0`. So the slice of this shape will be zero.

mvafin: Shape will be not zero, it will be [], so slice will return empty tensor. Could you add a test for empty list?

gkrivor (approved):

src/frontends/tensorflow_common/src/op/tensor_list_operations.cpp
rkazants: Test was added

mvafin (approved):

PR #19391, opened 2023-08-24
[OVC] Fix output parsing

pavel-esir:
### Details:
 - when `ovc.convert_model` is called from python output should not parsed, e.g. when output contains coma `ov.convert_model(..., output="x,y")` is should be treated as a single output name `"x,y"`.
 - For ovc cli tool parser is called therefore `ovc input_model.onnx ... output="x,y"` - means 2 outputs `"x"` and `"y"`. Moreover in cli tool spaces are ignored therefore:
  `ovc ... output="x,y"` == `ovc ... output="x, y"` == `ovc ... output=" x , y "`
 
### Tickets:
 - [CVS-118779](https://jira.devtools.intel.com/browse/CVS-118779)


slyalin (changes requested):
Apply changes for openvino.convert_model only: it shouldn't parse strings, with spaces or without spaces, single string always means single element. For ovc parsing should remain as usual.

pavel-esir:
> Apply changes for openvino.convert_model only: it shouldn't parse strings, with spaces or without spaces, single string always means single element. For ovc parsing should remain as usual.



> Apply changes for openvino.convert_model only: it shouldn't parse strings, with spaces or without spaces, single string always means single element. For ovc parsing should remain as usual.

done. please review

tools/ovc/openvino/tools/ovc/convert.py
```diff
@@ -72,7 +72,10 @@ def convert_model(
             do not add :0 to this name. The order of outputs in converted model is the
             same as order of specified operation names. For PaddlePaddle model represented
             as a Python object, you can specify outputs as a PaddlePaddle Python objects or
-            a list of such objects.
+            a list of such objects. Example with ovc.convert_model: output=["out_1", "out_2"].
```
slyalin, line 75: Don't refer to `convert_model` as `ovc.convert_model`. It is exposed to user as `openvino.convert_model`. Just use `convert_model`. In this case, change is not needed at all.

slyalin: Also move example with a list before PaddlePaddle case description.

tools/ovc/openvino/tools/ovc/convert_impl.py
```diff
@@ -86,10 +94,19 @@ def arguments_post_parsing(argv: argparse.Namespace):
...
-    argv.output = argv.output.split(',') if isinstance(argv.output, (str, pathlib.Path)) else argv.output
-
     log.debug("Placeholder shapes : {}".format(argv.placeholder_shapes))
 
+    if not hasattr(argv, 'output') or argv.output is None:
+        return argv
+
+    if argv.is_python_api_used:
+        error_msg = f"output '{argv.output}' is incorrect, it should be string or a list of strings"
+        assert isinstance(argv.output, (str, list)), error_msg
```
slyalin, line 104: List or tuple. We use them interchangeably in these parameters. @popovaan applied "list or tuple" feature for `input` parameter, please find it and make similar changes here.

tools/ovc/unit_tests/ovc/utils/cli_parser_test.py
```diff
@@ -343,12 +343,81 @@ def test_mo_convert_params(self):
...
 
         for arg, value in vars(argv).items():
             if arg not in args and arg != 'is_python_api_used':
                 assert value == cli_parser.get_default(arg)
 
+    def test_output_post_parsing_1(self):
+        args = {'input_model': os.path.dirname(__file__),
+                'input': "input_1[1,2,3]",
+                'output_model': os.getcwd() + "model.xml",
+                "framework": "onnx",
```
slyalin, line 356: Why do we need to write this parameter? `ovc` doesn't have it as a publicly visible.

tools/ovc/unit_tests/ovc/utils/cli_parser_test.py
```diff
@@ -343,12 +343,81 @@ def test_mo_convert_params(self):
...
+                'verbose': False,
+                'output': "a,b,c"}
+
+        argv = args_to_argv(**args)
+
+        argv.is_python_api_used = False
+        argv = arguments_post_parsing(argv)
```
slyalin, line 363: I don't think that just tests for one internal function is enough. Could we have tests that check outside visible behavior with clear separation between CLI and openvino.compile_model.

slyalin (changes requested):

tools/ovc/openvino/tools/ovc/convert_impl.py
pavel-esir: relaxed type assert, tuples are also allowed

tools/ovc/openvino/tools/ovc/convert.py
```diff
@@ -72,7 +72,10 @@ def convert_model(
...
+            When ovc cli tool is used outputs should be separated with coma (spaces are ignored):
+            Example 1: ovc ... output="out_1, out_2".
+            Example 2: ovc ... output="x,y,z" equivalent to ovc ... output="x, y, z".
```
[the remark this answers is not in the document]
pavel-esir: added ovc help/examples separately into specific place: `help.py`

tests/layer_tests/ovc_python_api_tests/test_complex_params.py
```diff
@@ -158,6 +159,126 @@ def test_mo_convert_tf_model_single_input_output(self, params, ie_device, precis
...
+        return model_path
+
+    @pytest.mark.nightly
+    @pytest.mark.precommit
+    def test_ovc_convert_model_with_coma_in_names(self, ie_device, precision, ir_version,
+                                                  temp_dir, use_new_frontend, use_old_api):
+        onnx_net_path = self.create_onnx_model_with_coma_in_names(temp_dir)
+        ref_model = self.create_ref_graph_with_coma_in_names()
+        test_params = {'output': 'relu_1,relu_2'}
+        test_params.update({'input_model': onnx_net_path})
```
pavel-esir, line 268:
```suggestion
        test_params.update({'input_model': onnx_net_path})
        
```

tools/ovc/unit_tests/ovc/utils/cli_parser_test.py
pavel-esir: Added a couple of more solid test cases.

One solely for `convert_model` for a model with output name `x,y`. (cli tool can not process such outputs)
And one case where 2 outputs separated b coma (for cli) or specified in a list (for convert_model).

https://github.com/openvinotoolkit/openvino/pull/19391/files#diff-e3878abd42ca567a18e9b90e4bd37a1394a1db1106502b0ee15b4985fc2fb856R263-R279

tools/ovc/unit_tests/ovc/utils/cli_parser_test.py
pavel-esir: This is a very good question! Unfortunately without this attribute `arguments_post_parsing` was failing because inside it calls `params_parsing` which checks `argv.framework`. Even in case of using `convert_model` is already initialized for us.

I didn't like introducing such a redundant attribute into tests. I detaste to such dangling placeholding attributes introduced for only reason that the test does not fall, i made check for `arvg.framework` safer and removed it:
https://github.com/openvinotoolkit/openvino/pull/19391/files#diff-fea8ab7df43344bcc52da931836648c9e47bf0fd173be1961d35d2718cbcb11aR364-R365

PS: Unfortunately even `convert_model` is bound to cli parser because `convert_model` uses `_convert(..)` which has a lot of references to cli_parser inside itself. Probably it would be better move all parser things into cli tool `__main__` https://github.com/openvinotoolkit/openvino/blob/f0300a36eb42115181a1509c1ca82324e81556cf/tools/ovc/openvino/tools/ovc/main.py#L19-L32 but for the moment we don't have time for that, imho.

pavel-esir:
@slyalin applied comments. Please review

tools/ovc/openvino/tools/ovc/convert_impl.py
```diff
@@ -76,20 +77,38 @@ def print_argv(argv: argparse.Namespace):
...
+    if argv.is_python_api_used:
+        error_msg = f"output '{argv.output}' is incorrect, it should be string or a list/tuple of strings"
+        assert isinstance(argv.output, (str, list, tuple)), error_msg
+        if isinstance(argv.output, list):
+            assert check_iterable_types(argv.output, str), error_msg
+        else:
+            argv.output = [argv.output]
+    else:
+        assert isinstance(argv.output, str)
+        argv.output = argv.output.replace(' ', '').split(',')
```
slyalin, line 111: This is not good. If the space is inside the name and there is no comma nearby it means the wrong name. For example: "na me" is not the same as "name", this is an error. Use `regex.split(r'\s*,\s*', argv.output.strip())` and check that the resulting elements don't contain spaces and are not empty, or raise an error with the explanation otherwise.

tools/ovc/openvino/tools/ovc/help.py
```diff
@@ -26,6 +26,13 @@ def get_convert_model_help_specifics():
                  'If the dimension is set as an integer (like 100 in [1,100]), such a dimension is not supposed '
                  'to be changed later, during a model conversion it is treated as a static value. '
                  'Example with unnamed inputs: \"[1,100],[1,?]\".'},
+        'output':
+            {'description':
+                'The name of the output operation of the model or list of names. For TensorFlow*, '
+                'do not add :0 to this name. The order of outputs in converted model is the '
+                'same as order of specified operation names. Outputs should be separated with coma (spaces are ignored): '
+                'Example 1: ovc ... output="out_1, out_2". '
+                'Example 2: ovc ... output="x,y,z" equivalent to ovc ... output="x, y, z".'},
```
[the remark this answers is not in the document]
slyalin: Reformulate `convert_model` description in another place accordingly (cannot make a suggestion there due to github limitations).

slyalin (changes requested):

tools/ovc/openvino/tools/ovc/help.py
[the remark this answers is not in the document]
pavel-esir: done

tools/ovc/openvino/tools/ovc/convert_impl.py
pavel-esir: done

slyalin (approved):

PR #19414, opened 2023-08-25
[Core] Use API 2.0 in evaluate for trigonometric operators

praasz:
### Details:
 - Migrate the evaluate member function in trigonometric operators to new version using `ov::Tensor`.
 - Drop using `ngraph` namespace in modified operators.
 - Refactor reference implementations for updated operators.

### Tickets:
 - [CVS-119471](https://jira.devtools.intel.com/browse/CVS-118471)


src/core/src/op/acos.cpp
```diff
@@ -28,53 +35,26 @@ std::shared_ptr<ov::Node> ov::op::v0::Acos::clone_with_new_inputs(const OutputVe
...
-template <ov::element::Type_t ET>
-inline bool evaluate(const ngraph::HostTensorPtr& arg0, const ngraph::HostTensorPtr& out, const size_t count) {
-    using T = typename ov::element_type_traits<ET>::value_type;
-    ngraph::runtime::reference::acos<T>(arg0->get_data_ptr<ET>(), out->get_data_ptr<ET>(), count);
-    return true;
-}
-
-bool evaluate_acos(const ov::HostTensorPtr& arg0, const ov::HostTensorPtr& out, const size_t count) {
-    bool rc = true;
-    out->set_unary(arg0);
```
mitruska, line 43: In the original evaluate, shape and element type of the output `HostTensor` were set.
Don't we need corresponding code to set the shape and element type of the output `ov::Tensor`?

https://github.com/openvinotoolkit/openvino/blob/679369c7071d624a334a2c52145cc0457a56e318/src/core/src/runtime/host_tensor.cpp#L151-L154
cc: @ilyachur

src/core/src/op/cos.cpp
```diff
@@ -2,81 +2,68 @@
...
     OV_OP_SCOPE(v0_Cos_evaluate);
-    OPENVINO_SUPPRESS_DEPRECATED_START
-    NGRAPH_CHECK(validate_host_tensor_vector(outputs, 1) && validate_host_tensor_vector(inputs, 1));
-    OPENVINO_SUPPRESS_DEPRECATED_END
-    return cosop::evaluate_cos(inputs[0], outputs[0], shape_size(inputs[0]->get_shape()));
+    using namespace ov::element;
+    return IfTypeOf<i32, i64, u32, u64, f16, f32>::apply<cos::Evaluate>(inputs[0].get_element_type(),
+                                                                        inputs[0],
+                                                                        outputs[0],
+                                                                        shape_size(inputs[0].get_shape()));
```
mitruska, line 50: The original check:
 `NGRAPH_CHECK(validate_host_tensor_vector(outputs, 1) && validate_host_tensor_vector(inputs, 1));`
 ensured the size of the input/outputs vector to safely get element at the index.
 
Currently `inputs[0]` and `outputs[0]` are reached without the `TensorVector` size check.

mitruska (changes requested):

src/core/src/op/acos.cpp
praasz: I'm not sure if this is really required for ov::Tensor @ilyachur can comment about it?

src/core/src/op/cos.cpp
praasz: This check was not there before and not sure if required at all.
When there is a graph the all inputs should be set and tensor should be created for them correctly.
@ilyachur do we really need check number of input/output tensors? or is done some upper layers when tensors are created?

mitruska: > This check was not there before and not sure if required at all.

The comment is directly related to the change, where such check was before:
image


and currently is on the master:
https://github.com/openvinotoolkit/openvino/blob/cf2c7da568934870c29acc961a4498ff9cbd8d9c/src/core/src/op/cos.cpp#L61-L63

The `evaluate` method is a part of public API of the Node class, also with Python API bindings, 
so it can be used directly by a user, with a TensorVector of any size. 
https://github.com/openvinotoolkit/openvino/blob/a45e5e03c5a2e412447d82989c9d69f7f51f9e41/src/bindings/python/tests/test_graph/test_basic.py#L442-L445
Such check can prevent unexpected issues, especially when the operator has several inputs and outputs.

praasz: But acos, acosh, atan... and there will be more ops which not implement this check.

ilyachur: From my point of view it will be better to check, because technically user can put empty vector of output tensors to the operation, it will be better to create required tensors or throw an exception than get a segfault.

src/core/src/op/acos.cpp
ilyachur: I wouldn't want to add such functionality to `ov::Tensor` I think the current functionality should be enough.
Maybe we can add `set_element_type()` to tensor

ilyachur: Yes we can use set_shape. Also as I see right now `ov::Tensor` doesn't have `set_element_type()` method. We can add it if it is needed.

src/core/src/op/asinh.cpp
```diff
@@ -2,77 +2,61 @@
...
+#include "element_visitor.hpp"
 #include "itt.hpp"
-#include "ngraph/op/util/elementwise_args.hpp"
-#include "ngraph/runtime/host_tensor.hpp"
-#include "ngraph/type/element_type.hpp"
 #include "openvino/reference/asinh.hpp"
 
-using namespace std;
-using namespace ngraph;
+namespace ov {
```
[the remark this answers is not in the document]
ilyachur: With using std in the cpp files we get a mix of styles: some places use std and have a code like `std::something` some places doesn't use and code looks like `something`.
Without using std we have one style inside one file.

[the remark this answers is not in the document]
praasz: I've changed it because of some automation.
It also is considered that `using namespace std` (or any other) as bad practice to use it in .cpp on global scope. What is recommended to use it in very limited scope (function, code block) to have control over it. For global use, import  only required features from namespace.

src/core/src/op/acos.cpp
praasz: The `set_shape` added. I think it can be required in case of constant folding (when lower/upper bounds of shape are evaluated) then the output can have dynamic shape and output tensor has to be set the correct shape.

src/core/src/op/asin.cpp
```diff
@@ -2,83 +2,62 @@
...
-        NGRAPH_TYPE_CASE(evaluate_asin, u64, arg0, out, count);
-        NGRAPH_TYPE_CASE(evaluate_asin, f16, arg0, out, count);
-        NGRAPH_TYPE_CASE(evaluate_asin, f32, arg0, out, count);
-    default:
-        rc = false;
-        break;
-    }
-    return rc;
+    return std::make_shared<Asin>(new_args.at(0));
 }
```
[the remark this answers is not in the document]
praasz: [See](https://github.com/openvinotoolkit/openvino/pull/19414#discussion_r1306874390)

mitruska (approved):

src/core/src/op/acos.cpp
mitruska: Great, I've approved the changes, the need of `set_element_type` can be considered separately.

src/core/reference/include/openvino/reference/acosh.hpp
```diff
@@ -4,25 +4,35 @@
...
+namespace ov {
 namespace reference {
-template <typename T, typename std::enable_if<!std::is_integral<T>::value, bool>::type = true>
-void acosh(const T* arg, T* out, size_t count) {
-    for (size_t i = 0; i < count; i++) {
-        out[i] = static_cast<T>(std::acosh(arg[i]));
-    }
+namespace func {
+template <class T, typename std::enable_if<ov::is_floating_point<T>()>::type* = nullptr>
+T acosh(const T in) {
```
ilyachur, line 16: Why do we have different implementations for `acos` and `acosh`?
Should we align the approach?

praasz: I didn't want change functionality to current implementation only refactor it and assumed that cast is intentional.
I will make separate PR for it as rounding gives more accurate result for integral values.

ilyachur (approved):

PR #19422, opened 2023-08-25
[GPU] add sqrt activation support in cpu_impl

wilson-seok:
### Details:
 - add sqrt activation support in cpu_impl
 
### Tickets:
 - 117829


yeonbok:
Paired PR for master: https://github.com/openvinotoolkit/openvino/pull/19421

PR #19430, opened 2023-08-27
LPT opset12::Pad negative indexes support

evkotov:
### Details:
 - Update LPT PadTransformation to support v12::Pad negative indexes
 - Add unit test for negative and mixed cases

### Tickets:
 - 115474

### DIFF
--- a/src/common/low_precision_transformations/src/pad.cpp
+++ b/src/common/low_precision_transformations/src/pad.cpp
@@ -178,15 +178,22 @@ bool PadTransformation::canBeTransformed(const TransformationContext& context, s
         return false;
     }
 
-    if (!hasPositiveIndexes(pad))
-        return true;
-
     const auto dequantization = NetworkHelper::getDequantization(op, defaultPrecisions);
     if (dequantization.empty()) {
         return false;
     }
 
     const auto mode = pad->get_pad_mode();
+    if (mode != op::PadMode::CONSTANT &&
+        mode != op::PadMode::EDGE &&
+        mode != op::PadMode::REFLECT &&
+        mode != op::PadMode::SYMMETRIC) {
+        return false;
+    }
+
+    if (!hasPositiveIndexes(pad))
+        return true;
+
     if (mode == op::PadMode::CONSTANT) {
         auto padAndDqByTheSameDimension = [&](const std::shared_ptr<ov::opset1::Constant>& deqConst) {
             const auto padsBegin = pad->get_pads_begin();
--- a/src/common/low_precision_transformations/src/pad.cpp
+++ b/src/common/low_precision_transformations/src/pad.cpp
@@ -178,16 +178,16 @@ bool PadTransformation::canBeTransformed(const TransformationContext& context, s
         return false;
     }
 
-    const auto dequantization = NetworkHelper::getDequantization(op, defaultPrecisions);
-    if (dequantization.empty()) {
-        return false;
-    }
-
     const auto mode = pad->get_pad_mode();
     if (mode != op::PadMode::CONSTANT &&
         mode != op::PadMode::EDGE &&
         mode != op::PadMode::REFLECT &&
         mode != op::PadMode::SYMMETRIC) {
+        return false;
+    }
+
+    const auto dequantization = NetworkHelper::getDequantization(op, defaultPrecisions);
+    if (dequantization.empty()) {
         return false;
     }
 
--- a/src/common/low_precision_transformations/src/pad.cpp
+++ b/src/common/low_precision_transformations/src/pad.cpp
@@ -37,6 +37,21 @@ PadTransformation::PadTransformation(const Params& params) : LayerTransformation
     this->register_matcher(m, callback);
 }
 
+namespace {
+    bool hasNegativeIndexes(const std::shared_ptr<ov::op::util::PadBase>& pad) {
+        const auto padsBegin = pad->get_pads_begin();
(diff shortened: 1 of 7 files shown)


src/common/low_precision_transformations/tests/pad_transformation.cpp
```diff
@@ -533,15 +595,121 @@ INSTANTIATE_TEST_SUITE_P(
...
+        {
+            ngraph::element::i8,
+            {
+                {ngraph::element::f32},
+                {{64.f, 64.f, 64.f, 32.f, 32.f, 32.f}, ngraph::element::f32, {1, 1, 6, 1}},
+                {{3.f, 3.f, 3.f, 2.f, 2.f, 2.f}, ngraph::element::f32, {1, 1, 6, 1}}
+            }
+        },
+        {
+            ngraph::element::i8,
```
ilyachur, line 685: Please use `ov` namespace instead of `ngraph`

src/common/low_precision_transformations/src/pad.cpp
```diff
@@ -179,10 +181,6 @@ bool PadTransformation::canBeTransformed(const TransformationContext& context, s
         return false;
     }
 
-    if (hasNegativeIndexes(pad)) {
-        return false;
-    }
-
     const auto dequantization = NetworkHelper::getDequantization(op, defaultPrecisions);
     if (dequantization.empty()) {
         return false;
```
v-Golubev, line 186: Below we have checks for Constant and Reflect pad modes: they are actual only in case of positive or mixed indexes. If all indexes are negative, we can return true before these checks

src/common/low_precision_transformations/tests/pad_transformation.cpp
```diff
@@ -707,12 +951,88 @@ INSTANTIATE_TEST_SUITE_P(
...
+INSTANTIATE_TEST_SUITE_P(
+    smoke_LPT,
+    PadTransformation,
+    ::testing::Combine(
+        ::testing::ValuesIn(inputShapes),
+        ::testing::Values(padsNegativeBySpatialDimensions),
+        ::testing::Values(ngraph::op::PadMode::SYMMETRIC),
+        ::testing::Values(0.f),
+        ::testing::ValuesIn(testValuesForSymetricMode)),
+    PadTransformation::getTestCaseName);
```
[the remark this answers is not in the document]
evkotov: I would like to have tests as many as possible. If you don't mind I would like to have them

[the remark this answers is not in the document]
v-Golubev: I don't insist, let's leave these test cases

src/common/low_precision_transformations/tests/pad_transformation.cpp
```diff
@@ -73,6 +72,8 @@ class PadTransformation : public LayerTransformation, public testing::WithParamI
             precisionAfterActualOp,
             { {}, {}, {} });
 
+        auto actualFunctionReference = actualFunction->clone();
+
```
v-Golubev, line 76: Please remove

v-Golubev (approved):
LGTM

This would be great if we ran experiments on big model scope to detect a performance impact

src/common/low_precision_transformations/src/pad.cpp
```diff
@@ -179,9 +178,8 @@ bool PadTransformation::canBeTransformed(const TransformationContext& context, s
...
     }
 
-    if (hasNegativeIndexes(pad)) {
-        return false;
-    }
+    if (!hasPositiveIndexes(pad))
+        return true;
```
mateusztabaka, line 182: Is it ok to return true here? Can Pad be transformed if it doesn't have positive pad values?

eshoguli (commented):
Can you accordantly update LPT tests for plugins, please:
* CPU: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/tests/functional/shared_tests_instances/low_precision_transformations/pad_transformation.cpp
* GPU: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_gpu/tests/functional/shared_tests_instances/low_precision_transformations/pad_transformation.cpp

src/common/low_precision_transformations/src/pad.cpp
evkotov: This code is a fix after @v-Golubev comment "Below we have checks for Constant and Reflect pad modes: they are actual only in case of positive or mixed indexes. If all indexes are negative, we can return true before these checks"

eshoguli (approved):
LGTM

src/common/low_precision_transformations/src/pad.cpp
eshoguli: dequantization operations check:
```
    const auto dequantization = NetworkHelper::getDequantization(op, defaultPrecisions);
    if (dequantization.empty()) {
        return false;
    }
```
has to be before this condition

eshoguli (changes requested):
1) There is comment in `canBeTransformed` which should be fixed.
2) Can we ask you to improve `canBeTransformed`: add only supported `mode` check: `op::PadMode::CONSTANT`, `op::PadMode::EDGE`, `op::PadMode::REFLECT` or `op::PadMode::SYMMETRIC`. If other, then `canBeTransformed` has to return `false`.

eshoguli (approved):
LGTM

PR #19435, opened 2023-08-28
[ShapeInfer] Remove old API from shape inference sources

praasz:
### Details:
 - Remove `ngraph` namespace from shape inference source code.
 - Temporary add `ngraph` includes in operators which will be removed when operator migrated to use API 2.0.

### Tickets:
 - [CVS-119067]([https://jira.devtools.intel.com/browse/CVS-119067)


jane-intel (approved):
I don't get why in some ops you've added ngraph/validation_utils.hpp header -- however we will  be reviewing ops later on so I don't mind it.

PR #19443, opened 2023-08-28
[PyOV] Update `scipy` to avoid security vulnerability and limit `cython`

p-wysocki:
### Tickets:
 - 117438
 - 119079

PR #19444, opened 2023-08-28
[CPU][ARM] Disable default fp16 inference precision for non-CNN

alvoron:
Reverts https://github.com/openvinotoolkit/openvino/pull/19069 because of accuracy issues.


alvoron:
Requires https://github.com/openvinotoolkit/openvino/pull/19443 to make CI green

dmitry-gorokhov (approved):

PR #19446, opened 2023-08-28
Disallow LeakyReluFusion when alpha is greater than one

mateusztabaka:
Tickets: CVS-118898, CVS-82454


src/common/transformations/src/transformations/common_optimizations/leaky_relu_fusion.cpp
```diff
@@ -18,7 +18,7 @@
 ov::pass::LeakyReluFusion::LeakyReluFusion() {
     MATCHER_SCOPE(LeakyReluFusion);
     auto data_pattern = pass::pattern::any_input();
-    auto alpha_pattern = pass::pattern::any_input(pattern::has_static_shape());
+    auto alpha_pattern = pass::pattern::wrap_type<op::v0::Constant>();
```
itikhono, line 21: it looks like the previous version worked for subgraphs that can be const folded
If we postpone  fp16 "const folding" of these subgraphs until the CPU phase, does it mean that this transformation won't be executed or CPU runs const folding before the transformation pipeline?
Could you clarify it with @pavel-esir ?

mateusztabaka: Right, this transformation won't be executed. We postpone fp16 Constant->Convert subgraphs constantfolding to first inference.

src/common/transformations/src/transformations/common_optimizations/leaky_relu_fusion.cpp
```diff
@@ -30,6 +30,17 @@ ov::pass::LeakyReluFusion::LeakyReluFusion() {
         if (shape_size(original_alpha_pattern.get_shape()) != 1)
             return false;
 
+        auto constant = ov::as_type_ptr<op::v0::Constant>(original_alpha_pattern.get_node_shared_ptr());
+        if (!constant)
+            return false;
+
+        float value;
```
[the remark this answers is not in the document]
mateusztabaka: it can be, but fp64 models are very rare and the constant would have to be at minimum something like 1.0000001 to make it slightly incorrect. Also given that plugins convert f64 to f32 - we won't see the difference anyway.

itikhono (approved):

PR #19456, opened 2023-08-28
[GPU] Create inner program without CNNNetwork

ahnyoung-paul:
### Details:
 - *Create inner program without CNNNetwork. Some inner network has only empty dimension constant. However, CNNNetwork is not allowed empty dimension. Thus, create inner program without CNNNetwork.*
 - *Add functional test case for empty dimension.*

### Tickets:
 - *116928*


src/plugins/intel_gpu/src/plugin/program_builder.cpp
```diff
@@ -318,6 +293,106 @@ ProgramBuilder::ProgramBuilder(cldnn::engine& engine, const ExecutionConfig& con
...
+                                std::shared_ptr<ov::threading::IStreamsExecutor> task_executor)
+    : m_curBatch(-1)
+    , m_config(config)
+    , m_engine(engine)
+    , queryMode(false)
+    , m_task_executor(task_executor) {
+    if (m_task_executor == nullptr)
+        m_task_executor = cldnn::program::make_task_executor(m_config);
+
+    std::shared_ptr<ov::Model> ngraph_function(model);
```
yeonbok, line 307: minor comment but could you change the name of ngraph_ ? because it is deprecated naming. How about just ov_model ? 

BTW can't we just use model as is ?

src/plugins/intel_gpu/src/plugin/program_builder.cpp
```diff
@@ -318,6 +293,106 @@ ProgramBuilder::ProgramBuilder(cldnn::engine& engine, const ExecutionConfig& con
...
+            dims.resize(shape.size(), 0);
+            for (size_t i = 0; i < shape.size(); ++i) {
+                if (shape[i].get_max_length() != -1)  // dimension has an estimation
+                    dims[i] = shape[i].get_max_length();
+            }
+        }
+        auto rank = shape.rank().is_static() ? shape.rank().get_length() : -1;
+        const auto rankLayout = rank < 0 ? InferenceEngine::Layout::BLOCKED : TensorDesc::getLayoutByRank(rank);
+        const auto precision = InferenceEngine::details::convertPrecision(output.get_element_type());
+        ptr.reset(new InferenceEngine::Data(outName, {precision, dims, rankLayout}));
```
yeonbok, line 332: can't we create it as shared_ptr?

yeonbok: Also I think we can add assert that rank should be static because we dont support dynamic rank

src/plugins/intel_gpu/src/plugin/program_builder.cpp
ahnyoung-paul: Fix codes to use model instead of ngraph_function. thanks.

src/plugins/intel_gpu/src/plugin/program_builder.cpp
ahnyoung-paul: Add ASSERT to check rank is static and remove redundant codes. thanks.

src/plugins/intel_gpu/src/plugin/program_builder.cpp
```diff
@@ -318,6 +293,106 @@ ProgramBuilder::ProgramBuilder(cldnn::engine& engine, const ExecutionConfig& con
...
+    }
+
+    // inner network is only allowed single batch
+    m_max_batch = 1;
+
+    // Set input / output for Parameter and Result in cldnn
+    InferenceEngine::InputsDataMap networkInputs = {};
+    InferenceEngine::OutputsDataMap networkOutputs = {};
+
+    auto create_data_for_result = [&](const ngraph::Output<ngraph::Node>& output,
```
vladimir-paramuzov, line 318: [random spot] Does https://github.com/openvinotoolkit/openvino/pull/18920 solves the original issue? If yes, then maybe we can postpone this PR a little bit, as you use/modify some things that are going to be removed

ahnyoung-paul: @vladimir-paramuzov the main purpose of this PR is creating cldnn::program without CNNNetwork to fix zero dimension issue and I think https://github.com/openvinotoolkit/openvino/pull/18920 can solve the issue. Thus, I will postpone this PR until  https://github.com/openvinotoolkit/openvino/pull/18920 will be merged. thanks.

vladimir-paramuzov: @ahnyoung-paul  PR https://github.com/openvinotoolkit/openvino/pull/18920 was merged. So please update this one or close if it's not needed anymore.

PR #19465, opened 2023-08-28
[GPU] Fixed reordered memory cache not to contain original weight memory

yeonbok:
### Details:
 - Previously reordered_weights_cache contained original memory too, which is just reinterpreted
 - It is waste of reordered_weights_cache slot, and causes eviction of actually reordered weights which may be needed in the future. 
-  Fixed above issue 
### Tickets:
 - 119145

### DIFF
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -1152,7 +1152,7 @@ event::ptr primitive_inst::update_weights() {
             GPU_DEBUG_PROFILED_STAGE_CACHE_HIT(true);
             GPU_DEBUG_TRACE_DETAIL << id() << ": reinterpret original weights memory from " << original_layout.to_short_string()
                                            << " to " << expected_layout.to_short_string() << std::endl;
-            _impl_params->weights_layout = optional_layout(original_layout);
+            _impl_params->weights_layout = optional_layout(expected_layout);
             return nullptr;
         } else {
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1153,7 +1152,7 @@ event::ptr primitive_inst::update_weights() {
             GPU_DEBUG_PROFILED_STAGE_CACHE_HIT(true);
             GPU_DEBUG_TRACE_DETAIL << id() << ": reinterpret original weights memory from " << original_layout.to_short_string()
                                            << " to " << expected_layout.to_short_string() << std::endl;
-            _reordered_weights_cache.add(expected_layout, engine.reinterpret_buffer(*original_weights_memory, expected_layout));
+            _impl_params->weights_layout = optional_layout(original_layout);
```
e-ddykim, line 1155: Why is the `weights_layout` set to `original_layout` instead of `expected_layout`?

yeonbok: Ah, they are physically same (compatible) so no matter which is set. 
However I will set it as expected layout to be more explicitly. Thanks!

e-ddykim: Oh, if so, this line can be deleted, because it's already set by the above Line 1145.

yeonbok: Applied thanks!

e-ddykim (approved):

PR #19467, opened 2023-08-29
[GPU] Update strided_slice for partially dynamic shape case

kelvinchoi-intel:
### Details:
 - *Update strided_slice for partially dynamic shape case*
 - *Update func test to test mixed input with constant and non-constant*

### Tickets:
 - *118757*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/strided_slice.cpp
+++ b/src/plugins/intel_gpu/src/graph/strided_slice.cpp
@@ -44,8 +44,20 @@ std::vector<layout> strided_slice_inst::calc_output_layouts(strided_slice_node c
     if ((begin_data.empty() && !constant_mem.count(1))
         || (end_data.empty() && !constant_mem.count(2))
         || (strides_data.empty() && !constant_mem.count(3))) {
-        auto out_shape = ov::PartialShape::dynamic(input0_layout.get_partial_shape().size());
-        return { layout{out_shape, input0_layout.data_type, format::get_default_format(out_shape.rank().get_length())} };
+        auto input0_pshape = input0_layout.get_partial_shape();
+        auto input0_len = input0_pshape.size();
+        auto out_shape = ov::PartialShape::dynamic(input0_len);
+        if (input0_layout.is_dynamic()) {
+            // fill with static shape until it finds dynamic
+            for (size_t i = 0; i < input0_len; i++) {
+                if (input0_pshape[i].is_static())
+                    out_shape[i] = input0_pshape[i];
+                else
+                    break;
+            }
+        }
+
+        return { layout{out_shape, input0_layout.data_type, format::get_default_format(input0_len)} };
     }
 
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/graph/strided_slice.cpp
```diff
@@ -45,7 +45,7 @@ std::vector<layout> strided_slice_inst::calc_output_layouts(strided_slice_node c
         || (end_data.empty() && !constant_mem.count(2))
         || (strides_data.empty() && !constant_mem.count(3))) {
         auto out_shape = ov::PartialShape::dynamic(input0_layout.get_partial_shape().size());
-        return { layout{out_shape, input0_layout.data_type, format::get_default_format(out_shape.rank().get_length())} };
+        return { layout{input0_layout.get_partial_shape(), input0_layout.data_type, format::get_default_format(out_shape.rank().get_length())} };
```
wilson-seok, line 48: Is there any potential issue when input rank is not same with output rank?

kelvinchoi-intel: I think there is no issue during compiling of dynamic shape. I've just used input0 layout as a default.

src/plugins/intel_gpu/src/graph/strided_slice.cpp
```diff
@@ -44,8 +44,20 @@ std::vector<layout> strided_slice_inst::calc_output_layouts(strided_slice_node c
     if ((begin_data.empty() && !constant_mem.count(1))
         || (end_data.empty() && !constant_mem.count(2))
         || (strides_data.empty() && !constant_mem.count(3))) {
-        auto out_shape = ov::PartialShape::dynamic(input0_layout.get_partial_shape().size());
-        return { layout{out_shape, input0_layout.data_type, format::get_default_format(out_shape.rank().get_length())} };
+        auto input0_pshape = input0_layout.get_partial_shape();
+        auto input0_len = input0_pshape.size();
+        auto out_shape = ov::PartialShape::dynamic(input0_len);
+        if (input0_layout.is_dynamic()) {
```
yeonbok, line 50: I think its always dynamic, isnt it?

kelvinchoi-intel: input0 could be static while input1/input2/input3 is dynamic

ahnyoung-paul: If input0 could be static, how to set out_shape? in this code, out_shape[] is not filled when input0_layout is static.
it only fills out_shape[] when input0_layout is dynamic.

yeonbok: Yeah at first I misread it as output. BTW I also would like to ask how static input could be handled as Paul mentioned? I think input's dynamism does not matter.

yeonbok (approved):

PR #19471, opened 2023-08-29
TransposeSinking - add support for ShapeOf

mateusztabaka:
It's often in the model that transposes have two consumers: ShapeOf and another transpose and with ShapeOf absent - the transpose pair could be eliminated. This patch's approach is to propagate a transpose through ShapeOf by creating "ShapeOf->Gather" subgraph and replacing ShapeOf's input with transpose input.

Ticket: CVS-118896

### DIFF
--- a/src/common/transformations/include/transformations/transpose_sinking/ts_shape_of.hpp
+++ b/src/common/transformations/include/transformations/transpose_sinking/ts_shape_of.hpp
@@ -1,0 +1,28 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include "transformations/transpose_sinking/ts_base.hpp"
+#include "transformations_visibility.hpp"
+
+namespace ov {
+namespace pass {
+namespace transpose_sinking {
+
+class TRANSFORMATIONS_API TSShapeOfForward;
+
+}  // namespace transpose_sinking
+}  // namespace pass
+}  // namespace ov
+
+/**
+ * @ingroup ie_transformation_common_api
+ * @brief TSShapeOfForward transformation sinks Transpose through ShapeOf in the forward direction.
+ */
+class ov::pass::transpose_sinking::TSShapeOfForward : public ov::pass::transpose_sinking::TSForwardBase {
+public:
+    OPENVINO_RTTI("TSShapeOfForward", "0");
+    TSShapeOfForward();
+};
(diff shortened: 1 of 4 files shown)


src/common/transformations/include/transformations/transpose_sinking/ts_shape_of.hpp
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include "transformations/transpose_sinking/ts_base.hpp"
+#include "transformations_visibility.hpp"
+
+namespace ov {
+namespace pass {
+namespace transpose_sinking {
+
+class TRANSFORMATIONS_API TSShapeOfForward;
```
itikhono, line 14: could you add a "backward" transformation?

src/common/transformations/tests/transpose_sinking/ts_shape_of_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "transformations/transpose_sinking/ts_shape_of.hpp"
+
+#include "common_test_utils/ngraph_test_utils.hpp"
+#include "transformations/transpose_sinking/ts_fuse.hpp"
+
+using namespace ov;
+
+using TSShapeOfForward = TransformationTestsF;
+
+TEST_F(TSShapeOfForward, v0ShapeOf) {
```
itikhono, line 14: you can use /home/tikhonov/OpenVINO/openvino/src/common/transformations/tests/transpose_sinking/ts_common_test.cpp
for simple scenarios

src/common/transformations/include/transformations/transpose_sinking/ts_shape_of.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+}  // namespace transpose_sinking
+}  // namespace pass
+}  // namespace ov
+
+/**
+ * @ingroup ie_transformation_common_api
+ * @brief TSShapeOfForward transformation sinks Transpose through ShapeOf in the forward direction.
```
itikhono, line 22: please mention Gather here, maybe it's better to draw a graph before and after

itikhono (approved):

src/common/transformations/include/transformations/transpose_sinking/ts_shape_of.hpp
mateusztabaka: I think we don't need it.

For example if we spot ShapeOf->Gather and then TSShapeOfBackward adds a new transpose on ShapeOf's input: Transpose->ShapeOf - even we if have another transpose that could be eliminated - we wouldn't be able to do that since the first transpose has two consumers now. But even if we could, at the time TSShapeOfBackward is executed, we have no guarantee that there's another transpose that can be eliminated. So we can be left with Transpose->ShapeOf instead of ShapeOf->Gather - in general the former would be more expensive.

itikhono: what about this case? 
image

PR #19477, opened 2023-08-29
PullReshapeThroughReduce - skip transformation if Reshape doesn't unsqueeze input

mateusztabaka:
Ticket: CVS-118905


src/common/transformations/src/transformations/common_optimizations/pull_through_reduce.cpp
```diff
@@ -198,8 +198,12 @@ ov::pass::PullReshapeThroughReduce::PullReshapeThroughReduce() {
...
             return false;
         }
-        const auto unsqueeze_axes =
-            try_get_unsqueeze_axes_from_reshape(reshape_node->get_shape(), input_node.get_shape());
+        const auto& input_shape = input_node.get_shape();
+        const auto& target_shape = reshape_node->get_shape();
+        if (target_shape.size() <= input_shape.size()) {
+            return false;
+        }
+        const auto unsqueeze_axes = try_get_unsqueeze_axes_from_reshape(target_shape, input_shape);
```
itikhono, line 206: what does  this function return in case of "target_shape.size() <= input_shape.size()"?
maybe it's better to update the function to return {}?

mateusztabaka: > what does this function return in case of "target_shape.size() <= input_shape.size()"?

It depends, but for the case I implemented https://github.com/openvinotoolkit/openvino/pull/19477/files#diff-cd8971be43bbb6771b4a08f53415ae5d428f8728ea0b3f95740fac90cb0b6489R363, it returns {1}. 


> maybe it's better to update the function to return {}?

done

itikhono (approved):

PR #19486, opened 2023-08-29
Refactor BroadcastLayerTest and GRUSequenceTest

olpipi:
### Details:
- Refactor BroadcastLayerTest  and GRUSequenceTest to use SubgraphBaseTest
- Update cpu plugin instantiation of these tests to use new implementation
- Disable GRUSequenceTest  due to fail on arm64. Fix in scope of CVS-120329

### Tickets:
 - CVS-111364


src/tests/functional/shared_test_classes/include/shared_test_classes/single_layer/gru_sequence.hpp
```diff
@@ -43,3 +45,36 @@ class GRUSequenceTest : public testing::WithParamInterface<GRUSequenceParams>,
...
+        //size_t,                                 // input size
+        std::vector<std::string>,                 // activations
+        float,                                    // clip
+        bool,                                     // linear_before_reset
+        ov::op::RecurrentSequenceDirection,       // direction
+        ov::test::utils::InputLayerType,          // WRB input type (Constant or Parameter)
+        ov::element::Type,                        // Network precision
+        std::string>;                             // Device name
+
+class GRUSequenceTestNew : public testing::WithParamInterface<GRUSequenceParams>,
```
ilyachur, line 67: I think you can avoid New postfix, new classes are located in new namespace

src/tests/functional/shared_test_classes/src/subgraph/get_output_before_activation.cpp
```diff
@@ -82,3 +83,92 @@ InferenceEngine::Blob::Ptr OutputBeforeActivation::GenerateInput(const Inference
...
+    case midOutputType::Mul:
+        return (os << "Mul");
+    default:
+        return (os << "Unknown");
+    }
+}
+
+std::string OutputBeforeActivationNew::getTestCaseName(const testing::TestParamInfo<outputBeforeActivationParams> &obj) {
+    std::string targetDevice;
+    ov::element::Type netPrecision;
```
ilyachur, line 104:
```suggestion
    ov::element::Type input_type;
```
We don't have network precision since v10 IR.

src/tests/functional/shared_test_classes/src/subgraph/get_output_before_activation.cpp
```diff
@@ -82,3 +83,92 @@ InferenceEngine::Blob::Ptr OutputBeforeActivation::GenerateInput(const Inference
...
+    size_t inputSize;
+    midOutputType outputType;
+    std::map<std::string, std::string> config;
+    std::tie(targetDevice, netPrecision, inputSize, outputType, config) = obj.param;
+    std::ostringstream result;
+
+    result << "netPrecision=" << netPrecision.get_type_name() << "_";
```
ilyachur, line 111:
```suggestion
    result << "IT=" << netPrecision.get_type_name() << "_";
```
Input Type

src/tests/functional/shared_test_classes/src/subgraph/get_output_before_activation.cpp
```diff
@@ -82,3 +83,92 @@ InferenceEngine::Blob::Ptr OutputBeforeActivation::GenerateInput(const Inference
...
+    }
+    default:
+        GTEST_FAIL() << "Unknown midOutputType";
+    }
+
+    auto act = ngraph::builder::makeActivation(midLayer, netPrecision, ngraph::helpers::ActivationTypes::Tanh);
+    outputs.insert(outputs.end(), {midLayer, act});
+    function = std::make_shared<ov::Model>(outputs, params, "output_before_activation");
+
+    std::vector<ov::test::InputShape> input_shapes;
```
ilyachur, line 158: Why not `ov::Shape`????

olpipi: init_input_shapes requires std::vector<ov::test::InputShape> as parameter

src/tests/functional/shared_test_classes/include/shared_test_classes/single_layer/gru_sequence.hpp
olpipi: It is OK from C++ perspective. But not from Gtest.
Namespace doesn't take part in Gtest test name mangling process. So we have to change these names.

I can just use original name for new class GRUSequenceTest and rename old class GRUSequenceTestLegacy

olpipi: I cannot change legacy test names because these names are still used in openvino_contrib repo

ilyachur: See my comment to `src/plugins/intel_cpu/tests/functional/shared_tests_instances/single_layer_tests/broadcast.cpp`

src/plugins/intel_cpu/tests/functional/shared_tests_instances/single_layer_tests/broadcast.cpp
```diff
@@ -39,24 +39,24 @@ std::vector<std::vector<size_t>> targetShapesNumpy1D = {
...
-        ::testing::Values(ngraph::op::BroadcastType::NUMPY),
+        ::testing::Values(ov::AxisSet{}), //not used in numpy mode
+        ::testing::Values(ov::op::BroadcastType::NUMPY),
         ::testing::Values(std::vector<size_t>{1}),
         ::testing::ValuesIn(inputPrecisions),
         ::testing::Values(ov::test::utils::DEVICE_CPU)
 );
 
-INSTANTIATE_TEST_CASE_P(smoke_TestNumpyBroadcast1D, BroadcastLayerTest, numpyBroadcast1DInputParams, BroadcastLayerTest::getTestCaseName);
+INSTANTIATE_TEST_CASE_P(smoke_TestNumpyBroadcast1D, BroadcastLayerTestNew, numpyBroadcast1DInputParams, BroadcastLayerTestNew::getTestCaseName);
```
ilyachur, line 49: I don't believe that you need to introduce the new class. Here you can use `BroadcastLayerTest` name. In case if gtest has any issues with namespaces I think it already should be solved in line 10 by:
```
using namespace ov::test;
```

olpipi:
There is a logic mistake in using init_input_shapes()
Need to change it

src/plugins/intel_cpu/tests/functional/shared_tests_instances/single_layer_tests/broadcast.cpp
olpipi: Unfortunately, gtest does not want to work with equal names even if I rename testcases:

`
 Attempted redefinition of test suite BroadcastLayerTest.
All tests in the same test suite must use the same test fixture
class.  However, in test suite BroadcastLayerTest, you tried
to define a test using a fixture class different from the one
used earlier. This can happen if the two fixture classes are
from different namespaces and have the same name. You should
probably rename one of the classes to put the tests into different
test suites.
`

olpipi: Gtest doesn't work if there are 2 or more fixtures with equal name (even in different namespace). I tried to rename testcase or testsuite - doesn't help

`Attempted redefinition of test suite BroadcastLayerTest.
All tests in the same test suite must use the same test fixture
class.  However, in test suite BroadcastLayerTest, you tried
to define a test using a fixture class different from the one
used earlier. This can happen if the two fixture classes are
from different namespaces and have the same name. You should
probably rename one of the classes to put the tests into different
test suites.`

olpipi: We can rename it back after all plugins migrate from old shared_test_classes to new ones

olpipi:
All issues are fixed

src/plugins/intel_cpu/tests/functional/shared_tests_instances/single_layer_tests/broadcast.cpp
```diff
@@ -85,35 +91,46 @@ std::vector<std::vector<size_t>> targetShapesNumpy3D = {
...
+        ::testing::Values(ov::AxisSet{}), //not used in numpy mode
+        ::testing::Values(ov::op::BroadcastType::NUMPY),
+        ::testing::ValuesIn(ov::test::static_shapes_to_test_representation(input_shapes_3d_static)),
         ::testing::ValuesIn(inputPrecisions),
         ::testing::Values(ov::test::utils::DEVICE_CPU)
 );
 
 INSTANTIATE_TEST_CASE_P(smoke_TestNumpyBroadcast3D, BroadcastLayerTest, numpyBroadcast3DInputParams, BroadcastLayerTest::getTestCaseName);
 
+const std::vector<std::vector<ov::Shape>> ngraph_evaluate_shapes_static = {
```
ilyachur, line 109: Please remove `ngraph` word

ilyachur (approved):
In general LGTM, one common comment: Please also move template tests to new classes where it is applicable

ilyachur:
@olpipi Please fix code style

iefode (approved):

PR #19492, opened 2023-08-30
Override Slice::constant_folding to allow to extract static dimensions from ShapeOf subgraphs

mateusztabaka:
Some subgraphs like ShapeOf->StridedSlice/Slice can be constantfolded if slicing is done on static dimensions. We have this functionality in StridedSlice::constant_fold, but it's missing in Slice.

Ticket: CVS-118902

### DIFF
--- a/src/core/include/openvino/op/slice.hpp
+++ b/src/core/include/openvino/op/slice.hpp
@@ -51,6 +51,7 @@ public:
     bool evaluate_lower(TensorVector& outputs) const override;
     bool evaluate_upper(TensorVector& outputs) const override;
     bool evaluate_label(TensorLabelVector& output_labels) const override;
+    bool constant_fold(OutputVector& output_values, const OutputVector& input_values) override;
 
     std::shared_ptr<v0::Constant> get_default_const_axes(const Output<Node>& start) const;
 };
--- a/src/core/src/op/slice.cpp
+++ b/src/core/src/op/slice.cpp
@@ -11,6 +11,8 @@
 #include "ngraph/attribute_visitor.hpp"
 #include "ngraph/graph_util.hpp"
 #include "ngraph/op/constant.hpp"
+#include "openvino/core/rt_info.hpp"
+#include "openvino/pass/constant_folding.hpp"
 #include "openvino/reference/slice.hpp"
 #include "slice_shape_inference.hpp"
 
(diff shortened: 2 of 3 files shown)


itikhono (approved):

jane-intel:
We already have this functionality. Why do we invent the wheel?

jane-intel:
We already have this functionality. Why do we reinvent the wheel?

itikhono:
> 

Wheel was invented for strided slice already
image
We just re-used that

jane-intel (changes requested):
I suggest taking different approach.
This sub-graph comes to some operation which should trigger partial value propagation.
Since it didn't fold -- it means some part of the graph couldn't propagate through. So I propose to search for the faulty op(s) instead of doing this temporary fix.

praasz:
> > 
> 
> The wheel was invented for strided slice already image We just re-used that @praasz do you remember any details?

Just short description for strided slice. The custom const fold check inputs 1, 2 (start, end) if used values have const value (no interval) and then these inputs are replaced by constants. 
Then strided slice sub-graph is check if its output can replaced by constant (abort if any node in this sub-graph has const folding disabled).

I think test case ` slice_shape_of_with_dynamic_shape_on_static_dimension`  shows the issue. 
There are input data for slice but it has got some interval values, so without proposed modification the slice cannot be folded even if selected values for slice are not interval values `2, 3`. If in this test `param` will be replaced to have only only static shape then slice will fold.

I think similar case can be reproduced for strided slice as well, as the input 0 is not checked in this way.

@jane-intel 
Maybe the slice and strided slice should have same feature to check data input. 
If sliced values has no interval values (no const fold disabled) then it should be possible to fold this operator.

praasz:
> > > 
> > 
> > 
> > The wheel was invented for strided slice already image We just re-used that @praasz do you remember any details?
> 
> Just short description for strided slice. The custom const fold check inputs 1, 2 (start, end) if used values have const value (no interval) and then these inputs are replaced by constants. Then strided slice sub-graph is check if its output can replaced by constant (abort if any node in this sub-graph has const folding disabled).
> 
> I think test case ` slice_shape_of_with_dynamic_shape_on_static_dimension` shows the issue. There are input data for slice but it has got some interval values, so without proposed modification the slice cannot be folded even if selected values for slice are not interval values `2, 3`. If in this test `param` will be replaced to have only only static shape then slice will fold.
> 
> I think similar case can be reproduced for strided slice as well, as the input 0 is not checked in this way.
> 
> @jane-intel Maybe the slice and strided slice should have same feature to check data input. If sliced values has no interval values (no const fold disabled) then it should be possible to fold this operator.

I've just checked the strided slice with similar case as `slice_shape_of_with_dynamic_shape_on_static_dimension`, it looks like it works correctly and no modification is required.

mateusztabaka:
> I suggest taking different approach. This sub-graph comes to some operation which should trigger partial value propagation. Since it didn't fold -- it means some part of the graph couldn't propagate through. So I propose to search for the faulty op(s) instead of doing this temporary fix.

So the scenario (in pseudocode) looks like follows:

```
param = Parameter(shape={-1, 2, 3, -1})
shapeof = ShapeOf(param)
slice = Slice(shapeof, start=[1], stop=[3], stride=[1], axis=[0])
model = Model(slice, [param])
manager.register_pass<ConstantFolding>()
manager.run_passes(model)
```

The Slice was not constantfolded because ShapeOf was not constantfolded and this is due its input shape being dynamic.
But the Slice can be constantfolded because it slices the part of the shape that is static.
Also if you use StridedSlice instead of Slice, this subgraph can be constfolded since [StridedSlice has custom constant_fold function](https://github.com/openvinotoolkit/openvino/blob/master/src/core/src/op/strided_slice.cpp#L302).

mateusztabaka:
I think I should expand on the actual scenario. In the model I'm working on, some ShapeOf subgraphs are not used for shape inference - they're the inputs to some compute layers in the model. So lower and upper bound is never set for such ShapeOfs. Therefore those subgraphs cannot be folded by ConstantFolding.

praasz (approved):

jane-intel:
Since Slice has axis attribute -- I agree to introduce constant_fold for it. However, I would rather have better evaluate_lower / upper for it, so that output of this Slice would have upper==lower and ConstantFolding pass would fold it without Slice::constant_fold change

PR #19495, opened 2023-08-30
Handle negative values in GroupedSliceToVSplitOptimization

mateusztabaka:
CVS-118897


src/common/transformations/src/transformations/common_optimizations/optimize_strided_slice.cpp
```diff
@@ -274,30 +274,47 @@ struct SliceWithAttrs {
...
+        int64_t value;
+        if (!get_scalar(op->get_input_node_shared_ptr(i), value))
             return false;
-        else if (i == 1)
+        if (i == 3) {
+            if (value != 1)
+                return false;
+            continue;
+        }
+        value = value >= 0 ? value : value + dimension;
```
itikhono, line 310: I think the previous implementation had better readability.
could we use "get_scalar(" lambda to process negative values?
I think we need to pass "rank" or "dimension" to this lambda as an additional argument
maybe somethink like that
vector<size_t> right_border_values = {dim, dim, dim, dim, rank};
after I hope we will be able to handle all inputs in one loop.
Could you also use unnamed enum instead of indexes for inputs?
enum { BEGIN = 1, END, STRIDE, AXIS };
if (BEGIN == i) etc

mateusztabaka: > could we use "get_scalar(" lambda to process negative values?
> I think we need to pass "rank" or "dimension" to this lambda as an additional argument
> maybe somethink like that
> vector<size_t> right_border_values = {dim, dim, dim, dim, rank};
> get_scalar(..., right_border_values[i])


I'd prefer not to, since we can incorrectly handle stride then. For example stride == -1 (which is not suitable for optimization), dim == 2 and after get_scalar, stride == 1 (which now is suitable for optimization). 
Also we need to know axis first to get the dimension.

But what do you think of something like this:
```
    enum {
        START = 1,
        STOP,
        STRIDE,
        AXIS
    };

    int64_t stride;
    if (!get_scalar(op->get_input_node_shared_ptr(STRIDE), stride) || stride != 1)
        return false;
    if (!get_scalar(op->get_input_node_shared_ptr(AXIS), attrs.axis))
        return false;
    attrs.axis = attrs.axis >= 0 ? attrs.axis : attrs.axis + rank;

    if (input_shape[attrs.axis].is_dynamic())
        return false;
    const auto dimension = input_shape[attrs.axis].get_length();

    for (int i = START; i <= STOP; i++) {
        int64_t value;
        if (!get_scalar(op->get_input_node_shared_ptr(i), value))
            return false;
        value = value >= 0 ? value : value + dimension;
        value = std::min(value, dimension);
        if (i == START)
            attrs.start = value;
        else if (i == STOP)
            attrs.stop = value;
    }
```

itikhono: we can use 0 as a right border for STRIDE for example
value + 0

itikhono: it looks like we need to get axis before the loop in any case
the current impl looks good

itikhono (approved):

src/common/transformations/src/transformations/common_optimizations/optimize_strided_slice.cpp
```diff
@@ -274,30 +274,47 @@ struct SliceWithAttrs {
...
+    const auto dimension = input_shape[attrs.axis].get_length();
 
-        if (((i == 1 || i == 2) && value < 0) || (i == 3 && value != 1))
+    for (int i = START; i <= STOP; i++) {
+        int64_t value;
+        if (!get_scalar(op->get_input_node_shared_ptr(i), value))
             return false;
-        else if (i == 1)
+        value = value >= 0 ? value : value + dimension;
+        value = std::min(value, dimension);
```
jane-intel, line 311: In case my dimension is 10 and start is -100 what would happen here? I think we still need to have the check for value being negative.

itikhono: Good point! it looks like we need to clamp it first
```
value = max(min(value, dimension), -dimension); // maybe max(..., -dimension +1) ?
value = value >= 0 ? value : value + dimension;
```
@mateusztabaka could you check this?

mateusztabaka: yes, we should clamp the value

itikhono (approved):

jane-intel (approved):

PR #19504, opened 2023-08-30
Get rid of fp16/bf16 evaluates

itikhono:
### Details:
Get rid of fp16/bf16 evaluates

### Tickets:
 - *CVS-109239*


ilyachur:
@itikhono Please take a look to the PR: https://github.com/openvinotoolkit/openvino/pull/19830/

In this PR I did the same experiments, but looks like we need to fix some transformations first of all. Right now I have an issue inside the LPT tests, maybe we need to add transformation which will be able to add f16 conversion before the multiply

CC @p-durandin

PR #19505, opened 2023-08-30
[DOCS] Fix for Install from Docker Image for master

msmykx-intel:
- Removed duplicated note
- Fixed the link for Bare Metal

### DIFF
--- a/docs/install_guides/installing-openvino-docker-linux.md
+++ b/docs/install_guides/installing-openvino-docker-linux.md
@@ -22,12 +22,7 @@ You can get started easily with pre-built and published docker images, which are
 * `Red Hat Ecosystem Catalog (development image) <https://catalog.redhat.com/software/containers/intel/openvino-dev/613a450dc9bc35f21dc4a1f7>`__
 * `Azure Marketplace <https://azuremarketplace.microsoft.com/en-us/marketplace/apps/intel_corporation.openvino>`__
 
-.. note::
-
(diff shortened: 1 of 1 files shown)


docs/install_guides/installing-openvino-docker-linux.md
```diff
@@ -37,11 +32,11 @@ To start using Dockerfiles the following conditions must be met:
 
 .. note::
 
-   OpenVINO's `Docker <https://docs.docker.com/>`__ and `Bare Metal <https://docs.openvino.ai/2023.0/ovms_docs_deploying_server.html#doxid-ovms-docs-deploying-server>`__ distributions are identical, so the documentation applies to both. 
+   OpenVINO's `Docker <https://docs.docker.com/>`__ and `Bare Metal <https://docs.openvino.ai/2023.0/ovms_docs_deploying_server.html#deploying-model-server-on-baremetal-without-container>`__ distributions are identical, so the documentation applies to both.
```
ngaloppo, line 35: Since this is documentation about installing OpenVINO native tools, I don't understand why we're linking to OVMS documentation. IMO a link to the OV dev/runtime bare metal installations would be a better solution here: 

https://docs.openvino.ai/2023.0/openvino_docs_install_guides_overview.html

msmykx-intel: The link here is for the user to get information about Bare Metal in OpenVINO, and since we have information about it in OVMS for now, that's why I put it there. If You think it is not necessary to have it in here I can remove it.

ngaloppo: In my opinion the link I provided above is a better reference to information about bare metal installation than the OVMS link for bare metal.

msmykx-intel: Changed in https://github.com/openvinotoolkit/openvino/pull/19505/commits/996f027c571d8827f7d4a63fc5865681f00a398a

ngaloppo (approved):

sgolebiewski-intel (approved):

PR #19508, opened 2023-08-30
Do not clean batch setting if proxy plugin

songbell:
### Details:
 - fix ov::hint::allow_auto_batching(false) does not take into effect for GPU

### Tickets:
 - *ticket-id*


src/inference/src/dev/core_impl.cpp
```diff
@@ -1095,8 +1095,9 @@ std::shared_ptr<const ov::Model> ov::CoreImpl::apply_auto_batching(const std::sh
             const auto disabled = batch_mode->second.as<std::string>() == CONFIG_VALUE(NO);
             // virtual plugins like AUTO/MULTI will need the config
             // e.g. to deduce the #requests correctly
+            // proxy plugin should also keep the config
             // otherwise, no need for this config key in the rest of loading
-            if (!is_virtual_device(deviceName))
+            if (!is_virtual_device(deviceName) && !is_proxy_device(deviceName))
```
ilyachur, line 1100: For proxy plugin we have a check in the line 1109.

Could you please add new tests? 
Why do we need these changes?

songbell: does this mean with proxy enabled, currently GPU will not enable auto batch (no matter which batch config applied)?
e.g user compiles to GPU with throughput hint, no auto batching will be checked/applied if proxy is there?

ilyachur: Why do you think that GPU doesn't work with auto batch?
Right now all tests on GPU with auto batch are passed.
We add auto batch between proxy and GPU:
Proxy->Batch->GPU. It should work for the master branch

songbell: sorry, let me share the case more clearly:
considering below case：
core.compile_model("GPU", ov::hint::performance_mode("THROUGHPUT"), **ov::allow_auto_batching(false)**)；
above expect batch to be turned off
yet on current master with proxy plugin enabled, it's not the case
you can try below command: ./benchmark_app -m ~/work/mount/mobilenet-v2/onnx/onnx/FP16/INT8/mobilenet-v2.xml -d GPU -b 1
here -b 1 will add ov::allow_auto_batching(false) to compile property
but still get batch compiled model:
[ INFO ]   NETWORK_NAME: torch-jit-export
[ INFO ]   SUPPORTED_CONFIG_KEYS: AUTO_BATCH_TIMEOUT
[ INFO ]   EXECUTION_DEVICES: OCL_GPU.0
[ INFO ]   AUTO_BATCH_TIMEOUT: 1000

ilyachur: @songbell Could you add a test on this behaviour?

src/plugins/proxy/tests/batch_compliance_test.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include "openvino/proxy/properties.hpp"
+#include "openvino/runtime/properties.hpp"
+#include "proxy_tests.hpp"
+
+using namespace ov::proxy::tests;
+
+TEST_F(ProxyTests, can_parse_and_inherit_batch_property) {
+    register_plugin_support_reshape(core,
+                                        "MOCK_DEVICE",
```
songbell, line 13: added test.
ilya, @ilyachur  another question, if alias name not same as real device name, e.g register_plugin("real_name", "alias_name")
is_proxy_device(real_name) will report false in this case? because only alias name updated in registy:
pluginRegistry[alias] = desc;
https://github.com/openvinotoolkit/openvino/blob/master/src/inference/src/dev/core_impl.cpp#L411C18-L411C18

src/inference/src/dev/core_impl.cpp
songbell: update tests

src/plugins/proxy/tests/batch_compliance_test.cpp
ilyachur: Proxy is a virtual plugin and we use only alias to operate with this plugin

ilyachur (approved):

src/plugins/proxy/tests/batch_compliance_test.cpp
songbell: > Proxy is a virtual plugin and we use only alias to operate with this plugin

so if in future we have another plugin with device name = "VPU", alias name = "NPU" for example, we might have problem again?
because  is_proxy_plugin("VPU") will report false and the allow_auto_batch property will be erased again?

songbell: @yangwang201911

PR #19516, opened 2023-08-31
[GPU] Add permute primitive instead of manual copy for deconv weights

vladimir-paramuzov:
### Details:
 - Insert explicit permute primitive for deconv weights in all cases to simplify Constant op handling

### DIFF
--- a/src/plugins/intel_gpu/src/plugin/ops/constant.cpp
+++ b/src/plugins/intel_gpu/src/plugin/ops/constant.cpp
@@ -65,39 +65,59 @@ static cldnn::tensor getConstTensor(const ov::Shape constDims) {
 
 struct ConstProperties {
     bool needsBatchInterpretation;
-    bool swapOI;
-    bool hasGroupDimension;
 };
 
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/plugin/ops/constant.cpp
```diff
@@ -64,40 +64,61 @@ static cldnn::tensor getConstTensor(const ov::Shape constDims) {
...
+        p.profiling_ids.push_back(initialconstPrimID);
+    } else {
+        GPU_DEBUG_LOG << "[" << initialconstPrimID << ": constant]" << std::endl;
+        cldnn::memory::ptr mem = p.get_engine().allocate_memory(constLayout, false);
+        auto& stream = p.get_engine().get_service_stream();
+        cldnn::mem_lock<char> lock{mem, stream};
+        auto buf = lock.data();
+        auto bufSize = constLayout.bytes_count();
+
+        // Do actual weights reorder and change O and I channels order
```
sshlyapn, line 106: Looks like this comment is not needed anymore

sshlyapn (approved):

PR #19517, opened 2023-08-31
Restored opset1::Reshape label propagation for -1 special value

jane-intel:
### Details:
 - *Auto-batch feature got turned off for models with Reshape operation which has -1 value in it*
 - *I estimate that performance of high number of models is impacted on async GPU inference*

### Tickets:
 - *To be updated after @songbell confirmation*

The check for `ov::DimensionTracker::get_label(dim)` is restored in the shape inference of `opset1::Reshape` operation since `FindBatch` transformation sets original static shape back to the model with only batch being labeled. This way batch has static value and has to propagate label through the `opset1::Reshape`

When batch is not yet restored to the static value previous code worked as expected:
image

However, after we restored it to the static dimension ([here](https://github.com/openvinotoolkit/openvino/blob/master/src/common/transformations/src/transformations/common_optimizations/dimension_tracking.cpp#L294)) shape inference filed to propagate the label through:
image

The issue existed since April 10th (https://github.com/openvinotoolkit/openvino/pull/15853/files)


src/core/tests/type_prop/reshape.cpp
```diff
@@ -20,6 +20,18 @@ TEST(type_prop, static_value_propagation) {
     ASSERT_EQ(r->get_shape(), (Shape{1, 2, 3}));
 }
 
+TEST(type_prop, tricky_label_propagation_for_auto_batch_case) {
+    auto shape = ov::PartialShape({1, 1280, 1, 1});
+    ov::DimensionTracker::set_label(shape[0], 1);
```
praasz, line 25: If labels will be set on dimensions e.g. 0 and 2 (both 1), which label will passed to output, or will it be lost?

praasz (approved):

src/core/tests/type_prop/reshape.cpp
```diff
@@ -20,6 +20,18 @@ TEST(type_prop, static_value_propagation) {
...
+    auto param = make_shared<op::Parameter>(element::f32, shape);
+    auto pattern = op::v0::Constant::create(element::i64, {2}, {-1, 1280});
+    auto r = make_shared<op::v1::Reshape>(param, pattern, false);
+
+    ASSERT_EQ(r->get_element_type(), element::f32);
+    ASSERT_EQ(r->get_shape(), (Shape{1, 1280}));
+    ASSERT_EQ(1, ov::DimensionTracker::get_label(r->get_output_partial_shape(0)[0]));
+}
+
```
praasz, line 34: Consider somethin similar to:
```suggestion
TEST(type_prop, reshape_label_propagation_minus_one_special_zero_case_1) {
    auto data_shape = PartialShape{4, -1, 2, 1, 3};
    set_shape_labels(data_shape, 10);

    auto input = make_shared<op::v0::Parameter>(element::f32, data_shape);
    auto output_pattern = make_shared<op::v0::Constant>(element::i64, Shape{3}, std::vector<int64_t>{-1, 12, 0});
    output_pattern->get_default_output().get_tensor().set_value_label({20, 21, 22});

    const auto reshape = make_shared<op::v1::Reshape>(input, output_pattern, true);

    auto output_shape = reshape->get_output_partial_shape(0);
    EXPECT_EQ(output_shape, PartialShape({-1, 12, 2}));
    EXPECT_THAT(get_shape_labels(output_shape), ElementsAre(ov::no_label, 21, 12));
}
```

With previous version the labels on output were, losing label will be ok now?
```    EXPECT_THAT(get_shape_labels(output_shape), ElementsAre(11, 21, 12));```

jane-intel:
@praasz I have changed the approach to the previous one. Having static dimension with label means one of two things: it was labeled by mistake, or our dynamic dimension which we labeled at the beginning of the model become static during shape propagation which means this dimension is restricted to the only static value (this can happen with channel dimension when it faces Convolution for example). Since this dimension is restricted -- then it could not be changed in any way without causing shape inference issues -- therefore we can consider it a constant static dimension of this model and operate as if there is no label on it.

songbell (approved):

praasz (approved):

wenjiew:
Should we go through the process to merge this to 2023.1 branch?

jane-intel:
Hello @wenjiew, I cannot provide validation proof that this fix has to go in the release -- this is why I didn't proceed with committee approval.

wenjiew:
@peterchen-intel @jane-intel If it is still the validation report blocking the merge, is there anything we can help from the internal testing side? Thanks!

jane-intel:
Closing since 23.1.1 is not yet confirmed.

PR #19519, opened 2023-08-31
[CI] [GHA] Specify save and restore key name for the Linux pipeline

akashchi:
### Details:
This PR specifies the cache key for the Linux pipeline in order to avoid confusion with cache entries from other pipelines.

This way, the key for the Linux pipeline would be `ccache-linux-ubuntu-*timestamp*`.


ilya-lavrenov (approved):

mryzhov (approved):

.github/workflows/linux.yml
```diff
@@ -129,9 +129,9 @@ jobs:
           # github.ref_name is 'ref/PR_#' in case of the PR, and 'branch_name' when executed on push
           save: ${{ github.ref_name == 'master' && 'true' || 'false'  }}
           verbose: 2
-          key: ${{ github.job }}-linux
+          key: linux-ubuntu
```
mryzhov, line 132: just an option, we can use runner's info:
cache-${{ runner.os }}-${{ steps.system-info.outputs.platform}}

akashchi: All the GHA Linux pipelines are using the same runner group (`ubuntu-20.04-*`) but have different `cmake` commands and/or install different packages (building for different OSes/arches) so the `${{ runner.os }}` would be the same for all current/potential GHA pipelines.

PR #19527, opened 2023-09-01
[CPU] Eliminate dependency for emitters tails process

chenhu-wang:
### Details:
 - *unblock RAW dependency for loop of tails process*
 - *1/2/3/4 byte tail number is unblocked previously, this pr extend to all tail number*

### Tickets:
 - *CVS-119576*

### DIFF
--- a/src/plugins/intel_cpu/src/emitters/x64/jit_load_store_emitters.cpp
+++ b/src/plugins/intel_cpu/src/emitters/x64/jit_load_store_emitters.cpp
@@ -234,11 +234,13 @@ void jit_load_emitter::load_bytes(const Vmm &vmm, const Xbyak::Reg64 &reg, int o
             has_xmm_block = true;
         }
 
+        if (!one_of(bytes_to_load, 0, 1, 2, 3, 4, 8, 16)) {
+            h->uni_vpxor(xmm, xmm, xmm);
+        }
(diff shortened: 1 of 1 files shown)


chenhu-wang:

dmitry-gorokhov:
@a-sidorova Please take a look

a-sidorova (approved):

src/plugins/intel_cpu/src/emitters/x64/jit_load_store_emitters.cpp
```diff
@@ -234,11 +234,13 @@ void jit_load_emitter::load_bytes(const Vmm &vmm, const Xbyak::Reg64 &reg, int o
             has_xmm_block = true;
         }
 
+        if (!one_of(bytes_to_load, 0, 1, 2, 3, 4, 8, 16)) {
```
dmitry-gorokhov, line 237: That would be awesome if you can write a comment why we need `vpxor` here and why it is inserted only for some byte sizes.

chenhu-wang: @dmitry-gorokhov the explanation comment is added.

dmitry-gorokhov (approved):

PR #19539, opened 2023-09-01
[GPU] Add FullyConnected internal op and some related transformations

vladimir-paramuzov:
### Details:
 - Added internal FullyConnected op and MatMul -> FC transformation 
 - Added transformation which removed Reshape op in compressed weights pattern
 - Enabled transpose fusion into matmul (by changing matmul transpose flags)
 - Added test for compressed Matmul

### Tickets:
 -  Part of *118099*

### DIFF
--- a/src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
@@ -3,7 +3,7 @@
 //
 
 #pragma once
-#include <string>
+#include <cstring>
 #include <mutex>
 #include <vector>
 #include <set>
--- a/src/plugins/intel_gpu/src/plugin/transformations/convert_matmul_to_fc.cpp
+++ b/src/plugins/intel_gpu/src/plugin/transformations/convert_matmul_to_fc.cpp
@@ -1,0 +1,172 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "intel_gpu/op/fully_connected.hpp"
+#include "convert_matmul_to_fc.hpp"
+#include "openvino/op/matmul.hpp"
+#include "openvino/op/convert.hpp"
+#include "openvino/op/transpose.hpp"
+#include "openvino/op/reshape.hpp"
+#include "openvino/core/rt_info.hpp"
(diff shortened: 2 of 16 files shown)


src/plugins/intel_gpu/include/intel_gpu/runtime/debug_configuration.hpp
```diff
@@ -3,7 +3,7 @@
 //
 
 #pragma once
-#include <string>
+#include <string.h>
```
sshlyapn, line 6: Shouldn't it be `<cstring>` instead?

src/plugins/intel_gpu/src/plugin/transformations/convert_matmul_to_fc.cpp
```diff
@@ -0,0 +1,172 @@
...
+        bool success = true;
+        ov::PartialShape shape_a_aligned;
+        ov::PartialShape shape_b_aligned;
+        std::tie(success, shape_a_aligned, shape_b_aligned) = get_aligned_shapes();
+        if (!success) {
+            return false;
+        }
+
+        auto aligned_a_rank = shape_a_aligned.rank();
+        auto aligned_b_rank = shape_a_aligned.rank();
```
sshlyapn, line 135:
```suggestion
        auto aligned_b_rank = shape_b_aligned.rank();
```
But I'm not sure we need this check at all, because we already have `OPENVINO_ASSERT(shape_b.is_static());` aboce

src/plugins/intel_gpu/src/plugin/transformations/convert_matmul_to_fc.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "openvino/core/rt_info.hpp"
+#include "openvino/pass/pattern/op/wrap_type.hpp"
+#include "transformations/utils/utils.hpp"
+
+namespace ov {
+namespace intel_gpu {
+
+ConvertMatMulToFullyConnected::ConvertMatMulToFullyConnected() {
+    auto static_rank_gt_1 = [](const ov::Output<ov::Node>& output) {
+        const auto& r =  output.get_partial_shape().rank();
```
sshlyapn, line 20: nit:
```suggestion
        const auto& r = output.get_partial_shape().rank();
```

sshlyapn (approved):

PR #19550, opened 2023-09-01
binding pcore for stream calculation

xufang-lisa:
### Details:
 - *binding pcore for stream calculation to ensure that the streams number is consistent on hybrid cpu*

### Tickets:
 - *121959*


src/plugins/intel_cpu/src/cpu_streams_calculation.cpp
```diff
@@ -520,11 +522,33 @@ std::vector<std::vector<int>> generate_stream_info(const int streams,
...
-void get_num_streams(const int streams, const std::shared_ptr<ngraph::Function>& ngraphFunc, Config& config) {
+void get_num_streams(const int streams,
+                     const std::shared_ptr<ngraph::Function>& ngraphFunc,
+                     Config& config,
+                     const std::shared_ptr<InferenceEngine::IInferencePlugin>& plugin) {
     InferenceEngine::IStreamsExecutor::Config& executor_config = config.streamExecutorConfig;
     std::vector<std::vector<int>> proc_type_table = get_proc_type_table();
 
-    generate_stream_info(streams, ngraphFunc, config, proc_type_table);
+    InferenceEngine::IStreamsExecutor::Config streamsConfig;
```
riverlijunjie, line 532: Could we use MT2.0 interfaces?
ov::threading::IStreamsExecutor::Config

src/plugins/intel_cpu/src/onednn/dnnl.cpp
```diff
@@ -148,5 +148,35 @@ unsigned get_cache_size(int level, bool per_core) {
     DNNL_THROW_ERROR(dnnl_unimplemented, "get_cache_size has no mode per_core == false");
 }
 
+unsigned get_cache_size_for_current_core(int level, bool per_core) {
```
riverlijunjie, line 151: This function is very similar with `dnnl::impl::cpu::platform::get_per_core_cache_size` and  `dnnl::utils::get_cache_size()`, maybe mix them or other.
So could we implement this `get_cache_size_for_current_core` by calling above dnnl function?

wangleis (approved):

src/plugins/intel_cpu/src/cpu_streams_calculation.cpp
xufang-lisa: updated

src/plugins/intel_cpu/src/onednn/dnnl.cpp
xufang-lisa: The dnnl function uses a static variable to get cpu cache size, and the initialization of this static variable can not be guaranteed to be performed on Pcore. So here we reacquire the cpu information in the current thread.

src/plugins/intel_cpu/src/cpu_streams_calculation.cpp
```diff
@@ -520,11 +547,33 @@ std::vector<std::vector<int>> generate_stream_info(const int streams,
...
+    ov::threading::IStreamsExecutor::Config streamsConfig;
+    streamsConfig._name = "CPUStreamsExecutor";
+    streamsConfig._streams = 1;
+    streamsConfig._threads = 1;
+    if (proc_type_table[0][MAIN_CORE_PROC] > 0) {
+        streamsConfig._streams_info_table.push_back({1, MAIN_CORE_PROC, 1, 0, 0});
+    }
+
+    if (streamsConfig._streams_info_table.empty()) {
+        generate_stream_info(streams, ngraphFunc, config, proc_type_table);
```
sunxiaoxia2022, line 566: I think it should not binding core when there is no Ecore.

xufang-lisa: updated

sunxiaoxia2022 (approved):

riverlijunjie (approved):
LGTM

xufang-lisa:
The changes in this PR are included in https://github.com/openvinotoolkit/openvino/pull/20814

wangleis:
Will extend this PR to include https://github.com/openvinotoolkit/openvino/pull/20814 for issue CVS-121959.

Will cover Pcore only and Ecore only cases on hybrid platform as well

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -171,6 +171,24 @@ Engine::Engine() :
...
+    if (proc_type_table[0][MAIN_CORE_PROC] > 0 && proc_type_table[0][EFFICIENT_CORE_PROC] > 0) {
+        streamsConfig._streams_info_table.push_back({1, MAIN_CORE_PROC, 1, 0, 0});
+    }
+    if (!streamsConfig._streams_info_table.empty()) {
+        std::shared_ptr<ov::threading::ITaskExecutor> taskExecutor =
+            std::make_shared<ov::threading::CPUStreamsExecutor>(streamsConfig);
+        std::vector<Task> tasks;
+        tasks.emplace_back([&] {
+            dnnl::impl::cpu::x64::cpu();
+        });
```
praasz, line 189:
```suggestion
        std::vector<Task> tasks{dnnl::impl::cpu::x64::cpu};
```
If dnnl::impl::cpu::x64::cpu is free function.

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -171,6 +171,24 @@ Engine::Engine() :
...
+    streamsConfig._name = "CPUStreamsExecutor";
+    streamsConfig._streams = 1;
+    streamsConfig._threads = 1;
+    //Initialize Xbyak::util::Cpu object on Pcore for hybrid cores machine
+    if (proc_type_table[0][MAIN_CORE_PROC] > 0 && proc_type_table[0][EFFICIENT_CORE_PROC] > 0) {
+        streamsConfig._streams_info_table.push_back({1, MAIN_CORE_PROC, 1, 0, 0});
+    }
+    if (!streamsConfig._streams_info_table.empty()) {
+        std::shared_ptr<ov::threading::ITaskExecutor> taskExecutor =
+            std::make_shared<ov::threading::CPUStreamsExecutor>(streamsConfig);
```
praasz, line 185:
```suggestion
        auto taskExecutor = std::make_shared<ov::threading::CPUStreamsExecutor>(streamsConfig);
```

src/plugins/intel_cpu/src/plugin.cpp
```diff
@@ -171,6 +171,24 @@ Engine::Engine() :
     deviceFullName(getDeviceFullName()),
     specialSetup(new CPUSpecialSetup) {
     set_device_name("CPU");
+    auto proc_type_table = get_proc_type_table();
```
dmitry-gorokhov, line 174: As I mentioned on the sync that makes a perfect sense to extract task executor creation logic into separate function (ideally places in threading interface, not the plugin). Function signature might take core type and std::fucntion. Then from plugin perspective it might be one-line usage (or smt similar):
```
execute_task(PCORE, []{dnnl::impl::cpu::x64::cpu()});
```

src/plugins/intel_cpu/src/plugin.cpp
dmitry-gorokhov, line 1: We also discussed whole model compilation proccess assigned to Pcore (to guarantee correct streams calculation). Is it still required?

src/plugins/intel_cpu/src/plugin.cpp
xufang-lisa: updated

src/plugins/intel_cpu/src/plugin.cpp
xufang-lisa: The function was extracted and was placed in class ExecutorManager.

src/plugins/intel_cpu/src/plugin.cpp
xufang-lisa: I think that initializing Xbyak::util::Cpu object here can guarantee correct streams calculation.

dmitry-gorokhov (approved):

praasz (approved):

PR #19568, opened 2023-09-04
bug fix update

ahnyoung-paul:
### Details:
 - *Check output shape after op.is_dynamic()*
 - *Fix zero dimension allocation, set {1} dims to zero dimension shape*
 - *Add debug log for condition operation*

### Tickets:
 - *116928*

### DIFF
--- a/src/plugins/intel_gpu/src/plugin/ops/constant.cpp
+++ b/src/plugins/intel_gpu/src/plugin/ops/constant.cpp
@@ -97,6 +97,11 @@ static void create_data(ProgramBuilder& p, const ov::Shape& constDims, const std
         p.primitive_ids[initialconstPrimID] = constPrimID;
         p.profiling_ids.push_back(initialconstPrimID);
     } else {
+        if (constLayout.is_static() && constLayout.count() == 0) {
+            // Convert zero dimension constant layout to 1 dimension to fix the issue
+            // that memory allocation is failed on windows when constant layout is zero dimension.
+            constLayout = cldnn::layout(ov::PartialShape({1}), constLayout.data_type, constLayout.format);
+        }
         cldnn::memory::ptr mem = p.get_engine().allocate_memory(constLayout, false);
         GPU_DEBUG_LOG << "[" << initialconstPrimID << ": constant] layout: "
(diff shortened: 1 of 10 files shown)


src/plugins/intel_gpu/src/plugin/ops/constant.cpp
```diff
@@ -97,8 +97,14 @@ static void create_data(ProgramBuilder& p, const ov::Shape& constDims, const std
         p.primitive_ids[initialconstPrimID] = constPrimID;
         p.profiling_ids.push_back(initialconstPrimID);
     } else {
-        GPU_DEBUG_LOG << "[" << initialconstPrimID << ": constant]" << std::endl;
+        if (constLayout.is_static() && constLayout.count() == 0) {
```
yeonbok, line 100: I think constLayout should always static ?

src/plugins/intel_gpu/src/plugin/ops/split.cpp
```diff
@@ -27,6 +27,13 @@ static void CreateCommonSplitOp(ProgramBuilder& p, const std::shared_ptr<ov::Nod
             ov::Shape start_offset(input_pshape.size());
             for (size_t i = 0; i < op->get_output_size(); i++) {
                 const auto outPartialShape = op->get_output_partial_shape(i);
+                // op->is_dynamic() does not check if output shape is dynamic. it only check dynamism for input shapes
+                // Even if op->is_dynamic() is false, output shape can be dynamic.
+                // Thus, it is necessary to check if output shape is dynamic.
+                if (outPartialShape.is_dynamic()) {
+                    offsets.clear();
+                    break;
```
yeonbok, line 35: Shoundn't we prevent to go into the loop (at line28) at all and make it to proceed to the dynamic pass?

src/plugins/intel_gpu/src/plugin/ops/constant.cpp
ahnyoung-paul: Fixed it. thanks

src/plugins/intel_gpu/src/plugin/ops/split.cpp
ahnyoung-paul: Fixed it. thanks.

yeonbok (approved):

PR #19569, opened 2023-09-04
[GPU] FC weights reorder's output uses 2d-reshaped layout at kernel_params

kelvinchoi-intel:
### Details:
 - *FC weights reorder's output uses 2d-reshaped layout at kernel_params*

### Tickets:
 - *118372*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/graph_optimizer/post_optimize_weights.cpp
+++ b/src/plugins/intel_gpu/src/graph/graph_optimizer/post_optimize_weights.cpp
@@ -92,9 +92,16 @@ void post_optimize_weights::optimize_weights(T& node, program& p) {
                 // Need to restore the original shape
                 auto updated_output_layout = weights_reorder_params->get_output_layout();
                 auto orig_rank = prev_node.get_output_layout().get_partial_shape().size();
-                auto weight_format_dims = format::dimension(weights_reorder_params->get_output_layout().format);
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gpu/src/graph/graph_optimizer/post_optimize_weights.cpp
```diff
@@ -89,12 +89,18 @@ void post_optimize_weights::optimize_weights(T& node, program& p) {
...
+                if (node.type() == fully_connected::type_id()) {
+                    // Restore to 2d-reshaped layout at kernel_params
+                    updated_output_layout = node.get_input_layout(1);
+                } else {
+                    // Need to restore the original shape
+                    auto orig_rank = prev_node.get_output_layout().get_partial_shape().size();
+                    auto weight_format_dims = format::dimension(weights_reorder_params->get_output_layout().format);
+                    updated_output_layout.set_partial_shape(
+                        updated_output_layout.get_tensor().get_partial_shape(orig_rank, weight_format_dims));
+                }
```
yeonbok, line 103: This (is_dynamic()) block hits only when the node is fully connected. 
I think we need to check more detail prior history of the weight layout. 
(e.g., what was it in IR? how was it created in cldnn topology ? what was the weight_reorder_params dim and formt? 
Maybe we need to keep 3d but there was an error in get_partial_shape.. (maybe)

PR #19575, opened 2023-09-04
[CPU]unify oc block to optimize peak memory.

luweizhou2016:
### Details:
 - Unify brgemm oc_block=32 when weight size >=8M on amx bf16
 - When weight size >= 16M on avx512, disable 1x1 conv replacing FC  and unify fc brgemm  the oc_block size = 64.
 - When weight size >= 16M on avx512,  unify fc brgemm the oc_block size = 24.
### Tickets:
 - *CVS-119650*


yuxu42:
@luo-cheng2021  Could you please take a review? Thanks!

luweizhou2016:
onednn PR:https://github.com/openvinotoolkit/oneDNN/pull/210

luo-cheng2021 (approved):
LGTM.
An open idea, if there is a mechanism which supports showing the performance warning even if the `CPU_DEBUG_CAPS` is not enabled after enabling a switch at runtime, it will be great.

dmitry-gorokhov:
@luweizhou2016  As we discussed the PR should also cover client systems (ADL, RPL) and non-AMX servers (CLX, ICX).

luweizhou2016:
@dmitry-gorokhov , Update the changes on avx512 and avx2. Test on i9-13900K RPL and Xeon(R) Gold 6330 CPU

luweizhou2016:
> LGTM. An open idea, if there is a mechanism which supports showing the performance warning even if the `CPU_DEBUG_CAPS` is not enabled after enabling a switch at runtime, it will be great.

What I can think is to have an unified logger level control environment. By default on error is printed out, here the log should be setting to be inform level.

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -1004,7 +1017,8 @@ bool FullyConnected::canBeExecutedInConv1x1() const {
         widthInConv = srcDims[inRank - 2];
         K = srcDims[inRank - 1];
         N = weightDims[0];
-
+        if (weightMemPtr->getSize() >= (16 * 1 << 20))
```
dmitry-gorokhov, line 1020: Lets add comment that we are disabling Conv1x1 to avoid different weights layout usage and as a consuquence we decrease memory consumptuion. It also it worth mention that condition is limited 16 MB (bit shifting is not so easy to understand).

luweizhou2016: Applied.

dmitry-gorokhov (approved):

PR #19589, opened 2023-09-04
[CPU] Fix input output tensor binding

maxnick:
### Details:
In this PR a more conservative condition of using the double buffering is applied. In this PR when the output blob is used as an input, the second buffer is always used. Though the solution is not perfectly optimal (there are still cases when just one buffer would be enough), it doesn't bring any coupling between the infer request and the graph levels, thus it's more reliable and robust to the graph level changes. In the most LLM applications the double buffering is applied anyway due to the topological specifics, so we doesn't lose much in terms of optimal memory use there.
A more elegant solution would be by far more complex and require developing auxiliary abstraction layers, so following the Pareto principle I decided to sacrifice some theoretical memory usage benefits in favor to simplicity and robustness.

### Tickets:
 - 119568


dmitry-gorokhov:
@ceciliapeng2011 Could you please take a look on the fix?

ceciliapeng2011 (approved):

PR #19593, opened 2023-09-04
Squeeze fix

jane-intel:
### Details:
 - *Fixes shape inference of Squeeze operation. Previously Squeeze of tensor with shape [1, 2, -1, 4] had output shape of [2, 4]. Now we recognize that dynamic dimensions may become 1 or may become another dimension. This is why with current changes such Squeeze would have output shape of dynamic rank*
 - *Squeeze shape inference was previously altered to suite PDPD FE needs expressing unique behavior of Slice operation. I removed unnecessary change in the common Squeeze operation and put all the specifics of the PDPD operator in the translation rule*

### Tickets:
 - *CVS-119569*

### DIFF
--- a/src/core/shape_inference/include/squeeze_shape_inference.hpp
+++ b/src/core/shape_inference/include/squeeze_shape_inference.hpp
@@ -56,8 +56,7 @@ std::vector<TRShape> shape_infer(const Squeeze* op,
                 OPENVINO_SUPPRESS_DEPRECATED_END
                 unique_axes.reset(new std::set<int64_t>(axes->cbegin(), axes->cend()));
             } else if (arg_rank.get_length() > 0 && shape_size(axes_shape.to_shape()) == 1) {
-                // The `axes` input must be a Parameter with single element to ensure uniqueness of axes
-                // only rank is deduced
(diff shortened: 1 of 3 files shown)


ilyachur (approved):

src/core/shape_inference/include/squeeze_shape_inference.hpp
```diff
@@ -80,14 +79,21 @@ std::vector<TRShape> shape_infer(const Squeeze* op,
...
-                         [](const DimType& dim) {
-                             return !dim.compatible(1);
-                         });
+            // if only first input provided or axes are empty remove all dimensions equal to 1.
+            if (std::any_of(arg_shape.cbegin(), arg_shape.cend(), [](const DimType& d) {
+                    return d.is_dynamic() && d.compatible(1);
+                })) {
+                // we are unsure if dynamic dimensions would be equal to 1 or not, so we set dynamic output rank
+                output_shape = PartialShape::dynamic();
+                return output_shapes;
```
praasz, line 88:
```suggestion
```
I think after set the output shape function will exit with return at end.

praasz (approved):

PR #19601, opened 2023-09-05
[GPU] Use preferred output format if node impl type is onednn

jade-cho:
### Details:
 - *Conditions*
   - onednn is enabled on dGPU.
   - The model is partial dynamic shape.
   - The impl type of gemm and FC node is selected to onednn.
 - *Problems*
   - The input format of reshape is selected to b_fs_yx_fsv16.
   - To reshape from b_fs_yx_fsv16 to bfwzyx is not supported.
   - This is due to not selecting a preferred format in gemm and fc nodes.
   - Target pattern
     - Expected: MatMul (bfyx) -> Reshape (bfwzyx)
     - Actual:     MatMul (b_fs_yx_fsv16) -> Reshape (bfwzyx)

### Tickets:
 - *117969*

### DIFF
--- a/src/plugins/intel_gpu/src/graph/fully_connected.cpp
+++ b/src/plugins/intel_gpu/src/graph/fully_connected.cpp
@@ -161,9 +161,7 @@ std::vector<layout> fully_connected_inst::calc_output_layouts(fully_connected_no
     format::type output_format = is_static && !allow_new_shape_infer ? get_preferred_format(node, impl_param) :
                                               input_layout.format.value;
 
-    // Select preferred output format if the impl_type of FC node is onednn.
-    if (node.get_preferred_impl_type() == impl_types::onednn
-        && node.get_preferred_output_fmt() != format::any)
+    if (node.get_preferred_output_fmt() != format::any)
         output_format = node.get_preferred_output_fmt();
 
     return { layout{output_shapes[0], output_type, output_format} };
--- a/src/plugins/intel_gpu/src/graph/fully_connected.cpp
+++ b/src/plugins/intel_gpu/src/graph/fully_connected.cpp
@@ -161,6 +161,11 @@ std::vector<layout> fully_connected_inst::calc_output_layouts(fully_connected_no
     format::type output_format = is_static && !allow_new_shape_infer ? get_preferred_format(node, impl_param) :
                                               input_layout.format.value;
 
+    // Select preferred output format if the impl_type of FC node is onednn.
+    if (node.get_preferred_impl_type() == impl_types::onednn
+        && node.get_preferred_output_fmt() != format::any)
+        output_format = node.get_preferred_output_fmt();
+
     return { layout{output_shapes[0], output_type, output_format} };
 }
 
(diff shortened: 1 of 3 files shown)


isanghao:
LGTM. Let me check performance and let's have a review from another person.

src/plugins/intel_gpu/src/graph/gemm.cpp
```diff
@@ -108,7 +108,13 @@ std::vector<layout> gemm_inst::calc_output_layouts(gemm_node const& /*node*/, co
 
     std::vector<ShapeType> output_shapes = ov::op::v0::shape_infer(&op, input_shapes);
 
-    return { layout{output_shapes[0], output_type, input0_layout.format, prim->output_paddings[0]} };
+    // Select preferred output format if the impl_type of gemm node is onednn.
+    cldnn::format output_format = input0_layout.format;
+    if (node.get_preferred_impl_type() == impl_types::onednn
```
ahnyoung-paul, line 113: How about use kernel_impl_param instead of node? 
because we are going to avoid using node itself in calc_output_layouts.

```suggestion
    if (impl_param.preferred_impl_type == impl_types::onednn
```

ahnyoung-paul: You can add  preferred_impl_type and preferred_output_fmt to the kernel_impl_params if necessary.

ahnyoung-paul: We will change this code when we remove program node from calc_output_layouts.
You can just apply it as is.

ahnyoung-paul (approved):

src/plugins/intel_gpu/src/graph/fully_connected.cpp
```diff
@@ -161,6 +161,11 @@ std::vector<layout> fully_connected_inst::calc_output_layouts(fully_connected_no
     format::type output_format = is_static && !allow_new_shape_infer ? get_preferred_format(node, impl_param) :
                                               input_layout.format.value;
 
+    // Select preferred output format if the impl_type of FC node is onednn.
+    if (node.get_preferred_impl_type() == impl_types::onednn
+        && node.get_preferred_output_fmt() != format::any)
+        output_format = node.get_preferred_output_fmt();
+
```
yeonbok, line 168: BTW, preferred_output_fmt is set not only for onednn, but also dynamic shape scenario. 
I think we can set output format with node.get_prererred_output_fmt if it is not empty or any, not only if it is onednn. How do you think?

yeonbok:
Could you add a unittest?

isanghao:
No perf issue from daily test

src/plugins/intel_gpu/src/graph/fully_connected.cpp
jade-cho: I agree but I'm not sure. I will change and then try jenkins test.

jade-cho: Changed

isanghao (approved):

PR #19612, opened 2023-09-05
[TF Hub][TF FE] Clean-up all downloaded files for TF Hub models validation

rkazants:
**Details:** Clean-up all downloaded files for TF Hub models validation

**Ticket:** 117711

### DIFF
--- a/tests/model_hub_tests/models_hub_common/constants.py
+++ b/tests/model_hub_tests/models_hub_common/constants.py
@@ -3,6 +3,13 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+import tempfile
+
+if os.environ.get('TFHUB_CACHE_DIR') is not None:
+    tf_hub_cache_dir = os.environ['TFHUB_CACHE_DIR']
+else:
+    tf_hub_cache_dir = os.path.join(tempfile.gettempdir(), "tfhub_modules")
+    os.environ['TFHUB_CACHE_DIR'] = tf_hub_cache_dir
 
 # supported_devices : CPU, GPU, GNA
 test_device = os.environ.get('TEST_DEVICE', 'CPU;GPU').split(';')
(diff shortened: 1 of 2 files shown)


popovaan (approved):

tests/model_hub_tests/models_hub_common/constants.py
```diff
@@ -3,6 +3,13 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+import tempfile
+
+if os.environ.get('TFHUB_CACHE_DIR') is not None:
```
gkrivor, line 8: Looks like it is just a dictionary, I may suggest to use "if 'TFHUB_CACHE_DIR' in os.environ"? As an alternate way if you want to use get, I suggest to use it like: tf_hub_cache_dir = os.environ.get('TFHUB_CACHE_DIR', os.path.join(...))

tests/model_hub_tests/models_hub_common/constants.py
```diff
@@ -3,6 +3,13 @@
...
+    tf_hub_cache_dir = os.environ['TFHUB_CACHE_DIR']
+else:
+    tf_hub_cache_dir = os.path.join(tempfile.gettempdir(), "tfhub_modules")
+    os.environ['TFHUB_CACHE_DIR'] = tf_hub_cache_dir
```
gkrivor, line 12: Doublecheck: do you really need to update environment variable? I'm not sure I understand the reason if you use a tf_hub_cache_dir

gkrivor (approved):

tests/model_hub_tests/models_hub_common/constants.py
rkazants: Just to secure that this env variable is set up for sure. Most likely, TF Hub will always set it up but I did not find that they guarantee it for sure.

PR #19622, opened 2023-09-05
Migrate ReduceL1, ReduceL2 to new API

praasz:
### Details:
 - Migrate ReduceL1, ReduceL2 to new API
 - Add new shape utils to replace deprecated functions.

### Tickets:
 - [CVS-118473](https://jira.devtools.intel.com/browse/CVS-118473)


src/core/src/op/reduce_l1.cpp
```diff
@@ -2,84 +2,72 @@
...
-    OPENVINO_SUPPRESS_DEPRECATED_START
-    NGRAPH_CHECK(validate_host_tensor_vector(inputs, 2));
-    NGRAPH_CHECK(validate_host_tensor_vector(outputs, 1));
+    OPENVINO_ASSERT(outputs.size() == 1);
+    OPENVINO_ASSERT(inputs.size() == 2);
 
-    const auto reduction_axes =
-        get_normalized_axes_from_tensor(inputs[1], inputs[0]->get_partial_shape().rank(), get_friendly_name());
-    OPENVINO_SUPPRESS_DEPRECATED_END
+    std::cout << "Reduce L1 eval " << this << std::endl;
```
ilyachur, line 47: Please remove debug messages

src/core/dev_api/shape_util.hpp
```diff
@@ -7,6 +7,31 @@
 #include "openvino/core/shape.hpp"
 
 namespace ov {
+
+template <class TContainer, class TAxes>
+TContainer reduce_container(const TContainer& input, const TAxes& axes) {
```
ilyachur, line 12: Why do we need to add these new API to developer API?
Why cannot it be internal API?

praasz: My intension was to provide this template to create this reduction for different type of containers or in same template implementation use it as template. But I think is not required to be public right now, I can hide it and consider to make it public if will be required.

Will update it.

src/core/shape_inference/include/reduce_shape_inference.hpp
```diff
@@ -1,16 +1,42 @@
...
+#include "openvino/op/util/arithmetic_reductions_keep_dims.hpp"
+#include "openvino/op/util/logical_reduction_keep_dims.hpp"
 #include "utils.hpp"
 
 namespace ov {
 namespace op {
+namespace util {
+
+template <class T>
+result_shape_t<T> reduce_shape(const T& input_shape, std::vector<int64_t>& axes, const bool keep_dims) {
```
mitruska, line 17: Is it possible to reuse common reduce helper here, or to use shape_infer in the evaluate?

src/core/src/shape_util.cpp
```diff
@@ -88,5 +88,21 @@ bool is_dynamic_shape(const Shape& s) {
...
+    return keep_dims ? reduce_keep_dims(input, axes) : reduce(input, axes);
+}
+
+std::vector<size_t> reduce(const std::vector<size_t>& input, const AxisSet& axes) {
+    return ov::reduce_container(input, axes);
+}
+
+Shape reduce_keep_dims(const Shape& input, const AxisSet& axes) {
+    return ov::replace_container(input, axes);
+}
```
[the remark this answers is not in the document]
praasz: No, the purpose is to have them separate and if `keep_dims` not required the user can select correct implementation, and additional check is not required.
There is difference be also when using for shape or for vector.

src/core/dev_api/shape_util.hpp
```diff
@@ -7,6 +7,31 @@
...
+        if (std::find(axes.begin(), axes.end(), axis) == axes.end()) {
+            result.emplace_back(input[axis]);
+        }
+    }
+
+    return result;
+}
+
+template <class TContainer, class TAxes>
+TContainer replace_container(const TContainer& input, const TAxes& axes) {
```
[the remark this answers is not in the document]
praasz: It is not internal stuff so it dosn't matter now. These function scope is one *.cpp.

src/core/shape_inference/include/reduce_shape_inference.hpp
praasz: It give not good result in term of binary size. Current solution is sufficient in evaluate.

src/core/dev_api/shape_util.hpp
```diff
@@ -7,6 +7,31 @@
...
+    auto result = input;
+    for (auto&& axis : axes) {
+        result[axis] = 1;
+    }
```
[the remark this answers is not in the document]
praasz: The `auto&&` will deduce in more generic way, Here there will be no special advantages or disadvantages.

src/core/dev_api/shape_util.hpp
```diff
@@ -26,5 +51,11 @@ OPENVINO_API Shape make_dynamic_shape();
  */
 OPENVINO_DEPRECATED("This function is deprecated and will be removed soon.")
 OPENVINO_API bool is_dynamic_shape(const Shape& s);
+
+OPENVINO_API Shape reduce(const Shape& input, const AxisSet& axes);
+OPENVINO_API Shape reduce(const Shape& input, const AxisSet& axes, const bool keep_dims);
+OPENVINO_API std::vector<size_t> reduce(const std::vector<size_t>& input, const AxisSet& axes);
+
+OPENVINO_API Shape reduce_keep_dims(const Shape& input, const AxisSet& axes);
```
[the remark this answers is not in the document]
praasz: Yes, the `OPENVINO_API`  is required if want to use them in other libraries. They have to be exported.

Regarding source file, I think is OK, the dev_api specifies functions which are not in general public API but in developer API, but still stored in main ov library.

src/core/reference/include/openvino/reference/reduce_l1.hpp
```diff
@@ -7,33 +7,55 @@
...
+namespace func {
+template <class T, typename std::enable_if<std::is_unsigned<T>::value>::type* = nullptr>
+constexpr T abs(const T num) {
+    return num;
+}
+
+template <class T, typename std::enable_if<std::is_signed<T>::value || ov::is_floating_point<T>()>::type* = nullptr>
+T abs(const T num) {
+    return std::abs(num);
+}
```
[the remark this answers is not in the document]
praasz: `abs` is not scope of this PR.

src/core/reference/include/openvino/reference/reduce_l1.hpp
```diff
@@ -7,33 +7,55 @@
...
+ * @param in             Input iterator to data.
+ * @param out            Output iterator to results.
+ * @param in_shape       Input shape.
+ * @param reduction_axes Axes on which reduction is applied.
+ */
+template <class InputIt, class OutputIt>
+void reduce_l1(InputIt in, OutputIt out, const Shape& in_shape, const AxisSet& reduction_axes) {
+    using T = typename std::iterator_traits<OutputIt>::value_type;
+    static_assert(std::is_same<typename std::iterator_traits<InputIt>::value_type, T>::value,
+                  "Assume in and out same type.");
```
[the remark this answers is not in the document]
praasz: The goal is to migrate to new API, the refactor is optional on this stage unless is relative easy.

In cpp the recommended way is use iterators over raw pointers when same iteration over data are done.
The iterators give more flexibility, as iterator can be e.g. using iterator adaptor. 
In reduce there could be some `stride iterator` adaptor which  could work similar as coordinate iterator but instead generate coordinate and calculate product just shift iterator to next element. Then instead loop some stl algo can be used which can impact on reducing code size and some performance as less multiplication will be used.
Using iterator also avoid do pointer arithmetic in explicit way, which is also not recommended to do.

I've check impact on binary size and here this change not affect it (but I've seen on some other implementations there is some change which increase size, then will leave pointer until root cause)

Regarding const T* can be achieved by providing const iterator or iterator from const container, then if function want change input data there will be an error.

I'm not sure if this approach can be applied to all reference but if there is no size penalty it gives more flexibility as template implementation.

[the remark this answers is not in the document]
mitruska: > I'm not sure if this approach can be applied to all reference but if there is no size penalty it gives more flexibility as template implementation.

There shouldn't be any size impact until new usages with other in/out template types.
Of course, I agree that usage of the iterators inside the reference implementation is good, but in fact the raw pointers are still passed to the function, because this is what `tensor.data()` returns.

mitruska (approved):

src/core/dev_api/shape_util.hpp
```diff
@@ -26,5 +26,11 @@ OPENVINO_API Shape make_dynamic_shape();
...
```
ilyachur, line 34: I don't see that anybody uses `reduce_keep_dims` method outside the common library, so looks like it can be removed from the developer API

praasz: The plan was to replace it for example in [mvn refernce](https://github.com/openvinotoolkit/openvino/blob/25b1b4e26c265bd0c11f52c34ba45bd69bb1e78c/src/core/reference/include/openvino/reference/mvn.hpp#L29).

I can hide but where should I put declaration to be available in reference implementaion?

And just maybe as TODO in separate PR. 
Should the `shape_utils.hpp` be moved into sub-folder `openvino/core` in dev_api?

ilyachur: Yes it will be better to move to sub folder

ilyachur (approved):

PR #19634, opened 2023-09-05
[CPU][ARM] Reorder FullyConnected weights in scope of compile_model

EgorDuplensky:
### Details:
 - Basically move weights reordering from first inference to compile_model

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
+++ b/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
@@ -399,7 +399,7 @@ static dnnl::primitive_desc createPrimitiveDesc(const FCKey& key, const dnnl::en
         const bool found = DnnlExtensionUtils::find_implementation(prim_desc, brgconv_avx512_1x1);
 
         if (found)
-            return prim_desc;
+            return std::move(prim_desc);
     }
 
     // fallback to normal inner product primitive
@@ -441,9 +441,9 @@ static dnnl::primitive_desc createPrimitiveDesc(const FCKey& key, const dnnl::en
     const bool found = DnnlExtensionUtils::find_implementation(prim_desc, key.implType);
 
     if (found)
-        return prim_desc;
+        return std::move(prim_desc);
 
-    return first_desc;
+    return std::move(first_desc);
 }
 
 #if defined(OV_CPU_WITH_ACL)
--- a/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
+++ b/src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
@@ -392,7 +392,7 @@ createDescriptorInternalForConv(DnnlMemoryDescCPtr inputDescPtr,
     }
 }
 
-static dnnl::primitive_desc createPrimitiveDesc(const FCKey& key, const dnnl::engine engine) {
+static dnnl::primitive_desc createPrimitiveDesc(const FCKey& key, const dnnl::engine& engine) {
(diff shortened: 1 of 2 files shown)


dmitry-gorokhov:
@maxnick Could you please take a look?

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -329,6 +330,172 @@ void FullyConnected::prepackMLASWeight() {
...
+            convInDesc, convWeightDescAny, convOutDesc,
+            dnnl::memory::dims{1},   // stride
+            dnnl::memory::dims{0},   // dilation
+            dnnl::memory::dims{0},   // paddingL
+            dnnl::memory::dims{0},   // paddingR
+            attr);
+    }
+}
+
+static dnnl::primitive_desc createPrimitiveDesc(const FCKey& key, const dnnl::engine engine) {
```
maxnick, line 395:
```suggestion
static dnnl::primitive_desc createPrimitiveDesc(const FCKey& key, const dnnl::engine& engine) {
```

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -329,6 +330,172 @@ void FullyConnected::prepackMLASWeight() {
...
+
+    DnnlMemoryDescPtr weightDesc = MemoryDescUtils::convertToDnnlMemoryDesc(weightDescIP);
+    DnnlMemoryDescCPtr biasDesc = nullptr;
+    if (biasMemPtr) {
+        biasDesc = biasMemPtr->getDescWithType<DnnlMemoryDesc>();
+    }
+    DnnlMemoryDescCPtr inDesc = MemoryDescUtils::convertToDnnlMemoryDesc(
+        MemoryDescUtils::makeDummyDesc(*srcMemPtr->getDescWithType<DnnlMemoryDesc>()));
+    DnnlMemoryDescCPtr outDesc = MemoryDescUtils::convertToDnnlMemoryDesc(
+        MemoryDescUtils::makeDummyDesc(*dstMemPtr->getDescWithType<DnnlMemoryDesc>()));
```
maxnick, line 479: I would say it is more reliable to use memory descriptors from the port descriptors in `selected_pd `. The idea is that these memory descriptors are always undefined in the case of dynamic shapes and used specially for this purpose: creating defined memory descriptors from undefined ones with a specific shape. Please refer to: https://github.com/openvinotoolkit/openvino/blob/eacdc24d541eb800284b8d49d6fb1fae37cdce0b/src/plugins/intel_cpu/src/node.cpp#L634

EgorDuplensky: actually this was the first thing I have tried.
But onednn primitive_desc failed to create using memory descs from the selected pd

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
EgorDuplensky: done

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
EgorDuplensky: Updated as discussed (better to use selected_pd).

maxnick (approved):

EgorDuplensky:
rebased

dmitry-gorokhov (approved):

PR #19651, opened 2023-09-06
[CPU] improve reorder to support any precision

riverlijunjie:
### Details:
 - make reorder node to support arbitrary precision

### Tickets:
 - 120207


src/plugins/intel_cpu/src/nodes/reorder.cpp
```diff
@@ -189,11 +189,62 @@ void Reorder::createReorderPrimitive(const dnnl::memory::desc& srcDesc,
...
+        return false;
+    };
+
     const auto engine = getEngine();
     src_blocked = std::make_shared<Memory>(engine, DnnlExtensionUtils::makeDescriptor(srcDesc), srcPtr, false);
-
     dst_blocked = std::make_shared<Memory>(engine, DnnlExtensionUtils::makeDescriptor(dstDesc), dstPtr, false);
 
+    isPreConvert = !isSupportedDataType(srcDesc.get_data_type());
+    isPostConvert = !isSupportedDataType(dstDesc.get_data_type());
```
maxnick, line 213: First. This is not the best solution though. First of all, on this level we are working with oneDNN entities already. Thus it doesn't help us to avoid an exception during an attempt to create oneDNN memory descriptor, since on the cpu_plugin::Node level there might be an arbitrary precision, but  when we call `srcMemPtr->getDescWithType<DnnlMemoryDesc>()` on L179 we implicitly convert an arbitrary memory descriptor to an oneDNN memory desc, which is not guarantied to complete successfully in all the cases. Thus the check must be made even before the memory descriptor conversion.
Second. Using F32 as an intermediate type unconditionally  is not a good idea, since there might be rounding of integral values that may break accuracy. Thus. When both input and output types are not supported, we should simply reinterpret the smallest in size of the two types as any type of the same size supported by oneDNN and do the reordering with this type. Then reinterpret the type back and do the conversion.
Third. Since we have `Reorder::reorderData`, which is aimed to perform instant reordering on demand and it also have to support arbitrary precisions, I strongly encourage you to develop a Reorder Executor, similar to the existing DnnlExecutor class. For the intermediate buffer, the Scratchpad may be used when possible to reduce the amount of memory reallocations.

src/plugins/intel_cpu/src/dnnl_extension_utils.cpp
```diff
@@ -60,6 +62,10 @@ memory::data_type DnnlExtensionUtils::IEPrecisionToDataType(const InferenceEngin
             return memory::data_type::f16;
         case InferenceEngine::Precision::UNSPECIFIED:
             return memory::data_type::undef;
+        case InferenceEngine::Precision::FP64:
+        case InferenceEngine::Precision::U64:
+        case InferenceEngine::Precision::I64:
+            return memory::data_type::f64;
```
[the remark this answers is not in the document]
riverlijunjie: Is there a good way to solve the precision mismatch question? We know that dnnl has a limited precision range, but we need match full range of IE Precision to the limited dnnl precision.

src/plugins/intel_cpu/src/nodes/reorder.cpp
riverlijunjie: OK, let's do：
1.  Precision check should be put before MemDesc conversion to oneDNN mem desc
2. Choose intermediate type: smallest size of input/output types, and the same size supported by oneDNN
3. Develop reorderExecutor class
4. Use Scratchpad for intermediate conversion buffer


For 2, we need know which type (same size)is supported, do we also use 'try create dnnl::reorder' to find out the support data type? Maybe a redefined list is better?

src/plugins/intel_cpu/src/nodes/reorder.cpp
```diff
@@ -189,11 +189,62 @@ void Reorder::createReorderPrimitive(const dnnl::memory::desc& srcDesc,
     if (!selectedPD)
         IE_THROW() << "Preferable primitive descriptor is not set.";
 
+    // How to query what data_types are supported by dnnl::reorder
```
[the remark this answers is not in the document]
maxnick: From the performance perspective it's better to use some a kind of a list of supported precisions. The serious drawback of this solution is the necessity to manually maintain such a list. May be this is not a big problem. Another thing is that oneDNN supports not all the precision combinations in the reorder implementation. Thus, we would have to maintain a list of supported combinations as well. Thus, may be we need to do the following. Maintain a list of supported by oneDNN openvino types (may be simply modify DnnlExtensionUtils::IEPrecisionToDataType so that it return a kind of error code instead of exception, so we can check whether the type is supported by oneDNN), then, when we proved that the current precisions are supported, we need to try to make the reorder and check the result, and if it is not created it means that the current precisions combination is not supported and we have to do the reordering with only one type and then perform the conversion.

src/plugins/intel_cpu/src/nodes/reorder.cpp
maxnick: > For 2, we need know which type (same size)is supported, do we also use 'try create dnnl::reorder' to find out the support data type? Maybe a redefined list is better?
Please refer to my previous comment https://github.com/openvinotoolkit/openvino/pull/19651#discussion_r1317393915. In short, to detect unsupported precisions we can reuse `DnnlExtensionUtils::IEPrecisionToDataType` with minor modifications. To detect unsuported precisions combinations we still need to try to instance a reorder primitve.

src/plugins/intel_cpu/src/nodes/reorder.cpp
```diff
@@ -417,6 +414,380 @@ void Reorder::reorderData(const IMemory &input, const IMemory &output, MultiCach
...
+
+        // Allocate or resize input memeory
+        auto src_desc = dst->getDesc().cloneWithNewPrecision(post_converter->getInputPrec());
+        auto src_mem = post_converter->getInputMem();
+        if (src_mem) {
+            src_mem->redefineDesc(src_desc);
+        } else {
+            auto mem = std::make_shared<Memory>(engine, src_desc, nullptr, false);
+            post_converter->setInputMem(mem);
+        }
```
maxnick, line 617: As I mentioned above, it's enough to have just one intermediate storage. Moreover we can use `scratchpad` for this purpose. Thus, we don't need this step at all.

src/plugins/intel_cpu/src/nodes/reorder.h
```diff
@@ -62,18 +62,105 @@ class Reorder : public Node {
...
 private:
-    dnnl::reorder::primitive prim;
-    std::shared_ptr<MemoryDesc> input;
-    std::shared_ptr<MemoryDesc> output;
-
-    std::vector<int> src_permutation;
-
-    MemoryPtr dst_blocked;
-    MemoryPtr src_blocked;
+    class ReorderExecutor {
```
maxnick, line 68: If the executor is supposed to be shared between nodes, it shall have constant internal state. Otherwise we can have unexpected behaviour. The thing is the `prepareParams` method modifies the internal state, but it's not called every time for the other node every infer request since the `Reorder::prepareParams` method is called only when the shape change. Thus, some other node, which got the same executer, may have completelly different memory in the `primArgs` map for example with say nothing about the other parameters, which is of course incorrect.

src/plugins/intel_cpu/src/nodes/reorder.cpp
```diff
@@ -175,77 +175,71 @@ void Reorder::prepareParams() {
...
-                                     const dnnl::memory::desc& dstDesc,
-                                     void* dstPtr) {
+struct ReorderExecutorKey {
+    dnnl::memory::desc src;
+    dnnl::memory::desc dst;
+    InferenceEngine::Precision src_prc;
+    InferenceEngine::Precision dst_prc;
+    size_t hash() const;
+    bool operator==(const ReorderExecutorKey& rhs) const;
+};
```
maxnick, line 189: We should think about the need to cache this executor. The thing is, the reorder primitives are already cached, additional converts initialization isn't costly. The only thing that may be amortized with caching is the attempts to instance the reorder primitive to test whether the precision combination is supported.
The most correct solution is storing the base `MemoryDesc` types in the key, of course this will make the hashing and comparison far more complicated, but it allows to describe an arbitrary reorder node, and switch from direct use of the reorder primitive to reorder executor.

src/plugins/intel_cpu/src/nodes/reorder.cpp
riverlijunjie: It seems we need 2 intermediate storages, for example {input=FP64, output=FP64} oneDNN doesn't support, we need convert to FP32 and then do reorder:

FP64(input) -> convert  -> FP32 -> reorder -> FP32 -> convert to FP64(output)

It needs 2 FP32 memory. Is it right?

src/plugins/intel_cpu/src/nodes/reorder.cpp
```diff
@@ -417,6 +414,380 @@ void Reorder::reorderData(const IMemory &input, const IMemory &output, MultiCach
...
+        return;
+    if (output)
+        post_converter->setOutputPrec(output->getPrecision());
+    post_converter->convert();
+}
+
+bool Reorder::ReorderExecutor::exec(dnnl::stream strm) {
+    if (prim) {
+        preConvert();
+        prim.execute(strm, primArgs);
```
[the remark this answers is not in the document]
riverlijunjie: Can compare input desc and output desc to judge whether it don't need reordering?

src/plugins/intel_cpu/src/nodes/reorder.cpp
```diff
@@ -417,6 +414,380 @@ void Reorder::reorderData(const IMemory &input, const IMemory &output, MultiCach
...
+
+        // Choose data type with the smallest data size
+        near_data_type = find_near_data_type(src_prc, dst_prc);
+        OPENVINO_ASSERT(isSupportedDataType(engine, cache, src, near_data_type));
+        pre_converter->setOutputPrec(DnnlExtensionUtils::DataTypeToIEPrecision(near_data_type));
+        post_converter->setInputPrec(DnnlExtensionUtils::DataTypeToIEPrecision(near_data_type));
+        break;
+    default:
+        OPENVINO_THROW("Incorrect converter mask!");
+        break;
```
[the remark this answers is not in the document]
riverlijunjie: Supposed F32->FP64 reorder, we should apply FP32 as intermediate memory to do reorder (not sure can do in place), then we will allocate buffer to store FP32->FP32 reordering result, which will as  input to convert to FP64 destination.

While for FP64->FP64 reorder (oneDNN doesn't support), We need allocate 2x FP32 buffer for FP32 reorder:

      FP64(input)->FP32(allocated) -> reorder -> FP32(allocated)->FP64(output)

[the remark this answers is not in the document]
riverlijunjie: It seems oneDNN doesn't support reorder: {fp64 -> fp64}, {bin -> bin}

src/plugins/intel_cpu/src/nodes/reorder.h
riverlijunjie: I see! But I thought we have provided - Reorder::reorderData - for other node, and don't need share it to other nodes.
Is it necessary to share it to other nodes?

src/plugins/intel_cpu/src/nodes/reorder.cpp
riverlijunjie: Agree, it is the most optimal solution. The cost is that hash computation will become complicated.
I also think maybe not cache executor will be a balance solution.

src/plugins/intel_cpu/src/nodes/reorder.h
```diff
@@ -62,18 +62,89 @@ class Reorder : public Node {
     static std::string getReorderArgs(const MemoryDesc &parentDesc, const MemoryDesc &childDesc);
 
     static void reorderData(const IMemory &input, const IMemory &output, MultiCachePtr cache = nullptr);
+    static void reorderData2(const IMemory &input, const IMemory &output, MultiCachePtr cache = nullptr);
```
ceciliapeng2011, line 65: any comments to reorderData2 since the two functions are exactly of the same paramters?

src/plugins/intel_cpu/src/nodes/reorder.h
```diff
@@ -62,18 +62,89 @@ class Reorder : public Node {
...
+        std::shared_ptr<IntermConverter> post_converter = nullptr;
+        DnnlScratchPadPtr scratch_ptr;
+
+        MemoryCPtr dst_blocked;
+        MemoryCPtr src_blocked;
+
+        MemoryDescPtr input;
+        MemoryDescPtr output;
+
+        bool need_reorder;
```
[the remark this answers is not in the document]
riverlijunjie: ok, will add comments for these class members.

src/plugins/intel_cpu/src/nodes/reorder.h
riverlijunjie: Method reorderData2() is old implement and reorderData() is new implement by reorder executor.
We can remove old implement and only keep the new implement.

src/plugins/intel_cpu/src/graph.cpp
```diff
@@ -599,32 +572,7 @@ void Graph::InitEdges() {
...
-                reorderStatusInternal = convertNode->getChildEdgeAt(0)->needReorder();
-                if (reorderStatusInternal != Edge::ReorderStatus::No)
-                    edge = convertNode->getChildEdgeAt(0);
-            }
-            if (reorderStatusInternal != Edge::ReorderStatus::No) {
-                insertReorder(edge, reorderStatusInternal == Edge::ReorderStatus::Optimized);
-            }
+            insertReorder(edge, false);
             updateEdge(i);
         } else if (reorderStatus == Edge::ReorderStatus::Optimized) {
```
ceciliapeng2011, line 577: We could simplify this to two lines :
insertReorder(edge, reorderStatus == Edge::ReorderStatus::Optimized);
updateEdge(i);

ceciliapeng2011 (changes requested):

src/plugins/intel_cpu/src/graph.cpp
riverlijunjie: We need consider the status of `Edge::ReorderStatus::No`, which will not need insert reorder node.

src/plugins/intel_cpu/src/nodes/reorder.cpp
```diff
@@ -376,45 +331,283 @@ void Reorder::reorderData(const IMemory &input, const IMemory &output, MultiCach
...
+
+    dnnl::memory::desc src_desc, dst_desc;
+    if (src_data_type != dnnl::memory::data_type::undef && dst_data_type != dnnl::memory::data_type::undef) {
+        // Check whether dnnl::reorder directly support input->output
+        src_desc = updateSrcDesc(engine, src_permutation);
+        dst_desc = dst_blocked->getPrimitive().get_desc();
+        auto result = getReorderPrim(cache, engine, src_desc, dst_desc);
+
+        if (result) {
+            prim = result;
```
luo-cheng2021, line 436: I suppose this is a hot path for many cases, so can we change line 399 to avoid unnecessary construction?

src/plugins/intel_cpu/src/dnnl_extension_utils.cpp
```diff
@@ -58,10 +60,12 @@ memory::data_type DnnlExtensionUtils::IEPrecisionToDataType(const InferenceEngin
             return memory::data_type::bin;
         case InferenceEngine::Precision::FP16:
             return memory::data_type::f16;
+        case InferenceEngine::Precision::FP64:
+            return memory::data_type::f64;
         case InferenceEngine::Precision::UNSPECIFIED:
             return memory::data_type::undef;
         default: {
-            IE_THROW() << "The plugin does not support " << prec.name();
+            return memory::data_type::undef;
```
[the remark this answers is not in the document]
riverlijunjie: We need this function can return status to make reorder be aware of whether dnnl support this precision. If throw exception, it cannot get this information.

src/plugins/intel_cpu/src/nodes/reorder.cpp
```diff
@@ -376,45 +331,283 @@ void Reorder::reorderData(const IMemory &input, const IMemory &output, MultiCach
...
-            if (!reorder) {
-                IE_THROW() << "No reorder available for the following tensor descriptors: "
-                    << input.getDesc().serializeFormat() << " and " << output.getDesc().serializeFormat();
-            }
+        auto engine = output.getPrimitive().get_engine();
+        std::vector<int> src_permutation = {};
+
+        auto src = std::make_shared<Memory>(engine, input.getDescPtr(), input.getData(), false);
+        auto dst = std::make_shared<Memory>(engine, output.getDescPtr(), output.getData(), false);
+        auto executor = std::make_shared<ReorderExecutor>(engine, cache, src, dst, src_permutation);
```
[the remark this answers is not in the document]
riverlijunjie: It creates Memory objects with allocated buffer(reuse it), not consume too much additional memory, so no too much impac on performance.
About `ReorderExecutor`, it has applied dnnl::reorder cache, and encapsulates the logic of precision convert and dnnl::reorder, which is the most direct way to do different precision reoder operation. And it is almost the same with the previous implement of this function.

src/plugins/intel_cpu/src/nodes/reorder.cpp
riverlijunjie: line 399 is used to construct an empty dnnl::reorder, which is an empty object and is used to check whether need do reorder (in case of only do convert).

src/plugins/intel_cpu/src/dnnl_extension_utils.cpp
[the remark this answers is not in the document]
riverlijunjie: Another method is to use try...catch clause to catch the unsupported precision during check whether this precision is supported.

src/plugins/intel_cpu/src/nodes/reorder.cpp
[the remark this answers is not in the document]
riverlijunjie: `ReorderExecutor` can be created in stack, but `src, dst` still need use `std::make_shared` to construct due to it is required by `ReorderExecutor`

luo-cheng2021 (approved):
LGTM

riverlijunjie:
closed, and is replaced by https://github.com/openvinotoolkit/openvino/pull/18124

PR #19658, opened 2023-09-06
[TF Hub][GA] Set correct scheduler and model scope for regular validation

rkazants:
**Details:** Set correct scheduler and model scope for regular validation.

The current scheduler is working as expected to run the correct list of models: https://github.com/openvinotoolkit/openvino/actions?query=event%3Aschedule

**Ticket:** 117711


.github/workflows/linux.yml
```diff
@@ -1,8 +1,8 @@
 name: Tests on Linux (Ubuntu 22.04, Python 3.11)
 on:
   schedule:
-    # - cron: '0 20/72 * * *'
-    - cron: '10 * * * *'
+    # at 00:00 on Wednesday and Saturday
+    - cron: '00 00 * * 3,6'
```
akashchi, line 5:
```suggestion
    - cron: '0 0 * * 3,6'
```

akashchi, line 5:
```suggestion
    - cron: '0 0 * * 3,6'
```

akashchi (changes requested):

PR #19664, opened 2023-09-07
[core] Migrate the Assign operator to new API

praasz:
### Details:
 - Migrate the Assign operator to new API

### Tickets:
 - [CVS-118477](https://jira.devtools.intel.com/browse/CVS-118477)


src/core/src/op/assign.cpp
```diff
@@ -33,93 +31,95 @@ void op::v3::Assign::validate_and_infer_types() {
...
-    outputs[0]->set_unary(inputs[0]);
+    auto buffer = var_value->get_state();
+    buffer.set_shape(inputs[0].get_shape());
+    outputs[0].set_shape(inputs[0].get_shape());
 
-    void* input = inputs[0]->get_data_ptr();
-    outputs[0]->write(input, outputs[0]->get_size_in_bytes());
-    buffer->write(input, buffer->get_size_in_bytes());
+    inputs[0].copy_to(outputs[0]);
+    inputs[0].copy_to(buffer);
```
praasz, line 110: @ilyachur 
```suggestion
    std::memcpy(outputs[0].data(), inputs[0].data(), inputs[0].get_byte_size());
    std::memcpy(buffer.data(), inputs[0].data(), inputs[0].get_byte_size());
```
Or use mem copy is sufficient here (it produce smaller binary size)?

ilyachur: I think memcpy will be enough for this case, `copy_to` is a sugar in order to make a copy of tensor to new tensor with different layout.
As I understand, here inputs and outputs has the same layouts and shapes

praasz: OK, thanks for explanation. Will change to `memcpy`

ilyachur (approved):

mitruska (approved):

PR #19666, opened 2023-09-07
Gather: removed indices normalization pass and added some checks for Gather-8 -> Gather-7 conversion

antonvor:
### Details:
In case of Gather-8 the GatherNegativeConstIndicesNormalize() transformation did indices normalization and we got incorrect output values. Correct behavior according to [Gather-8](https://docs.openvino.ai/2023.0/openvino_docs_ops_movement_Gather_8.html) specification (example 7) is to fill out of bounds values with zeros

#### Changes:
- removed GatherNegativeConstIndicesNormalize() pass
- added checks into ConvertGather8ToGather7() pass
- added shared and transformation tests
- fixed cpu tests with out of bound negative indices

### Tickets:
117837


antonvor:
@itikhono may I ask you to take a look on this change from a transformations perspective

nshchego (approved):

antonvor:
@nshchego thanks, @dmitry-gorokhov could you take a look please?

antonvor:
@itikhono @pavel-esir as discussed I removed GatherNegativeConstIndicesNormalize() pass and added more checks into ConvertGather8ToGather7 pass, take a look please

src/common/transformations/src/transformations/op_conversions/convert_gather_downgrade.cpp
```diff
@@ -48,6 +48,40 @@ pass::ConvertGather8ToGather7::ConvertGather8ToGather7() {
...
+            return false;
+
+        auto axis = axis_constant->cast_vector<int64_t>();
+        if (axis.size() != 1) {
+            return false;
+        }
+        auto axis_value = axis[0];
+        // normalize `axis` value if it is negative
+        if (axis_value < 0) {
+            axis_value = axis_value + data.get_partial_shape().rank().get_length();
```
itikhono, line 66: please check that "data.get_partial_shape().rank()" is static

src/common/transformations/src/transformations/op_conversions/convert_gather_downgrade.cpp
```diff
@@ -48,6 +48,40 @@ pass::ConvertGather8ToGather7::ConvertGather8ToGather7() {
...
+        if (data.get_partial_shape().rank().get_length() < axis_value) {
+            return false;
+        }
+        // check `axis` dimension of data tensor is static
+        if (!data.get_partial_shape()[axis_value].is_static()) {
+            return false;
+        }
+        auto axis_dim = data.get_partial_shape()[axis_value].get_length();
+
+        auto indices = indices_constant->cast_vector<int32_t>();
```
itikhono, line 77: why not "int64_t"?

src/common/transformations/src/transformations/op_conversions/convert_gather_downgrade.cpp
```diff
@@ -48,6 +48,40 @@ pass::ConvertGather8ToGather7::ConvertGather8ToGather7() {
...
+        // check all the indices are not negative and not out of bound
+        for (size_t i = 0; i < indices.size(); i++) {
+            if (indices[i] < 0 || indices[i] >= axis_dim) {
```
itikhono, line 80: it looks like we can normalize indices < 0 
normalized = indices[i] + axis_dim 
if normalized < axis_dim, it's a valid case for conversion

antonvor: Yes, we can, but I don't know why we should do this?
According to [Gather-8 specification](https://docs.openvino.ai/2023.0/openvino_docs_ops_movement_Gather_8.html) Gather-8 should support negative indices. Also the indices can be out of bound, so, I think, that the plugins should support negative values processing if they support Gather-8 as is.
According to [Gather-7 specification](https://docs.openvino.ai/2023.0/openvino_docs_ops_movement_Gather_7.html) Gather-7 expects only non-negative indices (in other words already normalized)
And, finally, in this PR we're getting rid of `GatherNegativeConstIndicesNormalize()` trnasformation, that which mostly created the original problem and confusion.
I suppose to leave it as is and in case of needed we can easily add indices normalization later

src/common/transformations/src/transformations/op_conversions/convert_gather_downgrade.cpp
antonvor: changed to `int64_t`, thanks

src/common/transformations/src/transformations/op_conversions/convert_gather_downgrade.cpp
antonvor: added check, thanks

src/common/transformations/src/transformations/op_conversions/convert_gather_downgrade.cpp
itikhono: I would prefer this normalization to be implemented in this PR, it shouldn't affect the complexity of the transformation, just a couple additional lines in this `for` loop. 
The case when all indices of Gather8 after normalization satisfy the condition 0 <= norm_index < axis_dim is not something special and can occur in usual models.

antonvor: added normalization, thanks

itikhono (approved):

src/common/transformations/src/transformations/op_conversions/convert_gather_downgrade.cpp
itikhono: thanks! resolved

antonvor:
@itikhono ci is green

PR #19667, opened 2023-09-07
[GPU] Fix canonicalization for fused dep's shape

andrew-k-park:
### Details:
 - Fix custom canonicalize shapes for Gather to avoid unintentional shape extension for 1D fused dep's shape

### Tickets:
 - 116597, 118222

### DIFF
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/gather.cpp
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/gather.cpp
@@ -116,7 +116,12 @@ public:
             out_layout.format = format::adjust_to_rank(out_layout.format, output_pshape.size());
         }
 
-        return primitive_impl::static_canonicalize_shapes(updated_impl_params);
+        for (auto& input_layout : updated_impl_params.input_layouts) {
+            input_layout.set_partial_shape(extend_shape_to_rank_from_end(input_layout.get_partial_shape()));
+        }
+        out_layout.set_partial_shape(extend_shape_to_rank_from_end(out_layout.get_partial_shape()));
+
+        return updated_impl_params;
     }
 
     kernel_impl_params canonicalize_shapes(const kernel_impl_params& impl_params) const override {
--- a/src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.h
+++ b/src/plugins/intel_gpu/src/graph/impls/ocl/kernel_selector_helper.h
@@ -263,7 +263,9 @@ inline kernel_impl_params canonicalize_fused_shapes(const kernel_impl_params& im
                 auto dep_shape = dep_layout.get_partial_shape();
 
                 if (!broadcastable(dep_shape, out_pshape, use_new_shape_infer)) {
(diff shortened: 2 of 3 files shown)


andrew-k-park:
@Lyamin-Roman This PR is fix for case where the layout of fused dep is 1D tensor, which extends into an unintended shape. So could you please review if you have any concerns about canonicalization through this PR?

Lyamin-Roman:
> @Lyamin-Roman This PR is fix for case where the layout of fused dep is 1D tensor, which extends into an unintended shape. So could you please review if you have any concerns about canonicalization through this PR?

As far as I remember, extend shape from begin for eltwise is needed to simulate numpy broadcasting (in case of non-broadcastable shapes) and this change could potentially cause problems in other models
Also, could you please change the test because it doesn't demonstrate the problem, it works on master too

andrew-k-park:
> > @Lyamin-Roman This PR is fix for case where the layout of fused dep is 1D tensor, which extends into an unintended shape. So could you please review if you have any concerns about canonicalization through this PR?
> 
> As far as I remember, extend shape from begin for eltwise is needed to simulate numpy broadcasting (in case of non-broadcastable shapes) and this change could potentially cause problems in other models Also, could you please change the test because it doesn't demonstrate the problem, it works on master too

@Lyamin-Roman Test case has been updated so that issue can be reproduced on master. The topology for issue as follows:
image
Each gather has the same data input, and `gather_first` has `gather_second` as fused dep input. The problem occurs when `static_canonicalize_shapes` of `gather_first` is executed.
image
As shape of fused dep for `gather_second` expands from `[4624]` 1D tensor to `[1, 4624]` 2D tensor. Therefore, the issue is that the index order of `GET_INPUT_INDEX` for fused op is executed differently than intended. In conclusion, as of now, when canonicalizing the shape of the fused dep, it has been changed to extend shape from end rather than begin. This is issue related to the accuracy of customer model. Could you please review this from your perspective?

yeonbok:
Yeah this will affect global models and I I thinks we need more investigation using various cases.. 
@andrew-k-park  Could you add more dynamic functional tests and check the internal behvior? 
Various shapes where gather input / output ranks differs + eltwise with more variations to be fused? 
What if the original gather output shape is [1, 4096] and eltwise is 1d [4096] ?

andrew-k-park:
> Yeah this will affect global models and I I thinks we need more investigation using various cases.. @andrew-k-park Could you add more dynamic functional tests and check the internal behvior? Various shapes where gather input / output ranks differs + eltwise with more variations to be fused? What if the original gather output shape is [1, 4096] and eltwise is 1d [4096] ?

@yeonbok Fixed the root cause with Gather's `static_canonicalize_shapes`(custom canonicalize shapes) instead of changes with global impact.

@Lyamin-Roman Fixed the root cause where `canonicalize_fused_shapes` is unnecessarily called again by `primitive_impl::static_canonicalize_shapes` when `static_canonicalize_shapes` for gather is finished. This causes unintentional shape extension of fused dep's shape. Could you review this PR again?

Lyamin-Roman (approved):

yeonbok (approved):

Issue #19617, opened 2023-09-05
Add a clear error message when creating an empty `Constant`


p-wysocki:
### Context

Current Constant implementation does not allow for creating empty Constants:
https://docs.openvino.ai/2023.0/openvino_docs_ops_infrastructure_Constant_1.html

Trying to do so results in a cryptic error message:
```python
IndexError: index 0 is out of bounds for axis 0 with size 0
```

### What needs to be done?

1. Add a clear error message for the user trying to create an empty Constant
This can be done in either Python wrapper: https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/src/openvino/runtime/opset1/ops.py#L334-L347
Or in C++ to Python binding: https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/src/pyopenvino/graph/ops/constant.cpp

It's up to you to decide where the error message implementation will be a good fit.

2. Make sure the error message appears
This can be done either manually or by writing a test.

3. Create a Pull Request

### Resources 

- [Contribution guide](https://github.com/openvinotoolkit/openvino/blob/master/CONTRIBUTING.md)
- [How to extend OpenVINO Python API](https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/docs/code_examples.md)
- [How to test OpenVINO Python API](https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/docs/test_examples.md)
- [Blog post on contributing to OpenVINO](https://medium.com/openvino-toolkit/how-to-contribute-to-an-ai-open-source-project-c741f48e009e)

### Contact points

@p-wysocki
@jiwaszki
@akuporos

Don't hesitate to reach out, we're here to help!

Ticket: 114374

p-wysocki:
PR: https://github.com/openvinotoolkit/openvino/pull/19674

p-wysocki:
The PR has been merged. Thank you @siddhant-0707!

siddhant-0707:
Glad to help! 😃

p-wysocki:
I'm moving the issue to the completed section in Contributions Welcome list.

PR #19674, opened 2023-09-07
Add a error message when creating an empty Constant

siddhant-0707:
### Details:
 - Added an error message when creating an empty constant to the constant wrapper

### Tickets:
 - #19617


p-wysocki (approved):
Thank you for your contribution! LGTM. @siddhant-0707 have you tested it locally to see if the ValueError is raised?

There are currently some issues with CI, please do not worry that some jobs are failing.

siddhant-0707:
@p-wysocki Yes, ValueError raised correctly. I tested it on Python 3.10.12

akuporos:
Can you please also extend the test [here](https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/tests/test_graph/test_ops.py#L122), that error raised?

akuporos (commented):
Can you please also extend the test [here](https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/tests/test_graph/test_ops.py#L122), that error raised?

src/bindings/python/tests/test_graph/test_ops.py
```diff
@@ -117,9 +117,13 @@ def test_broadcast():
...
     Constant(Type.f32, Shape([3, 3]), list(range(9))),
     ov.constant(np.arange(9).reshape(3, 3), Type.f32),
-    ov.constant(np.arange(9).reshape(3, 3), np.float32)
+    ov.constant(np.arange(9).reshape(3, 3), np.float32),
+    None
 ])
 def test_constant(node):
+    if node is None:
+        with pytest.raises(ValueError):
+            ov.constant(None)
```
akuporos, line 126:
```suggestion
    ov.constant(np.arange(9).reshape(3, 3), np.float32)
])
def test_constant(node):
    with pytest.raises(ValueError) as e:
        ov.constant(None)
    assert 'Cannot create an empty Constant. Please provide valid data.' in str(e.value)
```

src/bindings/python/tests/test_graph/test_ops.py
```diff
@@ -117,9 +117,13 @@ def test_broadcast():
 @pytest.mark.parametrize("node", [
     Constant(Type.f32, Shape([3, 3]), list(range(9))),
     ov.constant(np.arange(9).reshape(3, 3), Type.f32),
-    ov.constant(np.arange(9).reshape(3, 3), np.float32)
+    ov.constant(np.arange(9).reshape(3, 3), np.float32),
+    None
```
akuporos, line 121: I suggest to rewrite a little bit this test using [this feature](https://docs.pytest.org/en/7.1.x/example/parametrize.html#parametrizing-conditional-raising).

```suggestion
    ov.constant(None)
```

src/bindings/python/tests/test_graph/test_ops.py
```diff
@@ -117,9 +117,13 @@ def test_broadcast():
 @pytest.mark.parametrize("node", [
     Constant(Type.f32, Shape([3, 3]), list(range(9))),
     ov.constant(np.arange(9).reshape(3, 3), Type.f32),
```
akuporos, line 119:
```suggestion
@pytest.mark.parametrize(("const", "args", "expectation"), [
    (Constant, (Type.f32, Shape([3, 3]), list(range(9))), does_not_raise()),
    (ov.constant, (np.arange(9).reshape(3, 3), Type.f32), does_not_raise()),
```

src/bindings/python/tests/test_graph/test_ops.py
```diff
@@ -117,9 +117,13 @@ def test_broadcast():
...
+    ov.constant(np.arange(9).reshape(3, 3), np.float32),
+    ov.constant(None)
 ])
 def test_constant(node):
+    if node is None:
+        with pytest.raises(ValueError):
+            ov.constant(None)
     assert node.get_type_name() == "Constant"
     assert node.get_output_size() == 1
     assert list(node.get_output_shape(0)) == [3, 3]
```
akuporos, line 129:
```suggestion
    (ov.constant, (np.arange(9).reshape(3, 3), np.float32), does_not_raise()),
    (ov.constant, [None], pytest.raises(ValueError)),
])
def test_constant(const, args, expectation):
    with expectation:
        node = const(*args)
        assert node.get_type_name() == "Constant"
        assert node.get_output_size() == 1
        assert list(node.get_output_shape(0)) == [3, 3]
```

akuporos: please, also add indentation to the line with `assert node.get_output_element_type(0) == Type.f32`, this assert need to be under `with expectation`.

and please, add `from contextlib import nullcontext as does_not_raise` in the begging of this file

akuporos (changes requested):

akuporos (approved):

jiwaszki (approved):

siddhant-0707:
Hey, was wondering if I need to do something else also to merge this PR?

akuporos:
> Hey, was wondering if I need to do something else also to merge this PR?

Actually no, just wait for the last job will pass

akuporos:
build_jenkins

PR #19676, opened 2023-09-07
[PT FE] Separate tracing and scripting modes

mvafin:
### Details:
 - *Currently we will attempt scripting if tracing failed, but frequently tracing fails because user provided incorrect `example_input` and if scripting will succeed we will use scripted model which may not convert, because tracing is better supported.*
 - *One solution is to do scripting only when user didn't provide us `example_input`. In this case we will always fail on incorrect `example_input` argument.*

### Tickets:
 - *None*


mvafin:
Adding do_not_merge until this is tested thoroughly

andrei-kochin (approved):

PR #19720, opened 2023-09-10
[GPU] Added zero input handling for Pad

Lyamin-Roman:
### Details:
- *Using fake input buffer for zero input*
 - *Fixed dump graph failure in strided_slice*

### Tickets:
 - *100132*

### DIFF
--- a/src/plugins/intel_gpu/tests/unit/test_cases/border_gpu_test.cpp
+++ b/src/plugins/intel_gpu/tests/unit/test_cases/border_gpu_test.cpp
@@ -1684,12 +1684,12 @@ public:
         blt_size_f = p.lt[1];
         blt_size_y = p.lt[2];
         blt_size_x = p.lt[3];
-    
+
         brb_size_b = p.rb[0];
         brb_size_f = p.rb[1];
         brb_size_y = p.rb[2];
         brb_size_x = p.rb[3];
-    
+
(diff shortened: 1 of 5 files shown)


p-durandin (approved):

sshlyapn (approved):

src/plugins/intel_gpu/tests/unit/test_cases/border_gpu_test.cpp
```diff
@@ -1825,5 +1825,90 @@ TEST_P(border_dynamic_test, border_dynamic_test) {}
 INSTANTIATE_TEST_SUITE_P(border_dynamic_test,
                          border_dynamic_test,
                          ::testing::ValuesIn(dynamic_params));
-};  // namespace
 
+TEST(border_gpu, basic_zero_input_dynamic) {
+    auto& engine = get_test_engine();
+
+    ov::PartialShape zero_input_static = {0, 1};
+    auto input = engine.allocate_memory({zero_input_static, data_types::f32, format::bfyx});
```
vladimir-paramuzov, line 1833: As I remember that may cause crash on some platforms due to attempt to allocate 0 bytes for USM memory. So you'd better allocate at least 1 byte and reinterpret it to required layout with zero dim.

PR #19731, opened 2023-09-11
[CPU]fix thread count in OMP

zhangYiIntel:
Let parallel_nt decides number of thread

### Details:
 - *Avoid using parallel_get_num_threads which returns 1 in OpenMP*
 - *Let parallel_nt decides the number of threads*

### Tickets:
 - *CVS-119830*


usstq:
```cpp
#elif OV_THREAD == OV_THREAD_OMP
#    include <omp.h>

#    include <algorithm>
#    include <cstdlib>
#    include <string>

/* MSVC still supports omp 2.0 only */
#    if defined(_MSC_VER) && !defined(__INTEL_COMPILER)
#        define collapse(x)
#    endif  // defined(_MSC_VER) && !defined(__INTEL_COMPILER)
inline int parallel_get_max_threads() {
    return omp_get_max_threads();
}
inline int parallel_get_num_threads() {
    return omp_get_num_threads();
}
```
according to  https://stackoverflow.com/a/49602765/9292588, omp_get_num_threads() works only inside parallel region, so your fix LGTM.

usstq (approved):

yuxu42:
@usstq could you please take a review? Thanks!

dmitry-gorokhov (approved):

PR #19733, opened 2023-09-11
[CPU] Empty tensor used with a custom op leads to CPU plugin exception.

nshchego:
### Details:
 - *Reference node was not executed if one of input tensor was empty.*

### Tickets:
 - *113021*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/reference.cpp
+++ b/src/plugins/intel_cpu/src/nodes/reference.cpp
@@ -123,7 +123,11 @@ ov::TensorVector Reference::prepareOutputs() const {
         void *dstDataPtr = getChildEdgesAtPort(i)[0]->getMemory().getData();
         ov::Shape shape = ovCoreNode->get_output_partial_shape(i).rank().get_length() == 0 ?
                 ov::Shape{} : getChildEdgesAtPort(i)[0]->getMemory().getStaticDims();
-        outputs.push_back(ov::Tensor(ovCoreNode->get_output_element_type(i), shape, dstDataPtr));
+        if (dstDataPtr != nullptr) {
+            outputs.push_back(ov::Tensor(ovCoreNode->get_output_element_type(i), shape, dstDataPtr));
+        } else {
+            outputs.push_back(ov::Tensor(ovCoreNode->get_output_element_type(i), shape));
+        }
     }
     return outputs;
 }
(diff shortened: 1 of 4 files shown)


dmitry-gorokhov:
@maxnick Please take a look

src/plugins/intel_cpu/src/nodes/reference.cpp
```diff
@@ -136,7 +136,11 @@ ov::TensorVector Reference::prepareOutputs() const {
         void *dstDataPtr = getChildEdgesAtPort(i)[0]->getMemory().getData();
         ov::Shape shape = ngraphOp->get_output_partial_shape(i).rank().get_length() == 0 ?
                 ov::Shape{} : getChildEdgesAtPort(i)[0]->getMemory().getStaticDims();
-        outputs.push_back(ov::Tensor(ngraphOp->get_output_element_type(i), shape, dstDataPtr));
+        if (dstDataPtr != nullptr) {
+            outputs.push_back(ov::Tensor(ngraphOp->get_output_element_type(i), shape, dstDataPtr));
+        } else {
+            outputs.push_back(ov::Tensor(ngraphOp->get_output_element_type(i), shape));
+        }
```
maxnick, line 143: Isn't it more reliable to check the shape on presence of zeroes? There can be other reasons (and not all of them are legit) `dstDataPtr` being `nullptr`, but only in the context of an empty tensor, it's ok to allocate tensors on the ngraph side. Also what about inputs?

src/plugins/intel_cpu/src/nodes/reference.cpp
```diff
@@ -108,22 +108,38 @@ bool Reference::needShapeInfer() const {
...
         ov::Shape shape = ovCoreNode->get_input_partial_shape(i).rank().get_length() == 0 ?
                 ov::Shape{} : getParentEdgesAtPort(i)[0]->getMemory().getStaticDims();
-        inputs.push_back(ov::Tensor(ovCoreNode->get_input_element_type(i), shape, srcDataPtr));
+
+        if (std::any_of(shape.begin(), shape.end(), [](const size_t dim) { return dim == 0lu; } )) {
+            inputs.push_back(ov::Tensor(ovCoreNode->get_input_element_type(i), shape));
+        } else {
+            if (!srcDataPtr) {
+                THROW_CPU_NODE_ERR("has empty input data on port ", i);
+            }
```
maxnick, line 121: NIT. This is rather an assert construction. Should we introduce `CPU_NODE_ASSERT`?

maxnick (approved):

src/plugins/intel_cpu/src/nodes/reference.cpp
nshchego: Done

slyalin:
Any progress?

PR #19735, opened 2023-09-11
Fixed scalar creating in tf.Graph decoder 

popovaan:
Root cause analysis: 
In tf.Graph decoder scalars are wrapped to list, then to nd.array before passing to ov.Tensor() constructor.
Creating of list leads to creating of additional object in memory which is deallocated before C++ part access it. 
This leads to corruption of scalar value. 

Solution: 
Do not create additional list and create nd.array directly from scalar.

Ticket: 117250

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update


src/bindings/python/src/openvino/frontend/tensorflow/node_decoder.py
```diff
@@ -152,7 +152,8 @@ def get_attribute(self, name):
             if self.m_parsed_content.size == 1:
                 if isinstance(self.m_parsed_content, np.ndarray):
                     return OVAny(Tensor(self.m_parsed_content))
-                return OVAny(Tensor(np.array([self.m_parsed_content]), shape=[1]))
+                self.m_parsed_content = np.array(self.m_parsed_content)
+                return OVAny(Tensor(self.m_parsed_content))
             ov_tensor = Tensor(self.m_parsed_content, shared_memory=self.m_shared_memory)
```
rkazants, line 157: Are we ready to remove the skip here: https://github.com/openvinotoolkit/openvino/blob/master/tests/model_hub_tests/tf_hub_tests/precommit_models#L9?

Or is rootcause different?

popovaan: Rootcause is different, mobilenet_v1_100_224 still fails.

src/bindings/python/src/openvino/frontend/tensorflow/node_decoder.py
```diff
@@ -152,7 +152,8 @@ def get_attribute(self, name):
...
             ov_tensor = OVAny(ov_tensor)
             return ov_tensor
```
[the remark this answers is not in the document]
popovaan: As discussed offline, there are no examples of models, so it was decided to not add this test.

rkazants (approved):

PR #19742, opened 2023-09-11
[OVC] do not parse inputs for py_api

pavel-esir:
### Details:
In cases like below ovc should not parse input and an error should be thrown:
 ```
from openvino.tool.ovc import ocnvert_model
...
 ov_model = ov.convert_model('model.onnx', input="input.1[5,3,224,224]")
```
 
### Tickets:
 - [CVS-119474](https://jira.devtools.intel.com/browse/CVS-119474)

### CODE REVIEW

File: tools/ovc/openvino/tools/ovc/cli_parser.py [hunk 1]
```diff
@@ -116,7 +110,20 @@ def is_single_input(input: [tuple, list]):
...
+    for input_value in split_inputs(inputs):
+
+        # Parse string with parameters for single input
+        node_name, shape = parse_input_value(input_value)
+        # pylint: disable=no-member
+        inputs_list.append((node_name, shape))
+    return inputs_list
+
+
+def input_to_input_cut_info(input: [dict, tuple, list]):
```
popovaan, line 125: This method still parses input_value in 140 line. 
Looks like we need to remove lines 134-144.

tests/layer_tests/ovc_python_api_tests/test_pytorch.py
```diff
@@ -1093,6 +1093,21 @@ def test_example_inputs(self):
...
+        with self.assertRaisesRegex(Exception, ".*No node with name.*"):
+            convert_model(pytorch_model, input='input1[1, 10]')
+
+    def test_incorrect_inputs_2(self):
+        from openvino.tools.ovc import convert_model
+        pytorch_model, _, _ = create_pytorch_nn_module_case1('')
+
+        # check that it accepts specified names as is without parsing into 2 different inputs
+        with self.assertRaisesRegex(Exception, 'No node with name input1\[1, 10\],input2\[2, 100\]'):
+            convert_model(pytorch_model, input='input1[1, 10],input2[2, 100]')
```
popovaan, line 1109: Please add negative test for
convert_model(pytorch_model, input='input1,input2')

tools/ovc/openvino/tools/ovc/cli_parser.py
pavel-esir: oouups, i actually removed them, but here they left by mistake. Will remove them

popovaan: Corrected my comment. Let's not remove them, but replace with the suggested code. As single string should be treated as single node name.

tests/layer_tests/ovc_python_api_tests/test_pytorch.py
popovaan: And this negative test:
convert_model(pytorch_model, input='input1[1,10]','input2[2,100]')

tools/ovc/openvino/tools/ovc/cli_parser.py
pavel-esir: removed this lines.
discussed offline, after all i decided to leave checking if it's a string outside of `input_to_input_cut_info`, e.g.
https://github.com/openvinotoolkit/openvino/blob/372009b6f02aa6300c616895bd45ce107c46f849/tools/ovc/openvino/tools/ovc/convert_impl.py#L97-L101

tests/layer_tests/ovc_python_api_tests/test_pytorch.py
pavel-esir: done

pavel-esir: added test for `convert_model(pytorch_model, input=['input1[1,10]','input2[2,100]'])` as well.

popovaan (approved):

src/bindings/python/src/openvino/frontend/tensorflow/utils.py
```diff
@@ -264,10 +266,19 @@ def tf_function(args):
 
         input_needs_packing = True
 
+    def are_shapes_defined(shape: Union[List, Dict]):
+        if shape is None:
+            return False
+
+        if isinstance(shape, list):
```
rkazants, line 273: What will be if the list is empty?

pavel-esir: Agreed, better to add a check. Normally we should not have an empty list argument here: if user specified input=[] then `normalize_inputs` will convert them to `None`
https://github.com/openvinotoolkit/openvino/blob/372009b6f02aa6300c616895bd45ce107c46f849/tools/ovc/openvino/tools/ovc/convert_impl.py#L356

But if something goes wrong and this function is called with empty list from somewhere else then there will be a problem because
```
In [4]: np.all([])
Out[4]: True
```
it will try to create input tensors but it should not do that for empty shapes. Therefore added check for emptiness as well.

tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -37,18 +37,15 @@ def is_shape_type(value):
...
     :param input: input cut parameters of single input
     :return: InputCutInfo
     """
     if isinstance(input, str):
-        # Parse params from string
-        node_name, shape = parse_input_value(input)
         # pylint: disable=no-member
-        return _InputCutInfo(node_name,
-                             PartialShape(shape) if shape is not None else None)
+        return _InputCutInfo(input, None)
```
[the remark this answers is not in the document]
pavel-esir: Returned back `str` to signature. I removed them in the first place because implemented solution without strings, but it turned out string args are still necessary but i forgot to return them back to signature. Thanks for noticing

PR #19761, opened 2023-09-11
[apiConformance] Remove bf16 precision as unsupported

sbalandi:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *CVS-119749*


sbalandi:
bf16 was deleted not only from ov_infer_request_mandatory , but from all places

src/tests/functional/plugin/shared/include/behavior/ov_infer_request/io_tensor.hpp
```diff
@@ -59,7 +59,6 @@ struct OVInferRequestCheckTensorPrecision : public testing::WithParamInterface<O
 
     std::vector<ov::element::Type> precisions = {
```
iefode, line 60: Should be moved to optional, not removed

iefode: @sbalandi ?

sbalandi: fixed

PR #19763, opened 2023-09-11
[core]Migrate the Abs operator to new API

praasz:
### Details:
 - Migrate Abs operator to new API.
 - Move abs function from reduce_l1 to abs.

### Tickets:
 -  [CVS-118475](https://jira.devtools.intel.com/browse/CVS-118475)

### DIFF
--- a/src/core/include/openvino/op/abs.hpp
+++ b/src/core/include/openvino/op/abs.hpp
@@ -30,9 +30,7 @@ public:
 
     std::shared_ptr<Node> clone_with_new_inputs(const OutputVector& new_args) const override;
 
-    OPENVINO_SUPPRESS_DEPRECATED_START
-    bool evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const override;
-    OPENVINO_SUPPRESS_DEPRECATED_END
+    bool evaluate(TensorVector& outputs, const TensorVector& inputs) const override;
     bool has_evaluate() const override;
 };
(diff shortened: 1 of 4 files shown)


ilya-lavrenov (commented):
Could you please explain why do we need all this refactoring? 
To migrate to new API, we need:
- HostTensor -> Tensor
- get_data_ptr -> data

Which is 10 lines of code, while current PR has 140 changed lines.

jane-intel (approved):
LGTM.

@ilya-lavrenov please take into consideration that legacy namespaces are also sorted out. and with Evaluate structure we get rid of switch-cases.

praasz:
> Could you please explain why do we need all this refactoring? To migrate to new API, we need:
> 
> * HostTensor -> Tensor
> * get_data_ptr -> data
> 
> Which is 10 lines of code, while current PR has 140 changed lines.

I've just maybe not good enough description for PR.

The main goal is migration to new API (replace Host tensor, includes, deprecated tools, using namespace)
Additionally If reference implementation is simple I'll check if applied changes reduce binary size (if yes then I modify it).

For more complex operator the migration will be done and reference implementation will be refactored later if there will be benefit for binary size reduction (big templates or used in many operators).

For ops which not have evaluate, the migration can be done in batch as it will be only removing legacy code.
For reference implementation if not used in operators then only remove legacy code will be applied (can also be done in batch), they not impact on ov binary size but only for template plugin size.

PR #19767, opened 2023-09-12
[GPU] Do not use usm_host memory buffers for PVC as a device inputs

sshlyapn:
### Details:
 - This PR prohibits usm_host memory buffers usage as a device inputs since sometimes it causes significant perf degradations 
 
### Tickets:
 - 121649

### DIFF
--- a/src/plugins/intel_gpu/src/plugin/sync_infer_request.cpp
+++ b/src/plugins/intel_gpu/src/plugin/sync_infer_request.cpp
@@ -28,6 +28,19 @@
 
 namespace {
 
+inline bool can_use_usm_host(const cldnn::engine& engine) {
+    auto can_use_usm = engine.use_unified_shared_memory();
+
+    // WA: Disable USM host memory for infer request`s tensors for PVC as
+    // it has performance issues in case of host <-> device data transfers inside kernels
+    // Use unsupported SIMD8 as unique attribute of PVC
+    auto supported_simd_sizes = engine.get_device_info().supported_simd_sizes;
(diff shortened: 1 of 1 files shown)


src/plugins/intel_gpu/src/plugin/infer_request.cpp
```diff
@@ -25,6 +25,18 @@ const char str_device_output_unsupported_blob[] = "Device output is of an unsupp
 const char str_input_not_allocated[] = "Input data was not allocated.";
 const char str_output_not_allocated[] = "Output data was not allocated.";
 
+static bool can_use_usm_host(const cldnn::engine& engine) {
+    auto can_use_usm = engine.use_unified_shared_memory();
+
+    // WA: Disable USM host memory for infer request`s tensors for PVC as
+    // it has performance issues in case of host <-> device data transfers inside kernels
+    auto supported_simd_sizes = engine.get_device_info().supported_simd_sizes;
+    if (std::find(supported_simd_sizes.begin(), supported_simd_sizes.end(), 8) == supported_simd_sizes.end())
```
vladimir-paramuzov, line 34: Why not checking `gfx_ver` instead? Do we still have version parsing issues?

sshlyapn: @vladimir-paramuzov, how fixed version is supposed to look like? Something like "12.8.0"?

vladimir-paramuzov: Yes, 12.8.x:
https://github.com/intel/compute-runtime/blob/master/opencl/source/xe_hpc_core/cl_gfx_core_helper_xe_hpc_core.cpp#L29

p-durandin (approved):

PR #19768, opened 2023-09-12
[GPU] Fixed static init order for serialization

vladimir-paramuzov:
### Details:
 - `type_for_serialization` could be initialized after `BinaryOutputBuffer<>` and save/load functions are registered with "" as a key instead of type identifier. Reproduced with clang-16
 -  `DECLARE_OBJECT_TYPE_SERIALIZATION ` is now called for all primitives within `CLDNN_DECLARE_PRIMITIVE` macro
 - Updated `DECLARE_OBJECT_TYPE_SERIALIZATION` call in primitive impls to pass impl name

### Tickets:
 - *119938*


ilya-lavrenov (approved):
Tested on the reported configuration - no crash with this PR 👍 
Thanks!

e-ddykim (approved):
Looks good to me

PR #10469, opened 2022-02-17
DetectionOutput perf fix

chenhu-wang:
### Details:
 - *DetectionOutput perf fix*

### Tickets:
 - *79085*

PR #19773, opened 2023-09-12
[CPU] DetectionOutput perf fix

chenhu-wang:
### Details:
 - *In dense implementation, letting filter+topk+nms process in same loop should be better from data access perspective.*
 - *Continue for closed pr https://github.com/openvinotoolkit/openvino/pull/10469, can not reopen after pushed before reopen*

### Tickets:
 - *CVS-79085*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/detection_output.cpp
+++ b/src/plugins/intel_cpu/src/nodes/detection_output.cpp
@@ -123,7 +123,7 @@ void DetectionOutput::prepareParams() {
         confInfoForPrior.resize(imgNum * priorsNum);
 
     // confs...count...indices for caffe style and sparsity case.
-    // caffe: conf_info for sparsity or indices for dense --> topk(buffer) --> nms(indices)
+    // caffe: filter(conf_info for sparsity or indices for dense) --> topk(buffer) --> nms(indices)
     //        --> g_topk(vector<>(all detections) --> indices per class))
     // MXNet: max conf for prior within img, filter(indices) --> topk_img(buffer) --> nms_cls(indices)
     //        --> g_topk(vector<>(all detections) --> indices per class))
@@ -203,9 +203,7 @@ void DetectionOutput::execute(dnnl::stream strm) {
     if (!isSparsityWorthwhile) {
         confReorderDense(confData, ARMConfData, reorderedConfData);
 
-        if (!decreaseClassId) {
-            confFilterCF(reorderedConfData, indicesData, indicesBufData, detectionsData);
-        } else {
+        if (decreaseClassId) {
             confFilterMX(confData, ARMConfData, reorderedConfData, indicesData, indicesBufData, detectionsData);
         }
     } else { // sparsity
@@ -244,7 +242,7 @@ void DetectionOutput::execute(dnnl::stream strm) {
             }
         } else {
             for (int c = 0; c < locNumForClasses; ++c) {
-                if (c == backgroundClassId) {
+                if (c == static_cast<size_t>(backgroundClassId)) {
                     continue;
                 }
                 int locShift = n * priorsNum * locNumForClasses;
@@ -270,9 +268,14 @@ void DetectionOutput::execute(dnnl::stream strm) {
             // Caffe style
             parallel_for(classesNum, [&](int c) {
                 if (c != backgroundClassId) {  // Ignore background class
-                    int *pindices    = indicesData + n * classesNum * priorsNum + c * priorsNum;
-                    int *pbuffer     = indicesBufData + n * classesNum * priorsNum + c * priorsNum;
+                    int off = n * priorsNum * classesNum + c * priorsNum;
+                    const float *pconf = reorderedConfData + off;
+                    int *pindices = indicesData + off;
+                    int *pbuffer = indicesBufData + off;
                     int *pdetections = detectionsData + n * classesNum + c;
+
+                    if (!isSparsityWorthwhile)
+                        confFilterCF(pconf, pindices, pbuffer, pdetections, n);
 
                     const float *pboxes;
                     const float *psizes;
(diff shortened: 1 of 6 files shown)


chenhu-wang:
@a-sidorova , could you please help take a look?

src/plugins/intel_cpu/tests/functional/single_layer_tests/detection_output.cpp
```diff
@@ -8,16 +8,39 @@
...
 using namespace CPUTestUtils;
 using namespace ov;
 using namespace test;
+using namespace dnnl;
 
 namespace CPULayerTestsDefinitions {
 
 using namespace ov::test;
 
+unsigned get_cache_size(int level, bool per_core) {
```
a-sidorova, line 24: Is it possible to reuse [get_cache_size](https://github.com/openvinotoolkit/openvino/blob/b00fbd04c5c259672f65a29eeaec4b7ad6a0d433/src/plugins/intel_cpu/src/onednn/dnnl.h#L15) here to avoid code duplication?

src/plugins/intel_cpu/src/nodes/detection_output.cpp
```diff
@@ -289,9 +288,16 @@ void DetectionOutput::execute(dnnl::stream strm) {
             });
         } else {
             // MXNet style
-            int *pbuffer = indicesBufData + n * classesNum * priorsNum;
+            int offImg = n * priorsNum * classesNum;
```
a-sidorova, line 291: minor
```suggestion
            const int offImg = n * priorsNum * classesNum;
```
You can add `const` for all variables which are constant

src/plugins/intel_cpu/tests/functional/single_layer_tests/detection_output.cpp
```diff
@@ -111,6 +138,18 @@ class DetectionOutputLayerCPUTest : public testing::WithParamInterface<Detection
         return result.str();
     }
 
+    static void scale_input_shapes(std::vector<ov::test::InputShape> &inShapes) {
```
a-sidorova, line 141: Could you please add the comment why we have this function?

a-sidorova (commented):
The main comment: could you tell me please if you validated your fix on real models and see perf improvements?
The ticket contains the model in the description. May I ask you to attach the perf results of the whole model and only DO op for this model to the ticket (no in PR, for sure)?

chenhu-wang:
> The main comment: could you tell me please if you validated your fix on real models and see perf improvements? The ticket contains the model in the description. May I ask you to attach the perf results of the whole model and only DO op for this model to the ticket (no in PR, for sure)?
@a-sidorova ,  result is update on tickets.

src/plugins/intel_cpu/tests/functional/single_layer_tests/detection_output.cpp
chenhu-wang: added, thanks.

src/plugins/intel_cpu/src/nodes/detection_output.cpp
chenhu-wang: updated, thanks Alexandra.

src/plugins/intel_cpu/tests/functional/single_layer_tests/detection_output.cpp
chenhu-wang: 1. This function([get_cache_size](https://github.com/openvinotoolkit/openvino/blob/b00fbd04c5c259672f65a29eeaec4b7ad6a0d433/src/plugins/intel_cpu/src/onednn/dnnl.h#L15)) is just kind of wrap of ondnn call, and not exported as a cpu_plugin API. I am not sure it's make sense to export it as a plugin API just for testing.
2. I used to try export it and have some dependency issue.
So if it's need, I would like do it in a separate PR. what do you think?

a-sidorova: I'd suggest to add this static method to [CPUTestsBase](https://github.com/openvinotoolkit/openvino/blob/54609e7b72d9faa5a38fe7ac15c84aae77fbbc82/src/plugins/intel_cpu/tests/functional/test_utils/cpu_test_utils.hpp#L123) class to have opportunity to reuse it in other node tests at least. Moreover, this class already has the wrappers for `fmt2str` and `str2fmt` that are wraps of onednn call as well.
What do you think? If you would like to do in the separate PR and CI is green, let's do so. Otherwise I'd suggest to apply it in the current PR 😃

a-sidorova (approved):

src/plugins/intel_cpu/tests/functional/single_layer_tests/detection_output.cpp
chenhu-wang: Agree, have done in [6e572c1](https://github.com/openvinotoolkit/openvino/pull/19773/commits/6e572c1a880e4b58fcb11a4a53e0a8cd907b1302)

chenhu-wang:
@dmitry-gorokhov , could you please final review?

src/plugins/intel_cpu/tests/functional/single_layer_tests/detection_output.cpp
```diff
@@ -111,6 +115,21 @@ class DetectionOutputLayerCPUTest : public testing::WithParamInterface<Detection
         return result.str();
     }
 
+    // There are two major implemenation path for small tensor and large tensor(confidence tensor) compared with cache size.
+    // This function is to scale input shape based on current machine cache size,
+    // such that tensor size is large enough to cover sparsity implementation.
+    static void scale_input_shapes(std::vector<ov::test::InputShape> &inShapes) {
```
dmitry-gorokhov, line 121: I would strongly advise to avoid such cache size dependency in tests. Can we just add more test cases with different shapes to make sure sparse impl will be covered on each system?

dmitry-gorokhov (changes requested):

src/plugins/intel_cpu/tests/functional/single_layer_tests/detection_output.cpp
chenhu-wang: done, thanks Dmitry.

dmitry-gorokhov (approved):

PR #19775, opened 2023-09-12
[Snippets] Created common static method is_supported_fq

a-sidorova:
### Details:
 - *`FQ` is supported in Snippets only be decomposition. The PR adds common static method `is_supported_fq` that returns `true` if `FQ` will be decomposed and consequently is supported by Snippets. It means that we can reuse this check in tokenization.*
 - *Refactored related to `FQ` decomposition code a little bit.*

### Tickets:
 - *119707*


src/common/snippets/src/pass/validate.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    for (const auto& op : m->get_ordered_ops()) {
+        if (const auto constant = as_type_ptr<ov::op::v0::Constant>(op)) {
+            VALIDATE(is_support_constant, constant);
+        } else if (const auto convert = ov::as_type_ptr<ov::op::v0::Convert>(op)) {
+            VALIDATE(is_support_convert, convert);
+        } else if (const auto matmul = ov::as_type_ptr<ov::op::v0::MatMul>(op)) {
+            VALIDATE(is_supported_matmul, matmul);
+        } else if (ov::is_type<ov::op::v8::Softmax>(op) ||
+                   ov::is_type<ov::op::v1::Softmax>(op)) {
```
IvanNovoselov, line 78: Over here, for example. All other operations are handled through `as_type_ptr`, but this approach doesn't work well when you need to cover several ops. Probably the easiest way to align with other ops is to hide these casts inside validator or under the define.

src/common/snippets/src/pass/validate.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+bool is_support_convert(const std::shared_ptr<const ov::op::v0::Convert>& convert) {
+    return ov::is_type<op::ConvertTruncation>(convert) || ov::is_type<op::ConvertSaturation>(convert);
+}
+
+bool is_supported_matmul(const std::shared_ptr<const ov::op::v0::MatMul>& matmul) {
+    return !matmul->get_transpose_a() && !matmul->get_transpose_b();
+}
+
+bool is_support_softmax(const std::shared_ptr<const ov::Node>& softmax) {
```
IvanNovoselov, line 46:
```suggestion
bool is_supported_softmax(const std::shared_ptr<const ov::Node>& softmax) {
```

src/common/snippets/src/pass/validate.cpp
```diff
@@ -0,0 +1,90 @@
...
+bool is_support_constant(const std::shared_ptr<const ov::op::v0::Constant>& constant) {
+    const auto consumers = constant->get_output_target_inputs(0);
+    return ov::shape_size(constant->get_output_shape(0)) == 1 ||
+           (consumers.size() == 1 &&
+            (ov::is_type<const ov::op::v1::Transpose>(consumers.begin()->get_node()) ||
+             ov::is_type<const ov::op::v1::Broadcast>(consumers.begin()->get_node()) ||
+             ov::is_type<const ov::op::v3::Broadcast>(consumers.begin()->get_node())));
+}
+
+bool is_support_convert(const std::shared_ptr<const ov::op::v0::Convert>& convert) {
```
IvanNovoselov, line 38:
```suggestion
bool is_supported_convert(const std::shared_ptr<const ov::op::v0::Convert>& convert) {
```

src/common/snippets/src/pass/validate.cpp
```diff
@@ -0,0 +1,90 @@
...
+        } else if (const auto convert = ov::as_type_ptr<ov::op::v0::Convert>(op)) {
+            VALIDATE(is_support_convert, convert);
+        } else if (const auto matmul = ov::as_type_ptr<ov::op::v0::MatMul>(op)) {
+            VALIDATE(is_supported_matmul, matmul);
+        } else if (ov::is_type<ov::op::v8::Softmax>(op) ||
+                   ov::is_type<ov::op::v1::Softmax>(op)) {
+            VALIDATE(is_support_softmax, op);
+        } else if (ov::is_type<ov::op::v0::FakeQuantize>(op) ||
+                   ov::is_type<ov::op::v1::Reshape>(op)) {
+            VALIDATE(is_supported, op);
```
IvanNovoselov, line 82: Minor: By analogy with other methods, one might think that FQ is actually supported, which is a bit confusing. Do you think it's worth renaming `is_supported` to `not_supported` 🙃 ?

IvanNovoselov (approved):

src/common/snippets/src/pass/validate.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+namespace ov {
+namespace snippets {
+namespace pass {
+
+namespace {
+#define VALIDATE(op, op_type, validator)    \
+    if (ov::is_type<op_type>(op)) \
+        OPENVINO_ASSERT(validator(op), "Snippets validation of OV body has been failed: " + \
+                                       std::string(op->get_type_name()) + " op " + op->get_friendly_name() + " is not supported");
```
IvanNovoselov, line 28: You probably need `else` here as well (as was in my comment). Otherwise there will be a bunch of redundant casts for every successfully validated op

a-sidorova: For sure, I didn't notice `else` in your comment. And unfortunately, I didn't add it myself. Thanks!

dmitry-gorokhov:
@a-sidorova As we discussed offline Validation pass should be refactored to avoid potential issues caused by backend customization

a-sidorova:
> @a-sidorova As we discussed offline Validation pass should be refactored to avoid potential issues caused by backend customization

Implemented. Could you please take a look one more time?

dmitry-gorokhov (approved):

PR #19779, opened 2023-09-12
[PT FE] Support aten::one_hot

mvafin:
### Details:
 - *Support aten::one_hot*

### Tickets:
 - *119340*

### DIFF
--- a/src/frontends/pytorch/src/op/one_hot.cpp
+++ b/src/frontends/pytorch/src/op/one_hot.cpp
@@ -1,0 +1,50 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "openvino/op/one_hot.hpp"
+
+#include "openvino/frontend/pytorch/node_context.hpp"
+#include "openvino/op/constant.hpp"
+#include "openvino/op/greater.hpp"
+#include "openvino/op/reduce_max.hpp"
+#include "openvino/op/select.hpp"
+#include "openvino/op/convert.hpp"
+#include "openvino/op/add.hpp"
+#include "utils.hpp"
+
+namespace ov {
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+using namespace ov::op;
+
+OutputVector translate_one_hot(const NodeContext& context) {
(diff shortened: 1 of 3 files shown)


src/frontends/pytorch/src/op/one_hot.cpp
```diff
@@ -0,0 +1,50 @@
...
+    auto on_value = context.mark_node(v0::Constant::create(element::i64, Shape{}, {1}));
+    auto off_value = context.mark_node(v0::Constant::create(element::i64, Shape{}, {0}));
+    Output<Node> num_classes;
+    if (context.input_is_none(1)) {
+        num_classes = context.mark_node(v0::Constant::create(element::i64, Shape{}, {-1}));
+    } else {
+        num_classes = context.get_input(1);
+    num_classes = context.mark_node(std::make_shared<v0::Convert>(num_classes, element::i64));
+    }
+    auto zero = context.mark_node(v0::Constant::create(element::i64, Shape{}, {0}));
```
PiotrKrzem, line 37: It might be less descriptive, however, we could reuse on- and off-values as zero and one (ie, auto zero = off-value) to avoid creating multiple constants

src/frontends/pytorch/src/op/one_hot.cpp
```diff
@@ -0,0 +1,50 @@
...
+    auto max_class = context.mark_node(std::make_shared<v1::ReduceMax>(x, axes));
+    max_class = context.mark_node(std::make_shared<v1::Add>(max_class, one));
+    num_classes = context.mark_node(std::make_shared<v1::Select>(greater, num_classes, max_class));
+    return {context.mark_node(std::make_shared<v1::OneHot>(x, num_classes, on_value, off_value, -1))};
+};
+
+}  // namespace op
+}  // namespace pytorch
+}  // namespace frontend
+}  // namespace ov
```
PiotrKrzem, line 50:
```suggestion
}  // namespace ov

```

PiotrKrzem (approved):
clang, otherwise LGTM!

src/frontends/pytorch/src/op/one_hot.cpp
mvafin: Agree, good to reuse

PR #19782, opened 2023-09-13
[GPU] low accuracy problem in stable diffusion on igpu

sungeunk:
The f16 accumulation of ocl conv kernels has an accuracy problem.
For unet/f16/iGPU, 4 conv kernels are included.
- convolution_gpu_ref.cl
- convolution_gpu_bfyx_os_iyx_osv16.cl
- convolution_gpu_bfyx_to_fs_byx_fsv32.cl
- convolution_gpu_fs_byx_fsv32.cl

### Tickets:
 - 119157

### DIFF
--- a/src/plugins/intel_gpu/src/kernel_selector/kernels/convolution/convolution_kernel_base.cpp
+++ b/src/plugins/intel_gpu/src/kernel_selector/kernels/convolution/convolution_kernel_base.cpp
@@ -462,4 +462,10 @@ Datatype ConvolutionKernelBase::GetAccumulatorType(const convolution_params& par
     return params.inputs[0].GetDType();
 }
 
+Datatype ConvolutionKernelBase::GetAccumulationType(const convolution_params& params) const {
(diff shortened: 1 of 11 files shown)


src/plugins/intel_gpu/src/kernel_selector/kernels/convolution/convolution_kernel_base.cpp
```diff
@@ -462,4 +462,10 @@ Datatype ConvolutionKernelBase::GetAccumulatorType(const convolution_params& par
     return params.inputs[0].GetDType();
 }
 
+Datatype ConvolutionKernelBase::GetAccumulationType(const convolution_params& params) const {
+    auto accumulator_dt = GetAccumulatorType(params);
+    // WA: f16 accumulation occurs an accuracy problem.
+    return (accumulator_dt == Datatype::F16) ? Datatype::F32 : accumulator_dt;
```
vladimir-paramuzov, line 468: I suppose that will cause performance drop on many other models.

PR #19783, opened 2023-09-13
fix CPU interpolate bug and add related test case

meiyang-intel:
### Details:
 - *Fix CPU interpolate bug and add related test case*

### Tickets:
 - *CVS-97985*

### DIFF
--- a/src/plugins/intel_cpu/tests/functional/single_layer_tests/interpolate.cpp
+++ b/src/plugins/intel_cpu/tests/functional/single_layer_tests/interpolate.cpp
@@ -205,9 +205,15 @@ protected:
             inputShapes.push_back(InputShape({static_cast<int64_t>(axes.size())}, std::vector<ov::Shape>(dataShape.second.size(), {axes.size()})));
         }
 
+        auto simple_post_ops = std::find_if(fusedOps.begin(), fusedOps.end(),
+            [](const std::string& ops){
+                return ops.compare("Multiply") == 0 || ops.compare("Add") == 0;
+            });
         if (additionalConfig[InferenceEngine::PluginConfigParams::KEY_ENFORCE_BF16] == InferenceEngine::PluginConfigParams::YES) {
             inType = outType = ngPrc = ElementType::bf16;
             rel_threshold = 1e-2f;
+        } else if (simple_post_ops != std::end(fusedOps)) {
+            rel_threshold = 3e-2f;
         } else {
             inType = outType = ngPrc;
         }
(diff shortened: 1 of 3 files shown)


yuxu42:
@chenhu-wang  Could you please take a review? Thanks!

chenhu-wang:
could you please update bin_cov as well here: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/nodes/bin_conv.cpp#L589-L590

src/plugins/intel_cpu/tests/functional/single_layer_tests/interpolate.cpp
```diff
@@ -331,6 +337,8 @@ const std::vector<fusingSpecificParams> interpolateFusingParamsSet{
         emptyFusingSpec,
 #if defined(OPENVINO_ARCH_X86) || defined(OPENVINO_ARCH_X86_64)
         fusingSwish,
+        fusingAddPerTensor,
+        fusingMultiplyPerTensor,
```
chenhu-wang, line 341: we need per-channel post-ops to cover this fix.

chenhu-wang: let's remove this two lines as per-tensor post-ops is already covered, which doubled smoke test time. Interpolate smoke test and nightly test is heavy.

src/plugins/intel_cpu/tests/functional/single_layer_tests/interpolate.cpp
```diff
@@ -462,13 +471,19 @@ const std::vector<ShapeParams> shapeParams4D_fixed_C = {
 };
 
 #if defined(OPENVINO_ARCH_X86) || defined(OPENVINO_ARCH_X86_64)
+const std::vector<fusingSpecificParams> interpolateFusingParamsSet_fixed_C{
+        fusingFakeQuantizePerChannelRelu,
+        fusingAddPerChannel,
+        fusingMultiplyPerChannel,
```
chenhu-wang, line 477: leave one per-channel depthwise post-ops is enough;-)

meiyang-intel:
> could you please update bin_cov as well here: https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/nodes/bin_conv.cpp#L589-L590

Done! But there is no single layer test for bin_conv. Have created a ticket to track it. https://jira.devtools.intel.com/browse/CVS-120644

chenhu-wang (approved):

dmitry-gorokhov (approved):

PR #19785, opened 2023-09-13
Check HolderTests under the proxy

ilyachur:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


src/plugins/proxy/src/compiled_model.hpp
```diff
@@ -38,7 +38,10 @@ class CompiledModel : public ov::ICompiledModel {
     }
 
     ov::Any get_property(const std::string& name) const override {
-        return m_compiled_model->get_property(name);
+        auto property = m_compiled_model->get_property(name);
+        if (!property._so)
+            property._so = m_compiled_model._so;
+        return property;
```
songbell, line 44: in which case property need the so as well?

ilyachur: Yes because proxy returns the properties of the hw compiled model.

songbell: yes I understand, I mean why do the property need the so, generally they are of common types and hardware independent. maybe I'm missing something?

ilyachur: Any contains the shared pointer this can be allocated on the plugin side.

p-durandin (approved):

ilya-lavrenov (approved):

PR #19786, opened 2023-09-13
[GPU] use input memory buffer as output memory when input1/2 are empty

wilson-seok:
### Details:
 - Use input memory buffer as output memory when input1/2 are empty

### Tickets:
 - 100495


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -761,6 +764,25 @@ bool primitive_inst::has_inner_networks() const {
     return (_impl_params->inner_nets.size() > 0);
 }
 
+void primitive_inst::do_runtime_handle_empty_inputs() {
```
yeonbok, line 767: It seems that we don't need this additional handle, how about just do on_execute() at the returning point for empty WI? on_execute is for reuse inputs for optimizable nodes and empty tensor is optimizable.

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -783,6 +805,8 @@ event::ptr primitive_inst::execute(const std::vector<event::ptr>& events) {
             return ev;
```
yeonbok, line 805: So here, 
1) if output layout is empty 
2) set prim_inst->can_be_optimized = true 
3) run on_execute()
4) set prim_inst->can_be_optimized = false

src/plugins/intel_gpu/tests/functional/single_layer_tests/dynamic/scatter_nd_update.cpp
```diff
@@ -179,12 +220,51 @@ const std::vector<ElementType> constantPrecisions = {
...
+        ScatterUpdateShapes{
+            {{-1, -1, -1, -1}, {{ 100, 256, 14, 14 }}},
+            {{-1, -1, 14, 14}, {{ 0, 256, 14, 14 }}},
+            {{-1, 256, 14, 14}, {{ 0, 256, 14, 14 }}},
+            {{1}, {{0}}}
+        },
+        IndicesValues{ 0 },
+        Scatterupdate_type::Elements
+    },
+#endif
```
yeonbok, line 244: How about separate the TC and add prefix DISABLE instead of if 0?

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
wilson-seok: Updated. Thanks!

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
wilson-seok: Updated. Thanks!

src/plugins/intel_gpu/tests/functional/single_layer_tests/dynamic/scatter_nd_update.cpp
wilson-seok: Updated. Thanks!

src/plugins/intel_gpu/src/graph/scatter_elements_update.cpp
```diff
@@ -52,5 +52,30 @@ std::string scatter_elements_update_inst::to_string(scatter_elements_update_node
 }
 
 scatter_elements_update_inst::typed_primitive_inst(network& network, scatter_elements_update_node const& node) : parent(network, node) {}
+void scatter_elements_update_inst::on_execute() {
+    auto input1_shape = _impl_params->input_layouts[1].get_partial_shape();
+    auto input2_shape = _impl_params->input_layouts[2].get_partial_shape();
+
+    if (!input1_shape.is_dynamic() && !input2_shape.is_dynamic()) {
```
yeonbok, line 59: on execute, all inputs should not dynamic, always needs to be static. So we don't need this condition.

src/plugins/intel_gpu/src/graph/scatter_elements_update.cpp
```diff
@@ -52,5 +52,30 @@ std::string scatter_elements_update_inst::to_string(scatter_elements_update_node
...
+        if ((input1_shape.get_shape()[0] == 0) || (input2_shape.get_shape()[0] == 0)) {
```
yeonbok, line 60: What is the shape is like [1, 0]? 
I think we sould check the ov::shape_size(input1_shape.to_shape())

src/plugins/intel_gpu/src/graph/scatter_elements_update.cpp
wilson-seok: Removed it. Thanks!

src/plugins/intel_gpu/src/graph/scatter_elements_update.cpp
wilson-seok: It is much better to check empty! I updated it. Thanks!

yeonbok (approved):

PR #19806, opened 2023-09-13
Remove an option to disable transformations from MOC for plugins

e-nugmanova:
### Details:
 - *Stops passing pass config from plugins inside MOC*

### Tickets:
 - *CVS-76923*

FYI @itikhono


jane-intel:
This was a showcase PR to prove that plugins depend on passing pass config to common transformations.

itikhono:
Thanks!
We will try to process this

itikhono:
build_jenkins

PR #19316, opened 2023-08-21
[Snippets] Fixed memory leak in LinearIR

a-sidorova:
### Details:
 - *Changed `shared_ptr<Expression>` in `ExpressionPort` to `weak_ptr<Expression>` to remove retain cycle in Linear IR*
image

 - *[PR](https://github.com/openvinotoolkit/openvino/pull/19317) to release branch*

### Tickets:
 - *117366*

PR #19817, opened 2023-09-13
Add `bert-base-ner` in MemLeak tests

vurusovs:
### Details:
 - Cover https://github.com/openvinotoolkit/openvino/pull/19316
 
 ### Tickets:
 - 117366

### DIFF
--- a/tests/lib/src/common_utils.h
+++ b/tests/lib/src/common_utils.h
@@ -10,6 +10,13 @@
 #include <random>
 
 
+template <typename T>
+using uniformDistribution = typename std::conditional<
+    std::is_floating_point<T>::value,
+    std::uniform_real_distribution<T>,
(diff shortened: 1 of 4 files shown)


vurusovs:
@akladiev please, note that `stress` job execution time will increase

ilyachur (commented):
Why do we need to add this model?
Maybe it can be in the nightly scope? In case if we have an impact on the pre-commit job...

vurusovs:
> Why do we need to add this model? Maybe it can be in the nightly scope? In case if we have an impact on the pre-commit job...

The motivation is to cover most actively changed features. For the moment CPU snippets are not covered (example - https://github.com/openvinotoolkit/openvino/pull/19316), dynamism also not covered

Tests execution time is increased from 5 to 10 minutes. If it's unacceptable, we can fully replace `mobilenet-ssd` with `bert-base-ner` in precommit

ilyachur (approved):

PR #19832, opened 2023-09-14
 [core] fix memory leak issue imported by #18868

tiger100256-hu:
fix the memory leak issue that CpuStreamExecuter is released, but there are still thread id in t_stream_count_map

### Details:
 - *fix memory issue in https://github.com/openvinotoolkit/openvino/commit/85609d48813ea43f4470ea27404e31a14f306fa2*

### Tickets:
 - *CVS-120233*


vurusovs:
@tiger100256-hu please, verify your changes with `recreate_and_infer_in_thread` test from StressMemLeaksTests https://github.com/openvinotoolkit/openvino/blob/master/tests/stress_tests/README.md on Windows to be sure no regressions added

src/inference/src/dev/threading/cpu_streams_executor.cpp
```diff
@@ -341,6 +339,17 @@ struct CPUStreamsExecutor::Impl {
...
+            static thread_local std::map<void*, std::shared_ptr<std::thread::id>> t_stream_count_map;
+            // fix the memory leak issue that CPUStreamsExecutor is already released,
+            // but still exists thread id in t_stream_count_map
+            for (auto it = t_stream_count_map.begin(); it != t_stream_count_map.end();) {
+                if (this != it->first && it->second.use_count() == 1) {
+                    t_stream_count_map.erase(it++);
+                } else {
+                    it++;
+                }
+            }
```
maxnick, line 352: Actually this is not a really reliable solution since it introduces coupling between this class and some external behavior (threads management). Looks like a kind of a standard problem of avoiding extra ownership of a resource. Why can't we use a weak_ptr here `std::map<void*, std::weak_ptr<std::thread::id>>`? Another question is why do we use bin tree when the key type is a 8 bit integer? Isn't it better to use hash table instead?

tiger100256-hu: Yes I agree with you，the root issue is that cpuStreamExecuter need  Impl::Stream instance,  but Impl::Stream instance has two owner, one is thread, and one is cpuStreamExecuter, one of the ower is destroyed, impl::stream should be destroyed.
the origin code use tbb thread local class to store Impl::Stream, tbb thread local is a member of cpuStreamExecuter, but when customer thread is destroyed, tbb thread local don't know it, the memory leak issue happen, this is origin pr https://github.com/openvinotoolkit/openvino/pull/19191.  At the start of the PR 19191, I try to use c++ key word thread local to store Impl::Stream, but there are lifecycle issue, cpuStreamExecuter is destroy but Impl::Stream is still in thread local.
return back to the question.
1 I can't use std::weak_ptr here, it will lost the connection between impl::stream and thread.
2 about hash table and map, below is just my opinion, maybe wrong 
the key of map is the pointer of cpuStreamExecuter::Impl, I think there will not be many cpuStreamExecuter in same thread, so I am not sure hash table is more efficient here, hash table maybe use more memory than map and customer may has many threads. 
sorry I didn't get the point of relation between 8 byte and hash table

maxnick: > 1 I can't use std::weak_ptr here, it will lost the connection between impl::stream and thread.

How exactly it loses connection? It looks like the only reason of having a pointer to `std::thread::id` is tracking dynamically changing value, but with a guaranty that it's valid. So why do we need the ownership?

> 2 about hash table and map, below is just my opinion, maybe wrong
the key of map is the pointer of cpuStreamExecuter::Impl, I think there will not be many cpuStreamExecuter in same thread, so I am not sure hash table is more efficient here, hash table maybe use more memory than map and customer may has many threads.

Ok, when the amount of elements is low, we can use bin tree.

>sorry I didn't get the point of relation between 8 byte and hash table

Hash function already exists and trivial.

tiger100256-hu:
@vurusovs  already test it on my local windows machine, StressMemLeaksTests test all pass with config
`
<cases>
    <device name="CPU" iterations="30" processes="1" threads="1">
        <model name="mobilenet-ssd" precision="FP32" source="omz" framework="caffe" subdirectory="public/mobilenet-ssd" path="C:\Users\yuanhu1\Downloads\test\mobilenet-ssd.xml"        full_path="C:\Users\yuanhu1\Downloads\test\mobilenet-ssd.xml" />
    </device>
</cases>
`

src/inference/src/dev/threading/cpu_streams_executor.cpp
tiger100256-hu: before PR19191, thread didn't hold impl::stream, so it will cause memory leak when create and destroy thread frequently
```
thread -------------------------|
                                |
                             not hold
                                |
                                v
cpuStreamExecuter ---hold---> impl::stream
```
after PR 19191, if the thread is destroyed, the count of `std::shared_ptr<std::thread::id>` will became 1 and  when  call local(), cpuStreamExecuter  will check the count of `std::shared_ptr<std::thread::id>`, if it became 1, cpuStreamExecuter will    reuse  `impl::stream` or destroy `impl::stream`
```
thread ----shared_ptr--hold -> std::thread::id
                                       ^
                                       |
    |------shared_ptr-----hold-------- |                 
    |					
cpuStreamExecuter ---hold---> impl::stream
```

maxnick: Relaying on the `std::shared_ptr ` ref counter is a really bad idea. Suppose at some point of time, in some place in the code yet another copy of the shared_ptr happens, then this assumption is broken and we will run into hardly detecting memory leak again. We need a better solution.

tiger100256-hu: @maxnick  if you have new idea or solution, I think we can talk with wanglei.
about how `std::shared_ptr<std::thread::id>` controls `impl::stream` lifetime, 
cpuStreamExecuter  use  member  `_stream_map` to hold `impl::stream`,  and the key of ` _stream_map` is the shared_ptr of `std::thread::Id`, the value of `_stream_map` is the ` impl::stream`. if the count of `std::shared_ptr<std::thread::id>`  became 1, 
local() will reuse `impl::stream` or destroy `impl::stream`, please check the code https://github.com/openvinotoolkit/openvino/pull/19191
```
thread ----shared_ptr--hold -> std::thread::id
                                       ^
                                       |	
                                    shared_ptr
                                       |		
cpuStreamExecuter --------_stream_map<key ,value>---> impl::stream
```

wenjiew:
We are being asked to port this PR to 2023.1, as the PR https://github.com/openvinotoolkit/openvino/pull/20329. We will need to discuss a proper solution if needed.

src/inference/src/dev/threading/cpu_streams_executor.cpp
maxnick: Ok, I see. But I'm still against using the std::shared_ptr ref counter as a condition to perform any actions. As I described above this counter is out of the implementation control so an error may be easily introduced.
If I correctly understand your explanation, the problem may be formulated as follows:
We need to have a mechanism which releases `impl::stream` object when the corresponding thread is destroyed. To this end you may employ some RAII technics, or indeed some sort of an atomic counter, but it shall be protected from unintentional modifications (in in opposite to `std::shared_ptr` ref counter which can be modified by coping the object).

tiger100256-hu: > We need to have a mechanism which releases impl::stream object when the corresponding thread is destroyed

yes，it is the PR [19191](https://github.com/openvinotoolkit/openvino/pull/19191) want to fix 

> std::shared_ptr ref counter which can be modified by coping the object

`std::shared_ptr<std::thread::id>` only used in local() function and t_stream_count_map.  t_stream_count_map is also defined in local(), so the count of `std::shared_ptr<std::stread::id>` only can be modified by local() function. if local() do well, there will be no issue. if there is still concern on it, I think I can add assert for it, how do you think ?
```suggestion
            for (auto it = t_stream_count_map.begin(); it != t_stream_count_map.end();) {
                if (this != it->first && it->second.use_count() == 1) {
                    t_stream_count_map.erase(it++);
                } else {
                    openvino_assert(it->second.use_count() <= 2, 
                            "the count should not be greater than 2")
                    it++;
                }
            }
```

> RALL technics

could you give more detail about this technics, what is RALL technics?

maxnick: > could you give more detail about this technics, what is RALL technics?

https://en.cppreference.com/w/cpp/language/raii

tiger100256-hu: @maxnick  @wangleis I implement  `CustomThreadLocal ::ThreadID` which is a simple version of  `shard_ptr<std::thread::id>`, please help to continue to review

src/inference/src/dev/threading/cpu_streams_executor.cpp
```diff
@@ -335,27 +334,75 @@ struct CPUStreamsExecutor::Impl {
...
+            explicit ThreadId(std::thread::id& id)
+                : _id_ptr(new std::thread::id(id)),
+                  _count_ptr(new std::atomic_int(1)) {}
+            ~ThreadId() {
+                if (_count_ptr->fetch_sub(1) == 1) {
+                    delete _id_ptr;
+                    delete _count_ptr;
+                }
+            }
+            ThreadId* copy() {
```
maxnick, line 352: Rename to `clone` and return `std::shared_ptr` instead of raw pointer to avoid any memory leaks.

src/inference/src/dev/threading/cpu_streams_executor.cpp
```diff
@@ -335,27 +334,75 @@ struct CPUStreamsExecutor::Impl {
...
+
+        private:
+            // disable all copy and move semantics, user only can use copy()
+            // to create a new instance with a shared count num;
+            ThreadId(ThreadId const&) = default;
+            ThreadId(ThreadId&&) = delete;
+            ThreadId& operator=(ThreadId const&) = delete;
+            ThreadId& operator=(ThreadId&&) = delete;
+            std::thread::id* _id_ptr;
+            std::atomic_int* _count_ptr;
```
maxnick, line 373: It is simplier to use `shared_ptr` so that to avoid a situation of dereferencing already deleted object.

src/inference/src/dev/threading/cpu_streams_executor.cpp
```diff
@@ -335,27 +334,75 @@ struct CPUStreamsExecutor::Impl {
...
+                auto pre_valule = new_ptr->_count_ptr->fetch_add(1);
+                OPENVINO_ASSERT(pre_valule == 1, "this value must be 1, please check code CustomThreadLocal::local()");
+                return new_ptr;
+            }
+            std::thread::id& get_id() {
+                return *_id_ptr;
+            }
+            int count() {
+                return *_count_ptr;
+            }
```
maxnick, line 363:
```suggestion
            int count() const {
                return *_count_ptr;
            }
```

src/inference/src/dev/threading/cpu_streams_executor.cpp
```diff
@@ -335,27 +334,75 @@ struct CPUStreamsExecutor::Impl {
     // will be counted by thread_local t_stream_count_map.
     // when the customer's thread is destoryed, the stream's count will became 1,
     // Call local() will reuse one of them, and release others.
+    // it's only a workaround for ticket CVS-111490, please be carefully when need to modify
+    // CustomeThreadLocal::local(), especially like operations that will affect the count of
+    // CustomThreadLocal::ThreadId
     class CustomThreadLocal : public ThreadLocal<std::shared_ptr<Stream>> {
+        class ThreadId {
+        public:
+            explicit ThreadId(std::thread::id& id)
```
maxnick, line 343:
```suggestion
            explicit ThreadId(const std::thread::id& id)
```

maxnick (changes requested):

src/inference/src/dev/threading/cpu_streams_executor.cpp
maxnick: May be name `fetch` would even more suitable.

src/inference/src/dev/threading/cpu_streams_executor.cpp
```diff
@@ -335,27 +334,72 @@ struct CPUStreamsExecutor::Impl {
...
+        class ThreadTracker {
+        public:
+            explicit ThreadTracker(const std::thread::id& id)
+                : _id(id),
+                  _count_ptr(std::make_shared<std::atomic_int>(1)) {}
+            ~ThreadTracker() {
+                 _count_ptr->fetch_sub(1);
+            }
+            std::shared_ptr<ThreadTracker> fetch() {
+                ThreadTracker* new_ptr = new ThreadTracker(*this);
```
maxnick, line 350:
```suggestion
                auto new_ptr = std::make_shared<ThreadTracker>(*this);
```

tiger100256-hu: I tried it but it show compile error, the compile said it's a private function

tiger100256-hu: fix it, create the shared_ptr before assert

maxnick (approved):

tiger100256-hu:
@maxnick  @wangleis  CI is already green

wangleis (approved):

PR #19839, opened 2023-09-14
[CPU] [ARM] Enable MatMul SLT tests on ARM

alvoron:


src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/common/matmul.cpp
```diff
@@ -271,7 +100,11 @@ std::vector<CPUSpecificParams> filterSpecificParams_Brgconv1x1() {
 std::vector<CPUSpecificParams> filterSpecificParams_MLAS() {
     // replace with mlas primitive type
     std::vector<CPUSpecificParams> specificParams;
+#if defined(OPENVINO_ARCH_ARM64) || defined(OPENVINO_ARCH_ARM)
```
EgorDuplensky, line 103: I think we need to have a separate instance under x86/mlas to avoid #ifdef

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/common/matmul.cpp
```diff
@@ -231,8 +57,11 @@ const std::vector<ElementType> netPRCs {
 
 std::vector<CPUSpecificParams> filterSpecificParams() {
     std::vector<CPUSpecificParams> specificParams;
+#if defined(OPENVINO_ARCH_ARM64) || defined(OPENVINO_ARCH_ARM)
```
EgorDuplensky, line 60: Can we create a filter function which will filter out extra implementations for common scope.
So we insert both "acl" and "jit_gemm" info the params list, but the filter function will remove extra params depending on the architecture.

EgorDuplensky: filterCpuInfo() {
    filterCpuInfoForArch()
    filterCPUInfoForDevice()
}

src/plugins/intel_cpu/tests/functional/test_utils/arm/cpu_test_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+std::vector<CPUSpecificParams> filterCPUInfoForArch(const std::vector<CPUSpecificParams>& CPUParams) {
+    std::vector<CPUSpecificParams> resCPUParams;
+    const int selectedTypeIndex = 3;
+
+    for (auto param : CPUParams) {
+        auto selectedTypeStr = std::get<selectedTypeIndex>(param);
+
+        if (selectedTypeStr.find("jit") != std::string::npos)
```
EgorDuplensky, line 27: I think the arm part should not care / know about x86, so it should filter out everything non-arm related (non-acl for a current state)
So if not acl (maybe also ref?) -> than remove.

src/plugins/intel_cpu/tests/functional/test_utils/cpu_test_utils.hpp
```diff
@@ -177,6 +177,8 @@ const std::map<std::string, std::string> cpuBF16PluginConfig =
 
 // utility functions
 std::vector<CPUSpecificParams> filterCPUSpecificParams(const std::vector<CPUSpecificParams>& paramsVector);
+std::vector<CPUSpecificParams> filterCPUInfo(const std::vector<CPUSpecificParams>& CPUParams);
```
EgorDuplensky, line 180: I think we should have a separate file, called filter_cpu_info.cpp and a separate header filter_cpu_info.hpp

src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
```diff
@@ -217,6 +217,8 @@ std::vector<std::string> disabledTestPatterns() {
     retVector.emplace_back(R"(smoke_NegativeQuantizedMatMulMultiplyFusion.*)");
     // int8 specific
     retVector.emplace_back(R"(smoke_Quantized.*)");
+    //ACL does not support FakeQuantize fusing for MatMul
+    retVector.emplace_back(R"(.*_(FC|MM)_.*Fused=.*FakeQuantize.*)");
```
EgorDuplensky, line 221: Why don't we make fusing instances to be x64 specific?
I guess ACL won't support this fusing in the nearest future.

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/common/matmul.cpp
alvoron: fixed

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/common/matmul.cpp
alvoron: implemented

src/plugins/intel_cpu/tests/functional/test_utils/cpu_test_utils.hpp
alvoron: implemented

src/plugins/intel_cpu/tests/functional/shared_tests_instances/skip_tests_config.cpp
alvoron: addressed

src/plugins/intel_cpu/tests/functional/test_utils/x64/filter_cpu_info.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+std::vector<CPUSpecificParams> filterCPUInfoForArch(const std::vector<CPUSpecificParams>& CPUParams) {
+    std::vector<CPUSpecificParams> resCPUParams;
+    const int selectedTypeIndex = 3;
+
+    for (auto param : CPUParams) {
+        auto selectedTypeStr = std::get<selectedTypeIndex>(param);
+
+        if (selectedTypeStr.find("acl") != std::string::npos)
```
EgorDuplensky, line 28: I think we need to have "allow" list instead of "disallow".
So, we don't need to update this code when the new backends/implementations are introduced.
But we need to have better naming for the implementations then.
This will be partially done in scope of executors refactoring.
So, I am personally ok to stick with the current implementation for now.

src/plugins/intel_cpu/tests/functional/test_utils/filter_cpu_info.hpp
```diff
@@ -0,0 +1,92 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include <vector>
+#include <string>
+
+namespace CPUTestUtils {
```
EgorDuplensky, line 10: Why do we need to move this to a new header file?

src/plugins/intel_cpu/tests/functional/test_utils/filter_cpu_info.hpp
```diff
@@ -0,0 +1,92 @@
...
+        /// 4D RNN bias tensor in the format (num_layers, num_directions,
+        /// num_gates, output_channels).
+        ///
+        ///  - For LSTM cells, the gates order is input, forget, candidate
+        ///    and output gate.
+        ///  - For GRU cells, the gates order is update, reset and output gate.
+        ldgo = abcd,
+    } cpu_memory_format_t;
+
+using CPUSpecificParams =  std::tuple<
```
EgorDuplensky, line 82: And this one

src/plugins/intel_cpu/tests/functional/test_utils/filter_cpu_info.hpp
alvoron: fixed

src/plugins/intel_cpu/tests/functional/test_utils/filter_cpu_info.hpp
alvoron: fixed

src/plugins/intel_cpu/tests/functional/single_layer_tests/instances/arm/matmul.cpp
```diff
@@ -0,0 +1,80 @@
...
+    // replace with mlas primitive type
+    std::vector<CPUSpecificParams> specificParams;
+    specificParams.push_back(CPUSpecificParams{{}, {}, {"acl"}, "acl"});
+    return specificParams;
+}
+
+std::vector<fusingSpecificParams> fusingParamsSet3D_MLAS_smoke {
+        emptyFusingSpec,
+        fusingBias,
+        fusingMultiplyPerChannel
```
EgorDuplensky, line 29: emptyFusingSpec -> common
fusingBias -> x86
fusingMultiplyPerChannel -> x86

alvoron: addressed the commend by moving MLAS specific tests to `x64\mlas` directory
MLAS specific tests should not be executed on ARM since MLAS Matmul is not supported on ARM.

alvoron:
Due to complex rebase new PR has been created: https://github.com/openvinotoolkit/openvino/pull/20923

PR #19557, opened 2023-09-01
Multinomial operator specification

mmikolajcz:
### Details:
 - *Multinomial operator specification*
 - *...*

### Tickets:
 - *118978*

PR #19866, opened 2023-09-15
[Spec][Opset13] Multinomial specification

PiotrKrzem:
### Details:
 - Add multinomial operator specification
 - Continuation of [Mateusz's PR](https://github.com/openvinotoolkit/openvino/pull/19557)

### Tickets:
 - 118978


docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,181 @@
...
+Given a list of probabilities x1,x2, ..., xn:
+* If *log_probs* is true:
+  * For each probability x, replace it with a value e^x
+* Create a cumulative sum of those probabilities, ie. create an array of values where the ith value is the sum of the probabilities x1, ..., xi
+* Divide the created array by its last (maximum) value to normalize the cumulative probabilities between [0, 1]
+* Randomly generate a sequence of floats in the range range [0,1]
+* For each generated number, assign the class with the lowest index for which the cumulative probability is less or equal the generated value.
+* If *with_replacement* is False (sampling without replacement)
+  * Set the class probability to 0 by adjusting the cumulative sum
+  * Assume class with index i has been selected - then every cdf value starting at i-th index should be lowered by the probability of the selected class
```
praasz, line 27:
```suggestion
  * Assume class with index i has been selected - then every CDF value starting at i-th index should be lowered by the probability of the selected class
```

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,181 @@
...
+  * Afterwards, divide the cdf by its last (maximum) value to normalize the cumulative probabilities between [0, 1]
```
praasz, line 28:
```suggestion
  * Afterwards, divide the CDF by its last (maximum) value to normalize the cumulative probabilities between [0, 1]
```

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,181 @@
...
+* Convert the output indices to *output_type*
+* Return output indices
+
+**Example computations**:
+Example 1 - 1D tensor
+* Let *input* = ``[0.1, 0.5, 0.4]``, *num_samples* = 5, *log_probs* = false, *replacement* = true
+* CDF of *input* = ``[0.1, 0.1 + 0.5, 0.1 + 0.5 + 0.4]`` = ``[0.1, 0.6, 1]``
```
praasz, line 35: Is there any description of this acronym `CDF`?
If not can you add it (also for any other used in this document), or add link where they are described.

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,181 @@
...
+**Detailed description**: *Multinomial* operation generates a sequence of class indices sampled from a Multinomial distribution. In this context, the *input* values represent the probabilities associated with each class within the Multinomial distribution. When performing the operation, it randomly selects a class based on these probabilities. Subsequently, the index of the chosen class in the *input* array is appended to the *output* sequence in the corresponding batch.
+
+**Algorithm formulation**:
+
+Given a list of probabilities x1,x2, ..., xn:
+* If *log_probs* is true:
+  * For each probability x, replace it with a value e^x
+* Create a cumulative sum of those probabilities, ie. create an array of values where the ith value is the sum of the probabilities x1, ..., xi
+* Divide the created array by its last (maximum) value to normalize the cumulative probabilities between [0, 1]
+* Randomly generate a sequence of floats in the range range [0,1]
```
praasz, line 23:
```suggestion
* Randomly generate a sequence of floats in the range range [0, 1]
```
Please clarify if this range must be a `float` number or can be any other type like `double`, or some `fixed-point`.
Maybe it should be mention here that it has to be any real number between [0, 1]? 
Consider to use some mathematical notation to explain the range value.

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,181 @@
...
+
+**Detailed description**: *Multinomial* operation generates a sequence of class indices sampled from a Multinomial distribution. In this context, the *input* values represent the probabilities associated with each class within the Multinomial distribution. When performing the operation, it randomly selects a class based on these probabilities. Subsequently, the index of the chosen class in the *input* array is appended to the *output* sequence in the corresponding batch.
+
+**Algorithm formulation**:
+
+Given a list of probabilities x1,x2, ..., xn:
+* If *log_probs* is true:
+  * For each probability x, replace it with a value e^x
+* Create a cumulative sum of those probabilities, ie. create an array of values where the ith value is the sum of the probabilities x1, ..., xi
+* Divide the created array by its last (maximum) value to normalize the cumulative probabilities between [0, 1]
```
praasz, line 22:
```suggestion
* Divide the created array by its maximum value to normalize the cumulative probabilities between [0, 1]
```
Will not be enough? Is important to mention that position of maximum value. Maybe it should be mention that these cumulative probabilities is sorted array?.
Add description what values type are mentioned range [0, 1], like any real value?

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,181 @@
...
+* Take the next randomly generated float, here 0.6, and repeat until all random samples have assigned classes
+* Assigned classes = ``[1, 2]``
+
+
+**Attributes**:
+
+* ``output_type``
+
+  * **Description**: the type of the output. Determines generation algorithm and affects resulting values. Output numbers generated for different values of *output_type* may not be equal.
+  * **Range of values**: "i32", "i64", "f16", "bf16", "f32", "f64".
```
praasz, line 65: Please check this PR #19830  regarding removal of f16, bf16, f64 from evaluates.
Maybe it will impact on this value type list.

docs/ops/generation/Multinomial_13.md
PiotrKrzem: Clarified to `double` precision. Added mathematical notation  [a, b] => {x ~ R, a<=x<=b}

docs/ops/generation/Multinomial_13.md
PiotrKrzem: Updated!

docs/ops/generation/Multinomial_13.md
PiotrKrzem: Updated!

docs/ops/generation/Multinomial_13.md
PiotrKrzem: Removed the aforementioned values, updated the spec & ref

docs/ops/generation/Multinomial_13.md
```diff
@@ -62,7 +62,7 @@ Example 3 - 1D tensor, without replacement
 * ``output_type``
 
   * **Description**: the type of the output. Determines generation algorithm and affects resulting values. Output numbers generated for different values of *output_type* may not be equal.
-  * **Range of values**: "i32", "i64", "f16", "bf16", "f32", "f64".
+  * **Range of values**: "i4", "i8", "i16", "i32", "i64", "f32", "u4", "u8", "u16", u32", "u64".
```
praasz, line 65: Just curious. Will be there dedicated implementation for `binary` precision `i4`, `u4` in reference implementation?

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,181 @@
...
+
+*   **2**: ``num_samples`` - scalar or 1D tensor with 1 element of type `T_SAMPLES` specifying the number of samples to draw from Multinomial distribution. **Required.**
+
+**Outputs**:
+
+* **1**:  ``output`` A tensor with type specified by the attribute *output_type* and shape depending on the rank of *input*, either ``[num_samples]`` for one-dimensional *input* or ``[batch_size, num_samples]`` for the two-dimensional one.
+
+**Types**
+
+* **T_IN**: 32-bit floating point numbers.
```
mitruska, line 115: Is there specific reason for 32-bit float restriction for the probs input type?
If not, let's keep it as usual:
```suggestion
* **T_IN**: any supported floating-point type.
```
As the supported types are often plugin or implementation dependent, the specification can be more generic.
Also for example ONNX Multinomial supports `tensor(double), tensor(float), tensor(float16)` as the first input type, so with fp32 restriction additional Convert would be needed to align them.

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,181 @@
...
+
+  * **Description**: operational seed value.
+  * **Range of values**: non-negative integers
+  * **Type**: `unsigned int 64-bit`
+  * **Default value**: 0
+  * **Required**: *Yes*
+
+**Inputs**:
+
+*   **1**: ``input`` - 1D or 2D tensor of type `T_IN` and shape `[class_size]` or `[batch_size, class_size]` with probabilities. Allowed values depend on the *log_probs* attribute. The values are internally normalized to have values in the range of `[0, 1]` with the sum of all probabilities in the given batch equal to 1. **Required.**
```
mitruska, line 105: Is there any more specific name suitable to be assigned as the "input" name like "probs"?
It's a detail, but input with name "input" when there are actually two inputs is a little bit misleading;)

docs/ops/generation/Multinomial_13.md
PiotrKrzem: Afaik no, the conversion is simply applied to the output indices at the last step, so I believe it should work just fine (as long as no overflow is present)

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,181 @@
...
+  * **Default value**: 0
+  * **Required**: *Yes*
+
+* ``op_seed``
+
+  * **Description**: operational seed value.
+  * **Range of values**: non-negative integers
+  * **Type**: `unsigned int 64-bit`
+  * **Default value**: 0
+  * **Required**: *Yes*
```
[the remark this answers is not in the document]
PiotrKrzem: Right, updated!

docs/ops/generation/Multinomial_13.md
PiotrKrzem: I changed it to `f32` because of the PR Paweł mentioned https://github.com/openvinotoolkit/openvino/pull/19830
If I understand correctly, it is supposed to remove support for floating-point type inputs `f64`, `f16` and `bf16`, so the only one remaining would be `f32`

docs/ops/generation/Multinomial_13.md
PiotrKrzem: That's true, `probs` is a great idea!

docs/ops/generation/Multinomial_13.md
praasz: The mention PR removes from operators evaluate these precisions as it is used for shape_inference, constant folding or sometimes as fallback path for data inference.
The operator can still provide support for these precisions but maybe is not required to provide them into op's evaluate (plugin can still implement it).

This removal is more like trigger to double check if these precision should be supported and if yes it can be still added in specification.

praasz (approved):

mmikolajcz (changes requested):
In built doc there is few formatting issues

docs/ops/generation/Multinomial_13.md
mitruska: Also as index value is always an integer, it seems like supporting "i32" and "i64" as usual for such `output_type` attribute seems to be enough. Example in [Unique_10.md](https://github.com/openvinotoolkit/openvino/blob/master/docs/ops/movement/Unique_10.md) spec.
Then conversion of this values to any other type is possible as additional step.
But it can be easier to clarify that a little bit later, based on the tests and implementation.

docs/ops/opset13.md
```diff
@@ -121,6 +121,9 @@ Table of Contents
 * :doc:`MulticlassNMS <openvino_docs_ops_sort_MulticlassNonMaxSuppression_9>`
 * :doc:`Multiply <openvino_docs_ops_arithmetic_Multiply_1>`
 * :doc:`Negative <openvino_docs_ops_arithmetic_Negative_1>`
+* :doc:`Multiply <openvino_docs_ops_arithmetic_Multiply_1>`
+* :doc:`Multinomial <openvino_docs_ops_generation_Multinomial_13>`
+* :doc:`Negative <openvino_docs_ops_arithmetic_Negative_1>`
```
mmikolajcz, line 126:
```suggestion
* :doc:`Multinomial <openvino_docs_ops_generation_Multinomial_13>`
* :doc:`Multiply <openvino_docs_ops_arithmetic_Multiply_1>`
* :doc:`Negative <openvino_docs_ops_arithmetic_Negative_1>`
```
Repeated Multiply and Negative

PiotrKrzem: Great catch!

docs/ops/generation/Multinomial_13.md
PiotrKrzem: Removed other values except for i32 and i64, as it also helps with Paweł's comment regarding overload of template implementations. Updated the spec!

mlukasze (approved):

mlukasze:
> In built doc there is few formatting issues

@mmikolajcz 
I believe we can assume that issue has been solved, and comment could be marked as resolved as well?

mmikolajcz (approved):

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,200 @@
...
+
+  * Assume a class with index i has been selected - then every CDF value starting at i-th index should be lowered by the original probability of the selected class. This effectively sets the probability of sampling the given class to 0.
+  * Afterwards, divide the CDF by its last (maximum) value to normalize the cumulative probabilities between the real values in the range [0, 1]
+
+* Convert the output indices to *convert_type*
+* Return output indices
+
+**Example computations**:
+Example 1 - 1D tensor
+
```
tsavina, line 44:
```suggestion
**Example computations**:

Example 1 - 1D tensor

```

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,200 @@
...
+  * **Type**: string
+  * **Required**: *Yes*
+
+* ``with_replacement``
+
+  * **Description**: controls whether to sample with replacement (classes can be sampled multiple times).
+  * **Range of values**: `true`, `false`
+  
+      * ``true`` - class indices can be sampled multiple times.
+      * ``false`` - class indices will not repeat in the output and the size of ``probs``' ``class_size`` dimension is required to be larger or equal to *num_samples* value. Might affect performance.
```
tsavina, line 90:
```suggestion
    * ``true`` - class indices can be sampled multiple times.
    * ``false`` - class indices will not repeat in the output and the size of ``probs``' ``class_size`` dimension is required to be larger or equal to *num_samples* value. Might affect performance.
```

docs/ops/generation/Multinomial_13.md
```diff
@@ -0,0 +1,200 @@
...
+  * **Type**: `bool`
+  * **Required**: *Yes*
+
+* ``log_probs``
+
+  * **Description**: allows to control whether *inputs* should be treated as probabilities or unnormalized log probabilities.
+  * **Range of values**: `true`, `false`
+
+      * ``true`` - set values in *inputs* are unnormalized log probabilities that can be any real number.
+      * ``false`` - probabilities in *inputs* are expected to be non-negative, finite and have a non-zero-sum.
```
tsavina, line 101:
```suggestion
    * ``true`` - set values in *inputs* are unnormalized log probabilities that can be any real number.
    * ``false`` - probabilities in *inputs* are expected to be non-negative, finite and have a non-zero-sum.
```

PR #19873, opened 2023-09-15
fix compilation issue on Windows arm64

xufang-lisa:
### Details:
 - *fix compilation issue on Windows arm64*

### Tickets:
 - *ticket-id*


src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -519,7 +519,11 @@ Snippet::SnippetJitExecutor::SnippetJitExecutor(const SnippetAttrs& attrs, bool
         snippet_for_generation->set_friendly_name(snippetAttrs.snippet->get_friendly_name());
         auto host_isa = dnnl::impl::cpu::x64::mayiuse(dnnl::impl::cpu::x64::avx512_core) ?
             dnnl::impl::cpu::x64::avx512_core : dnnl::impl::cpu::x64::avx2;
+#if defined(OPENVINO_ARCH_X86_64)
```
ilya-lavrenov, line 522: why is it Windows arm64 specific?

xufang-lisa: CPUGenerator is excluded on ARM64 by the code (https://github.com/openvinotoolkit/openvino/blob/2cd130810432662f9974521e76418d876ef470e1/src/plugins/intel_cpu/CMakeLists.txt#L81)

src/plugins/intel_cpu/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ elseif(AARCH64)
 endif()
 set(OV_CPU_ARM_TARGET_ARCH ${OV_CPU_ARM_TARGET_ARCH_DEFAULT})
 
-if(X86 OR X86_64 OR AARCH64)
+if(X86 OR X86_64 OR (AARCH64 AND NOT WIN32))
```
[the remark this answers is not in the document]
xufang-lisa: I use MSVC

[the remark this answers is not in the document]
ilya-lavrenov: have you been able to compile ARM Compute library with MSVC?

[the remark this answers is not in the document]
ilya-lavrenov: Let's try to compile everything with clang. And if it passes, we need to exclude MLAs only for MSVC compiler instead of whole WIN32 condition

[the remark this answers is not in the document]
xufang-lisa: Ok, I will try it.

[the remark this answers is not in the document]
xufang-lisa: @ilya-lavrenov do you know how to compile openvino with clang, I set "CC=clang && CXX=clang++", then run command "cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ..", encountered the following error "FAILED: C:/Users/openvino-ci-64/xufang/openvino-arm64/openvino/bin/arm64/Release/protoc.exe"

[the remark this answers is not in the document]
ilya-lavrenov: looks like codebase is updated and we need to apply more fixes to make a compilation phase passing.
Regarding MLAS - I also failed to compile it with Clang-CL on Windows, so let's disable it at all for now.

cmake/developer_package/IEDevScriptsConfig.cmake
```diff
@@ -39,6 +39,10 @@ set_ci_build_number()
 
 include(target_flags)
 
+if(WIN32 AND AARCH64 AND OV_COMPILER_IS_CLANG)
+    message(FATAL_ERROR "For Windows on ARM, Clang compiler is not supported at the moment, please use MSVC as compiler")
+endif()
```
ilya-lavrenov, line 44: I don't think we need such global error, maybe is it enough to disable CPU plugin in this case?

src/plugins/intel_cpu/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ elseif(AARCH64)
...
     # disable mlas with webassembly
     if(EMSCRIPTEN OR MINGW OR (CMAKE_COMPILER_IS_GNUCXX AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7))
```
ilya-lavrenov, line 39:
```suggestion
    if(EMSCRIPTEN OR WIN32 OR MINGW OR (CMAKE_COMPILER_IS_GNUCXX AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7))
```
maybe it's better to move the check here?

xufang-lisa: Doesn't Windows on x86 support mlas?

ilya-lavrenov: corrected

Maybe we need to refactor these conditions to have 1 level of conditons..

cmake/developer_package/IEDevScriptsConfig.cmake
xufang-lisa: The compiler check is moved to CMakeLists.txt in CPU plugin directory.

src/plugins/intel_cpu/src/nodes/subgraph.cpp
```diff
@@ -693,12 +698,12 @@ bool Snippet::SnippetJitExecutor::optimizeExecDomain(std::vector<VectorDims>& in
 
 void Snippet::SnippetJitExecutor::generate(const jit_snippets_compile_args* jcp) {
     using Manager = snippets::pass::Manager;
+    std::vector<Manager::PositionedPass> backend_passes;
+#if defined(OPENVINO_ARCH_X86_64)
```
a-sidorova, line 702:
```suggestion
#if defined(OPENVINO_ARCH_X86_64)
#define SNIPPETS_REGISTER_PASS(PASS_POS, PASS, ...) \
        backend_passes.emplace_back(PASS_POS, std::make_shared<PASS>(__VA_ARGS__))
#else
#define SNIPPETS_REGISTER_PASS(PASS_POS, PASS, ...)
#endif // OPENVINO_ARCH_X86_64

SNIPPETS_REGISTER_PASS(PassPosition(Place::PipelineStart), ConvertToSwishCPU);
....
#undef SNIPPETS_REGISTER_PASS
```

Probably, it's a matter of taste. But for me it looks better.

xufang-lisa: updated

a-sidorova (approved):

PR #19880, opened 2023-09-15
Add EnableShapeOfConstantFolding transformation

mateusztabaka:
Transpose sinking (that is used in TF frontend) disables ShapeOf constant folding which prevents some optimizations further in the pipeline. This patch introduces EnableShapeOfConstantFolding that removes DisableConstantFolding from ShapeOf nodes.

Ticket: CVS-118890


src/common/transformations/include/transformations/common_optimizations/enable_shapeof_constant_folding.hpp
```diff
@@ -0,0 +1,20 @@
...
+//
+
+#pragma once
+
+#include <openvino/pass/graph_rewrite.hpp>
+#include <transformations_visibility.hpp>
+
+namespace ov {
+namespace pass {
+
```
evkotov, line 12: Could you please add some comments here about transformation. What does it actually do? Maybe why we need it? Any use cases? Just in a couple of words.

src/common/transformations/src/transformations/common_optimizations/enable_shapeof_constant_folding.cpp
```diff
@@ -0,0 +1,23 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <ngraph/pattern/op/wrap_type.hpp>
```
itikhono, line 5: please use openvino headers

evkotov (approved):

src/common/transformations/include/transformations/common_optimizations/enable_shapeof_constant_folding.hpp
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include <openvino/pass/graph_rewrite.hpp>
```
itikhono, line 7: minor: the same about <

src/common/transformations/src/transformations/transpose_sinking/ts_general.cpp
```diff
@@ -77,6 +78,7 @@ bool TSGeneral::run_on_model(const std::shared_ptr<ov::Model>& f) {
         manager.register_pass<TSGeneralBackward>();
```
itikhono, line 78: Should we disable ShapeOf here? in Backward transformations batch I mean

itikhono (approved):

src/common/transformations/src/transformations/transpose_sinking/ts_general.cpp
mateusztabaka: I can see that DisableShapeOfConstantFolding is registered at line 76.

itikhono: I mean do we actually need it? or should we run it once in Forward transformations?

PR #19887, opened 2023-09-15
[template] Add faked hint property

sbalandi:
### Details:
 - *added faked props `ov::hint::inference_precision` `ov::hint::execution_mode` `ov::hint::num_requests`*
 - *fixed default `perf_count` value, it is False now*
 - *fixed test for property ov::hint::model_priority in apiConformance. Property was tested by core->get_property in apiConformance , but it is implemented as compiled_model property, so this test is removed now . Another existing test, where getting property tested with compiled_model->get_property, added to apiConformance scope. Test with setting this property will be added after with task CVS-109419*

### Tickets:
 - *CVS-103948*


ilya-lavrenov (approved):

ilyachur (approved):
Please fix code style

iefode (approved):

PR #19888, opened 2023-09-15
[PT FE] Add tests for detectron2 models

mvafin:
### Details:
 - *Add tests for detectron2 models*

### Tickets:
 - *120265*


ilya-lavrenov (approved):

tests/model_hub_tests/torch_tests/test_detectron2.py
```diff
@@ -0,0 +1,95 @@
...
+import pytest
+import torch
+import tempfile
+import torchvision.transforms.functional as F
+from models_hub_common.test_convert_model import TestConvertModel
+from openvino import convert_model
+from models_hub_common.utils import get_models_list, compare_two_tensors
+
+
+class TestTorchHubConvertModel(TestConvertModel):
```
mvafin, line 14:
```suggestion
class TestDetectron2ConvertModel(TestConvertModel):
```

tests/model_hub_tests/torch_tests/test_detectron2.py
```diff
@@ -0,0 +1,95 @@
...
+            l = min(len(cur_fw_res), len(cur_ov_res))
+            assert l > 0 or len(cur_fw_res) == len(cur_ov_res), "No boxes were selected."
+            print(f"fw_re: {cur_fw_res};\n ov_res: {cur_ov_res}")
+            is_ok = compare_two_tensors(cur_ov_res[:l], cur_fw_res[:l], fw_eps)
+        assert is_ok, "Accuracy validation failed"
+
+    @pytest.mark.parametrize("name,type,mark,reason",
+                             get_models_list(os.path.join(os.path.dirname(__file__), "detectron2_precommit")))
+    @pytest.mark.precommit
+    def test_convert_model_all_models(self, name, type, mark, reason, ie_device):
```
mvafin, line 88:
```suggestion
    def test_detectron2_all_models(self, name, type, mark, reason, ie_device):
```

tests/model_hub_tests/torch_tests/test_detectron2.py
```diff
@@ -0,0 +1,95 @@
...
+        self.run(name, None, ie_device)
+
+    @pytest.mark.parametrize("name,type,mark,reason",
+                             get_models_list(os.path.join(os.path.dirname(__file__), "detectron2_models")))
+    @pytest.mark.nightly
+    def test_convert_model_all_models(self, name, type, mark, reason, ie_device):
```
mvafin, line 94:
```suggestion
    def test_detectron2_all_models(self, name, type, mark, reason, ie_device):
```

.github/workflows/linux.yml
```diff
@@ -924,8 +924,10 @@ jobs:
         run: |
           python3 -m pip install openvino --find-links=${{ env.INSTALL_DIR }}/tools
       - name: PyTorch Models Tests
+        # detectron2 requires torch to be alredy installed
         run: |
           python3 -m pip install -r ${{ env.MODEL_HUB_TESTS_INSTALL_DIR }}/torch_tests/requirements.txt
+          python3 -m pip install git+https://github.com/facebookresearch/detectron2.git
```
akladiev, line 930: Did you consider adding this dependency to a requirement file? https://pip.pypa.io/en/stable/topics/vcs-support

akladiev: You can still create a separate requirement file for detectron (like `requirements_detectron.txt`). It may be useful for local runs, especially if more dependencies will be added or if we'll have to pin a version of this dependency due to some regression (people won't need to look up a working version in the pipeline)

tests/model_hub_tests/models_hub_common/utils.py
```diff
@@ -32,7 +32,8 @@ def compare_two_tensors(ov_res, fw_res, eps):
     is_ok = True
     if not np.allclose(ov_res, fw_res, atol=eps, rtol=eps, equal_nan=True):
         is_ok = False
-        print("Max diff is {}".format(np.array(abs(ov_res - fw_res)).max()))
+        max_diff = np.abs(ov_res.astype(np.float32) - fw_res.astype(np.float32)).max()
```
rkazants, line 35: Disagree with this change, let us check that types of both outputs are the same before computation of `max_diff`

rkazants (changes requested):
Please revert add check that both results `ov_res` and `fw_res` to be the same

tests/model_hub_tests/models_hub_common/utils.py
mvafin: That is not what this code does. It just reports max diff when np.allclose failed and this will fail for bool outputs, because `-` is not supported for bools

tests/model_hub_tests/torch_tests/test_detectron2.py
```diff
@@ -0,0 +1,95 @@
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import pytest
+import torch
+import tempfile
```
rkazants, line 7: Looks redundant
```suggestion
```

tests/model_hub_tests/torch_tests/test_detectron2.py
```diff
@@ -0,0 +1,95 @@
...
+import torchvision.transforms.functional as F
```
rkazants, line 8: not used
```suggestion
```

rkazants (approved):

.github/workflows/linux.yml
mvafin: done

PR #19900, opened 2023-09-18
Introduce NF4 data type

andreyanufr:
Support of nf4 type for quantization.


src/bindings/c/include/c_api/ie_c_api.h
```diff
@@ -73,1092 +73,1128 @@
...
- * @param device_name A name of the device to load the network to.
- * @param config Device configuration.
- * @param exe_network A pointer to the newly created executable network.
- * @return Status code of the operation: OK(0) for success.
- */
-INFERENCE_ENGINE_C_API(IE_NODISCARD IEStatusCode)
-ie_core_import_network_from_memory(ie_core_t* core,
-                                   const uint8_t* content,
-                                   size_t content_size,
+            typedef struct ie_network ie_network_t;
```
AlexKoff88, line 76: @andreyanufr, can you check the code alignment in this file? It does not seem correct. Git cannot reflect the changes as well.

src/core/include/openvino/core/type/nf4.hpp
```diff
@@ -0,0 +1,47 @@
...
+        uint8_t val = quantize_nf4(static_cast<float>(src[idx]));
+        const size_t byte_idx = idx / 2;
+        const uint8_t bit_shift = 4 * (++idx % 2);
+        dst[byte_idx] &= ~(0xF << bit_shift);         // half byte zeroed
+        dst[byte_idx] |= ((val & 0xF) << bit_shift);  // set 1's
+    }
+
+    template <typename T, typename std::enable_if<std::is_integral<T>::value, bool>::type = true>
+    static void pack(uint8_t* dst, T* src, std::size_t idx) {
+        uint8_t val = static_cast<uint8_t>(src[idx]);
```
AlexKoff88, line 35: do we need type cast to nf4 here?

src/bindings/c/include/c_api/ie_c_api.h
```diff
@@ -73,1092 +73,1128 @@
 #endif
 
 typedef struct ie_core ie_core_t;
-typedef struct ie_network ie_network_t;
```
ilya-lavrenov, line 76: please, revert most of the changes in this file.

src/core/reference/include/openvino/reference/convert.hpp
```diff
@@ -70,6 +71,11 @@ TO get_value(const uint8_t* buf, size_t idx, element::Type from_type) {
     if (from_type == element::i4) {
         return detail::get_i4(buf, idx);
     }
+
+    if (from_type == element::nf4) {
+        return detail::get_u4(buf, idx);
```
ilya-lavrenov, line 76: do we need get_nf4 ?

src/core/src/type/nf4.cpp
```diff
@@ -0,0 +1,81 @@
...
+// LICENSE file in the root directory of this source tree.
+//==============================================================================
+
+#include "openvino/core/type/nf4.hpp"
+
+using namespace ov;
+
+float ConvertNF4::dequantize_nf4(uint8_t val) {
+    static float lookup[16] = {-1.0,
+                               -0.6961928009986877,
```
ilya-lavrenov, line 23:
```suggestion
                               -0.6961928009986877f,
```

src/core/include/openvino/core/type/nf4.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <iostream>
+#include <limits>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "openvino/core/core_visibility.hpp"
+#include "openvino/core/type/float16.hpp"
+
+#define ROUND_MODE_TO_NEAREST_EVEN
```
ilya-lavrenov, line 17: why do we need it?

src/core/include/openvino/core/type/nf4.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace ov {
+class OPENVINO_API ConvertNF4 {
```
ilya-lavrenov, line 20: do we need it in public API?
I see it's used internally only. So, proper place is this folder https://github.com/openvinotoolkit/openvino/tree/0915a150c38fa3c29d08fa99ea0b88ef9776f4d1/src/core/dev_api

src/core/include/openvino/core/type/nf4.hpp
```diff
@@ -0,0 +1,47 @@
...
+public:
+    ConvertNF4() = default;
+
+    template <typename T, typename std::enable_if<!std::is_integral<T>::value, bool>::type = true>
+    static void pack(uint8_t* dst, T* src, std::size_t idx) {
+        uint8_t val = quantize_nf4(static_cast<float>(src[idx]));
+        const size_t byte_idx = idx / 2;
+        const uint8_t bit_shift = 4 * (++idx % 2);
+        dst[byte_idx] &= ~(0xF << bit_shift);         // half byte zeroed
+        dst[byte_idx] |= ((val & 0xF) << bit_shift);  // set 1's
```
praasz, line 30: The `dst` can be int8_t but the `val` is `uint8_t`, consider to add explicit cast as conversion is intentional and there is also risk of compile warning regarding conversion and precision lost.

src/core/src/type/nf4.cpp
```diff
@@ -0,0 +1,81 @@
...
+                               0.33791524171829224,
+                               0.44070982933044434,
+                               0.5626170039176941,
+                               0.7229568362236023,
+                               1.0};
+    return lookup[val];
+}
+
+uint8_t ConvertNF4::quantize_nf4(float x) {
+    if (x > 0.03979014977812767f)
```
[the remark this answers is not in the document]
AlexKoff88: @praasz, thanks for the hint. This is a reference implementation taken from the framework. I would like to keep it the same.

src/core/include/openvino/core/type/nf4.hpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    template <typename T, typename std::enable_if<std::is_integral<T>::value, bool>::type = true>
+    static void pack(uint8_t* dst, T* src, std::size_t idx) {
+        uint8_t val = static_cast<uint8_t>(src[idx]);
+        const size_t byte_idx = idx / 2;
+        const uint8_t bit_shift = 4 * (++idx % 2);
```
AlexKoff88, line 37: Let's change the order of elements in the byte so that the zero element will be in the first place and the second on on the second place.

AlexKoff88:
Guys, thanks for the comments. We need this to be integrated ASAP because there is a huge interest in this feature. And since the second part of the contribution is NNCF PR  that depends on this, we need this feature to fall into the next prerelease.

src/core/include/openvino/core/type/nf4.hpp
andreyanufr: yes we need this to reduce the number of warnings

src/core/reference/include/openvino/reference/convert.hpp
andreyanufr: Removed this code for changing the order of elements in packing.

src/core/src/type/nf4.cpp
andreyanufr: Done

src/core/src/type/nf4.cpp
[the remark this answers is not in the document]
AlexKoff88: btw, the plan is to introduce optimizations in the follow-up PR. But this is out of the scope of our team.

src/core/src/type/nf4.cpp
```diff
@@ -0,0 +1,81 @@
...
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//==============================================================================
+
+#include "openvino/core/type/nf4.hpp"
+
+using namespace ov;
+
+float ConvertNF4::dequantize_nf4(uint8_t val) {
```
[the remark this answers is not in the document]
andreyanufr: @praasz What do you mean on:  " Then literals can be converted during compile time." ?

src/core/src/type/nf4.cpp
```diff
@@ -0,0 +1,81 @@
...
+    static float lookup[16] = {-1.0,
```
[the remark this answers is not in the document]
andreyanufr: Done

src/core/include/openvino/core/type/nf4.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include "openvino/core/core_visibility.hpp"
+#include "openvino/core/type/float16.hpp"
+
+#define ROUND_MODE_TO_NEAREST_EVEN
+
+namespace ov {
+class OPENVINO_API ConvertNF4 {
+public:
+    ConvertNF4() = default;
```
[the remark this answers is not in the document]
andreyanufr: We are not sure about non static class members now, but may be in the future we will extend class functionality with other fp4 types and these changes may require class members.

src/core/include/openvino/core/type/nf4.hpp
```diff
@@ -0,0 +1,47 @@
...
+    template <typename T, typename std::enable_if<std::is_integral<T>::value, bool>::type = true>
+    static void pack(uint8_t* dst, T* src, std::size_t idx) {
+        uint8_t val = static_cast<uint8_t>(src[idx]);
+        const size_t byte_idx = idx / 2;
+        const uint8_t bit_shift = 4 * (++idx % 2);
+        dst[byte_idx] &= ~(0xF << bit_shift);         // half byte zeroed
+        dst[byte_idx] |= ((val & 0xF) << bit_shift);  // set 1's
+    }
+
+    static float dequantize_nf4(uint8_t val);
```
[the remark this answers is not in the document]
andreyanufr: Done

src/core/include/openvino/core/type/nf4.hpp
andreyanufr: Removed this function.

src/core/src/type/nf4.cpp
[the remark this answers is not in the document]
praasz: If function is `constexpr` then using it like `auto y =  ConvertNF4::dequantize_nf4(0)` can result as `float y = 0.0f` and be resolved during compile time as value (no runtime calculations). The runtime calculation will be preformed only for values not know at compile time.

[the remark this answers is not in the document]
andreyanufr: I think all quantize/dequantize will work only on runtime and only on the tests the usage of ConvertNF4::dequantize_nf4(0) is possible. Do you think that tests it is the case for constexpr usage ?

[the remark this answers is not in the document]
praasz: This was just suggestion that maybe this function can work on compile time.. But as you comment if mainly used on runtime so maybe is not worthy to spent time to refactor it.

src/core/reference/include/openvino/reference/convert.hpp
```diff
@@ -8,6 +8,7 @@
 
 #include "ngraph/type/element_type.hpp"
 #include "ngraph/type/float16.hpp"
+#include "ngraph/type/nf4.hpp"
```
praasz, line 11: Can `openvino` include be used here instead `ngraph` for `nf4`?

src/core/src/type/nf4.cpp
```diff
@@ -0,0 +1,81 @@
...
+// implementation.
+// Copyright notice from original source file is as follows.
+
+//*******************************************************************************
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//==============================================================================
+
```
mlukasze, line 16: please do not forget to add information about 3rd party code into 
https://github.com/openvinotoolkit/openvino/blob/master/licensing/third-party-programs.txt
and report it to OS PDT for the review.

src/plugins/intel_cpu/thirdparty/onednn
ilya-lavrenov, line 1: should we revert submodule update?

src/core/src/type/nf4.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "openvino/core/type/nf4.hpp"
+
+using namespace ov;
+
+float ConvertNF4::dequantize(uint8_t val) {
+    static std::array<float, 16> lookup = {-1.0f,
```
ilya-lavrenov, line 22:
```suggestion
    static const std::array<float, 16> lookup = {-1.0f,
```

src/core/reference/include/openvino/reference/convert.hpp
andreyanufr: Done

src/bindings/python/tests/test_transformations/test_compression_4bit.py
```diff
@@ -0,0 +1,50 @@
...
+              -0.18477343022823334, -0.09105003625154495,
+              0.0, 0.07958029955625534, 0.16093020141124725,
+              0.24611230194568634, 0.33791524171829224,
+              0.44070982933044434, 0.5626170039176941,
+              0.7229568362236023, 1.0]
+    target = np.array(target)
+    
+    diff = np.max(np.abs(target - uniq))
+    
+    assert diff < 0.001
```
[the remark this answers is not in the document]
andreyanufr: Is it better to remove this test and add extra C++ test?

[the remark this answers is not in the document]
ilya-lavrenov: yes, let's create s file similar to https://github.com/openvinotoolkit/openvino/blob/master/src/core/tests/uint4.cpp and put tests there

src/core/include/openvino/core/type/nf4.hpp
andreyanufr: Done

andreyanufr:
@slyalin , @dmitry-gorokhov , @jane-intel can you please look at the PR?

src/core/tests/nf4.cpp
```diff
@@ -0,0 +1,98 @@
...
+                     0.16093020141124725,
+                     0.24611230194568634,
+                     0.33791524171829224,
+                     0.44070982933044434,
+                     0.5626170039176941,
+                     0.7229568362236023,
+                     1.0};
+
+    auto model = get_model<T>(const_data, ov_type);
+    ov::CompiledModel compiled_model = core.compile_model(model, "CPU");
```
ilya-lavrenov, line 69: we should not perform inference on CPU.
These are unit tests for ngraph component, which should check only primitives you added - conversion from nf4 to float and back.

thirdparty/open_model_zoo
ilya-lavrenov, line 1: please, revert

src/core/include/openvino/op/constant.hpp
```diff
@@ -670,6 +682,50 @@ class OPENVINO_API Constant : public Op {
...
+              typename StorageDataType = fundamental_type_for<Type>,
+              typename std::enable_if<Type == element::Type_t::nf4 &&
+                                          (std::is_floating_point<T>::value || std::is_same<T, bfloat16>::value ||
+                                           std::is_same<T, float16>::value),
+                                      bool>::type = true>
+    void write_buffer(const std::vector<T>& source) {
+        auto p = get_data_ptr_nc<Type>();
+        size_t i = 0;
+        for (; i < source.size() / 2; i++) {
+            const auto idx1 = ConvertNF4::quantize(static_cast<float>(source[i * 2]));
```
praasz, line 715: Just detail if there will be any change consider rename, the `idx` name as it is not index but quantized value.

praasz (approved):

AlexKoff88 (approved):

src/core/include/openvino/op/constant.hpp
andreyanufr: Ok, in the next PR I will change it. Is name like qnt1, qnt2 appropriate ?

PR #19903, opened 2023-09-18
tf.Graph decoder accuracy fixes

popovaan:
Description:

- Fixed output names passed from tf.Graph decoder. 
- Fixed order of internal names in output_names_map.
- Added support of variables of non-resource type.
- Removed usage of _name attribute for comparison of EagerTensor's.

Ticket: 119718

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update


src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -25,17 +25,24 @@ std::vector<T> reorder_ops_by_names(const std::vector<std::string>& names, const
...
+
+    size_t ind = 0;
+    for (const auto& name : names) {
+        for (const auto& op : ops) {
+            if (op->get_friendly_name() == name) {
+                resulted_ops[ind] = op;
+                ind++;
+                break;
+            }
+        }
```
rkazants, line 45: Not sure about this change. How is it different from the current version?

rkazants (approved):

src/frontends/tensorflow/src/translate_session.cpp
rkazants: Please refactor this code a bit in the next PR as discussed offline

PR #19913, opened 2023-09-18
[core] API 2.0/Migrate LogicalXor to new API

praasz:
### Details:
 - Migrate `LogicalXor` operator (v0, v1) to new API. Remove legacy, headers, functions, includes. Modify for bin size reduction.
 - Update the reference implementation to reduce binary size.

### Tickets:
 - [CVS-118685](https://jira.devtools.intel.com/browse/CVS-118685)

### DIFF
--- a/src/core/src/op/xor.cpp
+++ b/src/core/src/op/xor.cpp
@@ -2,105 +2,103 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-#include "ngraph/op/xor.hpp"
+#include "openvino/op/xor.hpp"
 
+#include "element_visitor.hpp"
 #include "itt.hpp"
-#include "ngraph/runtime/host_tensor.hpp"
-#include "ngraph/validation_util.hpp"
+#include "openvino/op/logical_xor.hpp"
 #include "openvino/reference/xor.hpp"
+#include "shape_util.hpp"
 
-using namespace std;
-using namespace ngraph;
-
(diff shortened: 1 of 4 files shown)


t-jankowski (approved):

p-durandin (approved):

src/core/src/op/xor.cpp
```diff
@@ -2,105 +2,103 @@
...
-    default:
-        rc = false;
-        break;
-    }
-    return rc;
+bool evaluate(TensorVector& outputs, const TensorVector& inputs, const AutoBroadcastSpec& broadcast_spec) {
+    OPENVINO_ASSERT(outputs.size() == 1);
+    OPENVINO_ASSERT(inputs.size() == 2);
+
+    outputs[0].set_shape(ov::util::get_broadcast_shape(inputs[0].get_shape(), inputs[1].get_shape(), broadcast_spec));
```
jane-intel, line 44: can we reuse shape infer?

jane-intel (approved):

src/core/src/op/xor.cpp
praasz: Will check.

praasz: I think it can be replaced. 
In #19984 will add helper function which will work on ov::Shape and use shape inference to calculate shape. THen will replace get_broadcast_shape function.

PR #19921, opened 2023-09-18
[GPU] updated fully_connected_gpu_bf_tiled to use plain format weight

e-ddykim:
### Details:
 - This PR modifies the `fully_connected_gpu_bf_tiled` kernel to use plain format weight.
 - The purpose of this PR is to prevent runtime weight reordering when dynamic models are executed on dGPU.

### Tickets:
 - 119159

### DIFF
--- a/src/plugins/intel_gpu/src/kernel_selector/kernels/fully_connected/fully_connected_kernel_bf_tiled.cpp
+++ b/src/plugins/intel_gpu/src/kernel_selector/kernels/fully_connected/fully_connected_kernel_bf_tiled.cpp
@@ -385,16 +385,10 @@ KernelsData FullyConnected_bf_tiled::GetTunedKernelsDataByIndex(const Params &pa
 
     tune_params tparams = GetAutoTuneParams(fc_params, autoTuneIndex);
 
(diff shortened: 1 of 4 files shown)


src/plugins/intel_gpu/src/kernel_selector/kernels/fully_connected/fully_connected_kernel_bf_tiled.cpp
```diff
@@ -385,16 +385,10 @@ KernelsData FullyConnected_bf_tiled::GetTunedKernelsDataByIndex(const Params &pa
...
-    if (tparams.tile_ofm * simd == 32)
-        weights_layout = WeightsLayout::os_iyx_osv32;
-    else if (tparams.tile_ofm * simd == 64)
-        weights_layout = WeightsLayout::os_iyx_osv64;
-
     return GetCommonKernelsData(params,
                                 options,
                                 fc_params.inputs[0].GetLayout(),
-                                weights_layout,
+                                WeightsLayout::oiyx,
```
sshlyapn, line 391: Did you have a chance to evaluate performance impact of this change on previous generations as well? Maybe it makes sense to keep weights_layout as a parametrized option under some conditions (e.g. type of GPU or something else)

PR #19930, opened 2023-09-18
[GPU] Fix gpu memory handling

yeonbok:
### Details:
 - Disabled enqueue zero size memcpy 
 - Allow host memory usage larger than physical memory for Windows 
 - Added more debug prints

### DIFF
--- a/src/plugins/intel_gpu/src/runtime/ocl/ocl_engine.cpp
+++ b/src/plugins/intel_gpu/src/runtime/ocl/ocl_engine.cpp
@@ -133,11 +133,48 @@ memory::ptr ocl_engine::allocate_memory(const layout& layout, allocation_type ty
                     "Requested ", layout.bytes_count(), " bytes "
                     "but max alloc size is ", get_device_info().max_alloc_mem_size, " bytes");
 
-    auto used_mem = get_used_device_memory(allocation_type::usm_device) + get_used_device_memory(allocation_type::usm_host);
-    OPENVINO_ASSERT(layout.bytes_count() + used_mem <= get_max_memory_size(),
+    auto dev_type = get_device_info().dev_type;
+    auto used_usm_device = get_used_device_memory(allocation_type::usm_device);
+    auto used_usm_host = get_used_device_memory(allocation_type::usm_host);
+    if (dev_type == device_type::integrated_gpu) {
+        auto total_used_mem = used_usm_device + used_usm_host;
+        if (layout.bytes_count() + total_used_mem > get_max_memory_size()) {
+#ifdef __unix__
+            OPENVINO_ASSERT(false,
                     "[GPU] Exceeded max size of memory allocation: ",
-                    "Required ", (layout.bytes_count() + used_mem), " bytes "
-                    "but memory size is ", get_max_memory_size(), " bytes");
+                    "Required ", (layout.bytes_count() + total_used_mem), " bytes "
+                    "but memory size of host memory is ", get_max_memory_size(), " bytes");
+#else
+            GPU_DEBUG_COUT << "[GPU][Warning] Exceeded max size of host memory: "
+                    << "Required " << (layout.bytes_count() + total_used_mem) << " bytes "
+                    << "but memory size of host memory is " << get_max_memory_size() << " bytes"
+                    << std::endl;
+            GPU_DEBUG_COUT << "=> Performance might be affected due to memory swap" << std::endl;
+#endif
+        }
+    } else { // discrete_gpu
+        if (type == allocation_type::usm_device) {
+            auto max_dev_mem = get_device_info().max_global_mem_size;
+            OPENVINO_ASSERT(layout.bytes_count() + used_usm_device <= max_dev_mem,
+                    "[GPU] Exceeded max size of memory object allocation for device mem: ",
+                    "Requested ", layout.bytes_count(), " bytes "
+                    "but max alloc size of device memory is ", max_dev_mem, " bytes");
+        } else {
+#ifdef __unix__
+        // Prevent from being killed by Ooo Killer of Linux
+            OPENVINO_ASSERT(layout.bytes_count() + used_usm_host <= get_host_memory_size(),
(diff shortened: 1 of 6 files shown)


src/plugins/intel_gpu/src/runtime/ocl/ocl_engine.cpp
```diff
@@ -133,11 +133,48 @@ memory::ptr ocl_engine::allocate_memory(const layout& layout, allocation_type ty
                     "Requested ", layout.bytes_count(), " bytes "
                     "but max alloc size is ", get_device_info().max_alloc_mem_size, " bytes");
 
-    auto used_mem = get_used_device_memory(allocation_type::usm_device) + get_used_device_memory(allocation_type::usm_host);
-    OPENVINO_ASSERT(layout.bytes_count() + used_mem <= get_max_memory_size(),
+    auto dev_type = get_device_info().dev_type;
+    auto used_usm_device = get_used_device_memory(allocation_type::usm_device);
+    auto used_usm_host = get_used_device_memory(allocation_type::usm_host);
```
vladimir-paramuzov, line 138: 1. Could you extract those checks to separate method (can_allocate() or somethine like that)
2. Please add a comment which describes OS/device specific limitations on allocation size
3. Can the code be simplified? The error message is duplicated many times with minor modifications, so I think it can be parameterized and then saved to stringstream for instance.

src/plugins/intel_gpu/src/runtime/ocl/ocl_memory.cpp
```diff
@@ -414,6 +414,10 @@ void gpu_usm::unlock(const stream& /* stream */) {
 }
 
 event::ptr gpu_usm::fill(stream& stream, unsigned char pattern) {
+    if (_bytes_count == 0) {
+        GPU_DEBUG_TRACE_DETAIL << "Skip gpu_usm::fill for 0 size tensor" << std::endl;
```
vladimir-paramuzov, line 418: Does that condition mean that we have allocations with 0 bytes? Or allocation size is always non-zero but memory can be reinterpreted to empty tensor? IIRC usm allocations with 0 bytes may crash on some platforms, so if we have such case somewhere in the pipeline, then we need to work-around that too.

src/plugins/intel_gpu/src/runtime/ocl/ocl_engine.cpp
yeonbok: Yes, I was going to reduce the condition because we need to allow swapping for device memory too.

src/plugins/intel_gpu/src/runtime/ocl/ocl_memory.cpp
yeonbok: Memory allocation size is non zero but shape is zero size.

src/plugins/intel_gpu/src/runtime/ocl/ocl_engine.cpp
yeonbok: Updated, could you review again?

vladimir-paramuzov (approved):

andrew-k-park (approved):

ahnyoung-paul (approved):

src/plugins/intel_gpu/src/runtime/ocl/ocl_memory.cpp
```diff
@@ -439,6 +443,10 @@ event::ptr gpu_usm::fill(stream& stream) {
 }
 
 event::ptr gpu_usm::copy_from(stream& stream, const memory& other, bool blocking) {
+    if (_bytes_count == 0) {
+        GPU_DEBUG_TRACE_DETAIL << "Skip EnqueueMemcpy for 0 size tensor" << std::endl;
+        return nullptr;
```
sshlyapn, line 448: I think it's safer to return user event instead of nullptr, at least while we have `event->wait()` calls in plugin (w/o any pointer validity checks)

src/plugins/intel_gpu/src/runtime/ocl/ocl_memory.cpp
```diff
@@ -462,6 +470,10 @@ event::ptr gpu_usm::copy_from(stream& stream, const memory& other, bool blocking
 }
 
 event::ptr gpu_usm::copy_from(stream& stream, const void* host_ptr, bool blocking) {
+    if (_bytes_count == 0) {
```
sshlyapn, line 473: Just out of curiosity: shouldn't we implement the same condition for OpenCL buffers as well?

src/plugins/intel_gpu/src/runtime/ocl/ocl_memory.cpp
yeonbok: I wanted to reduce the even creation overhead too, but I agree safeness is important too. I will update.

src/plugins/intel_gpu/src/runtime/ocl/ocl_memory.cpp
yeonbok: Hmm my basic intention of removing this "0 byte memcpy" is for performance matter rather than the functionality, because AFAIK the driver actually should support 0 byte memcopy. So I thought it is just a "nice to have" feature, but for the consistency, I will apply it for that case too :)

src/plugins/intel_gpu/src/runtime/ocl/ocl_memory.cpp
sshlyapn: Yeah, agree, not sure how much it affects performance, but if the impact is indeed noticeable, we may consider to implement all wait calls with stream::wait_for_events() and check pointers there

src/plugins/intel_gpu/src/runtime/ocl/ocl_memory.cpp
sshlyapn: Oh, got it, I thought we were supposed to address this issue on our side, thanks for clarification

yeonbok:
@sshlyapn Applied all the comments, please review again

Issue #19833, opened 2023-09-14
[Feature Request]: Improve the ov_core_read_model_from_memory API Definition in OpenVINO Library


sdcb:
### Request Description

The current OpenVINO library API ov_core_read_model_from_memory seems to have a design issue. Specifically, the model_str parameter, which appears to be a string, actually points to the model file in memory. Internally, it is then converted into a string for processing. This leads to a problem where if the model file includes a null character ('\0'), the read-in model size will be smaller than the actual size. Conversely, if the model file does not contain a null character, it will inevitably cause a memory error.

```c
/**
 * @brief Reads models from IR / ONNX / PDPD / TF / TFLite formats.
 * @ingroup ov_core_c_api
 * @param core A pointer to the ie_core_t instance.
 * @param model_str String with a model in IR / ONNX / PDPD / TF / TFLite format.
 * @param weights Shared pointer to a constant tensor with weights.
 * @param model A pointer to the newly created model.
 * Reading ONNX / PDPD / TF / TFLite models does not support loading weights from the @p weights tensors.
 * @note Created model object shares the weights with the @p weights object.
 * Thus, do not create @p weights on temporary data that can be freed later, since the model
 * constant data will point to an invalid memory.
 * @return Status code of the operation: OK(0) for success.
 */
OPENVINO_C_API(ov_status_e)
ov_core_read_model_from_memory(const ov_core_t* core,
                               const char* model_str,
                               const ov_tensor_t* weights,
                               ov_model_t** model);
```

I propose to modify the API by renaming the model_str parameter to model_data, and adding a new integer parameter model_data_size after it, to accurately represent the memory size of model_data.

```c
OPENVINO_C_API(ov_status_e)
ov_core_read_model_from_memory(const ov_core_t* core,
                               const char* model_data,
                               const int model_data_size,
                               const ov_tensor_t* weights,
                               ov_model_t** model);
```

### Feature Use Case

This proposed change would prevent potential memory errors and inaccurate model size readings that can occur if model files contain or lack null characters. It would increase the robustness of the API, making it more reliable and easier to use for developers. This improvement could also potentially increase the efficiency of model loading operations, as it provides clearer information about the memory size required for the operation.

### Issue submission checklist

- [X] The feature request or improvement must be related to OpenVINO

riverlijunjie:
Char string contains '\0' at the end to mean string is ended. So for char string, we don't need data size.
And in C application, we suggest adding a '\0' at the end of string.

[...]

PR #19931, opened 2023-09-19
[C API] support string size for char pointer

riverlijunjie:
### Details:
 - Add string size for C API reading model from memory

### Tickets:
 - Closes https://github.com/openvinotoolkit/openvino/issues/19833


src/bindings/c/include/openvino/c/ov_core.h
```diff
@@ -189,6 +189,28 @@ ov_core_read_model_from_memory(const ov_core_t* core,
...
+ * @param weights Shared pointer to a constant tensor with weights.
+ * @param model A pointer to the newly created model.
+ * Reading ONNX / PDPD / TF / TFLite models does not support loading weights from the @p weights tensors.
+ * @note Created model object shares the weights with the @p weights object.
+ * Thus, do not create @p weights on temporary data that can be freed later, since the model
+ * constant data will point to an invalid memory.
+ * @return Status code of the operation: OK(0) for success.
+ */
+OPENVINO_C_API(ov_status_e)
+ov_core_read_model_from_memory_with_size(const ov_core_t* core,
```
ilyachur, line 208: Maybe we can come up the better name... For example: `ov_core_read_model_from_memory_buffer`

ilyachur: Also we can deprecate old method, because these methods provide the same functionality

riverlijunjie: But if deprecate old method maybe lead to some compatibility problem if this API has been used by user application.

ilyachur: Yes in this case user will get a deprecation warning, but it is a normal process and without it we cannot change or remove legacy API. I don't talk about deletion of legacy method immediately, only about the deprecation and removing later.
From my point of view two methods will confuse users...

ilyachur (approved):

ilya-lavrenov (approved):

PR #19932, opened 2023-09-19
Fix warning in template plugin

riverlijunjie:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


ilya-lavrenov (approved):
why does not CI reproduce these warnings?

riverlijunjie:
> why does not CI reproduce these warnings?
In local build, I didn't use Ninja for CMake with CMAKE_COMPILE_WARNING_AS_ERROR=ON, it will reports above error.
But if use Ninja, it will report the similar warning but not treat it as error and building going on.

PR #19934, opened 2023-09-19
[core] Api 2.0/migrate swish op to new API

praasz:
### Details:
 - Migrate swish operator to new API. Remove legacy headers, namespace, and function. Migrate evaluate and reduce binary size.
 - Remove `visit_attributes` override as it has got same implementation as base class
 - Refactor reference implementation to reduce binary size

### Tickets:
 - [CVS-118678](https://jira.devtools.intel.com/browse/CVS-118678)


t-jankowski (approved):

src/core/src/op/swish.cpp
```diff
@@ -50,81 +59,52 @@ void op::v4::Swish::validate_and_infer_types() {
...
+                                                      shape_size(inputs[0].get_shape()));
 }
 
-bool op::v4::Swish::has_evaluate() const {
+bool Swish::has_evaluate() const {
     OV_OP_SCOPE(v4_Swish_has_evaluate);
     switch (get_input_element_type(0)) {
-    case ngraph::element::f16:
-    case ngraph::element::f32:
+    case element::f16:
```
t-jankowski, line 101: Should it be dropped in context of PR #19830?

src/core/src/op/swish.cpp
```diff
@@ -50,81 +59,52 @@ void op::v4::Swish::validate_and_infer_types() {
...
-    OPENVINO_SUPPRESS_DEPRECATED_END
-    return swish::evaluate_swish(inputs, outputs[0]);
+    OPENVINO_ASSERT(outputs.size() == 1);
+    OPENVINO_ASSERT(swish::has_1_or_2_inputs(inputs.size()));
+
+    outputs[0].set_shape(inputs[0].get_shape());
+    const auto& arg1 = inputs.size() == 2 ? inputs[1] : Tensor();
+
+    using namespace ov::element;
+    return IfTypeOf<f16, f32>::apply<swish::Evaluate>(inputs[0].get_element_type(),
```
t-jankowski, line 91: Should it be only `fp32` in context of PR #19830?

src/core/src/op/swish.cpp
praasz: I've tried to drop it but then there are functional test failures in template and CPU plugin. And not sure if is enough to remove this precision from these tests or ir is required. This is reason why this drop is not introduced here and I think test issues will be resolved in #19830.

src/core/src/op/swish.cpp
praasz: Same as above.

vurusovs (approved):

PR #19951, opened 2023-09-19
[Snippets] Added SubgraphPass and SubgraphManager support

a-sidorova:
### Details:
 - *Added support of `SubgraphPass` and `SubgraphManager` to allow developers to create passes for `Subgraph` op. Since it's needed only for `CommonOptimizations` and should be run only in this pass, these classes are declared in `CommonOptimizations` classes*
 - *Needed for updating of `SpitDimensionM` optimization: make development easier*

### Tickets:
 - *N/A*


a-sidorova:
Close since this feature is part of https://github.com/openvinotoolkit/openvino/pull/20013 now

PR #19957, opened 2023-09-19
[Snippets] Changed BrgemmCopyB shape inference

v-Golubev:
### Details:
 - *BrgemmCopyB shape inference changed to PassThroughShapeInfer*
 - *Allocation shape assignment for Buffers after BrgemmCopyB moved to a separate transformation*


src/plugins/intel_cpu/src/transformations/snippets/x64/op/brgemm_copy_b.hpp
```diff
@@ -42,8 +42,11 @@ class BrgemmCopyB : public snippets::op::MemoryAccess {
     size_t get_n_block_size() const { return m_N_blk; }
     void set_k_block_size(size_t block_size) { m_K_blk = block_size; }
     void set_n_block_size(size_t block_size) { m_N_blk = block_size; }
+    size_t get_k_rounded() const { return m_K_rounded; }
+    size_t get_n_rounded() const { return m_N_rounded; }
```
a-sidorova, line 46: Do we really these setters? To be honest for me `_rounded` sounds strange 😄 
I'd suggest to create two direct methods `get_data_repacking_shape` and `get_compensation_shape` at least.
Could you please verify that it's more comfortable than using these getters?
Moreover, as I understand these class fields are depended on shapes. It's not very cool to store them in class so when we are working on dynamism support.

v-Golubev: Agree with you. The mentioned methods are created and reused in `SetBrgemmCopyBBuffersShape` transformation

a-sidorova (approved):

src/plugins/intel_cpu/src/transformations/snippets/x64/op/brgemm_copy_b.cpp
```diff
@@ -57,39 +57,31 @@ void BrgemmCopyB::custom_constructor_validate_and_infer_types(std::vector<size_t
     // During ctor call, BrgemmCopyB doesn't know his port descriptors.
     // So we use port descs from source inputs
     const auto element_type = get_input_element_type(0);
-    const auto pshape = snippets::utils::get_planar_pshape(get_input_partial_shape(0), layout_input);
-    validate(pshape, element_type);
+    const auto& pshape = get_input_partial_shape(0);
+    const auto planar_pshape = snippets::utils::get_planar_pshape(pshape, layout_input);
+    set_output_type(0, element_type, planar_pshape);
+    if (is_with_compensations()) {
```
IvanNovoselov, line 63: Minor: this is not a straightforward behavior, so let's add a comment here on why it's done this way

src/plugins/intel_cpu/src/transformations/snippets/x64/op/brgemm_cpu.cpp
```diff
@@ -181,13 +169,7 @@ size_t BrgemmCPU::get_offset_scratch() const {
...
 }
 
-BrgemmCPU::ShapeInfer::ShapeInfer(const std::shared_ptr<ov::Node>& n) : Brgemm::ShapeInfer(n) {
-    const auto& brg = ov::as_type_ptr<BrgemmCPU>(n);
-    OPENVINO_ASSERT(brg, "Got invalid node in BrgemmCPU::ShapeInfer");
-    const auto brgemm_copy = brg->is_with_data_repacking() ? brg->get_brgemm_copy() : nullptr;
-    if (brgemm_copy)
-        m_io_layouts[1] = snippets::lowered::PortDescriptorUtils::get_port_descriptor_ptr(brgemm_copy->input(0))->get_layout();
-}
+BrgemmCPU::ShapeInfer::ShapeInfer(const std::shared_ptr<ov::Node>& n) : Brgemm::ShapeInfer(n) {}
```
IvanNovoselov, line 172: If shapeInfer is declared inside a node, it means that it is unique for this particular node. If the same shapeInfer is shared between several nodes, then it must be implemented separately in shape_infer_instances, and used in the factory's mapping for both of the nodes. This would help to keep shapeInfer pipeline concise and transparent, so let's try to keep it this way.
Note that we can split shapeInfer implementations into several files if necessary.

IvanNovoselov (approved):
Approve, but need to move Brgemm (and BrgemmCPU) shapeInfer to shape_infer_instances

src/plugins/intel_cpu/src/transformations/snippets/x64/op/brgemm_cpu.cpp
v-Golubev: Done, thanks

src/plugins/intel_cpu/src/transformations/snippets/x64/op/brgemm_copy_b.cpp
v-Golubev: Done, thanks

v-Golubev:
@dmitry-gorokhov could you please take a look?

src/plugins/intel_cpu/src/transformations/snippets/x64/pass/lowered/set_brgemm_copy_b_buffers_shape.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include "snippets/lowered/pass/pass.hpp"
+
+namespace ov {
+namespace intel_cpu {
+namespace pass {
+
+/**
+ * @interface SetBrgemmCopyBBuffersShape
+ * @brief Sets the allocation shape for the Buffers after BrgemmCopyB node using BrgemmCopyB parameters
```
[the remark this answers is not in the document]
v-Golubev: @dmitry-gorokhov you are right. But if you don't mind I will apply this comment in the next PR: #19335. This one could be merged since CI is green.

PR #19962, opened 2023-09-20
[CPU]Correct flags for clang compiler

zhangYiIntel:
### Details:
 - *Enable Warning as Error in Android Compilation*
 - *Change Unsupported clang flags from `-Wno-delete-non-abstract-non-virtual-dtor` to `-Wno-delete-non-virtual-dtor`*

### Tickets:
 - *CVS-120605*

### DIFF
--- a/cmake/developer_package/features.cmake
+++ b/cmake/developer_package/features.cmake
@@ -18,7 +18,7 @@ else()
     ie_option(USE_BUILD_TYPE_SUBFOLDER "Create dedicated sub-folder per build type for output binaries" ON)
 endif()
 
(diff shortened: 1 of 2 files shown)


zhangYiIntel:
@ilya-lavrenov  For `Android_x86_64` build, if enabling `CMAKE_COMPILE_WARNING_AS_ERROR`, the GPU part also fails. The problem exists not only in CPU Plugins. I suggest to invite colleagues from GPU plugin to take a look too.

zhangYiIntel:
The warning is not caused by MLAS

PR #19977, opened 2023-09-20
[apiConformance] Change test for ov::hint::inference_precision RW property

sbalandi:
### Details:
- *move the ov::hint::inference_precision check to separate test. Now we check if property supported in general and if at least one precision is implemented. But if some precision is not maintained, it will not be a reason for the test to fail*

### Tickets:
 - *CVS-103948*

### DIFF
--- a/src/tests/functional/plugin/shared/src/behavior/ov_plugin/properties_tests.cpp
+++ b/src/tests/functional/plugin/shared/src/behavior/ov_plugin/properties_tests.cpp
@@ -240,18 +240,6 @@ std::vector<ov::AnyMap> OVPropertiesTestsWithCompileModelProps::configurePropert
 std::vector<ov::AnyMap> OVPropertiesTestsWithCompileModelProps::getRWMandatoryPropertiesValues(std::vector<std::string> props) {
     std::vector<ov::AnyMap> res;
 
-    if (props.empty() || std::find(props.begin(), props.end(), ov::hint::inference_precision.name()) != props.end()) {
-        const std::vector<ov::element::Type> ovElemTypes = {
-            ov::element::f64, ov::element::f32, ov::element::f16, ov::element::bf16,
-            ov::element::i64, ov::element::i32, ov::element::i16, ov::element::i8, ov::element::i4,
-            ov::element::u64, ov::element::u32, ov::element::u16, ov::element::u8, ov::element::u4,  ov::element::u1,
-            ov::element::boolean, ov::element::undefined, ov::element::dynamic
-        };
-        for (auto &precision : ovElemTypes) {
-            res.push_back({{ov::hint::inference_precision(precision)}});
-        }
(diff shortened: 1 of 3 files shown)


src/tests/functional/plugin/shared/src/behavior/ov_plugin/properties_tests.cpp
```diff
@@ -401,6 +389,45 @@ TEST_P(OVCheckChangePropComplieModleGetPropTests_DEVICE_ID, ChangeCorrectDeviceP
...
+    std::vector<ov::PropertyName> supported_properties;
+    OV_ASSERT_NO_THROW(supported_properties = core->get_property(target_device, ov::supported_properties));
+    auto supported = util::contains(supported_properties, ov::hint::inference_precision);
+    ASSERT_TRUE(supported) << "property is not supported: " << ov::hint::inference_precision;
+
+    ov::Any default_property;
+    OV_ASSERT_NO_THROW(default_property = core->get_property(target_device, ov::hint::inference_precision));
+    ASSERT_FALSE(default_property.empty());
+
+    const std::vector<ov::element::Type> ovElemTypes = {
```
iefode, line 402: Do we have any OV interface to get all types?

src/tests/functional/plugin/shared/src/behavior/ov_plugin/properties_tests.cpp
```diff
@@ -401,6 +389,45 @@ TEST_P(OVCheckChangePropComplieModleGetPropTests_DEVICE_ID, ChangeCorrectDeviceP
...
+        ov::element::boolean, ov::element::undefined, ov::element::dynamic
+    };
+
+    bool any_supported = false;
+    for (ov::element::Type type : ovElemTypes) {
+        try {
+            core->set_property(target_device, ov::hint::inference_precision(type));
+            core->compile_model(model, target_device, compileModelProperties);
+        } catch (const Exception& ex) {
+            continue;
```
iefode, line 415: ?

iefode (approved):

src/tests/functional/plugin/shared/src/behavior/ov_plugin/properties_tests.cpp
sbalandi: I have not found (

src/tests/functional/plugin/shared/src/behavior/ov_plugin/properties_tests.cpp
sbalandi: my logic is: If something went wrong with some precision it is okay, we want to make sure that it is possible to set any of them(at least one precision)
I thought that we could check err message, if it have close meaning to "unsupported precision". But the err messages are different in different plugins.

iefode (approved):

PR #19981, opened 2023-09-21
[PT FE]: extend logical operations support

eaidova:
### Details:
 - *added support aten::logical_and, aten::logical_or, aten::logical_not, aten::logical_xor*

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/frontends/pytorch/src/op/logical.cpp
+++ b/src/frontends/pytorch/src/op/logical.cpp
@@ -4,7 +4,9 @@
 
 #include "openvino/frontend/pytorch/node_context.hpp"
 #include "openvino/op/logical_and.hpp"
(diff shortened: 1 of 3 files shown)


mvafin (approved):
Does `aten::logical_*` support non-boolean inputs? Do they work as bitwise ops in that case or boolean ops?

eaidova:
> Does `aten::logical_*` support non-boolean inputs? Do they work as bitwise ops in that case or boolean ops?



> Does `aten::logical_*` support non-boolean inputs? Do they work as bitwise ops in that case or boolean ops?

bitwise accepts only bool and int8 and for bool works as boolean and for int as bitwise
logical accept different types and convert them to bool

PR #19986, opened 2023-09-21
Fix `StressMemLeaksTests` with several models

vurusovs:
### Details:
 - Because `ie_wrapper` was captured by reference, tests actually execute only last model instead of several models

### DIFF
--- a/tests/stress_tests/common/ie_pipelines/pipelines.cpp
+++ b/tests/stress_tests/common/ie_pipelines/pipelines.cpp
@@ -55,11 +55,9 @@ create_compiled_model(const std::string &model, const std::string &target_device
     };
 }
 
-std::function<void()> recreate_compiled_model(std::shared_ptr<InferApiBase> &ie_wrapper, const std::string &model,
+std::function<void()> recreate_compiled_model(std::shared_ptr<InferApiBase> &ie_wrapper,
                                               const std::string &target_device, const int &api_version) {
-    return [&] {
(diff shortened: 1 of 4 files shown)


ilya-lavrenov (approved):

tests/stress_tests/common/ie_pipelines/pipelines.cpp
```diff
@@ -97,14 +95,14 @@ infer_request_inference(const std::string &model, const std::string &target_devi
 
 
 std::function<void()> reinfer_request_inference(std::shared_ptr<InferApiBase> &ie_wrapper) {
-    return [&] {
+    return [=] {
         ie_wrapper->infer();
```
tiger100256-hu, line 99: I am not sure if it's right here, from my understand of the test code, maybe I am wrong, even if use [=] here, when test_params.numthreads > 1, there will be multi threads use same ie_wrapper to call infer(); one of them will do infer and other threads will throw request busy. so please help to set test_params.numthreads=10 and check if the test case is ok, thanks.

vurusovs: Thank you for detailed review! The original issue was in consecutive inference, each inference request in separate thread (code snippet may be found in CVS-111490). So it shouldn't cause issues with `ie_wrapper`

`test_params.numthreads` is a legacy from `StressUnitTests`, `StressMemLeakTests` weren't designed to test parallel execution because it is non-trivial

PR #19988, opened 2023-09-21
[core] Api 2.0/migrate operators without evaluate

praasz:
### Details:
 - Migrate operators without evaluate to new API. remove legacy includes, namespaces, remove using std namespace in ops.
 - Use OPENVINO_ASSERT instead NGRAPH_CHECK in ops
 - Remove `visit_attributes` in modified ops where this function has got same implementation as base class.
 - Move `shape_utils.hpp` from `dev_api` to `dev_api/openvino/core/`

### Tickets:
 - [CVS-120795](https://jira.devtools.intel.com/browse/CVS-120795)


src/core/src/op/lrn.cpp
```diff
@@ -2,22 +2,19 @@
...
-using namespace std;
-using namespace ngraph;
+#include "openvino/core/attribute_visitor.hpp"
+#include "openvino/core/validation_util.hpp"
+#include "openvino/op/constant.hpp"
+#include "openvino/op/multiply.hpp"
 
-op::LRN::LRN(const Output<Node>& arg, double alpha, double beta, double bias, size_t size)
+namespace ov {
+op::v0::LRN::LRN(const Output<Node>& arg, double alpha, double beta, double bias, size_t size)
```
t-jankowski, line 14: What about nesting in such way?
```suggestion
namespace ov {
namespace op {
namespace v0 {
LRN::LRN(const Output<Node>& arg, double alpha, double beta, double bias, size_t size)
```

t-jankowski (approved):

src/core/src/op/lrn.cpp
t-jankowski: However I think it'd be better to not expand this PR ;)

vurusovs (approved):

PR #19991, opened 2023-09-21
[PT FE] Fix aten::repeat regression

mvafin:
### Details:
 - *Fix `aten::repeat` regression introduced by #19926*

### Tickets:
 - *CVS-121034, CVS-121082, CVS-121152*


src/frontends/pytorch/src/op_table.cpp
```diff
@@ -134,6 +134,7 @@ OP_CONVERTER(translate_randn_like);
 OP_CONVERTER(translate_reciprocal);
 OP_CONVERTER(translate_relu6);
 OP_CONVERTER(translate_remainder);
+OP_CONVERTER(translate_repeat);
```
mvafin, line 137:
```suggestion
```

eaidova (approved):

tests/layer_tests/requirements.txt
```diff
@@ -4,6 +4,7 @@ numpy
 requests
 torch
 torchvision
+transformers
```
eaidova, line 7: do you really need install transformers in this PR?
I believe it is related to another PR

mvafin: I do, I added test for primitive from transformers

PR #19992, opened 2023-09-21
[CC] fix issue of creating pritive desc failure

riverlijunjie:
### Details:
 - oneDNN https://github.com/openvinotoolkit/oneDNN/pull/214
 - Fix issue of failure during creating primitive desc in case of conditional compilation:
    oneDNN requires forward pd to exists until primitive is created, so such primitives need to be kept to create its primitive_desc.

### Tickets:
 - 121046


riverlijunjie:
Replaced by https://github.com/openvinotoolkit/openvino/pull/20571

PR #19994, opened 2023-09-21
[PT FE]: support aten::broadcast_tensors

eaidova:
### Details:
 - *support aten::broadcast_tensors operation via transformation with prim::ListUnpack*
 
### Tickets:
 - *ticket-id*


src/frontends/pytorch/src/op/broadcast_tensors.cpp
```diff
@@ -0,0 +1,49 @@
...
+namespace frontend {
+namespace pytorch {
+namespace op {
+
+using namespace ov::op;
+
+OutputVector translate_broadcast_tensors(const NodeContext& context) {
+    num_inputs_check(context, 1, 1);
+    auto list_tensors = context.get_input(0).get_node_shared_ptr();
+    if (cast_fw_node(list_tensors, "prim::ListConstruct")) {
```
mvafin, line 25: if input is not ListConstruct you will delete op from graph, is it expected?

src/frontends/pytorch/src/transforms/prim_list_unpack_replacer.cpp
```diff
@@ -173,6 +173,32 @@ PrimListUnpackReplacer::PrimListUnpackReplacer() {
...
+            auto tensors = cast_fw_node(broadcast_tensors->input_value(0).get_node_shared_ptr(), "prim::ListConstruct");
+            if (!tensors) {
+                add_exception_to_fw_node(input_node,
+                                         "aten::broadcast_tensors: only prim::ListConstruct supported as input.");
+                return false;
+            }
+            auto zero = opset10::Constant::create(element::i32, Shape{}, {0});
+            Output<Node> final_shape_t = zero;
+            for (auto input : tensors->inputs()) {
+                auto tensor_shape = std::make_shared<opset10::ShapeOf>(input.get_source_output());
```
mvafin, line 186: Please use rg.make instead of make_shared and call copy_runtime_info_and_name afterwards

src/frontends/pytorch/src/transforms/prim_list_unpack_replacer.cpp
```diff
@@ -173,6 +173,32 @@ PrimListUnpackReplacer::PrimListUnpackReplacer() {
...
+                auto zero_broadcasted =
+                    std::make_shared<opset10::Broadcast>(zero, tensor_shape, ov::op::BroadcastType::BIDIRECTIONAL);
+                final_shape_t = std::make_shared<opset10::Add>(final_shape_t, zero_broadcasted);
```
mvafin, line 189: As far as I see you use `Add` here with zero to broadcast many shapes to broadcasted zero, meaning you use 2 broadcast ops, can it be 1 broadcast op?

src/frontends/pytorch/src/transforms/prim_list_unpack_replacer.cpp
```diff
@@ -173,6 +173,32 @@ PrimListUnpackReplacer::PrimListUnpackReplacer() {
...
+            }
+            auto final_shape = std::make_shared<opset10::ShapeOf>(final_shape_t, element::i32);
+            OutputVector outputs;
+            for (auto input : tensors->inputs()) {
+                outputs.push_back(std::make_shared<opset10::Broadcast>(input.get_source_output(),
+                                                                       final_shape,
+                                                                       ov::op::BroadcastType::BIDIRECTIONAL));
```
mvafin, line 196: Is it really bidirectional?

src/frontends/pytorch/src/transforms/prim_list_unpack_replacer.cpp
eaidova: done

src/frontends/pytorch/src/transforms/prim_list_unpack_replacer.cpp
eaidova: I think it is not possible as you do not know maximal shape before starting broadcasting, but I replaced first broadcast with just add

src/frontends/pytorch/src/transforms/prim_list_unpack_replacer.cpp
eaidova: replaced on numpy

mvafin (approved):

PR #19999, opened 2023-09-22
[cpu] fix memory leak in binconv

tiger100256-hu:
### Details:
 - *use std::shared_ptr instead of pointer*

### Tickets:
 - *CVS-120867*

### DIFF
--- a/src/plugins/intel_cpu/src/nodes/bin_conv.cpp
+++ b/src/plugins/intel_cpu/src/nodes/bin_conv.cpp
@@ -8,6 +8,7 @@
 #include "eltwise.h"
 #include "fake_quantize.h"
 #include "conv.h"
+#include <memory>
 #include <string>
 #include <vector>
 #include <dnnl_types.h>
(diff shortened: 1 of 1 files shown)


src/plugins/intel_cpu/src/nodes/bin_conv.cpp
```diff
@@ -63,10 +64,10 @@ struct jit_uni_bin_conv_kernel_f32 : public jit_uni_bin_conv_kernel, public jit_
         for (int i = 0; i < end_idx; i++) {
             auto &post_op = p.entry_[i];
             if (post_op.is_eltwise()) {
-                eltwise_injectors.push_back(new jit_uni_eltwise_injector_f32<isa>(
+                eltwise_injectors.push_back(std::make_shared<jit_uni_eltwise_injector_f32<isa>>(
```
zhangYiIntel, line 67: `eltwise_injectors` When will this vector be destroyed ? It seems that this vector could only be inserted, in such case, the data inside stays until process exits.

tiger100256-hu: the vector is a member of jit_uni_bin_conv_kernel_f32, so it will destroy when jit_uni_bin_conv_kernel_f32 destroy.

zhangYiIntel: Ok, then it's really a memory leak

zhangYiIntel (approved):
LGTM

ilya-lavrenov (approved):

vurusovs (approved):

PR #20005, opened 2023-09-22
[Opset13][CPP] Initialization of Opset13 

mitruska:
### Details:
 - Initialization of Opset13 in C++ API
 - To reduce number of repeated changes introducing Opset13 in each PR related to new operator
 
### Tickets:
 - N/A


src/core/tests/opset.cpp
```diff
@@ -69,7 +70,8 @@ INSTANTIATE_TEST_SUITE_P(opset,
                                          OpsetTestParams{ov::get_opset9, 173},
                                          OpsetTestParams{ov::get_opset10, 177},
                                          OpsetTestParams{ov::get_opset11, 177},
-                                         OpsetTestParams{ov::get_opset12, 178}),
+                                         OpsetTestParams{ov::get_opset12, 178},
+                                         OpsetTestParams{ov::get_opset13, 178}),
```
praasz, line 74:
```suggestion
                                         OpsetTestParams{ov::get_opset13, 179}),
```
?

mitruska: This is a number of the operations in the opset, as the initial point it's the same value as the Opset12.
But yes it will be updated +1 with introduction of each new operator.

mmikolajcz (commented):
I think that there is couple more places that need update to opset13, like:

- `src/tests/functional/plugin/conformance/test_runner/op_conformance_runner/src/op_impl_check/single_op_graph.cpp`,
- `src/frontends/common/include/openvino/frontend/extension/op.hpp`
- `src/core/tests/op.cpp`

mmikolajcz (approved):

mitruska:
> I think that there is couple more places that need update to opset13, like:
> 
> * `src/tests/functional/plugin/conformance/test_runner/op_conformance_runner/src/op_impl_check/single_op_graph.cpp`,
> * `src/frontends/common/include/openvino/frontend/extension/op.hpp`
> * `src/core/tests/op.cpp`

@mmikolajcz  Thanks! Updated 🚀

praasz (approved):

PR #20007, opened 2023-09-22
[PyOV] Make openvino properties behave as python property object

akuporos:
### Details:
 - dynamically import `openvino.properties` and its submodules
 - dynamically wrap `openvino.properties` so they behave like python properties

### Tickets:
 - 119377


src/bindings/python/src/openvino/properties/intel_gpu/hint/__init__.py
```diff
@@ -3,10 +3,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Properties
-from openvino._pyopenvino.properties.intel_gpu.hint import queue_throttle
-from openvino._pyopenvino.properties.intel_gpu.hint import queue_priority
-from openvino._pyopenvino.properties.intel_gpu.hint import host_task_priority
-from openvino._pyopenvino.properties.intel_gpu.hint import available_device_mem
+import openvino._pyopenvino.properties.intel_gpu.hint as __hint
+from openvino.properties.properties import __make_properties
```
jiwaszki, line 7:
```suggestion
from openvino.properties._properties import __make_properties
```
Would it be possible to make it more "private"? I know most of the functions inside are done this way but either changing it to `_properties` or other name (i.e. `impl`/`details`) would help to distinguish module and implementation from each other.

src/bindings/python/tests/test_runtime/test_properties.py
```diff
@@ -28,7 +28,7 @@ def test_properties_ro_base():
 
 
 def test_properties_rw_base():
-    assert props.cache_dir() == "CACHE_DIR"
+    assert props.cache_dir == "CACHE_DIR"
```
jiwaszki, line 31:
```suggestion
    assert ov.properties.cache_dir == "CACHE_DIR"
```
Can one or two "longer" imports be checked?

src/bindings/python/src/openvino/properties/properties.py
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2018-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import sys
+from types import BuiltinFunctionType, ModuleType
+from typing import Callable, Any, Union
+
+
+class Property(str):
```
jiwaszki, line 10: Please add some docs here, as it is derivative of a standard string - it might be confusing why it can be called.

jiwaszki (commented):
Overall LGTM, just some minor things to discuss.

Docs updates will appear in the next PR right?
Plus, please consider extending developer docs with "how-to" add new properties module and how it corresponds with pybind internals. It might be worth it to take notes on that.

src/bindings/python/src/openvino/properties/properties.py
akuporos: added

src/bindings/python/tests/test_runtime/test_properties.py
akuporos: added

src/bindings/python/src/openvino/properties/intel_gpu/hint/__init__.py
akuporos: marked it as `_properties`.

akuporos:
@jiwaszki  all docs/snippets update I'll make in the next PR

jiwaszki (approved):

PR #20009, opened 2023-09-22
Added support of shapes and types from original FW in ov.convert_model()

popovaan:
Description: 
Added support of shapes and types from original PyTorch, TF and PaddlePaddle in ov.convert_model()
Ticket: 117391

Code:
* [x]  Comments
* [x]  Code style (PEP8)

Validation:
* [x]  Unit tests
* [x]  Transformation tests

Documentation:
* [x]  User guide update


eaidova (approved):

tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -25,9 +26,18 @@
 # Helper class for storing input cut information
 _InputCutInfo = namedtuple("InputCutInfo", ["name", "shape", "type", "value"], defaults=[None, None, None, None])
 
+
 def is_shape_type(value):
     if isinstance(value, PartialShape):
         return True
+    if 'tensorflow' in sys.modules:
```
rkazants, line 33: Do we need to add for PyTroch? It seems to be `torch.Tensor.Size`?

tools/ovc/openvino/tools/ovc/cli_parser.py
```diff
@@ -25,9 +26,18 @@
 # Helper class for storing input cut information
 _InputCutInfo = namedtuple("InputCutInfo", ["name", "shape", "type", "value"], defaults=[None, None, None, None])
 
+
 def is_shape_type(value):
```
rkazants, line 30: let us have these utils functions in `utils` file because `cli_parser` is quite overloaded. Please check if we have already such routine because we have already implemented quite similar logic that relates also to types.

tests/layer_tests/ovc_python_api_tests/test_complex_params.py
```diff
@@ -70,13 +79,27 @@ def create_tf_model_single_input_output(tmp_dir):
         # save model to .pb and return path to the model
         return save_to_pb(tf_net, tmp_dir)
 
-    test_data = [
+    paddle_test_data_set_1 = [
```
rkazants, line 82: I would suggest to split these tests into separate files by FWs

tools/ovc/openvino/tools/ovc/cli_parser.py
popovaan: `torch.Tensor.Size` represents Python `list`. So it works on master, I added test for this type in [test_complex_params.py](https://github.com/openvinotoolkit/openvino/pull/20009/files#diff-e3878abd42ca567a18e9b90e4bd37a1394a1db1106502b0ee15b4985fc2fb856)

rkazants (approved):

tests/layer_tests/ovc_python_api_tests/test_complex_params.py
popovaan: Done.

PR #20014, opened 2023-09-22
[PyOV] fix if_op return types

akuporos:
### Details:
 - wrap `get_function` methods of `if-op`, `loop`, `tensor_iterator` classes to return `openvino.runtime.Model` object instead of `openvino._pyopenvino.Model`.

### Tickets:
 - CVS-121115


p-wysocki (commented):
I guess `isinstance` asserts in tests could be added in order to avoid this in the future.

There are some wild errors in CI - maybe adding casting to pure Python types would be easier to implement error-free in pure Python wrappers? Just an idea, I'm not certain if good.

src/bindings/python/src/pyopenvino/graph/model.cpp
```diff
@@ -998,6 +998,18 @@ void regclass_graph_Model(py::module m) {
...
+    model.def(
+        "_get_raw_address",
+        [](ov::Model& self) {
+            return reinterpret_cast<uint64_t>(&self);
+        },
+        R"(
+        Returns raw address of the Model object.
+
+        :return: a raw address of the Model object.
+        :rtype: int
```
jiwaszki, line 1010:
```suggestion
        Returns a raw address of the Model object from C++.
        
        Use this function to compare underlying C++ addresses instead of using `__eq__` in Python.

        :return: a raw address of the Model object.
        :rtype: int
```

jiwaszki (approved):
LGTM, just need some alignment and couple of tests outside the scope of ops (within just simple Model API tests). It can be done in the separate PR.

jiwaszki:
> There are some wild errors in CI - maybe adding casting to pure Python types would be easier to implement error-free in pure Python wrappers? Just an idea, I'm not certain if good.

@p-wysocki The idea here is to import `openvino.Model` directly and apply it on the pybind's level -- so the types are matching. The problem is that some wrappers are existing on Python level already, so either everything is moved to pybind or find some solution to WA around it.

Pure Python wrappers are not ideally either. There is no currently known way (in this scenario of Py-Cpp wrappers mix) to return same object 2+ times without comparing underlying pointers and/or other rich comparison methods.

Having this in mind, this solution is the best choice currently. One topic that bugs me are parallel executions of such code. There is implicit GIL capturing to allow Python's types handling. I wasn't able to break it myself but I see that it can be a little tricky to navigate in future. (*)

@akuporos (*) -- maybe let's put some future feature on our list to create "templated"/parametrized Python interpreter calls with conversions like this __init__ trick? So the solution will be centralized a little bit.

PR #20015, opened 2023-09-22
[GPU] Added int32 input/output support for reorder_weights kernels

Lyamin-Roman:
### Tickets:
 - *120301*


steve-y:
@Lyamin-Roman The i32 type error was not reproduced in debug configuration, and the e2e tests passed. Thanks

vladimir-paramuzov (approved):

PR #20036, opened 2023-09-25
[CONFORMANCE] Provide plugin impl status based on QM instead of compile_model

iefode:
### Details:
 - *Merge QM + Compile model impl status to one case*

CC @gkrivor DB update


src/tests/functional/plugin/conformance/test_runner/op_conformance_runner/include/op_impl_check/op_impl_check_test.hpp
```diff
@@ -28,20 +28,30 @@ TEST_P(OpImplCheckTest, checkPluginImplementationCompileModel) {
         crashHandler.StartTimer();
         summary.setDeviceName(targetDevice);
         try {
-            auto executableNetwork = core->compile_model(function, targetDevice, configuration);
```
sbalandi, line 31: Dose it needed to completely remove compile_model or it should be stayed as qm ?
And in https://github.com/openvinotoolkit/openvino/pull/20035/ it should be set summary.setDowngradeCoefficient(3); 2 ?

iefode: Because `query_model` provides supported op

sbalandi (approved):

PR #20039, opened 2023-09-25
[CPU] Group & NF4 decompression support

v-Golubev:
### Details:
 - *NF4 precision support in decompression related transformations*
 - *Group decompression support in transformations*
 - *Updated CPU test class with new instances to cover the changes*

This PR can be merged only after #19900

### DIFF
--- a/src/plugins/intel_cpu/src/graph_optimizer.cpp
+++ b/src/plugins/intel_cpu/src/graph_optimizer.cpp
@@ -286,7 +286,7 @@ void GraphOptimizer::FuseConvMatmulFCDeconvAndDQScales(Graph &graph) {
 }
 
 void GraphOptimizer::FuseFCAndWeightsDecompression(Graph &graph) {
-    const std::set<InferenceEngine::Precision> supportedWeightsPrecisions{InferenceEngine::Precision::U8};
+    std::set<InferenceEngine::Precision> supportedWeightsPrecisions{InferenceEngine::Precision::U8, InferenceEngine::Precision::NF4};
     const std::set<InferenceEngine::Precision> supportedDataPrecisions{InferenceEngine::Precision::FP32, InferenceEngine::Precision::BF16};
     auto expectedNode = [](NodePtr node, Type expectedType) {
         return node->getType() == expectedType && node->getChildEdges().size() == 1;
@@ -301,11 +301,21 @@ void GraphOptimizer::FuseFCAndWeightsDecompression(Graph &graph) {
         if (fcNode == nullptr)
             continue;
 
-        const auto parent = fcNode->getParentEdgesAtPort(1)[0]->getParent();
+        auto parent = fcNode->getParentEdgesAtPort(1)[0]->getParent();
         const bool withTranspose = parent->getType() == Type::Transpose;
         const NodePtr transposeNode = withTranspose ? parent : nullptr;
+        if (transposeNode)
+            parent = transposeNode->getParentEdgesAtPort(0)[0]->getParent();
 
-        const auto multiplyNode = withTranspose ? parent->getParentEdgesAtPort(0)[0]->getParent() : parent;
+        const bool withReshape = parent->getType() == Type::Reshape;
+        const auto reshapeNode = withReshape ? parent : nullptr;
+        if (reshapeNode) {
+            if (reshapeNode->getInputShapeAtPort(0).getRank() != 3 && reshapeNode->getOutputShapeAtPort(0).getRank() != 2)
+                continue;
+            parent = reshapeNode->getParentEdgesAtPort(0)[0]->getParent();
+        }
+
+        const auto multiplyNode = parent;
         if (!expectedNode(multiplyNode, Type::Eltwise) || multiplyNode->getAlgorithm() != Algorithm::EltwiseMultiply ||
             !multiplyNode->isConstant())
             continue;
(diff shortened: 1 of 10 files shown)


src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -301,11 +301,21 @@ void GraphOptimizer::FuseFCAndWeightsDecompression(Graph &graph) {
...
         const bool withTranspose = parent->getType() == Type::Transpose;
         const NodePtr transposeNode = withTranspose ? parent : nullptr;
+        if (transposeNode)
+            parent = transposeNode->getParentEdgesAtPort(0)[0]->getParent();
 
-        const auto multiplyNode = withTranspose ? parent->getParentEdgesAtPort(0)[0]->getParent() : parent;
+        const bool withReshape = parent->getType() == Type::Reshape;
+        const auto reshapeNode = withReshape ? parent : nullptr;
+        if (reshapeNode) {
+            if (reshapeNode->getInputShapeAtPort(0).getRank() != 3 && reshapeNode->getOutputShapeAtPort(0).getRank() != 2)
```
dmitry-gorokhov, line 313: We have shape limitation section below. Can this condition be moved there?

src/plugins/intel_cpu/src/graph_optimizer.cpp
```diff
@@ -373,9 +399,9 @@ void GraphOptimizer::FuseFCAndWeightsDecompression(Graph &graph) {
         }
 
         // Fusion processing
-        fcNode->fuseDecompressionMultiply(multiplyConstNode);
+        fcNode->fuseDecompressionMultiplyPtr(multiplyConstNode);
```
dmitry-gorokhov, line 402: We still fuse decompression node itself, not ptr. So lets get rid of the suffix :)

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
```diff
@@ -1123,27 +1123,20 @@ bool FullyConnected::useSparseWeightsDecompression() {
...
-void FullyConnected::fuseDecompressionConstant(const NodePtr& constData, std::vector<float>& decompressionValues) {
+void FullyConnected::fuseDecompressionConstantPtr(const NodePtr& constData, MemoryCPtr& decompressionValuesPtr) {
     auto *constInputNode = dynamic_cast<node::Input *>(constData.get());
     if (!constInputNode) {
         IE_THROW() << "Cannot cast " << constData->getName() << " to Input";
     }
-    auto constBlob = constInputNode->getMemoryPtr();
-    const auto elementsCount = constBlob->getDescWithType<BlockedMemoryDesc>()->getPaddedElementsCount();
-    decompressionValues.resize(elementsCount);
-    cpu_convert(constBlob->getData(),
```
dmitry-gorokhov, line 1142: Don't we need precision convertion anymore? Looking on the test zero-point is expressed as low precision constant.

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/matmul_weights_decompression.cpp
```diff
@@ -94,34 +95,64 @@ class MatmulWeightsDecompression : public testing::WithParamInterface<MatmulWeig
...
+                std::swap(*result_shape.rbegin(), *(result_shape.rbegin() + 1));
+            return result_shape;
         };
 
-        auto weights_shape = transpose_if_necessary(inputShapes[1].to_shape());
-        auto weights = ngraph::builder::makeConstant<uint8_t>(weights_precision, weights_shape, {}, true);
+        // Weights has shape [I, O], where
+        // I - input channels
+        // O - output channels
+        // If group size greater than 1, input channels dimension are split into 2: I -> [N, G], where
```
dmitry-gorokhov, line 115: This logic introduces confusion. I expect that in case G = 1, N should be equal IC. Can we use '-1' as default value for G which will trigger test to init N = 1?

src/plugins/intel_cpu/src/graph_optimizer.cpp
v-Golubev: Yes, we can move it. Done

src/plugins/intel_cpu/src/graph_optimizer.cpp
v-Golubev: No problem, the suffix is removed)

src/plugins/intel_cpu/src/nodes/fullyconnected.cpp
v-Golubev: You are right: we still need conversion, but only if constant precision is not equal to decompression precision (FP32 for the moment). I added this logic
However, as I remember, we planned to store nf4 decompression constants as is, without conversion. So this logic should be changed when nf4 support on oneDNN side is added

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/matmul_weights_decompression.cpp
v-Golubev: Default value is introduced, thanks for the suggestion. I also introduced boolean variable `group_decompression` to make it easier to match explanatory comments with the code

dmitry-gorokhov (approved):

PR #20042, opened 2023-09-26
[GPU] Gather needs to keep the original input/output rank

sungeunk:
- because the parameters as indices, batch_dims and axis depend on the rank.
- add input_shape to gather primitive.

### DIFF
--- a/src/plugins/intel_gpu/src/plugin/ops/gather.cpp
+++ b/src/plugins/intel_gpu/src/plugin/ops/gather.cpp
@@ -119,6 +119,7 @@ void CreateGatherOpBase(ProgramBuilder& p, const std::shared_ptr<T>& op, const i
                                         reordered_inputs[0],
                                         reordered_inputs[1],
                                         axis,
+                                        input_rank,
                                         out_shape,
                                         batch_dim,
                                         support_neg_ind);
--- a/src/plugins/intel_gpu/include/intel_gpu/primitives/gather.hpp
+++ b/src/plugins/intel_gpu/include/intel_gpu/primitives/gather.hpp
@@ -21,6 +21,7 @@ struct gather : public primitive_base<gather> {
     /// @param dict Input dictionary primitive id.
     /// @param idx Input indexes primitive id.
     /// @param axis Gathering axis.
+    /// @param input_rank Input rank.
     /// @param output_shape Output shape.
     /// @param batch_dim Batch_dim
     /// @param support_neg_ind Support negative indexes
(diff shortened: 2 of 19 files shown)


src/plugins/intel_gpu/src/plugin/ops/gather.cpp
```diff
@@ -119,6 +119,7 @@ void CreateGatherOpBase(ProgramBuilder& p, const std::shared_ptr<T>& op, const i
                                         reordered_inputs[0],
                                         reordered_inputs[1],
                                         axis,
+                                        input_shape.to_shape(),
                                         out_shape,
                                         batch_dim,
```
isanghao, line 124: could you add unittest for this target usage?

sungeunk: There  is the mixed_ranks_gather at src/plugins/intel_gpu/tests/unit/passes/reorder_inputs_test.cpp.
That's enough to check it.

ngaloppo: If the test was already there, why did it not catch the bug?

sungeunk: At bad commit, the input/output rank were compared incorrectly as this old code.

src/plugins/intel_gpu/tests/unit/passes/reorder_inputs_test.cpp
```c++
TEST(reorder_inputs, mixed_ranks_gather) {
    // Topology:
    // (4d) -> conv -> (4d) -> border -> (4d) -> gather -> (5d) -> gather -> (6d) -> permute (6d)
    // In case when preferred format for convolution is selected as byxf (in the test it's enforced)
    // it could be propagated to border and gathers, but dimensions are handled incorrectly
    // and the second gather may complain that axis >= rank
    // So here we expect that input format for gather is aligned with actual output rank and format
...
    ASSERT_EQ(gather1_node.get_input_layouts()[0].format, format::bfzyx);   <<< 5d input. It should be 4d (bfyx)
    ASSERT_EQ(gather1_node.get_output_layout().format, format::bfzyx);      <<< 5d output

    ASSERT_EQ(gather2_node.get_input_layouts()[0].format, format::bfwzyx);  <<< 6d input. It should be 5d (bfzyx)
    ASSERT_EQ(gather2_node.get_output_layout().format, format::bfwzyx);     <<< 6d output
...
```

isanghao (approved):
LGTM

isanghao (approved):

PR #20043, opened 2023-09-26
[core]Api 2.0/migrate logical reduce ops to new API

praasz:
### Details:
 - Migrate ReduceLogicalAnd, ReduceLogicalOr to new API. Remove legacy namespace, utils, evaluate.
 - Refactor ops to reduce binary size.
 - Refactor reference to reduce binary size.

### Tickets:
 - [CVS-120966](https://jira.devtools.intel.com/browse/CVS-120966)
 - [CVS-120967](https://jira.devtools.intel.com/browse/CVS-120967)

### DIFF
--- a/src/core/src/op/reduce_logical_and.cpp
+++ b/src/core/src/op/reduce_logical_and.cpp
@@ -2,76 +2,61 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-#include "ngraph/op/reduce_logical_and.hpp"
+#include "openvino/op/reduce_logical_and.hpp"
 
-#include <ngraph/validation_util.hpp>
-
+#include "element_visitor.hpp"
 #include "itt.hpp"
-#include "ngraph/log.hpp"
-#include "ngraph/op/util/evaluate_helpers.hpp"
-#include "ngraph/runtime/host_tensor.hpp"
+#include "openvino/core/shape_util.hpp"
+#include "openvino/op/util/axes_util.hpp"
 #include "openvino/reference/logical_reduction.hpp"
-#include "openvino/util/log.hpp"
 
-using namespace ngraph;
-using namespace std;
+namespace ov {
+namespace op {
+namespace reduce_and {
+struct Evaluate : element::NoAction<bool> {
+    using element::NoAction<bool>::visit;
 
-op::v1::ReduceLogicalAnd::ReduceLogicalAnd(const Output<Node>& data,
-                                           const Output<Node>& reduction_axes,
-                                           const bool keep_dims)
+    template <element::Type_t ET>
+    static result_type visit(const Tensor& in, Tensor& out, const AxisSet& reduction_axes) {
+        using T = fundamental_type_for<ET>;
+        reference::reduce_logical_and(in.data<const T>(), out.data<T>(), in.get_shape(), reduction_axes);
+        return true;
+    }
+};
+}  // namespace reduce_and
+
+namespace v1 {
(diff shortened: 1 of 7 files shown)


src/core/src/op/reduce_logical_and.cpp
```diff
@@ -2,76 +2,61 @@
...
-    const auto& out = outputs[0];
-    if (data->get_element_type() != element::boolean || !axes->get_element_type().is_integral_number()) {
-        return false;
-    }
-    const auto reduction_axes =
-        get_normalized_axes_from_tensor(axes, data->get_partial_shape().rank(), get_friendly_name());
-    OPENVINO_SUPPRESS_DEPRECATED_END
-    return reduce_and::evaluate_reduce_logical_and(data, out, reduction_axes, get_keep_dims());
+    const auto reduction_axes = get_normalized_axes_from_tensor(this, inputs[1], inputs[0].get_shape().size());
+    outputs[0].set_shape(ov::util::reduce(inputs[0].get_shape(), reduction_axes, get_keep_dims()));
```
vurusovs, line 47: May it corrupt output shape in case of unsupported type? Previous implementation applied `set_shape` after type validation

src/core/src/op/reduce_logical_and.cpp
```diff
@@ -2,76 +2,61 @@
...
     OPENVINO_SUPPRESS_DEPRECATED_START
-    OPENVINO_ASSERT(validate_host_tensor_vector(inputs, 2));
-    OPENVINO_ASSERT(validate_host_tensor_vector(outputs, 1));
+    OPENVINO_ASSERT(inputs.size() == 2);
+    OPENVINO_ASSERT(outputs.size() == 1);
 
-    const auto& data = inputs[0];
-    const auto& axes = inputs[1];
-    const auto& out = outputs[0];
-    if (data->get_element_type() != element::boolean || !axes->get_element_type().is_integral_number()) {
```
vurusovs, line 65: Is `!axes->get_element_type().is_integral_number()` will be handled somewhere in scope of `evaluate`?

praasz: In others reduction this check is not done in evaluate and `get_normalized_axes_from_tensor` do cast to integral and normalization. The reduction will contain only integral values.
Is also done to more consistent with other reduce operators.

src/core/src/op/reduce_logical_and.cpp
praasz: It will resize, but if evaluate return false anything in output cannot be treated as valid data or tensor. And usually fail in evaluate stop the whole  data processing.

I've seen these two approaches in evaluate, change shape before validation or after, in both cases there were no issues but check after type validation is part of template implementation which increase binary size.

Is also remove outside template implementation to reduce size.

jane-intel: I propose to use has_evaluate at the begining of all the evaluates to solve it.

jane-intel (approved):

praasz:
@jane-intel  Regarding `has_evaluate`.

It is used usually to check if evaluate can be called. Also `has_evaluate` assumes there are inputs set and evaluate will not work when operator created with default ctor.

vurusovs (approved):

src/core/src/op/reduce_logical_and.cpp
praasz: It is used usually to check if evaluate can be called. Also `has_evaluate` assumes there are inputs set and evaluate will not work when operator has been created by default ctor.

t-jankowski (approved):

PR #20044, opened 2023-09-26
[PD FE] loading weight from ov::tensor

xczhai:
### Details:
 - *support model in ov::tensor format*

### Tickets:
 - *121052*
 - https://github.com/openvinotoolkit/openvino/issues/19833

### DIFF
--- a/src/frontends/paddle/src/frontend.cpp
+++ b/src/frontends/paddle/src/frontend.cpp
@@ -135,29 +135,29 @@ bool normalize_framework_node(const std::shared_ptr<FrameworkNode>& node,
     return true;
 }
 
-std::istream* variant_to_stream_ptr(const ov::Any& variant, std::ifstream& ext_stream, std::stringstream& ext_stream2) {
+std::istream* variant_to_stream_ptr(const ov::Any& variant,
+                                    std::ifstream& ext_ifstream,
+                                    std::stringstream& ext_sstream) {
     if (variant.is<std::istream*>()) {
         return variant.as<std::istream*>();
     } else if (variant.is<std::shared_ptr<ngraph::runtime::AlignedBuffer>>()) {
         auto& aligned_weights_buffer = variant.as<std::shared_ptr<ngraph::runtime::AlignedBuffer>>();
-        auto filesize = aligned_weights_buffer->size();
-        auto buf_ptr = aligned_weights_buffer->get_ptr<char>();
-        ext_stream2.write(buf_ptr, filesize);
-        FRONT_END_INITIALIZATION_CHECK(ext_stream2 && ext_stream2.good(), "[dbg] Cannot open model file.");
-        return &ext_stream2;
+        ext_sstream.write(aligned_weights_buffer->get_ptr<char>(), aligned_weights_buffer->size());
+        FRONT_END_INITIALIZATION_CHECK(ext_sstream && ext_sstream.good(), "Cannot open ov::tensor.");
+        return &ext_sstream;
     } else if (variant.is<std::string>()) {
         const auto& model_path = variant.as<std::string>();
-        ext_stream.open(model_path, std::ios::in | std::ifstream::binary);
-    } 
+        ext_ifstream.open(model_path, std::ios::in | std::ifstream::binary);
+    }
 #if defined(OPENVINO_ENABLE_UNICODE_PATH_SUPPORT) && defined(_WIN32)
     else if (variant.is<std::wstring>()) {
         const auto& model_path = variant.as<std::wstring>();
-        ext_stream.open(model_path.c_str(), std::ios::in | std::ifstream::binary);
+        ext_ifstream.open(model_path.c_str(), std::ios::in | std::ifstream::binary);
     }
 #endif
 
-    FRONT_END_INITIALIZATION_CHECK(ext_stream && ext_stream.is_open(), "Cannot open model file.");
-    return &ext_stream;
+    FRONT_END_INITIALIZATION_CHECK(ext_ifstream && ext_ifstream.is_open(), "Cannot open model file.");
+    return &ext_ifstream;
 }
 }  // namespace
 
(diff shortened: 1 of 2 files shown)


src/frontends/paddle/src/frontend.cpp
```diff
@@ -135,22 +135,32 @@ bool normalize_framework_node(const std::shared_ptr<FrameworkNode>& node,
     return true;
 }
 
-std::istream* variant_to_stream_ptr(const ov::Any& variant, std::ifstream& ext_stream) {
+OPENVINO_SUPPRESS_DEPRECATED_START
```
riverlijunjie, line 138: It seems `ngraph::runtime::AlignedBuffer` will be deprecated, but this function will not, right?

xczhai: > It seems `ngraph::runtime::AlignedBuffer` will be deprecated, but this function will not, right?

@riverlijunjie Yes, your understanding is correct. The function won't be deprecated.

src/frontends/paddle/src/frontend.cpp
```diff
@@ -135,22 +135,32 @@ bool normalize_framework_node(const std::shared_ptr<FrameworkNode>& node,
...
+std::istream* variant_to_stream_ptr(const ov::Any& variant,
+                                    std::ifstream& ext_ifstream,
+                                    std::stringstream& ext_sstream) {
```
riverlijunjie, line 141: Can we merge `std::ifstream& ext_ifstream` and `std::stringstream& ext_sstream` into together, because we can see that each branch only uses one of them.

src/frontends/paddle/src/frontend.cpp
```diff
@@ -135,22 +135,32 @@ bool normalize_framework_node(const std::shared_ptr<FrameworkNode>& node,
...
     if (variant.is<std::istream*>()) {
         return variant.as<std::istream*>();
+    } else if (variant.is<std::shared_ptr<ngraph::runtime::AlignedBuffer>>()) {
```
ilyachur, line 144: Could you please add new tests

xczhai: Yes. Add it.

src/frontends/paddle/src/frontend.cpp
xczhai: > Can we merge `std::ifstream& ext_ifstream` and `std::stringstream& ext_sstream` into together, because we can see that each branch only uses one of them.

@riverlijunjie I simplify the function variant_to_stream_ptr. Please review when you are free.

src/frontends/paddle/src/frontend.cpp
```diff
@@ -135,22 +135,32 @@ bool normalize_framework_node(const std::shared_ptr<FrameworkNode>& node,
...
 
-std::istream* variant_to_stream_ptr(const ov::Any& variant, std::ifstream& ext_stream) {
+OPENVINO_SUPPRESS_DEPRECATED_START
+void variant_to_stream_ptr(const ov::Any& variant, std::stringstream& ss) {
     if (variant.is<std::istream*>()) {
-        return variant.as<std::istream*>();
+        ss << variant.as<std::istream*>()->rdbuf();
+    } else if (variant.is<std::shared_ptr<ngraph::runtime::AlignedBuffer>>()) {
+        auto& aligned_weights_buffer = variant.as<std::shared_ptr<ngraph::runtime::AlignedBuffer>>();
+        ss.write(aligned_weights_buffer->get_ptr<char>(), aligned_weights_buffer->size());
```
riverlijunjie, line 144: It will bring extra memory copying, which may impact performance.
Could we avoid memory copying?

xczhai: @riverlijunjie it is a trade-off here.
Clean code style may conflict with performance in this case because this function deals with several different IO data type.
So, it is hard to meet the two requirements at the same time.

src/frontends/paddle/src/frontend.cpp
```diff
@@ -392,9 +402,17 @@ bool FrontEnd::supported_impl(const std::vector<ov::Any>& variants) const {
 #endif
     else if (variants[0].is<std::istream*>()) {
         // Validating first stream, it must contain a model
-        auto p_model_stream = variants[0].as<std::istream*>();
+        // step 1:
+        // PDPD API ParseFromIstream always deconstructs the context in model stream.
+        // So, make a copy for variants[0] to avoid breaking the context in variants[0].
+        const auto p_model_stream = variants[0].as<std::istream*>();
+        std::istream copy_model_stream(p_model_stream->rdbuf());
```
riverlijunjie, line 409: This is another issue fixing? Do we have other idea to avoid context desconstruction except for doing copy?

xczhai: > This is another issue fixing? Do we have other idea to avoid context desconstruction except for doing copy?

@riverlijunjie Yes, it is a bug and no test case to cover. 
Actually, it can be avoided but don't worry too much. The copy behavior is just about model file instead of weight file so the negative impact is little.

riverlijunjie (approved):
LGTM

src/frontends/paddle/src/frontend.cpp
ceciliapeng2011: @ilyachur @xczhai Is it okay to declare OPENVINO_SUPPRESS_DEPRECATED_START and OPENVINO_SUPPRESS_DEPRECATED_END only for the branch of AlignedBuffer instead of the whole fucntion?

src/frontends/paddle/src/frontend.cpp
```diff
@@ -135,22 +135,32 @@ bool normalize_framework_node(const std::shared_ptr<FrameworkNode>& node,
     return true;
 }
 
-std::istream* variant_to_stream_ptr(const ov::Any& variant, std::ifstream& ext_stream) {
+OPENVINO_SUPPRESS_DEPRECATED_START
+std::istream* variant_to_stream_ptr(const ov::Any& variant, std::fstream& fs, std::stringstream& ss) {
```
ceciliapeng2011, line 139: Can we use fstream to handle case of AlignedBuffer? Any particular reason to use stringstream?

src/frontends/paddle/src/frontend.cpp
ceciliapeng2011: @riverlijunjie any idea to share the AlignedBuffer's memory to avoid copy?

src/frontends/paddle/src/frontend.cpp
xczhai: Maybe not. @ceciliapeng2011 
- fstream should be open before use but not file can be open.
- The return type is `istream` so we should return a derived type pointer. AFAIK, stringstream as a derived type can be filled with an internal buffer.

sdcb:
@ceciliapeng2011 @riverlijunjie 
We should also include this into 2023.2 release since https://github.com/openvinotoolkit/openvino/pull/19931 is also included.

PR #20047, opened 2023-09-26
[Transformations] Fix unroll if transformation rt info

nkogteva:
### Details:
 - *Don't copy rt info to operations which are initially were inside operation body in ov::pass::UnrollIf*
 - *Keep original operation name during unrolling in ov::pass::UnrollIf*

### Tickets:
 - *ticket-id*


src/common/transformations/src/transformations/control_flow/unroll_if.cpp
```diff
@@ -42,6 +42,9 @@ bool ov::pass::UnrollIf::run_on_model(const std::shared_ptr<ov::Model>& f) {
         auto body = (cond_value[0]) ? if_node->get_then_body() : if_node->get_else_body();
         auto input_descriptions = if_node->get_input_descriptions(static_cast<int>(!cond_value[0]));
         auto output_descriptions = if_node->get_output_descriptions(static_cast<int>(!cond_value[0]));
+        // copy rt info before reconnection
+        for (auto& op : body->get_ops())
+            copy_runtime_info({op, if_node}, op);
```
itikhono, line 47: why do we need to copy rt info from "op" to "op"?
I think it should be
copy_runtime_info(if_node, op);

nkogteva: > why do we need to copy rt info from "op" to "op"? I think it should be copy_runtime_info(if_node, op);

`copy_runtime_info(if_node, op)` replaces rt_info of `op` by rt_info from `if_node`

nkogteva: > I don't think so, `copy_runtime_info` calls this function inside image As I can see, only repeating attributes will be overwritten, other attributes should remain

I have added test checks which are failed with `copy_runtime_info(if_node, op)`

itikhono: I've double checked. 
`copy_runtime_info({op, if_node}, op);`  
calls `merge` method of `FusedNames` attribute for `{op, if_node}` and doesn't call it otherwise.
Looks correct. Resolved.

itikhono (approved):

PR #20077, opened 2023-09-27
【Hackathon 5th No.95】add paddle unique op

AndPuQing:
### Details:
 - add paddle unique op convert support

### reference

- https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/unique_en.html
- https://docs.openvino.ai/2023.1/openvino_docs_ops_movement_Unique_10.html

image

### Note

dtype **float32, int32, int64** would be supported, **float64** ov [TestConvert not be supported](https://github.com/openvinotoolkit/openvino/blob/master/src/frontends/tests/frontend/shared/src/op_fuzzy.cpp#L71)


src/frontends/paddle/src/op/unique.cpp
```diff
@@ -0,0 +1,41 @@
...
+    NamedOutputs named_outputs;
+
+    auto axis = node.get_attribute<std::vector<int32_t>>("axis");
+
+    auto dtype_str = node.get_attribute<std::string>("dtype");
+    auto dtype = dtype_str == "int32" ? element::i32 : element::i64;
+
+    if (axis.size() != 0) {
+        auto axis_node = std::make_shared<default_opset::Constant>(element::i32, Shape{}, axis);
+        outputs = std::make_shared<ov::opset10::Unique>(x, axis_node, true, dtype, dtype)->outputs();
```
OpenVINO-dev-contest, line 26: Can we use default_opset other than opset10 ?

src/frontends/paddle/tests/test_models/gen_scripts/generate_unique.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def unique(
+    name: str,
+    x,
+    reture_index=False,
+    reture_inverse=False,
+    reture_counts=False,
+    dtype="int64",
+    axes=None,
```
OpenVINO-dev-contest, line 22: Can we leave these parameter's name as origin one, which can be loaded to ```paddle.unique``` directly

src/frontends/paddle/src/op/unique.cpp
AndPuQing: To the best of my knowledge, the unique operator is defined in opset10

- https://docs.openvino.ai/2023.1/openvino_docs_ops_movement_Unique_10.html

OpenVINO-dev-contest:
@yuxu42 Could you help to have a review on this PR, thanks

src/frontends/paddle/tests/test_models/gen_scripts/generate_unique.py
AndPuQing: Already fixed.

yuxu42:
@ceciliapeng2011 could you please help reivew? Thanks!

src/frontends/paddle/tests/test_models/gen_scripts/generate_unique.py
```diff
@@ -0,0 +1,75 @@
...
+    with paddle.static.program_guard(paddle.static.Program(), paddle.static.Program()):
+        node_x = paddle.static.data(name="x", shape=x.shape, dtype=data_type)
+
+        unique_outs = paddle.unique(node_x, **op_args)
+        if isinstance(unique_outs, tuple):
+            outputs = []
+            for out in unique_outs:
+                if out is not None:
+                    if out.dtype == paddle.int64 or out.dtype == paddle.int32:
+                        out = paddle.cast(out, "float32")
```
xczhai, line 27: I can't figure out why `cast` operation is needed. In my understanding, we should keep the original `unique` output for verification.

src/frontends/paddle/src/op/unique.cpp
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+        outputs = std::make_shared<ov::opset10::Unique>(x, true, dtype, dtype)->outputs();
+    }
+
+    named_outputs["Index"] = {outputs[2]};
+    named_outputs["Indices"] = {outputs[1]};
+    named_outputs["Counts"] = {outputs[3]};
```
xczhai, line 33: - In OpenVINO unique spec, `outputs[3]` type is controled by `count_element_type`. https://docs.openvino.ai/2023.1/openvino_docs_ops_movement_Unique_10.html.
- In PaddlePaddle spec, I don't find the `return_counts` dtype. I am worried because I run into the similar issue before. Could you confirm it here? I

src/frontends/paddle/src/op/unique.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace ov {
+namespace frontend {
+namespace paddle {
+namespace op {
+NamedOutputs unique(const NodeContext& node) {
+    auto x = node.get_input("X");
+
+    std::vector<Output<Node>> outputs;
+    NamedOutputs named_outputs;
```
xczhai, line 17: As far as I know, the `NamedOutputs` is a `std::map`. 
In this case, you'd better use constructor with initailization list insteading of inserting k-v one by one, I think.

src/frontends/paddle/tests/test_models/gen_scripts/generate_unique.py
AndPuQing: AFAIK, the openvino test [compare_results](https://github.com/openvinotoolkit/openvino/blob/master/src/tests/test_utils/common_test_utils/src/test_case.cpp#L169-L179) only support fp32 & int32, so for the int64 index we should convert to float32/int32

src/frontends/paddle/src/op/unique.cpp
AndPuQing: In paddle, the `count_element_type` and `index_element_type` both be controlled by [attr_dtype](https://github.com/PaddlePaddle/Paddle/blob/release/2.5/python/paddle/tensor/manipulation.py#L2514-L2522)

src/frontends/paddle/src/op/unique.cpp
AndPuQing: Done

AndPuQing:
> @AndPuQing could you please fix the issues and rebase the code? Thanks!

Done

src/frontends/paddle/tests/test_models/gen_scripts/generate_unique.py
xczhai: @AndPuQing Thanks for your very detailed analysis. Your finding makes sense.
Maybe the test case framework is not perfect enough to call. Don't worry at all. If you think it is not reasonable, just pull PR to fix or optimize it, which is much appreciated. We should NOT skip it using a work-around.
If you have no time for this case, let me know.

AndPuQing: Sorry, it seems I may not have the time to rectify the test case framework.

xczhai (approved):

OpenVINO-dev-contest:
Could you change the PR's title as 【Paddle Hackathon 5th No.95】?

yuxu42:
@ceciliapeng2011  can we merge the PR? Thanks!

PR #20082, opened 2023-09-27
[CPU] Postpone throwing error for dnnl memory until it's requested

maxnick:
### Details:
StaticMemory object tries to create dnnl::memory object in the constructor. However, this memory object is used only by dnnl primitives and may be never requested if it isn't an input of a Node that uses a oneDNN primitive. However, it's still better to try to create the object inside the constructor to avoid data race issues. So the solution is to try to create dnnl::memory object and throw only when it's requested instead of throwing from the constructor.

### Tickets:
 - 121363
 - 120279


maxnick:
@riverlijunjie , FIY

src/plugins/intel_cpu/src/cpu_memory.cpp
```diff
@@ -288,24 +288,28 @@ StaticMemory::StaticMemory(const dnnl::engine& eng, MemoryDescPtr desc, const vo
...
+        //
+        // ========================
+        if (pads_zeroing)
+            m_prim.set_data_handle(m_pMemMngr->getRawPtr());
+        else
+            m_prim.set_data_handle_no_pads_proc(m_pMemMngr->getRawPtr());
+    }
+    catch (const std::exception& exc) {
+        dnnlErrorCtx = exc.what();
+    }
```
riverlijunjie, line 312: It does can handle "const->result" graph for unsupported precision, but it seems it cannot handle "const->convert->result" graph, such as:
`Constant(u16)->convert(u16->f32)->result(f32)`

Could you create such test cases?

src/plugins/intel_cpu/src/dnnl_extension_utils.cpp
```diff
@@ -68,7 +68,7 @@ memory::data_type DnnlExtensionUtils::IEPrecisionToDataType(const InferenceEngin
         case InferenceEngine::Precision::UNSPECIFIED:
             return memory::data_type::undef;
         default: {
-            OPENVINO_THROW("The plugin does not support ", prec.name());
+            IE_THROW() << "OneDNN doesn't support " << prec.name();
```
riverlijunjie, line 71: It will be better to restore this change.

maxnick: Yep, agree, we shouldn't use `IE_THROW()` as a deprecated macro.

src/plugins/intel_cpu/src/cpu_memory.cpp
maxnick: Aren't the enabled tests enough? Your proposed subgraph will likely be folded during the constant folding pass.

maxnick:
@dmitry-gorokhov , could you please take a look?

dmitry-gorokhov (approved):

PR #20086, opened 2023-09-27
`ROIPooling`, `ROIAlign` layer tests to API2.0

vurusovs:


src/tests/functional/plugin/shared/include/single_op_tests/roi_align.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include "shared_test_classes/single_op/roi_align.hpp"
+
+namespace ov {
+namespace test {
+TEST_P(ROIAlignLayerTest, Inference) {
+    run();
+}
+
+TEST_P(ROIAlignV9LayerTest, Inference) {
```
vurusovs, line 15: @iefode please, note that the test isn't instantiated. Who should enable it?

src/tests/functional/shared_test_classes/src/single_op/roi_align.cpp
```diff
@@ -0,0 +1,201 @@
...
+}
+
+static int randInt(int low, int high) {
+    std::random_device rd;
+    std::mt19937 gen(rd());
+    std::uniform_int_distribution<int> dis(low, high);
+    return dis(gen);
+}
+
+void ROIAlignLayerTest::fillCoordTensor(std::vector<float>& coords, int height, int width,
```
iefode, line 56: Hm. Coordinates? In this case why we use `float`? Looks like should be renamed

iefode (approved):

src/tests/functional/plugin/shared/include/single_op_tests/roi_align.hpp
iefode: Could you please check who is author and ask him? If the author would not provide correct answer, remove this case from lib. Thanks for your attention :)

vurusovs: @iefode it's used in GPU tests, so I'll leave it as is

src/tests/functional/shared_test_classes/src/single_op/roi_align.cpp
vurusovs: I have no idea how it should be renamed, so will port as is. If you have some suggestions, I'm open to apply it

PR #20105, opened 2023-09-28
[TF FE] Support TF1 While Control flow

rkazants:
**Details:** TensorFlow models can have low control flow While operation that is represented as cyclic graph with low control flow operations: Switch, Merge, NextIteration, Enter, Exit, and LoopCond.

Example - TF with Control flow While:
image

**Ticket:** 108581

### DIFF
--- a/src/frontends/tensorflow/src/input_model.cpp
+++ b/src/frontends/tensorflow/src/input_model.cpp
@@ -137,6 +137,10 @@ void InputModel::InputModelTFImpl::load_places() {
         auto op_name = node_decoder->get_op_name();
         auto op_type = node_decoder->get_op_type();
 
+        if (op_type == "Placeholder" && op_name.rfind("unused_control_flow_input", 0) != std::string::npos) {
+            continue;
+        }
+
         if (m_telemetry) {
             op_statistics[op_type]++;
         }
@@ -320,9 +324,6 @@ std::vector<std::shared_ptr<OpPlace>> InputModel::InputModelTFImpl::topologicall
     std::stack<std::shared_ptr<OpPlace>> ops_to_do;
     std::unordered_set<std::shared_ptr<OpPlace>> ops_done;
 
-    // TODO: implement logic to check direct cycles in the graph
-    // and break them
-    // probably not only NextIteration can generate cycles
     for (const auto& output_place : m_outputs) {
         FRONT_END_GENERAL_CHECK(output_place->get_names().size() > 0, "TensorPlace must have at least one name.");
         auto output_place_name = output_place->get_names()[0];
@@ -334,6 +335,23 @@ std::vector<std::shared_ptr<OpPlace>> InputModel::InputModelTFImpl::topologicall
                                 "Custom specified output is incorrect: " + output_place_name);
         auto output_operation_place = m_op_places_map.at(operation_name);
         ops_to_do.push(output_operation_place);
+    }
+
+    // walk through all NextIteration nodes and put their producers into ops_to_do
+    // this is needed to avoid missed nodes in the body graph of TF1 While operation
+    for (const auto& op_place : m_op_places) {
+        auto op_decoder = op_place->get_decoder();
+        if (op_decoder->get_op_type() == "NextIteration") {
+            std::string producer_name;
+            std::string producer_output_port_name;
+            size_t producer_output_port_idx;
+            op_decoder->get_input_node(0, producer_name, producer_output_port_name, producer_output_port_idx);
+            FRONT_END_GENERAL_CHECK(m_op_places_map.count(producer_name),
+                                    "[TensorFlow Frontend] internal error or inconsistent model: producer of "
+                                    "NextIteration is not found among operation places " +
+                                        producer_name);
+            ops_to_do.push(m_op_places_map.at(producer_name));
+        }
     }
 
     // the traversing algorithm to compute topologically sorted nodes is taken from topological_sort in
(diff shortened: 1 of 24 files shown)


src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -154,6 +159,130 @@ size_t get_flat_index_by_name_and_id(const ov::frontend::NamedOutputVector& outp
...
+        FRONT_END_GENERAL_CHECK(
+            switch_node->get_output_target_inputs(0).size() < 2,
+            "[TensorFlow Frontend] internal error or inconsistent model: Switch node has more than one Exit nodes");
+        if (switch_node->get_output_target_inputs(0).size() == 1) {
+            auto exit_node = (*switch_node->get_output_target_inputs(0).begin()).get_node();
+            ov_outputs[ind] = exit_node->output(0);
+            output_tensor_names[ind] = exit_node->get_friendly_name() + ":0";
+        } else {
+            // no Exit node, use Output with no consumers
+            // ov_outputs[ind] = ov::Output<ov::Node>();
```
rkazants, line 215: need to remove

gkrivor (commented):
Need more time for full review...

src/frontends/tensorflow/src/input_model.cpp
```diff
@@ -336,6 +337,23 @@ std::vector<std::shared_ptr<OpPlace>> InputModel::InputModelTFImpl::topologicall
...
+    // walk through all NextIteration nodes and put their producers into ops_to_do
+    // this is needed to avoid missed nodes in the body graph of TF1 While operation
+    for (const auto& op_place : m_op_places) {
+        auto op_decoder = op_place->get_decoder();
+        if (op_decoder->get_op_type() == "NextIteration") {
+            std::string producer_name;
+            std::string producer_output_port_name;
+            size_t producer_output_port_idx;
+            op_decoder->get_input_node(0, producer_name, producer_output_port_name, producer_output_port_idx);
+            FRONT_END_GENERAL_CHECK(m_op_places_map.count(producer_name),
```
gkrivor, line 349: You use map.count + map.at (not only here), why don't you use find? It allows to do item lookup only once. Correct me if I miss something in this logic.

src/frontends/tensorflow/src/tf_utils.cpp
```diff
@@ -369,6 +372,115 @@ bool propagate_conditional_flow(const OutputVector& ov_inputs,
...
+                            "[TensorFlow Frontend] Internal error or inconsistent model: body graph "
+                            " must have the same number of Parameter nodes as a number of inputs to While.");
+    FRONT_END_GENERAL_CHECK(body_results.size() == input_size,
+                            "[TensorFlow Frontend] Internal error or inconsistent model: body graphs "
+                            " must have the same number of Result nodes as a number of inputs to While.");
+    FRONT_END_GENERAL_CHECK(cond_params.size() == input_size,
+                            "[TensorFlow Frontend] Internal error or inconsistent model: condition graph "
+                            " must have the same number of Parameter nodes as a number of inputs to While.");
+    for (size_t param_ind = 0; param_ind < cond_params_size; ++param_ind) {
+        cond_params[param_ind]->output(0).replace(body_results[param_ind]->input_value(0));
```
gkrivor, line 423: Just curious, by using 0-index we may expect it could have more than 1 input/output. Do we have any reason to unify this operation by using additional loop for replacing all outputs by inputs? or verify it mandatory has only one input/output?

src/frontends/tensorflow/src/translate_session.cpp
```diff
@@ -154,6 +159,124 @@ size_t get_flat_index_by_name_and_id(const ov::frontend::NamedOutputVector& outp
...
+        auto switch_node = ov::as_type_ptr<Switch>(consuming_input.get_node()->shared_from_this());
+        FRONT_END_GENERAL_CHECK(switch_node,
+                                "[TensorFlow Frontend] internal error or inconsistent model: consumer of LoopCond "
+                                "output is not Switch operation");
+        switch_nodes.push_back(switch_node);
+    }
+
+    // collect all output tensors for Loop
+    // the created Loop node outputs will be connected with ov_outputs
+    size_t num_inputs = switch_nodes.size();
```
gkrivor, line 193: Do we need to check it is greater than 0?

gkrivor (commented):
Mostly completed...

src/frontends/tensorflow/src/op/exit.cpp
```diff
@@ -0,0 +1,34 @@
...
+using namespace ov;
+using namespace ov::frontend::tensorflow;
+
+namespace ov {
+namespace frontend {
+namespace tensorflow {
+namespace op {
+
+OutputVector translate_exit_op(const NodeContext& node) {
+    auto node_name = node.get_name();
```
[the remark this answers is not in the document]
rkazants: Fixed.

src/frontends/tensorflow/src/op/enter.cpp
```diff
@@ -0,0 +1,34 @@
...
+using namespace ov;
+using namespace ov::frontend::tensorflow;
+
+namespace ov {
+namespace frontend {
+namespace tensorflow {
+namespace op {
+
+OutputVector translate_enter_op(const NodeContext& node) {
+    auto node_name = node.get_name();
```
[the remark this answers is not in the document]
rkazants: Fixed.

src/frontends/tensorflow/src/translate_session.cpp
rkazants: No, that is because `LoopCond` cannot exist without `Switch` consuming node(s)

src/frontends/tensorflow/src/tf_utils.cpp
rkazants: No, we don't need because we replace a value (tensor value) with another one for all consumers of this value.
This is a logic of `replace` method. ov::inputs will be automatically re-connected to the new value.

src/frontends/tensorflow/src/input_model.cpp
rkazants: Yes, you are correct that it will be slightly faster to use iterators but I do not like to work with STL iterators. The code looks worse create iterator, dereference it and use subscript operator.

gkrivor (approved):

PR #20108, opened 2023-09-28
[core]Api 2.0/migrate Subtract op to new API

praasz:
### Details:
 - Migrate Subtract operator to new API. Remove legacy headers, namespace and classes.
 - Refactor operator and reference implementation to reduce binary size.

### Tickets:
 - [CVS-118673](https://jira.devtools.intel.com/browse/CVS-118673)


src/core/src/op/subtract.cpp
```diff
@@ -2,89 +2,78 @@
...
-bool op::v1::Subtract::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs) const {
+bool Subtract::evaluate(TensorVector& outputs, const TensorVector& inputs) const {
     OV_OP_SCOPE(v1_Subtract_evaluate);
-    return subtract::evaluate_subtract(inputs[0], inputs[1], outputs[0], get_autob());
+    OPENVINO_ASSERT(outputs.size() == 1);
+    OPENVINO_ASSERT(inputs.size() == 2);
+
+    outputs[0].set_shape(infer_broadcast_shape(this, inputs[0].get_shape(), inputs[1].get_shape()));
+    using namespace ov::element;
+    return IfTypeOf<bf16, f16, f32, i8, i32, i64, u8, u32, u64>::apply<subtract::Evaluate>(inputs[0].get_element_type(),
```
vurusovs, line 55: Out of PR scope question - why evaluate supports i8/u8, but it isn't in `has_evaluate`?

@ilyachur @ilya-lavrenov @jane-intel

vurusovs (approved):

src/core/src/op/subtract.cpp
praasz: Compared to `Add` operator the `Subtract` not supports `u16`, `i16`. 
Should be added?

t-jankowski (approved):

src/core/src/op/subtract.cpp
ilyachur: I think we need to align has_evaluate and evaluate methods at least.

ilyachur (approved):

PR #20109, opened 2023-09-28
[PyOV] Fix docs and add test for Model._get_raw_address

akuporos:
### Details:
 - follow up of the https://github.com/openvinotoolkit/openvino/pull/20014

### Tickets:
 - *ticket-id*


jiwaszki (approved):
🙌

PR #20117, opened 2023-09-28
[CPU] FullyConnected: sparse weights fix

antonvor:
### Tickets:
 - [120485](https://jira.devtools.intel.com/browse/CVS-120485)


antonvor:
@dmitry-gorokhov take a look please

vuiseng9:
If it makes sense, please consider implementing tests to prevent the recurrence of the issue. Thank you for the fix!

antonvor:
@vuiseng9 there is a MatMulSparseCPUTest (https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/tests/functional/single_layer_tests/matmul_sparse.cpp) but it requires the host to support AMX isa

dmitry-gorokhov (approved):

PR #20149, opened 2023-09-29
Improve Commit slider's handling of non-decodable characters

IvanNovoselov:
### Details:
 - *Commit slider could have troubles writing to stdout non-unicode characters that it failed to decode. This patch fixes this behavior, so the non-decodable characters will be replaced with "?" now.*

### Tickets:
 - *-*


yury-intel (approved):
Good idea

IvanNovoselov:
@dmitry-gorokhov please take a look

PR #20167, opened 2023-10-02
[PT FE]: support aten::log1p, fixes for where and linalg_norm

eaidova:
### Details:
 - *support aten::log1p*
 - *extended linal_norm for tensors with rang > 2 if ord not provided*
 - *fixed aten::where translation for mixed precision if/else tensors*
 
verified with owl-vit models family

### Tickets:
 - *ticket-id*


src/frontends/pytorch/src/op/norm.cpp
```diff
@@ -259,11 +259,8 @@ OutputVector translate_linalg_norm(const NodeContext& context) {
             auto input_rank = x.get_partial_shape().rank();
             if (input_rank.is_static() && input_rank.get_length() == 2) {
                 result = frobenius_norm(context, x, dim, keep_dim);
-            } else if (input_rank.is_static() && input_rank.get_length() == 1) {
-                result = norm_vector(context, x, dim, 2, keep_dim);
             } else {
-                FRONT_END_OP_CONVERSION_CHECK(false,
```
mvafin, line 265: Does it change the behaivior or just allows dynamic rank? If the later I think it makes sense to reverse the validation. Previously it was working for `input_rank.is_static() && input_rank.get_length() == 1`. Replace it with working if `input_rank.is_dynamic() || input_rank.get_length() == 1`

tests/layer_tests/pytorch_tests/test_log.py
```diff
@@ -42,7 +44,9 @@ def forward(self, x):
                              ["log_", "float32"], 
                              ["log2", "float32"], 
                              ["log2", "int32"], 
-                             ["log2_", "float32"]])
+                             ["log2_", "float32"],
+                             ["log1p", "int32"],
```
rkazants, line 48: Do we need to test `log1p` for float32?

src/frontends/pytorch/src/op/norm.cpp
eaidova: I'm not sure that it is my case, I'll check. I thought that in my case rank known but it is greater 2

eaidova: yes, it works in my case, updated

tests/layer_tests/pytorch_tests/test_log.py
eaidova: added, thanks

mvafin (approved):

PR #20171, opened 2023-10-02
[CPU] RandomUniform-8 implementation.

nshchego:
### Details:
 - *RandomUniform operation implementation based on the Philox algorithm*
 - *JIT kernel for x86*

### Tickets:
 - *111202*


nshchego:
@a-sidorova, could you please take a look?

src/core/src/op/random_uniform.cpp
```diff
@@ -92,6 +92,14 @@ bool RandomUniform::evaluate(TensorVector& outputs, const TensorVector& inputs)
     const auto& t_out = get_out_type();
     OPENVINO_ASSERT(validate::out_et(t_out), "Unsupported type of RandomUniform: " + t_out.get_type_name());
 
+    auto in0_data = inputs[0].data();
+    Shape output_shape;
+    for (int i = 0; i < inputs[0].get_size(); i++) {
+        output_shape.push_back(out_dims[i]);
+    }
+
+    outputs[0].set_shape(output_shape);
```
mitruska, line 101: Can the `out_shape` variable (the result of `shape_infer` L89) be reused here instead of copying from `out_dims` vector?
```suggestion
    outputs[0].set_shape(out_shape);
```

nshchego: Right, thanks

src/plugins/intel_cpu/src/nodes/random_uniform.cpp
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    if (m_const_inputs[MIN_VAL]) {
+        initEdgeValues(m_min_val, as_type<op::v0::Constant>(op->get_input_node_ptr(MIN_VAL))->get_data_ptr(), m_output_prc);
+    }
+    if (m_const_inputs[MAX_VAL]) {
+        initEdgeValues(m_max_val, as_type<op::v0::Constant>(op->get_input_node_ptr(MAX_VAL))->get_data_ptr(), m_output_prc);
+    }
+
+    if (m_algo == STL) {
```
a-sidorova, line 62: Could you tell me please if `m_algo` is just debug parameter? I don't see this parameter in [spec](https://github.com/openvinotoolkit/openvino/blob/master/docs/ops/generation/RandomUniform_8.md).

src/plugins/intel_cpu/src/nodes/random_uniform.hpp
```diff
@@ -0,0 +1,120 @@
...
+    // Philox algorithm returns 4 elements of RNG sequence per each invocation
+    static constexpr uint64_t PHILOX_GROUP_SIZE = 4lu;
+
+    // Output elements number threshold to execute on one thread.
+    static constexpr uint64_t PHILOX_PARALLEL_EXECUTION_THRESHOLD = 1000lu;
+
+    uint64_t m_skip_count = 0lu;
+    /////////////////////////////////////////////////////////////////////////////////
+
+    std::shared_ptr<kernel::JitKernel<kernel::RandomUniformCompileParams, kernel::RandomUniformCallArgs>> m_jit_kernel;
```
a-sidorova, line 115: What's about to add `using RandomUniformKernel = kernel::JitKernel<kernel::RandomUniformCompileParams, kernel::RandomUniformCallArgs>;`?

a-sidorova: Moreover, I'm not sure that it will be built on ARM machine since `jit_kernel` is not compiled on these machines

src/plugins/intel_cpu/src/nodes/random_uniform.cpp
```diff
@@ -0,0 +1,513 @@
...
+        EL_CASE(i64)
+        EL_CASE(f64)
+        default:
+            THROW_CPU_NODE_ERR("has unsupported output precision: ", output_type);
+    }
+
+#undef EL_CASE
+}
+
+std::string RandomUniform::getPrimitiveDescriptorType() const {
```
a-sidorova, line 452: Why do we need to override this method?

a-sidorova (commented):
First part

src/plugins/intel_cpu/src/nodes/random_uniform.cpp
nshchego: The base fn choses the input precision first, but the output precision is important here.

src/plugins/intel_cpu/src/nodes/random_uniform.cpp
nshchego: This flag allows to switch between TF and ONNX ways, check accuracy and performance. At this moment we support just TF op, thus yes, it can be called Debug parameter.

src/frontends/onnx/frontend/src/op/random_uniform.cpp
```diff
@@ -29,6 +29,7 @@ OutputVector random_uniform(const Node& node) {
 
     const auto target_type = common::get_ngraph_element_type(dtype);
     const uint64_t global_seed = 0;
+    // TODO: This multiplication leads to a mismatch in accuracy, thus need to implement a different WA here.
```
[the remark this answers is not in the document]
nshchego: It should be done in scope of ONNX RandomUniform op implementation.

src/plugins/intel_cpu/src/node.h
```diff
@@ -41,6 +41,8 @@
 #include "nodes/executors/mvn_list.hpp"
 #include "nodes/executors/executor.hpp"
 
+#define THROW_CPU_NODE_ERR(...) OPENVINO_THROW(getTypeStr(), " node with name '", getName(), "' ", __VA_ARGS__)
```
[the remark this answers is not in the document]
nshchego: There are just about 10 nodes with such definition. Most of nodes have `errorPrefix` member. In my opinion it more convenient to define it in one place:
- It will be easy to modify that if OPENVINO_THROW will be changed again,
- No need to define macros or errorPrefix in each node.

src/plugins/intel_cpu/src/nodes/kernels/x64/jit_kernel_base.hpp
```diff
@@ -140,7 +166,9 @@ class JitKernelBase: public dnnl::impl::cpu::x64::jit_generator {
         return dnnl::impl::cpu::x64::mayiuse(isa);
     }
 
+    const dnnl::impl::cpu::x64::cpu_isa_t m_isa;
     RegistersPool::Ptr registersPool;
+    size_t vlen;
```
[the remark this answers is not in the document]
nshchego: The variable can be more easily reused in the kernel implementation. Please look at the src/plugins/intel_cpu/src/nodes/kernels/x64/random_uniform.cpp

src/plugins/intel_cpu/src/nodes/kernels/x64/jit_kernel_base.hpp
```diff
@@ -155,5 +183,71 @@ class JitKernelBase: public dnnl::impl::cpu::x64::jit_generator {
...
+    void operator()(const CallArgs& args) const {
+        this->operator()(&args);
+    }
+
+    template <template<dnnl::impl::cpu::x64::cpu_isa_t isa> typename KernelT>
+    static std::shared_ptr<JitKernel<CompileParams, CallArgs>> createInstance(const CompileParams& jcp) {
+        std::shared_ptr<JitKernel<CompileParams, CallArgs>> res;
+
+        try {
+#define IF_ISA_CASE(ISA)                             \
```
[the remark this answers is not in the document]
nshchego: Modified, thanks

src/plugins/intel_cpu/src/nodes/reference.cpp
```diff
@@ -81,18 +69,18 @@ void Reference::executeDynamicImpl(dnnl::stream strm) {
...
         }
     } else {
          IE_THROW(Unexpected) <<
             "Unexpected shape infer result status during the inference of a node with type " <<
             getTypeStr() << " and name " << getName();
     }
-    if (!ngraphOp->evaluate(outputs, inputs)) {
-        IE_THROW() << "Evaluation failed on node of type: " << std::string(ngraphOp->get_type_name()) << " name: " << getName();
+    if (!ovCoreNode->evaluate(outputs, inputs)) {
+        IE_THROW() << "Evaluation failed on node of type: " << std::string(ovCoreNode->get_type_name()) << " name: " << getName();
```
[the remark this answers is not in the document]
nshchego: Done

src/plugins/intel_cpu/src/shape_inference/custom/random_uniform.cpp
```diff
@@ -0,0 +1,41 @@
...
+        auto data = reinterpret_cast<const int64_t*>(mem->getData());
+        dims.assign(data, data + rank);
+    }
+
+    return {{dims}, ShapeInferStatus::success};
+}
+
+RandomUniformShapeInferFactory::RandomUniformShapeInferFactory(const std::shared_ptr<ov::Node>& op) : m_op(op) {
+    if (!ov::is_type<const op::v8::RandomUniform>(m_op)) {
+        OPENVINO_THROW("Unexpected op type in RandomUniform shape inference factory: ", m_op->get_type_name());
```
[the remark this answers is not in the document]
nshchego: Done

src/plugins/intel_cpu/tests/functional/single_layer_tests/random_uniform.cpp
```diff
@@ -0,0 +1,388 @@
...
+        bool,                            // Is 3rd input constant
+        ov::AnyMap,                      // Additional plugin configuration
+        CPUTestUtils::CPUSpecificParams
+> RandomUniformLayerTestCPUParamSet;
+
+class RandomUniformLayerTestCPU : public testing::WithParamInterface<RandomUniformLayerTestCPUParamSet>,
+                                  public SubgraphBaseTest, public CPUTestUtils::CPUTestsBase {
+public:
+    static std::string getTestCaseName(const testing::TestParamInfo<RandomUniformLayerTestCPUParamSet>& obj) {
+        const auto& out_sahpe        = std::get<0>(obj.param);
```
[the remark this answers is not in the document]
nshchego: Done

src/plugins/intel_cpu/tests/functional/single_layer_tests/random_uniform.cpp
```diff
@@ -0,0 +1,388 @@
...
+        std::ostringstream result;
+        result << "IS={" << out_sahpe.size();
+        result << "}_OS=" << out_sahpe;
+        result << "_Min=" << std::get<0>(min_max);
+        result << "_Max=" << std::get<1>(min_max);
+        result << "_ShapePrc=" << shape_prc;
+        result << "_OutPrc=" << output_prc;
+        result << "_GlobalSeed=" << global_seed;
+        result << "_OperationalSeed=" << operational_seed;
+        result << "_ConstIn={" << (const_in_1 ? "True," : "False,") << (const_in_2 ? "True," : "False,") << (const_in_3 ? "True}" : "False}");
```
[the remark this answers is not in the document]
nshchego: Done

src/plugins/intel_cpu/src/nodes/random_uniform.hpp
nshchego: `using RandomUniformKernel` was added.
About the ARM: there is another execution branch without the kernel.

a-sidorova: I understand that we have another execution branch. But `JitKernel` is defined in [jit_kernel_base.hpp](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/nodes/kernels/x64/jit_kernel_base.hpp). But this file is not [compiled](https://github.com/openvinotoolkit/openvino/blob/a13cc8c733c443bf51091e3cf8c9bddbeb11ccef/src/plugins/intel_cpu/CMakeLists.txt#L82) on ARM. I think that this code won't be compiled on ARM. Maybe I missed something?

src/plugins/intel_cpu/src/nodes/kernels/x64/jit_kernel_base.cpp
```diff
@@ -127,6 +144,45 @@ void JitKernelBase::uni_vpsubd(const Xbyak::Ymm& vDst,
     }
 }
 
+void JitKernelBase::uni_vsubpd(const Xbyak::Xmm& v_dst,
```
a-sidorova, line 147: Somewhere you use `vDst`, somewhere - `v_dst`. Can we align the variable names? Thanks

src/plugins/intel_cpu/src/nodes/kernels/x64/jit_kernel_base.cpp
```diff
@@ -93,6 +97,19 @@ void JitKernelBase::uni_vpaddd(const Xbyak::Ymm& vDst,
     }
 }
 
+void JitKernelBase::uni_vpaddq(const Xbyak::Xmm& vDst,
+                               const Xbyak::Xmm& vSrc,
+                               const Xbyak::Operand& op) {
+    if (isValidIsa(x64::avx2)) {
```
a-sidorova, line 103: Seems like `avx` [supports](https://www.felixcloutier.com/x86/paddb:paddw:paddd:paddq) this instruction as well, doesn't it? Could you verify please?

src/plugins/intel_cpu/src/nodes/kernels/x64/random_uniform.cpp
```diff
@@ -0,0 +1,642 @@
...
+    v_n_inc        = getVmm();
+    v_range        = getVmm();
+    v_min          = getVmm();
+    v_key_64       = getVmm();
+    v_counter_64   = getVmm();
+    v_n_64         = getVmm();
+    v_res_perm     = getVmm();
+
+    // Initialize constants.
+    mov(r64_aux, 0xd2511f53);
```
a-sidorova, line 59: Could you please leave comments what is the constants?

a-sidorova: Moreover, I see you use these constants in different places. Maybe we can create static class fields with clear names with that are equal to these values?

src/plugins/intel_cpu/src/nodes/kernels/x64/random_uniform.cpp
```diff
@@ -0,0 +1,642 @@
...
+    vpbroadcastq(v_max_mul_n_64, r64_aux);
+
+    mov(r64_aux, 0xcd9e8d57);
+    vpbroadcastq(v_max_mul_c_64, r64_aux);
```
a-sidorova, line 63: Seems like you can create method `mov_scalar` (or smth else) that will be do these two lines. Because this code is repeated several times.

src/plugins/intel_cpu/src/nodes/random_uniform.cpp
```diff
@@ -0,0 +1,513 @@
...
+    }
+}
+
+void RandomUniform::initSupportedPrimitiveDescriptors() {
+    auto shape_prc = getOriginalInputPrecisionAtPort(SHAPE);
+    if (!one_of(shape_prc, InferenceEngine::Precision::I32, InferenceEngine::Precision::I64)) {
+        shape_prc = InferenceEngine::Precision::I32;
+    }
+
+    auto out_prc = getOriginalOutputPrecisionAtPort(0);
```
a-sidorova, line 82: As I can see kernel support bf16 only for `avx512_core_bf16` platforms. If it's so, should we add check here? Correct me please if I'm wrong

nshchego: If the kernel does not support any precision, it will not be created silently and execution will go through the CPU ref, which actually supports f16 and bf16.

src/plugins/intel_cpu/src/nodes/kernels/x64/jit_kernel_base.cpp
nshchego: This instruction works with Xmm vectors only on the AVX arch and requires implementation like [this](https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_cpu/src/nodes/kernels/x64/jit_kernel_base.cpp#L70). Taking into account the fact that we don't support AVX, it's better to avoid it here.

src/plugins/intel_cpu/src/nodes/kernels/x64/jit_kernel_base.cpp
nshchego: Done

src/plugins/intel_cpu/src/nodes/kernels/x64/random_uniform.cpp
nshchego: Done

src/plugins/intel_cpu/src/nodes/random_uniform.hpp
a-sidorova: As I can see, it's built successfully. But I think we should avoid these situations with x64 code compilation, shouldn't we?

@dmitry-gorokhov correct me please if I'm wrong.

src/plugins/intel_cpu/src/nodes/kernels/x64/random_uniform.hpp
```diff
@@ -0,0 +1,95 @@
...
+    const void* key_ptr;
+    const void* counter_ptr;
+    const void* n_ptr;
+    const void* min_ptr;
+    const void* max_ptr;
+    uint64_t work_amount = 0lu;
+};
+
+template <dnnl::impl::cpu::x64::cpu_isa_t isa>
+class RandomUniform : public JitKernel<RandomUniformCompileParams, RandomUniformCallArgs> {
```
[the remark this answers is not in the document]
nshchego: It's already in the `kernel` namespace, so such renaming is redundant.

src/plugins/intel_cpu/src/nodes/kernels/x64/jit_kernel_base.hpp
```diff
@@ -155,5 +183,71 @@ class JitKernelBase: public dnnl::impl::cpu::x64::jit_generator {
...
+
+        try {
+#define IF_ISA_CASE(ISA)                                \
+            if (dnnl::impl::cpu::x64::mayiuse(ISA))     \
+                res.reset(new KernelT<ISA>(jcp));       \
+            else
+
+                IF_ISA_CASE(dnnl::impl::cpu::x64::avx512_core)
+                IF_ISA_CASE(dnnl::impl::cpu::x64::avx2)
+                IF_ISA_CASE(dnnl::impl::cpu::x64::sse41);
```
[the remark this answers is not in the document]
nshchego: Done

src/plugins/intel_cpu/src/nodes/kernels/x64/random_uniform.hpp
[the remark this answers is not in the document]
a-sidorova: Do we really need it?

One of my main concerns in this PR: you add something new for several code parts but use only in new code and update nowhere:
- New macros for node exception: somewhere we don't have macros, somewhere we have personal macros for nodes (defined in the corresponding source files). Now we add `THROW_CPU_NODE_ERROR` macros that we use only in `RandomUniform`: we have 3 ways to throe exception. If new developer implement new op, he will have new question: how I should throw exception?
- Only kernel classes `GridSample` and `RandomUniform` are derived from `JitKernelBase`. But these classes are defined as `ov::intel_cpu::GridSampleKernel` and `ov::intel_cpu::kernel::RandomUniform` (other kernel classes are derived from `jit_generator` but have `kernel` in name). We have different implementation ways again. I'm not sure that it's OK situation as well. Moreover, there is another question: if we use namespace `kernel`, should we add namespace `x64`?

a-sidorova (approved):
Approved. But anyway I have two concerns:
- we add new namespace and exception macros but not update in other nodes and kernels. Let's do it in the follow-up PR to avoid other developer misunderstandings;
- included header for x64 and arm from folder `x64/kernel` that shouldn't be compiled on arm machines.

nshchego:
@dmitry-gorokhov, could you please take a look?

src/tests/functional/shared_test_classes/src/base/ov_subgraph.cpp
```diff
@@ -258,15 +258,15 @@ void SubgraphBaseTest::infer() {
 precisions_map SubgraphBaseTest::get_ref_precisions_convert_map() {
     //TODO: remove this conversions as soon as function interpreter fully support bf16 and f16
     precisions_map precisions = {
-            { ngraph::element::bf16, ngraph::element::f32 }
+            { ov::element::bf16, ov::element::f32 }
```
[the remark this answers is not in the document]
nshchego: Done

src/plugins/intel_cpu/src/nodes/random_uniform.cpp
```diff
@@ -0,0 +1,536 @@
...
+        } break;
+        case element::i32: {
+            generateData<int32_t, std::uniform_int_distribution<int32_t>>(
+                    std::uniform_int_distribution<int32_t>{m_min_val.i32, m_max_val.i32}, out, work_amount);
+        } break;
+        case element::i64: {
+            generateData<int64_t, std::uniform_int_distribution<int64_t>>(
+                    std::uniform_int_distribution<int64_t>{m_min_val.i64, m_max_val.i64}, out, work_amount);
+        } break;
+        case element::f64: {
```
dmitry-gorokhov, line 416: The plugin doesn't support f64 and there are no such plans. So lets remove f64 instances

src/plugins/intel_cpu/src/nodes/random_uniform.cpp
```diff
@@ -0,0 +1,536 @@
...
+        m_skip_count = m_out_el_num * SKIP_CONST;
+
+        if (m_out_el_num < PHILOX_PARALLEL_EXECUTION_THRESHOLD) {
+            m_threads_num = 1;
+        } else {
+            m_threads_num = parallel_get_max_threads();
+        }
+        m_thread_params.resize(m_threads_num);
+
+        parallel_nt(m_threads_num, [&](const int ithr, const int nthr) {
```
dmitry-gorokhov, line 149: Why we need to have it on prepareParams stage? The preparation step seems to be very simple and should significantly affect execution time - so can be united with execute call for simplification.

src/frontends/onnx/frontend/src/op/random_uniform.cpp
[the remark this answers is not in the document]
dmitry-gorokhov: Anyway there should be ticket number here.

src/plugins/intel_cpu/src/nodes/random_uniform.cpp
nshchego: We have a couple of "fast" models where the RandomUniform is a quite small.  One of them took less than 1 μs and it is sensitive to such work.

src/plugins/intel_cpu/src/shape_inference/custom/random_uniform.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include "shape_inference/shape_inference_cpu.hpp"
+#include <node.h>
+
+#pragma once
+
+namespace ov {
+namespace intel_cpu {
+namespace node {
+
+class RandomUniformShapeInfer : public ShapeInferEmptyPads {
```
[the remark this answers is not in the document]
nshchego: Core shape inference supports only constant inputs.

src/plugins/intel_cpu/src/nodes/random_uniform.cpp
nshchego: Done

src/plugins/intel_cpu/src/shape_inference/custom/random_uniform.hpp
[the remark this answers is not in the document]
dmitry-gorokhov: So it sounds valuable to extend core impl, instead of introduction custom one

src/plugins/intel_cpu/tests/functional/single_layer_tests/random_uniform.cpp
```diff
@@ -0,0 +1,381 @@
+// Copyright (C) 2018-2023 Intel Corporation
```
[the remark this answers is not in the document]
dmitry-gorokhov: Primitive type check doesn't seems to be very important here, so making this test shared is preferable (given common parts are not trivial). If prim type check is still required we can reuse common test via inheritance.
It also needs to be mentioned that we have [new structure](https://github.com/openvinotoolkit/openvino/tree/master/src/plugins/intel_cpu/tests#structure-of-the-platform-specific-tests) for CPU single layer tests, so all new tests should be aligned.

src/frontends/onnx/frontend/src/op/random_uniform.cpp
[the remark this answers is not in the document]
nshchego: Done

src/plugins/intel_cpu/src/shape_inference/custom/random_uniform.hpp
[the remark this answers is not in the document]
nshchego: I've created the ticket for that.

src/plugins/intel_cpu/tests/functional/single_layer_tests/random_uniform.cpp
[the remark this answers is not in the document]
nshchego: I've divided tests as per new layout.
About the Shared tests. There is already implementation for RandomUniform, but it is so restricted. Its modification may lead to conflicts with other plugins.

dmitry-gorokhov (approved):

PR #20185, opened 2023-10-02
[apiConformance] Remove 1.0 api tests from API Conformance

sbalandi:
### Details:
 - *Task for restoring missing tests CVS-121889*
 
### Tickets:
 - *CVS-109784*


iefode:
@sbalandi We need to remove these tests from shared lib also after the plugins will migrate the instances to api2.0.

Just FYI @ilyachur @ilya-lavrenov

iefode (approved):

sbalandi:
> @sbalandi We need to remove these tests from shared lib also after the plugins will migrate the instances to api2.0.

@iefode yes, task for tracking that https://jira.devtools.intel.com/browse/CVS-110598

PR #20218, opened 2023-10-03
adding python pot quantization and sparsity sample notebook

ronandorrepaal:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


MaximProshin:
@ilya-lavrenov , it was an OV notebook initially but we agreed to move it to POT samples instead as POT is deprecated now and we don't want to show it in notebooks (https://github.com/openvinotoolkit/openvino_notebooks/pull/1269). 

@ronandorrepaal , please reopen your PR to POT samples (https://github.com/openvinotoolkit/openvino/tree/master/tools/pot/openvino/tools/pot/api/samples) and assign it to @AlexKoff88 .

PR #20223, opened 2023-10-04
[GPU] fix issue of optimized out reorder during event sync

wilson-seok:
### Details:
 - fix issue of runtime reorder optimized out during event sync in dGPU

### Tickets:
 - 100495

### DIFF
--- a/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
+++ b/src/plugins/intel_gpu/src/graph/primitive_inst.cpp
@@ -685,6 +685,12 @@ void primitive_inst::do_runtime_skip_reorder() {
             GPU_DEBUG_TRACE_DETAIL << "[do runtime skip reorder] update shape for user " << u->id() << std::endl;
             u->update_shape();
             u->update_shape_done_by_other = true;
+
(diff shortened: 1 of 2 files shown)


src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -685,6 +685,12 @@ void primitive_inst::do_runtime_skip_reorder() {
             GPU_DEBUG_TRACE_DETAIL << "[do runtime skip reorder] update shape for user " << u->id() << std::endl;
             u->update_shape();
             u->update_shape_done_by_other = true;
+
+            // Opt out reorder which has _needs_completion_event = true causes syncronization failed in dGPU.
+            if (_needs_completion_event == false && u->_needs_completion_event == true) {
+                _needs_completion_event = true;
+            }
```
yeonbok, line 692: I think this should be applied only when user is actually optimized (at line 707)

wilson-seok: Updated it. Thanks!

yeonbok (approved):

PR #20240, opened 2023-10-04
support aten::channel_shuffle

eaidova:
### Details:
 - *support aten::channel_shuffle operation conversion*

enables fbnet models from [mobile-vision](https://github.com/facebookresearch/mobile-vision)

### Tickets:
 - *CVS-110686*


src/frontends/pytorch/src/op/pixel_shuffle.cpp
```diff
@@ -67,6 +69,38 @@ OutputVector translate_pixel_shuffle(const NodeContext& context) {
...
+    Output<Node> shape;
+    Output<Node> rank;
+    std::tie(shape, rank) = get_shape_rank(context, x, true);
+    // PyTorch realization uses assumption that channels dim is always 1
+    auto indices = context.mark_node(v0::Constant::create(element::i32, Shape{2}, {0, 1}));
+    auto dims = context.mark_node(std::make_shared<v8::Gather>(shape, indices, zero));
+    auto dims_splitted = context.mark_node(std::make_shared<v1::Split>(dims, zero, 2));
+    auto c = dims_splitted->output(1);
+    auto n = dims_splitted->output(0);
+    auto rank_1d = context.mark_node(std::make_shared<v0::Unsqueeze>(rank, zero));
```
mvafin, line 89: Please use `get_shape_rank` helper with `false` to have 1d rank value

eaidova: current realization does not use rank at all, so removed this line

mvafin (approved):

PR #20242, opened 2023-10-04
[Snippets] Deep copy interface

IvanNovoselov:
### Details:
 - *During review of PR18563 it was established that it is hard to perform correct deep copy of LIR, because it owns a range of pointers that in turn needs to be deep copied. In this PR we introduce a convenient and safe way to copy LIR or deep_copy a range of expressions*

### Tickets:
 - *-*
 
### Prerequisites:
- [x] Snippets dynamic pipeline reorganization [18563](https://github.com/openvinotoolkit/openvino/pull/18563)

PR between branches for review: https://github.com/IvanNovoselov/openvino/pull/14

### DIFF
--- a/src/common/snippets/include/snippets/lowered/expression.hpp
+++ b/src/common/snippets/include/snippets/lowered/expression.hpp
@@ -26,7 +26,6 @@ class Expression : public std::enable_shared_from_this<Expression> {
 
 public:
     Expression() = default;
-    Expression(const Expression& other);
     virtual ~Expression() = default;
 
     std::shared_ptr<Node> get_node() const;
@@ -62,12 +61,11 @@ public:
     ExpressionPtr clone_with_new_inputs(const ExressionMap& expr_map, const std::shared_ptr<Node>& new_node) const;
 
 protected:
+    Expression(const Expression& other);
     // Note: The constructor initialization is private since an expression can be created only by Linear IR.
     //       The method must be used only by Linear IR builder of expressions!
     Expression(const std::shared_ptr<Node>& n, const std::shared_ptr<IShapeInferSnippetsFactory>& factory);
-    static void update_node_and_connectors(const ExpressionPtr& expr,
-                                           const std::vector<PortConnectorPtr>& new_inputs,
-                                           const std::shared_ptr<Node>& new_node);
+    void update_node_and_connectors(const std::vector<PortConnectorPtr>& new_inputs, const std::shared_ptr<Node>& new_node);
 
     void replace_input(size_t port, PortConnectorPtr to);
 
@@ -88,7 +86,6 @@ class IOExpression : public Expression {
 
 public:
     enum class io_type {INPUT, OUTPUT, UNDEFINED};
-    IOExpression(const IOExpression& other) = default;
     ExpressionPtr clone_with_new_inputs(const std::vector<PortConnectorPtr>& new_inputs,
                                         const std::shared_ptr<Node>& new_node) const override;
     int64_t get_index() const  { return m_index; }
@@ -96,6 +93,7 @@ public:
     // Result needs shapeInfer to copy shape from Parent's output to this expr input
     bool needShapeInfer() const override {return m_type == io_type::OUTPUT; }
 private:
+    IOExpression(const IOExpression& other) = default;
     explicit IOExpression(const std::shared_ptr<ov::opset1::Parameter>& n, int64_t index, const std::shared_ptr<IShapeInferSnippetsFactory>& factory);
     explicit IOExpression(const std::shared_ptr<ov::opset1::Result>& n, int64_t index, const std::shared_ptr<IShapeInferSnippetsFactory>& factory);
 
--- a/src/common/snippets/include/snippets/lowered/expression.hpp
+++ b/src/common/snippets/include/snippets/lowered/expression.hpp
@@ -19,12 +19,14 @@ namespace lowered {
 
 class LinearIR;
 using ExpressionPtr = std::shared_ptr<Expression>;
+using ExressionMap = std::unordered_map<Expression*, ExpressionPtr>;
 class Expression : public std::enable_shared_from_this<Expression> {
     friend class LinearIR;
     friend class ExpressionPort;
 
 public:
     Expression() = default;
+    Expression(const Expression& other);
     virtual ~Expression() = default;
 
     std::shared_ptr<Node> get_node() const;
@@ -55,11 +57,17 @@ public:
 
     std::vector<size_t> get_loop_ids() const;
     void set_loop_ids(const std::vector<size_t>& loops);
+    virtual ExpressionPtr clone_with_new_inputs(const std::vector<PortConnectorPtr>& new_inputs,
+                                                const std::shared_ptr<Node>& new_node) const;
+    ExpressionPtr clone_with_new_inputs(const ExressionMap& expr_map, const std::shared_ptr<Node>& new_node) const;
 
 protected:
     // Note: The constructor initialization is private since an expression can be created only by Linear IR.
     //       The method must be used only by Linear IR builder of expressions!
     Expression(const std::shared_ptr<Node>& n, const std::shared_ptr<IShapeInferSnippetsFactory>& factory);
+    static void update_node_and_connectors(const ExpressionPtr& expr,
+                                           const std::vector<PortConnectorPtr>& new_inputs,
+                                           const std::shared_ptr<Node>& new_node);
 
     void replace_input(size_t port, PortConnectorPtr to);
 
@@ -80,7 +88,9 @@ class IOExpression : public Expression {
 
 public:
     enum class io_type {INPUT, OUTPUT, UNDEFINED};
-
+    IOExpression(const IOExpression& other) = default;
+    ExpressionPtr clone_with_new_inputs(const std::vector<PortConnectorPtr>& new_inputs,
+                                        const std::shared_ptr<Node>& new_node) const override;
     int64_t get_index() const  { return m_index; }
     io_type get_type() const { return m_type; }
     // Result needs shapeInfer to copy shape from Parent's output to this expr input
--- a/src/common/snippets/include/snippets/lowered/expression_port.hpp
+++ b/src/common/snippets/include/snippets/lowered/expression_port.hpp
@@ -25,6 +25,11 @@ public:
 
     ExpressionPort() = default;
     explicit ExpressionPort(const std::shared_ptr<Expression>& expr, Type type, size_t port);
+    /**
+    * @interface clone
+    * @brief Creates similar Expression port, but for new expression
+     */
+    std::shared_ptr<ExpressionPort> clone_for_new_expr(const std::shared_ptr<Expression>& new_expr) const;
 
     std::shared_ptr<Expression> get_expr() const;
     Type get_type() const { return m_type; }
--- a/src/common/snippets/include/snippets/lowered/expression_port.hpp
+++ b/src/common/snippets/include/snippets/lowered/expression_port.hpp
@@ -26,10 +26,10 @@ public:
     ExpressionPort() = default;
     explicit ExpressionPort(const std::shared_ptr<Expression>& expr, Type type, size_t port);
     /**
-    * @interface clone
+    * @interface clone_with_new_expr
     * @brief Creates similar Expression port, but for new expression
      */
-    std::shared_ptr<ExpressionPort> clone_for_new_expr(const std::shared_ptr<Expression>& new_expr) const;
+    std::shared_ptr<ExpressionPort> clone_with_new_expr(const std::shared_ptr<Expression>& new_expr) const;
 
     std::shared_ptr<Expression> get_expr() const;
     Type get_type() const { return m_type; }
--- a/src/common/snippets/include/snippets/lowered/linear_ir.hpp
+++ b/src/common/snippets/include/snippets/lowered/linear_ir.hpp
@@ -49,7 +49,10 @@ public:
 
     ExpressionPtr create_expression(const std::shared_ptr<Node>& n, const std::vector<PortConnectorPtr>& inputs);
 
-    static LinearIR::container deep_copy_range(LinearIR::container::const_iterator begin, LinearIR::container::const_iterator end);
+    std::shared_ptr<LinearIR> clone() const;
+    static LinearIR::container deep_copy_range(LinearIR::container::const_iterator begin,
+                                               LinearIR::container::const_iterator end,
+                                               ExressionMap& expression_map);
 
     const container& get_ops() const {return m_expressions; }
     const io_container& get_IO_ops() const {return m_io_expressions; }
@@ -116,7 +119,6 @@ public:
     IShapeInferSnippets::Result shape_infer(const std::vector<VectorDimsRef>& input_shapes);
     const std::shared_ptr<ShapeInferSnippetsNode>& get_shape_infer_instance() const {return m_shape_infer; }
     VectorDims get_master_shape() const;
-    LinearIR deep_copy() const;
 
 private:
     std::shared_ptr<ShapeInferSnippetsNode> m_shape_infer = nullptr;
--- a/src/common/snippets/src/lowered/expression.cpp
+++ b/src/common/snippets/src/lowered/expression.cpp
@@ -27,6 +27,22 @@ Expression::Expression(const std::shared_ptr<Node>& n, const std::shared_ptr<ISh
     }
 }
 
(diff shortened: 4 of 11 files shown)


IvanNovoselov:
@a-sidorova, @v-Golubev could you please start the review?

src/common/snippets/src/lowered/loop_manager.cpp
```diff
@@ -15,8 +15,17 @@
 namespace ov {
 namespace snippets {
 namespace lowered {
+using LoopManager = LinearIR::LoopManager;
+using LoopPort = LoopManager::LoopPort;
+using LoopInfo = LoopManager::LoopInfo;
+
+std::shared_ptr<LoopPort> LoopPort::clone_for_new_expr(const ExpressionPtr& new_expr) const {
```
a-sidorova, line 22: Now we can clone `LoopPort` 🎆  May I ask you to update the WA [there](https://github.com/openvinotoolkit/openvino/blob/3403e6c02873bb0fa9f0cf410d24bbb79991e93c/src/common/snippets/src/lowered/loop_manager.cpp#L323) please?

src/common/snippets/include/snippets/lowered/linear_ir.hpp
```diff
@@ -49,7 +49,10 @@ class LinearIR {
 
     ExpressionPtr create_expression(const std::shared_ptr<Node>& n, const std::vector<PortConnectorPtr>& inputs);
 
-    static LinearIR::container deep_copy_range(LinearIR::container::const_iterator begin, LinearIR::container::const_iterator end);
+    std::shared_ptr<LinearIR> clone() const;
```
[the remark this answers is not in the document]
a-sidorova: And for other classes. Because sometimes we can work with only instances. I'd like to avoid situations aka `auto new_smth = *{smth->clone()}` like [there](https://github.com/openvinotoolkit/openvino/blob/f62fd778a5c51ef426138e19f70aa59f1800709d/src/common/snippets/src/op/subgraph.cpp#L494).

src/common/snippets/src/lowered/expression.cpp
```diff
@@ -103,6 +119,54 @@ void Expression::set_loop_ids(const std::vector<size_t>& loops) {
     m_loop_ids = loops;
 }
 
+void Expression::update_node_and_connectors(const ExpressionPtr& expr,
+                                            const std::vector<PortConnectorPtr>& new_inputs,
+                                            const std::shared_ptr<Node>& new_node) {
+    if (new_node) {
```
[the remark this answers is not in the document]
a-sidorova: Seems like in `InsertTailLoop`...
If we want to have valid `LinearIR` till code emission, let's make really deep copy for tail loops with node copying. What do you think?

src/common/snippets/include/snippets/lowered/expression_port.hpp
```diff
@@ -25,6 +25,11 @@ class ExpressionPort {
 
     ExpressionPort() = default;
     explicit ExpressionPort(const std::shared_ptr<Expression>& expr, Type type, size_t port);
+    /**
+    * @interface clone
+    * @brief Creates similar Expression port, but for new expression
+     */
+    std::shared_ptr<ExpressionPort> clone_for_new_expr(const std::shared_ptr<Expression>& new_expr) const;
```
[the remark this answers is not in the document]
IvanNovoselov: Good idea, renamed.

src/common/snippets/include/snippets/lowered/expression.hpp
```diff
@@ -19,12 +19,14 @@ namespace lowered {
...
 class LinearIR;
 using ExpressionPtr = std::shared_ptr<Expression>;
+using ExressionMap = std::unordered_map<Expression*, ExpressionPtr>;
 class Expression : public std::enable_shared_from_this<Expression> {
     friend class LinearIR;
     friend class ExpressionPort;
 
 public:
     Expression() = default;
+    Expression(const Expression& other);
```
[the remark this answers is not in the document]
IvanNovoselov: We can, but note that it will make `Expressions` non-copyable outside of `Expression::` or friendly classes. This is not a bad thing, since copying generally produces invalid expression (because internal pointers are not updated), but we should keep it in mind.

src/common/snippets/src/lowered/expression.cpp
```diff
@@ -103,6 +119,54 @@ void Expression::set_loop_ids(const std::vector<size_t>& loops) {
     m_loop_ids = loops;
 }
 
+void Expression::update_node_and_connectors(const ExpressionPtr& expr,
```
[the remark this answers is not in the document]
IvanNovoselov: Yes, it looks better, let's do it

src/common/snippets/src/lowered/expression.cpp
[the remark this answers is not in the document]
IvanNovoselov: The idea here was to allow updates of expression's connections without deep-coping of nGraph IR. For example, if we need to move Expression inside LIR and to change its connectors.
On the other hand, we can always pass `expr->get_node()` to this method, so it's Ok to remove the `if` I guess.

src/common/snippets/src/lowered/expression.cpp
```diff
@@ -103,6 +119,54 @@ void Expression::set_loop_ids(const std::vector<size_t>& loops) {
...
+    OPENVINO_ASSERT(new_inputs.size() == expr->m_input_port_descriptors.size(),
+                    "Can't create Expression with new inputs: invalid number of input port connectors passed");
+    expr->m_input_port_connectors = new_inputs;
+    for (size_t i = 0; i < expr->m_input_port_descriptors.size(); i++) {
+        const auto& i_con = new_inputs[i];
+        const auto& i_port = expr->get_input_port(i);
+        if (!i_con->found_consumer(i_port))
+            i_con->add_consumer(i_port);
+    }
+    auto& new_outputs = expr->m_output_port_connectors;
```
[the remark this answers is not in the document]
IvanNovoselov: Theoretically yes, it would be safer, so lets do it
BTW I thought that you're going to suggest `std::transform(...)` 😆

src/common/snippets/include/snippets/lowered/linear_ir.hpp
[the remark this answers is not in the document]
IvanNovoselov: Yes, I see your point, but LIR is a different story. The thing with LIR is that it's handled by value and passed to transformations by reference almost everywhere in the pipeline. So when we write `auto a = m_linear_ir`, we actually expect a shallow copy, not a deep one. For example here, in [op::Kernel](https://github.com/openvinotoolkit/openvino/blob/9fafcabb7c7b221821b0b77616d9967c8d13223f/src/common/snippets/include/snippets/op/kernel.hpp#L23-L26), we don't need to deep copy of the whole IR (which is really expensive). 
In fact, we need to deep copy LIR only once, and that's the place you linked. And it looks a bit unnatural just because `LinearIR::clone()` is aligned with the rest of the `clone` methods and returns `shared_ptr` of the object, while LIR (despite the rest of the classes) is managed by value, not by pointer.

src/common/snippets/src/lowered/linear_ir.cpp
```diff
@@ -44,6 +44,35 @@ LinearIR::LinearIR(const std::shared_ptr<ov::Model>& model, const std::shared_pt
...
+    cloned->m_config = m_config;
+    // Clone underlying ngraph nodes
+    NodeVector original_nodes;
+    original_nodes.reserve(m_expressions.size());
+    for (const auto& expr : m_expressions)
+        original_nodes.push_back(expr->get_node());
+    // node_map and expr_map map original node pointer (expression) to a new pointer (expression)
+    ngraph::NodeMap node_map;
+    OPENVINO_SUPPRESS_DEPRECATED_START
+    ngraph::clone_nodes(original_nodes,  node_map);
```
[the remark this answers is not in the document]
IvanNovoselov: 👍

src/common/snippets/src/lowered/loop_manager.cpp
IvanNovoselov: I tried, but please double-check. I'm not 100% confident that I got it right 😅

src/common/snippets/src/lowered/expression.cpp
```diff
@@ -103,6 +119,54 @@ void Expression::set_loop_ids(const std::vector<size_t>& loops) {
...
+        expr->m_source_node = new_node;
+    }
+    OPENVINO_ASSERT(new_inputs.size() == expr->m_input_port_descriptors.size(),
+                    "Can't create Expression with new inputs: invalid number of input port connectors passed");
+    expr->m_input_port_connectors = new_inputs;
+    for (size_t i = 0; i < expr->m_input_port_descriptors.size(); i++) {
+        const auto& i_con = new_inputs[i];
+        const auto& i_port = expr->get_input_port(i);
+        if (!i_con->found_consumer(i_port))
+            i_con->add_consumer(i_port);
```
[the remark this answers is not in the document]
IvanNovoselov: As discussed offline, it's a good idea, but should be implemented in a separate PR. I've created a dedicated ticket **122690**

src/common/snippets/include/snippets/lowered/linear_ir.hpp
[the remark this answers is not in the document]
a-sidorova: I got it, good explained! Thanks a lot!

src/common/snippets/src/lowered/loop_manager.cpp
a-sidorova: You did exactly what I meant. Thanks!

a-sidorova (approved):

src/common/snippets/src/lowered/expression_factory.cpp
```diff
@@ -91,7 +91,7 @@ ExpressionPtr LinearIR::ExpressionFactory::create(const std::shared_ptr<op::Loop
                                                   const std::vector<PortConnectorPtr>& inputs,
                                                   const LinearIR& linear_ir) {
     OPENVINO_ASSERT(inputs.empty(), "LoopBegin cannot have inputs");
-    auto expr = std::make_shared<Expression>(Expression(n, linear_ir.m_shape_infer_factory));
+    auto expr = std::shared_ptr<Expression>(new Expression(n, linear_ir.m_shape_infer_factory));
```
v-Golubev, line 94: What is the reason for this change? We can't use `make_shared` for expressions anymore?

src/common/snippets/src/lowered/linear_ir.cpp
```diff
@@ -44,6 +44,35 @@ LinearIR::LinearIR(const std::shared_ptr<ov::Model>& model, const std::shared_pt
...
+    // Clone underlying ngraph nodes
+    NodeVector original_nodes;
+    original_nodes.reserve(m_expressions.size());
+    for (const auto& expr : m_expressions)
+        original_nodes.push_back(expr->get_node());
+    // node_map and expr_map map original node pointer (expression) to a new pointer (expression)
+    ngraph::NodeMap node_map;
+    OPENVINO_SUPPRESS_DEPRECATED_START
+    ngraph::clone_nodes(original_nodes, node_map);
+    OPENVINO_SUPPRESS_DEPRECATED_END
```
v-Golubev, line 59: It looks like I missed something: why do we need to clone `ov::Node` here if we clone it inside `deep_copy_range` too?

src/common/snippets/src/lowered/loop_manager.cpp
v-Golubev: @IvanNovoselov  @a-sidorova Actually we can't just use `clone_for_new_expr` instead of old logic because 
new expression ports in `update_loop_port` might have different `port_idx` besides expression. Let's look at a specific example. E.g. we use `update_loops_port` when we insert Load operation on 1st Add's input: in this case new ExpressionPort has idx=0 whereas Add's ExpressionPort has idx=1.

So we can return old logic or introduce a new API for LoopPort: `clone_with_new_expr_port`

src/common/snippets/src/lowered/loop_manager.cpp
```diff
@@ -319,13 +357,11 @@ void LinearIR::LoopManager::update_loop_port(size_t loop_id, const ExpressionPor
...
-                   [](LoopPort loop_port, const ExpressionPort& expr_port) {
-                       LoopPort copy = std::move(loop_port);  // to save loop port parameters
-                       copy.expr_port = std::make_shared<ExpressionPort>(expr_port);
-                       return copy;
-                   });
+    std::vector<LoopPort> target_loop_ports;
+    target_loop_ports.reserve(target_ports.size());
+    for (const auto& p : target_ports)
+        target_loop_ports.push_back(*port_it->clone_with_new_expr(p.get_expr()));
+
```
v-Golubev, line 364: Please note that we need to change not only expression but port_idx too. See more detailed explanations in the [discussion](https://github.com/openvinotoolkit/openvino/pull/20242#discussion_r1354963045)

src/common/snippets/src/lowered/loop_manager.cpp
a-sidorova: @v-Golubev Oh, I see your point. Great catch! Thanks a lot!

@IvanNovoselov I'm sorry that I confused you. Could you please revert this change? It's invalid since we call `clone_for_new_expr` with new expr that isn't the same as previous expr (they may have different input count, output count). I think that in this case we should add `clone_with_new_expr_port`. You may just revert this change and don't introduce new API with `clone_with_new_expr_port`. It's just small suggestion.

src/common/snippets/src/lowered/expression_factory.cpp
IvanNovoselov: The reason is that we implemented a [protected copy constructor](https://github.com/IvanNovoselov/openvino/blob/1e43b55755396b62f3227b9a33860f2a787fdefa/src/common/snippets/include/snippets/lowered/expression.hpp#L64) instead of a default public one.

src/common/snippets/src/lowered/linear_ir.cpp
```diff
@@ -44,6 +44,35 @@ LinearIR::LinearIR(const std::shared_ptr<ov::Model>& model, const std::shared_pt
     m_shape_infer = std::make_shared<LIRShapeInfer>(m_expressions, m_io_expressions);
 }
 
+std::shared_ptr<LinearIR> LinearIR::clone() const {
+    auto cloned = std::make_shared<LinearIR>();
+    cloned->m_config = m_config;
+    // Clone underlying ngraph nodes
```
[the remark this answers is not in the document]
IvanNovoselov: Even from comments? Wow, that's harsh 🙂 
You can delete it from my comments, but you would never delete it from my heart 🤣

src/common/snippets/src/lowered/linear_ir.cpp
IvanNovoselov: No-no, you're right it was there well before a proper deep_copy_range was implemented, but we don't need it anymore. Thanks a lot!

src/common/snippets/src/lowered/loop_manager.cpp
IvanNovoselov: All right, I reverted the changes

src/common/snippets/src/lowered/loop_manager.cpp
IvanNovoselov: Agreed. Good job!

src/common/snippets/src/lowered/expression_factory.cpp
v-Golubev: Thanks for the explanation

v-Golubev:
BTW do we really have dependency from https://github.com/openvinotoolkit/openvino/pull/18563? When I merged "Deep copy interface" changes with my branch (which do not contain #18563 changes), I did not meet any difficulties. If we did not have the dependency, we would probably merge this PR faster since it's quite small in comparison with #18563

IvanNovoselov:
> BTW do we really have dependency from #18563? When I merged "Deep copy interface" changes with my branch (which do not contain #18563 changes), I did not meet any difficulties. If we did not have the dependency, we would probably merge this PR faster since it's quite small in comparison with #18563

In principle, we can merge it before #18563, since this PR changes are quite isolated. But I would not suggest it because I expect that rebasing #18563 over this PR would be error-prone and could be painful as well.

IvanNovoselov:
@a-sidorova, @v-Golubev thanks a lot for your comments!
@dmitry-gorokhov could you take a look please?

dmitry-gorokhov (approved):

PR #20250, opened 2023-10-05
[GPU] add shape infer in scatter elements update

wilson-seok:
### Details:
 - add shape infer in scatter elements update

### Tickets:
 - 100201


src/plugins/intel_gpu/src/graph/scatter_elements_update.cpp
```diff
@@ -34,6 +36,32 @@ layout scatter_elements_update_inst::calc_output_layout(scatter_elements_update_
...
+        input2_layout.get<ShapeType>(),     // updates_shape,
+        input3_layout.get<ShapeType>(),     // axis_shape,
+    };
+
+    std::vector<ShapeType> output_shapes = {ShapeType()};
+
+    ov::op::v3::ScatterElementsUpdate op;
+    output_shapes = shape_infer(&op, input_shapes);
+
+    return { layout{output_shapes[0], input0_layout.data_type, input0_layout.format} };
```
yeonbok, line 59: I think scatter update 's output shape is always same as original input, isn't it?

yeonbok: So I think we just can do it as static shape just replacing the access to the "tensor()" to get_partial_shape().

ahnyoung-paul (approved):

yeonbok (approved):

PR #20256, opened 2023-10-05
Fix SubgraphsDumper compilation Win

iefode:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/tests/functional/plugin/conformance/subgraphs_dumper/src/utils/model.cpp
+++ b/src/tests/functional/plugin/conformance/subgraphs_dumper/src/utils/model.cpp
@@ -52,7 +52,8 @@ find_models(const std::vector<std::string> &dirs, const std::string& regexp) {
                 } else {
                     continue;
                 }
-            } catch (std::exception& e) {
+            } catch (...) {
+            // } catch (std::exception& e) {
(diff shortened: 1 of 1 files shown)


src/tests/functional/plugin/conformance/subgraphs_dumper/src/utils/model.cpp
```diff
@@ -52,7 +52,8 @@ find_models(const std::vector<std::string> &dirs, const std::string& regexp) {
                 } else {
                     continue;
                 }
-            } catch (std::exception& e) {
+            } catch (...) {
```
praasz, line 55:
```suggestion
            } catch (std::exception&) {
```
Will it it work also?

src/tests/functional/plugin/conformance/subgraphs_dumper/src/utils/model.cpp
```diff
@@ -52,7 +52,8 @@ find_models(const std::vector<std::string> &dirs, const std::string& regexp) {
...
+            // } catch (std::exception& e) {
```
praasz, line 56: Remove the commented code ?

praasz (approved):

PR #20257, opened 2023-10-05
Moved apply auto padding to dev API

ilyachur:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


praasz (approved):

src/core/dev_api/validation_util.hpp
```diff
@@ -47,5 +54,29 @@ OPENVINO_API std::shared_ptr<op::v0::Constant> constantfold_subgraph(const Outpu
...
+/// \param filter_dilations  The dilations of applied padding.
+/// \param pad_type          The type of padding. Auto padding is applied only
+///                          for SAME_UPPER and SAME_LOWER mode.
+/// \param padding_above     The beginning of padding shape.
+/// \param end               The beginning of padding shape.
+///
+/// \return true if auto padding was applied successfully (all needed informations such as
+///         spatial dims are known), false otherwise.
+OPENVINO_API
+bool try_apply_auto_padding(const PartialShape& image_shape,
```
praasz, line 73: Also consider to change deprecated to be removed in 2024.
The reason is that this function is used in plugins (GNA, GPU) in pooling, convolution ops but current version of shape_infer handle this padding internally.

Also some front end use functions which relay on this auto padding but I think FE could also be reworked.

Also functions like:
- infer_convolution_forward
- infer_batched_pooling_forward
- infer_batch_norm_forward_helper
- validate_conv_params_spatial_dimensions
- infer_windowed_reduction_output_shape
- infer_slice_shape
etc.

For me are candidates for removal,
 They are big and  most of them are implemented in shape_infer, Plugins could be reworked to re-use shape_infer and maybe FE will not require it all.

The binary size reduction also will be visible.

The function `infer_batch_norm_forward` is used as shape inference in operator but can be migrated also to shape_inference library.

PR #20258, opened 2023-10-05
[core]Migrate Eye to new API

praasz:
### Details:
 - Migrate Eye operator to new API. Remove legacy headers, namespace, functions, classes.
 - Refactor Eye for binary size reduction.

### Tickets:
 - [CVS-118568](https://jira.devtools.intel.com/browse/CVS-118568)


t-jankowski (approved):

src/core/include/openvino/op/util/evaluate_helpers.hpp
```diff
@@ -0,0 +1,23 @@
...
+namespace op {
+namespace util {
+
+/**
+ * @brief Get the tensors shapes as ov::PartialSHape
+ *
+ * @param tensors  Input tensors vector to get its shapes.
+ * @return Vector of partial shapes sam size as input tensor vector.
+ */
+std::vector<PartialShape> get_tensors_shapes(const TensorVector& tensors);
```
ilyachur, line 20: Why do we need this function? Is it a common case for operations?
Moreover `ov::Tensor` doesn't contain `ov::PartialShape` and has only `ov::Shape`.

praasz: It is common when `shape_infer` used for output shape calculations. It helps get shapes as partial shape instead manual rewrite tensor shapes into vector and also in case when op has variadic inputs count it collect all of them in same way.

Will rename it to  `get_tensors_partial_shapes` to explicitly give information what is the return value. When accepted in one PR and merged then propagate this change for others.

Function `shape_infer` not working on `ov::Shape` as this shape has not compatible interface with `ov::PartialShape`.

src/core/include/openvino/op/util/evaluate_helpers.hpp
```diff
@@ -0,0 +1,23 @@
...
+namespace op {
+namespace util {
+
+/**
+ * @brief Get the tensors shapes as ov::PartialShape.
+ *
+ * @param tensors  Input tensors vector to get its shapes.
+ * @return Vector of partial shapes sam size as input tensor vector.
+ */
+std::vector<PartialShape> get_tensors_partial_shapes(const TensorVector& tensors);
```
ilyachur, line 20: I think it should be a part of dev API. 
Moreover you need to export this method if you want to have it as a part of public API.

ilyachur: We need to export public API from the library. Because in other case users won't be able to use these methods. API should be consistant.

PR #20265, opened 2023-10-05
[Opset13][pyAPI] Python API BitwiseNot-13

mmikolajcz:
### Details:
 - *Add pyAPI for BitwiseNot-13*
 - *...*

### Tickets:
 - *118959*


src/bindings/python/src/openvino/runtime/opset13/ops.py
```diff
@@ -19,6 +19,23 @@
...
+    node: NodeInput,
+    name: Optional[str] = None,
+) -> Node:
+    """Return node which performs bitwise NOT operation on input node element-wise.
+
+    For boolean input tensors, operator is equivalent to logical_not.
+
+    :param node: Tensor of integer or boolean datatype providing data.
+    :param name: The optional new name for output node.
+    :return: The node performing bitwise NOT operation with given tensor.
```
mitruska, line 33: Consider:
```suggestion
    :return: The node performing bitwise NOT operation on the given tensor.
```

PR #20270, opened 2023-10-05
[TF FE] Provide full support of TF1 Control flow and TensorArray ops

rkazants:
**Details:** it completes the feature proving support TF1 Control flow for While and If operations with TensorArray operations.
Since this moment, there is no longer need to specify JSON config for conversion of TF1 OD models and experience problems with conversion of models with TF1 control flow. All should work OOB.

**Ticket:** 108581

### DIFF
--- a/tools/mo/unit_tests/moc_tf_fe/conversion_basic_models_test.py
+++ b/tools/mo/unit_tests/moc_tf_fe/conversion_basic_models_test.py
@@ -241,11 +241,8 @@ class TestMoFreezePlaceholderTFFE(unittest.TestCase):
                    None, None, True, True, False, False)
 
     def test_conversion_failure_fallback_use_new_frontend(self):
-        with self.assertRaisesRegex(Exception,
-                                    "\[TensorFlow Frontend\] Internal error, no translator found for operation\(s\)\: "
-                                    "TensorArrayGatherV3\, TensorArrayReadV3\, TensorArraySizeV3\, TensorArrayV3\, TensorArrayWriteV3"):
-            self.basic("ctc_model_based.pbtxt", None, None, None, None,
-                       None, None, True, True, True, False)
+        self.basic("ctc_model_based.pbtxt", None, None, None, None,
+                   None, None, True, True, True, False)
 
     @unittest.skip("88349: Fix auto-pruning in legacy FE")
     def test_conversion_model_oneshot_iterator_use_legacy_frontend(self):
--- a/src/frontends/tensorflow/src/op/tensor_array_operations.cpp
+++ b/src/frontends/tensorflow/src/op/tensor_array_operations.cpp
@@ -1,0 +1,363 @@
+// Copyright (C) 2018-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "common_op_table.hpp"
+#include "helper_ops/enter.hpp"
+#include "helper_ops/tensor_array.hpp"
+#include "openvino/frontend/tensorflow/node_context.hpp"
+#include "openvino/op/add.hpp"
+#include "openvino/op/broadcast.hpp"
+#include "openvino/op/concat.hpp"
+#include "openvino/op/constant.hpp"
+#include "openvino/op/convert.hpp"
+#include "openvino/op/gather.hpp"
+#include "openvino/op/maximum.hpp"
+#include "openvino/op/multiply.hpp"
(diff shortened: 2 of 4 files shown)


src/common/transformations/src/transformations/common_optimizations/moc_transformations.cpp
```diff
@@ -160,7 +160,7 @@ bool ov::pass::MOCTransformations::run_on_model(const std::shared_ptr<ov::Model>
     REGISTER_PASS(manager, PullThroughReduce)
 
     // GRUCellFusion and SequenceFusion should be before NopElimination
-    REGISTER_PASS(manager, GRUCellFusion)
+    //REGISTER_PASS(manager, GRUCellFusion)
```
rkazants, line 163:
```suggestion
    // REGISTER_PASS(manager, GRUCellFusion)
```

tools/mo/unit_tests/moc_tf_fe/conversion_basic_models_test.py
```diff
@@ -241,11 +241,8 @@ def test_conversion_failure_fallback_default(self):
 
     @unittest.skipIf(platform == 'darwin', reason="Ticket - 122182")
     def test_conversion_failure_fallback_use_new_frontend(self):
```
popovaan, line 243: Can we remove unittest.skipIf here? 
And may be rename the test, as it does not fail anymore.

src/common/transformations/src/transformations/common_optimizations/gru_cell_fusion.cpp
```diff
@@ -148,6 +148,15 @@ ov::pass::GRUCellFusion::GRUCellFusion() {
             Bh = rg.make<ov::op::v0::Constant>(WRh.get_element_type(), Shape{1, static_cast<size_t>(hidden_size)}, 0);
         }
 
+        // perform additional check for applicability of the transformation
+        // without this check, process_weights can fail
+        if (WR.get_partial_shape()[1] != (hidden_size + input_size)) {
```
popovaan, line 153: Can shape size be less than 2 here? Should we add check?

popovaan:
Will this functionality work for importing from memory case? Can we add a test?

src/common/transformations/src/transformations/common_optimizations/gru_cell_fusion.cpp
rkazants: It is preliminary checked in the pattern definition. Please see here: https://github.com/openvinotoolkit/openvino/blob/master/src/common/transformations/src/transformations/common_optimizations/gru_cell_fusion.cpp#L91

tools/mo/unit_tests/moc_tf_fe/conversion_basic_models_test.py
rkazants: Fixed.

rkazants:
> Will this functionality work for importing from memory case? Can we add a test?

No in this PR please, because I think we need to adopt layer tests to work in two modes (from disk and from memory). And it will cover automatic testing for the future new operations and functionality.

popovaan (approved):

src/common/transformations/src/transformations/common_optimizations/remove_multi_subgraph_op_dangling_params.cpp
```diff
@@ -157,6 +159,14 @@ bool ov::pass::RemoveMultiSubGraphOpDanglingParamsResults::run_on_model(const st
...
                     }
                 }
+                for (auto input_index : required_inputs_indices) {
+                    if (input_index > removed_loop_idx) {
+                        new_required_inputs_indices.insert(input_index - 1);
+                    } else {
+                        new_required_inputs_indices.insert(input_index);
+                    }
+                }
+                required_inputs_indices = new_required_inputs_indices;
```
itikhono, line 169: I thinks it's better to erase/insert then a creation a new set

itikhono (approved):
Transformations part looks good

src/common/transformations/src/transformations/common_optimizations/remove_multi_subgraph_op_dangling_params.cpp
rkazants: I am afraid that we iterator through container and erasing in the fly (during iteration) is not good

rkazants:
build_jenkins

ababushk:
build_jenkins

PR #20282, opened 2023-10-06
[core]Migrate squeeze operator to new API

praasz:
### Details:
 - Migrate Squeeze operator to new API, remove legacy classes, headers, and util functions.
 - Refactor for binary size reduction.

### Tickets:
 - [CVS-118671](https://jira.devtools.intel.com/browse/CVS-118671)


src/core/src/op/squeeze.cpp
```diff
@@ -40,115 +60,69 @@ void op::Squeeze::validate_and_infer_types() {
...
-        default:
-            break;
-        }
-        return false;
-    } else if (get_input_size() == 1) {
-        return true;
-    } else {
-        return false;
-    }
+    return (get_input_size() < 2) || validate::axes_type(get_input_element_type(1));
```
ilyachur, line 90: nit: Maybe we can have internal lambda instead of introducing ` validate::axes_type`

src/core/include/openvino/op/util/evaluate_helpers.hpp
```diff
@@ -0,0 +1,23 @@
...
+namespace op {
+namespace util {
+
+/**
+ * @brief Get the tensors shapes as ov::PartialSHape
+ *
+ * @param tensors  Input tensors vector to get its shapes.
+ * @return Vector of partial shapes sam size as input tensor vector.
+ */
+std::vector<PartialShape> get_tensors_shapes(const TensorVector& tensors);
```
ilyachur, line 20: Why not `ov::Shape`?

praasz: As it easier to use with `shape_infer` functions.
Will rename it to `get_tensors_partial_shapes` to be more explicit what this function will return.

t-jankowski (approved):

ilyachur (approved):

PR #20308, opened 2023-10-09
[Snippets][CPU] Disabled MHA tokenization with infer precision f16

a-sidorova:
### Details:
 - *CPU Plugin Subgraph supports BRGEMM only with precisions `f32`, `bf16`, `i8`. If precision is `f16`, MHA should be executed with this precision as separate node set on plugin side. Thus, the PR enabled MHA tokenization only with infer precision `f32`, `bf16`*

### Tickets:
 - *N/A*

### DIFF
--- a/src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
+++ b/src/plugins/intel_cpu/src/transformations/transformation_pipeline.cpp
@@ -591,7 +591,7 @@ void Transformations::MainSnippets(void) {
 
     // - MHA has BRGEMM that is supported only on AVX512 platforms
     // - CPU Plugin Subgraph supports only f32, bf16 (and quantized) BRGEMM
-    //   TODO[122494]: Need to add support of f16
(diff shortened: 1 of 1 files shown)


a-sidorova:
@IvanNovoselov @dmitry-gorokhov may I ask to take a look please?

The PR includes the fix for transformer models for cases when infer precision `f16`. There are two points:

- At the moment, after canonicalization we insert `Convert` ops on inputs `f16->f32`:
image
These `Transposes` are not decomposed and not fused into `BRGEMM` -> generator doesn't support explicit `Transpose` op.

- Since we don't support `BRGEMM` on `f16` precision, let it be executed on plugin side on `f16` precision for best performance.

IvanNovoselov (approved):
Please add this image and explanation to the ticket

a-sidorova:
@IvanNovoselov thank you for the approve and reminder. Added.

@dmitry-gorokhov take a look please

PR #20323, opened 2023-10-09
fix random layer names and count

evkotov:
### Details:
 - add sorting by instance id to shared_node_optimization() SharedOpOptimization transformation that causes random output layer names

### Tickets:
 - 118730


src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -51,6 +51,10 @@ bool shared_node_optimization(const shared_ptr<Model>& model, const rules_t& rul
                     continue;
                 const auto& are_equal = rules.at(item.first);
 
+                std::sort(shared_nodes.begin(), shared_nodes.end(), [](Node* l_node, Node* r_node) {
```
itikhono, line 54: is it possible to add some test?

src/common/transformations/src/transformations/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -51,6 +51,10 @@ bool shared_node_optimization(const shared_ptr<Model>& model, const rules_t& rul
...
+                    return l_node->get_instance_id() < r_node->get_instance_id();
```
itikhono, line 55: I looks like the actual root cause is in `get_target_inputs` function
image
image
It sorts the resulting Inputs by the address of the nodes
So maybe it's better to fix this function.
@jane-intel @evkotov what do you think?

evkotov: @ilyachur could you also comment this problem?

evkotov: I am afraid of changing basic functionality in library, since it could cause some side effects.

evkotov: Created [PR](https://github.com/openvinotoolkit/openvino/pull/20365) to check side effects of changing Input/Output comparison

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -433,3 +433,64 @@ TEST_F(SharedTransformationTestsF, SharedShapeOfTestMixed) {
...
+        if (node->get_friendly_name() == name)
+            return node;
+    }
+
+    return {};
+}
+}  // namespace
+
+TEST(TransformationTests, SharedShapeOfTestRandomOrder) {
+    // manager.register_pass<pass::SharedOpOptimization>();
```
itikhono, line 484: please delete the comment

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -433,3 +433,64 @@ TEST_F(SharedTransformationTestsF, SharedShapeOfTestMixed) {
...
 }
+
+namespace {
+template <typename T, typename... NodeArgTypes>
+OutputVector createNodesInMemory(const std::vector<size_t>& node_order,
+                                 std::shared_ptr<T[]>& memory,
+                                 const std::string& node_name_prefix,
+                                 NodeArgTypes&&... node_args) {
+    OutputVector outputs;
+    memory.reset(new T[node_order.size()]);
```
itikhono, line 444: are you sure that destructor of ShapeOf objects will be called correctly if they stored as raw array?
could you double check?

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -433,3 +433,64 @@ TEST_F(SharedTransformationTestsF, SharedShapeOfTestMixed) {
...
+    std::shared_ptr<v3::ShapeOf[]> buffer;
+
+    for (const auto& shapes_order : std::vector<std::vector<size_t>>{{0, 1}, {1, 0}}) {
+        auto model = createModelWithShapes(input_shape, shapes_order, buffer);
+
+        ov::pass::Manager manager;
+        manager.register_pass<pass::SharedOpOptimization>();
+        manager.run_passes(model);
+        ASSERT_TRUE(findNodeByName(model, "Shape_0") != nullptr) << "node Shape_0 is not found in model";
+    }
```
itikhono, line 495: please add an additional check that after the transformation both these graphs have to be equal

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -433,3 +433,64 @@ TEST_F(SharedTransformationTestsF, SharedShapeOfTestMixed) {
...
+    return {};
+}
+}  // namespace
+
+TEST(TransformationTests, SharedShapeOfTestRandomOrder) {
+    // manager.register_pass<pass::SharedOpOptimization>();
+    Shape input_shape{120, 4};
+    std::shared_ptr<v3::ShapeOf[]> buffer;
+
+    for (const auto& shapes_order : std::vector<std::vector<size_t>>{{0, 1}, {1, 0}}) {
```
itikhono, line 488: it's better to create a separate variable for `std::vector<std::vector<size_t>>{{0, 1}, {1, 0}}`
it's not so clear what do {0, 1} and {1, 0} mean here without additional analysis of createNodesInMemory function. So maybe it's better to add a comment here

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
itikhono, line 444: could you check that a destructor of each ShapeOf obj in this array will be called in this case?

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -433,3 +433,78 @@ TEST_F(SharedTransformationTestsF, SharedShapeOfTestMixed) {
...
+namespace {
+template <typename T, typename... NodeArgTypes>
+OutputVector createNodesInMemory(const std::vector<size_t>& node_order_in_memory,
+                                 std::shared_ptr<void>& memory,
+                                 const std::string& node_name_prefix,
+                                 NodeArgTypes&&... node_args) {
+    OutputVector outputs;
+    memory.reset(::malloc(node_order_in_memory.size() * sizeof(T)), ::free);
+    for (size_t i = 0; i < node_order_in_memory.size(); ++i) {
+        T* node_addr = static_cast<T*>(memory.get()) + node_order_in_memory[i];
```
itikhono, line 446: are T* and static_cast<T*> necessary here?

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -433,3 +433,78 @@ TEST_F(SharedTransformationTestsF, SharedShapeOfTestMixed) {
         model_ref = std::make_shared<Model>(NodeVector{concat}, ParameterVector{input});
     }
 }
+
+namespace {
+template <typename T, typename... NodeArgTypes>
+OutputVector createNodesInMemory(const std::vector<size_t>& node_order_in_memory,
+                                 std::shared_ptr<void>& memory,
+                                 const std::string& node_name_prefix,
+                                 NodeArgTypes&&... node_args) {
```
itikhono, line 442: NodeArgTypes&&...
it looks like it's always shared_ptr<Node>, so maybe it's better to change argument type and get rid of std::forward

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
```diff
@@ -433,3 +433,78 @@ TEST_F(SharedTransformationTestsF, SharedShapeOfTestMixed) {
...
+    for (const auto& shape_node : shape_nodes) {
+        auto node = std::make_shared<v0::Convert>(shape_node, element::i64);
+        inputs_of_concat.push_back(node);
+    }
+
+    auto concat = std::make_shared<v0::Concat>(inputs_of_concat, 0);
+    return std::make_shared<Model>(NodeVector{concat}, ParameterVector{input});
+}
+
+std::shared_ptr<Node> findNodeByName(const std::shared_ptr<Model>& model, const std::string& name) {
```
itikhono, line 477: why not std::find_if?

src/common/transformations/tests/common_optimizations/shared_ops_optimization.cpp
evkotov: This is to avoid warning: "pointer of type ‘void *’ used in arithmetic [-Wpointer-arith]"

itikhono (approved):

ilyachur (approved):

mateusztabaka (approved):

PR #20329, opened 2023-10-10
[core] fix memory leak issue imported by #18868

tiger100256-hu:
fix the memory leak issue that CpuStreamExecuter is released, but there are still thread id in t_stream_count_map

### Details:
 - *fix memory issue in https://github.com/openvinotoolkit/openvino/commit/85609d48813ea43f4470ea27404e31a14f306fa2*
 - *port from https://github.com/openvinotoolkit/openvino/pull/19832*

### Tickets:
 - *CVS-120233*

### DIFF
--- a/src/inference/src/dev/threading/cpu_streams_executor.cpp
+++ b/src/inference/src/dev/threading/cpu_streams_executor.cpp
@@ -4,6 +4,7 @@
 
 #include "openvino/runtime/threading/cpu_streams_executor.hpp"
 
+#include <atomic>
 #include <condition_variable>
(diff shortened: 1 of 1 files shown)


vurusovs (approved):
We have a leak issue https://github.com/openvinotoolkit/openvino/issues/20633, and I want to confirm these changes not related to it

tiger100256-hu:
@vurusovs sorry for  this late reply,  this pr is not the root cause of issue https://github.com/openvinotoolkit/openvino/issues/20633

PR #20333, opened 2023-10-10
Move create model sample to the last opset

ilyachur:
### Details:
 - Create model sample should use only one opset

### Tickets:
 - *ticket-id*


samples/cpp/model_creation_sample/main.cpp
```diff
@@ -108,7 +107,7 @@ std::shared_ptr<ov::Model> create_model(const std::string& path_to_weights) {
     Shape padBeginShape{0, 0};
     Shape padEndShape{0, 0};
 
-    auto maxPoolingNodeFirst = std::make_shared<opset1::MaxPool>(addNodeFirst->output(0),
+    auto maxPoolingNodeFirst = std::make_shared<opset8::MaxPool>(addNodeFirst->output(0),
```
mitruska, line 110: The CI is failing because the `v8::MaxPool` constructor requires additional attr `dilations` to be provided. To keep compatibility with `v1` it can be filled with ones:



```
auto maxPoolingNodeFirst = std::make_shared<opset8::MaxPool>(
addNodeFirst->output(0),
Strides{2, 2},

Strides{1, 1}, // <- Additional attr "dilations" in v8::MaxPool

padBeginShape,
padEndShape,
Shape{2, 2},
op::RoundingType::CEIL);
```
Upgrade transformation for MaxPool as reference: [transformations/op_conversions/convert_maxpool_upgrade.cpp](https://github.com/openvinotoolkit/openvino/blob/b345f3c324ae700d55211d4a4d071152a47c4449/src/common/transformations/src/transformations/op_conversions/convert_maxpool_upgrade.cpp#L26C1-L40)

samples/cpp/model_creation_sample/main.cpp
mitruska, line 1: The title of the PR should be aligned with the current changes 
(unification to `opset8` instead of moving to the "last opset" (which is `opset13`).

mitruska (commented):
I've left a comment regarding the difference between `v1::MaxPool` and `v8::MaxPool`  constructors, as the reason of failing CI.

ilya-lavrenov (approved):

PR #20340, opened 2023-10-10
[Opset13][TF FE] Enable tensorflow bitwise operators

mmikolajcz:
### Details:
 - *Enable tensorflow bitwise operators: BitwiseAnd, BitwiseOr, BitwiseXor and Invert*
 - *Extend tensorflow unary and binary test cases*

### Tickets:
 - *118977*
 - *118969*
 - *118961*
 - *118953*


tests/layer_tests/tensorflow_tests/test_tf_BinaryOps.py
```diff
@@ -123,7 +132,7 @@ def create_add_placeholder_const_net(self, x_shape, y_shape, ir_version, op_type
                               'Equal', 'NotEqual', 'Mod', 'Greater', 'GreaterEqual', 'Less',
                               'LessEqual',
                               'LogicalAnd', 'LogicalOr', 'LogicalXor', 'FloorMod', 'FloorDiv',
-                              'Xdivy'])
+                              'Xdivy', 'BitwiseAnd', 'BitwiseOr', 'BitwiseXor',])
     @pytest.mark.nightly
     @pytest.mark.precommit
```
rkazants, line 137: please create a separate test suite for Bitwise because you have added to one which is not included into the pre-commit (no `precommit_tf_fe` mark ).

tests/layer_tests/tensorflow_tests/test_tf_UnaryOps.py
```diff
@@ -152,6 +158,7 @@ def create_net_with_unary_op(self, shape, ir_version, op_type, use_new_frontend)
                                          'LogicalNot',
                                          'Square',
                                          'Erf',
+                                         'BitwiseNot'
```
rkazants, line 161: the same comment

tests/layer_tests/tensorflow_tests/test_tf_BinaryOps.py
mmikolajcz: From what I see, 1 case from test data is marked as precommit_tf_fe in inputs list just above first pytest decorator. If this 1 test is not enough, maybe it will be worth to add few more tests for cover all binary ops?

tests/layer_tests/tensorflow_tests/test_tf_UnaryOps.py
mmikolajcz: Similar case as in binary, in L189 single case is marked as precommit_tf_fe, in addition there is also separate test case with just "precommit" mark that seem to just duplicate second test. Maybe they should be merged together?

rkazants (approved):

mitruska (approved):

PR #20345, opened 2023-10-10
Handle Reshape in SplitSqueezeConcatFusion

mateusztabaka:
SplitSqueezeConcatFusion fuses Split->(multiple) Squeeze->Concat subgraphs. This change extends its functionality to handle Split->(multiple) Reshape->Concat, in situations where Reshape works the same as Squeeze. 

This change is motived by some recent changes to transformation pipeline.
Recently, NopElimination was added to SimplifyShapeOfSubGraph, which happes to run before SplitSqueezeConcatFusion.
One of transformations that is part of NopElimination replaces Squeeze with Reshape, which ultimately breaks SplitSqueezeConcatFusion "pattern".
Therefore SplitSqueezeConcatFusion  can be extended to support Reshape as well - so any model with Split->(multiple) Reshape->Concat subgraphs can benefit from that.

Ticket: CVS-122455


src/common/transformations/src/transformations/common_optimizations/split_squeeze_concat_fusion.cpp
```diff
@@ -120,3 +107,40 @@ ov::pass::SplitSqueezeConcatFusion::SplitSqueezeConcatFusion() {
...
+
+    // check if squeezed dimension equals to 1
+    if (input_shape[axis].is_dynamic() || input_shape[axis] != 1)
+        return false;
+
+    // check if the dimensions surrounding squeezed axis matches:
+    // input shape = [..., a, 1, b, ...]
+    // output shape = [..., x, y, ...]
+    // function returns false if a != x or b != y
+    if (axis > 0) {
```
itikhono, line 132: I believe we need to check all dims, not only x,y

Maybe this approach is better:
1. delete dim at axis from input_shape
2. compare all dims in input_shape and output_shape, they have to be equal

mateusztabaka: done

src/common/transformations/src/transformations/common_optimizations/split_squeeze_concat_fusion.cpp
```diff
@@ -120,3 +107,65 @@ ov::pass::SplitSqueezeConcatFusion::SplitSqueezeConcatFusion() {
...
+    // check if output_rank == input_rank - 1
+    // to make sure the node actually squeezes a dimension
+    if (input_rank != output_rank + 1)
+        return false;
+
+    // check if squeezed dimension equals to 1
+    if (input_shape[axis].is_dynamic() || input_shape[axis] != 1)
+        return false;
+
+    if (ov::is_type<ov::op::v1::Reshape>(squeeze_node)) {
```
itikhono, line 128: it looks like we need to check 2nd input to Reshape operation because it can contain these values:
image
we might break reshapability in some cases

itikhono (changes requested):

andrei-kochin (changes requested):
No approval yet -> 2023.3 scope
Let's wait for 2023.2 branch out before merge.

src/common/transformations/src/transformations/common_optimizations/split_squeeze_concat_fusion.cpp
itikhono: fixed

itikhono (approved):

evkotov (approved):

andrei-kochin (approved):

PR #20354, opened 2023-10-10
[MO][Opset13] NMSRotated-13 support in MO IR Reader

mitruska:
### Details:
 - Add NMSRotated-13 to MO (support in MO IR Reader)
 - Extend `check_graph_can_save` test class method to return ov model and enable reloaded model testing and avoid files collision (changes from: https://github.com/openvinotoolkit/openvino/pull/20446)

### Tickets:
 - 115304


mmikolajcz (approved):

tools/mo/openvino/tools/mo/ops/nms_rotated.py
```diff
@@ -0,0 +1,80 @@
...
+        if num_of_outputs >= 2 and node.has_port('out', 1):
+            node.out_port(1).data.set_shape(
+                shape_array([dynamic_dimension_value, 3]))
+        if num_of_outputs >= 3 and node.has_port('out', 2):
+            node.out_port(2).data.set_shape(shape_array([1]))
+
+    @staticmethod
+    def type_infer(node: Node):
+        node.out_port(1).set_data_type(np.float32)
+        assert node.has_valid('output_type')
```
mvafin, line 74: Please add message to this assert

mitruska: Actually `has_valid` helper checks only if the key exists and is not None, so moved the check to the `if` below, 
as the `output_type` attr is optional and `i64` by default.

mvafin (approved):

jiwaszki (approved):

PR #20361, opened 2023-10-10
[nGraph Transformatoin] Copy rtinfo from all replaced nodes in GroupedSliceToVSplitOptimization

nkogteva:
### Details:
 - *Copy rtinfo from all replaced nodes in GroupedSliceToVSplitOptimization (currently last one overrides rt info)*

### Tickets:
 - *ticket-id*

### DIFF
--- a/src/common/transformations/src/transformations/common_optimizations/optimize_strided_slice.cpp
+++ b/src/common/transformations/src/transformations/common_optimizations/optimize_strided_slice.cpp
@@ -258,7 +258,7 @@ bool ov::pass::GroupedStridedSliceOptimizer::run_on_model(const std::shared_ptr<
             }
             ++i;
         }
-        copy_runtime_info(ops_to_replace, variadic_split);
+        copy_runtime_info(ops_to_replace, {variadic_split, axis_const, size_splits_const});
     }
     return graph_rewritten;
 }
@@ -405,12 +405,14 @@ bool ov::pass::GroupedSliceToVSplitOptimization::run_on_model(const std::shared_
         auto variadic_split = std::make_shared<op::v1::VariadicSplit>(output, axis_const, split_lengths_const);
 
         auto i = 0;
+        NodeVector ops_to_replace;
         for (auto& slice_with_attrs : attributes) {
-            graph_rewritten |=
-                ov::replace_output_update_name(slice_with_attrs.slice->output(0), variadic_split->output(i));
-            ov::copy_runtime_info(slice_with_attrs.slice, variadic_split);
+            graph_rewritten = true;
+            slice_with_attrs.slice->output(0).replace(variadic_split->output(i));
+            ops_to_replace.push_back(slice_with_attrs.slice);
             ++i;
         }
+        ov::copy_runtime_info(ops_to_replace, {variadic_split, axis_const, split_lengths_const});
     }
     return graph_rewritten;
 }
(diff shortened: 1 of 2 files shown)


evkotov (approved):

evkotov:
I don't see ticket id in the description. Is that how it's meant to be?

src/common/transformations/src/transformations/common_optimizations/optimize_strided_slice.cpp
```diff
@@ -258,7 +258,7 @@ bool ov::pass::GroupedStridedSliceOptimizer::run_on_model(const std::shared_ptr<
             }
             ++i;
         }
-        copy_runtime_info(ops_to_replace, variadic_split);
+        copy_runtime_info(ops_to_replace, {variadic_split, axis_const, size_splits_const});
```
itikhono, line 261: do we need to update rt_info checks in the tests to catch this?

nkogteva: > do we need to update rt_info checks in the tests to catch this?

Actually it is hard to catch. There is a workaround in copy_runtime_info which adds Constant to targets if they have no rt_info yet https://github.com/openvinotoolkit/openvino/blob/9bedafb560fd4bc0e4627c8cac67cc0236a72e86/src/core/src/rt_info.cpp#L85 But I think it is better to add them explicitly.

src/common/transformations/src/transformations/common_optimizations/optimize_strided_slice.cpp
```diff
@@ -405,12 +405,14 @@ bool ov::pass::GroupedSliceToVSplitOptimization::run_on_model(const std::shared_
         auto variadic_split = std::make_shared<op::v1::VariadicSplit>(output, axis_const, split_lengths_const);
 
         auto i = 0;
+        NodeVector ops_to_replace;
         for (auto& slice_with_attrs : attributes) {
-            graph_rewritten |=
-                ov::replace_output_update_name(slice_with_attrs.slice->output(0), variadic_split->output(i));
-            ov::copy_runtime_info(slice_with_attrs.slice, variadic_split);
+            graph_rewritten = true;
+            slice_with_attrs.slice->output(0).replace(variadic_split->output(i));
```
itikhono, line 411: ".replace" method doesn't preserve some legacy names (friendly names, legacy tensor names)  unlike replace_output_update_name method.
@ilyachur do we still need to preserve these legacy names?

nkogteva: @itikhono Method `replace` is used in lot of transformations meanwhile.

nkogteva: The same transformation several linses above https://github.com/openvinotoolkit/openvino/blob/5f9c913d6444f27c311bdf73ba9885b334eedc71/src/common/transformations/src/transformations/common_optimizations/optimize_strided_slice.cpp#L256

mateusztabaka: I think you should still use replace_output_update_name. Otherwise if Slice output is a model output, you can encounter issues like: "Cannot find blob with name:"

> The same transformation several lines above

That can be an issue as well.

PR #20365, opened 2023-10-10
fix Output and Input comparison to use instance id instead of addresses

evkotov:
### Details:
 - fix Output and Input comparison operators
 - addresses differ on each run instead of instance id

### Tickets:
 - 118730

### DIFF
--- a/src/core/src/node_input.cpp
+++ b/src/core/src/node_input.cpp
@@ -61,11 +61,13 @@ bool Input<Node>::operator!=(const Input& other) const {
     return !(*this == other);
 }
 bool Input<Node>::operator<(const Input& other) const {
-    return m_node < other.m_node || (m_node == other.m_node && m_index < other.m_index);
+    return m_node->get_instance_id() < other.m_node->get_instance_id() ||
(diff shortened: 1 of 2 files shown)


src/core/src/node_input.cpp
```diff
@@ -61,11 +61,13 @@ bool Input<Node>::operator!=(const Input& other) const {
     return !(*this == other);
 }
 bool Input<Node>::operator<(const Input& other) const {
-    return m_node < other.m_node || (m_node == other.m_node && m_index < other.m_index);
```
itikhono, line 64: also for `operator == `

evkotov: Is it needed? if addresses are equal, than instance id also should be equal? Or not?

evkotov: In operators< and > I also don't use instance id for checking equivalence

itikhono: ok, agree

itikhono:
Looks good, CI is green
Could you add a new test?

evkotov:
This experiment is seemed to be successfull.

PR #20368, opened 2023-10-10
[CPU] Dynamic shapes support for memory states with zero copy

maxnick:
### Details:
The purpose of the PR is to provide dynamic shapes support for the memory states, as well as to minimize copying of internal memory while maintaining the functionality of the subsystem.

### Tickets:
 - *124784*


src/plugins/intel_cpu/src/memory_state.h
```diff
@@ -17,15 +17,38 @@ namespace intel_cpu {
...
     void Reset() override;
+    void SetState(const InferenceEngine::Blob::Ptr& newState) override;
+    InferenceEngine::Blob::CPtr GetState() const override;
+
+    void SwapBuffer(); //controversial thing, increase probability of error
+    MemoryPtr InternalMem() const {
+        return m_internal_mem[buffer_num];
+    }
+
+    MemoryDescPtr InitalDesc() const {
```
maxnick, line 33:
```suggestion
    MemoryDescPtr OriginalDesc() const {
```

src/plugins/intel_cpu/src/node.cpp
```diff
@@ -1714,7 +1719,13 @@ int Node::inPlaceInputPort(int portIdx) const {
 
     return conf.inConfs[portIdx].inPlace();
 }
+
 int Node::inPlaceOutPort(int portIdx) const {
+    if (outputShapes.empty()) {
+        //special cas - a dead end node
```
maxnick, line 1725:
```suggestion
        //special case - a dead end node
```

src/plugins/intel_cpu/src/nodes/memory.hpp
```diff
@@ -9,6 +9,8 @@
 #include "ie_algorithm.hpp"
 #include "input.h"
 #include <node.h>
+#include <memory_state.h>
```
maxnick, line 12: Remove

src/plugins/intel_cpu/src/infer_request.cpp
```diff
@@ -95,48 +96,26 @@ void InferRequestBase::pushInput(const std::string& inputName, InferenceEngine::
     graph->PushInputData(inputName, needConvert ? iconv : inputBlob);
 }
 
-void InferRequestBase::PushStates() {
+// state -> node
+void InferRequestBase::AssignStates() {
     for (auto &node : graph->GetNodes()) {
         if (node->getType() == Type::MemoryInput) {
-            auto cur_node = dynamic_cast<node::MemoryInput*>(node.get());
+            auto cur_node = std::dynamic_pointer_cast<node::MemoryInput>(node);
```
maxnick, line 103: Could be avoided by saving the `MemInput` nodes in the graph in a separate container, just like for the I/O nodes.

src/plugins/intel_cpu/src/infer_request.cpp
```diff
@@ -47,19 +47,20 @@ void InferRequestBase::CreateInferRequest() {
...
             // Remove suffix with pair ID. Internal information.
             auto suffix_idx = state_name.find("/id=");
-            if (suffix_idx != std::string::npos)
+            if (suffix_idx != std::string::npos) {
                 state_name = state_name.substr(0, suffix_idx);
+            }
 
-            memoryStates.emplace_back(new VariableState(state_name, state_store));
+            memoryStates.emplace_back(
+                std::make_shared<VariableStateDoubleBuffer>(state_name, memoryNode->memoryBuilder(), memoryNode->getMemoryPtr()));
```
maxnick, line 63: It might make sense to move the instantiation of the state objects to the MemoryInput nodes. That will allow to hide the selection of the concrete state type on the graph/node level.

src/plugins/intel_cpu/src/infer_request.cpp
```diff
@@ -192,6 +168,10 @@ void InferRequestBase::InferImpl() {
         }
     }
 
+    if (!memoryStates.empty()) {
+        CommitStates();
+    }
+
```
maxnick, line 174: Move to `MemoryOutput::execute`?

src/plugins/intel_cpu/src/infer_request.cpp
maxnick: done

src/plugins/intel_cpu/src/nodes/reorder.cpp
```diff
@@ -457,9 +457,24 @@ void Reorder::reorderData(const IMemory &input, const IMemory &output, MultiCach
...
+            auto reshapedSrcMemDesc = srcMemoryDesc.reshape(dstMemoryDesc.get_dims(), allowEmpty);
+            if (reshapedSrcMemDesc) {
+                srcMemoryDesc = reshapedSrcMemDesc;
+                srcMemory = dnnl::memory(srcMemoryDesc, engine, srcMemory.get_data_handle());
+            }
+        }
+
         // try directly reorder
-        reorder = getReorderPrim(cache, dstMemory.get_engine(), srcMemory.get_desc(), dstMemory.get_desc());
+        reorder = getReorderPrim(cache, engine, srcMemoryDesc, dstMemoryDesc);
```
ceciliapeng2011, line 477: When will this happen? Is it safe to simply reshape source mem desc with destination's?

maxnick: It may happen for example when ReadValue output shape is `{2, 1, 2}`, but Assign node input shape `{2, 2}`. Ngraph allows  such combinations. Yep, it's safe as long as such a reshape is allowed, since it doen't require tossing elements in memory.

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/stateful_model.cpp
```diff
@@ -0,0 +1,547 @@
...
+
+        // The ReadValue/Assign operations must be used in pairs in the model.
+        // For each such a pair, its own variable object must be created.
+        const std::string variable_name("variable0");
+        auto variable = std::make_shared<ov::op::util::Variable>(
+            ov::op::util::VariableInfo{ov::PartialShape::dynamic(), ov::element::dynamic, variable_name});
+
+        // Creating ov::Model
+        auto read = std::make_shared<ov::op::v6::ReadValue>(arg, variable);
+        std::vector<std::shared_ptr<ov::Node>> args = {arg, read};
```
ceciliapeng2011, line 304: Do we expect a behavior difference of the two lines below?
`auto read = std::make_shared<ov::op::v6::ReadValue>(arg, variable);`
and
`auto read = std::make_shared<ov::op::v6::ReadValue>(variable);`   (which is newly added by pull/20404)

Should the first object be initialized by `arg`? If so, the reference state output after the first iteration should be concat(2.44f, add(2.44f, 2.44f)), instead of concat(2.44f, add(2.44f, 0.f)).

@itikhono Would you please comment on the behavior of state with a initial state from a subgraph?

maxnick: To my understanding, in this particular tests, the behavior is the same. On L340 `states.front().set_state(init_tensor);` with 0 value is explicitly called, thus we always start with `concat(2.44f, add(2.44f, 0.f))`.

src/plugins/intel_cpu/src/nodes/memory.cpp
```diff
@@ -118,57 +235,189 @@ MemoryInput::MemoryInput(const std::shared_ptr<ov::Node>& op, const GraphContext
...
+            getName(),
+            " unexpected state shape: ",
+            vec2str(newDims),
+            ", while the expected shape: ",
+            vec2str(expectedDims));
+
+        internDesc = getBaseMemDescAtOutputPort(0);
+    }
+
+    OPENVINO_ASSERT(memMngr,
```
ceciliapeng2011, line 363: Do we expext memMngr is always valid here? How do we make sure MemoryInput::resolveInPlaceEdges is always involked and hence memMngr available?

src/plugins/intel_cpu/src/nodes/memory.cpp
```diff
@@ -60,39 +57,159 @@ MemoryOutput::MemoryOutput(const std::shared_ptr<ov::Node>& op, const GraphConte
...
+    OPENVINO_ASSERT(extMemDesc,
+        "MemoryOutput ",
+        getName(),
+        " assigned state has null base mem desc ptr");
+
+    if (!memMngr) { return; } //nothing to do, edge memory isn't under control
+    auto inpDesc = getBaseMemDescAtInputPort(0);
+
+    if (inpDesc->isCompatible(*extMemDesc)) {
+        memMngr->setMemMngrResize(assignedMem->getMemoryMngr());
```
ceciliapeng2011, line 161: Why does it need setMemMngrResize instead of setMemMngr here? The resize stage of its parent node will resize it, won't it?

src/plugins/intel_cpu/src/infer_request.h
```diff
@@ -81,7 +82,7 @@ class SyncInferRequest : public ov::ISyncInferRequest {
         }
 
         void update() {
-            m_proxyMemMngr->setMemMngr(currentMemMngr());
+            m_proxyMemMngr->setMemMngrResize(currentMemMngr());
```
ceciliapeng2011, line 85: Why cannot the previous setMemMngr work now?

src/plugins/intel_cpu/src/nodes/memory.cpp
```diff
@@ -118,57 +235,189 @@ MemoryInput::MemoryInput(const std::shared_ptr<ov::Node>& op, const GraphContext
...
+        inPortConfig.inPlace(-1);
+        inPortConfig.constant(false);
+        inPortConfig.setMemDesc(descCreators.at(LayoutType::ncsp)->createSharedDesc(precision, shape));
+
+        config.inConfs.push_back(std::move(inPortConfig));
+    }
+
+    PortConfig outPortConfig;
+
+    outPortConfig.inPlace(0);
```
ceciliapeng2011, line 269: I think here we simply reuses the interface of inPlace, but changes its original meaning, right?
outPortConfig.inPlace(0) originally means this outport reuses memory from input port 0. But obviously, it is not. The memory is from state variable instead.

ceciliapeng2011: That's why you `resolve inplace dead end nodes` in the beginning of `Graph::Allocate()`, right?

src/plugins/intel_cpu/src/nodes/reorder.cpp
ceciliapeng2011: When is a Reorder need for this ReadValue/Assign pair?

src/plugins/intel_cpu/src/nodes/memory.cpp
```diff
@@ -118,57 +235,189 @@ MemoryInput::MemoryInput(const std::shared_ptr<ov::Node>& op, const GraphContext
...
+    }
+
+    auto selected_pd = getSelectedPrimitiveDescriptor();
+    OPENVINO_ASSERT(selected_pd,
+        "MemoryInput ",
+        getName(),
+        " failed getSelectedPrimitiveDescriptor() call, preferable primitive descriptor is not set");
+
+    auto memDesc = selected_pd->getConfig().outConfs.front().getMemDesc();
+    memMngr = std::make_shared<ProxyMemoryMngr>();
```
ceciliapeng2011, line 316: Can we use the same way of change_default_ptr to MemoryInput, instead of using a proxy mem manager?

src/plugins/intel_cpu/src/memory_state.cpp
```diff
@@ -1,18 +1,131 @@
...
+    if (prime_mem()->getShape() != shape) {
+        auto new_desc = m_internal_desc->cloneWithNewDims(shape.getStaticDims());
+        prime_mem()->redefineDesc(new_desc);
+    }
+
+    auto src = m_state->data();
+
+    Memory mem(get_engine(), state_desc, src);
+    prime_mem()->load(mem);
+}
```
ceciliapeng2011, line 56: Do we need a check to check if `state` is actually `prime_mem()`? If so we can avoid a copy here. e.g., in a user call sequence `state = get_state()` -> do some change in place -> `set_state(state)`.

src/plugins/intel_cpu/src/nodes/memory.cpp
maxnick: It's always invoked by the graph logic as the node sets inPlace port number different form `-1`. But, to handle any unexpected situations, this assert is there.

src/plugins/intel_cpu/src/nodes/memory.cpp
maxnick: The reason is the same as for the Output nodes memory. The resize step may be skipped if the `needShapeInferCall()` return false, so we may not rely on the parent node.

src/plugins/intel_cpu/src/infer_request.h
maxnick: Because it calls resize unconditionally, but this applicable in not all the contexts. Especially in the case when we transfer the input memory. Sometimes the input memory size is less, than the previous one, and the old logic would trigger resize unconditionally with loosing the underlining memory values.

src/plugins/intel_cpu/src/nodes/memory.cpp
maxnick: Yep, exactly. This memory substitution fits into the inplace memory model completely, except the constraints on its description. So there was to options: 1. Add a fake edge with the memory from the state 2. Remove constraints on persistence of the edge that is referenced. First option is more bulky as requires a special treatment from the graph logic, while the second one requires only one additional condition for the dead end nodes.

src/plugins/intel_cpu/src/nodes/memory.cpp
maxnick: In theory yes. This is rather an implementation details hidden inside the MemoryInput node, they doesn't affect anything outside. And neither of the solution is better then the other in this case.

src/plugins/intel_cpu/src/memory_state.cpp
maxnick: Yep, we can have some optimizations on this level. For example we can try to reuse the user's memory directly as a ro input on the first iteration, but it requires some additional checks to prevent spoiling user's memory. With no doubt it can be optimized later within a separate task.

src/plugins/intel_cpu/src/nodes/reorder.cpp
maxnick: The memory state tensor may have only one shape either {2, 1, 2} or {2, 2} , it may not have both at the same time, thus when either ReadValue or Assign edge is processed, a `load()` call happens, thus the reorder is invoked. Also this is useful in a general case, when we have memory objects of shape {2, 1, 2} and {2, 2}. Having this change we can conveniently call load, without additional checks and memory desc conversions.

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/stateful_model.cpp
```diff
@@ -0,0 +1,547 @@
...
+
+        auto arg = std::make_shared<ov::op::v0::Parameter>(netPrc, ov::Shape{1, 1});
+        auto init_const = ov::op::v0::Constant::create(netPrc, ov::Shape{1, 1}, {0});
+
+        // The ReadValue/Assign operations must be used in pairs in the model.
+        // For each such a pair, its own variable object must be created.
+        const std::string variable_name("variable0");
+        auto variable = std::make_shared<ov::op::util::Variable>(
+            ov::op::util::VariableInfo{ov::PartialShape::dynamic(), ov::element::dynamic, variable_name});
+
```
luo-cheng2021, line 85: According to PR #20404 , the output shape will be [got](https://github.com/openvinotoolkit/openvino/blob/6adc6064b0c0340dade3844be89710d9a92c9f0c/src/core/src/op/read_value.cpp#L62) from the state instead of `init_cost`, then a dynamic rank node will be created and the cpu plugin may raise error.

src/plugins/intel_cpu/src/memory_state.cpp
```diff
@@ -1,18 +1,131 @@
...
+            cpu_convert(prime_mem()->getData(), mem->getData(), internal_prc, external_prc, elements_to_convert);
+            return std::make_shared<Tensor>(mem);
+        }
+    }
+
+    //reorder
+    auto mem = std::make_shared<Memory>(get_engine(), current_ext_desc);
+    mem->load(*(prime_mem()));
+    return std::make_shared<Tensor>(mem);
+}
```
ceciliapeng2011, line 91: Do you have a unit test that triggers this precision conversion?

I checked that the existing PR implementation will always have the same external_desc and internal_desc, as both them comes from getOriginalOutputPrecisionAtPort(0). Is it what we expect? 
I have a [comment](https://github.com/openvinotoolkit/openvino/pull/20404#discussion_r1395091262) and try to figure out if there any way to let a user declare his wanted precision to a state. It is useful we use half-precision (FP16) kvcaches for platforms of avx2/avx512.

src/plugins/intel_cpu/tests/functional/subgraph_tests/src/stateful_model.cpp
maxnick: Yep. With no doubts it will do that. Thus, depending on which PR is merged earlier, some changes need to be introduced.

src/plugins/intel_cpu/src/memory_state.cpp
maxnick: We don't. I think we may do it in a separate PR since this one is necessary to unblock some other activities.

luo-cheng2021 (approved):

ceciliapeng2011 (approved):

dmitry-gorokhov (approved):

PR #20371, opened 2023-10-10
Partial GPTQ int4 models conversion support

slyalin:


src/frontends/pytorch/src/frontend.cpp
```diff
@@ -135,8 +136,17 @@ std::shared_ptr<Model> FrontEnd::convert(const ov::frontend::InputModel::Ptr& mo
...
     FRONT_END_OP_CONVERSION_CHECK(is_conversion_successful, pack_detailed_failure_report(unconverted_ops, norm_err));
+    #else
+    if(!is_conversion_successful) {
+        std::cerr << "*******************************************************************************\n";
+        std::cerr << "[ DEBUG ] Model cannot be converted completely by PyTorch FE.\n[ DEBUG ] Debug mode is active -- leaving model as-is with unsupported operations. Compilation of the model will fail.\n";
+        std::cerr << pack_detailed_failure_report(unconverted_ops, norm_err) << "\n";
+        std::cerr << "*******************************************************************************\n";
+    }
     return converted_model;
+    #endif
```
slyalin, line 149: Revert.

src/frontends/pytorch/src/op/cat.cpp
```diff
@@ -20,6 +20,7 @@ namespace op {
 
 using namespace ov::op;
 
+
```
slyalin, line 23:
```suggestion
```

src/frontends/pytorch/src/transforms/aten_stack_list_construct_replacer.cpp
```diff
@@ -38,22 +41,28 @@ AtenStackListConstructReplacer::AtenStackListConstructReplacer() {
...
             const auto& list_inputs = list_construct_node->input_values();
-            OutputVector node_vector;
-            auto zero = v0::Constant::create(element::i32, Shape{}, {0});
-            // Iterate over values in ListConstruct
-            for (const auto& list_input : list_inputs) {
-                auto node = concat_list_construct(list_input);
-                auto unsqueezed_node = std::make_shared<v0::Unsqueeze>(node, axis_const);
-                node_vector.push_back(unsqueezed_node);
+            std::shared_ptr<Node> node;
+            if (auto compression = u4_compression_stack(nullptr, std::deque<Output<Node>>(list_inputs.begin(), list_inputs.end()), axis[0])) {
```
slyalin, line 49: `u4_compressoin_stack` is used in this place only, no need to use `deque` anymore.

src/frontends/pytorch/src/utils_quantize.cpp
```diff
@@ -168,6 +169,57 @@ std::shared_ptr<QuantizedPtNode> cast_quantized_fw_node(std::shared_ptr<Node> no
...
+    if (list_elems.size() != 2)
+        return nullptr;
+    auto bitwise_and = cast_fw_node(list_elems[0].get_node_shared_ptr(), "aten::bitwise_and");
+    if (!bitwise_and)
+        return nullptr;
+    auto bitwise_shift = cast_fw_node(list_elems[1].get_node_shared_ptr(), "aten::bitwise_right_shift");
+    if (!bitwise_shift)
+        return nullptr;
+
+    // TODO: Use context (if not nullptr) to mark newly created nodes
```
slyalin, line 186: The result of this compression is a single u4 Constant, it can be marked outside of this function, `context` is not required to be passed.

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
```diff
@@ -0,0 +1,95 @@
...
+using namespace ov::op;
+using namespace ov::pass::pattern;
+
+U4BlockRepack::U4BlockRepack() {
+    const auto m_constant = ov::pass::pattern::wrap_type<ov::op::v0::Constant>();
+    const auto m_reshape1 = ov::pass::pattern::wrap_type<ov::op::v1::Reshape>({m_constant, any_input()});
+    const auto m_transpose = ov::pass::pattern::wrap_type<ov::op::v1::Transpose>({m_reshape1, any_input()});
+    const auto m_reshape2 = ov::pass::pattern::wrap_type<ov::op::v1::Reshape>({m_transpose, any_input()});
+
+    auto pack_byte = [](uint8_t lo, uint8_t hi) { return (lo << 4) | hi; };  // swap halfs because Convert op assumes this layout
```
AlexKoff88, line 32: I think it is safer to do the following: `return (lo << 4) | (hi && 0x0F)`. This should be changed when the right order is implemented.

src/bindings/python/src/openvino/frontend/pytorch/ts_decoder.py
```diff
@@ -85,7 +86,12 @@ def _get_scripted_model(self, pt_module, example_inputs=None, skip_freeze=False)
...
                 input_parameters, input_signature, pt_module, self._input_is_list = prepare_example_inputs_and_model(example_inputs, input_params, pt_module)
-                scripted = torch.jit.trace(pt_module, **input_parameters, strict=False)
+                try:
+                    gptq.patch_model(pt_module)
+                    scripted = torch.jit.trace(
+                        pt_module, **input_parameters, strict=False)
+                except:
+                    gptq.unpatch_model(pt_module)
             if not skip_freeze:
                 for n in scripted.inlined_graph.nodes():
```
AlexKoff88, line 96: `scripted` variable is no longer accessible here if you get an exception in the logic above.

src/bindings/python/src/openvino/frontend/pytorch/gptq.py
```diff
@@ -0,0 +1,120 @@
...
+    mask = torch.tensor(15, dtype=torch.uint8).to(weights.device)
+    return torch.stack((torch.bitwise_and(weights, mask), torch.bitwise_right_shift(weights, 4)), dim=-1)
+
+
+def patched_forward(self, *args, **kwargs):
+    if hasattr(self, "_hf_hook"):
+        args, kwargs = self._hf_hook.pre_forward(self, *args, **kwargs)
+
+    dtype = x.dtype
+    x = args[0]
```
AlexKoff88, line 34:
```suggestion
    x = args[0]
    dtype = x.dtype
```

slyalin: A bit of trashed commits, sorry for that.

src/bindings/python/src/openvino/frontend/pytorch/gptq.py
```diff
@@ -0,0 +1,120 @@
...
+            # already patched, skipping
+            continue
+        # TODO: Check module type
+        if hasattr(m, "QUANT_TYPE") and m.QUANT_TYPE == "exllama":
+            if m.bits != 4:
+                raise ValueError(
+                    f"Exllama kernel supports only bits=4, requested bits={m.bits} for module {name}. Something is wrong in the model initialization.")
+
+            int4_in_int32 = 8
+            groups = m.qzeros.shape[0]
```
AlexKoff88, line 75:
```suggestion
            groups = m.qzeros.shape[0]
            m.width = m.qweight.shape[1]
```

src/bindings/python/src/openvino/frontend/pytorch/gptq.py
```diff
@@ -0,0 +1,120 @@
...
+        out = self._hf_hook.post_forward(self, out)
+    return out
+
+
+def patch_model(model):
+    for name, m in model.named_modules():
+        if hasattr(m, "_openvino_patch_orig_forward"):
+            # already patched, skipping
+            continue
+        # TODO: Check module type
```
AlexKoff88, line 68:
```suggestion
        # TODO: Check module type
        is_quantized = getattr(m, "is_quantized", None)
        if not is_quantized is None:
            m.is_quantized = False
        m.float()
```

AlexKoff88: This is required to make it work w/o GPU.

src/bindings/python/src/openvino/frontend/pytorch/gptq.py
slyalin: Why does it work without this change? Other models don't have correct `width` defined?

src/frontends/pytorch/src/utils_quantize.cpp
```diff
@@ -168,6 +170,53 @@ std::shared_ptr<QuantizedPtNode> cast_quantized_fw_node(std::shared_ptr<Node> no
...
+    auto u8_shape = weights_u8->get_shape();
+    size_t full_size = shape_size(u8_shape);
+    auto src = weights_u8->get_data_ptr<uint8_t>();
+
+    auto u4_shape = u8_shape;
+    u4_shape.push_back(2);
+    auto new_const = std::make_shared<v0::Constant>(element::u4, u4_shape);
+    auto dst = const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(new_const->get_data_ptr()));
+
+    std::copy(src, src + full_size, dst);  // TODO: Avoid copying, reuse the same constant
```
slyalin, line 215: @ilyachur, @jane-intel, please suggest better way to reinterpret content of one constant as a new constant with different shape and data type.

src/frontends/pytorch/src/transforms/aten_stack_list_construct_replacer.cpp
```diff
@@ -38,22 +39,28 @@ AtenStackListConstructReplacer::AtenStackListConstructReplacer() {
         auto axis_node = pattern_map.at(axis).get_node_shared_ptr();
         auto axis_const = std::dynamic_pointer_cast<v0::Constant>(axis_node);
         auto axis = axis_const->cast_vector<int64_t>();
+        OPENVINO_ASSERT(axis.size() == 1);
```
jane-intel, line 42: nit: move check to pattern

tools/mo/openvino/tools/mo/ops/Cast.py
```diff
@@ -83,10 +89,12 @@ def custom_type_casting_and_packing(node: Node, value, dst_type):
...
-        bit_order_little = (padded[:, None] & (1 << np.arange(num_bits)) > 0).astype(np.uint8)
-        bit_order_big = np.flip(bit_order_little, axis=1)
-        bit_order_big_flattened = bit_order_big.flatten()
-        packed = np.packbits(bit_order_big_flattened)
+        bit_order_little = (padded[:, None] & (
+            1 << np.arange(num_bits)) > 0).astype(np.uint8)
+        bit_order_big_flattened = bit_order_little.flatten()
+        # u1 still has reversed bit order:
+        packed = np.packbits(bit_order_big_flattened,
+                             bitorder='little' if num_bits > 1 else 'big')
```
jane-intel, line 97: num_bits always == 4 for this function

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
```diff
@@ -0,0 +1,96 @@
...
+        return (hi << 4) | (lo & 0x0F);
+    };  // swap halfs because Convert op assumes this layout
+
+    auto get_u4 = [](const uint8_t* src, size_t idx) {
+        const size_t byte_idx = idx / 2;
+        const uint8_t bit_shift = 4 * (idx % 2);
+        return (src[byte_idx] >> bit_shift) & 0xF;
+    };
+
+    register_matcher(
```
jane-intel, line 40: why don't you use same style for pattern matching?
```
ov::matcher_pass_callback callback = [=](Matcher& m) {
        ...
};
auto m = std::make_shared<Matcher>(m_reshape2, "ov::frontend::pytorch::pass::U4BlockRepack");
```

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
```diff
@@ -0,0 +1,96 @@
...
+        std::make_shared<ov::pass::pattern::Matcher>(m_reshape2, "ov::frontend::pytorch::pass::U4BlockRepack"),
+        [=](ov::pass::pattern::Matcher& m) {
+            auto& pattern_to_output = m.get_pattern_value_map();
+            auto constant =
+                std::dynamic_pointer_cast<ov::op::v0::Constant>(pattern_to_output[m_constant].get_node_shared_ptr());
```
jane-intel, line 45:
```suggestion
            auto constant =
                std::dynamic_pointer_cast<ov::op::v0::Constant>(pattern_to_output[m_constant].get_node_shared_ptr());
            if (!constant)
                return false;
```

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
```diff
@@ -0,0 +1,96 @@
...
+            auto reshape1 = pattern_to_output[m_reshape1].get_node_shared_ptr();
+            auto transpose = pattern_to_output[m_transpose].get_node_shared_ptr();
+            auto reshape2 = pattern_to_output[m_reshape2].get_node_shared_ptr();
+
+            if (constant->get_element_type() != element::u4)
+                return false;
```
jane-intel, line 51: You can check it in the pattern

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+            // FIXME: Check reshape/transpose/reshape target shapes and axes permutation; now they are supposed to be
+            // always in expected form
+
+            auto source_shape = reshape1->get_output_shape(0);
+
+            if (source_shape.size() != 3)
+                return false;
```
jane-intel, line 59: Please check in the pattern -- it makes transformation code less cluttered.

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+            auto destination_shape = reshape2->get_output_shape(0);
```
jane-intel, line 61: please put checks in place for output shapes to be static. put them in the pattern.

jane-intel (commented):
@ilyachur could you share BKMs on how to make copy-less creation of Constant of other type and shape?

src/bindings/python/src/openvino/frontend/pytorch/ts_decoder.py
```diff
@@ -84,8 +85,24 @@ def _get_scripted_model(self, pt_module, example_inputs=None, skip_freeze=False)
             if example_inputs is None:
                 scripted = torch.jit.script(pt_module)
             else:
-                input_parameters, input_signature, pt_module, self._input_is_list = prepare_example_inputs_and_model(example_inputs, input_params, pt_module)
-                scripted = torch.jit.trace(pt_module, **input_parameters, strict=False)
+                input_parameters, input_signature, pt_module, self._input_is_list = prepare_example_inputs_and_model(
+                    example_inputs, input_params, pt_module)
+                gptq_patched = False
+
```
AlexKoff88, line 91: I would recommend the following check before trying GPTQ patching:
`if getattr(model.config, "quantization_config", None) and model.config.quantization_config.quant_method == "gptq":`

src/frontends/pytorch/src/transforms/aten_stack_list_construct_replacer.cpp
slyalin: I don't want to suppress the error message in this case.

tools/mo/openvino/tools/mo/ops/Cast.py
slyalin: Good point, but the code around was written in a way that `num_bits` was treated as a parameter (look at `np.arange(num_bits)`) so I have just left the same "portable" style.

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
slyalin: It wasn't my invention, have copied and pasted from another transformation. Is it worse? Looks shorter.

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
slyalin: It is hard to remember how to do that. Good code snippets?

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
slyalin: Please make a code suggestion.

src/bindings/python/src/openvino/frontend/pytorch/gptq.py
```diff
@@ -0,0 +1,125 @@
...
+        is_quantized = getattr(m, "is_quantized", None)
+        if not is_quantized is None:
```
mvafin, line 70:
```suggestion
        if is_quantized is not None:
```

src/bindings/python/src/openvino/frontend/pytorch/gptq.py
```diff
@@ -0,0 +1,125 @@
...
+            m.is_quantized = False
+        m.float()
```
mvafin, line 72: Why we do that for each module? We will change every model this way won't we?

src/frontends/pytorch/src/frontend.cpp
```diff
@@ -182,6 +183,8 @@ void FrontEnd::normalize(const std::shared_ptr<ov::Model>& model) const {
     manager.register_pass<ov::frontend::pytorch::pass::AtenCatToConcat>();
     manager.register_pass<ov::frontend::pytorch::pass::AppendListUnpackReplacer>();
     manager.register_pass<ov::frontend::pytorch::pass::AtenStackListConstructReplacer>();
+    manager.register_pass<ov::pass::MarkDequantizationSubgraph>(
```
mvafin, line 186: Why do we need this transformation here 2nd time? We already have it on line 177

src/frontends/pytorch/src/transforms/aten_stack_list_construct_replacer.cpp
mvafin: I don't like to throw exception in transformation pipeline. Yes we have something wrong here, but better just return false in that case and add message to fw_node using `add_exception_to_fw_node` helper

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include "u4_block_repack.hpp"
+
+#include "openvino/core/rt_info.hpp"
+#include "openvino/op/constant.hpp"
+#include "openvino/op/reshape.hpp"
+#include "openvino/op/transpose.hpp"
+#include "openvino/pass/pattern/matcher.hpp"
+#include "openvino/pass/pattern/op/wrap_type.hpp"
+#include "utils.hpp"
+#include "utils_quantize.hpp"
```
mvafin, line 14: I don't see usage of `u4_compression_stack` in this file. Do you really need this include?

src/frontends/pytorch/src/transforms/aten_stack_list_construct_replacer.cpp
```diff
@@ -38,22 +39,28 @@ AtenStackListConstructReplacer::AtenStackListConstructReplacer() {
...
             const auto& list_inputs = list_construct_node->input_values();
-            OutputVector node_vector;
-            auto zero = v0::Constant::create(element::i32, Shape{}, {0});
-            // Iterate over values in ListConstruct
-            for (const auto& list_input : list_inputs) {
-                auto node = concat_list_construct(list_input);
-                auto unsqueezed_node = std::make_shared<v0::Unsqueeze>(node, axis_const);
-                node_vector.push_back(unsqueezed_node);
+            std::shared_ptr<Node> node;
+            if (auto compression = u4_compression_stack(list_inputs, axis[0])) {
```
mvafin, line 47: You use this function only in this file as I can see. Could you have it in this file instead of adding to `utils_quantize`?

mvafin:
I would like to see a test with int4 model as input even if it will not work without cuda GPU.

ilyachur (approved):
Core part LGTM

src/bindings/python/src/openvino/frontend/pytorch/gptq.py
AlexKoff88: This is required to export the model without GPU on board. Otherwise, we are getting errors such as "layer_norm does not support float16 on CPU" and similar.

jane-intel (approved):
Core part + MO LGTM

src/frontends/pytorch/src/frontend.cpp
slyalin: We have produced new sub-graph with constant->convert in `AtenStackListConstructReplacer` that should be preserved.

mvafin: But it needs to be protected only against `ConstantFolding` and next one will be called in moc transformations which also have `MarkDequantizationSubgraph` transformation before `ConstantFolding`, so it should be safe to remove it from here.

src/bindings/python/src/openvino/frontend/pytorch/gptq.py
mvafin: at least we need to check that module is float16 before doing that, right? Also this code is called for every model, not only u4 one

mvafin (changes requested):
Please use patching only for u4/i4 models

src/frontends/pytorch/src/transforms/u4_block_repack.cpp
slyalin: Decided to leave as-is.

src/frontends/pytorch/src/transforms/aten_stack_list_construct_replacer.cpp
slyalin: Decided to leave as-is.

src/bindings/python/src/openvino/frontend/pytorch/ts_decoder.py
```diff
@@ -84,8 +85,27 @@ def _get_scripted_model(self, pt_module, example_inputs=None, skip_freeze=False)
...
+                    try:
+                        gptq.patch_model(pt_module)
+                        gptq_patched = True
+                    except Exception as error:
+                        print('[ WARNING ] Failed patching of AutoGPTQ model. Error message:\n', error)
+                        print('[ WARNING ] Tracing of the model will likely be unsuccesfull or incorrect')
+                        gptq.unpatch_model(pt_module)
+                        gptq_patched = False
+
+                try:
```
mvafin, line 102: Do you need try-finally here? If trace fails we don't have any recovery from it

slyalin: We need to un-patch the model anyway regardless of success of the tracing.

mvafin: Then you should rethrow an exception

mvafin (approved):
Looks good from pytorch frontend point of view

AlexKoff88 (approved):

src/bindings/python/src/openvino/frontend/pytorch/ts_decoder.py
slyalin: Not catching it in `finally`, so no need to re-raise -- it floats up automatically.

slyalin:
@ilya-lavrenov, could we merge this PR with part of the checks still pending due to partially not functional CI? Core changes checks passed 2 days ago, and since then we have done changes in the PyTorch part only, which is validated for the most recent commit. The remaining part of the jobs that are pending shouldn't bring much value in addition to what we already have covered.

src/bindings/python/src/openvino/frontend/pytorch/gptq.py
```diff
@@ -0,0 +1,140 @@
...
+def patch_model(model):
+    for name, m in model.named_modules():
+        if hasattr(m, '_openvino_patch_orig_forward'):
+            # already patched, skipping
+            continue
+        # TODO: Check module type
+        is_quantized = getattr(m, 'is_quantized', None)
+        if is_quantized is not None:
+            m.is_quantized = False
+        m.float()  # enables tracing on CPU, applied for all modules
```
slyalin, line 76: Leftover noticed by @mvafin: should we revert the effect of this step during unpatching?

tools/mo/openvino/tools/mo/ops/Cast.py
```diff
@@ -63,6 +65,10 @@ def custom_type_casting_and_packing(node: Node, value, dst_type):
             we would pad them to 6 element with the last element as zero and we would pack them into 3 uint8 values
         """
         assert dst_type in [packed_U4, packed_I4]
+        # TODO: Remove this comment when it's clear that we can fix it easily
+        # raise Exception("Packing of u4/i4 data is no longer supported in mo because it is now incompatible with the new "
+        #                 "order of the halfs of a byte that was introduced in OpenVINO runtime recently. Use ovc "
+        #                 "command line tool or openvino.convert_model python function instead.")
```
slyalin, line 71: Remove.

PR #20380, opened 2023-10-11
TorchFX: Constant value pass without copy

cavusmustafa:
Creating Constant node by passing numpy array instead of copying data


src/bindings/python/src/openvino/frontend/pytorch/fx_decoder.py
```diff
@@ -224,11 +223,7 @@ def as_constant(self):
         if self.pt_module.op == 'get_attr':
             # Extract Constant from FX module field
             ret = fetch_attr(self.fx_gm, self.pt_module.target)
-            ovshape = PartialShape(ret.size())
-            ovtype = pt_to_ov_type_map[str(ret.type())]
-            c_type = ctypes.POINTER(ov_to_c_type_map[ovtype])
-            data_c_ptr = ctypes.cast(ret.data_ptr(), c_type)
-            ov_const = op.Constant(ovtype, ovshape.get_shape(), data_c_ptr[:ret.nelement()])
+            ov_const = make_constant(ret.numpy())
```
jiwaszki, line 226:
```suggestion
            ov_const = op.Constant(ret.numpy(), shared_memory=True)
```
Is it really improving performance with claimed "without copy"?

BTW I think that `make_constant` is absolutely redundant function. It only creates overhead of passing arguments around and calling function frames which can only build up in the runtime.

jiwaszki (changes requested):

src/bindings/python/src/openvino/frontend/pytorch/fx_decoder.py
jiwaszki: If you want to use function, there is a better one which do the trick:
https://github.com/cavusmustafa/openvino/blob/3736fcaa121c045d9ad45aaa1ef4c873cf45e77f/src/bindings/python/src/openvino/frontend/pytorch/utils.py#L61-L72

cavusmustafa: Thanks for the suggestion. I added passing shared_memory=True.

For make_constant, it is used in multiple locations in fx_decoder currently. We will review and raise another PR to completely remove it if it is not needed.

mvafin (approved):

PR #20398, opened 2023-10-11
[TESTS] Check the shape of output in OVInferRequestDynamicTests::checkOutput

briangrifiin:
### Details:
 - Check that the shape of the output matches the reference value
 - This allows to catch cases when an output has the same number of elements as the reference tensor, but their shapes differ

### Tickets:
 - [CVS-122578](https://jira.devtools.intel.com/browse/CVS-122578)


ilyachur:
@iefode Please take a look

iefode (approved):

PR #20399, opened 2023-10-11
[GPU] skip excessive mem alloc request in build

wilson-seok:
### Details:
 - skip excessive mem alloc request in build

### Tickets:
 - 84485

### DIFF
--- a/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
@@ -543,10 +543,9 @@ private:
         }
 
         auto& e = typ_node.get_program().get_engine();
-        auto max_mem = e.get_max_memory_size();
-        auto required_mem = typ_node.get_output_layout().bytes_count();
+        auto output_layout = typ_node.get_output_layout();
         // When dynamic shape node has huge upper boundary which causes bigger mem size than system max mem size, return false.
-        if (required_mem >= max_mem) {
+        if (!e.check_allocatable(output_layout, e.get_lockable_preferred_memory_allocation_type(output_layout.format.is_image_2d()))) {
             return false;
         }
 
--- a/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
@@ -541,14 +541,6 @@ private:
             typ_node.get_users().front()->can_be_optimized()) {  // check if the only user is concat
             return false;
         }
-
-        auto& e = typ_node.get_program().get_engine();
-        auto output_layout = typ_node.get_output_layout();
-        // When dynamic shape node has huge upper boundary which causes bigger mem size than system max mem size, return false.
-        if (!e.check_allocatable(output_layout, e.get_lockable_preferred_memory_allocation_type(output_layout.format.is_image_2d()))) {
-            return false;
-        }
-
         return true;
     }
 };
--- a/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
+++ b/src/plugins/intel_gpu/src/graph/include/primitive_inst.h
@@ -541,6 +541,15 @@ private:
             typ_node.get_users().front()->can_be_optimized()) {  // check if the only user is concat
             return false;
         }
+
+        auto& e = typ_node.get_program().get_engine();
+        auto max_mem = e.get_max_memory_size();
+        auto required_mem = typ_node.get_output_layout().bytes_count();
+        // When dynamic shape node has huge upper boundary which causes bigger mem size than system max mem size, return false.
+        if (required_mem >= max_mem) {
+            return false;
+        }
+
         return true;
     }
 };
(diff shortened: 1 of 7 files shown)


src/plugins/intel_gpu/src/graph/include/primitive_inst.h
```diff
@@ -541,6 +541,15 @@ class typed_primitive_inst_base : public primitive_inst {
             typ_node.get_users().front()->can_be_optimized()) {  // check if the only user is concat
             return false;
         }
+
+        auto& e = typ_node.get_program().get_engine();
+        auto max_mem = e.get_max_memory_size();
```
vladimir-paramuzov, line 546: Wouldn't checking of max **used** memory be better here?

yeonbok: I think we can do like this 
- fix ocl_engine::check_allocatable not to throw exception but return false if not allocatable 
- add the required assertion in ocl_engine::allocate_memory() like OPENVINO_ASSERT(check_allocatable(layout,type), true)
- use check_allocatable here

yeonbok: Though basically I think chcking used memory here for dynamic shape is not that much meaningful because the filtering here is to avoid too much memory for bounded shape while actual exception will be happen eventually in the allocate_output function.

wilson-seok: Updated! Thanks!

src/plugins/intel_gpu/src/runtime/ocl/ocl_engine.cpp
```diff
@@ -125,29 +125,26 @@ allocation_type ocl_engine::detect_usm_allocation_type(const void* memory) const
...
-            "[GPU] Exceeded max size of memory allocation: ",
-            "Required ", layout.bytes_count(), " bytes, already occupied : ", used_mem, " bytes, ",
-            "but available memory size is ", get_max_memory_size(), " bytes");
-#else
+
     if (layout.bytes_count() + used_mem > get_max_memory_size()) {
         GPU_DEBUG_COUT << "[Warning] [GPU] Exceeded max size of memory allocation: " << "Required " << layout.bytes_count() << " bytes, already occupied : "
                        << used_mem << " bytes, but available memory size is " << get_max_memory_size() << " bytes" << std::endl;
         GPU_DEBUG_COUT << "Please note that performance might drop due to memory swap." << std::endl;
+        return false;
```
yeonbok, line 136: This is changing behavior. 
In linux it is not warning but error. Also in windows, it should not return false.

wilson-seok: Updated! Thanks!

src/plugins/intel_gpu/src/runtime/ocl/ocl_engine.cpp
```diff
@@ -125,29 +125,29 @@ allocation_type ocl_engine::detect_usm_allocation_type(const void* memory) const
...
-    // Prevent from being killed by Ooo Killer of Linux
-    OPENVINO_ASSERT(layout.bytes_count() + used_mem <= get_max_memory_size(),
-            "[GPU] Exceeded max size of memory allocation: ",
-            "Required ", layout.bytes_count(), " bytes, already occupied : ", used_mem, " bytes, ",
-            "but available memory size is ", get_max_memory_size(), " bytes");
-#else
+
     if (layout.bytes_count() + used_mem > get_max_memory_size()) {
-        GPU_DEBUG_COUT << "[Warning] [GPU] Exceeded max size of memory allocation: " << "Required " << layout.bytes_count() << " bytes, already occupied : "
+        GPU_DEBUG_COUT << "[GPU] Exceeded max size of memory allocation: " << "Required " << layout.bytes_count() << " bytes, already occupied : "
```
yeonbok, line 133: This print is shown for both  unix and windows. 
e.g., if unix, first this message is printed and then printed again at the exception throwing. 
Please let the message to be shown only once. 
linux : as a exception 
windows: as a warning

wilson-seok: Updated. Thanks!

src/plugins/intel_gpu/src/graph/primitive_inst.cpp
```diff
@@ -1012,7 +1012,14 @@ primitive_inst::primitive_inst(network& network, program_node const& node, bool
                 if (user->is_type<mutable_data>())
                     _outputs[0] = user->as<mutable_data>().get_attached_memory_ptr();
         } else {
-            _outputs = allocate_outputs();
+            auto& e = network.get_engine();
+            auto output_layout = node.get_output_layout();
+            // When dynamic shape node has huge upper boundary which causes bigger mem size than system max allocable mem size, return false.
+            if (!e.check_allocatable(output_layout, e.get_lockable_preferred_memory_allocation_type(output_layout.format.is_image_2d()))) {
```
yeonbok, line 1018: 1. I think you can just pass usm_host to this check. (actual alloc type will be checked again in allocate_output but here, we just filter out the tensors which cannot be allocated even for host) 
2. Also we need to do this check only for dynamic shape, because static shape check will be done in allocate_output. 
3. So my suggestion is
-  // not to set _mem_allocated in the constructors initializer (get rid of line 978) 
- if (node.is_dynamic() && (!check_allocatable(output_layout, usm_host)) 
        allocate_memory = false;
   _mem_allocated = allocate_memory;
   if (allocate_memory) {
        // original code
   }

wilson-seok: Updated! Thanks!

yeonbok (approved):

ahnyoung-paul (approved):

PR #20404, opened 2023-10-11
Support dynamic shapes in ReadValue/Assign ops and MakeStateful transformation

itikhono:
Support dynamic shapes in ReadValue/Assign ops and MakeStateful transformation

**Changes in ov operations:** TBA
ReadValue op: 
Assign op:

**Changes in python API:** TBA
ReadValue op
Workaround
Fixed

**Changes in tests:** TBA

**Changes in ov transformations:** TBA

image

### Tickets:
 - *CVS-111978*


src/core/src/op/assign.cpp
```diff
@@ -66,9 +66,28 @@ Assign::Assign(const Output<Node>& new_value, const std::shared_ptr<util::Variab
 
 void Assign::validate_and_infer_types() {
     OV_OP_SCOPE(v6_Assign_validate_and_infer_types);
-    m_variable->update({get_input_partial_shape(0), get_input_element_type(0), m_variable->get_info().variable_id});
+    auto variable_info = m_variable->get_info();
```
itikhono, line 69: move it to a separate shape  inference function

src/core/tests/type_prop/assign.cpp
```diff
@@ -8,10 +8,11 @@
 
 #include "common_test_utils/type_prop.hpp"
 #include "openvino/core/model.hpp"
-#include "openvino/op/read_value.hpp"
+#include "openvino/op/assign.hpp"
```
itikhono, line 11: todo: revert

src/core/include/openvino/op/read_value.hpp
```diff
@@ -48,6 +48,12 @@ class OPENVINO_API ReadValue : public util::ReadValueBase {
     OPENVINO_OP("ReadValue", "opset6", util::ReadValueBase);
     ReadValue() = default;
 
+    /// \brief Constructs a ReadValue operation.
+    ///
+    /// \param variable Class for storing and synchronizing element types, shapes and
+    /// identifiers between pairs of Assign/ReadValue nodes.
+    ReadValue(const std::shared_ptr<util::Variable>& variable);
```
ilya-lavrenov, line 55:
```suggestion
    explicit ReadValue(const std::shared_ptr<util::Variable>& variable);
```

itikhono: done

src/core/src/op/assign.cpp
```diff
@@ -66,9 +66,26 @@ Assign::Assign(const Output<Node>& new_value, const std::shared_ptr<util::Variab
...
+    auto variable_type = variable_info.data_type;
+    auto variable_shape = variable_info.data_shape;
+
+    auto input_type = get_input_element_type(0);
+    auto input_shape = get_input_partial_shape(0);
+    bool compatible_type = variable_type.is_dynamic() || input_type == variable_type;
+    bool compatible_shape = input_shape.rank().compatible(variable_shape.rank());
```
slyalin, line 76: Something wrong here...

src/core/src/op/assign.cpp
```diff
@@ -66,9 +66,26 @@ Assign::Assign(const Output<Node>& new_value, const std::shared_ptr<util::Variab
...
+
+    if (compatible_shape && input_shape.rank().is_static() && variable_shape.rank().is_static()) {
+        OPENVINO_ASSERT(input_shape.rank().get_length() == variable_shape.rank().get_length(),
+                        "Ranks of initial_shape and variable_shape do not match.");
+        for (int64_t i = 0; i < variable_shape.rank().get_length(); ++i) {
+            compatible_shape = compatible_shape && variable_shape[i].compatible(input_shape[i]);
```
slyalin, line 82: Something wrong in the same way here...

src/core/src/op/assign.cpp
```diff
@@ -66,9 +66,26 @@ Assign::Assign(const Output<Node>& new_value, const std::shared_ptr<util::Variab
...
+        }
+    }
 
-    set_output_type(0, get_input_element_type(0), get_input_partial_shape(0));
+    OPENVINO_ASSERT(compatible_shape, "The shape specified in the Variable is not compatible with the input shape.");
+    OPENVINO_ASSERT(compatible_type, "The type specified in the Variable doesn't match the type of the input.");
```
slyalin, line 87: Please provide more information: print shapes, node/variable names etc.

src/bindings/python/src/openvino/runtime/opset13/__init__.py
```diff
@@ -129,7 +129,7 @@
 from openvino.runtime.opset4.ops import range
 from openvino.runtime.opset8.ops import random_uniform
 from openvino.runtime.opset9.ops import rdft
-from openvino.runtime.opset3.ops import read_value
+from openvino.runtime.opset6.ops import read_value
```
slyalin, line 132: Good catch.

src/core/src/op/read_value.cpp
```diff
@@ -50,19 +55,40 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+    if (get_input_size() > 0) {
+        const auto initial_type = get_input_element_type(0);
+        const auto& initial_shape = get_input_partial_shape(0);
+
+        // Variable shape/type determines a permissible range of values for shape/type inferred from initial_subgraph.
+        // If initial_subgraph is set, then we need to check that shape/type inferred from initial_subgraph
+        // is within the permissible range.
+
+        bool compatible_type = variable_type.is_dynamic() || initial_type == variable_type;
+        bool compatible_shape = variable_shape.rank().relaxes(initial_shape.rank());
```
slyalin, line 74: Also check this one.

slyalin (changes requested):
Please check if all cases for introduced set operations for types/shapes are really covered by tests and check if they are really correct (discussed during offline code review).

src/core/src/op/assign.cpp
itikhono: done

src/core/src/op/assign.cpp
itikhono: done

src/core/src/op/assign.cpp
itikhono: done

src/core/src/op/read_value.cpp
itikhono: done

src/core/src/op/read_value.cpp
```diff
@@ -50,19 +55,40 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+                compatible_shape = compatible_shape && variable_shape[i].relaxes(initial_shape[i]);
+            }
+        }
+        OPENVINO_ASSERT(compatible_shape,
+                        "The shape specified in the Variable have to extend (relax) the shape "
+                        "inferred from the initializing subgraph.");
+        OPENVINO_ASSERT(compatible_type,
+                        "The type specified in the Variable have to extend (relax) the type "
+                        "inferred from the initializing subgraph.");
+    }
```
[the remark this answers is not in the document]
itikhono: fixed

src/core/src/op/read_value.cpp
```diff
@@ -50,36 +55,70 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+
+    set_output_type(0, variable_type, variable_shape);
 }
 
 shared_ptr<Node> op::v6::ReadValue::clone_with_new_inputs(const OutputVector& new_args) const {
     OV_OP_SCOPE(v6_ReadValue_clone_with_new_inputs);
     check_new_args_count(this, new_args);
-    return make_shared<ReadValue>(new_args.at(0), m_variable);
+    if (new_args.size() == 1) {
```
praasz, line 98: Consider use similar pattern like in Squeeze operator
https://github.com/openvinotoolkit/openvino/blob/9e97d7ef1ad30200c5ca404697a081e4d413a65e/src/core/src/op/squeeze.cpp#L52

src/core/src/pass/low_latency.cpp
```diff
@@ -93,13 +93,13 @@ std::shared_ptr<ov::opset9::Assign> replace_with_memory(const ov::Input<ov::Node
...
-    VariableInfo var_info{ov::PartialShape::dynamic(), ov::element::dynamic, variable_name};
-    auto variable = std::make_shared<Variable>(var_info);
-
     ov::Output<ov::Node> read_value_in = input.get_source_output();
     if (use_const_initializer) {
         read_value_in = create_init_subgraph(read_value_in, to);
     }
+
+    VariableInfo var_info{read_value_in.get_partial_shape(), read_value_in.get_element_type(), variable_name};
+    auto variable = std::make_shared<Variable>(var_info);
```
praasz, line 102:
```suggestion
    auto variable = std::make_shared<Variable>(read_value_in.get_partial_shape(), read_value_in.get_element_type(), variable_name);
```
the `std::make_shared` should use ctor and make dynamic object in-place.

src/bindings/python/src/pyopenvino/graph/dict_attribute_visitor.cpp
```diff
@@ -109,8 +109,14 @@ void util::DictAttributeDeserializer::on_adapter(const std::string& name, ov::Va
                     ov::op::util::VariableInfo{ov::PartialShape::dynamic(), ov::element::dynamic, variable_id});
             }
             a->set(m_variables[variable_id]);
+        } else if (const auto& a = ov::as_type<ov::AttributeAdapter<ov::PartialShape>>(&adapter)) {
+            if (py::isinstance<ov::PartialShape>(m_attributes[name.c_str()])) {
+                a->set(m_attributes[name.c_str()].cast<ov::PartialShape>());
+            } else {
+                OPENVINO_ASSERT(false, "No AttributeVisitor support for accessing attribute named: ", name);
```
praasz, line 116:
```suggestion
                OPENVINO_THROW("No AttributeVisitor support for accessing attribute named: ", name);
```

src/bindings/python/src/pyopenvino/graph/dict_attribute_visitor.cpp
```diff
@@ -109,8 +109,14 @@ void util::DictAttributeDeserializer::on_adapter(const std::string& name, ov::Va
...
+            }
         } else {
-            NGRAPH_CHECK(false, "No AttributeVisitor support for accessing attribute named: ", name);
+            OPENVINO_ASSERT(false, "No AttributeVisitor support for accessing attribute named: ", name);
```
praasz, line 119:
```suggestion
            OPENVINO_THROW("No AttributeVisitor support for accessing attribute named: ", name);
```

src/bindings/python/src/pyopenvino/graph/model.cpp
```diff
@@ -60,6 +60,23 @@ static std::vector<std::shared_ptr<ov::Node>> cast_to_node_vector(const ov::Sink
     return nodes;
 }
 
+// Assign operations created via Assign py binding have Variables which are not connected to
+// ReadValue operations. This function attempts to resolve this situation by finding correct Variables
+// for Assigns.
+static void set_correct_variables_for_assign_ops(const std::shared_ptr<ov::Model>& model, const ov::SinkVector& sinks) {
+    auto variables = model->get_variables();
```
praasz, line 67:
```suggestion
    const auto& variables = model->get_variables();
```
Or copy of this vector is requried?

src/core/tests/type_prop/assign.cpp
```diff
@@ -78,3 +79,99 @@ TEST(type_prop, variable_comparison) {
     ASSERT_FALSE(variable1->get_info() == variable4->get_info());
     ASSERT_FALSE(variable1->get_info() == variable5->get_info());
 }
+
+TEST(type_prop, assign_v6_static_shape_match) {
+    auto input = make_shared<ov::op::v0::Parameter>(element::f32, Shape{1, 2, 64, 64});
+    auto variable = std::make_shared<op::util::Variable>(
+        op::util::VariableInfo{PartialShape{1, 2, 64, 64}, element::f32, "variable_id"});
+    std::shared_ptr<ov::op::v6::Assign> assign;
+    EXPECT_NO_THROW(assign = std::make_shared<ov::op::v6::Assign>(input, variable));
```
praasz, line 88:
```suggestion
    auto assign = std::make_shared<ov::op::v6::Assign>(input, variable);
```
I think there is no much benefit of this assertion as there is assumption the the operator should be created with success if there will be error test fill fail anyway.

src/core/tests/type_prop/assign.cpp
```diff
@@ -78,3 +79,99 @@ TEST(type_prop, variable_comparison) {
...
+    std::shared_ptr<ov::op::v6::Assign> assign;
+    EXPECT_ANY_THROW(assign = std::make_shared<ov::op::v6::Assign>(input, variable));
+}
+
+TEST(type_prop, assign_v6_static_types_do_not_match) {
+    auto input = make_shared<ov::op::v0::Parameter>(element::i32, Shape{1, 2, 64, 64});
+    auto variable_info = op::util::VariableInfo{PartialShape{1, 2, 64, 64}, element::f32, "variable_id"};
+    auto variable = std::make_shared<op::util::Variable>(variable_info);
+    std::shared_ptr<ov::op::v6::Assign> assign;
+    EXPECT_ANY_THROW(assign = std::make_shared<ov::op::v6::Assign>(input, variable));
```
praasz, line 107:
```suggestion
    OV_EXPECT_THROW(std::ignore= std::make_shared<ov::op::v6::Assign>(input, variable), exp_exception, testing::HasSubstr(""));
```
As alternative this helper macro can be used from `test_asssertion.hpp` to check if expected exception has been thrown and also, optionally message of exception can be checked.

src/core/tests/type_prop/assign.cpp
```diff
@@ -78,3 +79,99 @@ TEST(type_prop, variable_comparison) {
...
+    auto variable = std::make_shared<op::util::Variable>(
+        op::util::VariableInfo{PartialShape{1, 2, 64, 64}, element::f32, "variable_id"});
+    std::shared_ptr<ov::op::v6::Assign> assign;
+    EXPECT_NO_THROW(assign = std::make_shared<ov::op::v6::Assign>(input, variable));
+
+    ASSERT_EQ(assign->get_element_type(), element::f32);
+    ASSERT_EQ(assign->get_shape(), (Shape{1, 2, 64, 64}));
+}
+
+TEST(type_prop, assign_v6_static_shapes_do_not_match) {
```
praasz, line 94: Add test to check if dimensions labels are propagated or not from inputs to outputs?

src/core/src/op/read_value.cpp
```diff
@@ -50,36 +55,70 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+        return make_shared<ReadValue>(new_args.at(0), m_variable);
+    } else if (new_args.empty()) {
+        return make_shared<ReadValue>(new_args.at(0), m_variable);
```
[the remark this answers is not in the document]
slyalin: This bug makes the whole feature not usable in LLMs...

src/core/tests/type_prop/read_value.cpp
```diff
@@ -15,4 +15,84 @@ TEST(type_prop, read_value_deduce) {
...
+}
+
+TEST(type_prop, read_value_v6_init_shape_is_not_in_range_2) {
+    auto input = make_shared<ov::op::v0::Parameter>(element::f32, PartialShape{{1, 2}, 2, 64, 64});
+
+    auto variable_info = op::util::VariableInfo{PartialShape{1, 2, 64, 64}, element::f32, "variable_id"};
+    auto variable = std::make_shared<op::util::Variable>(variable_info);
+
+    std::shared_ptr<ov::op::v6::ReadValue> read_value;
+    EXPECT_ANY_THROW(read_value = std::make_shared<ov::op::v6::ReadValue>(input, variable));
```
slyalin, line 97: I don't see any tests for a newly created ctor with only variable id and without initialization node. Please cover this case.

itikhono: done

src/core/src/op/assign.cpp
```diff
@@ -66,9 +66,26 @@ Assign::Assign(const Output<Node>& new_value, const std::shared_ptr<util::Variab
...
 void Assign::validate_and_infer_types() {
     OV_OP_SCOPE(v6_Assign_validate_and_infer_types);
-    m_variable->update({get_input_partial_shape(0), get_input_element_type(0), m_variable->get_info().variable_id});
-
-    set_output_type(0, get_input_element_type(0), get_input_partial_shape(0));
+    auto variable_info = m_variable->get_info();
+    auto variable_type = variable_info.data_type;
+    auto variable_shape = variable_info.data_shape;
+
+    auto input_type = get_input_element_type(0);
```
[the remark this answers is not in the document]
itikhono: done

src/core/src/op/read_value.cpp
```diff
@@ -50,36 +55,70 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+    // If no inputs provided, it means this ReadValue doesn't have initial subgraph. This is valid.
+    if (get_input_size() > 0) {
+        const auto initial_type = get_input_element_type(0);
+        const auto& initial_shape = get_input_partial_shape(0);
+
+        // Variable's shape/type determine a permissible range of values for shape/type inferred from initial_subgraph.
+        // If initial_subgraph is set, then we need to check that shape/type inferred from initial_subgraph
+        // is within the permissible range.
+
+        bool compatible_type = variable_type.is_dynamic() || initial_type == variable_type;
```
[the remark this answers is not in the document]
itikhono: done

src/core/src/op/read_value.cpp
```diff
@@ -50,36 +55,70 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+
+    set_output_type(0, variable_type, variable_shape);
 }
 
 shared_ptr<Node> op::v6::ReadValue::clone_with_new_inputs(const OutputVector& new_args) const {
     OV_OP_SCOPE(v6_ReadValue_clone_with_new_inputs);
     check_new_args_count(this, new_args);
-    return make_shared<ReadValue>(new_args.at(0), m_variable);
+    if (new_args.size() == 1) {
+        return make_shared<ReadValue>(new_args.at(0), m_variable);
```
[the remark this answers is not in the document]
itikhono: done

src/core/src/op/read_value.cpp
itikhono: done

src/core/src/pass/low_latency.cpp
itikhono: double-checked, it doesn't work

src/bindings/python/src/pyopenvino/graph/dict_attribute_visitor.cpp
itikhono: done

src/core/src/op/read_value.cpp
```diff
@@ -50,36 +55,70 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+    } else if (new_args.empty()) {
+        return make_shared<ReadValue>(new_args.at(0), m_variable);
+    }
+    OPENVINO_ASSERT(false,
```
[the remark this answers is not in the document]
itikhono: done

src/core/src/op/read_value.cpp
```diff
@@ -42,6 +42,11 @@ bool op::v3::ReadValue::visit_attributes(AttributeVisitor& visitor) {
     return true;
 }
 
+op::v6::ReadValue::ReadValue(const shared_ptr<Variable>& variable) {
+    m_variable = variable;
```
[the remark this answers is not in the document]
itikhono: the current hierarchy looks like 
ReadValue <- ReadValueBase <- VariableExtension (protected: m_variable)
we can't use initializer list in this situation.
it's possible to introduce an additional ctor for ReadValueBase, it's better to implement it in the next PR

src/bindings/python/src/pyopenvino/graph/dict_attribute_visitor.cpp
itikhono: done

src/core/tests/type_prop/read_value.cpp
```diff
@@ -105,4 +106,13 @@ TEST(type_prop, read_value_v6_no_init) {
     ASSERT_EQ(read_value->get_element_type(), element::f32);
     ASSERT_EQ(read_value->get_output_partial_shape(0), (PartialShape{1, 2, 64, 64}));
     ASSERT_EQ(read_value->get_variable_id(), "variable_id");
+}
+
+TEST(type_prop, DISABLED_read_value_labels_propagation) {
+    auto variable_pshape = PartialShape{1, 2, 64, 64};
+    auto variable_info = op::util::VariableInfo{variable_pshape, element::f32, "variable_id"};
+    auto variable = std::make_shared<op::util::Variable>(variable_info);
+    set_shape_labels(variable_pshape, 10);
```
slyalin, line 115: Move it before variable_info.

itikhono: fixed

src/core/src/op/read_value.cpp
```diff
@@ -50,36 +55,72 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+
+        // Variable's shape/type determine a permissible range of values for shape/type inferred from initial_subgraph.
+        // If initial_subgraph is set, then we need to check that shape/type inferred from initial_subgraph
+        // is within the permissible range.
+
+        const auto compatible_type = variable_type.is_dynamic() || initial_type == variable_type;
+        const auto compatible_shape = variable_shape.relaxes(initial_shape);
+
+        OPENVINO_ASSERT(compatible_shape,
+                        "The shape specified in the Variable have to extend (relax) the shape "
```
ceciliapeng2011, line 77:
```suggestion
                        "The shape specified in the Variable has to extend (relax) the shape "
```

src/bindings/python/src/pyopenvino/graph/model.cpp
```diff
@@ -60,6 +60,23 @@ static std::vector<std::shared_ptr<ov::Node>> cast_to_node_vector(const ov::Sink
     return nodes;
 }
 
+// Assign operations created via Assign py binding have Variables which are not connected to
+// ReadValue operations. This function attempts to resolve this situation by finding correct Variables
+// for Assigns.
```
ilya-lavrenov, line 65: why is it done on python bindings level? is it python specific?

itikhono: Assign/ReadValue pair has to be initialized with the same obj of Variable class, this a requirement.
The previous solution is a` map <string, shared_ptr<Variable>> ` inside NodeFactory class. This map allows to get Variable by name, if it exists, or stores a created Variable, if it doesn't. So it helps us to use the same Variable obj for Assign/ReadValue classes.
But a separate py binding was created for Assign class couple Releases ago, it means that we use the class constructor to create Assign operation instead of NodeFactory. It broke the logic above and now it's impossible to create a correct ov::Model with Assign/ReadValue via py api.

itikhono: It was decided to use Model class binding because Assign binding doesn't have access to the model and we can' t get all Variables on this level.

src/core/src/op/read_value.cpp
```diff
@@ -116,6 +116,18 @@ bool op::v6::ReadValue::visit_attributes(AttributeVisitor& visitor) {
     auto variable_info = m_variable->get_info();
     visitor.on_attribute("variable_type", variable_info.data_type);
     visitor.on_attribute("variable_shape", variable_info.data_shape);
+
+    // workaround:
+    // dynamic rank/type can be derived from the IRs generated via the prev versions of OV,
+    // but dynamic rank/type are not supported in plugins,
+    // so we are trying to fix them here using the rank/type of ReadValue 1st input, if it exists
+    if (get_input_size() > 0 && variable_info.data_shape.rank().is_dynamic() && variable_info.data_type.is_dynamic()) {
```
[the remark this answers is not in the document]
slyalin: Apply reverse type/shape propagation?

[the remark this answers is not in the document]
slyalin: Difficulties of this kind is a reason to introduce a new version of ReadValue, but a new version will provide a more precise shape inference only, so upgrading the version for this only reason looks to be an overkill.

ceciliapeng2011:
@itikhono Would you please comment to 
https://github.com/openvinotoolkit/openvino/pull/20368#discussion_r1392120938 ? It is not clear to me when a State object is constructed with a initial state which is a subgraph (or simply a Parameter).

src/core/src/op/read_value.cpp
```diff
@@ -50,36 +55,84 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+    }
 }
 
 bool op::v6::ReadValue::visit_attributes(AttributeVisitor& visitor) {
     OV_OP_SCOPE(v6_ReadValue_visit_attributes);
     visitor.on_attribute("variable_id", m_variable);
+
+    auto variable_info = m_variable->get_info();
+    visitor.on_attribute("variable_type", variable_info.data_type);
+    visitor.on_attribute("variable_shape", variable_info.data_shape);
```
ceciliapeng2011, line 118: Can we have the following interfaces for ReadValue? 
`get_variable_type` and `get_variable_shape`? Then plugin can retrieve them easily.

ceciliapeng2011: @ilya-lavrenov @ilyachur @slyalin @itikhono We need to clarify a problem for stateful implementation & optimization in plugin -
1. The variable has a data type. 
2. The ngraph node ReadValue has an output data type. 
 
What's their relationship of these types? Do we expect them to be the same type? If not, which is the data type a user is assuming? Which data type should the plugin assume it should return this type as it is asked by user?

src/core/src/op/read_value.cpp
```diff
@@ -50,36 +55,84 @@ op::v6::ReadValue::ReadValue(const Output<Node>& init_value, const shared_ptr<Va
...
+        OPENVINO_ASSERT(compatible_type,
+                        "The type specified in the Variable has to extend (relax) the type "
+                        "inferred from the initializing subgraph.",
+                        " Variable type: ",
+                        variable_type,
+                        " Initialization type: ",
+                        initial_type);
+    }
+
+    set_output_type(0, variable_type, variable_shape);
```
ceciliapeng2011, line 92: @maxnick    Here ReadValue infers its shape and datatype from variable now, instead of input prot 0 or original ngraph node precision (with ppp we can change the node precision here).

maxnick: Yes, we discussed it with @itikhono, so this change is expected.

evkotov (approved):

src/bindings/python/tests/test_runtime/test_infer_request.py
```diff
@@ -678,15 +680,14 @@ def test_query_state_write_buffer(device, input_shape, data_type, mode):
     from openvino.runtime.utils.types import get_dtype
 
     model = create_model_with_memory(input_shape, data_type)
+    model.validate_nodes_and_infer_types()
```
jiwaszki, line 683: My only concern: why is this call needed here? Is it going to be needed for all cases when model has `read_value` as operator?

jiwaszki (commented):
Overall looks good, only one concern about usage.

akuporos (approved):

src/core/src/op/assign.cpp
itikhono: I will move them in the next PR

src/bindings/python/tests/test_runtime/test_infer_request.py
itikhono: This method checks the model consistency and it throwed an exception without the fixes.
So I've just added it be sure that it passes.
Maybe we need to add this call for other tests as well.

itikhono:
> @itikhono, what is ETA for merge?



> @itikhono, what is ETA for merge?

today or tomorrow
CPU part has already been merged, the issues with integration were fixed. Waiting for CI.

Could you review the latest changes in this PR please?

slyalin (approved):

src/bindings/python/src/openvino/runtime/opset6/ops.py
```diff
@@ -124,17 +124,66 @@ def mvn(
...
 
 
+@singledispatch
 @nameable_op
-def read_value(init_value: NodeInput, variable_id: str, name: Optional[str] = None) -> Node:
+def read_value(init_value: NodeInput,
+               variable_id: str,
+               variable_type: Optional[Union[NumericType, Type, str]] = None,
+               variable_shape: Optional[TensorShape] = None,
+               name: Optional[str] = None) -> Node:
```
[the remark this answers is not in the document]
akuporos: @itikhono from the point of API compatibility it would be better to have `name` before added arguments.
From the other hand it looks more logical how it is done right now.
If it isn't critical, I vote for keeping it as is.

PR #20413, opened 2023-10-12
[CPU] Enabled custom MHA fusion for BF16 and INT8+FP32 cases

a-sidorova:
### Details:
 - *Some MHA patterns were fused into custom MHA node but after Snippets MHA tokenization enabling for BF16, INT8 models these subgraphs aren't tokenized into single node because of heuristic conditions: if there are big shapes, Snippets don't guarantee performance because of BRGEMM blocking for BF16 and INT8 is not implemented . The PR returns custom MHA fusion for these subgraphs only for BF16 and INT8+FP32 cases: FP32 case is faster on Snippets (due to blocking support) and INT8+BF16 case is supported only by Snippets(Custom MHA supports only INT8+FP32 that slower). Seems like custom MHA is matched only on models with small shapes that we don't have performance regressions*

### Tickets:
 - *N/A*


a-sidorova:
Closed since the feature is used locally (not in the production)

PR #20416, opened 2023-10-12
[core]Migrate VariadicSplit op to new API

praasz:
### Details:
 - Migrate VariadicSplit to new API, remove legacy headers, namespaces, classes.
 - Refactor VariadicSplit to reduce binary size.
 - Remove private member functions as can be internal implementation only.

### Tickets:
 - [CVS-118684](https://jira.devtools.intel.com/browse/CVS-118684)


src/core/src/op/variadic_split.cpp
```diff
@@ -2,33 +2,69 @@
...
+                             static_cast<char*>(output.data()),
+                             data_tensor.get_shape(),
+                             lower_bounds,
+                             upper_bounds,
+                             default_strides,
+                             out_shape,
+                             data_tensor.get_element_type().size());
+        }
+
+        lower_bounds[axis] = upper_bounds[axis];
```
t-jankowski, line 53: Is this assignment needed in every iteration?

t-jankowski (approved):

src/core/src/op/variadic_split.cpp
praasz: Yes is like  `new start` for next iteration.

mitruska (approved):

PR #20431, opened 2023-10-12
[PT FE] Implement custom op for types alignment

mvafin:
### Details:
 - *Implement special operation for types alignment*

### Tickets:
 - *122693*


src/core/include/openvino/op/util/framework_node.hpp
```diff
@@ -81,7 +81,7 @@ class OPENVINO_API FrameworkNode : public MultiSubGraphOp {
 
     explicit FrameworkNode(const OutputVector& inputs, size_t output_size = 1, size_t num_subgraphs = 0);
```
rkazants, line 83: Ant test or new model to enable by this implementation?

mvafin: This change should be tested by existing layer tests for types

src/bindings/python/src/openvino/runtime/utils/types.py
```diff
@@ -23,6 +23,7 @@
 
 openvino_to_numpy_types_map = [
     (Type.boolean, bool),
+    (Type.boolean, np.bool_),
```
mvafin, line 26: This change is needed for `get_element_type` to work on boolean numpy tensors

src/frontends/pytorch/src/helper_ops/internal_op.hpp
```diff
@@ -0,0 +1,56 @@
...
+    const size_t m_num_outputs;
+};
+
+class InternalOperation : public PtFrameworkNode {
+protected:
+    InternalOperation(const std::string& op_type,
+                      const OutputVector& inputs,
+                      size_t num_outputs,
+                      const std::string& no_conversion_reason)
+        : PtFrameworkNode(std::make_shared<InternalOpDecoder>(op_type, num_outputs), inputs) {
```
slyalin, line 48: Can we just avoid passing a decoder to such ops? We don't have anything in the original graph that corresponds to such ops, why do we need a decoder in this case?

src/frontends/pytorch/src/helper_ops/internal_op.hpp
```diff
@@ -0,0 +1,56 @@
...
+        return m_decoder_type;
+    }
+
+private:
+    const std::string m_op_type;
+    const std::string m_decoder_type = "internal_op";
+    const size_t m_num_outputs;
+};
+
+class InternalOperation : public PtFrameworkNode {
```
slyalin, line 42: `InternalOp` doesn't always correspond to a node in the original graph, why should we derive it from `PtFrameworkNode`?

slyalin (changes requested):

src/frontends/pytorch/src/helper_ops/internal_op.hpp
mvafin: To avoid issues with reporting such operations when they are left in graph, otherwise that reporting should be reworked, which is out of scope of these changes

src/frontends/pytorch/src/helper_ops/internal_op.hpp
mvafin: Since we derive from `PtFrameworkNode` we have to provide decoder.

src/bindings/python/src/openvino/runtime/utils/types.py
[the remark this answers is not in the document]
mvafin: We have a search by second value, so that should be fine, we have both `bool` and `np.bool_` mapped to `Type.boolean`

jiwaszki (approved):

slyalin (approved):

PR #20459, opened 2023-10-13
SDPA Softmax Temp. Perf. Fix

cavusmustafa:
- Temporary solution Softmax performance issue which is part of SDPA decomposition
- Converts 4D input of Softmax into a 3D input


suryasidd:
build_jenkins

src/frontends/pytorch/src/op/scaled_dot_product_attention.cpp
```diff
@@ -15,13 +15,15 @@
...
 #include "openvino/op/range.hpp"
+#include "openvino/op/reshape.hpp"
 #include "openvino/op/select.hpp"
 #include "openvino/op/shape_of.hpp"
 #include "openvino/op/softmax.hpp"
 #include "openvino/op/sqrt.hpp"
 #include "openvino/op/squeeze.hpp"
 #include "openvino/op/transpose.hpp"
 #include "openvino/op/unsqueeze.hpp"
+#include "openvino/op/util/framework_node.hpp"
```
mvafin, line 26: why this include is needed?

src/frontends/pytorch/src/op/scaled_dot_product_attention.cpp
```diff
@@ -99,11 +101,43 @@ OutputVector translate_scaled_dot_product_attention(const NodeContext& context)
...
+    if (!(scaled_atten->is_dynamic()) && scaled_atten->get_output_partial_shape(0).rank() == 4) {
+        const std::vector<size_t>* scaled_atten_v = &(scaled_atten->get_shape());
+        std::vector<size_t> softmax_shape_v((*scaled_atten_v).begin() + 1, (*scaled_atten_v).end());
+        softmax_shape_v[0] *= (*scaled_atten_v)[0];
+        std::vector<size_t> matmul_shape_v(*scaled_atten_v);
+        auto softmax_shape = context.mark_node(v0::Constant::create(element::i32, Shape{3}, softmax_shape_v));
+        auto reshape_3d = context.mark_node(std::make_shared<v1::Reshape>(scaled_atten, softmax_shape, false));
+        scaled_atten = context.mark_node(std::make_shared<v8::Softmax>(reshape_3d, -1));
+        auto matmul_shape = context.mark_node(v0::Constant::create(element::i64, Shape{4}, matmul_shape_v));
+        scaled_atten = context.mark_node(std::make_shared<v1::Reshape>(scaled_atten, matmul_shape, false));
```
mvafin, line 113:
```suggestion
    auto satten_out_pshape = scaled_atten->get_output_partial_shape(0);
    if (satten_out_pshape.is_static() && satten_out_pshape.rank().get_length() == 4) {
        auto scaled_atten_v = satten_out_pshape.to_shape();
        Shape softmax_shape_v(scaled_atten_v.begin() + 1, scaled_atten_v.end());
        softmax_shape_v[0] *= scaled_atten_v[0];
        auto softmax_shape = context.mark_node(v0::Constant::create(element::i32, Shape{3}, softmax_shape_v));
        auto reshape_3d = context.mark_node(std::make_shared<v1::Reshape>(scaled_atten, softmax_shape, false));
        scaled_atten = context.mark_node(std::make_shared<v8::Softmax>(reshape_3d, -1));
        auto matmul_shape = context.mark_node(v0::Constant::create(element::i32, Shape{4}, scaled_atten_v));
        scaled_atten = context.mark_node(std::make_shared<v1::Reshape>(scaled_atten, matmul_shape, false));
```
That should work, but I didn't check

mvafin (changes requested):

src/frontends/pytorch/src/op/scaled_dot_product_attention.cpp
cavusmustafa: Seems to be working as expected

src/frontends/pytorch/src/op/scaled_dot_product_attention.cpp
mvafin: But it should be already included in `utils.hpp`

cavusmustafa: I just pushed an update to include this in utils.hpp

mvafin:
build_jenkins

slyalin (changes requested):
Please, hold. We are moving this part to another place.

Also the CPU part is likely very sensitive to any changes in the decomposition. @jane-intel, please comment.

jane-intel:
Per e-mail thread: GPU plugin will address this performance insufficiency on their side. Since this PR changes model for all the plugins and there is no performance data for other plugins on other models -- I suggest to close this PR and wait for the fix from GPU plugin side.

Cc: @vladimir-paramuzov @p-durandin

p-durandin (changes requested):
Perf pre tests on CPU and GPU required

cavusmustafa:
@jane-intel @p-durandin @mvafin @slyalin 

I agree that a proper fix should be implemented on GPU plugin side, do you have any ETA for it? The performance regression on Stable Diffusion impacts our A1111 integration which is used by Tom's Hardware's performance tests. The reason we want this PR to be merged as soon as possible and consider this as a critical fix is to prevent the performance regression to appear on the press. If the GPU fix will not be available for the next nightly build, I would suggest including this change as a temporary fix.

I will collect some performance test results for Stable Diffusion UNET model on CPU and GPU.

PR #20462, opened 2023-10-14
[ARM CPU] Some fixes for behavior tests (fp16)

allnes:
### Details:
 - *item1*
 - *...*

### Tickets:
 - *ticket-id*


src/plugins/intel_cpu/tests/functional/behavior/ov_plugin/properties.cpp
```diff
@@ -173,13 +173,18 @@ TEST_F(OVClassConfigTestCPU, smoke_PluginSetConfigAffinityCore) {
     ASSERT_EQ(false, value);
 }
 
+#if defined(OV_CPU_ARM_ENABLE_FP16)
+const auto expected_precision_for_performance_mode = ov::element::f16;
```
EgorDuplensky, line 177: From my understanding this should fail with the current master, since we use f32 precision by default on ARM at the moment. Let's wait for the corresponding PR to be merged.

PR #20472, opened 2023-10-15
choose Pcore to compile model for GPU plugin

xufang-lisa:
### Details:
 - *choose Pcore to compile model for GPU plugin*

### Tickets:
 - *122426*
 - *121161*


wangleis (approved):

src/plugins/intel_gpu/src/graph/program.cpp
```diff
@@ -122,6 +123,31 @@ static void adjust_num_cores(ov::threading::IStreamsExecutor::Config& config) {
     }
 
     config._streams = std::min(config._streams, num_cores);
+
+    //create stream_info_table based on core type
+    const auto proc_type_table = ov::get_org_proc_type_table();
+    std::vector<int> stream_info(ov::CPU_STREAMS_TABLE_SIZE, 0);
+    stream_info[ov::THREADS_PER_STREAM] = 1;
+    stream_info[ov::STREAM_NUMA_NODE_ID] = 0;
+    stream_info[ov::STREAM_SOCKET_ID] = 0;
```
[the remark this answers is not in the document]
wangleis: @vladimir-paramuzov May I know if you expect IStreamExecutor provide simple interface like `make_task_executor_config(number_of_streams, number_of_threads_per_stream, cpu_core_type)`?

sunxiaoxia2022 (approved):

src/plugins/intel_gpu/src/graph/program.cpp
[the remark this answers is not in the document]
vladimir-paramuzov: @wangleis yes, something like this. If such config is not needed anywhere else, then I'm OK with current change, but I have a feeling that this code may be useful in other components too.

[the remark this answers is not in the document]
wangleis: @vladimir-paramuzov Understood. With streams_info_table, HW plugin have full feasibility to create any kind of stream on any CPU core. Meanwhile, IStreamExecutor will also provide simple API for most of normal cases, which will be easy to use in HW plugin.

yeonbok:
I have some questions
1) Does this change intends to optimize for processor binding only?
i.e., if the binidng is ANY, then no impact? 
i.e., user (including optimum) should apply model_priority to set the core binding?  (Is that planned too alongside this change??)
2) Is the baseline assumption same for both ADL and MTL?

src/plugins/intel_gpu/src/graph/program.cpp
```diff
@@ -104,39 +105,39 @@
...
+    case ov::hint::Priority::MEDIUM:
+        task_executor_config = ov::threading::IStreamsExecutor::Config::update_executor_config(
+            task_executor_config,
+            task_executor_config._streams,
+            1,
+            ov::threading::IStreamsExecutor::Config::ANY,
+            false);
+        break;
+    case ov::hint::Priority::HIGH:
+        task_executor_config = ov::threading::IStreamsExecutor::Config::update_executor_config(
```
vladimir-paramuzov, line 130: I think this part can be refactored to avoid code duplication:
```cpp
auto preferred_core_type = ov::threading::IStreamsExecutor::Config::ANY;
switch (priority) {
        case ov::hint::Priority::LOW: preferred_core_type = ov::threading::IStreamsExecutor::Config::LITTLE; break;
        case ov::hint::Priority::MEDIUM: preferred_core_type = ov::threading::IStreamsExecutor::Config::ANY; break;
        case ov::hint::Priority::HIGH: preferred_core_type = ov::threading::IStreamsExecutor::Config::BIG; break;
        default: OPENVINO_ASSERT(false, "[GPU] Can't create task executor: invalid host task priority value: ", priority);
}
task_executor_config = ov::threading::IStreamsExecutor::Config::update_executor_config(
    task_executor_config,
    task_executor_config._streams,
    1,
    preferred_core_type,
    false);
```

xufang-lisa: thanks, updated

src/inference/dev_api/openvino/runtime/threading/istreams_executor.hpp
```diff
@@ -150,6 +150,12 @@ class OPENVINO_RUNTIME_API IStreamsExecutor : virtual public ITaskExecutor {
               _threadPreferredCoreType(threadPreferredCoreType),
               _streams_info_table{streamsInfoTable},
               _cpu_reservation{cpuReservation} {}
+
+        static Config update_executor_config(const Config& initial,
```
vladimir-paramuzov, line 154: Why is it static method? It makes a copy of the config, and I don't really see any reason to do it.

src/inference/dev_api/openvino/runtime/threading/istreams_executor.hpp
```diff
@@ -150,6 +150,12 @@ class OPENVINO_RUNTIME_API IStreamsExecutor : virtual public ITaskExecutor {
               _threadPreferredCoreType(threadPreferredCoreType),
               _streams_info_table{streamsInfoTable},
               _cpu_reservation{cpuReservation} {}
+
```
vladimir-paramuzov, line 153: Please add a description of this method. It's not obvious how exactly the config is updated

src/inference/src/dev/threading/istreams_executor.cpp
```diff
@@ -553,5 +553,109 @@ IStreamsExecutor::Config IStreamsExecutor::Config::reserve_cpu_threads(const ISt
     return config;
 }
 
+IStreamsExecutor::Config IStreamsExecutor::Config::update_executor_config(
```
wangleis, line 556: Please add smoke test cases for this feature

src/inference/dev_api/openvino/runtime/threading/istreams_executor.hpp
xufang-lisa: updated

src/inference/dev_api/openvino/runtime/threading/istreams_executor.hpp
xufang-lisa: Static function is not necessary, it has been changed to non-static function.

src/inference/src/dev/threading/istreams_executor.cpp
xufang-lisa: updated

vladimir-paramuzov (approved):

src/plugins/intel_gpu/src/graph/program.cpp
```diff
@@ -135,7 +116,10 @@ static ov::threading::IStreamsExecutor::Config make_task_executor_config(const E
         default: OPENVINO_ASSERT(false, "[GPU] Can't create task executor: invalid host task priority value: ", priority);
     }
 
-    adjust_num_cores(task_executor_config);
+    task_executor_config.update_executor_config(task_executor_config._streams,
+                                                1,
+                                                task_executor_config._threadPreferredCoreType,
+                                                false);
```
ilya-lavrenov, line 122: why do we need `update` method? can all required valid values be set via class ctor?

